PR #206, opened 2025-12-16
refactor: replace genAI model sdk by okHttp

### LINKED ISSUE #96: [Refactor] Use okhttp to replace GenAI Model SDK

Aias00:
## AgentScope-Java Version

[The version of AgentScope-Java you are working on, e.g. 1.0.2, check your pom.xml dependency version or run `mvn dependency:tree | grep agentscope-parent:pom`(only mac/linux)]

## Description

 Introduce dedicated DTOs for Gemini API request and response structures and update related components.
fixes #96 
## Checklist


agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiMessageConverter.java
```diff
@@ -80,63 +66,50 @@ public GeminiMessageConverter() {
...
+                            new GeminiFunctionCall(tub.getId(), tub.getName(), tub.getInput());
+
+                    // Build Part
+                    GeminiPart part = new GeminiPart();
+                    part.setFunctionCall(functionCall);
+
+                    // Note: Thought signature currently not directly supported in simple DTOs
+                    // unless we add it
+                    // The SDK supported it, but it might be an internal detail.
+                    // If needed, we can add it to GeminiPart DTO later.
```
AlbumenJ: `Thought signature` is really important for Gemini-3-pro

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiResponseParser.java
```diff
@@ -69,42 +57,47 @@ public GeminiResponseParser() {
...
+                        metadata.getPromptTokenCount() != null ? metadata.getPromptTokenCount() : 0;
+                int totalOutputTokens =
+                        metadata.getCandidatesTokenCount() != null
+                                ? metadata.getCandidatesTokenCount()
+                                : 0;
+
+                // Note: thinking tokens field might not be in generic UsageMetadata unless we
+                // add it
+                // Assuming it's not crucial or we add it to DTO if needed.
+                // For now, use totalOutputTokens.
```
AlbumenJ: Please fix it

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiResponseParser.java
```diff
@@ -117,7 +110,7 @@ public ChatResponse parseResponse(GenerateContentResponse response, Instant star
             }
 
             return ChatResponse.builder()
-                    .id(response.responseId().orElse(null))
+                    // Response ID is not always present in simple JSON or might be different key
```
AlbumenJ: Should support it

AlbumenJ (changes requested):

Aias00:
gemini-2.0-flush
截屏2025-12-17 09 52 54

gemini-3-pro-preview
截屏2025-12-17 09 57 20

AlbumenJ:
image

Please make sure all the E2E Tests pass

AlbumenJ:
Your current logic for handling the Base URL differs from the GenAI SDK's approach, specifically regarding the handling of the `GOOGLE_API_BASE_URL` key.

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiChatFormatter.java
```diff
@@ -48,12 +53,12 @@
...
  */
 public class GeminiChatFormatter
-        extends AbstractBaseFormatter<
-                Content, GenerateContentResponse, GenerateContentConfig.Builder> {
+        extends AbstractBaseFormatter<GeminiContent, GeminiResponse, GeminiRequest> {
 
     private final GeminiMessageConverter messageConverter;
     private final GeminiResponseParser responseParser;
     private final GeminiToolsHelper toolsHelper;
+    private GeminiContent systemInstruction;
```
JGoP-L: The Formatter itself is required to be stateless, yet the instance variable systemInstruction stores temporary state, which leads to race conditions when the instance is shared across multiple threads

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiResponseParser.java
```diff
@@ -69,51 +69,109 @@ public GeminiResponseParser() {}
      * @param startTime Request start time for calculating duration
      * @return AgentScope ChatResponse
      */
-    public ChatResponse parseResponse(GenerateContentResponse response, Instant startTime) {
+    public ChatResponse parseResponse(GeminiResponse response, Instant startTime) {
         try {
+            // Log raw response for debugging
+            try {
+                String responseJson = JsonUtils.getJsonCodec().toJson(response);
+                System.out.println("=== Raw Gemini response: " + responseJson);
```
JGoP-L: The raw API responses are printed in the production environment

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiResponseParser.java
```diff
@@ -129,49 +187,88 @@ public ChatResponse parseResponse(GenerateContentResponse response, Instant star
...
-            if (part.thought().isPresent() && part.thought().get() && part.text().isPresent()) {
-                String thinkingText = part.text().get();
-                if (thinkingText != null && !thinkingText.isEmpty()) {
-                    blocks.add(ThinkingBlock.builder().thinking(thinkingText).build());
+    protected void parsePartsToBlocks(List<GeminiPart> parts, List<ContentBlock> blocks) {
+        // Debug: Log the parts received from Gemini
+        if (org.slf4j.LoggerFactory.getLogger(this.getClass()).isDebugEnabled()) {
+            try {
+                org.slf4j.Logger log = org.slf4j.LoggerFactory.getLogger(this.getClass());
+                log.debug("=== Parsing {} parts from Gemini response", parts.size());
```
JGoP-L: Code redundancy (the private static final Logger log has already been defined at the top of the class)

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiMessageConverter.java
```diff
@@ -112,51 +114,56 @@ public List<Content> convertMessages(List<Msg> msgs) {
...
+                                tub.getMetadata().get(ToolUseBlock.METADATA_THOUGHT_SIGNATURE);
+                        if (thoughtSig instanceof String) {
+                            part.setThoughtSignature((String) thoughtSig);
                         }
                     }
 
-                    parts.add(partBuilder.build());
+                    parts.add(part);
+
+                } else if (block instanceof ThinkingBlock tb) {
```
JGoP-L: Can the following else if (block instanceof ThinkingBlock) (line 181) be removed?

Aias00: why remove?

JGoP-L: > why remove?

Since line 137 has already been executed, this check on line 181 will never be triggered

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiMessageConverter.java
```diff
@@ -112,51 +114,42 @@ public List<Content> convertMessages(List<Msg> msgs) {
...
+                    GeminiPart part = new GeminiPart();
+                    part.setFunctionCall(functionCall);
+
+                    // Restore thoughtSignature from metadata if present (required for Gemini 2.5+)
+                    if (tub.getMetadata() != null
+                            && tub.getMetadata()
+                                    .containsKey(ToolUseBlock.METADATA_THOUGHT_SIGNATURE)) {
+                        Object thoughtSig =
+                                tub.getMetadata().get(ToolUseBlock.METADATA_THOUGHT_SIGNATURE);
+                        if (thoughtSig instanceof String) {
```
AlbumenJ, line 161: Should also support byte[] here for backward-compability

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiResponseParser.java
```diff
@@ -69,51 +69,108 @@ public GeminiResponseParser() {}
...
                                 .build();
             }
 
             return ChatResponse.builder()
-                    .id(response.responseId().orElse(null))
+                    // Use actual response ID if available, otherwise generate one
+                    .id(
+                            response.getResponseId() != null
+                                    ? response.getResponseId()
+                                    : java.util.UUID.randomUUID().toString())
```
AlbumenJ: Replace FQN here

agentscope-core/src/main/java/io/agentscope/core/model/GeminiChatModel.java
```diff
@@ -55,109 +70,79 @@
...
-    private final HttpOptions httpOptions;
-    private final GoogleCredentials credentials;
-    private final ClientOptions clientOptions;
-    private final Client client;
     private final GenerateOptions defaultOptions;
-    private final Formatter<Content, GenerateContentResponse, GenerateContentConfig.Builder>
-            formatter;
+    private final Formatter<GeminiContent, GeminiResponse, GeminiRequest> formatter;
+    private final OkHttpClient httpClient;
+    private final ObjectMapper objectMapper;
```
AlbumenJ: Use `JsonUtils` instead

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiToolsHelper.java
```diff
@@ -63,34 +59,51 @@ public GeminiToolsHelper() {}
...
-                    builder.parameters(schema);
+                    // Clean schema to remove Gemini-incompatible fields
+                    Map<String, Object> cleanedParams =
+                            cleanSchemaForGemini(toolSchema.getParameters());
+                    declaration.setParameters(cleanedParams);
+
+                    // Debug: Log the cleaned schema
+                    try {
+                        String schemaJson =
+                                new com.fasterxml.jackson.databind.ObjectMapper()
```
AlbumenJ: Use `JsonUtils` instead

AlbumenJ (changes requested):

Aias00:
image

LearningGp:
Considering compatibility and the principle of minimizing dependencies, it’s better to switch to the JDK HTTP transport.

Aias00:
> Considering compatibility and the principle of minimizing dependencies, it’s better to switch to the JDK HTTP transport.

It’s like this — let me explain first: the original task in this issue was to replace the generated SDK with OkHttp.

I can handle the work of using java HTTP in a new PR.

agentscope-core/src/main/java/io/agentscope/core/model/GeminiChatModel.java
```diff
@@ -174,79 +195,456 @@ protected Flux<ChatResponse> doStream(
...
-                                                        }
-                                                    });
-                                } else {
-                                    // Use non-streaming API
-                                    GenerateContentResponse response =
-                                            client.models.generateContent(
-                                                    modelName, formattedMessages, config);
+                                // 2. Serialize Request
+                                String requestJson = jsonCodec.toJson(requestDto);
+                                log.info(
```
LearningGp: ​This should not be an INFO log. Please audit and update similar occurrences.

agentscope-core/src/main/java/io/agentscope/core/model/GeminiChatModel.java
```diff
@@ -15,24 +15,51 @@
  */
```
LearningGp, line 15: It feels like GeminiChatModel is doing too much. We could split it up—for example, by moving the structured output logic to a separate component.

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiMessageConverter.java
```diff
@@ -307,8 +396,11 @@ private Source extractSourceFromBlock(ContentBlock block) {
     /**
      * Save base64 data to a temporary file.
      *
-     * <p>The file extension is extracted from the MIME type (e.g., "audio/wav" → ".wav").
-     * The file is created with prefix "agentscope_" and will not be automatically deleted.
+     * <p>
+     * The file extension is extracted from the MIME type (e.g., "audio/wav" →
+     * ".wav").
+     * The file is created with prefix "agentscope_" and will not be automatically
```
LearningGp, line 402: Since these aren't automatically deleted, does it pose a risk of a storage leak?

Aias00: Yes, it can be deleted automatically.

The minimal change is to add one line after creating the temp file:

Path tempFile = Files.createTempFile("agentscope_", extension);
tempFile.toFile().deleteOnExit(); // delete when JVM exits
But note:

deleteOnExit() does not mean “delete right after use”; it deletes only when the process exits.
It is not ideal for long-running services (files can accumulate until shutdown).
A safer approach is:

Keep deleteOnExit() as a fallback.
Add a scheduled cleanup job to remove agentscope_* temp files older than N hours/days.

agentscope-core/src/main/java/io/agentscope/core/model/GeminiTransport.java
```diff
@@ -0,0 +1,148 @@
...
+    GeminiTransport(
+            OkHttpClient httpClient,
+            JsonCodec jsonCodec,
+            Formatter<GeminiContent, GeminiResponse, GeminiRequest> formatter) {
+        this.httpClient = httpClient;
+        this.jsonCodec = jsonCodec;
+        this.formatter = formatter;
+    }
+
+    Flux<ChatResponse> handleUnaryResponse(Request request, Instant startTime) {
```
LearningGp, line 59: The Response object in handleUnaryResponse should be managed using try-with-resources. Currently, if response.isSuccessful() is false, the Response itself remains unclosed.

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiResponseParser.java
```diff
@@ -129,49 +194,80 @@ public ChatResponse parseResponse(GenerateContentResponse response, Instant star
...
      */
     protected void parseToolCall(
-            FunctionCall functionCall, byte[] thoughtSignature, List<ContentBlock> blocks) {
+            GeminiFunctionCall functionCall, String thoughtSignature, List<ContentBlock> blocks) {
         try {
-            String id = functionCall.id().orElse("tool_call_" + System.currentTimeMillis());
-            String name = functionCall.name().orElse("");
+            String id = functionCall.getId();
+            if (id == null || id.isEmpty()) {
+                id = "tool_call_" + System.currentTimeMillis(); // Fallback if ID is missing
```
LearningGp: Potential ID duplication issue during concurrent execution. Consider using a thread-safe generator (e.g., UUID) to prevent collisions.

agentscope-core/src/main/java/io/agentscope/core/model/GeminiTransport.java
```diff
@@ -0,0 +1,148 @@
...
+            try (ResponseBody responseBody = response.body()) {
+                String bodyString = responseBody != null ? responseBody.string() : null;
+                if (!response.isSuccessful() || bodyString == null) {
+                    String errorBody = bodyString != null ? bodyString : "null";
+                    throw new GeminiApiException(response.code(), errorBody);
+                }
+
+                GeminiResponse geminiResponse =
+                        jsonCodec.fromJson(bodyString, GeminiResponse.class);
+                log.info("Gemini Response JSON: {}", bodyString);
```
LearningGp: Should this be downgraded to DEBUG level? It might be worth re-evaluating other INFO logs as well to avoid excessive log noise in production.

agentscope-core/src/main/java/io/agentscope/core/model/GeminiResponseGuard.java
```diff
@@ -0,0 +1,125 @@
...
+        if (finishReason == null
+                || finishReason.isEmpty()
+                || EXPECTED_FINISH_REASONS.contains(finishReason)) {
+            // Normal completion or streaming chunk - don't add fallback text or retry.
+            return response;
+        }
+
+        // For Gemini 3 models, throw exception on problematic finish reasons
+        // to trigger retry logic (workaround for API instability).
+        if (modelName.toLowerCase().contains("gemini-3")) {
```
LearningGp, line 86: Does this need to match gemini-3.x as well? Just checking if it's expected.

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiToolsHelper.java
```diff
@@ -63,34 +64,63 @@ public GeminiToolsHelper() {}
...
+                // For tools with no parameters, omit the parameters field entirely
+                Map<String, Object> parameters = toolSchema.getParameters();
+
+                // Special handling for generate_response tool:
+                // We previously unwrapped the schema for Gemini compatibility, but this caused
+                // conflicts with Gemini 2.5+ which follows the system prompt instructions
+                // (which request a "response" object) more strictly.
+                // We now keep the schema as-is (wrapped) to ensure prompt and schema alignment.
+                /*
+                if (StructuredOutputCapableAgent.STRUCTURED_OUTPUT_TOOL_NAME.equals(
```
LearningGp: Suggest removing this.

LearningGp (commented):
There are many unrelated formatting changes in this PR. Would it be possible to move these into a separate PR? This would help maintain commit cohesion and make the main logic easier to review.

Aias00:
> There are many unrelated formatting changes in this PR. Would it be possible to move these into a separate PR? This would help maintain commit cohesion and make the main logic easier to review.

find. I will handle other review comments first, and separate formatting changes later

agentscope-core/src/main/java/io/agentscope/core/model/GeminiResponseGuard.java
Aias00: Yes, this is intentional. contains("gemini-3") is meant to cover the whole Gemini 3 family, including gemini-3.x variants.

Aias00:
@AlbumenJ @LearningGp hi, pls help me reopen this pr

LearningGp (approved):
LGTM

LearningGp (changes requested):
For Gemini 3, are we supposed to use thinkingLevel instead of thinkingBudget now? Curious about the specific reason why "thinking" is restricted on Gemini 3 Flash.

Aias00:
> For Gemini 3, are we supposed to use thinkingLevel instead of thinkingBudget now? Curious about the specific reason why "thinking" is restricted on Gemini 3 Flash.

For Gemini 3, yes, Google now recommends using `thinkingLevel` rather than `thinkingBudget`. `thinkingBudget` is still accepted for backward compatibility, but the official guidance is: use `thinkingLevel` for Gemini 3 and later, and use `thinkingBudget` for Gemini 2.5. Google also notes that using `thinkingBudget` with Gemini 3 Pro can lead to unexpected performance, so it should not be treated as the preferred control going forward.

For the `Gemini 3 Flash + structured output` case, the current attempt to suppress thinking is not because Gemini Flash is officially defined as incompatible with thinking. The reason is that, in our earlier implementation, we observed that once thinking was enabled on `gemini-3-flash`, the tool-based structured output path became less stable and less deterministic. Since AgentScope currently implements structured output through the framework’s unified `generate_response` tool flow, we introduced this workaround to make Gemini 3 Flash align more reliably with the framework-defined structured output capability.

In other words, the current “disable thinking” behavior is a framework-compatibility mitigation: it was added to make `gemini-3-flash + structured output tool` behave more predictably within the existing framework contract, rather than because Gemini’s native design requires thinking to be turned off.

Going forward, we should evaluate whether Gemini structured output can be handled in a more native way, for example through `responseSchema`, while still adapting the result back into the framework’s common structured output semantics. If that works well, this workaround may no longer be necessary.

Official references:
- [Gemini Thinking](https://ai.google.dev/gemini-api/docs/thinking)
- [Gemini Structured Output](https://ai.google.dev/gemini-api/docs/structured-output)

Issue #265, opened 2025-12-19
[Feature]: Add options to `StreamOption` to filter out Reasoning events


AlbumenJ:
* **`includeReasoningChunk`**: The **incremental delta** of the reasoning process during streaming.
* **`includeReasoningResult`**: The **final, consolidated** reasoning output in the response.

sunxiang0918:
发现 如果 includeReasoningResult = false 了以后. 
就拿不到 ToolUseBlock 了. 以及 无法获取到 ChatStreamEvent.EventType.TOOL_CALL_START 事件.
从而,无法配对 TOOL_CALL_START与 TOOL_CALL_END. 界面无法渲染工具调用的块.

PR #269, opened 2025-12-20
feat: add reasoning chunk and result filtering options to StreamOptions

shlokmestry:
This PR adds explicit configuration options to StreamOptions to allow consumers to control how reasoning-related streaming events are emitted.

Some providers emit both incremental reasoning deltas (chunks) and a final consolidated reasoning result during streaming. Previously, these could not be filtered independently. 
This change introduces two new flags:
1. includeReasoningChunk: controls emission of incremental reasoning deltas
2. includeReasoningResult: controls emission of the final reasoning output

The defaults preserve existing behavior to avoid breaking changes.
Additional Javadoc and TODO notes reference Issue #265 and guide future integration in the stream mapping layer.

How to test:
1. Build the project: mvn clean install
2. Create StreamOptions via the builder and toggle the new flags
3. Verify flags are accessible and default behavior remains unchanged


shlokmestry:
All CI checks are now passing.  
Thanks for reviewing,  happy to make any changes if needed.

PR #271, opened 2025-12-21
perf(core): cache sorted hooks to optimize HookNotifier execution

shlokmestry:
1.0.4-SNAPSHOT

Background: `ReActAgent.HookNotifier` previously re-sorted the hook list on every hook
event notification. In streaming reasoning scenarios (e.g. 100+ chunks),
this resulted in unnecessary CPU overhead because the hook list is immutable
after agent construction.

Changes:
- Pre-sort hooks once during `ReActAgent` construction
- Cache the sorted hook list as an immutable field
- Reuse the cached list across all hook notifications
- Preserve hook priority ordering and existing behavior

Benefits:
- Eliminates repeated sorting overhead
- Improves streaming hook throughput
- Reduces CPU usage under high concurrency
- No functional or behavioral changes

How to Test:
```bash
mvn -pl agentscope-core spotless:apply
mvn -pl agentscope-core test

### CODE REVIEW

**Comment by shlokmestry** [comment]
Regarding Codecov missing lines:
The uncovered lines are in ReActAgent.HookNotifier and relate to simple pass-through / defensive paths that are unchanged in behavior.
This PR is a performance only refactor (caching pre-sorted hooks) with no functional logic change, and the critical execution paths are already covered by existing tests.
Additional tests would add limited value here and are intentionally omitted to keep the change minimal and low risk.

File: agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java [hunk 1]
```diff
@@ -146,6 +149,9 @@ public class ReActAgent extends AgentBase {
     private ReActAgent(Builder builder) {
         super(builder.name, builder.description, builder.checkRunning, builder.hooks);
 
+        this.sortedHooks =
+                builder.hooks.stream().sorted(Comparator.comparingInt(Hook::priority)).toList();
```
AlbumenJ: 1. Hooks are managed within **AgentBase** and should not be directly modified in **ReActAgent**.
2. Currently, many extensions add or remove hooks after a **ReActAgent** is instantiated. A mechanism is needed to support this capability.

shlokmestry:
Thanks for the review!

I’ve addressed the requested changes by:
- Moving hook sorting & caching into AgentBase
- Adding explicit add/remove hook APIs that invalidate the cache
- Updating ReActAgent to delegate hook access via AgentBase instead of managing its own sorted list

All checks are now passing locally and in CI.  
Please let me know if anything else needs adjustment.

agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
```diff
@@ -410,7 +428,15 @@ public List<Hook> getHooks() {
...
     protected List<Hook> getSortedHooks() {
-        return hooks.stream().sorted(java.util.Comparator.comparingInt(Hook::priority)).toList();
+        if (!hooksDirty.get() && cachedSortedHooks != null) {
+            return cachedSortedHooks;
+        }
+
+        List<Hook> sorted = hooks.stream().sorted(Comparator.comparingInt(Hook::priority)).toList();
+
+        cachedSortedHooks = sorted;
+        hooksDirty.set(false);
```
AlbumenJ, line 464: Concurrent issue here

agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
```diff
@@ -400,7 +408,17 @@ protected Mono<Void> doObserve(Msg msg) {
      * @return List of hooks
      */
     public List<Hook> getHooks() {
-        return hooks;
+        return List.copyOf(hooks);
```
AlbumenJ, line 418: Also fix here

Image

AlbumenJ (changes requested):

shlokmestry:
I’ve addressed all the requested changes from the review and updated the implementation accordingly.
The remaining CI failure appears to be coming from an unrelated test in the extensions module; I’m not sure about the root cause yet, but it doesn’t seem connected to the changes in this PR.

Issue #208, opened 2025-12-16
[Feature]: Allow `Msg.getStructuredData()` to accept a run-time JSON Schema instead of a compiled Class


kidbei:
**<u>AgentScope-Java is an open-source project. To involve a broader community, we recommend asking your questions in English.</u>**

**Is your feature request related to a problem? Please describe.**  
Yes.  
In our use-case, end-users are allowed to design the JSON schema they expect the agent to return (via a visual schema editor on the front-end).  
At run-time we only have this user-supplied schema (usually expressed as a `JsonNode` or raw JSON string), **not** a compiled Java class.  
Currently `Msg.getStructuredData(Class<?> clazz)` only accepts a compiled `Class` object, so we have to:

1. Generate a Java source file from the schema at run-time.
2. Compile it in-memory.
3. Load the resulting `Class` and pass it to `getStructuredData`.

This is slow, fragile and impossible on some restricted runtimes.  
We would like to skip the “generate-and-compile” step and let `getStructuredData` work directly with the user-defined schema.

**Describe the solution you'd like**  
Add an overloaded method (or an optional parameter) that accepts a **JSON Schema** instead of a `Class`, e.g.

```java
public <T> T Msg.getStructuredData(JsonNode schema);  
// or  
public <T> T Msg.getStructuredData(String jsonSchema);
```

Internally the framework would use the same binding / validation engine it already uses (Jackson, JSON-B, etc.) but would build the Java object dynamically (e.g. `JsonNode`, `Map<String,Object>`, or a generated in-memory type) instead of requiring a pre-compiled POJO.

If the returned type must be generic, `JsonNode` or `Map<String,Object>` is perfectly acceptable for us; we just need the validation and conversion to be done by AgentScope-Java.

**Describe alternatives you've considered**  
1. Pre-generate one POJO per possible schema – impossible because users create new schemas at any time.  
2. Generate and compile a POJO at runtime – works, but adds heavy JDK compiler dependencies and slows the agent response.  
3. Call `getText()` and manually validate / convert – duplicates the logic that AgentScope-Java already has inside `getStructuredData`.

**Additional context**  
*none*

Jamison929611:
@AlbumenJ please assign it to me

Issue #300, opened 2025-12-22
Map.copyOf(metadata) throws a null pointer exception when a null field appears in the structured output object.


harri2012:
https://github.com/agentscope-ai/agentscope-java/blame/82d8840b7f0ab01f46e03222f91b2bda4434eeaa/agentscope-core/src/main/java/io/agentscope/core/message/Msg.java#L90

As title mentioned, modifications are needed here to allow field values in the structured output object to be null, thereby preventing the Map.copyOf(metadata) method from throwing a NullPointerException when the metadata contains null values.

Jamison929611:
please assign it to me

AlbumenJ:
Fixed in https://github.com/agentscope-ai/agentscope-java/pull/303

harri2012:
This issue need more modifications: 

- `List.copyOf(content)` : throws NullPointerException if content is null, or if it contains any nulls.
- `Map.copyOf(metadata)` : throws NullPointerException if metadata is null, or if it contains any null keys or values.

@AlbumenJ

PR #357, opened 2025-12-25
feat(output): add support for structured data dynamic define

chiangzeon:
## AgentScope-Java Version

[The version of AgentScope-Java you are working on, e.g. 1.0.3, check your pom.xml dependency version or run `mvn dependency:tree | grep agentscope-parent:pom`(only mac/linux)]

## Description
Support dynamic structured data definition.

Summary

1.io.agentscope.core.agent.Agent#call(io.agentscope.core.message.Msg, java.lang.Class<?>) The reason for passing POJO is because it is called io.agentscope.core.ReActAgent#doCall(java.util.List<io.agentscope.core.message.Msg>, java.lang.Class<?>) We need to register a tool to summarize the output.

2. Need to overload  getStructuredData method, as the dynamic type return can only be KV data and metadata already contains the original response. For concurrency security, the input parameter provides a Boolean type for the caller to choose whether it can be modified.

3. The core call is in io.agentscope. core. tracking. Tracer # callModel. Based on the protocols of different providers, the override io. agentscope. core. model. ChatModelBase # doStream method is required.

If there is any mistake, please point it out promptly. Thank you

[https://github.com/agentscope-ai/agentscope-java/issues/208](url)

## Checklist

### DIFF
--- a/agentscope-core/src/test/java/io/agentscope/core/agent/StructuredOutputDynamicDefineTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/agent/StructuredOutputDynamicDefineTest.java
@@ -16,136 +16,140 @@
 
 package io.agentscope.core.agent;
 
-import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.agentscope.core.ReActAgent;
-import io.agentscope.core.agent.test.TestUtils;
-import io.agentscope.core.e2e.providers.DashScopeCompatibleProvider;
+import io.agentscope.core.agent.test.MockModel;
+import io.agentscope.core.memory.InMemoryMemory;
+import io.agentscope.core.memory.Memory;
 import io.agentscope.core.message.Msg;
+import io.agentscope.core.message.MsgRole;
+import io.agentscope.core.message.TextBlock;
+import io.agentscope.core.message.ToolUseBlock;
+import io.agentscope.core.model.ChatResponse;
+import io.agentscope.core.model.ChatUsage;
 import io.agentscope.core.tool.Toolkit;
-import java.time.Duration;
-import java.util.Collection;
+import java.util.List;
 import java.util.Map;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /** Tests for dynamic define json schema in StructuredOutputHandler (deferred forcing mode). */
 public class StructuredOutputDynamicDefineTest {
 
-    private static final Duration TEST_TIMEOUT = Duration.ofSeconds(60);
+    private Toolkit toolkit;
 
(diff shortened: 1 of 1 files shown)


agentscope-core/src/main/java/io/agentscope/core/message/Msg.java
```diff
@@ -87,8 +88,20 @@ private Msg(
...
+        this.metadata =
+                Objects.nonNull(metadata)
+                                && metadata.entrySet().stream()
+                                        .allMatch(
+                                                entry ->
+                                                        Objects.nonNull(entry.getKey())
+                                                                && Objects.nonNull(
+                                                                        entry.getValue()))
+                        ? Map.copyOf(metadata)
+                        : Map.of();
```
AlbumenJ: Why change this?

agentscope-core/src/main/java/io/agentscope/core/agent/StructuredOutputHandler.java
AlbumenJ: Will fail here

AlbumenJ (changes requested):

agentscope-core/src/main/java/io/agentscope/core/message/Msg.java
Jamison929611: > Why change this?

[https://github.com/agentscope-ai/agentscope-java/issues/300](url) new bug report

agentscope-core/src/main/java/io/agentscope/core/agent/StructuredOutputHandler.java
```diff
@@ -270,9 +287,16 @@ public Mono<ToolResultBlock> callAsync(ToolCallParam param) {
...
+                                    && responseData != null) {
                                 try {
-                                    OBJECT_MAPPER.convertValue(responseData, targetClass);
+                                    // Dynamic types are not verified, and the data needs to be
+                                    // reconstructed according to schemaDesc first, and then
+                                    // verified attribute by attribute. It is impossible to exhaust
+                                    // all the rules and leave them to the calling side
+                                    if (Objects.nonNull(targetClass)) {
+                                        OBJECT_MAPPER.convertValue(responseData, targetClass);
+                                    }
```
AlbumenJ, line 326: Can we validate the `schemaDesc` here?

agentscope-core/src/main/java/io/agentscope/core/message/Msg.java
AlbumenJ: Please submit a new PR to fix it

AlbumenJ (changes requested):

agentscope-core/src/main/java/io/agentscope/core/agent/StructuredOutputHandler.java
Jamison929611: Yes, this is technically feasible. However, I would generally advise against it because a dynamically generated validation mechanism cannot comprehensively cover all business rules. Implementing this feature would incur significant development and maintenance overhead. Therefore, I propose that the responsibility for parameter validation should lie with the calling client. Do you agree with this perspective?

AlbumenJ: Given that current Large Language Models (LLMs) still exhibit limitations in generating complex structures, it is essential to implement data format validation within the Agent. This facilitates timely prompting of the model for error correction.

Jamison929611: > Given that current Large Language Models (LLMs) still exhibit limitations in generating complex structures, it is essential to implement data format validation within the Agent. This facilitates timely prompting of the model for error correction.

Done

PR #377, opened 2025-12-27
refactor(core): refactor toolkit and skillBox registration

fang-tech:
## AgentScope-Java Version

1.0.4

## Description

Refactor toolkit and skillBox registration
Now, for the situation where multiple tools are registered simultaneously, an error occurs when apply().
Passing in an empty value is no longer regarded as a valid call.

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/Toolkit.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/Toolkit.java
@@ -684,7 +684,6 @@ public class Toolkit extends StateModuleBase {
         private ExtendedModel extendedModel;
         private List<String> enableTools;
         private List<String> disableTools;
-        private int toolCount = 0;
 
         private ToolRegistration(Toolkit toolkit) {
             this.toolkit = toolkit;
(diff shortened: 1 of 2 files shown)


fang-tech:
CI failed reason is not related to this PR

agentscope-core/src/main/java/io/agentscope/core/tool/Toolkit.java
```diff
@@ -696,12 +697,8 @@ private ToolRegistration(Toolkit toolkit) {
...
          */
         public ToolRegistration tool(Object toolObject) {
-            if (this.agentTool != null
-                    || this.mcpClientWrapper != null
-                    || this.subAgentProvider != null) {
-                throw new IllegalStateException(
-                        "Cannot set multiple registration types. Use only one of: tool(),"
-                                + " agentTool(), mcpClient(), or subAgent().");
+            if (toolObject != null) {
+                toolCount++;
```
AlbumenJ: If the user returns `null`, `toolCount` will not be decremented. It is recommended to move the `toolCount` calculation into the `build` method.

AlbumenJ (changes requested):

agentscope-core/src/main/java/io/agentscope/core/tool/Toolkit.java
fang-tech: yep, have fixed

Issue #430, opened 2026-01-04
[Bug]: [Boba tea shop] - Failed to get list: autoContextMemory_workingMessages


zhliuka2024:
**<u>AgentScope-Java is an open-source project. To involve a broader community, we recommend asking your questions in English.</u>**

**Describe the bug**

Deserialize issue when load autoContextMemory - workingMessages from db.

**To Reproduce**

1. You code

<img width="740" height="398" alt="Image" src="https://github.com/user-attachments/assets/dad7003d-7d6c-476b-afa6-973875ee5d0c" />

2. How to execute

<img width="991" height="698" alt="Image" src="https://github.com/user-attachments/assets/3fa773a2-84b5-4d49-922a-f2d12ff8880a" />

3. See error
System processing error, please try again later.

**Expected behavior**
continue to conversion normally.

**Error messages**
Detailed error messages.
2026-01-04T13:23:26.718+08:00 ERROR 33900 --- [supervisor-agent] [io-10008-exec-2] i.a.e.b.s.c.SupervisorAgentController    : Failed to process user query: ????????

java.lang.RuntimeException: Failed to get list: autoContextMemory_workingMessages
        at io.agentscope.core.session.mysql.MysqlSession.getList(MysqlSession.java:593) ~[agentscope-extensions-session-mysql-1.0.6-SNAPSHOT.jar:na]
        at io.agentscope.core.memory.autocontext.AutoContextMemory.loadFrom(AutoContextMemory.java:1761) ~[agentscope-extensions-autocontext-memory-1.0.6-SNAPSHOT.jar:na]
        at io.agentscope.core.ReActAgent.loadFrom(ReActAgent.java:222) ~[agentscope-core-1.0.6-SNAPSHOT.jar:na]
        at io.agentscope.core.state.StateModule.loadIfExists(StateModule.java:104) ~[agentscope-core-1.0.6-SNAPSHOT.jar:na]
        at io.agentscope.core.state.StateModule.loadIfExists(StateModule.java:118) ~[agentscope-core-1.0.6-SNAPSHOT.jar:na]
        at io.agentscope.examples.bobatea.supervisor.agent.SupervisorAgent.stream(SupervisorAgent.java:96) ~[classes/:na]
        at io.agentscope.examples.bobatea.supervisor.controller.SupervisorAgentController.chat(SupervisorAgentController.java:82) ~[classes/:na]
[...]

PR #435, opened 2026-01-04
Solved #430

MentosL:
To address the deserialization issue, we are currently using the `@Ignore` annotation to skip certain fields.

Additionally, to maintain compatibility with the existing suspend logic, a new field has been added to handle this.

## AgentScope-Java Version

[The version of AgentScope-Java you are working on, e.g. 1.0.5, check your pom.xml dependency version or run `mvn dependency:tree | grep agentscope-parent:pom`(only mac/linux)]

## Description

Solved  #430 

To address the deserialization issue, we are currently using the @Ignore annotation to skip certain fields.

Additionally, to maintain compatibility with the existing suspend logic, a new field has been added to handle this.

If there are any problems, please point them out to me. thx

## Checklist


agentscope-core/src/main/java/io/agentscope/core/message/ToolResultBlock.java
```diff
@@ -119,7 +129,8 @@ public Map<String, Object> getMetadata() {
      * @return true if this result is suspended, false otherwise
      */
     public boolean isSuspended() {
```
AlbumenJ, line 135: how about add `@Transient` and `@JsonIgnore` here

MentosL: thx. that great. It seems that someone else has already committed this section of the code.

PR #560, opened 2026-01-14
fix(example): Inconsistent groupManager in ReActAgent.Builder#build due to toolkit copying

fishl3j:
During `ReActAgent.Builder#build`, the toolkit is copied to prevent state interference. However, this leads to a mismatch in `groupManager` instances under certain conditions:
- If `registerMetaTool()` is called on the original toolkit, the meta tool holds a reference to the **original** toolkit's `groupManager`.
- After copying, the new toolkit's `schemaProvider` uses the **copied** toolkit's `groupManager`.
- As a result, the `groupManager` used for tool group activation differs from the one used in schema generation, causing inconsistency.
**Fix:**
- Remove direct calls to `registerMetaTool()` on the original toolkit.
- Instead, use `Builder.enableMetaTool(true)` to register the meta tool **during build time** on the **copied** toolkit.
- This ensures both the meta tool and the schemaProvider use the **same** `groupManager` instance.


AlbumenJ:
Please translate title and description into English

PR #586, opened 2026-01-16
fix(agui): Fix duplicate TextMessageEnd emission (#562)

fishl3j:
# Background: Fix Duplicate AguiEvent.TextMessageEnd Issue

## Problem Description

When using AGUI protocol to request a ReActAgent with SKILL and TOOL capabilities, the `AguiAgentAdapter` was emitting duplicate `AguiEvent.TextMessageEnd` events, causing incorrect event closure in the AGUI event stream.

## Root Cause

The issue occurs in the `convertEvent` method of `AguiAgentAdapter` when processing `REASONING` events that contain both `TextBlock` and `ToolUseBlock` content blocks:

1. **First emission path**: When processing a `ToolUseBlock` ,checks if there's an active text message and attempts to end it before starting the tool call (lines 162-169).

2. **Second emission path**: When processing a `TextBlock` in the last event (`event.isLast() == true`), the code emits `TextMessageEnd` to close the text message (lines 152-158).

3. **The bug**: If a message contains both text content and tool usage, and the text block is processed as the last event, the `TextMessageEnd` is emitted. However, when the subsequent `ToolUseBlock` is processed, the code doesn't check if the message has already been ended, leading to a duplicate `TextMessageEnd` event.

## Impact

- **Issue**: [#562](https://github.com/agentscope-ai/agentscope-java/issues/562) - AGUI方式下请求使用Skill的ReActAgent，返回的AguiEvent存在闭合错误
- **Version affected**: 1.0.8-SNAPSHOT
- **Symptoms**: 
  - Duplicate `TextMessageEnd` events in the AGUI event stream
  - Incorrect event closure state
  - Potential issues with AGUI clients expecting proper event sequencing

## Solution

Added a guard check using `state.hasEndedMessage(messageId)` before emitting `TextMessageEnd` events. This ensures that:

1. The message end event is only emitted once per message
2. The state tracking correctly reflects whether a message has been closed
3. The event stream maintains proper sequencing and closure semantics

## Code Changes

The fix adds a conditional check before emitting `TextMessageEnd`:

```java
if (!state.hasEndedMessage(messageId)) {
    events.add(
        new AguiEvent.TextMessageEnd(
            state.threadId, state.runId, messageId));
    state.endMessage(messageId);
}
```

This prevents duplicate emissions while maintaining the existing logic for properly closing messages when tool calls are encountered.


AlbumenJ (changes requested):
Image

Please resolve conflicts

THX

PR #605, opened 2026-01-20
Feat add custom event

fishl3j:
# Add Custom Event Support to AguiEvent

## Summary
This issue implements support for Custom events in the AG-UI protocol, providing an extension mechanism for implementing features not covered by the standard event types.

## Motivation
The AG-UI protocol currently supports various standard event types (RUN_STARTED, TEXT_MESSAGE_START, TOOL_CALL_START, etc.) and a RAW event type for unstructured custom data. However, there is a need for a structured Custom event type that allows passing custom data with a name identifier, following the AG-UI protocol specification.

### Changes Made

1. **Added CUSTOM Event Type**
   - Added `CUSTOM` enum value to `AguiEventType`
   - Updated event type count from 12 to 13

2. **Implemented Custom Event Record**
   - Created `AguiEvent.Custom` record implementing `AguiEvent` interface
   - Fields:
     - `threadId` (String, required): Thread ID associated with the event
     - `runId` (String, required): Run ID associated with the event
     - `name` (String, required): Name of the custom event
     - `value` (Object, optional): Value associated with the event
   - Added proper null validation for required fields
   - Implemented JSON serialization/deserialization support via `@JsonCreator` and `@JsonProperty` annotations

3. **Updated Sealed Interface**
   - Added `AguiEvent.Custom` to the `permits` clause of the sealed `AguiEvent` interface
   - Added `@JsonSubTypes` annotation entry for Custom event type mapping

4. **Test Coverage**
   - Added comprehensive test suite in `CustomTest` nested class:
     - Event creation tests
     - Null value handling
     - Complex value handling
     - JSON serialization/deserialization
     - Null parameter validation
     - toString() method verification
   - Updated `testEventTypeCount` to expect 13 event types
   - Added CUSTOM to `testAllEventTypesExist` test

## Protocol Compliance
The implementation follows the AG-UI protocol specification:
- Custom events include a `name` field for event identification
- Custom events include a `value` field for event data
- All standard event properties (threadId, runId) are maintained

## Example Usage

```java
import io.agentscope.core.agui.event.AguiEvent;
import io.agentscope.core.agui.event.AguiEventType;
import io.agentscope.core.util.JsonUtils;
import java.util.Map;

// Create a custom event with a name and value
AguiEvent.Custom customEvent = new AguiEvent.Custom(
        "thread-1",
        "run-1",
        "test-event",
        Map.of("key", "value", "number", 42));

// Verify event properties
assertEquals(AguiEventType.CUSTOM, customEvent.getType());
assertEquals("thread-1", customEvent.getThreadId());
assertEquals("run-1", customEvent.getRunId());
assertEquals("test-event", customEvent.name());
assertNotNull(customEvent.value());

// Serialize to JSON
String json = JsonUtils.getJsonCodec().toJson(customEvent);
// JSON contains: "type":"CUSTOM", "name":"test-event", "value":{...}

// Deserialize from JSON
AguiEvent deserialized = JsonUtils.getJsonCodec().fromJson(json, AguiEvent.class);
assertTrue(deserialized instanceof AguiEvent.Custom);
AguiEvent.Custom deserializedCustom = (AguiEvent.Custom) deserialized;
assertEquals("test-event", deserializedCustom.name());
```

### DIFF
--- a/agentscope-extensions/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/event/AguiEvent.java
+++ b/agentscope-extensions/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/event/AguiEvent.java
@@ -49,7 +49,19 @@ import java.util.Objects;
     @JsonSubTypes.Type(value = AguiEvent.StateSnapshot.class, name = "STATE_SNAPSHOT"),
     @JsonSubTypes.Type(value = AguiEvent.StateDelta.class, name = "STATE_DELTA"),
     @JsonSubTypes.Type(value = AguiEvent.Raw.class, name = "RAW"),
-    @JsonSubTypes.Type(value = AguiEvent.Custom.class, name = "CUSTOM")
+    @JsonSubTypes.Type(value = AguiEvent.Custom.class, name = "CUSTOM"),
+    @JsonSubTypes.Type(value = AguiEvent.ReasoningStart.class, name = "REASONING_START"),
+    @JsonSubTypes.Type(
+            value = AguiEvent.ReasoningMessageStart.class,
+            name = "REASONING_MESSAGE_START"),
+    @JsonSubTypes.Type(
+            value = AguiEvent.ReasoningMessageContent.class,
+            name = "REASONING_MESSAGE_CONTENT"),
+    @JsonSubTypes.Type(value = AguiEvent.ReasoningMessageEnd.class, name = "REASONING_MESSAGE_END"),
+    @JsonSubTypes.Type(
+            value = AguiEvent.ReasoningMessageChunk.class,
+            name = "REASONING_MESSAGE_CHUNK"),
+    @JsonSubTypes.Type(value = AguiEvent.ReasoningEnd.class, name = "REASONING_END")
 })
 public sealed interface AguiEvent
         permits AguiEvent.RunStarted,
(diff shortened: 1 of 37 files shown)


AlbumenJ (changes requested):
Image

Please resolve conflicts

agentscope-extensions/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/event/AguiEvent.java
```diff
@@ -530,6 +532,40 @@ public String getRunId() {
         }
     }
 
+    /**
+     * The Custom event provides an extension mechanism for implementing
+     * features not covered by the standard event types.
+     */
+    record Custom(String threadId, String runId, String name, Object value) implements AguiEvent {
```
AlbumenJ, line 539: How to publish this event?

AlbumenJ (changes requested):

agentscope-extensions/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/event/AguiEvent.java
fishl3j: Currently, we have certain scenarios where, in addition to the AguiEvent output by AguiAgentAdapter, we need to send an extra card render event to the client for rendering card-based information. This event cannot be effectively carried using a RawEvent.

Moreover, since CustomEvent is already a defined type within the AG-UI protocol, it is theoretically necessary to extend or formalize this capability.

AlbumenJ: Currently, this POJO definition is not directly used within AgentScope. Therefore, my question is: how should users utilize it? Does it require directly modifying the implementation of AG-UI in AgentScope?

fishl3j: Yes, currently there is no place in AgentScope that actually uses the CustomEvent POJO.

I’m not fully leveraging all the AG-UI features provided by AgentScope—for example, the `AguiRestController`. Although it exposes a complete `RunAgentInput`POJO, some parameters (such as `RunAgentInput.context`) are not actually used or passed through, which prevents the agent from accessing the context content.

Therefore, I only use the `AguiRequestProcessor` to directly handle the AG-UI event stream and inject my own custom events.

This is the motivation and rationale behind adding the `CustomEvent` POJO, which is currently unused by AgentScope.

PR #800, opened 2026-02-14
fix(model): Fix NPE when OpenAIClient stream call handle error

aihai:
## AgentScope-Java Version
1.0.10

## Description
Fixes: #786 

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/model/exception/OpenAIException.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/exception/OpenAIException.java
@@ -54,6 +54,13 @@ public class OpenAIException extends RuntimeException {
         this.responseBody = responseBody;
     }
 
+    public OpenAIException(String message, String errorCode, String responseBody) {
+        super(message);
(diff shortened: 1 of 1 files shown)


agentscope-core/src/main/java/io/agentscope/core/model/exception/OpenAIException.java
```diff
@@ -64,7 +64,11 @@ public OpenAIException(String message, int statusCode, String errorCode, String
      * @return Appropriate exception subclass
      */
     public static OpenAIException create(
-            int statusCode, String message, String errorCode, String responseBody) {
+            Integer statusCode, String message, String errorCode, String responseBody) {
+        if (null == statusCode) {
+            // When no status code is available, the default is 418
+            statusCode = 418;
```
LearningGp: 418 is not a suitable default error code. Please consider constructing an OpenAIException for this case instead.

aihai: OK, I will do it.

PR #906, opened 2026-03-10
fix(openai): avoid setting max_tokens and max_completion_tokens together

### LINKED ISSUE #905: [Bug]:fix(openai): max_tokens and max_completion_tokens are set together causing request failures on some models
Background
In agentscope-core’s OpenAI Chat Completion HTTP integration, the current implementation sets both max_tokens and max_completion_tokens based on GenerateOptions.maxTokens. Some OpenAI / OpenAI-compatible providers treat these two parameters as mutually exclusive, which causes requests to fail.
Steps to reproduce
Use OpenAIChatModel (or any model/formatter that uses OpenAIChatFormatter, including subclasses such as DeepSeekFormatter and GLMFormatter).
Set GenerateOptions.maxTokens (e.g. GenerateOptions.builder().maxTokens(512).build()).
Send a chat completion request.
Actual result
The request body contains both max_tokens and max_completion_tokens, and some models/providers return an error like:
> The parameter max_tokens specified in the request are not valid: max_tokens and max_completion_tokens cannot be set at the same time
Expected result
If the caller sets only maxTokens, the request should include only max_tokens
If the caller sets only maxCompletionTokens, the request should include only max_completion_tokens
If the caller explicitly sets both values, both fields may be sent as-is (the SDK should not enforce exclusivity; the provider/model can decide whether it’s valid)
Root cause
In OpenAIChatFormatter.applyOptions, the maxTokens handling currently calls both:
request.setMaxTokens(maxTokens)
request.setMaxCompletionTokens(maxTokens)
This means even when the user only configures GenerateOptions.maxTokens, the request still carries max_completion_tokens.
Proposed fix (I can submit a PR)
Add a new field maxCompletionTokens to GenerateOptions (getter + builder + merge support).
Update OpenAIChatFormatter.applyOptions to read and set max_tokens and max_completion_tokens independently, avoiding mirroring one option into both request fields.
Impact assessment
Fix benefits OpenAIChatFormatter and all subclasses that inherit its option application (OpenAIMultiAgentFormatter, DeepSeekFormatter, GLMFormatter, etc.).
Other provider integrations (Gemini/DashScope/Anthropic/Ollama, etc.) are unaffected and can continue using maxTokens only.

darkwu:
Fixes #905

## Summary
- Add `GenerateOptions.maxCompletionTokens`
- Update `OpenAIChatFormatter` to set `max_tokens` and `max_completion_tokens` independently, without mirroring

## Behavior
- If only `maxTokens` is set -> only `max_tokens` is sent
- If only `maxCompletionTokens` is set -> only `max_completion_tokens` is sent
- If both are set -> both fields are sent as-is (SDK does not enforce exclusivity; provider can decide)

## Notes
- Affects `OpenAIChatFormatter` and its subclasses (e.g. `OpenAIMultiAgentFormatter`, `DeepSeekFormatter`, `GLMFormatter`)
- Other providers (Gemini/DashScope/Anthropic/Ollama, etc.) continue to use `maxTokens` only

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIChatFormatter.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIChatFormatter.java
@@ -101,13 +101,7 @@ public class OpenAIChatFormatter extends OpenAIBaseFormatter {
             request.setPresencePenalty(presencePenalty);
         }
 
-        // Apply max tokens
-        Integer maxTokens =
-                getOptionOrDefault(options, defaultOptions, GenerateOptions::getMaxTokens);
-        if (maxTokens != null) {
-            request.setMaxCompletionTokens(maxTokens);
-            request.setMaxTokens(maxTokens);
-        }
+        applyMaxTokens(request, options, defaultOptions);
 
         // Apply seed
         Long seed = getOptionOrDefault(options, defaultOptions, GenerateOptions::getSeed);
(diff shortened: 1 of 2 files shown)


darkwu:
同时set值调用部分模型的时候会出现如下异常：The parameter `max_tokens` specified in the request are not valid: max_tokens and max_completion_tokens cannot be set at the same time

PR #928, opened 2026-03-11
fix(openai): handle null statusCode in HttpTransportException to prevent NPE

### LINKED ISSUE #891: NPE in OpenAIClient: HttpTransportException.getStatusCode() null auto-unboxing
## Bug Description

`OpenAIClient.java` line 442-443 passes `HttpTransportException.getStatusCode()` (return type `Integer`, nullable) directly to `OpenAIException.create(int, ...)` (parameter type `int`). When the exception is constructed without a status code (e.g. connection failure, DNS error, timeout), `getStatusCode()` returns `null`, and Java auto-unboxing throws a `NullPointerException`.

## Stack Trace

```
Cannot invoke "java.lang.Integer.intValue()" because the return value of
"io.agentscope.core.model.transport.HttpTransportException.getStatusCode()" is null
```

## Root Cause

`HttpTransportException` has constructors that leave `statusCode = null`:

```java
public HttpTransportException(String message) {
    super(message);
    this.statusCode = null;  // no HTTP status code
}

public HttpTransportException(String message, Throwable cause) {
    super(message, cause);
    this.statusCode = null;  // no HTTP status code
}
```

But the streaming error handler in `OpenAIClient` assumes it's always non-null:

```java
.onErrorMap(ex -> {
    if (ex instanceof HttpTransportException) {
        return OpenAIException.create(
            ((HttpTransportException) ex).getStatusCode(),  // NPE here when null
            "HTTP transport error during streaming: " + ex.getMessage(),
            null,
            ((HttpTransportException) ex).getResponseBody());
    }
    return ex;
});
```

## Suggested Fix

```java
.onErrorMap(ex -> {
    if (ex instanceof HttpTransportException hte) {
        Integer code = hte.getStatusCode();
        String msg = "HTTP transport error during streaming: " + ex.getMessage();
        if (code != null) {
            return OpenAIException.create(code, msg, null, hte.getResponseBody());
        }
        return new OpenAIException(msg, ex);
    }
    return ex;
});
```

## Environment

- agentscope-core version: 1.0.10-SNAPSHOT
- Java: 21

shlokmestry:
1.0.10-SNAPSHOT

Fixes a NullPointerException in OpenAIClient.stream() caused by auto-unboxing 
a nullable Integer when HttpTransportException has no HTTP status code 
(e.g. connection failure, DNS error, timeout).

The existing code passed getStatusCode() (return type Integer, nullable) 
directly to OpenAIException.create(int, ...), which throws an NPE when 
statusCode is null.

Fix: Added a null check before accessing getStatusCode(). If null, falls back 
to new OpenAIException(msg, ex) instead of crashing.

Fixes #891

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -4,6 +4,7 @@ updates:
     directory: "/"
     open-pull-requests-limit: 20
     schedule:
-      interval: "daily"
+      interval: "weekly"
+      day: "sunday"
       time: "03:00"
       timezone: "Asia/Shanghai"
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -99,6 +99,12 @@ jobs:
           mkdir -Path "~\.mvnd" -Force
           Expand-Archive -Path "mvnd.zip" -DestinationPath "mvnd" -FORCE
           Copy-Item -Path "mvnd\maven-mvnd-${{ env.MVND_VERSION }}-windows-amd64\*" -Destination "~\.mvnd\" -Recurse
+          Set-Content -Path "~\.mvnd\conf\mvnd.properties" -Value @"
+          # mvnd client timeout settings
+          mvnd.connectTimeout=120000
+          mvnd.socketConnectTimeout=30000
+          mvnd.cancelConnectTimeout=30000
+          "@ -Encoding Utf8 -Force
           Remove-Item -Path "mvnd", "mvnd.zip" -Recurse -Force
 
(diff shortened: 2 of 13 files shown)


shlokmestry:
Ran OpenAIClientExceptionTest locally : 17/17 tests pass (Failures: 0, Errors: 0). The only failing test in the full suite (ShellCommandToolTest$BufferDeadlockTests) is unrelated to this change. CI failures are due to Codecov token not being configured for fork PRs and a Windows build timeout, not the code change itself.

PR #956, opened 2026-03-14
fix: handle tool execution timeout/error causing IllegalStateExceptio…

### LINKED ISSUE #951: [Bug]: java.lang.IllegalStateException: Cannot add messages without tool results when pending tool calls exist. Pending IDs: [call_xxx]
# ReActAgent Tool Execution Error Handling Enhancement

## Issue Summary

**Problem:** `ReActAgent` throws `IllegalStateException` when tool calls timeout or fail, causing the agent to crash and leaving pending tool call states that break subsequent requests.

**Error Message:**
```
java.lang.IllegalStateException: Cannot add messages without tool results when pending tool calls exist. Pending IDs: [call_xxx]
    at io.agentscope.core.ReActAgent.validateAndAddToolResults(ReActAgent.java:335)
    at io.agentscope.core.ReActAgent.doCall(ReActAgent.java:256)
```

### Scenario 1: Tool Execution Timeout/Error
```
1. Model returns tool_call (ID: call_xxx)
2. Toolkit.callTools() times out or throws exception
3. Exception propagates up → No tool result written to memory
4. Pending tool call state remains in memory
5. Agent crashes with IllegalStateException
```

### Scenario 2: Orphaned Pending State
```
1. Previous request's tool execution fails mid-way
2. Error occurs before tool result is written to memory
3. Pending state persists in memory
4. New request arrives → doCall() detects pending IDs
5. User message has no tool results → validateAndAddToolResults() throws
```

### Fix 1: Error Recovery in `executeToolCalls()`

Add `onErrorResume` to catch tool execution failures and generate error tool results instead of propagating exceptions.

**Location:** `ReActAgent.java` - `executeToolCalls()` method

```java
private Mono<List<Map.Entry<ToolUseBlock, ToolResultBlock>>> executeToolCalls(
        List<ToolUseBlock> toolCalls) {
    return toolkit.callTools(toolCalls, toolExecutionConfig, this, toolExecutionContext)
            .map(results -> IntStream.range(0, toolCalls.size())
                    .mapToObj(i -> Map.entry(toolCalls.get(i), results.get(i)))
                    .toList())
            // NEW: Error recovery - generate error results instead of failing
            .onErrorResume(error -> {
                log.error("Tool execution failed, generating error results for {} tool calls: {}",
                        toolCalls.size(), error.getMessage());
                List<Map.Entry<ToolUseBlock, ToolResultBlock>> errorResults = toolCalls.stream()
                        .map(toolCall -> {
                            ToolResultBlock errorResult = ToolResultBlock.builder()
                                    .id(toolCall.getId())
                                    .output(List.of(TextBlock.builder()
                                            .text("[ERROR] Tool execution failed: " + error.getMessage())
                                            .build()))
                                    .build();
                            return Map.entry(toolCall, errorResult);
                        })
                        .toList();
                return Mono.just(errorResults);
            });
}
```

### Fix 2: Auto-Recovery in `doCall()`

When pending tool calls are detected but user message contains no tool results, automatically generate error results to clear the pending state.

**Location:** `ReActAgent.java` - `doCall()` method

```java
@Override
protected Mono<Msg> doCall(List<Msg> msgs) {
    Set<String> pendingIds = getPendingToolUseIds();

    if (pendingIds.isEmpty()) {
        addToMemory(msgs);
        return executeIteration(0);
    }

    // NEW: Check if user provided tool results
    List<ToolResultBlock> providedResults = msgs == null ? List.of() :
            msgs.stream()
                    .flatMap(m -> m.getContentBlocks(ToolResultBlock.class).stream())
                    .toList();

    if (providedResults.isEmpty()) {
        // NEW: Auto-generate error results for orphaned pending calls
        log.warn("Pending tool calls detected without results, auto-generating error results. Pending IDs: {}", pendingIds);
        generateAndAddErrorToolResults(pendingIds);
        addToMemory(msgs);
        return executeIteration(0);
    }

    validateAndAddToolResults(msgs, pendingIds);
    return hasPendingToolUse() ? acting(0) : executeIteration(0);
}
```

### Fix 3: New Helper Method `generateAndAddErrorToolResults()`

**Location:** `ReActAgent.java` - new private method

```java
/**
 * Generate error tool results for pending tool calls and add them to memory.
 * This is used to recover from situations where tool execution failed without
 * properly writing results to memory.
 */
private void generateAndAddErrorToolResults(Set<String> pendingIds) {
    Msg lastAssistant = findLastAssistantMsg();
    if (lastAssistant == null) {
        return;
    }

    List<ToolUseBlock> pendingToolCalls = lastAssistant.getContentBlocks(ToolUseBlock.class).stream()
            .filter(toolUse -> pendingIds.contains(toolUse.getId()))
            .toList();

    for (ToolUseBlock toolCall : pendingToolCalls) {
        ToolResultBlock errorResult = ToolResultBlock.builder()
                .id(toolCall.getId())
                .output(List.of(TextBlock.builder()
                        .text("[ERROR] Previous tool execution failed or was interrupted. Tool: " + toolCall.getName())
                        .build()))
                .build();
        Msg toolResultMsg = ToolResultMessageBuilder.buildToolResultMsg(errorResult, toolCall, getName());
        memory.addMessage(toolResultMsg);
        log.info("Auto-generated error result for pending tool call: {} ({})", toolCall.getName(), toolCall.getId());
    }
}
```

## Behavior Changes

| Scenario | Before | After |
|----------|--------|-------|
| Tool execution timeout | `IllegalStateException` thrown, agent crashes | Error result generated, model receives error feedback, continues processing |
| Tool execution error | Exception propagates, pending state orphaned | Error result generated, agent continues normally |
| Orphaned pending state from previous request | New request fails with `IllegalStateException` | Auto-generates error results, clears pending state, processes normally |

### Log Indicators

After the fix, you should see these log messages instead of exceptions:

1. **Tool execution failure recovery:**
   ```
   ERROR - Tool execution failed, generating error results for X tool calls: <error message>
   ```

2. **Orphaned pending state recovery:**
   ```
   WARN  - Pending tool calls detected without results, auto-generating error results. Pending IDs: [...]
   INFO  - Auto-generated error result for pending tool call: <tool_name> (<tool_id>)
   ```

### Test Scenarios

1. **Timeout Test:** Configure a short `TOOL_TIMEOUT` and invoke a long-running tool
   - Expected: Agent returns error message instead of crashing

2. **Recovery Test:** Manually interrupt a tool execution mid-way, then send a new message
   - Expected: Agent auto-recovers and processes the new message

## Files Changed

- `src/main/java/io/agentscope/core/ReActAgent.java`
  - Modified: `executeToolCalls()` - added `onErrorResume` error handling
  - Modified: `doCall()` - added pending state detection and auto-recovery
  - Added: `generateAndAddErrorToolResults()` - new helper method

## Compatibility

- **Backward Compatible:** Yes
- **API Changes:** None
- **Breaking Changes:** None

The fix only adds error recovery logic without changing the normal execution path.

chensk0601:
…n (#951)

ReActAgent throws IllegalStateException when tool calls timeout or fail, because no tool result is written to memory, leaving orphaned pending tool call states that crash the agent on subsequent requests.

Root cause:
- Tool execution timeout/error propagates without writing results to memory
- Pending tool call state remains, blocking subsequent doCall() invocations
- validateAndAddToolResults() throws when user message has no tool results

Changes:
- doCall(): detect pending tool calls without user-provided results and auto-generate error results to clear the pending state
- executeToolCalls(): add onErrorResume to catch tool execution failures and generate error tool results instead of propagating exceptions
- Add generateAndAddErrorToolResults() helper to create error results for orphaned pending tool calls

This ensures the agent recovers gracefully from tool failures instead of crashing, and the model receives proper error feedback to continue processing.

Closes #951

## AgentScope-Java Version

[The version of AgentScope-Java you are working on, e.g. 1.0.9, check your pom.xml dependency version or run `mvn dependency:tree | grep agentscope-parent:pom`(only mac/linux)]

## Description

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -30,6 +30,7 @@ import io.agentscope.core.hook.PreSummaryEvent;
 import io.agentscope.core.hook.ReasoningChunkEvent;
 import io.agentscope.core.hook.SummaryChunkEvent;
 import io.agentscope.core.interruption.InterruptContext;
+import io.agentscope.core.interruption.InterruptSource;
 import io.agentscope.core.memory.InMemoryMemory;
 import io.agentscope.core.memory.LongTermMemory;
 import io.agentscope.core.memory.LongTermMemoryMode;
@@ -57,6 +58,9 @@ import io.agentscope.core.rag.RAGMode;
 import io.agentscope.core.rag.model.Document;
 import io.agentscope.core.rag.model.RetrieveConfig;
 import io.agentscope.core.session.Session;
+import io.agentscope.core.shutdown.AgentShuttingDownException;
+import io.agentscope.core.shutdown.GracefulShutdownManager;
+import io.agentscope.core.shutdown.PartialReasoningPolicy;
 import io.agentscope.core.skill.SkillBox;
 import io.agentscope.core.skill.SkillHook;
 import io.agentscope.core.state.AgentMetaState;
@@ -132,6 +136,8 @@ import reactor.core.publisher.Mono;
 public class ReActAgent extends StructuredOutputCapableAgent {
 
     private static final Logger log = LoggerFactory.getLogger(ReActAgent.class);
+    private static final GracefulShutdownManager shutdownManager =
+            GracefulShutdownManager.getInstance();
 
     // ==================== Core Dependencies ====================
 
@@ -227,7 +233,14 @@ public class ReActAgent extends StructuredOutputCapableAgent {
      * @param sessionKey the session identifier
      */
     @Override
+    public boolean loadIfExists(Session session, SessionKey sessionKey) {
+        shutdownManager.bindSession(this, session, sessionKey);
+        return super.loadIfExists(session, sessionKey);
+    }
+
+    @Override
     public void loadFrom(Session session, SessionKey sessionKey) {
+        shutdownManager.bindSession(this, session, sessionKey);
         // Load memory if managed
         if (statePersistence.memoryManaged()) {
             memory.loadFrom(session, sessionKey);
@@ -478,10 +491,19 @@ public class ReActAgent extends StructuredOutputCapableAgent {
                 .onErrorResume(
                         InterruptedException.class,
                         error -> {
-                            // Save accumulated message before propagating interrupt
                             Msg msg = context.buildFinalMessage();
                             if (msg != null) {
-                                memory.addMessage(msg);
+                                boolean discard =
+                                        getInterruptSource() == InterruptSource.SYSTEM
+                                                && shutdownManager
+                                                                .getConfig()
+                                                                .partialReasoningPolicy()
+                                                        == PartialReasoningPolicy.DISCARD;
+                                // Manually interruption will save the msg, while system
+                                // interruption will discard on specific config
+                                if (!discard) {
+                                    memory.addMessage(msg);
+                                }
                             }
                             return Mono.error(error);
                         })
@@ -1004,6 +1026,11 @@ public class ReActAgent extends StructuredOutputCapableAgent {
 
     @Override
     protected Mono<Msg> handleInterrupt(InterruptContext context, Msg... originalArgs) {
+        if (context.getSource() == InterruptSource.SYSTEM) {
+            shutdownManager.saveOnInterruptObserved(this);
+            return Mono.error(new AgentShuttingDownException());
+        }
+
         String recoveryText = "I noticed that you have interrupted me. What can I do for you?";
 
         Msg recoveryMsg =
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
@@ -23,6 +23,8 @@ import io.agentscope.core.hook.PreCallEvent;
 import io.agentscope.core.interruption.InterruptContext;
 import io.agentscope.core.interruption.InterruptSource;
 import io.agentscope.core.message.Msg;
+import io.agentscope.core.shutdown.GracefulShutdownHook;
+import io.agentscope.core.shutdown.GracefulShutdownManager;
 import io.agentscope.core.state.StateModule;
 import io.agentscope.core.tracing.TracerRegistry;
 import java.util.ArrayList;
@@ -93,7 +95,9 @@ public abstract class AgentBase implements StateModule, Agent {
     private final AtomicBoolean running = new AtomicBoolean(false);
     private final boolean checkRunning;
     private final List<Hook> hooks;
-    private static final List<Hook> systemHooks = new CopyOnWriteArrayList<>();
+    private static final List<Hook> systemHooks =
+            new CopyOnWriteArrayList<>(
+                    List.of(new GracefulShutdownHook(GracefulShutdownManager.getInstance())));
     private final Map<String, List<AgentBase>> hubSubscribers = new ConcurrentHashMap<>();
 
     // Interrupt state management (available to all agents)
@@ -101,6 +105,8 @@ public abstract class AgentBase implements StateModule, Agent {
     private final AtomicReference<Msg> userInterruptMessage = new AtomicReference<>(null);
     // Hook non-null
     private static final Comparator<Hook> HOOK_COMPARATOR = Comparator.comparingInt(Hook::priority);
+    private final AtomicReference<InterruptSource> interruptSource =
+            new AtomicReference<>(InterruptSource.USER);
 
     /**
      * Constructor for AgentBase.
@@ -165,14 +171,7 @@ public abstract class AgentBase implements StateModule, Agent {
     @Override
     public final Mono<Msg> call(List<Msg> msgs) {
         return Mono.using(
-                () -> {
-                    if (checkRunning && !running.compareAndSet(false, true)) {
-                        throw new IllegalStateException(
-                                "Agent is still running, please wait for it to finish");
-                    }
-                    resetInterruptFlag();
-                    return this;
-                },
+                this::acquireExecution,
                 resource ->
                         TracerRegistry.get()
                                 .callAgent(
@@ -185,7 +184,7 @@ public abstract class AgentBase implements StateModule, Agent {
                                                         .onErrorResume(
                                                                 createErrorHandler(
                                                                         msgs.toArray(new Msg[0])))),
-                resource -> running.set(false),
+                this::releaseExecution,
                 true);
     }
 
@@ -201,14 +200,7 @@ public abstract class AgentBase implements StateModule, Agent {
     @Override
     public final Mono<Msg> call(List<Msg> msgs, Class<?> structuredOutputClass) {
         return Mono.using(
-                () -> {
-                    if (checkRunning && !running.compareAndSet(false, true)) {
-                        throw new IllegalStateException(
-                                "Agent is still running, please wait for it to finish");
-                    }
-                    resetInterruptFlag();
-                    return this;
-                },
+                this::acquireExecution,
                 resource ->
                         TracerRegistry.get()
                                 .callAgent(
@@ -225,7 +217,7 @@ public abstract class AgentBase implements StateModule, Agent {
                                                         .onErrorResume(
                                                                 createErrorHandler(
                                                                         msgs.toArray(new Msg[0])))),
-                resource -> running.set(false),
+                this::releaseExecution,
                 true);
     }
 
@@ -241,14 +233,7 @@ public abstract class AgentBase implements StateModule, Agent {
     @Override
     public final Mono<Msg> call(List<Msg> msgs, JsonNode schema) {
         return Mono.using(
-                () -> {
-                    if (checkRunning && !running.compareAndSet(false, true)) {
-                        throw new IllegalStateException(
-                                "Agent is still running, please wait for it to finish");
-                    }
-                    resetInterruptFlag();
-                    return this;
-                },
+                this::acquireExecution,
                 resource ->
                         TracerRegistry.get()
                                 .callAgent(
@@ -261,7 +246,7 @@ public abstract class AgentBase implements StateModule, Agent {
                                                         .onErrorResume(
                                                                 createErrorHandler(
                                                                         msgs.toArray(new Msg[0])))),
-                resource -> running.set(false),
+                this::releaseExecution,
                 true);
     }
 
@@ -318,6 +303,7 @@ public abstract class AgentBase implements StateModule, Agent {
      */
     @Override
     public void interrupt() {
+        interruptSource.set(InterruptSource.USER);
         interruptFlag.set(true);
     }
 
@@ -329,10 +315,21 @@ public abstract class AgentBase implements StateModule, Agent {
      */
     @Override
     public void interrupt(Msg msg) {
+        interruptSource.set(InterruptSource.USER);
         interruptFlag.set(true);
         if (msg != null) {
             userInterruptMessage.set(msg);
         }
+    }
+
+    /**
+     * Interrupt execution with explicit source.
+     *
+     * @param source interruption source
+     */
+    public void interrupt(InterruptSource source) {
+        interruptSource.set(source != null ? source : InterruptSource.SYSTEM);
+        interruptFlag.set(true);
     }
 
     /**
@@ -376,6 +373,7 @@ public abstract class AgentBase implements StateModule, Agent {
     protected void resetInterruptFlag() {
         interruptFlag.set(false);
         userInterruptMessage.set(null);
+        interruptSource.set(InterruptSource.USER);
     }
 
     /**
@@ -386,9 +384,45 @@ public abstract class AgentBase implements StateModule, Agent {
      */
     private InterruptContext createInterruptContext() {
         return InterruptContext.builder()
-                .source(InterruptSource.USER)
+                .source(interruptSource.get())
                 .userMessage(userInterruptMessage.get())
                 .build();
+    }
+
+    /**
+     * Acquire execution resources for a {@code call()} invocation.
+     * Used as the {@code resourceSupplier} in {@link Mono#using} to guarantee that
+     * {@link #releaseExecution} is always called on completion, error, or cancellation.
+     *
+     * @return this agent instance
+     */
+    private AgentBase acquireExecution() {
+        if (checkRunning && !running.compareAndSet(false, true)) {
+            throw new IllegalStateException("Agent is still running, please wait for it to finish");
+        }
+        try {
+            resetInterruptFlag();
+            GracefulShutdownManager.getInstance().ensureAcceptingRequests();
+            GracefulShutdownManager.getInstance().registerRequest(this);
+        } catch (RuntimeException ex) {
+            if (checkRunning) {
+                running.set(false);
+            }
+            throw ex;
+        }
+        return this;
+    }
+
+    /**
+     * Release execution resources after a {@code call()} invocation.
+     * Used as the {@code resourceCleanup} in {@link Mono#using} — guaranteed to run
+     * regardless of how the reactive chain terminates (success, error, or cancel).
+     *
+     * @param resource the agent instance (ignored, uses {@code this})
+     */
+    private void releaseExecution(AgentBase resource) {
+        running.set(false);
+        GracefulShutdownManager.getInstance().unregisterRequest(this);
     }
 
     /**
@@ -418,6 +452,15 @@ public abstract class AgentBase implements StateModule, Agent {
      */
     protected AtomicBoolean getInterruptFlag() {
         return interruptFlag;
+    }
+
+    /**
+     * Get current interruption source.
+     *
+     * @return interruption source
+     */
+    protected InterruptSource getInterruptSource() {
+        return interruptSource.get();
     }
 
     /**
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/dashscope/dto/DashScopeMessage.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/dashscope/dto/DashScopeMessage.java
@@ -15,6 +15,7 @@
  */
 package io.agentscope.core.formatter.dashscope.dto;
 
+import com.fasterxml.jackson.annotation.JsonAlias;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
@@ -79,6 +80,7 @@ public class DashScopeMessage {
 
     /** Reasoning/thinking content (for assistant messages with thinking enabled). */
     @JsonProperty("reasoning_content")
+    @JsonAlias("reasoning")
     private String reasoningContent;
 
     /** Cache control configuration for prompt caching. */
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/dto/OpenAIMessage.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/dto/OpenAIMessage.java
@@ -15,6 +15,7 @@
  */
 package io.agentscope.core.formatter.openai.dto;
 
+import com.fasterxml.jackson.annotation.JsonAlias;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
@@ -76,10 +77,12 @@ public class OpenAIMessage {
     private List<OpenAIToolCall> toolCalls;
 
     /**
-     * Reasoning/thinking content (for o1 and similar reasoning models).
-     * This field contains the model's internal reasoning process.
+     * Reasoning/thinking content from the model.
+     * Accepts both {@code reasoning_content} (commercial APIs like DeepSeek, DashScope) and
+     * {@code reasoning} (vLLM deployments) during deserialization.
      */
     @JsonProperty("reasoning_content")
+    @JsonAlias("reasoning")
     private String reasoningContent;
 
     /**
--- a/agentscope-core/src/main/java/io/agentscope/core/model/transport/HttpTransportFactory.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/transport/HttpTransportFactory.java
@@ -47,8 +47,9 @@ import org.slf4j.LoggerFactory;
  * HttpTransportFactory.register(myTransport);
  * }</pre>
  *
- * <p>All registered transports and the default transport will be automatically
- * closed when the JVM shuts down.
+ * <p>All registered transports and the default transport will be closed
+ * during JVM shutdown, orchestrated by {@code AgentScopeJvmShutdownHook}
+ * after all agents have finished processing.
  */
 public final class HttpTransportFactory {
 
@@ -59,9 +60,6 @@ public final class HttpTransportFactory {
 
     /** List of all managed transports for cleanup. */
     private static final List<HttpTransport> managedTransports = new CopyOnWriteArrayList<>();
-
-    /** Whether the shutdown hook has been registered. */
-    private static volatile boolean shutdownHookRegistered = false;
 
     /** Lock object for synchronization. */
     private static final Object lock = new Object();
@@ -85,7 +83,6 @@ public final class HttpTransportFactory {
                 if (defaultTransport == null) {
                     defaultTransport = new JdkHttpTransport();
                     managedTransports.add(defaultTransport);
-                    registerShutdownHookIfNeeded();
                     log.debug("Created default HttpTransport: {}", defaultTransport);
                 }
             }
@@ -107,7 +104,6 @@ public final class HttpTransportFactory {
             if (transport != null && !managedTransports.contains(transport)) {
                 managedTransports.add(transport);
             }
-            registerShutdownHookIfNeeded();
             log.debug("Set default HttpTransport: {}", transport);
         }
     }
@@ -124,7 +120,6 @@ public final class HttpTransportFactory {
     public static void register(HttpTransport transport) {
         if (transport != null && !managedTransports.contains(transport)) {
             managedTransports.add(transport);
-            registerShutdownHookIfNeeded();
             log.debug("Registered HttpTransport for management: {}", transport);
         }
     }
@@ -144,30 +139,6 @@ public final class HttpTransportFactory {
             log.debug("Unregistered HttpTransport from management: {}", transport);
         }
         return removed;
-    }
-
-    /**
-     * Register the JVM shutdown hook if not already registered.
-     */
-    private static void registerShutdownHookIfNeeded() {
-        if (!shutdownHookRegistered) {
-            synchronized (lock) {
-                if (!shutdownHookRegistered) {
-                    Runtime.getRuntime()
-                            .addShutdownHook(
-                                    new Thread(
-                                            () -> {
-                                                log.debug(
-                                                        "Shutdown hook triggered, closing all"
-                                                                + " HttpTransports");
-                                                shutdown();
-                                            },
-                                            "HttpTransportFactory-ShutdownHook"));
-                    shutdownHookRegistered = true;
-                    log.debug("Registered HttpTransport shutdown hook");
-                }
-            }
-        }
     }
 
     /**
--- a/agentscope-core/src/main/java/io/agentscope/core/session/InMemorySession.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/session/InMemorySession.java
@@ -170,6 +170,15 @@ public class InMemorySession implements Session {
         sessions.remove(sessionKeyStr);
     }
 
+    @Override
+    public void delete(SessionKey sessionKey, String key) {
+        String sessionKeyStr = serializeSessionKey(sessionKey);
+        SessionData data = sessions.get(sessionKeyStr);
+        if (data != null) {
+            data.removeSingleState(key);
+        }
+    }
+
     /**
      * List all session keys.
      *
@@ -222,6 +231,10 @@ public class InMemorySession implements Session {
 
         State getSingleState(String key) {
             return singleStates.get(key);
+        }
+
+        void removeSingleState(String key) {
+            singleStates.remove(key);
         }
 
         void setListState(String key, List<? extends State> values) {
--- a/agentscope-core/src/main/java/io/agentscope/core/session/JsonSession.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/session/JsonSession.java
@@ -298,6 +298,15 @@ public class JsonSession implements Session {
         }
     }
 
+    @Override
+    public void delete(SessionKey sessionKey, String key) {
+        try {
+            Files.deleteIfExists(getStatePath(sessionKey, key));
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to delete state: " + key, e);
+        }
+    }
+
     /**
      * List all session keys.
      *
--- a/agentscope-core/src/main/java/io/agentscope/core/session/Session.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/session/Session.java
@@ -119,6 +119,16 @@ public interface Session {
     void delete(SessionKey sessionKey);
 
     /**
+     * Delete a single state entry within a session.
+     *
+     * @param sessionKey the session identifier
+     * @param key the state key to delete
+     */
+    default void delete(SessionKey sessionKey, String key) {
+        // Default no-op; implementations should override if they support per-key deletion
+    }
+
+    /**
      * List all session keys.
      *
      * @return set of all session keys
--- a/agentscope-core/src/main/java/io/agentscope/core/shutdown/ActiveRequestContext.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/shutdown/ActiveRequestContext.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.shutdown;
+
+import io.agentscope.core.agent.AgentBase;
+import io.agentscope.core.interruption.InterruptSource;
+import io.agentscope.core.session.Session;
+import io.agentscope.core.state.SessionKey;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Runtime context for a single active request.
+ */
+final class ActiveRequestContext {
+
+    private static final Logger log = LoggerFactory.getLogger(ActiveRequestContext.class);
+
+    private final String requestId;
+    private final AgentBase agent;
+    private final AtomicBoolean shutdownInterruptIssued = new AtomicBoolean(false);
+
+    private final Session session;
+    private final SessionKey sessionKey;
+
+    ActiveRequestContext(
+            String requestId, AgentBase agent, Session session, SessionKey sessionKey) {
+        this.requestId = requestId;
+        this.agent = agent;
+        this.session = session;
+        this.sessionKey = sessionKey;
+    }
+
+    String getRequestId() {
+        return requestId;
+    }
+
+    boolean hasSessionBinding() {
+        return session != null && sessionKey != null;
+    }
+
+    static final String SHUTDOWN_INTERRUPTED_KEY = "shutdown_interrupted";
+
+    void saveToSession() {
+        if (!hasSessionBinding()) {
+            return;
+        }
+        try {
+            agent.saveTo(session, sessionKey);
+            session.save(sessionKey, SHUTDOWN_INTERRUPTED_KEY, new ShutdownInterruptedState(true));
+        } catch (Exception e) {
+            log.warn("Failed to save agent state for request {}", requestId, e);
+        }
+    }
+
+    boolean interruptForShutdown() {
+        if (!shutdownInterruptIssued.compareAndSet(false, true)) {
+            return false;
+        }
+        agent.interrupt(InterruptSource.SYSTEM);
+        return true;
+    }
+}
--- a/agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentScopeJvmShutdownHook.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentScopeJvmShutdownHook.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.shutdown;
+
+import io.agentscope.core.model.transport.HttpTransportFactory;
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * JVM hook for SIGTERM graceful shutdown entry.
+ *
+ * <p>Shutdown order:
+ * <ol>
+ *   <li>Trigger graceful shutdown (reject new requests, interrupt at checkpoints)</li>
+ *   <li>Wait for all active agent requests to complete or timeout</li>
+ *   <li>Close HTTP transports (only after agents are done)</li>
+ * </ol>
+ */
+public final class AgentScopeJvmShutdownHook {
+
+    private static final Logger log = LoggerFactory.getLogger(AgentScopeJvmShutdownHook.class);
+    private static final AtomicBoolean REGISTERED = new AtomicBoolean(false);
+
+    private static final Duration INTERRUPT_GRACE_PERIOD = Duration.ofSeconds(5);
+
+    private AgentScopeJvmShutdownHook() {}
+
+    public static void register(GracefulShutdownManager manager) {
+        if (!REGISTERED.compareAndSet(false, true)) {
+            return;
+        }
+        Runtime.getRuntime()
+                .addShutdownHook(
+                        new Thread(
+                                () -> {
+                                    try {
+                                        manager.performGracefulShutdown();
+                                        Duration timeout = manager.getConfig().shutdownTimeout();
+                                        // Wait for shutdown timeout + an extra grace period,
+                                        // so agents have time to handle the interrupt and
+                                        // clean up before HTTP transports are closed.
+                                        Duration awaitTimeout =
+                                                timeout != null
+                                                        ? timeout.plus(INTERRUPT_GRACE_PERIOD)
+                                                        : null;
+                                        manager.awaitTermination(awaitTimeout);
+                                    } catch (Exception e) {
+                                        log.warn("Graceful shutdown hook failed", e);
+                                    } finally {
+                                        HttpTransportFactory.shutdown();
+                                    }
+                                },
+                                "agentscope-jvm-shutdown-hook"));
+    }
+}
--- a/agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentShuttingDownException.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentShuttingDownException.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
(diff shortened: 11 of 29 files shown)


agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -602,7 +669,41 @@ private Mono<List<Map.Entry<ToolUseBlock, ToolResultBlock>>> executeToolCalls(
...
+                                                                                        TextBlock
+                                                                                                .builder()
+                                                                                                .text(
+                                                                                                        "[ERROR]"
+                                                                                                            + " Tool"
+                                                                                                            + " execution"
+                                                                                                            + " failed:"
+                                                                                                            + " "
+                                                                                                                + error
+                                                                                                                        .getMessage())
```
LearningGp: error.getMessage() might return null, which could result in an unhelpful error message like [ERROR] Tool execution failed: null. Since ToolExecutor.executeWithInfrastructure() already uses ExceptionUtils.getErrorMessage(e), it might be better to align with that approach here for the sake of consistency.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -252,9 +252,72 @@ protected Mono<Msg> doCall(List<Msg> msgs) {
...
+    }
+
+    /**
+     * Generate error tool results for pending tool calls and add them to memory.
+     * This is used to recover from situations where tool execution failed without
+     * properly writing results to memory.
+     *
+     * @param pendingIds The set of pending tool use IDs
+     */
+    private void generateAndAddErrorToolResults(Set<String> pendingIds) {
```
LearningGp: The logic for manually constructing ToolResultBlock in both generateAndAddErrorToolResults and executeToolCalls.onErrorResume is nearly identical. I’d suggest extracting this into a shared helper method to reduce duplication and ensure a consistent error message format across the board.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -602,7 +669,41 @@ private Mono<List<Map.Entry<ToolUseBlock, ToolResultBlock>>> executeToolCalls(
                         results ->
                                 IntStream.range(0, toolCalls.size())
                                         .mapToObj(i -> Map.entry(toolCalls.get(i), results.get(i)))
-                                        .toList());
+                                        .toList())
+                .onErrorResume(
+                        error -> {
```
LearningGp, line 676: Unexpected errors from the Reactor infrastructure itself (such as OutOfMemoryError) are currently being silently converted into error results. I suggest narrowing the catch scope to exclude Error subclasses, allowing critical JVM failures to propagate as intended.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
chensk0601: create new function buildErrorToolResult to do this common work

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -252,9 +252,72 @@ protected Mono<Msg> doCall(List<Msg> msgs) {
...
+                    ToolResultMessageBuilder.buildToolResultMsg(errorResult, toolCall, getName());
+            memory.addMessage(toolResultMsg);
+            log.info(
+                    "Auto-generated error result for pending tool call: {} ({})",
+                    toolCall.getName(),
+                    toolCall.getId());
+        }
     }
 
     /**
```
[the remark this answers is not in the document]
chensk0601: use PostActingEvent to add tool result message.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
chensk0601: use Exception.class to replace of the Throwable Object

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -602,7 +669,41 @@ private Mono<List<Map.Entry<ToolUseBlock, ToolResultBlock>>> executeToolCalls(
...
-                                        .toList());
+                                        .toList())
+                .onErrorResume(
+                        error -> {
+                            // Generate error tool results for all pending tool calls
+                            log.error(
+                                    "Tool execution failed, generating error results for {} tool"
+                                            + " calls: {}",
+                                    toolCalls.size(),
+                                    error.getMessage());
```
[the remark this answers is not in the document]
chensk0601: have fixed it

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -602,7 +669,41 @@ private Mono<List<Map.Entry<ToolUseBlock, ToolResultBlock>>> executeToolCalls(
...
+                                                        ToolResultBlock errorResult =
+                                                                ToolResultBlock.builder()
+                                                                        .id(toolCall.getId())
+                                                                        .output(
+                                                                                List.of(
+                                                                                        TextBlock
+                                                                                                .builder()
+                                                                                                .text(
+                                                                                                        "[ERROR]"
+                                                                                                            + " Tool"
```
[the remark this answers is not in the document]
chensk0601: have fixed it

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
chensk0601: have fixed it

agentscope-core/src/test/java/io/agentscope/core/hook/HookStopAgentTest.java
```diff
@@ -368,15 +372,11 @@ void testNewMsgWithPendingToolUseContinuesActing() {
...
             Msg newMsg = createUserMsg("new message");
+            Msg result2 = agent.call(newMsg).block(TEST_TIMEOUT);
 
-            StepVerifier.create(agent.call(newMsg))
-                    .expectErrorMatches(
-                            e ->
-                                    e instanceof IllegalStateException
-                                            && e.getMessage().contains("pending tool calls"))
-                    .verify();
+            assertNotNull(result2, "Agent should auto-recover and return a result");
```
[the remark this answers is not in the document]
chensk0601: have enhanced this test case

agentscope-core/src/test/java/io/agentscope/core/hook/HookStopAgentTest.java
```diff
@@ -661,14 +665,10 @@ void testAgentHandlesPendingToolCallsGracefully() {
...
-            StepVerifier.create(agent.call(createUserMsg("new")))
-                    .expectErrorMatches(
-                            e ->
-                                    e instanceof IllegalStateException
-                                            && e.getMessage().contains("pending tool calls"))
-                    .verify();
+            // With new design, agent will auto-recover by generating error results
+            // for pending tool calls and continue processing
+            Msg result = agent.call(createUserMsg("new")).block(TEST_TIMEOUT);
+            assertNotNull(result, "Agent should auto-recover and return a result");
```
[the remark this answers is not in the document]
chensk0601: have fixed it

chensk0601:
> The changes to executeToolCalls look solid, as they allow the model to proceed even after a tool call exception. However, I’m not sure about adding automatic recovery to doCall. In my view, it might be better to surface these exceptions to the framework consumers instead. Open to discussion on this point!

This fix not only ensures execution continues after tool call timeouts and exceptions, but also resolves a critical issue where the entire agent would become unusable following a timeout, repeatedly throwing IllegalStateException. I have been running this updated version in our production environment for some time now, and it has been performing flawlessly.

LearningGp:
> > The changes to executeToolCalls look solid, as they allow the model to proceed even after a tool call exception. However, I’m not sure about adding automatic recovery to doCall. In my view, it might be better to surface these exceptions to the framework consumers instead. Open to discussion on this point!
> 
> This fix not only ensures execution continues after tool call timeouts and exceptions, but also resolves a critical issue where the entire agent would become unusable following a timeout, repeatedly throwing IllegalStateException. I have been running this updated version in our production environment for some time now, and it has been performing flawlessly.

It seems that the modifications to executeToolCalls should be sufficient to ensure a ToolResultBlock is present following a timeout or exception, which would prevent the persistent IllegalStateException.

However, I have some reservations about the auto-supplement logic in doCall. Since HITL (Human-In-The-Loop) workflows require a ToolResultBlock to be manually provided when resuming a conversation, automating this process might lead to unintended execution paths. It could also potentially mask improper HITL usage, making it harder to detect implementation errors.

It’s great to hear that this fix has been verified in production! A couple of follow-up questions:

- Have we encountered specific scenarios where the auto-supplement logic in doCall was actually triggered?

- In our current context, would it be possible to achieve the same goal by only modifying executeToolCalls?

chensk0601:
> > > The changes to executeToolCalls look solid, as they allow the model to proceed even after a tool call exception. However, I’m not sure about adding automatic recovery to doCall. In my view, it might be better to surface these exceptions to the framework consumers instead. Open to discussion on this point!
> > 
> > 
> > This fix not only ensures execution continues after tool call timeouts and exceptions, but also resolves a critical issue where the entire agent would become unusable following a timeout, repeatedly throwing IllegalStateException. I have been running this updated version in our production environment for some time now, and it has been performing flawlessly.
> 
> It seems that the modifications to executeToolCalls should be sufficient to ensure a ToolResultBlock is present following a timeout or exception, which would prevent the persistent IllegalStateException.
> 
> However, I have some reservations about the auto-supplement logic in doCall. Since HITL (Human-In-The-Loop) workflows require a ToolResultBlock to be manually provided when resuming a conversation, automating this process might lead to unintended execution paths. It could also potentially mask improper HITL usage, making it harder to detect implementation errors.
> 
> It’s great to hear that this fix has been verified in production! A couple of follow-up questions:
> 
> * Have we encountered specific scenarios where the auto-supplement logic in doCall was actually triggered?
> * In our current context, would it be possible to achieve the same goal by only modifying executeToolCalls?

Thanks for the thorough review! I appreciate the concern about HITL workflows — preserving the explicit contract for manual `ToolResultBlock` provision is indeed important.

Let me address your questions:

## Regarding scenarios where `doCall` auto-supplement was triggered

While `executeToolCalls`'s `onErrorResume` handles failures *within* tool execution, we've identified cases where pending states persist **outside** that coverage:

1. **Pre-execution failures**: Network timeouts before `toolkit.callTools()` begins, JVM crashes, or async scheduling failures where tool calls are queued but never executed
2. **Memory recovery**: When agent state is persisted and restored (e.g., after restart), orphaned pending IDs may exist without corresponding results

In these cases, `executeToolCalls` is never reached, so its error handling cannot apply. The `doCall` auto-recovery acts as a safety net for these edge cases.

## On achieving the goal with only `executeToolCalls` modifications

For failures occurring *during* `executeToolCalls`, yes — the `onErrorResume` logic is sufficient. However, for the scenarios above, we need the `doCall` layer to handle states that originate from outside the tool execution flow.

## Proposed compromise to protect HITL workflows

To address your concern about masking improper HITL usage, I suggest adding an explicit mode check:

```java
if (providedResults.isEmpty()) {
    if (isHITLMode()) {  // Check if conversation is in HITL pause state
        // Strict behavior: require manual ToolResultBlock
            "HITL workflow requires manual ToolResultBlock for pending IDs: " + pendingIds);
    }
    // Non-HITL: Auto-recover from orphaned states
    log.warn("Orphaned pending tool calls detected, auto-generating error results: {}", pendingIds);
    generateAndAddErrorToolResults(pendingIds);
}

**Comment by chensk0601** [comment]
> > > > The changes to executeToolCalls look solid, as they allow the model to proceed even after a tool call exception. However, I’m not sure about adding automatic recovery to doCall. In my view, it might be better to surface these exceptions to the framework consumers instead. Open to discussion on this point!
> > > 
> > > 
> > > This fix not only ensures execution continues after tool call timeouts and exceptions, but also resolves a critical issue where the entire agent would become unusable following a timeout, repeatedly throwing IllegalStateException. I have been running this updated version in our production environment for some time now, and it has been performing flawlessly.
> > 
> > 
> > It seems that the modifications to executeToolCalls should be sufficient to ensure a ToolResultBlock is present following a timeout or exception, which would prevent the persistent IllegalStateException.
> > However, I have some reservations about the auto-supplement logic in doCall. Since HITL (Human-In-The-Loop) workflows require a ToolResultBlock to be manually provided when resuming a conversation, automating this process might lead to unintended execution paths. It could also potentially mask improper HITL usage, making it harder to detect implementation errors.
> > It’s great to hear that this fix has been verified in production! A couple of follow-up questions:
> > 
> > * Have we encountered specific scenarios where the auto-supplement logic in doCall was actually triggered?
> > * In our current context, would it be possible to achieve the same goal by only modifying executeToolCalls?
> 
> Thanks for the thorough review! I appreciate the concern about HITL workflows — preserving the explicit contract for manual `ToolResultBlock` provision is indeed important.
> 
> Let me address your questions:
> 
> ## Regarding scenarios where `doCall` auto-supplement was triggered
> While `executeToolCalls`'s `onErrorResume` handles failures _within_ tool execution, we've identified cases where pending states persist **outside** that coverage:
> 
> 1. **Pre-execution failures**: Network timeouts before `toolkit.callTools()` begins, JVM crashes, or async scheduling failures where tool calls are queued but never executed
> 2. **Memory recovery**: When agent state is persisted and restored (e.g., after restart), orphaned pending IDs may exist without corresponding results
> 
> In these cases, `executeToolCalls` is never reached, so its error handling cannot apply. The `doCall` auto-recovery acts as a safety net for these edge cases.
> 
> ## On achieving the goal with only `executeToolCalls` modifications
> For failures occurring _during_ `executeToolCalls`, yes — the `onErrorResume` logic is sufficient. However, for the scenarios above, we need the `doCall` layer to handle states that originate from outside the tool execution flow.
> 
> ## Proposed compromise to protect HITL workflows
> To address your concern about masking improper HITL usage, I suggest adding an explicit mode check:
> 
> ```java
> if (providedResults.isEmpty()) {
>     if (isHITLMode()) {  // Check if conversation is in HITL pause state
>         // Strict behavior: require manual ToolResultBlock
>         throw new IllegalStateException(
>             "HITL workflow requires manual ToolResultBlock for pending IDs: " + pendingIds);
>     }
>     // Non-HITL: Auto-recover from orphaned states
>     log.warn("Orphaned pending tool calls detected, auto-generating error results: {}", pendingIds);
>     generateAndAddErrorToolResults(pendingIds);
> }
> ```

Thanks for the thorough review! I've carefully reconsidered this, and I believe this is fundamentally a **bug fix** rather than an enhancement — adding configuration parameters would not be the right approach.

## Why this is a critical bug (not just an enhancement)

The current behavior causes **complete conversation failure** on any tool execution error:

1. **First failure**: Tool timeout/exception → `IllegalStateException` → Agent crashes
2. **Cascading effect**: Pending state persists in memory  
3. **All subsequent requests**: Fail with same exception, making the agent unusable

This violates basic fault isolation principles — a single tool failure should not crash the entire agent permanently.

## Two key reasons why auto-supplement is necessary

**First**, tool execution failures should not leave the agent in a permanently broken state. Whether it's a timeout, network error, or unexpected exception, the agent must recover and continue functioning. The "auto-supplement" in `doCall` serves as a critical safety net for orphaned pending states that originate **outside** `executeToolCalls` coverage (e.g., JVM crashes, memory recovery, pre-execution failures).

**Second**, manually generating error results and feeding them to the LLM is **essential for proper decision-making**. Without this feedback, the model has no visibility into what happened. By providing explicit error messages (e.g., *"[ERROR] Tool execution failed: timeout"*), we enable the model to:
- Understand the failure context
- Decide on alternative approaches  
- Communicate meaningfully with the user

This is not masking errors — it's **proper error propagation** to the model layer.

## Regarding HITL concerns

The HITL workflow concern is valid, but I'd argue:
- HITL pause/resume should be handled at a **higher orchestration layer**, not by leaving the agent in a corrupted state
- If HITL requires manual `ToolResultBlock`, it should explicitly manage the conversation lifecycle (clearing/resuming state) rather than relying on low-level exceptions

## Summary

| Approach | Result |
|---------|--------|
| Only `executeToolCalls` fix | Covers runtime failures, but leaves recovery gaps |
| Configuration switches | Adds complexity without solving the root problem |
| **Current PR (both fixes)** | Complete fault isolation + proper LLM feedback |

The combination ensures robust error handling while keeping the implementation clean and deterministic. I'm happy to discuss alternative HITL integration patterns if needed, but I believe the core fix should remain as-is.

Would appreciate your thoughts on this perspective.

**Comment by LearningGp** [comment]
> > > > > The changes to executeToolCalls look solid, as they allow the model to proceed even after a tool call exception. However, I’m not sure about adding automatic recovery to doCall. In my view, it might be better to surface these exceptions to the framework consumers instead. Open to discussion on this point!
> > > > 
> > > > 
> > > > This fix not only ensures execution continues after tool call timeouts and exceptions, but also resolves a critical issue where the entire agent would become unusable following a timeout, repeatedly throwing IllegalStateException. I have been running this updated version in our production environment for some time now, and it has been performing flawlessly.
> > > 
> > > 
> > > It seems that the modifications to executeToolCalls should be sufficient to ensure a ToolResultBlock is present following a timeout or exception, which would prevent the persistent IllegalStateException.
> > > However, I have some reservations about the auto-supplement logic in doCall. Since HITL (Human-In-The-Loop) workflows require a ToolResultBlock to be manually provided when resuming a conversation, automating this process might lead to unintended execution paths. It could also potentially mask improper HITL usage, making it harder to detect implementation errors.
> > > It’s great to hear that this fix has been verified in production! A couple of follow-up questions:
> > > 
> > > * Have we encountered specific scenarios where the auto-supplement logic in doCall was actually triggered?
> > > * In our current context, would it be possible to achieve the same goal by only modifying executeToolCalls?
> > 
> > 
> > Thanks for the thorough review! I appreciate the concern about HITL workflows — preserving the explicit contract for manual `ToolResultBlock` provision is indeed important.
> > Let me address your questions:
> > ## Regarding scenarios where `doCall` auto-supplement was triggered
> > While `executeToolCalls`'s `onErrorResume` handles failures _within_ tool execution, we've identified cases where pending states persist **outside** that coverage:
> > 
> > 1. **Pre-execution failures**: Network timeouts before `toolkit.callTools()` begins, JVM crashes, or async scheduling failures where tool calls are queued but never executed
> > 2. **Memory recovery**: When agent state is persisted and restored (e.g., after restart), orphaned pending IDs may exist without corresponding results
> > 
> > In these cases, `executeToolCalls` is never reached, so its error handling cannot apply. The `doCall` auto-recovery acts as a safety net for these edge cases.
> > ## On achieving the goal with only `executeToolCalls` modifications
> > For failures occurring _during_ `executeToolCalls`, yes — the `onErrorResume` logic is sufficient. However, for the scenarios above, we need the `doCall` layer to handle states that originate from outside the tool execution flow.
> > ## Proposed compromise to protect HITL workflows
> > To address your concern about masking improper HITL usage, I suggest adding an explicit mode check:
> > ```java
> > if (providedResults.isEmpty()) {
> >     if (isHITLMode()) {  // Check if conversation is in HITL pause state
> >         // Strict behavior: require manual ToolResultBlock
> >         throw new IllegalStateException(
> >             "HITL workflow requires manual ToolResultBlock for pending IDs: " + pendingIds);
> >     }
> >     // Non-HITL: Auto-recover from orphaned states
> >     log.warn("Orphaned pending tool calls detected, auto-generating error results: {}", pendingIds);
> >     generateAndAddErrorToolResults(pendingIds);
> > }
> > ```
> 
> Thanks for the thorough review! I've carefully reconsidered this, and I believe this is fundamentally a **bug fix** rather than an enhancement — adding configuration parameters would not be the right approach.
> 
> ## Why this is a critical bug (not just an enhancement)
> The current behavior causes **complete conversation failure** on any tool execution error:
> 
> 1. **First failure**: Tool timeout/exception → `IllegalStateException` → Agent crashes
> 2. **Cascading effect**: Pending state persists in memory
> 3. **All subsequent requests**: Fail with same exception, making the agent unusable
> 
> This violates basic fault isolation principles — a single tool failure should not crash the entire agent permanently.
> 
> ## Two key reasons why auto-supplement is necessary
> **First**, tool execution failures should not leave the agent in a permanently broken state. Whether it's a timeout, network error, or unexpected exception, the agent must recover and continue functioning. The "auto-supplement" in `doCall` serves as a critical safety net for orphaned pending states that originate **outside** `executeToolCalls` coverage (e.g., JVM crashes, memory recovery, pre-execution failures).
> 
> **Second**, manually generating error results and feeding them to the LLM is **essential for proper decision-making**. Without this feedback, the model has no visibility into what happened. By providing explicit error messages (e.g., _"[ERROR] Tool execution failed: timeout"_), we enable the model to:
> 
> * Understand the failure context
> * Decide on alternative approaches
> * Communicate meaningfully with the user
> 
> This is not masking errors — it's **proper error propagation** to the model layer.
> 
> ## Regarding HITL concerns
> The HITL workflow concern is valid, but I'd argue:
> 
> * HITL pause/resume should be handled at a **higher orchestration layer**, not by leaving the agent in a corrupted state
> * If HITL requires manual `ToolResultBlock`, it should explicitly manage the conversation lifecycle (clearing/resuming state) rather than relying on low-level exceptions
> 
> ## Summary
> Approach	Result
> Only `executeToolCalls` fix	Covers runtime failures, but leaves recovery gaps
> Configuration switches	Adds complexity without solving the root problem
> **Current PR (both fixes)**	Complete fault isolation + proper LLM feedback
> The combination ensures robust error handling while keeping the implementation clean and deterministic. I'm happy to discuss alternative HITL integration patterns if needed, but I believe the core fix should remain as-is.
> 
> Would appreciate your thoughts on this perspective.

Thanks for the feedback. I agree with your take on the fault isolation principle.

One nuance to consider: the current autocompletion doesn't differentiate between the root causes of a missing ToolResult. This is precisely why I previously mentioned the HITL (Human-in-the-loop) mechanism.

We need to decide where to propagate exceptions: to the model (via autocompletion) or to the system/developer. Model propagation works well for recoverable errors (timeouts, execution failures) where the model can retry or select alternative tool. But for non-recoverable issues (bugs, API misuse or other system-level failures), system-side propagation is likely more appropriate.

Refining this would require mapping out specific scenarios, so for this iteration, propagating all cases to the model seems like a reasonable baseline.

@chickenlj PTAL

**Comment by chensk0601** [comment]
> > > > > > The changes to executeToolCalls look solid, as they allow the model to proceed even after a tool call exception. However, I’m not sure about adding automatic recovery to doCall. In my view, it might be better to surface these exceptions to the framework consumers instead. Open to discussion on this point!
> > > > > 
> > > > > 
> > > > > This fix not only ensures execution continues after tool call timeouts and exceptions, but also resolves a critical issue where the entire agent would become unusable following a timeout, repeatedly throwing IllegalStateException. I have been running this updated version in our production environment for some time now, and it has been performing flawlessly.
> > > > 
> > > > 
> > > > It seems that the modifications to executeToolCalls should be sufficient to ensure a ToolResultBlock is present following a timeout or exception, which would prevent the persistent IllegalStateException.
> > > > However, I have some reservations about the auto-supplement logic in doCall. Since HITL (Human-In-The-Loop) workflows require a ToolResultBlock to be manually provided when resuming a conversation, automating this process might lead to unintended execution paths. It could also potentially mask improper HITL usage, making it harder to detect implementation errors.
> > > > It’s great to hear that this fix has been verified in production! A couple of follow-up questions:
> > > > 
> > > > * Have we encountered specific scenarios where the auto-supplement logic in doCall was actually triggered?
> > > > * In our current context, would it be possible to achieve the same goal by only modifying executeToolCalls?
> > > 
> > > 
> > > Thanks for the thorough review! I appreciate the concern about HITL workflows — preserving the explicit contract for manual `ToolResultBlock` provision is indeed important.
> > > Let me address your questions:
> > > ## Regarding scenarios where `doCall` auto-supplement was triggered
> > > While `executeToolCalls`'s `onErrorResume` handles failures _within_ tool execution, we've identified cases where pending states persist **outside** that coverage:
> > > 
> > > 1. **Pre-execution failures**: Network timeouts before `toolkit.callTools()` begins, JVM crashes, or async scheduling failures where tool calls are queued but never executed
> > > 2. **Memory recovery**: When agent state is persisted and restored (e.g., after restart), orphaned pending IDs may exist without corresponding results
> > > 
> > > In these cases, `executeToolCalls` is never reached, so its error handling cannot apply. The `doCall` auto-recovery acts as a safety net for these edge cases.
> > > ## On achieving the goal with only `executeToolCalls` modifications
> > > For failures occurring _during_ `executeToolCalls`, yes — the `onErrorResume` logic is sufficient. However, for the scenarios above, we need the `doCall` layer to handle states that originate from outside the tool execution flow.
> > > ## Proposed compromise to protect HITL workflows
> > > To address your concern about masking improper HITL usage, I suggest adding an explicit mode check:
> > > ```java
> > > if (providedResults.isEmpty()) {
> > >     if (isHITLMode()) {  // Check if conversation is in HITL pause state
> > >         // Strict behavior: require manual ToolResultBlock
> > >         throw new IllegalStateException(
> > >             "HITL workflow requires manual ToolResultBlock for pending IDs: " + pendingIds);
> > >     }
> > >     // Non-HITL: Auto-recover from orphaned states
> > >     log.warn("Orphaned pending tool calls detected, auto-generating error results: {}", pendingIds);
> > >     generateAndAddErrorToolResults(pendingIds);
> > > }
> > > ```
> > 
> > 
> > Thanks for the thorough review! I've carefully reconsidered this, and I believe this is fundamentally a **bug fix** rather than an enhancement — adding configuration parameters would not be the right approach.
> > ## Why this is a critical bug (not just an enhancement)
> > The current behavior causes **complete conversation failure** on any tool execution error:
> > 
> > 1. **First failure**: Tool timeout/exception → `IllegalStateException` → Agent crashes
> > 2. **Cascading effect**: Pending state persists in memory
> > 3. **All subsequent requests**: Fail with same exception, making the agent unusable
> > 
> > This violates basic fault isolation principles — a single tool failure should not crash the entire agent permanently.
> > ## Two key reasons why auto-supplement is necessary
> > **First**, tool execution failures should not leave the agent in a permanently broken state. Whether it's a timeout, network error, or unexpected exception, the agent must recover and continue functioning. The "auto-supplement" in `doCall` serves as a critical safety net for orphaned pending states that originate **outside** `executeToolCalls` coverage (e.g., JVM crashes, memory recovery, pre-execution failures).
> > **Second**, manually generating error results and feeding them to the LLM is **essential for proper decision-making**. Without this feedback, the model has no visibility into what happened. By providing explicit error messages (e.g., _"[ERROR] Tool execution failed: timeout"_), we enable the model to:
> > 
> > * Understand the failure context
> > * Decide on alternative approaches
> > * Communicate meaningfully with the user
> > 
> > This is not masking errors — it's **proper error propagation** to the model layer.
> > ## Regarding HITL concerns
> > The HITL workflow concern is valid, but I'd argue:
> > 
> > * HITL pause/resume should be handled at a **higher orchestration layer**, not by leaving the agent in a corrupted state
> > * If HITL requires manual `ToolResultBlock`, it should explicitly manage the conversation lifecycle (clearing/resuming state) rather than relying on low-level exceptions
> > 
> > ## Summary
> > Approach	Result
> > Only `executeToolCalls` fix	Covers runtime failures, but leaves recovery gaps
> > Configuration switches	Adds complexity without solving the root problem
> > **Current PR (both fixes)**	Complete fault isolation + proper LLM feedback
> > The combination ensures robust error handling while keeping the implementation clean and deterministic. I'm happy to discuss alternative HITL integration patterns if needed, but I believe the core fix should remain as-is.
> > Would appreciate your thoughts on this perspective.
> 
> Thanks for the feedback. I agree with your take on the fault isolation principle.
> 
> One nuance to consider: the current autocompletion doesn't differentiate between the root causes of a missing ToolResult. This is precisely why I previously mentioned the HITL (Human-in-the-loop) mechanism.
> 
> We need to decide where to propagate exceptions: to the model (via autocompletion) or to the system/developer. Model propagation works well for recoverable errors (timeouts, execution failures) where the model can retry or select alternative tool. But for non-recoverable issues (bugs, API misuse or other system-level failures), system-side propagation is likely more appropriate.
> 
> Refining this would require mapping out specific scenarios, so for this iteration, propagating all cases to the model seems like a reasonable baseline.
> 
> @chickenlj PTAL

Thanks for the clear guidance on error categorization. I fully agree that differentiating recoverable vs non-recoverable errors is the right long-term direction.

In the latest commit, I've made several improvements that partially address this:

1. **Hook pipeline integration**: `generateAndAddErrorToolResults()` now routes error results through the `PostActingEvent` hook pipeline (same path as normal tool execution). This means developers can intercept, modify, or `stopAgent()` on error results via hooks — providing an extension point for custom error handling without a dedicated exception handler API.

2. **Error scope narrowing**: `onErrorResume` is now scoped to `Exception.class` only, so critical JVM errors (`OutOfMemoryError`, `StackOverflowError`, etc.) propagate to the system as expected.

3. **Better diagnostics**: Using `ExceptionUtils.getErrorMessage()` for non-null error messages and logging the full exception object (with stack trace) for production debugging.

4. **Stronger test coverage**: Added assertions verifying error `ToolResultBlock` is written to memory, model re-invocation happens, and response content matches expectations.

For the error categorization refinement (routing certain exception types to system-side propagation), I think that fits well as a follow-up iteration once we have concrete scenarios mapped out. The current hook-based extensibility should cover most custom handling needs in the meantime.

Could you take another look when you get a chance?

File: agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java [hunk 9]
```diff
@@ -606,7 +694,33 @@ private Mono<List<Map.Entry<ToolUseBlock, ToolResultBlock>>> executeToolCalls(
                         results ->
                                 IntStream.range(0, toolCalls.size())
                                         .mapToObj(i -> Map.entry(toolCalls.get(i), results.get(i)))
-                                        .toList());
+                                        .toList())
+                .onErrorResume(
+                        Exception.class,
```
LearningGp, line 675: Avoid swallowing InterruptedException. In the context of AgentScope, this exception must be propagated to ensure the interruption policy functions correctly.

LearningGp:
> > > > > > > The changes to executeToolCalls look solid, as they allow the model to proceed even after a tool call exception. However, I’m not sure about adding automatic recovery to doCall. In my view, it might be better to surface these exceptions to the framework consumers instead. Open to discussion on this point!对 executeToolCalls 的修改看起来很稳妥，因为它们允许模型在工具调用异常后继续执行。不过，我不确定是否应该在 doCall 中添加自动恢复机制。在我看来，将这些异常暴露给框架使用者可能更好。欢迎就此点展开讨论！
> > > > > > 
> > > > > > 
> > > > > > This fix not only ensures execution continues after tool call timeouts and exceptions, but also resolves a critical issue where the entire agent would become unusable following a timeout, repeatedly throwing IllegalStateException. I have been running this updated version in our production environment for some time now, and it has been performing flawlessly.此修复不仅确保了在工具调用超时和异常后执行能够继续进行，还解决了此前一个关键问题：即超时发生后整个代理会变得不可用，并持续抛出 IllegalStateException 异常。我已在我们的生产环境中运行此更新版本一段时间，表现完美无瑕。
> > > > > 
> > > > > 
> > > > > It seems that the modifications to executeToolCalls should be sufficient to ensure a ToolResultBlock is present following a timeout or exception, which would prevent the persistent IllegalStateException.似乎对 executeToolCalls 的修改已足够确保在超时或异常后仍会存在一个 ToolResultBlock，从而防止持续出现 IllegalStateException。
> > > > > However, I have some reservations about the auto-supplement logic in doCall. Since HITL (Human-In-The-Loop) workflows require a ToolResultBlock to be manually provided when resuming a conversation, automating this process might lead to unintended execution paths. It could also potentially mask improper HITL usage, making it harder to detect implementation errors.然而，我对 doCall 中的自动补充逻辑有些顾虑。由于 HITL（Human-In-The-Loop）工作流在恢复对话时需要手动提供 ToolResultBlock，自动化此过程可能导致意外的执行路径。这也可能掩盖不当的 HITL 使用方式，使得实现错误更难被发现。
> > > > > It’s great to hear that this fix has been verified in production! A couple of follow-up questions:很高兴听到这个修复已在生产环境中验证通过！有几个后续问题：
> > > > > 
> > > > > * Have we encountered specific scenarios where the auto-supplement logic in doCall was actually triggered?我们是否遇到过 doCall 中的自动补充逻辑实际上被触发的特定场景？
> > > > > * In our current context, would it be possible to achieve the same goal by only modifying executeToolCalls?在我们当前的上下文中，是否可以通过仅修改 executeToolCalls 来实现相同的目标？
> > > > 
> > > > 
> > > > Thanks for the thorough review! I appreciate the concern about HITL workflows — preserving the explicit contract for manual `ToolResultBlock` provision is indeed important.感谢详细的审查！我非常感谢你对 HITL 工作流的关切——保持对人工提供 `ToolResultBlock` 的显式契约确实很重要。
> > > > Let me address your questions:让我来回答你的问题：
> > > > ## Regarding scenarios where `doCall` auto-supplement was triggered关于触发 `doCall` 自动补全的场景
> > > > While `executeToolCalls`'s `onErrorResume` handles failures _within_ tool execution, we've identified cases where pending states persist **outside** that coverage:虽然 `executeToolCalls` 的 `onErrorResume` 处理了工具执行过程中的失败情况，但我们发现仍存在一些情况，即挂起状态会超出该覆盖范围而持续存在：
> > > > 
> > > > 1. **Pre-execution failures**: Network timeouts before `toolkit.callTools()` begins, JVM crashes, or async scheduling failures where tool calls are queued but never executed预执行失败：在 `toolkit.callTools()` 开始之前发生网络超时、JVM 崩溃，或异步调度失败（工具调用被放入队列但从未被执行）
> > > > 2. **Memory recovery**: When agent state is persisted and restored (e.g., after restart), orphaned pending IDs may exist without corresponding results内存恢复：当代理状态被持久化并恢复时（例如在重启后），可能存在没有对应结果的孤立待处理 ID
> > > > 
> > > > In these cases, `executeToolCalls` is never reached, so its error handling cannot apply. The `doCall` auto-recovery acts as a safety net for these edge cases.在这些情况下， `executeToolCalls` 永远不会被到达，因此其错误处理无法生效。 `doCall` 的自动恢复机制为这些边缘情况提供了一个安全网。
> > > > ## On achieving the goal with only `executeToolCalls` modifications仅通过 `executeToolCalls` 次修改实现目标
> > > > For failures occurring _during_ `executeToolCalls`, yes — the `onErrorResume` logic is sufficient. However, for the scenarios above, we need the `doCall` layer to handle states that originate from outside the tool execution flow.在 `executeToolCalls` 期间发生的故障，是的—— `onErrorResume` 的逻辑已经足够。然而，对于上述情况，我们需要 `doCall` 层来处理源自工具执行流程之外的状态。
> > > > ## Proposed compromise to protect HITL workflows提议的折衷方案，以保护人机协同工作流程
> > > > To address your concern about masking improper HITL usage, I suggest adding an explicit mode check:为解决您对不当使用 HITL 的担忧，我建议添加一个显式的模式检查：
> > > > ```java
> > > > if (providedResults.isEmpty()) {
> > > >     if (isHITLMode()) {  // Check if conversation is in HITL pause state
> > > >         // Strict behavior: require manual ToolResultBlock
> > > >         throw new IllegalStateException(
> > > >             "HITL workflow requires manual ToolResultBlock for pending IDs: " + pendingIds);
> > > >     }
> > > >     // Non-HITL: Auto-recover from orphaned states
> > > >     log.warn("Orphaned pending tool calls detected, auto-generating error results: {}", pendingIds);
> > > >     generateAndAddErrorToolResults(pendingIds);
> > > > }
> > > > ```
> > > 
> > > 
> > > Thanks for the thorough review! I've carefully reconsidered this, and I believe this is fundamentally a **bug fix** rather than an enhancement — adding configuration parameters would not be the right approach.感谢详细的审查！我已仔细重新考虑了这个问题，我认为这本质上是一个错误修复，而不是功能增强——添加配置参数并不是正确的做法。
> > > ## Why this is a critical bug (not just an enhancement)为什么这是一个严重错误（而不仅仅是一个改进）
> > > The current behavior causes **complete conversation failure** on any tool execution error:当前行为在任何工具执行错误时都会导致整个对话失败：
> > > 
> > > 1. **First failure**: Tool timeout/exception → `IllegalStateException` → Agent crashes首次失败：工具超时/异常 → `IllegalStateException` → 代理崩溃
> > > 2. **Cascading effect**: Pending state persists in memory级联效应：待处理状态在内存中持续存在
> > > 3. **All subsequent requests**: Fail with same exception, making the agent unusable所有后续请求：均因相同异常而失败，导致代理无法使用
> > > 
> > > This violates basic fault isolation principles — a single tool failure should not crash the entire agent permanently.这违反了基本的故障隔离原则——单个工具的失败不应当永久导致整个代理崩溃。
> > > ## Two key reasons why auto-supplement is necessary自动补充必要的两个主要原因
> > > **First**, tool execution failures should not leave the agent in a permanently broken state. Whether it's a timeout, network error, or unexpected exception, the agent must recover and continue functioning. The "auto-supplement" in `doCall` serves as a critical safety net for orphaned pending states that originate **outside** `executeToolCalls` coverage (e.g., JVM crashes, memory recovery, pre-execution failures).首先，工具执行失败不应使代理处于永久损坏状态。无论是超时、网络错误还是意外异常，代理都必须能够恢复并继续运行。 `doCall` 中的“自动补充”机制作为关键的安全网，用于处理超出 `executeToolCalls` 覆盖范围的孤立挂起状态（例如 JVM 崩溃、内存恢复、预执行失败等情况）。
> > > **Second**, manually generating error results and feeding them to the LLM is **essential for proper decision-making**. Without this feedback, the model has no visibility into what happened. By providing explicit error messages (e.g., _"[ERROR] Tool execution failed: timeout"_), we enable the model to:其次，手动生成错误结果并将其提供给 LLM 对于正确决策至关重要。如果没有这种反馈，模型就无法了解发生了什么。通过提供明确的错误消息（例如，“[ERROR] 工具执行失败：超时”），我们可以使模型能够：
> > > 
> > > * Understand the failure context理解失败的上下文
> > > * Decide on alternative approaches决定替代方案
> > > * Communicate meaningfully with the user与用户进行有意义的交流
> > > 
> > > This is not masking errors — it's **proper error propagation** to the model layer.这不是屏蔽错误，而是将错误正确地向上传播到模型层。
> > > ## Regarding HITL concerns  关于人工介入学习（HITL）的顾虑
> > > The HITL workflow concern is valid, but I'd argue:HITL 工作流程的问题是合理的，但我认为：
> > > 
> > > * HITL pause/resume should be handled at a **higher orchestration layer**, not by leaving the agent in a corrupted stateHITL 暂停/恢复应由更高层次的编排层处理，而不是让代理处于损坏状态
> > > * If HITL requires manual `ToolResultBlock`, it should explicitly manage the conversation lifecycle (clearing/resuming state) rather than relying on low-level exceptions如果 HITL 需要手动 `ToolResultBlock` ，则应显式管理对话生命周期（清除/恢复状态），而不是依赖底层异常
> > > 
> > > ## Summary  摘要
> > > Approach	Result  方法结果
> > > Only `executeToolCalls` fix	Covers runtime failures, but leaves recovery gaps仅修复了运行时失败，但留下了恢复的缺口
> > > Configuration switches	Adds complexity without solving the root problem配置开关 增加了复杂性，但并未解决根本问题
> > > **Current PR (both fixes)**	Complete fault isolation + proper LLM feedback当前 PR（两个修复）完成故障隔离 + 正确的 LLM 反馈
> > > The combination ensures robust error handling while keeping the implementation clean and deterministic. I'm happy to discuss alternative HITL integration patterns if needed, but I believe the core fix should remain as-is.这种组合确保了健壮的错误处理，同时保持实现的简洁性和确定性。如果需要，我很乐意讨论其他的人机协同（HITL）集成模式，但我认为核心修复应保持原样。
> > > Would appreciate your thoughts on this perspective.非常感谢您对此观点的反馈。
> > 
> > 
> > Thanks for the feedback. I agree with your take on the fault isolation principle.感谢反馈。我同意你对故障隔离原则的看法。
> > One nuance to consider: the current autocompletion doesn't differentiate between the root causes of a missing ToolResult. This is precisely why I previously mentioned the HITL (Human-in-the-loop) mechanism.需要注意的一个细节是：当前的自动补全功能无法区分 ToolResult 缺失的根本原因。这正是我之前提到 HITL（人机协同）机制的原因。
> > We need to decide where to propagate exceptions: to the model (via autocompletion) or to the system/developer. Model propagation works well for recoverable errors (timeouts, execution failures) where the model can retry or select alternative tool. But for non-recoverable issues (bugs, API misuse or other system-level failures), system-side propagation is likely more appropriate.我们需要决定异常应该传播到哪里：通过自动补全传播给模型，还是传播给系统/开发者。对于可恢复的错误（如超时、执行失败），将异常传播给模型效果较好，因为模型可以重试或选择其他工具。但对于不可恢复的问题（如 bug、API 使用不当或其他系统级故障），向系统侧传播异常可能更合适。
> > Refining this would require mapping out specific scenarios, so for this iteration, propagating all cases to the model seems like a reasonable baseline.进一步完善需要明确具体场景，因此在这一版本中，将所有情况传递给模型似乎是一个合理的基线方案。
> > @chickenlj PTAL  请查看
> 
> Thanks for the clear guidance on error categorization. I fully agree that differentiating recoverable vs non-recoverable errors is the right long-term direction.感谢您对错误分类的清晰指导。我完全同意，区分可恢复与不可恢复错误是正确的长期发展方向。
> 
> In the latest commit, I've made several improvements that partially address this:在最新的提交中，我进行了几项改进，部分解决了这个问题：
> 
> 1. **Hook pipeline integration**: `generateAndAddErrorToolResults()` now routes error results through the `PostActingEvent` hook pipeline (same path as normal tool execution). This means developers can intercept, modify, or `stopAgent()` on error results via hooks — providing an extension point for custom error handling without a dedicated exception handler API.钩子管道集成： `generateAndAddErrorToolResults()` 现在通过 `PostActingEvent` 钩子管道路由错误结果（与正常工具执行相同的路径）。这意味着开发者可以通过钩子拦截、修改或 `stopAgent()` 错误结果，从而在无需专用异常处理 API 的情况下提供自定义错误处理的扩展点。
> 2. **Error scope narrowing**: `onErrorResume` is now scoped to `Exception.class` only, so critical JVM errors (`OutOfMemoryError`, `StackOverflowError`, etc.) propagate to the system as expected.错误作用域缩小： `onErrorResume` 现在仅限于 `Exception.class` 的作用域，因此关键的 JVM 错误（ `OutOfMemoryError` 、 `StackOverflowError` 等）会如预期那样传播到系统中。
> 3. **Better diagnostics**: Using `ExceptionUtils.getErrorMessage()` for non-null error messages and logging the full exception object (with stack trace) for production debugging.更好的诊断：对于非空错误消息使用 `ExceptionUtils.getErrorMessage()` ，并在生产调试时记录完整的异常对象（包含堆栈跟踪）。
> 4. **Stronger test coverage**: Added assertions verifying error `ToolResultBlock` is written to memory, model re-invocation happens, and response content matches expectations.更强的测试覆盖率：添加了断言以验证错误 `ToolResultBlock` 是否已写入内存、模型是否重新调用，以及响应内容是否符合预期。
> 
> For the error categorization refinement (routing certain exception types to system-side propagation), I think that fits well as a follow-up iteration once we have concrete scenarios mapped out. The current hook-based extensibility should cover most custom handling needs in the meantime.对于错误分类的优化（将某些异常类型路由到系统侧传播），我认为这可以在我们明确具体场景后再作为后续迭代进行。目前基于钩子的可扩展性应该足以满足大部分自定义处理需求。
> 
> Could you take another look when you get a chance?你有空时能再看一下吗？

Works for me.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -255,9 +256,92 @@ protected Mono<Msg> doCall(List<Msg> msgs) {
...
+                                            "Previous tool execution failed or was interrupted."
+                                                    + " Tool: "
+                                                    + toolCall.getName());
+                            log.info(
+                                    "Auto-generated error result for pending tool call: {} ({})",
+                                    toolCall.getName(),
+                                    toolCall.getId());
+                            return notifyPostActingHook(Map.entry(toolCall, errorResult));
+                        })
+                .then();
```
chickenlj: I think ToolResult patching can be implemented in a separate Hook. That way, users can decide whether to enable patching or not when building ReActAgent. What do you think?

chensk0601: > I think ToolResult patching can be implemented in a separate Hook. That way, users can decide whether to enable patching or not when building ReActAgent. What do you think?

Good idea. Extracting it as a Hook would align well with the existing extensibility model and address the "expose to developer" concern more cleanly.
One consideration: since the original issue (#951) is a crash-level bug, I'd suggest making this Hook enabled by default in ReActAgent (similar to how StaticLongTermMemoryHook is auto-registered). Users who want custom handling (retry, HITL escalation, etc.) can replace or remove it.
Want me to refactor it in this PR, or would you prefer a follow-up PR for the Hook extraction?

chickenlj: > since the original issue (#951) is a crash-level bug, I'd suggest making this Hook enabled by default in ReActAgent (similar to how StaticLongTermMemoryHook is auto-registered).

I tend to make this hook optional by default, allowing users to enable it based on their needs. Meanwhile, we need a dedicated section on our official website to clearly explain the built-in hooks and their respective functionalities, enabling users who require them to activate them independently.

> Want me to refactor it in this PR, or would you prefer a follow-up PR for the Hook extraction?

Both sound good to me, please go ahead.

chensk0601: > > since the original issue (#951) is a crash-level bug, I'd suggest making this Hook enabled by default in ReActAgent (similar to how StaticLongTermMemoryHook is auto-registered).
> 
> I tend to make this hook optional by default, allowing users to enable it based on their needs. Meanwhile, we need a dedicated section on our official website to clearly explain the built-in hooks and their respective functionalities, enabling users who require them to activate them independently.
> 
> > Want me to refactor it in this PR, or would you prefer a follow-up PR for the Hook extraction?
> 
> Both sound good to me, please go ahead.

Thanks for the suggestion! I've extracted the pending tool recovery logic into a standalone PendingToolRecoveryHook that intercepts PreCallEvent.
Changes:
Created PendingToolRecoveryHook — detects orphaned pending tool calls and auto-generates error ToolResultBlocks before agent processing begins
Added enablePendingToolRecovery(boolean) to ReActAgent.Builder (default: false) so users can opt in explicitly
Simplified ReActAgent.doCall() — removed inline auto-recovery; now throws IllegalStateException if pending state is unresolved and the hook is not enabled
Removed generateAndAddErrorToolResults() and findLastAssistantMsg() from ReActAgent (logic moved to hook)
The hook skips patching when input messages are empty (resume scenario) or when user already provides ToolResultBlocks

PR #981, opened 2026-03-16
fix(core): Fix MediaUtils parse path extension

guanxuc:
## AgentScope-Java Version

1.0.11

## Description

* Closes: #977 
* Fix MediaUtils parse path extension

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/MediaUtils.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/MediaUtils.java
@@ -354,22 +354,19 @@ public class MediaUtils {
             return "";
         }
 
-        boolean localFile = isLocalFile(path);
-
         Path fileNamePath;
-        if (localFile) {
-            // treat as file
-            fileNamePath = Paths.get(path).normalize().getFileName();
-        } else {
-            // treat as url
-            URI uri;
-            try {
-                uri = URI.create(path).normalize();
-            } catch (IllegalArgumentException e) {
-                log.error("Invalid URL: {}", path);
-                return "";
+        try {
+            if (isLocalFile(path)) {
+                // treat as file
+                fileNamePath = Paths.get(path).normalize().getFileName();
+            } else {
+                // treat as url
+                URI uri = URI.create(path).normalize();
+                fileNamePath = Paths.get(uri.getPath()).getFileName();
             }
-            fileNamePath = Paths.get(uri.getPath()).getFileName();
+        } catch (Exception e) {
+            log.warn("Invalid path: {}", path, e);
+            return "";
         }
 
         if (fileNamePath == null) {


agentscope-core/src/main/java/io/agentscope/core/formatter/MediaUtils.java
```diff
@@ -354,8 +355,22 @@ public static String getExtension(String path) {
         }
         int dotIndex = path.lastIndexOf('.');
         int slashIndex = Math.max(path.lastIndexOf('/'), path.lastIndexOf('\\'));
+
```
LearningGp, line 356: The logic fails when query params are present (e.g., ?v=2.0). We should probably strip or parse the query string first.
https://cdn.example.com/avatar.jpg?v=2.0

guanxuc:
@LearningGp Thanks for your review and good suggestion! I have fixed the code by strip the path.

agentscope-core/src/main/java/io/agentscope/core/formatter/MediaUtils.java
```diff
@@ -352,8 +353,22 @@ public static String getExtension(String path) {
         if (path == null) {
             return "";
         }
+
+        // Check for query string or fragment
+        Set<String> masks = Set.of("?", "&", "#");
```
LearningGp: This should be a static final constant.

guanxuc:
@LearningGp I have changed the parse code logic. Use `java.net.URI` to parse url path and use `java.nio.file.Path` to parse file path, I think that will be better, PTAL.

agentscope-core/src/main/java/io/agentscope/core/formatter/MediaUtils.java
```diff
@@ -349,14 +350,37 @@ public static String inferAudioFormatFromMediaType(String mediaType) {
...
+        }
+
+        if (fileNamePath == null) {
+            return "";
+        }
+
+        String fileName = fileNamePath.toString();
+        int dotIndex = fileName.lastIndexOf('.');
+        // Ensure the dot exists and is not the last character
+        if (dotIndex > -1 && dotIndex < path.length() - 1) {
```
LearningGp: Would it be more readable to change this to dotIndex > -1 && dotIndex < fileName.length() - 1?

agentscope-core/src/test/java/io/agentscope/core/formatter/MediaUtilsTest.java
LearningGp, line 1: Suggest consolidating the test cases to keep the suite concise.

guanxuc:
@LearningGp Thanks for your review, I've cleaned up the code completely.

agentscope-core/src/main/java/io/agentscope/core/formatter/MediaUtils.java
```diff
@@ -349,14 +350,37 @@ public static String inferAudioFormatFromMediaType(String mediaType) {
...
+        }
+
+        if (fileNamePath == null) {
+            return "";
+        }
+
+        String fileName = fileNamePath.toString();
+        int dotIndex = fileName.lastIndexOf('.');
+        // Ensure the dot exists and is not the last character
+        if (dotIndex != -1 && dotIndex < path.length() - 1) {
```
LearningGp: Is there a specific reason for using path.length() here rather than fileName.length()?

guanxuc: Oh, I will fix it, thanks for your point.

PR #1003, opened 2026-03-20
fix(dashscope): avoid 400 for non-streaming qwen3.5 multimodal requests

### LINKED ISSUE #984: [Bug]:Error when calling qwen3.5-plus
2026-03-17 17:28:00.920 [] [] [] [boundedElastic-1] DEBUG i.a.core.model.DashScopeHttpClient - DashScope request to https://dashscope.aliyuncs.com/api/v1/services/aigc/multimodal-generation/generation: {"model":"qwen3.5-plus",

io.agentscope.core.model.DashScopeHttpClient$DashScopeHttpException: DashScope API request failed with status 400
at io.agentscope.core.model.DashScopeHttpClient.call(DashScopeHttpClient.java:198)
at io.agentscope.core.model.DashScopeChatModel.lambda$streamWithHttpClient$1(DashScopeChatModel.java:288)

rrrjqy66:
Fixes #984
## AgentScope-Java Version

1.0.11

### Background

When `DashScopeChatModel` is used with `qwen3.5-plus` in non-streaming mode, DashScope may return HTTP 400.

At first glance this can look like an endpoint routing issue, but the multimodal endpoint is actually correct here. According to DashScope's Qwen 3.5 model behavior, `qwen3.5-plus` should still be treated as a multimodal model. The real problem is that Qwen 3.5 enables thinking by default, while thinking mode requires streaming. In the previous implementation, when `stream=false` and `enableThinking` was not explicitly set, the request did not explicitly disable thinking, which could cause DashScope to reject the request.

## Checklist

### DIFF
--- a/agentscope-extensions/agentscope-extensions-mem0/src/main/java/io/agentscope/core/memory/mem0/Mem0ApiType.java
+++ b/agentscope-extensions/agentscope-extensions-mem0/src/main/java/io/agentscope/core/memory/mem0/Mem0ApiType.java
@@ -18,10 +18,12 @@ package io.agentscope.core.memory.mem0;
 /**
  * Enumeration of Mem0 API deployment types.
  *
- * <p>Mem0 provides two deployment options with different API endpoints:
+ * <p>Mem0 provides two deployment options with different API endpoints and authentication:
  * <ul>
- *   <li><b>PLATFORM</b>: Official cloud service with endpoints /v1/memories/ and /v2/memories/search/</li>
- *   <li><b>SELF_HOSTED</b>: Self-deployed service with endpoints /memories and /search</li>
+ *   <li><b>PLATFORM</b>: Official cloud service with endpoints /v1/memories/ and /v2/memories/search/,
+ *       using {@code Authorization: Token <apiKey>} header</li>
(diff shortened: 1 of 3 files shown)


agentscope-core/src/test/java/io/agentscope/core/model/DashScopeChatModelTest.java
```diff
@@ -746,6 +768,69 @@ void testDoNonStreamErrorHandling() throws Exception {
...
+        Map<String, Object> parameters = (Map<String, Object>) requestBody.get("parameters");
+        Map<String, Object> input = (Map<String, Object>) requestBody.get("input");
+        List<Map<String, Object>> messages = (List<Map<String, Object>>) input.get("messages");
+        Map<String, Object> firstMessage = messages.get(0);
+        List<Map<String, Object>> content = (List<Map<String, Object>>) firstMessage.get("content");
+
+        assertEquals(Boolean.FALSE, parameters.get("enable_thinking"));
+        assertEquals("test", content.get(0).get("text"));
+
+        mockServer.shutdown();
```
LearningGp: Wrap shutdown() in a try-finally block to prevent resource leaks if an assertion fails.

LearningGp:
I’m unable to reproduce this issue. Could you please provide a minimal reproducible example?

Issue #992, opened 2026-03-19
[Feature]: MysqlSession clearAllSessions should use truncate instead of delete from


lfz757077613:
<img width="1228" height="666" alt="Image" src="https://github.com/user-attachments/assets/c46f673d-896d-465f-9ccd-747090c170b9" />
Its better to clean table with truncate

Issue #991, opened 2026-03-19
[Feature]: MysqlSession should use datetime(3) instead of timestamp to create table


lfz757077613:
<img width="1850" height="774" alt="Image" src="https://github.com/user-attachments/assets/fa324cf2-77f2-4587-92f0-d9a7848aae3b" />

keep millisecond: 
    created_at      datetime(3)     DEFAULT CURRENT_TIMESTAMP(3)
    updated_at      datetime(3)     DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3)

PR #1006, opened 2026-03-20
feat(session): Improve mysql session sql

guanxuc:
## AgentScope-Java Version

1.0.11

## Description

- Closes: #991 
- Closes: #992 
- Use DATETIME instead of TIMESTAMP for mysql field
- Use TRUNCATE instead of DELETE from table

## Checklist

### DIFF
--- a/agentscope-extensions/agentscope-extensions-session-mysql/src/main/java/io/agentscope/core/session/mysql/MysqlSession.java
+++ b/agentscope-extensions/agentscope-extensions-session-mysql/src/main/java/io/agentscope/core/session/mysql/MysqlSession.java
@@ -699,9 +699,11 @@ public class MysqlSession implements Session {
      * Clear all sessions from the database (for testing or cleanup).
      *
      * @return Number of rows deleted
+     * @deprecated Use {@link #truncateAllSessions()} instead
      */
+    @Deprecated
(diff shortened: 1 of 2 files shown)


agentscope-extensions/agentscope-extensions-session-mysql/src/main/java/io/agentscope/core/session/mysql/MysqlSession.java
```diff
@@ -701,7 +701,7 @@ public DataSource getDataSource() {
      * @return Number of rows deleted
      */
     public int clearAllSessions() {
-        String clearSql = "DELETE FROM " + getFullTableName();
+        String clearSql = "TRUNCATE TABLE " + getFullTableName();
```
LearningGp: In MySQL, TRUNCATE TABLE via JDBC's executeUpdate() typically returns 0, which contradicts the Javadoc ("Number of rows deleted"). It also generally requires DROP privileges. Consider clarifying this in the docs.

guanxuc:
@LearningGp Thanks for your good advice, I have updated the code and the java docs. Keep both `clearAllSessions` and `truncateAllSessions` and deprecated `clearAllSessions`  method.

PR #1008, opened 2026-03-21
fix(ollama): ensure stable message ID for streaming responses

### LINKED ISSUE #897: [Bug]:Ollama streaming responses do not populate Msg.id (null in REASONING events)
**Describe the bug**
When using Ollama as the model backend, streamed `REASONING` events may contain `Msg.id == null`.

This causes integration issues for AG-UI adapters. AG-UI requires a non-null and stable `messageId` for one assistant message across the full streaming lifecycle:

- `TEXT_MESSAGE_START`
- multiple `TEXT_MESSAGE_CONTENT`
- `TEXT_MESSAGE_END`

When `Msg.id` is missing on the Ollama path, the adapter cannot rely on upstream ids and must generate/cache its own fallback `messageId`. Otherwise, downstream AG-UI clients can fail to correctly match message start/content/end events.

This issue does not reproduce in our Qwen setup.

**To Reproduce**
1. Use AgentScope-Java with an Ollama model.
2. Stream responses with:
   `StreamOptions.builder().eventTypes(EventType.ALL).incremental(true).build()`
3. In the streaming callback, inspect `event.getMessage().getId()` for `EventType.REASONING`.
4. Observe that `msg.getId()` is `null` on the Ollama path.
5. Convert the stream into AG-UI text events, where `messageId` is required for `TEXT_MESSAGE_START / CONTENT / END`.

**Expected behavior**
`Msg.id` should be non-null and stable for all chunks that belong to the same assistant message.

This is important for protocols such as AG-UI, where all chunks of one assistant response must share the same `messageId` so the frontend can correctly close the text stream before `RUN_FINISHED`.

**Actual behavior**
On the Ollama path, `Msg.id` can be `null`. As a result, AG-UI adapters must implement provider-specific fallback logic in the adapter layer to synthesize and hold a stable `messageId`.

Without that handling, downstream clients may see broken message sequencing, such as text streams that cannot be cleanly matched from start to end.

**Error messages**
AgentScope itself does not throw an exception here, but the downstream AG-UI client can fail because text messages remain logically unmatched.

In our case, the frontend reports errors similar to:
`Cannot send 'RUN_FINISHED' while text messages are still active`

- AgentScope-Java Version: 1.0.9
- Java Version: 17
- OS: Windows

**Additional context**
From local inspection, the Ollama response path appears not to set `ChatResponse.id`, and that `null` value is later propagated into `Msg.id`.

It would help if AgentScope could either:
1. generate a stable message id when the provider does not return one, or
2. preserve an internally generated id instead of allowing `Msg.id` to become `null`.

That would make integrations with AG-UI and similar streaming protocols much safer.

jujn:
## Description

Close #897 
Fixed the issue of Olama streaming output generating different UUIDs for each data block due to the underlying layer not returning an ID. By caching and reusing the ID of the first data block during the lifecycle of the stream, it ensures that downstream AG-UI clients can receive stable messageIDs.

## Checklist


agentscope-core/src/main/java/io/agentscope/core/model/OllamaChatModel.java
```diff
@@ -187,8 +187,33 @@ private Flux<ChatResponse> streamWithHttpClient(
...
+                                                            formatter.parseResponse(
+                                                                    response, Instant.now());
+
+                                                    if (finalStableId[0] == null) {
+                                                        finalStableId[0] = parsedResponse.getId();
+                                                        return parsedResponse;
+                                                    }
+
+                                                    return ChatResponse.builder()
+                                                            .id(finalStableId[0])
```
LearningGp: Manually copying fields will lead to silent data loss when new fields are added to ChatResponse, and this cannot be caught at compile-time. Consider adding a withId(String) convenience method to ChatResponse to encapsulate the copying logic within the class itself.

jujn: > Manually copying fields will lead to silent data loss when new fields are added to ChatResponse, and this cannot be caught at compile-time. Consider adding a withId(String) convenience method to ChatResponse to encapsulate the copying logic within the class itself.

Done. Thanks for the suggestion.

PR #1011, opened 2026-03-21
fix(generate_response): Fix flaky concurrent tests for generate_response

xuhuafeifei:
## AgentScope-Java Version

[The version of AgentScope-Java you are working on, e.g. 1.0.9, check your pom.xml dependency version or run `mvn dependency:tree | grep agentscope-parent:pom`(only mac/linux)]

### **Background**

The unit test code `testConcurrencyConflictStructuredOutput ` contains two `agent.call` actions.

The first thread's `agent.call` writes `generate_response` into the *Toolkit* and then clears it in the `doFinally` block.

The second thread's `agent.call` also writes `generate_response` into the *Toolkit* and then clears it in the `doFinally` block.

### **Problem**

The issue occurs when the first thread executes `doFinally` later than the second thread writes `generate_response`. In this case, the first thread will clear the content written by the second thread into the *Toolkit*, causing the second thread to fail to retrieve `generate_response`.

### **Objective**

Avoid "late cleanup" mistakenly deleting a tool with the same name that has already been overwritten by another call.

### **Implementation Approach**

Use Java's `ConcurrentHashMap.remove(Object key, Object value)` to ensure only the tool written by the current thread is removed, while preventing race conditions.

### **How to Verify**

Add a new test method: `io.agentscope.core.agent.ReActAgentStructuredOutputTest#testConcurrencyConflictStructuredOutput_repeated`.

Test the same method 25,000 times. If no exceptions occur, it can be considered basically problem-free.

## Checklist


agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentStructuredOutputTest.java
```diff
@@ -536,6 +538,26 @@ void testStructuredOutputPreservesThinkingBlock() {
...
     @Test
     void testConcurrencyConflictStructuredOutput() {
+        runSubscribeOnThenSequentialSecondCallStructuredOutputScenario(toolkit);
+    }
+
+    /**
+     * Reproduces the subscribeOn(elastic) vs delayed {@code doFinally} race: run many times until
+     * failure or increase confidence after a fix.
+     */
+    @RepeatedTest(25000)
```
LearningGp, line 550: Consider skipping this by default or reducing the frequency to avoid bloating the CI reports.

xuhuafeifei: I will use @EnabledIfSystemProperty to disable this test by default.

PR #1034, opened 2026-03-24
fix(skill): use SnakeYAML to parse complex frontmatter in SKILL.md

jujn:
## Description

Close #1030

This PR replaces the regex-based `SimpleYamlParser` with `SnakeYAML` to correctly parse complex frontmatter (such as lists and nested objects) in `SKILL.md`. 

## Checklist

Please check the following items before code is ready to be reviewed.

- [ ]  Code has been formatted with `mvn spotless:apply`
- [ ]  All tests are passing (`mvn test`)
- [ ]  Javadoc comments are complete and follow project conventions
- [ ]  Related documentation has been updated (e.g. links, examples, etc.)
- [ ]  Code is ready for review

PR #1053, opened 2026-03-27
Support standard YAML skill frontmatter

richuV:
## Summary
- replace the ad-hoc flat frontmatter parser with SnakeYAML-based parsing
- keep exposing only top-level scalar metadata values so the existing skill API stays compatible
- add regression tests for nested YAML fields and multiline scalar frontmatter

## Why
`MarkdownSkillParser` currently rejects common YAML frontmatter constructs such as lists and block scalars. That makes otherwise valid `SKILL.md` files fail to load even when the framework only needs scalar fields like `name` and `description`.

This change makes the parser more tolerant of standard YAML frontmatter while preserving the current `Map<String, String>` metadata contract.

## Testing
- `mvn -pl agentscope-core "-Dtest=MarkdownSkillParserTest,SkillFileSystemHelperTest" test`
- `mvn -pl agentscope-core spotless:check`

### DIFF
--- a/agentscope-core/pom.xml
+++ b/agentscope-core/pom.xml
@@ -140,5 +140,11 @@
             <groupId>com.networknt</groupId>
             <artifactId>json-schema-validator</artifactId>
         </dependency>
+
+        <!-- YAML frontmatter parsing for markdown-based skills -->
+        <dependency>
+            <groupId>org.yaml</groupId>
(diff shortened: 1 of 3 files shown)


jujn:
hi, you can take a look at this PR: https://github.com/agentscope-ai/agentscope-java/pull/1034

LearningGp:
image

PR #1079, opened 2026-03-30
fix: avoid blocking Reactor scheduler thread during AutoContext compression

zhangliyuangit:
When compression is triggered in handlePreReasoning, compressIfNeeded() internally calls .block() for LLM-based summarization. Running this on the Reactor scheduler thread stalls downstream streaming chunk dispatch, causing all chunks to be flushed at once instead of incrementally.

Fix: wrap compressIfNeeded() in Mono.fromCallable(...).subscribeOn(Schedulers.boundedElastic()) so blocking work runs on the elastic thread pool.


LearningGp:
PTAL, @shiyiyue1102

PR #1088, opened 2026-03-31
fix: remove race condition in shutdown state transition

uuuyuqi:
## Summary

- Fix a race condition in `GracefulShutdownManager` where the monitor thread (with `initialDelay=0`) calls `updateTerminatedIfNoRequests()` on every tick, which can advance the state from `SHUTTING_DOWN` to `TERMINATED` before `performGracefulShutdown()` even returns. This causes `performShutdownTransition` test to fail flakily.
- Move `updateTerminatedIfNoRequests()` from the monitor thread to two deterministic call sites: `unregisterRequest()` (when the last request completes) and `awaitTermination()` entry (for the "no active requests at shutdown time" edge case).

## Root Cause

The monitor thread was responsible for both timeout enforcement (its intended job) and state transition (an overreach). Since it runs with `initialDelay=0`, it could execute `updateTerminatedIfNoRequests()` before `performGracefulShutdown()` returned to the caller, making the `SHUTTING_DOWN` state unobservable.

## Test plan

- [x] `GracefulShutdownTest` (59 tests) all pass, including the previously flaky `performShutdownTransition`
- [x] No other files changed, minimal blast radius

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownManager.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownManager.java
@@ -167,6 +167,7 @@ public final class GracefulShutdownManager {
         }
         activeRequestsByAgentId.remove(agent.getAgentId());
         sessionBindings.remove(agent.getAgentId());
+        updateTerminatedIfNoRequests();
     }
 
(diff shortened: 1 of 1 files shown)


LearningGp (approved):
LGTM

Issue #1095, opened 2026-04-01
[Bug]:OpenAITextEmbedding 组件存在设计缺陷


jackxmcheng:
OpenAITextEmbedding 组件存在设计缺陷：强制要求传入 dimensions 向量维度参数，并默认携带该参数调用接口，未区分模型兼容性。
该参数仅 OpenAI 官方嵌入模型支持，BAAI/bge 等开源嵌入模型不支持 matryoshka representation 嵌套向量功能，携带 dimensions 调用会直接返回 400 错误，导致开源模型无法使用。
复现方式
使用 OpenAITextEmbedding 调用开源嵌入模型（如 BAAI/bge-large-zh-v1.5），框架自动携带 dimensions 参数，接口报错：
Model "BAAI/bge-large-zh-v1.5" does not support matryoshka representation...
影响范围
所有非 OpenAI 官方嵌入模型均无法通过 OpenAITextEmbedding 正常调用，严重限制框架对开源嵌入模型的支持。
修复建议
将 dimensions 改为非必选参数，默认不传递；
仅当用户显式配置时，才携带该参数；
增加模型兼容性判断，仅对 OpenAI 官方模型支持 dimensions 配置。

guslegend0510:
Hi, I’d like to work on this issue as my first contribution.

My understanding is:
- make `dimensions` optional in `OpenAITextEmbedding`
- only send it when explicitly configured
- add/update unit tests for this behavior

Please let me know if this direction looks good. Thanks!

Asaduddin18:
Hi, I wanted to work on this issue as my first contribution.

Thanks!

Blue-Berrys:
I'd like to pick this up and implement the fix: make `dimensions` optional in `OpenAITextEmbedding`, only send it when explicitly configured, and add/update unit tests. The previous assignee claim appears inactive — happy to coordinate if a PR is already in progress.

PR #1110, opened 2026-04-02
fix: make dimensions parameter optional in OpenAITextEmbedding (#1095)

egezturkk:
## AgentScope-Java Version
1.0.12-SNAPSHOT

## Description
Background & Purpose:
Currently, `OpenAITextEmbedding` enforces a mandatory `dimensions` parameter. This causes 400 Bad Request errors when using open-source embedding models (like BAAI/bge) through the OpenAI-compatible API, as these models do not support the Matryoshka representation parameter.

Changes:
- Changed `dimensions` type from `int` to `Integer` in `OpenAITextEmbedding` and its `Builder`.
- Removed the hardcoded default value of `1536` to make it optional.
- Updated the request builder logic to only include the `dimensions` field in the API payload if it is explicitly provided.
- Added null checks in `embed()`, `getDimensions()`, and validation logic to ensure stability.

How to test:
Initialize `OpenAITextEmbedding` without calling `.dimensions()`. Verify that the outgoing JSON request to the embedding endpoint does not contain the `dimensions` key.

## Checklist
- [x] All tests are passing (mvn test)

### DIFF
--- a/agentscope-extensions/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/embedding/openai/OpenAITextEmbedding.java
+++ b/agentscope-extensions/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/embedding/openai/OpenAITextEmbedding.java
@@ -49,7 +49,7 @@ public class OpenAITextEmbedding implements EmbeddingModel {
 
     private final String apiKey;
     private final String modelName;
-    private final int dimensions;
+    private final Integer dimensions;
     private final ExecutionConfig defaultExecutionConfig;
 
     private final String baseUrl;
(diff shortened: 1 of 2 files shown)


agentscope-extensions/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/embedding/openai/OpenAITextEmbedding.java
```diff
@@ -225,7 +230,7 @@ public String getModelName() {
 
     @Override
     public int getDimensions() {
-        return dimensions;
+        return dimensions != null ? dimensions : 0;
```
LearningGp, line 233: The Javadoc for EmbeddingModel.getDimensions() explicitly states that the default value is 1024 if not specified, but this implementation returns 0.

Downstream stores (InMemoryStore, QdrantStore, PgVectorStore) validate that dimensions > 0 and will throw an exception otherwise. If a user passes the result of OpenAITextEmbedding.getDimensions() to a Store builder, it will cause a runtime failure.

LearningGp (changes requested):

PR #1118, opened 2026-04-03
fix: handle Map type in getChatUsage() after deserialization

### LINKED ISSUE #1115: [Bug]: getChatUsage
# Bug: getChatUsage() returns null when metadata._chat_usage is a Map type
## Describe the bug
The `getChatUsage()` method in the message class returns `null` incorrectly when the `_chat_usage` value in `metadata` is a **Map type** (instead of a direct `ChatUsage` object). This causes the token usage data to be unreadable via the official getter method, even though the raw data exists in the metadata map.

### 1. Your code
```java
// Load agent from session (persisted session data)
agent.loadFrom(session, "user123");

// Iterate over memory messages and try to get chat usage
for (Msg message : agent.getMemory().getMessages()) {
    // Official getter returns null unexpectedly
    ChatUsage usage = message.getChatUsage();
    System.out.println("Official getter usage: " + usage); // Output: null

    // Manual check shows the data exists as a Map
    if (message.getMetadata() != null) {
        Object rawUsage = message.getMetadata().get("_chat_usage");
        if (rawUsage instanceof Map) {
            Map<String, Object> usageMap = (Map) rawUsage;
            System.out.println("Map usage data: inputTokens=" + usageMap.get("inputTokens") 
                + ", outputTokens=" + usageMap.get("outputTokens")); // Valid data printed
        }
    }
}
```

### 2. How to execute
1. Reload the agent from the session using `agent.loadFrom()`
2. Call `message.getChatUsage()` on the loaded messages

### 3. See error
`getChatUsage()` returns `null`, but the underlying `_chat_usage` field is a valid Map containing token usage data.

## Expected behavior
The `getChatUsage()` method should **convert the Map type to a ChatUsage object** (or return the non-null usage data) instead of returning null. The getter should reliably return the token usage regardless of whether the underlying type is `ChatUsage` or `Map`.

## Error messages
No exception thrown, but `getChatUsage()` returns `null` unexpectedly.

## Environment
- AgentScope-Java Version: 1.0.10
- Java Version: 17
- OS: Windows

## Additional context
- Root cause: The `_chat_usage` field is serialized/stored as a Map when loading from session, but the getter only checks for `instanceof ChatUsage`
- The raw Map contains all required fields: `inputTokens`, `outputTokens`, `totalTokens`
- Temporary workaround: Manually cast to Map and extract fields (shown in reproduction code)
- Affects functionality: Token usage statistics cannot be retrieved via the official API after session reload

---

### Suggested Fix (Optional)
Convert the Map to a ChatUsage object in the getter:
```java
@Transient
@JsonIgnore
public ChatUsage getChatUsage() {
    if (metadata == null) {
        return null;
    }
    Object usage = metadata.get(MessageMetadataKeys.CHAT_USAGE);
    if (usage instanceof ChatUsage) {
        return (ChatUsage) usage;
    } else if (usage instanceof Map) {
        // Convert Map to ChatUsage
        Map<String, Object> usageMap = (Map<String, Object>) usage;
        ChatUsage chatUsage = new ChatUsage();
        chatUsage.setInputTokens((Integer) usageMap.get("inputTokens"));
        chatUsage.setOutputTokens((Integer) usageMap.get("outputTokens"));
        return chatUsage;
    }
    return null;
}
```

mvanhorn:
## Summary

`getChatUsage()` returns null when `_chat_usage` metadata is a Map (after session persistence/reload) instead of a ChatUsage object.

## Why this matters

When messages are persisted via `agent.loadFrom()` and reloaded, JSON deserialization converts the `_chat_usage` metadata value to a `Map<String, Object>` instead of a `ChatUsage` instance. The existing `instanceof ChatUsage` check fails, returning null even though the token usage data exists in the metadata.

## Changes

`agentscope-core/src/main/java/io/agentscope/core/message/Msg.java`:

- Added Map handling in `getChatUsage()` (line 413): when the metadata value is a Map, extract `inputTokens`, `outputTokens`, and `time` fields, build a ChatUsage via the builder, and cache it back in metadata for subsequent calls
- Added `toInt()` and `toDouble()` helper methods for safe numeric conversion from deserialized values (handles Integer, Long, Double from JSON)

## Testing

- When `_chat_usage` is a ChatUsage object: unchanged behavior
- When `_chat_usage` is a Map (post-deserialization): converts to ChatUsage and returns it
- When `_chat_usage` is null or absent: returns null (unchanged)
- The converted ChatUsage is cached back in metadata so the conversion only happens once

Fixes #1115

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/message/Msg.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/message/Msg.java
@@ -421,11 +421,12 @@ public class Msg implements State {
         if (usage instanceof Map) {
             @SuppressWarnings("unchecked")
             Map<String, Object> map = (Map<String, Object>) usage;
-            ChatUsage chatUsage = ChatUsage.builder()
-                    .inputTokens(toInt(map.get("inputTokens")))
-                    .outputTokens(toInt(map.get("outputTokens")))
-                    .time(toDouble(map.get("time")))
-                    .build();
+            ChatUsage chatUsage =
+                    ChatUsage.builder()
+                            .inputTokens(toInt(map.get("inputTokens")))
+                            .outputTokens(toInt(map.get("outputTokens")))
+                            .time(toDouble(map.get("time")))
+                            .build();
             metadata.put(MessageMetadataKeys.CHAT_USAGE, chatUsage);
             return chatUsage;
         }
--- a/agentscope-core/src/main/java/io/agentscope/core/message/Msg.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/message/Msg.java
@@ -415,7 +415,35 @@ public class Msg implements State {
             return null;
         }
         Object usage = metadata.get(MessageMetadataKeys.CHAT_USAGE);
-        return usage instanceof ChatUsage ? (ChatUsage) usage : null;
+        if (usage instanceof ChatUsage) {
+            return (ChatUsage) usage;
+        }
+        if (usage instanceof Map) {
+            @SuppressWarnings("unchecked")
+            Map<String, Object> map = (Map<String, Object>) usage;
+            ChatUsage chatUsage = ChatUsage.builder()
+                    .inputTokens(toInt(map.get("inputTokens")))
+                    .outputTokens(toInt(map.get("outputTokens")))
+                    .time(toDouble(map.get("time")))
+                    .build();
+            metadata.put(MessageMetadataKeys.CHAT_USAGE, chatUsage);
+            return chatUsage;
+        }
+        return null;
+    }
+
+    private static int toInt(Object value) {
+        if (value instanceof Number) {
+            return ((Number) value).intValue();
+        }
+        return 0;
+    }
+
+    private static double toDouble(Object value) {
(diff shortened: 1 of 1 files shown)


yadavsharn:
Hey @mvanhorn, great fix 👍

I noticed the CI build is failing—this might be due to type casting issues when converting values from the Map. Since JSON deserialization can produce `Integer`, `Long`, or `Double`, direct casting might cause failures.

Maybe using `Number` and converting via `.intValue()` / `.doubleValue()` could make it safer.

Happy to help test or contribute improvements if needed 🚀

mvanhorn:
Thanks for the review! The `Number` casting approach is actually already in place (using `.intValue()` / `.doubleValue()`), so the type safety concern should be covered.

The CI failure was a spotless formatting issue with the builder chain indentation. Fixed in 536b3c8.

LearningGp (approved):
LGTM

mvanhorn:
Thanks for the merge, @LearningGp. Map handling in getChatUsage was a real corner case.

PR #1145, opened 2026-04-06
[#570] Preserve AutoContextMemory compression language

### LINKED ISSUE #570: [Bug]:AutoContextMemory的上下文压缩造成思考出现英文
**Describe the bug**
可能是由于上下文压缩默认采用的是英文提示词，在压缩后，产生的总结为英文，导致在上下文中的思考变成了英文了。

**如何解决**
在默认提示词中约定“使用用户提问的语言”来压缩内容。

Alexxigang:
## Summary
- infer the preferred compression language from the first user message (with fallback to the available message text)
- append explicit language-preservation requirements to AutoContextMemory compression and summarization prompts
- thread the language-aware prompt variants through previous-round summaries, tool compression, current-round compression, and large-message summarization
- add regression tests for Chinese, English, mixed-language, and fallback cases

## Why this fix
AutoContextMemory currently builds all compression prompts in English with no hint about the user's language. That makes the summarization model free to rewrite compressed context in English, which is especially harmful for Chinese addresses, names, and domain phrases because they may be translated or converted into pinyin during compression. This patch keeps the compression prompts language-aware so the compressed memory stays aligned with the original conversation language while still preserving multilingual technical fragments verbatim.

## Validation
- mvn -pl agentscope-extensions/agentscope-extensions-autocontext-memory -am "-Dtest=PromptProviderTest,CompressionLanguageHintResolverTest" test

Closes #570.

### DIFF
--- a/agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
+++ b/agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
@@ -582,7 +582,7 @@ public class AutoContextMemory implements StateModule, Memory, ContextOffLoader 
                                 TextBlock.builder()
                                         .text(
                                                 PromptProvider.getCurrentRoundLargeMessagePrompt(
-                                                        customPrompt))
+                                                        customPrompt, List.of(message)))
                                         .build())
                         .build());
         newMessages.add(message);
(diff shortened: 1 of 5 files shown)


LearningGp:
PTAL @shiyiyue1102

LearningGp (changes requested):
In 3/4 scenarios where the USER message cannot be found, the fallback logic concatenates all ASSISTANT/TOOL text for detection. Since tool call content is often saturated with English paths, code, and tool names, this leads to Chinese conversations being misidentified as English, which may actually exacerbate the issue.

jujn:

PR #1146, opened 2026-04-06
fix(autocontext): preserve follow-up turn shape after compression

Alexxigang:
## Summary
- preserve a non-assistant trailing turn after strategy-6 current-round compression
- mark synthetic current-round summaries so internal scans ignore them as real user turns
- add regression coverage for follow-up reasoning turn shape and synthetic-summary metadata

## Root cause
Strategy 6 replaced the post-user tool/result sequence with a synthetic assistant message. For providers that expect the next reasoning request to continue from a non-assistant turn, that leaves the conversation ending with assistant content and can produce an empty follow-up response.

## Fix
- store the current-round compressed summary as a synthetic USER message instead of an ASSISTANT message
- add `MsgUtils` helpers to detect synthetic current-round summaries and ignore them in user-turn scans / interaction extraction
- add regression tests covering the follow-up reasoning shape and metadata preservation

## Validation
- `mvn -pl agentscope-extensions/agentscope-extensions-autocontext-memory -am spotless:apply`
- `mvn -pl agentscope-extensions/agentscope-extensions-autocontext-memory -am "-Dtest=AutoContextMemoryTest,PromptProviderTest,CompressionLanguageHintResolverTest" test`


LearningGp:
PTAL @shiyiyue1102

jujn:

PR #1156, opened 2026-04-07
fix(anthropic): tolerate proxy responses without message ids

Alexxigang:
## Summary
- tolerate Anthropic proxy responses that omit the top-level `id` field
- reuse the SDK raw `_id()` field when available and fall back gracefully when it is missing
- add parser- and formatter-level regression coverage for both non-streaming and message-start streaming paths

## Root cause
`AnthropicResponseParser` called `message.id()` directly in both the non-streaming parse path and the streaming message-start path. Some proxy services strip the top-level `id` field, and the Anthropic SDK throws before AgentScope can build a `ChatResponse`.

## Fix
- add a safe message-id reader that prefers the SDK raw `_id()` field
- fall back to a generated `ChatResponse` id when the payload really has no `id`
- cover raw-id reuse and missing-id fallback in parser and formatter tests

## Validation
- `mvn -pl :agentscope-core -am "-Dtest=AnthropicResponseParserTest,AnthropicChatFormatterTest" test`

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/anthropic/AnthropicResponseParser.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/anthropic/AnthropicResponseParser.java
@@ -94,7 +94,11 @@ public class AnthropicResponseParser {
                         .time(Duration.between(startTime, Instant.now()).toMillis() / 1000.0)
                         .build();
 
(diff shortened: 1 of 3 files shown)


Alexxigang:
Rebased this PR onto the current `main` and reran the targeted Anthropic formatter/parser tests locally. The branch head is refreshed and ready for review again.

LearningGp (changes requested):

PR #1157, opened 2026-04-07
fix(core): sanitize interrupted tool call arguments

### LINKED ISSUE #1147: Interrupted streaming tool call leaves invalid JSON arguments in memory, causing 400 errors on subsequent requests
## Problem

When a streaming tool call response is interrupted (user manually interrupts the agent), the partially accumulated tool call arguments (an incomplete JSON string) are saved into memory. On the next request to the model, this invalid JSON is sent as the `arguments` field of the tool call in the conversation history, causing DashScope (百炼) API to reject the request with a 400 error.

## Root Cause Chain

1. **Streaming accumulation**: `ToolCallsAccumulator.ToolCallBuilder` appends each chunk's argument fragment to `rawContent` StringBuilder during streaming.

2. **Interrupt triggers `build()`**: When interrupted, `ReActAgent.reasoning()` catches `InterruptedException` and calls `context.buildFinalMessage()`, which triggers `ToolCallBuilder.build()`.

3. **`build()` preserves invalid JSON**: In `build()`, when JSON parsing of `rawContent` fails (as expected for incomplete JSON), `input` stays empty but `content` is still set to the raw broken string:
   ```java
   .content(rawContentStr.isEmpty() ? "{}" : rawContentStr)
   ```
   For example, `content` might be `{"query": "hello wor` — a clearly invalid JSON fragment.

4. **Saved to memory**: For user interruptions (non-SYSTEM source), the message with the broken `ToolUseBlock` is always saved to memory via `memory.addMessage(msg)`.

5. **`PendingToolRecoveryHook` doesn't fix arguments**: The recovery hook correctly generates an error `ToolResultBlock` to close the tool call cycle, but does NOT repair the broken `arguments` in the original `ToolUseBlock`.

6. **Converter sends invalid JSON as-is**: Both `DashScopeToolsHelper.convertToolCalls()` and `OpenAIMessageConverter` prioritize the `content` field over `input`:
   ```java
   if (toolUse.getContent() != null && !toolUse.getContent().isEmpty()) {
       argsJson = toolUse.getContent(); // broken JSON sent directly
   }
   ```

7. **API rejects**: DashScope API validates that tool call `arguments` must be valid JSON → **400 error**.

## Affected Models

Any model accessed via DashScope API (qwen3.5, qwen3.6, etc.) that validates historical tool call arguments.

## Proposed Fix

1. **Primary (root cause)**: In `ToolCallBuilder.build()`, validate that `rawContentStr` is valid JSON before using it as `content`. If invalid, fall back to `"{}"`.

2. **Secondary (defense in depth)**: In `DashScopeToolsHelper.convertToolCalls()` and `OpenAIMessageConverter`, validate `content` is valid JSON before using it as `arguments`. If invalid, fall back to serializing `input` or `"{}"`.

Alexxigang:
## Summary
- sanitize interrupted streaming tool-call argument payloads before they are persisted into memory
- add a shared JSON normalization helper used by the accumulator and provider converters
- add regression tests for accumulator recovery plus DashScope/OpenAI formatter fallback behavior

## Why this fix
When a streaming tool call is interrupted mid-JSON, AgentScope can persist a partial `ToolUseBlock.content` string into memory. On the next request, the DashScope/OpenAI formatter path reuses that invalid JSON as tool-call arguments, which can trigger provider-side 400 errors.

This patch fixes the bug at two layers:
- `ToolCallsAccumulator` now normalizes incomplete raw argument fragments to either serialized structured input or `{}` before the message is finalized
- `DashScopeToolsHelper` and `OpenAIMessageConverter` now defensively sanitize historical `ToolUseBlock.content` before sending it back to providers

Fixes #1147.

## Validation
- `mvn -pl :agentscope-core -am "-Dtest=ToolCallsAccumulatorTest,DashScopeToolsHelperComprehensiveTest,OpenAIMessageConverterTest,ToolCallJsonUtilsTest" test`

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
@@ -17,7 +17,7 @@ package io.agentscope.core.agent.accumulator;
 
 import io.agentscope.core.message.ContentBlock;
 import io.agentscope.core.message.ToolUseBlock;
-import io.agentscope.core.util.JsonUtils;
+import io.agentscope.core.util.ToolCallJsonUtils;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


Alexxigang:
Rebased this PR onto the current `main` and reran the targeted accumulator / formatter / JSON sanitization tests locally. The branch head is refreshed and ready for review again.

LearningGp:
FYI, #1148

PR #1162, opened 2026-04-08
refactor(core): remove redundant hasPendingToolUse check in doCall

xichaodong:
The check at line 275 is always true because the early return at line 268 already guarantees pendingIds is non-empty at that point. Simplify to call acting(0) directly.


LearningGp (approved):
LGTM

PR #1167, opened 2026-04-08
fix: resolve three bugs in core agent and tool management (#1164 #1165 #1166)

### LINKED ISSUE #1164: Bug: StaticLongTermMemoryHook injects memory at end of messages instead of beginning
## Bug Description

In `StaticLongTermMemoryHook`, the retrieved long-term memory message is appended to the **end** of the message list, despite the code comment explicitly stating it should be injected **at the beginning**.

## Affected File

`agentscope-core/src/main/java/io/agentscope/core/memory/StaticLongTermMemoryHook.java`, lines 157–160

## Current Behavior

```java
// Inject memory message at the beginning   // <-- comment says beginning
List<Msg> enhancedMessages = new ArrayList<>();
enhancedMessages.addAll(inputMessages);      // all messages added first
enhancedMessages.add(memoryMsg);             // memory appended at the END
event.setInputMessages(enhancedMessages);
```

The LLM receives messages in this order:
1. System prompt
2. Conversation history / user query
3. Retrieved memory context ← **appended last**

## Expected Behavior

The retrieved memory context should be injected **at the beginning** (or at least before the user query), so the LLM has the relevant context when processing the current turn:

```java
// Inject memory message at the beginning
List<Msg> enhancedMessages = new ArrayList<>();
enhancedMessages.add(0, memoryMsg);          // insert at index 0
enhancedMessages.addAll(inputMessages);
event.setInputMessages(enhancedMessages);
```

Or equivalently:

```java
enhancedMessages.addAll(inputMessages);
enhancedMessages.add(0, memoryMsg);          // insert at beginning after the fact
```

## Impact

- RAG (Retrieval-Augmented Generation) quality is degraded because the LLM receives the user query **before** the retrieved context
- The retrieved memory has less influence on the response since it appears after the question has already been posed in the message sequence
- The comment/code mismatch indicates this is an unintentional regression

## Steps to Reproduce

1. Configure a `ReActAgent` with `StaticLongTermMemoryHook`
2. Send a query that triggers a long-term memory retrieval
3. Inspect the message list passed to the LLM — the memory `SYSTEM` message will appear last instead of first

## Suggested Fix

Change line 160 in `StaticLongTermMemoryHook.java`:

```java
// Before
enhancedMessages.add(memoryMsg);

// After
enhancedMessages.add(0, memoryMsg);
```

### LINKED ISSUE #1165: Bug: ToolGroupManager.activeGroups is not thread-safe, causing race conditions in concurrent agent scenarios
## Bug Description

`ToolGroupManager.activeGroups` is declared as a plain `ArrayList`, while the other two maps in the same class use `ConcurrentHashMap`. In multi-agent or concurrent execution scenarios, multiple threads can read and write `activeGroups` simultaneously, leading to race conditions.

## Affected File

`agentscope-core/src/main/java/io/agentscope/core/tool/ToolGroupManager.java`, line 36

## Current Code

```java
private final Map<String, ToolGroup> toolGroups = new ConcurrentHashMap<>(); // thread-safe ✓
private final Map<String, Set<String>> tools = new ConcurrentHashMap<>();    // thread-safe ✓
private List<String> activeGroups = new ArrayList<>();                       // NOT thread-safe ✗
```

`activeGroups` is accessed (read or written) from multiple methods without synchronization:

| Line | Method | Operation |
|------|--------|-----------|
| 57–58 | `createToolGroup()` | read + write |
| 92–98 | `updateToolGroups()` | read + write |
| 132 | `removeToolGroups()` | write |
| 152 | `getActivatedNotes()` | iterate |
| 317–318 | `getActiveGroups()` | read |
| 327 | `setActiveGroups()` | **field reassignment** (not `volatile`) |
| 377 | `copyTo()` | write |

Additionally, `setActiveGroups()` replaces the field reference entirely (`this.activeGroups = new ArrayList<>(...)`), but the field is neither `volatile` nor `final`, so other threads may observe a stale reference.

## Expected Behavior

All accesses to `activeGroups` should be thread-safe, consistent with the rest of the class.

## Impact

When multiple agents share or concurrently modify a `ToolGroupManager`:
- `ConcurrentModificationException` during iteration in `getActivatedNotes()`
- Lost updates: tool group activation/deactivation silently dropped
- Stale reads: a thread may see the old list after `setActiveGroups()` replaces it
- Agents may execute with incorrect tool availability

## Suggested Fix

Replace `ArrayList` with `CopyOnWriteArrayList` (appropriate for read-heavy workloads) and make the field `final` to prevent unsafe reference replacement:

```java
// Before
private List<String> activeGroups = new ArrayList<>();

// After
private final List<String> activeGroups = new CopyOnWriteArrayList<>();
```

`setActiveGroups()` should then clear and re-add rather than reassign:

```java
public void setActiveGroups(List<String> activeGroups) {
    this.activeGroups.clear();
    this.activeGroups.addAll(activeGroups);
    // ... rest of the method
}
```

Similarly, `copyTo()` should use the same pattern instead of direct field assignment.

### LINKED ISSUE #1166: Bug: NullPointerException in MessageUtils.extractRecentToolCalls when assistant message has no name
## Bug Description

`MessageUtils.extractRecentToolCalls()` calls `.equals()` directly on `msg.getName()` without a null check. Since `Msg.name` is an optional field that can be `null`, this throws a `NullPointerException` whenever an `ASSISTANT` message with no name is present in the message list.

## Affected File

`agentscope-core/src/main/java/io/agentscope/core/util/MessageUtils.java`, line 54

## Current Code

```java
for (int i = messages.size() - 1; i >= 0; i--) {
    Msg msg = messages.get(i);
    if (msg.getRole() == MsgRole.ASSISTANT && msg.getName().equals(agentName)) {  // ← NPE here
        List<ToolUseBlock> toolCalls = msg.getContentBlocks(ToolUseBlock.class);
        if (!toolCalls.isEmpty()) {
            return toolCalls;
        }
        break;
    }
}
```

`msg.getName()` returns `null` when the message was constructed without a name (the field is not required by `Msg.builder()`). Calling `.equals()` on `null` throws `NullPointerException`.

## Expected Behavior

The method should handle `null` names gracefully and simply skip messages where the name does not match.

## Impact

- Any `ASSISTANT` message in memory that was created without a `name` field will crash `extractRecentToolCalls()`
- This method is called during the ReAct agent's acting phase to look up pending tool calls
- The crash propagates up and terminates the agent's reasoning loop
- Third-party integrations that construct `Msg` objects without names (a valid use of the API) are silently broken

## Steps to Reproduce

```java
// Create an assistant message without a name
Msg assistantMsg = Msg.builder()
    .role(MsgRole.ASSISTANT)
    // name intentionally omitted
    .content(TextBlock.builder().text("Hello").build())
    .build();

List<Msg> messages = List.of(assistantMsg);

// This throws NullPointerException
MessageUtils.extractRecentToolCalls(messages, "MyAgent");
```

## Suggested Fix

Use `Objects.equals()` which handles `null` safely:

```java
// Before
if (msg.getRole() == MsgRole.ASSISTANT && msg.getName().equals(agentName)) {

// After
if (msg.getRole() == MsgRole.ASSISTANT && Objects.equals(msg.getName(), agentName)) {
```

Fruank4:
## Summary

This PR fixes three bugs identified in the core library:

- **#1164** \`StaticLongTermMemoryHook\` injects retrieved memory at the beginning of the message list with a \`SYSTEM\` role, which breaks message sequence constraints and degrades RAG quality
- **#1165** \`ToolGroupManager.activeGroups\` uses a non-thread-safe \`ArrayList\`, causing race conditions in concurrent agent scenarios
- **#1166** \`MessageUtils.extractRecentToolCalls\` calls \`.equals()\` on \`msg.getName()\` without a null check, throwing \`NullPointerException\` when an assistant message has no name

### Fix #1164 — \`StaticLongTermMemoryHook.java\`

Change memory message role from \`SYSTEM\` to \`USER\` and append it at the end of the message list so the sequence remains valid and the memory context is closest to the model's generation point:

\`\`\`java
// Before
Msg memoryMsg = Msg.builder().role(MsgRole.SYSTEM)...build();
List<Msg> enhancedMessages = new ArrayList<>();
enhancedMessages.addAll(inputMessages);
enhancedMessages.add(memoryMsg);   // appended at end but wrong role

// After
Msg memoryMsg = Msg.builder().role(MsgRole.USER)...build();
List<Msg> enhancedMessages = new ArrayList<>(inputMessages);
enhancedMessages.add(memoryMsg);   // USER role, appended at end
\`\`\`

### Fix #1165 — \`ToolGroupManager.java\`

Replace the non-thread-safe \`ArrayList\` with \`CopyOnWriteArrayList\` (declared \`volatile\`) and update \`setActiveGroups()\` / \`copyTo()\` to replace the reference atomically, avoiding the transient empty-list window that \`clear()\` + \`addAll()\` would create:

\`\`\`java
// Before
private List<String> activeGroups = new ArrayList<>();
// ...
this.activeGroups = new ArrayList<>(activeGroups);
target.activeGroups = new ArrayList<>(this.activeGroups);

// After
private volatile List<String> activeGroups = new CopyOnWriteArrayList<>();
// ...
this.activeGroups = new CopyOnWriteArrayList<>(activeGroups);
target.activeGroups = new CopyOnWriteArrayList<>(this.activeGroups);
\`\`\`

### Fix #1166 — \`MessageUtils.java\`

Replace \`msg.getName().equals(agentName)\` with \`Objects.equals()\` to safely handle \`null\` names:

\`\`\`java
// Before
if (msg.getRole() == MsgRole.ASSISTANT && msg.getName().equals(agentName)) {

// After
if (msg.getRole() == MsgRole.ASSISTANT && Objects.equals(msg.getName(), agentName)) {
\`\`\`

## Test plan

- [x] Existing unit tests pass (\`mvn test -pl agentscope-core\`)
- [x] \`StaticLongTermMemoryHookTest\` updated to assert \`USER\` role and end-of-list position
- [ ] Verify no \`ConcurrentModificationException\` when multiple agents modify tool groups concurrently
- [ ] Verify no \`NullPointerException\` when assistant messages without names are present in memory


Fruank4:
三个问题：空指针，并发安全，记忆反序

Fruank4:
给孩子过了吧 呜呜呜

LearningGp:
lichuang34 这个账号没有签署 CLA，要不用Fruank4这个账号提交下后两个commit

Fruank4:
CLA验证了

agentscope-core/src/main/java/io/agentscope/core/tool/ToolGroupManager.java
```diff
@@ -324,7 +325,8 @@ public List<String> getActiveGroups() {
      * @param activeGroups List of group names to mark as active
      */
     public void setActiveGroups(List<String> activeGroups) {
-        this.activeGroups = new ArrayList<>(activeGroups);
+        this.activeGroups.clear();
```
LearningGp: clear() and addAll() are non-atomic. CopyOnWriteArrayList only ensures thread-safety for single operations, so readers may see a transient empty list in between. Suggest replacing the reference directly instead.

Note that the check-then-act logic in createToolGroup still has a potential race condition: a thread could write to a stale list if the reference is swapped mid-operation. This mirrors the legacy behavior, so I'm fine with ignoring it in this PR.

agentscope-core/src/main/java/io/agentscope/core/memory/StaticLongTermMemoryHook.java
```diff
@@ -156,8 +156,8 @@ private Mono<PreCallEvent> handlePreCall(PreCallEvent event) {
 
                             // Inject memory message at the beginning
                             List<Msg> enhancedMessages = new ArrayList<>();
-                            enhancedMessages.addAll(inputMessages);
                             enhancedMessages.add(memoryMsg);
```
LearningGp, line 159: Suggest changing the role for long-term memory messages to user, while ensuring they remain positioned at the end of the sequence.

LearningGp (changes requested):

agentscope-core/src/main/java/io/agentscope/core/tool/ToolGroupManager.java
Fruank4: ok

agentscope-core/src/main/java/io/agentscope/core/memory/StaticLongTermMemoryHook.java
Fruank4: ok

LearningGp (approved):
LGTM

Fruank4:
@LearningGp hibro，你们部门有招人计划吗，我想面一面

PR #1168, opened 2026-04-08
fix(extensions-agui): convert SUMMARY events to AG-UI messages (#1029)

### LINKED ISSUE #1029: [Bug]: [extensions-agui] 超过maxIters了,  是否也需要将SUMMARY事件转换为AG-UI事件
当超过 maxIters 时，用户端无法收到conversation总结内容，是否也需要将 SUMMARY 事件转换为 AG-UI 事件？

lynx009:
## Description

Fixes #1029.

When a ReAct agent reaches `maxIters`, AgentScope emits a `SUMMARY` event as a fallback response.
Before this change, `extensions-agui` only converted `REASONING` events into AG-UI text events, so the client only received `RUN_FINISHED` and could not see the summary content.

This PR updates the AG-UI adapter to also convert `EventType.SUMMARY` using the existing text-message mapping path.
As a result, when the agent stops because it exceeds `maxIters`, the frontend can still receive the final summary content through AG-UI events.

Changes included:
- Handle `EventType.SUMMARY` in the AG-UI adapter
- Reuse the existing text/reasoning event conversion flow
- Add regression tests for normal summary and streaming summary cases

How to verify:
- Configure an agent with `maxIters=1`
- Send a request that requires at least one tool call before producing the final answer
- Before the fix:
  - the AG-UI stream ends after `TOOL_CALL_RESULT` with `RUN_FINISHED`
  - no assistant summary text is delivered
- After the fix:
  - the AG-UI stream includes `TEXT_MESSAGE_START / TEXT_MESSAGE_CONTENT / TEXT_MESSAGE_END`
  - the summary content is visible to the client

## Checklist

- [x] Code has been formatted
- [x] All tests are passing 

### DIFF
--- a/agentscope-extensions/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java
+++ b/agentscope-extensions/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java
@@ -325,14 +325,14 @@ public class AguiAgentAdapter {
         StringBuilder sb = new StringBuilder();
         for (ContentBlock output : toolResult.getOutput()) {
             if (output instanceof TextBlock textBlock) {
-                if (sb.length() > 0) {
+                if (!sb.isEmpty()) {
                     sb.append("\n");
                 }
                 sb.append(textBlock.getText());
             }
         }
 
-        return sb.length() > 0 ? sb.toString() : null;
+        return !sb.isEmpty() ? sb.toString() : null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lynx009:
Setting maxIters to 1, the single round ReAct test samples are as follows, and the same results are obtained for multiple rounds of ReAct test samples.

Before the fix:
截屏2026-04-08 22 04 29
When reaching `maxIters`, it is necessary to inquire about the results again:
截屏2026-04-08 22 15 42
After the fix:
截屏2026-04-08 22 05 41

LearningGp (approved):
LGTM

PR #1169, opened 2026-04-08
fix(core): mirror summary thinking into reasoning events

### LINKED ISSUE #1152: DashScopeChatModel: Qwen3.5-plus thinking chunks may be emitted only via SummaryChunkEvent instead of ReasoningChunkEvent
When using DashScopeChatModel with qwen3.5-plus in thinking mode, reasoning output is not always emitted as
  ReasoningChunkEvent.

  In my case, the model does return reasoning content, but sometimes it is only present inside the incremental Msg carried by
  SummaryChunkEvent as ThinkingBlock, while no separate ReasoningChunkEvent is fired.

  This causes an integration problem for streaming UIs and hooks that follow the current event contract literally:

  - ReasoningChunkEvent -> render reasoning/thinking
  - SummaryChunkEvent -> render final answer text

  With qwen3.5-plus, this assumption can fail, because reasoning may be mixed into SummaryChunkEvent. As a result, the
  frontend may appear to receive no reasoning at all unless it additionally parses ThinkingBlock from SummaryChunkEvent.

  Environment:

  - AgentScope Java: 1.0.10
  - DashScope SDK: 2.0.0
  - Model: qwen3.5-plus
  - Streaming: true
  - Thinking: true
  - Additional body params:
      - result_format=message
      - incremental_output=true

  Simplified setup:

  DashScopeChatModel chatModel = DashScopeChatModel.builder()
          .apiKey(apiKey)
          .baseUrl("https://dashscope.aliyuncs.com/compatible-mode/v1")
          .modelName("qwen3.5-plus")
          .stream(true)
          .enableThinking(true)
          .defaultOptions(GenerateOptions.builder()
                  .additionalBodyParam("result_format", "message")
                  .additionalBodyParam("incremental_output", true)
                  .build())
          .build();

  Observed behavior:

  - Some reasoning text is only available from:
      - SummaryChunkEvent#getIncrementalChunk()
      - Msg#getContentBlocks(ThinkingBlock.class)
  - No corresponding ReasoningChunkEvent is emitted for that same reasoning delta

  Expected behavior:

  - If an incremental chunk contains ThinkingBlock, AgentScope should emit a ReasoningChunkEvent for it
  - Or, if this is intentional, the documentation should explicitly state that reasoning may arrive via SummaryChunkEvent and
    consumers must inspect ThinkingBlock manually

  Why this looks like a bug:

  - Other model/provider paths are easier to consume because reasoning arrives as ReasoningChunkEvent
  - For qwen3.5-plus, downstream consumers need model-specific workaround logic
  - This breaks the abstraction promised by hook event types

  Current workaround on my side:

  - Continue handling ReasoningChunkEvent
  - Also inspect SummaryChunkEvent#getIncrementalChunk() and manually split:
      - ThinkingBlock -> reasoning stream
      - TextBlock -> normal text stream

  It would be great if AgentScope could unify this behavior at the framework level.

Alexxigang:
## Summary
- mirror summary-phase `ThinkingBlock` chunks into `ReasoningChunkEvent`
- keep the existing `SummaryChunkEvent` behavior for summary streaming intact
- add regression tests for summary thinking delivery, accumulated reasoning content, and hook event order

## Why this fix
`ReActAgent` uses a dedicated summary streaming path after max-iteration fallback. In that path, a streamed `ThinkingBlock` currently only triggers `SummaryChunkEvent`, even though the same provider/model behavior is exposed as `ReasoningChunkEvent` in normal reasoning streams.

That breaks the hook abstraction for clients that render reasoning exclusively from `ReasoningChunkEvent`, especially with models like `qwen3.5-plus` that may place thinking deltas inside the summary stream.

This patch keeps summary hooks working as before, while also mirroring summary-phase thinking chunks into `ReasoningChunkEvent` with accumulated thinking content.

Fixes #1152.

## Validation
- `mvn -pl :agentscope-core -am "-Dtest=ReActAgentSummaryThinkingHookTest" test`
- `mvn -pl :agentscope-core -am "-Dtest=ReActAgentTest,ReActAgentThinkingCumulativeTest,ReActAgentSummaryThinkingHookTest" test`

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -936,45 +936,21 @@ public class ReActAgent extends StructuredOutputCapableAgent {
     }
 
     private Mono<Void> notifyReasoningChunk(Msg chunkMsg, ReasoningContext context) {
-        ContentBlock content = chunkMsg.getFirstContentBlock();
+        return notifyReasoningChunk(chunkMsg, context, null);
+    }
 
-        ContentBlock accumulatedContent = null;
-        if (content instanceof TextBlock) {
-            accumulatedContent = TextBlock.builder().text(context.getAccumulatedText()).build();
-        } else if (content instanceof ThinkingBlock) {
-            accumulatedContent =
-                    ThinkingBlock.builder().thinking(context.getAccumulatedThinking()).build();
-        } else if (content instanceof ToolUseBlock tub) {
-            // Support streaming ToolUseBlock events
-            ToolUseBlock accumulated = context.getAccumulatedToolCall(tub.getId());
-            if (accumulated != null) {
-                accumulatedContent = accumulated;
-            } else {
-                // If no accumulated data, use the current chunk directly
-                accumulatedContent = tub;
-            }
+    private Mono<Void> notifyReasoningChunk(
+            Msg chunkMsg, ReasoningContext context, GenerateOptions generateOptions) {
+        ContentBlock accumulatedContent = resolveAccumulatedReasoningContent(chunkMsg, context);
+        if (accumulatedContent == null) {
+            return Mono.empty();
         }
 
-        if (accumulatedContent != null) {
-            Msg accumulated =
-                    Msg.builder()
-                            .id(chunkMsg.getId())
-                            .name(chunkMsg.getName())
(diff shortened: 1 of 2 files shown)


Alexxigang:
Rebased this PR onto the current `main` and refreshed the head branch so it can be re-evaluated without the previous merge-conflict state.

LearningGp:

PR #1170, opened 2026-04-08
fix(tool): allow null for optional nested object fields

### LINKED ISSUE #1163: [Bug]:Tool parameter conversion bug
Passing an object as a tool parameter will definitely result in an error if any field is empty. Unless converted to an empty string or empty array, etc.

Image

Image

Image

JGoP-L:
## AgentScope-Java Version

1.0.12-SNAPSHOT

## Description

Fixes #1163

### Background

When a tool parameter contains a nested object, AgentScope-Java currently treats these two cases differently for optional fields:

1. The optional field is omitted
2. The optional field is explicitly set to `null`

Case 1 passes validation, but case 2 fails during JSON Schema validation with an error like:

`null found, string expected`

This makes tool calling brittle in practice, because LLMs often emit `null` for optional fields instead of omitting them entirely.

### Purpose

Make optional nested object fields behave consistently:
- missing optional field: valid
- optional field explicitly set to `null`: also valid

Required field validation should continue to work as before.

#### 1. Fix tool input validation for explicit `null` optional fields

Updated `ToolValidator.validateInput(...)` to normalize tool input before JSON Schema validation.

It now recursively removes `null`-valued object properties before validating the payload. This means:
- optional fields with `null` are treated the same as omitted fields
- required fields with `null` still fail validation, because the field is removed and then checked by `required`

Modified file:
- `agentscope-core/src/main/java/io/agentscope/core/tool/ToolValidator.java`

#### 2. Add regression tests for generated nested bean schema

Added tests to verify:
- missing optional nested bean field is accepted
- explicit `null` optional nested bean field is accepted

Modified file:
- `agentscope-core/src/test/java/io/agentscope/core/tool/ToolValidatorTest.java`

#### 3. Add protection tests for object parameter conversion

Added tests to confirm that nested bean parameter conversion itself already works correctly for:
- missing optional field
- explicit `null` optional field

This verifies the bug is in validation behavior rather than conversion behavior.

Modified file:
- `agentscope-core/src/test/java/io/agentscope/core/tool/ToolMethodInvokerTest.java`

### How to test

Run:

```bash
mvn -pl agentscope-core -Dtest=ToolValidatorTest,ToolMethodInvokerTest test

### CODE REVIEW

File: agentscope-core/src/main/java/io/agentscope/core/tool/ToolValidator.java [hunk 1]
```diff
@@ -95,6 +103,104 @@ public static String validateInput(String input, Map<String, Object> schema) {
...
+            JsonNode itemsSchema = resolvedSchema.get("items");
+            for (JsonNode item : arrayNode) {
+                if (itemsSchema != null) {
+                    pruneOptionalNullObjectFields(item, itemsSchema, schemaRoot);
+                }
+            }
+        }
+    }
+
+    private static JsonNode resolveSchemaNode(JsonNode schemaNode, JsonNode schemaRoot) {
```
LearningGp, line 181: Missing visited tracking for $ref here. Circular references could cause an infinite loop. Although victools schemas are currently safe, it's better to add a safeguard for defensive programming. This can be handled in a future PR.

LearningGp (approved):

PR #1171, opened 2026-04-08
fix(tool): support Flux-returning tool methods

Alexxigang:
## Summary
- add native Flux handling in ToolMethodInvoker instead of treating it as a synchronous object
- stream each emitted Flux item through ToolEmitter so tool chunk callbacks receive incremental updates
- aggregate Flux<String> items into the final tool result while preserving sensible behavior for single-value and multi-value non-string streams
- add regression coverage for invoker-level aggregation/chunk emission and toolkit-level async execution

## Why this fix
Issue #938 happens because invokeAsync(...) only special-cases CompletableFuture and Mono. When a tool method returns Flux<?>, the current code falls through the synchronous branch, serializes the Flux object itself, and never subscribes to the stream. That means the parent agent never waits for or consumes the streamed tool output.

This change subscribes to Flux results, forwards intermediate chunks to hooks, and converts the aggregated stream output into the final ToolResultBlock returned to the LLM-facing caller.

## Validation
- mvn -pl agentscope-core '-Dtest=AsyncToolTest,ToolMethodInvokerTest' test
- mvn -pl agentscope-core '-Dtest=ToolIntegrationTest' test

## AI Assistance
- Drafted with AI assistance and reviewed/validated locally before submission.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolMethodInvoker.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolMethodInvoker.java
@@ -25,8 +25,10 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 /**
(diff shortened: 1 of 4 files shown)


Alexxigang:

agentscope-core/src/main/java/io/agentscope/core/tool/ToolMethodInvoker.java
```diff
@@ -119,6 +124,101 @@ r, extractGenericType(method)))
...
+                            Flux<Object> flux = (Flux<Object>) method.invoke(toolObject, args);
+                            return flux != null ? flux : Flux.empty();
+                        })
+                .flatMap(
+                        flux ->
+                                flux.doOnNext(
+                                                item ->
+                                                        emitFluxChunk(
+                                                                emitter, converter, item, itemType))
+                                        .collectList()
```
LearningGp, line 152: Caution: collectList() collects the entire Flux into memory. Using this on an unbounded or extremely large stream (like a DB cursor) risks triggering an OOM.

LearningGp (changes requested):

PR #1172, opened 2026-04-09
fix: prevent presetParameters from being overridden by tool-call input

### LINKED ISSUE #1119: [Bug] presetParameters overridden by LLM input due to incorrect merge order in ToolExecutor
## Bug Description

In `ToolExecutor.java`, preset parameters registered via `Toolkit.presetParameters()` 
can be silently overridden by LLM-generated tool call inputs due to incorrect `putAll` 
merge order.

## Root Cause

The merge logic places `presetParameters` first and LLM input second, allowing LLM 
output to overwrite any preset value with the same key:

```java
// Current implementation (incorrect)
Map<String, Object> mergedInput = new HashMap<>();
if (registered != null) {
    mergedInput.putAll(registered.getPresetParameters()); // ① preset written first
}
if (param.getInput() != null && !param.getInput().isEmpty()) {
    mergedInput.putAll(param.getInput());                 // ② LLM input overwrites preset 
} else if (toolCall.getInput() != null) {
    mergedInput.putAll(toolCall.getInput());              // ③ same issue
}
```

The intended semantics of `presetParameters` is that they should be **immutable 
from the LLM's perspective** — values injected by the framework must not be 
overridable by model output.

## Expected Behavior

Preset parameters should take precedence over LLM-generated input. The correct 
merge order should be:

```java
// Correct implementation
Map<String, Object> mergedInput = new HashMap<>();
// LLM input first (lower priority)
if (param.getInput() != null && !param.getInput().isEmpty()) {
    mergedInput.putAll(param.getInput());
} else if (toolCall.getInput() != null) {
    mergedInput.putAll(toolCall.getInput());
}
// Preset parameters last — always wins, cannot be overridden by LLM 
if (registered != null) {
    mergedInput.putAll(registered.getPresetParameters());
}
```

## How to Trigger

Although `Toolkit.getParameters()` correctly filters preset keys from the JSON Schema 
sent to the LLM (so the model shouldn't know these fields exist), the override can 
still occur in the following scenarios:

1. **Conversation history leakage** — if a previous `tool_use` message in the context 
   window contained preset field names, the LLM may reproduce them in subsequent calls
2. **Description hints** — field descriptions containing words like "hidden", "JWT", 
   or "do not fill" can inadvertently guide the LLM to guess the field name
3. **LLM prior knowledge** — large models have strong priors on common API credential 
   parameter names (e.g., `Authorization`, `token`, `api_key`) and may generate them 
   unprompted

## Impact

- Security: credential fields (e.g., API keys, JWT tokens) injected via 
  `presetParameters` can be silently replaced by LLM-fabricated values
- Reliability: tool calls may fail intermittently due to unexpected credential 
  values, which are difficult to reproduce and diagnose
- Trust: the `presetParameters` API contract is violated — callers reasonably expect 
  preset values to be immutable during execution

## Suggested Fix

Swap the `putAll` order in `ToolExecutor.java` so preset parameters are applied last 
and always take precedence over LLM input (see Expected Behavior above).

Alternatively, consider providing a dedicated `ToolExecutionContext` injection 
mechanism (already present in the framework) as the recommended approach for 
injecting framework-controlled, LLM-invisible parameters into `@Tool` methods — 
bypassing the `input` map entirely.

JGoP-L:
## AgentScope-Java Version

1.0.12-SNAPSHOT

## Description

This PR fixes an issue where `presetParameters` could be overridden by tool-call input during tool execution.

Background:
`presetParameters` are intended for framework-controlled values such as API keys or other hidden parameters. While these fields are excluded from the tool schema shown to the LLM, the execution-time merge order in `ToolExecutor` allowed tool-call input to overwrite preset values when the same key appeared in both places.

Changes made:
- Updated `ToolExecutor` to merge tool-call input first and apply `presetParameters` last
- Ensured preset parameters take precedence during execution and cannot be overridden by LLM-generated or caller-provided input
- Updated the related unit test to reflect the intended precedence behavior

How to test:
- Run:
  - `mvn -pl agentscope-core -Dtest=io.agentscope.core.tool.ToolkitTest#testPresetParameters test`
  - `mvn -pl agentscope-core -Dtest=io.agentscope.core.tool.ToolkitTest#testPresetParametersOverride test`
  - `mvn -pl agentscope-core -Dtest=io.agentscope.core.tool.ToolkitTest test`
  - `mvn -pl agentscope-core -Dtest=ToolExecutorTest test`
- Verify that preset parameters are still injected correctly
- Verify that when the same parameter exists in both tool-call input and `presetParameters`, the preset value wins

Closes #1119

## Checklist


agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
```diff
@@ -219,16 +219,15 @@ private Mono<ToolResultBlock> executeCore(ToolCallParam param) {
...
-            mergedInput.putAll(registered.getPresetParameters());
-        }
-        if (param.getInput() != null && !param.getInput().isEmpty()) {
+        if (!param.getInput().isEmpty()) {
             mergedInput.putAll(param.getInput());
-        } else if (toolCall.getInput() != null) {
+        } else if (!toolCall.getInput().isEmpty()) {
             mergedInput.putAll(toolCall.getInput());
         }
+        mergedInput.putAll(registered.getPresetParameters());
```
LearningGp: Updates to tools and registeredTools are non-atomic. There’s a slim race condition where registered == null. Suggest keeping the guard or adding an assertion to safeguard against NPEs from any future registration logic.

LearningGp (changes requested):

LearningGp (approved):
LGTM

PR #1177, opened 2026-04-09
fix issue #1160

xuhuafeifei:
## AgentScope-Java Version

## Description

LongTermMemory will block the agent when the LLM stops responding. This commit enables the framework to record long term memory asynchronously.

Added `longTermMemoryAsyncRecord` option to `ReActAgent.Builder` to support fire-and-forget long-term memory recording in `STATIC_CONTROL` / `BOTH` mode.

When enabled, the `StaticLongTermMemoryHook` performs `record()` asynchronously without blocking the agent's response, reducing latency in `STATIC_CONTROL` mode. Agent-controlled recording (`AGENT_CONTROL`) remains synchronous.

## Checklist

- `ReActAgent.Builder` — new `longTermMemoryAsyncRecord(boolean)` setter
- `StaticLongTermMemoryHook` — overloaded constructor accepting `asyncRecord` flag; `handlePostCall()` splits sync/async paths
- `StaticLongTermMemoryHookTest` — added 3 async recording test cases

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/memory/StaticLongTermMemoryHook.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/memory/StaticLongTermMemoryHook.java
@@ -76,15 +76,29 @@ public class StaticLongTermMemoryHook implements Hook {
 
     private final LongTermMemory longTermMemory;
     private final Memory memory;
+    private final boolean asyncRecord;
 
     /**
-     * Creates a new StaticLongTermMemoryHook.
+     * Creates a new StaticLongTermMemoryHook with synchronous recording.
      *
      * @param longTermMemory The long-term memory instance for persistent storage
      * @param memory The agent's memory for accessing conversation history
      * @throws IllegalArgumentException if longTermMemory or memory is null
      */
     public StaticLongTermMemoryHook(LongTermMemory longTermMemory, Memory memory) {
+        this(longTermMemory, memory, false);
+    }
+
+    /**
+     * Creates a new StaticLongTermMemoryHook.
+     *
+     * @param longTermMemory The long-term memory instance for persistent storage
+     * @param memory The agent's memory for accessing conversation history
+     * @param asyncRecord Whether to record memories asynchronously (fire-and-forget)
+     * @throws IllegalArgumentException if longTermMemory or memory is null
+     */
+    public StaticLongTermMemoryHook(
+            LongTermMemory longTermMemory, Memory memory, boolean asyncRecord) {
         if (longTermMemory == null) {
             throw new IllegalArgumentException("Long-term memory cannot be null");
         }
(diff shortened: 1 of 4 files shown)


agentscope-core/src/main/java/io/agentscope/core/memory/StaticLongTermMemoryHook.java
```diff
@@ -191,16 +211,35 @@ private Mono<PostCallEvent> handlePostCall(PostCallEvent event) {
...
-                .record(allMessages)
-                .thenReturn(event)
-                .onErrorResume(
-                        error -> {
-                            // Log error but don't interrupt the flow
-                            log.warn(
-                                    "Failed to record to long-term memory: {}", error.getMessage());
-                            return Mono.just(event);
-                        });
+        if (asyncRecord) {
```
LearningGp, line 227: Reactor Context is missing in this async path. Bare .subscribe() calls result in an empty tracing context, bypassing the Hooks.onEachOperator setup in TracerRegistry. We should apply the same fix as in AgentBase.createEventStream()—using deferContextual to capture and contextWrite to propagate the context.

agentscope-core/src/main/java/io/agentscope/core/memory/StaticLongTermMemoryHook.java
```diff
@@ -191,16 +211,35 @@ private Mono<PostCallEvent> handlePostCall(PostCallEvent event) {
...
+                    .record(allMessages)
+                    .subscribeOn(Schedulers.boundedElastic())
+                    .onErrorResume(
+                            error -> {
+                                log.warn(
+                                        "Failed to asynchronously record to long-term memory: {}",
+                                        error.getMessage());
+                                return Mono.empty();
+                            })
+                    .subscribe();
```
LearningGp: Discarding the Disposable means in-progress records won't be cancelled on shutdown/destruction. Also, slow record() calls might lead to unbounded task queuing in boundedElastic. We should either document these trade-offs in the Javadoc or manage the Disposable lifecycle by cancelling it when the hook is detached.

LearningGp (changes requested):

agentscope-core/src/main/java/io/agentscope/core/memory/StaticLongTermMemoryHook.java
xuhuafeifei: I will fix it

agentscope-core/src/main/java/io/agentscope/core/memory/StaticLongTermMemoryHook.java
xuhuafeifei: I will fix it

LearningGp (approved):
LGTM

Issue #1176, opened 2026-04-09
[Bug]:Schema Validation Error with Reused Objects in Structured Output


killer9527:
# Bug Report: Schema Validation Error with Reused Objects in Structured Output

**<u>AgentScope-Java is an open-source project. To involve a broader community, we recommend asking your questions in English.</u>**

## Describe the bug

When using structured output with reused objects (i.e., the same class used multiple times in a parent class), the framework generates a JSON Schema with `$defs` and `$ref` references. However, during parameter validation, the schema validator fails to resolve these references, resulting in a `Schema validation error: Reference /$defs/Address cannot be resolved` error.

This prevents the use of reusable object structures in structured output, which is a common and important pattern for complex data extraction scenarios.

### 1. Your code

Create a structured output class with reused objects:

```java
/** Schema for order information with address reuse. */
public static class OrderInfo {
    public String orderId;
    public String orderStatus;
    public String customerName;
    public String customerPhone;
    public String customerEmail;
    public Address billingAddress;      // Reused Address class
    public Address shippingAddress;     // Reused Address class (same as billingAddress)
    public List<OrderItem> items;
    public Double subtotal;
    public Double tax;
    public Double shippingFee;
    public Double totalAmount;
    public String paymentMethod;
    public String createdAt;
    public String updatedAt;

    public OrderInfo() {}
}

/** Address information - reused in billing and shipping. */
public static class Address {
    public String province;
    public String city;
    public String district;
    public String street;
    public String postalCode;
    public String recipientName;
    public String recipientPhone;

    public Address() {}
}

/** Order item. */
public static class OrderItem {
    public String productName;
    public String productId;
    public Integer quantity;
    public Double unitPrice;
    public Double subtotal;

    public OrderItem() {}
}
```

Use the class with ReActAgent:

```java
/** Example 5: Extract order information with nested structures. */
    private static void runOrderExtractionExample(ReActAgent agent) {
        String query =
                "Order ID: ORD20260409001, Customer: Zhang San (Phone: 13800138000, Email: zhangsan@example.com). "
                        + "Shipping Address: 88 Jianguo Road, Chaoyang District, Beijing, Postal Code 100022, Recipient: Li Si, Phone: 13900139000. "
                        + "Billing Address: 1 Zhongguancun Street, Haidian District, Beijing, Postal Code 100080, Recipient: Zhang San, Phone: 13800138000. "
[...]

Issue #1175, opened 2026-04-09
[Bug]:结构化输出，会调用generate_response工具，但是如果存在多层嵌套，工具会报错


59467:
下面是工具的返回消息：
	{
		"content":[
			{
				"id":"call_-7727915998354861090",
				"metadata":{},
				"name":"generate_response",
				"output":[
					{
						"text":"Error: Parameter validation failed for tool 'generate_response': Schema validation error: Reference /$defs/DocumentExtract cannot be resolved\nPlease correct the parameters and try again."
					}
				]
			}
		],
		"id":"c80ea208-9a88-4ac2-a940-0556ea8a3c9e",
		"metadata":{},
		"name":"指标提取智能体",
		"role":"TOOL",
		"timestamp":"2026-04-09 12:39:05.771"
	}

Alexxigang:
I traced this to the structured-output tool schema assembly in `StructuredOutputCapableAgent`. The generated `generate_response` tool wraps the nested schema under `response`, but it does not hoist `$defs` / `definitions` to the tool-schema root, so nested `$ref` pointers resolve against the wrong document root and validation fails.

I'm working on a fix that normalizes the wrapped schema before registration and will add regression coverage for nested class-based schemas, dynamic JSON schemas, and the root-level `$defs` behavior.

Alexxigang:
Opened PR #1205: https://github.com/agentscope-ai/agentscope-java/pull/1205

PR #1181, opened 2026-04-09
fix(agent): hoist structured output defs to root

Alexxigang:
## Summary
- hoist nested $defs / definitions to the root when wrapping structured-output schemas under the temporary generate_response tool
- keep the wrapped 
esponse schema focused on its object body so #//... references resolve correctly during tool parameter validation
- add regression tests for reused nested objects both at the schema-validation layer and through an actual ReActAgent structured-output flow

## Why this fix
Issue #1176 happens because StructuredOutputCapableAgent wraps the generated output schema under a 
esponse property, but leaves shared definitions nested inside that property. The generated $ref values still point to #//..., which means validation fails once the schema is wrapped and the definitions are no longer located at the document root.

This change mirrors the same def-hoisting strategy already used in tool parameter schema generation, so reusable nested objects continue to work after the structured-output schema is wrapped for generate_response.

## Validation
- mvn -pl agentscope-core spotless:apply
- mvn -pl agentscope-core '-Dtest=StructuredOutputSchemaWrappingTest' test
- mvn -pl agentscope-core '-Dtest=ReActAgentStructuredOutputTest,ToolkitTest,ToolValidatorTest,JsonSchemaUtilsTest' test

## Notes
- This also appears to address the same unresolved-reference symptom reported in #1175.

## AI Assistance
- Drafted with AI assistance and validated locally before submission.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/StructuredOutputCapableAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/StructuredOutputCapableAgent.java
@@ -215,11 +215,7 @@ public abstract class StructuredOutputCapableAgent extends AgentBase {
 
             @Override
             public Map<String, Object> getParameters() {
-                Map<String, Object> params = new HashMap<>();
-                params.put("type", "object");
-                params.put("properties", Map.of("response", schema));
-                params.put("required", List.of("response"));
-                return params;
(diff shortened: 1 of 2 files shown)


Alexxigang:

LearningGp:

PR #1183, opened 2026-04-09
fix(tool): make active tool groups concurrency-safe

### LINKED ISSUE #1165: Bug: ToolGroupManager.activeGroups is not thread-safe, causing race conditions in concurrent agent scenarios

Alexxigang:
## Summary
- store active tool-group names in a concurrency-safe snapshot set instead of a plain ArrayList
- keep ToolGroup.active aligned with setActiveGroups(...) replacements and concurrent updates
- add regression coverage for concurrent read/write access, snapshot replacement, and copyTo(...) isolation

## Why this fix
ToolGroupManager already uses concurrent maps for group and tool indexes, but ctiveGroups remained a mutable ArrayList. In concurrent agent scenarios, activation/deactivation and snapshot reads such as getActiveGroups() / getActivatedNotes() can race, and the list can diverge from each ToolGroup's active flag. This patch makes the active-group snapshot safe for concurrent access while preserving the existing public API.

## Validation
- mvn -pl agentscope-core spotless:apply
- mvn -pl agentscope-core '-Dtest=ToolGroupManagerTest,ToolGroupManagerConcurrencyTest,ToolkitTest,ToolGroupsE2ETest' test

Closes #1165.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolGroup.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolGroup.java
@@ -43,7 +43,7 @@ public class ToolGroup {
 
     private final String name;
     private final String description;
-    private boolean active;
+    private volatile boolean active;
     private final Set<String> tools; // Tool names in this group
 
     private ToolGroup(Builder builder) {
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolGroupManager.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolGroupManager.java
@@ -17,10 +17,12 @@ package io.agentscope.core.tool;
 
 import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArraySet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -33,7 +35,7 @@ class ToolGroupManager {
 
     private final Map<String, ToolGroup> toolGroups = new ConcurrentHashMap<>(); // group -> tools
     private final Map<String, Set<String>> tools = new ConcurrentHashMap<>(); // tool -> groups
-    private List<String> activeGroups = new ArrayList<>();
+    private final Set<String> activeGroups = new CopyOnWriteArraySet<>();
 
     /**
      * Create tool groups and record them in the manager.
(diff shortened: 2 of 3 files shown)


Alexxigang:

LearningGp:
Please run mvn spotless:apply before committing.

PR #1185, opened 2026-04-09
Fix issue 1158

xuhuafeifei:
## AgentScope-Java Version

### Background

  Some skill repositories (for example Nacos) encode the skill source with a colon, such as nacos:public. `AgentSkill#getSkillId()`
  is built as name + "_" + source, so the id can contain :. On Windows, : is not allowed in a normal path segment, so
  `SkillBox#uploadSkillFiles()` could fail to write resources when it used Path#resolve(skillId) as a directory name.

### Solution

 Keep logical identifiers (`getSource()`, `getSkillId()`) unchanged, while making filesystem paths used during skill file upload safe
  on Windows (and generally for reserved filename characters).

## Checklist

  • Add `SkillBox.unifyToSafeSkillId` to normalize characters that are invalid in file names (\ / : * ? " < > |) to _, with a small
     fallback for blank ids.
  • Use the normalized id only when constructing upload directory paths in`uploadSkillFiles()`.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
@@ -795,8 +795,7 @@ public class SkillBox implements StateModule {
         if (skillId == null || skillId.isBlank()) {
             return "skill";
         }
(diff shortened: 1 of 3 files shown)


LearningGp:
PTAL @fang-tech

PR #1191, opened 2026-04-11
fix(a2a-server,core-tool): harden host fallback and guard schema definition conflicts

RuleViz:
This PR fixes two reliability issues across A2A server setup and tool schema generation.

1. In A2A deployment properties, host resolution now falls back to `localhost` when local IP lookup throws a `SocketException`, and logs a warning instead of leaving host as `null`.
2. In tool parameter schema generation, `$defs` hoisting now detects conflicting definitions for the same key and throws an explicit exception instead of silently overriding previous entries.

The change includes regression tests for both cases:
- deployment host fallback on local IP resolution failure
- `$defs` merge behavior for both conflicting and equivalent definitions


LearningGp (approved):
LGTM

PR #1205, opened 2026-04-12
fix(agent): hoist structured output defs to tool root

Alexxigang:
## Summary
- hoist nested `$defs` and legacy `definitions` to the generated `generate_response` tool schema root
- reuse the same schema normalization path for structured output wrapping and regular tool schema generation
- add regression coverage for nested dynamic JSON schemas, validation behavior, and input-schema immutability

## Root cause
`StructuredOutputCapableAgent` wraps the response schema under a top-level `response` property when it registers the temporary `generate_response` tool. When the wrapped schema contains nested `$ref` values like `#/$defs/Material`, those references resolve from the tool schema document root, but the definitions were still nested under `properties.response`. That makes validation fail for multi-level structured output schemas.

## Validation
- `mvn -pl agentscope-core '-Dtest=JsonSchemaUtilsTest,StructuredOutputDynamicDefineTest,ToolValidatorTest' test -q`
- `mvn -pl agentscope-core '-Dtest=JsonSchemaUtilsTest,StructuredOutputDynamicDefineTest,ToolValidatorTest,ReActAgentStructuredOutputTest,StructuredOutputHookTest' test -q`


LearningGp:

PR #1206, opened 2026-04-12
fix(agent): sync tool group updates for existing agents

Alexxigang:
## Summary
- sync external `Toolkit.updateToolGroups(...)` state into already-built `ReActAgent` instances at the start of each new invocation
- keep the per-agent copied toolkit isolation while propagating group activation and deactivation for subsequent calls
- add regression tests covering repeated calls, streaming calls, and copied-toolkit behavior

## Root cause
`ReActAgent.Builder.build()` deep-copies the provided `Toolkit` to avoid state interference between agents. After that, later calls to `toolkit.updateToolGroups(...)` mutate only the caller-owned toolkit, while the existing agent keeps using its stale internal toolkit snapshot. As a result, newly activated or deactivated tool groups are not reflected for later invocations on the same agent instance.

## Fix
Before a fresh invocation begins, the agent now diffs the source toolkit's active groups against its internal copied toolkit and applies the required group activations/deactivations. This keeps the existing per-call toolkit isolation and avoids resetting tool-group changes during an in-flight invocation.

## Validation
- `mvn -pl agentscope-core '-Dtest=ReActAgentToolGroupSyncTest' test -q`
- `mvn -pl agentscope-core '-Dtest=ReActAgentTest,ReActAgentToolGroupSyncTest,ToolkitTest,ToolGroupManagerTest' test -q`


jujn:
Hi. This PR fixes the current issue very well. However, the synchronization status is also limited to the activation and deactivation of the workgroup. For other internal states of Toolkit, such as tool list, MCP client, etc., the same issue still exists.

LearningGp:

PR #1215, opened 2026-04-14
fix(core): AgentScopeJvmShutdownHook conflict with spring (#1093)

### LINKED ISSUE #1093: [Bug]:1.0.11 AgentScopeJvmShutdownHook conflict with spring
spring工程的SIGTERM信号响应是由spring的ShutdownHook处理的，agentscope又注册了一个AgentScopeJvmShutdownHook。这两个hook之间是并行的，会造成spring容器没有完全关闭时会正常接受请求，但是agent已经关闭了，调用会报错返回Operation interrupted due to system shutting down, please retry。
按照我的理解，在spring工程中agentscope不需要这种hook，生命周期是由spring统一管理。
之前dubbo也有这种问题，提供了一个ShutdownHook和spring的ShutdownHook并发冲突。

Wzy19930507:
## AgentScope-Java Version

## Description

* Closes: #1093
* Fix AgentScopeJvmShutdownHook conflict with spring

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentScopeJvmShutdownHook.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentScopeJvmShutdownHook.java
@@ -40,7 +40,20 @@ public final class AgentScopeJvmShutdownHook {
 
     private AgentScopeJvmShutdownHook() {}
 
+    /**
+     * Resets the REGISTERED flag for testing purposes.
+     *
+     * <p>This method is intended for testing only. It allows tests to verify
+     * the JVM hook registration behavior in isolation.
+     */
+    static void resetForTesting() {
+        REGISTERED.set(false);
+    }
+
     public static void register(GracefulShutdownManager manager) {
+        if (!manager.getConfig().isRegister()) {
+            return;
+        }
         if (!REGISTERED.compareAndSet(false, true)) {
             return;
         }
--- a/agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentScopeJvmShutdownHook.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentScopeJvmShutdownHook.java
@@ -18,6 +18,7 @@ package io.agentscope.core.shutdown;
 import io.agentscope.core.model.transport.HttpTransportFactory;
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -35,6 +36,7 @@ public final class AgentScopeJvmShutdownHook {
 
     private static final Logger log = LoggerFactory.getLogger(AgentScopeJvmShutdownHook.class);
     private static final AtomicBoolean REGISTERED = new AtomicBoolean(false);
+    private static final AtomicReference<Thread> SHUTDOWN_HOOK_THREAD = new AtomicReference<>();
 
     private static final Duration INTERRUPT_GRACE_PERIOD = Duration.ofSeconds(5);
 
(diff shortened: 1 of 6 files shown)


uuuyuqi:
I recommend disabling automatic registration only in Spring-based projects. In non-Spring projects, AgentScopeJvmShutdownHook is necessary.

agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentScopeJvmShutdownHook.java
```diff
@@ -40,7 +40,20 @@ public final class AgentScopeJvmShutdownHook {
...
+     *
+     * <p>This method is intended for testing only. It allows tests to verify
+     * the JVM hook registration behavior in isolation.
+     */
+    static void resetForTesting() {
+        REGISTERED.set(false);
+    }
+
     public static void register(GracefulShutdownManager manager) {
+        if (!manager.getConfig().isRegister()) {
```
uuuyuqi: In projects not built with Spring, registration is still required, which may necessitate a detection mechanism.

agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentScopeJvmShutdownHook.java
```diff
@@ -40,7 +40,20 @@ public final class AgentScopeJvmShutdownHook {
...
     private AgentScopeJvmShutdownHook() {}
 
+    /**
+     * Resets the REGISTERED flag for testing purposes.
+     *
+     * <p>This method is intended for testing only. It allows tests to verify
+     * the JVM hook registration behavior in isolation.
+     */
+    static void resetForTesting() {
+        REGISTERED.set(false);
```
uuuyuqi: Removing Shutdown Hook may be necessary.

agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownConfig.java
```diff
@@ -25,16 +25,20 @@
...
  *   <li>Policy for handling partial reasoning results</li>
+ *   <li>Policy for control register jvm shutdown hook</li>
  * </ul>
  *
  * @param shutdownTimeout       maximum duration to wait for shutdown completion;
  *                              if null, the system will wait indefinitely for all operations
  *                              to complete; if specified, must be a positive duration
  * @param partialReasoningPolicy policy for handling incomplete reasoning results during shutdown;
  *                              cannot be null
+ * @param isRegister Policy for control register jvm shutdown hook
```
uuuyuqi: There seems to be a grammatical error here. Perhaps you could try using AI to review it yourself first.

agentscope-examples/graceful-shutdown/src/main/java/io/agentscope/examples/shutdown/e2e/AgentService.java
```diff
@@ -89,6 +89,9 @@ public class AgentService {
     @Value("${agent.shutdown-partial-reasoning-policy:save}")
     private String partialReasoningPolicy;
 
+    @Value("${agent.shutdown-hook-is-register:true}")
+    private boolean isRegister;
```
uuuyuqi: The variable and method name `isRegister` sound a bit odd. `needRegister` or `registerJvmShutdownHook` might be more appropriate.

uuuyuqi (changes requested):

agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentScopeJvmShutdownHook.java
Wzy19930507: Thanks! I plan to add a new API `AgentScopeJvmShutdownHook.unregister()` and remove `resetForTesting`.

agentscope-core/src/main/java/io/agentscope/core/shutdown/AgentScopeJvmShutdownHook.java
Wzy19930507: @uuuyuqi  Hello, I need some help.

Before setConfig is called, GracefulShutdownManager has already been initialized. At that point, needRegister uses its default value true, so the JVM shutdown hook has already been registered. Therefore, setting needRegister=false later cannot prevent that registration.

There are two possible directions:

### Option A: Change the configuration semantics
Change the option from “whether to register the JVM shutdown hook” to “whether to enable the graceful shutdown logic inside the JVM shutdown hook”. The hook is still registered, but if the option is false, it returns immediately when executed.

``` 
AgentScopeJvmShutdownHook {
    public static void register(GracefulShutdownManager manager) {
                new Thread(
                        () -> {
                            // enable the graceful shutdown 
                            if (!manager.getConfig().enableShutdownHook()) {
                                return;
                            }
                            try {
                                manager.performGracefulShutdown();
```

### Option B: Keep the registration semantics
Keep the meaning as “whether to register the JVM shutdown hook” and add unregister support. When setConfig receives needRegister=false, it should remove the already registered hook; when it receives needRegister=true, it should ensure the hook is registered.

---

I prefer Option A. It involves fewer changes and brings more stable behavior. What do you think?

Wzy19930507: The detection mechanism sounds good, but implementing it in `agentscope-core` is not suitable.

How about we handle the adaptation of GracefulShutdownConfig in the `agentscope-spring-boot-starter` module via a separate PR?

PR #1221, opened 2026-04-14
feat(tool): expose MCP output schemas in tool definitions

Alexxigang:
## Summary
- expose MCP `outputSchema` through `AgentTool`, `McpTool`, and `ToolSchema`
- preserve output schemas when registering MCP tools via `McpClientManager`
- add regression coverage for schema propagation and toolkit exposure

## Why this fix
`McpSchema.Tool` already provides `outputSchema()`, but the current registration flow drops it before the tool reaches AgentScope's tool schema layer. This prevents downstream consumers from seeing structured output contracts for MCP-backed tools.

## Validation
- `mvn -pl agentscope-core -Dtest=McpClientManagerTest,ExternalToolSupportTest,ToolSchemaProviderTest,McpToolTest,ToolkitTest test -q`

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/McpClientManager.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/McpClientManager.java
@@ -178,7 +178,12 @@ class McpClientManager {
                                                     toolPresetParams != null
                                                             ? toolPresetParams.keySet()
                                                             : Collections.emptySet()),
-                                            mcpClientWrapper);
+                                            mcpTool.outputSchema() != null
+                                                    ? new ConcurrentHashMap<>(
+                                                            mcpTool.outputSchema())
+                                                    : null,
+                                            mcpClientWrapper,
+                                            toolPresetParams);
 
                             // Register with group, MCP client name, and preset parameters via
                             // callback
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/McpClientManager.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/McpClientManager.java
@@ -183,7 +183,7 @@ class McpClientManager {
                                                             mcpTool.outputSchema())
                                                     : null,
                                             mcpClientWrapper,
-                                            toolPresetParams);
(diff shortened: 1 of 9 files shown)


agentscope-core/src/main/java/io/agentscope/core/tool/McpClientManager.java
```diff
@@ -178,7 +178,12 @@ Mono<Void> registerMcpClient(
                                                     toolPresetParams != null
                                                             ? toolPresetParams.keySet()
                                                             : Collections.emptySet()),
-                                            mcpClientWrapper);
+                                            mcpTool.outputSchema() != null
+                                                    ? new ConcurrentHashMap<>(
+                                                            mcpTool.outputSchema())
+                                                    : null,
+                                            mcpClientWrapper,
+                                            toolPresetParams);
```
LearningGp: When refactoring the McpClientManager constructor from 4 to 6 parameters, passing toolPresetParams into McpTool creates redundancy. These params now exist in both McpTool.presetArguments and RegisteredToolFunction.presetParameters.

This leads to double application in both ToolExecutor and McpTool.mergeArguments(). Furthermore, the merge priorities are inconsistent: one path has presets overriding inputs, while the other has inputs overriding presets. This breaks the "single source of truth" for the preset parameter injection path.

Recommendation:
Pass null as the last argument in the 6-param constructor. This ensures preset params continue to be injected solely through the RegisteredToolFunction → ToolExecutor path.

LearningGp (changes requested):

Alexxigang:
Addressed the review point in `McpClientManager`: preset params are now kept as a single source of truth through `RegisteredToolFunction -> ToolExecutor`, while MCP tool registration still uses them only for schema filtering and callback propagation.

Re-ran targeted validation:
- `mvn -pl agentscope-core '-Dtest=McpClientManagerTest,ToolkitTest,ToolSchemaProviderTest,McpToolTest' test`

LearningGp (approved):
LGTM

PR #1223, opened 2026-04-14
fix(model): normalize malformed streaming tool call fragments

### LINKED ISSUE #1216: [Bug]:执行工具调用是，解析的参数发生异常（When executing the tool call, an exception occurred in the parsed parameters）
**Describe the [bug**]
执行工具调用是，解析的参数发生异常（When executing the tool call, an exception occurred in the parsed parameters）
详细参考附件

**To Reproduce**
通过Langfuse 看到的信息,下面的一次tool_call因参数解析出问题，变成了两次tool_call，最终导致调用失败
（Based on the information obtained through Langfuse, the following issue occurred: due to a problem with parameter parsing, one tool_call was mistakenly executed twice, ultimately resulting in a failed call）
----------------------------trace info---------------------------
...
}
2: {4 Items
type: "tool_call"
id: "call_28281576ef7241a397970399"
name: "retrieveFromMemory"
arguments: "{"keywords": ["关注", "重点", "部类", "品牌", "偏好", "负责"]"
}
3: {4 Items
type: "tool_call"
id: "streaming_1776130606735"
name: "retrieveFromMemory"
arguments: "}"
}
]
finish_reason: "tool_calls"

----------------------------trace info---------------------------
**Expected behavior**

**Error messages**
Error: Parameter validation failed for tool 'retrieveFromMemory': Schema validation error: Invalid input
Please correct the parameters and try again.

- AgentScope-Java Version: [e.g. 1.0.11, check your pom.xml dependency version or run `mvn dependency:tree | grep agentscope-parent:pom`(only mac/linux)]
- Java Version: [e.g. 17]
- OS: [e.g. macos, windows]

**Additional context**
Add any other context about the problem [here.](url)

Alexxigang:
## Summary
- normalize malformed OpenAI-compatible streaming tool-call fragments with a non-null tool name but no tool-call id
- prevent trailing argument chunks such as `}` from becoming a new synthetic `streaming_` tool call
- add parser, formatter, and accumulation regression coverage for the malformed chunk shape reported in #1216

## Why this fix
Some OpenAI-compatible APIs can return the final argument fragment with `name` set even though `id` is null. The previous parser treated any non-empty name as a new tool call, generated a synthetic `streaming_` id, and prevented `ToolCallsAccumulator` from merging the closing fragment back into the original call. The parser now recognizes that malformed named chunk as a fragment so the original call keeps complete JSON arguments.

## Validation
- `mvn -pl agentscope-core -Dtest=OpenAIChatFormatterTest,OpenAIStreamingToolCallTest,OpenAIResponseParserTest,ToolCallsAccumulatorTest,OpenAIChatModelTest test -q`

Fixes #1216

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIResponseParser.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIResponseParser.java
@@ -52,6 +52,21 @@ public class OpenAIResponseParser {
     protected static final String FRAGMENT_PLACEHOLDER = "__fragment__";
 
     /**
+     * Some OpenAI-compatible providers occasionally emit a malformed trailing argument fragment
+     * with a non-null tool name but no tool call id. That chunk should still be treated as a
+     * fragment so it can merge back into the previously started tool call instead of becoming a
+     * new synthetic call.
+     */
+    private boolean isMalformedNamedStreamingFragment(
+            String toolCallId, String toolName, String arguments) {
+        return toolCallId == null
+                && toolName != null
+                && !toolName.isEmpty()
+                && arguments != null
+                && !arguments.isEmpty();
+    }
+
+    /**
      * Safely get prompt token count from usage, returning 0 if null or invalid.
      *
      * @param usage the OpenAI usage object (may be null)
@@ -461,14 +476,19 @@ public class OpenAIResponseParser {
                                         }
                                     }
 
-                                    if (toolCallId == null) {
-                                        toolCallId = "streaming_" + System.currentTimeMillis();
-                                    }
                                     if (toolName == null) {
                                         toolName = "";
                                     }
                                     if (arguments == null) {
                                         arguments = "";
+                                    }
+
+                                    boolean malformedNamedFragment =
+                                            isMalformedNamedStreamingFragment(
+                                                    toolCallId, toolName, arguments);
+
+                                    if (!malformedNamedFragment && toolCallId == null) {
+                                        toolCallId = "streaming_" + System.currentTimeMillis();
                                     }
 
                                     log.debug(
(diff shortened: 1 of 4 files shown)


Alexxigang:

Alexxigang:
Added targeted streaming edge-case coverage on top of this PR for missing-id first chunks, malformed complete JSON chunks, and thought-signature-only fragments.

Validation:
- `mvn -pl agentscope-core '-Dtest=OpenAIResponseParserTest,OpenAIStreamingToolCallTest,OpenAIChatFormatterTest' test`

LearningGp:
The index field in the OpenAI streaming protocol is specifically designed to correlate multiple chunks of the same tool call. Wouldn't it be more appropriate to leverage this mechanism to address the issue?

Alexxigang:
Friendly ping when maintainers have time. The branch is still ready on my side, including the added regression coverage, and I'm happy to make any follow-up adjustments if review feedback comes in. Thanks!

Alexxigang:
Checked this against the current implementation before changing anything: `index` is already part of the correlation path in `ToolCallsAccumulator.determineKey()` as the last fallback after `id`, non-placeholder `name`, and `lastToolCallKey`.

The cases this PR is targeting are the ones where correlation alone still leaves us with a broken block:
- first/only chunk has no stable `id`, so downstream still needs a generated tool-call id
- the final assembled arguments string is malformed JSON, so we need to preserve raw content instead of dropping it on parse failure
- Gemini-style thought-signature-only fragments add metadata without advancing the normal name/args fields

So I agree `index` is the right mechanism for chunk grouping, and the current accumulator already uses it. The remaining fix here is about making the assembled `ToolUseBlock` still usable when the provider sends incomplete or malformed fragments even after grouping.

Alexxigang:
Friendly follow-up on this one as well: the branch is still ready on my side with the regression coverage included, and I can adjust the implementation if you'd prefer a different direction around the streaming tool-call assembly path.

LearningGp:
PTAL @chickenlj

PR #1236, opened 2026-04-17
Fix #1102: place cache_control at content-item level for multimodal messages

### LINKED ISSUE #1102: [Bug]:cacheControl 设置的位置不对吧？
看文档里cacheControl是设置在message的content里，但是agentscope吧cacheControl放到了message里
Image

Image

wade-microblueworld:
### Problem

When using prompt caching (`cacheControl: true`) with multimodal messages (content is an array), the `cache_control` field was being set at the **message level**, but the OpenAI API specification requires `cache_control` to be set on individual **content array items**, not on the message object itself.

**Before (incorrect):**
```json
{
  "role": "user",
  "content": [
    {"type": "image_url", "image_url": {"url": "..."}},
    {"type": "text", "text": "What is this?"}
  ],
  "cache_control": {"type": "ephemeral"}  // ← wrong location
}
```

**After (correct):**
```json
{
  "role": "user",
  "content": [
    {"type": "image_url", "image_url": {"url": "..."}},
    {"type": "text", "text": "What is this?", "cache_control": {"type": "ephemeral"}}  // ← correct
  ]
}
```

### Changes

- **`OpenAIBaseFormatter.applyCacheControl()`**: For multimodal messages, find the last text content item and set `cache_control` on it. Fall back to message level if no text part exists.
- **`OpenAIMessageConverter.applyCacheControlFromMetadata()`**: Same logic applied for metadata-driven cache control marking.
- **`OpenAIContentPart`**: Added `cacheControl` field with `@JsonProperty("cache_control")`.

### Testing

Existing tests pass (text-only behavior unchanged). New test coverage for multimodal case was considered but the fix preserves backward compatibility for the text-only path.

Fixes #1102

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIBaseFormatter.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIBaseFormatter.java
@@ -16,6 +16,7 @@
 package io.agentscope.core.formatter.openai;
 
 import io.agentscope.core.formatter.AbstractBaseFormatter;
+import io.agentscope.core.formatter.openai.dto.OpenAIContentPart;
 import io.agentscope.core.formatter.openai.dto.OpenAIMessage;
 import io.agentscope.core.formatter.openai.dto.OpenAIRequest;
 import io.agentscope.core.formatter.openai.dto.OpenAIResponse;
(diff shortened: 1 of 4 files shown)


agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIBaseFormatter.java
```diff
@@ -183,13 +187,41 @@ public void applyCacheControl(List<OpenAIMessage> messages) {
...
+
+    /**
+     * Apply cache control to a single message.
+     *
+     * <p>For multimodal messages (content is an array), cache_control is added to the last text
+     * content item. For text-only messages, it is added at the message level.
+     *
+     * @param msg the message to apply cache control to
+     */
+    private void applyCacheControlToMessage(OpenAIMessage msg) {
```
LearningGp, line 209: applyCacheControl triggers redundant calls to applyCacheControlToMessage when a multimodal system message is the last in the list, causing duplicate cache_control at both the content-part and message levels.

agentscope-core/src/test/java/io/agentscope/core/formatter/openai/OpenAICacheControlTest.java
```diff
@@ -238,4 +242,162 @@ void systemMessageMetadata() {
...
+
+            List<OpenAIMessage> result = formatter.format(List.of(msg));
+
+            // Falls back to message level
+            assertEquals(EPHEMERAL, result.get(0).getCacheControl());
+        }
+
+        @Test
+        @DisplayName("should not set cache_control on multimodal when metadata flag is absent")
+        void multimodalNoMetadata() {
```
LearningGp, line 386: The applyCacheControlFromMetadata code path is never exercised because msg is manually constructed and doesn't go through formatter.format().

LearningGp (changes requested):

PR #1242, opened 2026-04-18
fix(mem0): fix memory search returning no results (#1224)

lzq986:
related issue 
#1224 

Fix the bug where Mem0 search requests fail to retrieve stored memories due to incorrect filter structure.

Root cause:
- The buildNestedFilters() method in Mem0SearchRequest generates an incorrect filter format when both entity filters (userId/agentId/runId) and metadata filters are present
- The original code incorrectly placed all conditions in a single OR level instead of the proper nested AND/OR structure required by Mem0 v2 API

Changes:
- Refactor buildNestedFilters() to correctly build the nested filter structure: {AND: [{OR: entityFilters}, {OR: metadataFilters}]}
- Add logging for debugging search failures
- Add unit tests for various filter combination scenarios

Affected files:
- Mem0SearchRequest
- Mem0LongTermMemory

## AgentScope-Java Version

[The version of AgentScope-Java you are working on, e.g. 1.0.11, check your pom.xml dependency version or run `mvn dependency:tree | grep agentscope-parent:pom`(only mac/linux)]

## Description

## Checklist

### DIFF
--- a/agentscope-extensions/agentscope-extensions-mem0/src/main/java/io/agentscope/core/memory/mem0/Mem0SearchRequest.java
+++ b/agentscope-extensions/agentscope-extensions-mem0/src/main/java/io/agentscope/core/memory/mem0/Mem0SearchRequest.java
@@ -17,6 +17,7 @@ package io.agentscope.core.memory.mem0;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -28,10 +29,35 @@ import java.util.Map;
  * to retrieve relevant memories based on a query string. The search uses semantic
  * similarity to find the most relevant memories.
  *
- * <p>The v2 API uses a filters object to specify search criteria (user_id, agent_id,
- * run_id, app_id, etc.), enabling proper memory isolation in multi-tenant scenarios.
- * The filters field is required by the API and will be sent as an empty object if no
- * filters are specified.
+ * <p>The v2 API requires filters to be structured with logical operators (AND, OR, NOT):
+ * <ul>
+ *   <li>Entity filters (user_id, agent_id, app_id, run_id) must be OR-connected
+ *       because a single memory cannot have multiple entity values</li>
+ *   <li>Metadata filters are OR-connected for multiple key-value pairs</li>
+ *   <li>Entity filters and metadata filters are AND-connected at the top level</li>
+ * </ul>
+ *
+ * <p>Example of correct filter structure:
+ * <pre>{@code
+ * {
+ *   "filters": {
+ *     "AND": [
+ *       {
+ *         "OR": [
+ *           { "user_id": "user123" },
+ *           { "agent_id": "agentA" }
+ *         ]
+ *       },
+ *       {
+ *         "OR": [
+ *           { "metadata": { "category": "travel" } },
+ *           { "metadata": { "priority": "high" } }
+ *         ]
+ *       }
+ *     ]
+ *   }
+ * }
+ * }</pre>
  */
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class Mem0SearchRequest {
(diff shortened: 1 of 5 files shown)


agentscope-extensions/agentscope-extensions-mem0/src/main/java/io/agentscope/core/memory/mem0/Mem0SearchRequest.java
```diff
@@ -355,11 +316,79 @@ public Mem0SearchRequest build() {
...
+            if (!entityFilters.isEmpty()) {
+                Map<String, Object> entityOr = new HashMap<>();
+                entityOr.put("OR", entityFilters);
+                andConditions.add(entityOr);
+            }
+
+            List<Map<String, Object>> metadataFilters = buildMetadataFilters();
+            if (!metadataFilters.isEmpty()) {
+                Map<String, Object> metadataOr = new HashMap<>();
+                metadataOr.put("OR", metadataFilters);
```
LearningGp, line 338: Suggest adding a config option for AND/OR logic in metadata filtering, as the current OR default might return too many matches.

agentscope-extensions/agentscope-extensions-mem0/src/main/java/io/agentscope/core/memory/mem0/Mem0SearchRequest.java
```diff
@@ -355,11 +316,79 @@ public Mem0SearchRequest build() {
...
+
+            request.setFilters(buildNestedFilters());
+
             return request;
         }
+
+        private Map<String, Object> buildNestedFilters() {
+            List<Map<String, Object>> andConditions = new ArrayList<>();
+
+            List<Map<String, Object>> entityFilters = buildEntityFilters();
```
LearningGp, line 328: OR logic in Entity filters risks cross-user memory leakage.

LearningGp (changes requested):
In the Mem0 v2 API, user_id, agent_id, and run_id are handled as mutually exclusive dimensions. This makes AND logic useless (always empty) and OR logic too permissive. We should consider enforcing user_id filtering by default or using metadata for more granular combined queries.

agentscope-extensions/agentscope-extensions-mem0/src/main/java/io/agentscope/core/memory/mem0/Mem0SearchRequest.java
lzq986: Thanks.You are very thoughtful.I will improve these  codes. But i still have some questions.
As mentioned above,in the Mem0 v2 API, user_id, agent_id, and run_id are handled as mutually exclusive dimensions.Therefore, If the user passes these mutually exclusive parameters, i think there may be two ways  to handle it as follows.

**1.** 
---

### Solution: Configurable Entity Filter Strategy

add a new `EntityFilterStrategy` enum with three options:

| Strategy | Behavior | Security |
|----------|----------|----------|
| `USER_ID_ONLY` (default) | Only use `user_id` for filtering, ignore other entity fields | ✅ Safest |
| `OR` | Combine entity filters with OR logic | ⚠️ May leak cross-user data |
| `AND` | Combine entity filters with AND logic | ❌ Usually returns empty results |

### Example Usage

```java
// Default (secure): Only filter by user_id
memory = builder()
    .userId("user123")
    .agentName("agentA")
    .build();

// Explicit OR (use with caution - may leak cross-user data)
memory = builder()
    .userId("user123")
    .agentName("agentA")
    .entityFilterStrategy(EntityFilterStrategy.OR)
    .build();
```

### Backward Compatibility

- Default is now `USER_ID_ONLY` (more secure)
- Users can explicitly opt-in to OR or AND behavior if needed

---
**Is this handling method what you want ?**

I have another way to handle this.
**2.**
---
Approach: If there is a user_id, use it; if not, use agent_id; if neither is available, use run_id
// only  userId
{ "filters": { "user_id": "user123" } }

// only  agentId
{ "filters": { "agent_id": "agentA" } }

---

LearningGp:
PTAL @shiyiyue1102

jujn:
Assign to @amyaxy (Review)

PR #1256, opened 2026-04-19
fix: sanitize null values in tool arguments to prevent empty field errors

Jim1874:
## Summary

Fix tool parameter conversion bug where passing an object with empty fields causes errors. Null values are now sanitized to empty strings or empty arrays before being passed to tool arguments.

## Changes

- Added null value sanitization in `DashScopeToolsHelper.java` and `OpenAIMessageConverter.java`
- Added `JacksonJsonCodec.java` with utility for safe null-to-empty conversion
- Ensures all object fields are properly initialized before tool call execution


LearningGp:
image

PR #1259, opened 2026-04-20
feat(subagent): support passing user-defined fields in SubAgentTool

### LINKED ISSUE #485: [Feature]: Support passing user-defined fields in `SubAgentTool`
**Problem Description**
Currently, `SubAgentTool` is implemented solely based on conversation history. Consequently, users cannot pass external parameters (e.g., `userId`) to the sub-agent **securely** or **accurately**.

**Proposed Solution**
We need to design a mechanism that allows users to define and pass specific fields to the `SubAgentTool` explicitly.

jujn:
## Description

Close #485 
This PR introduces a mechanism to pass user-defined external parameters (e.g., `userId`, `tenantId`) to sub-agents securely, addressing the issue where parameters could only be passed via the pure text message.

## Checklist


LearningGp:
image

agentscope-core/src/main/java/io/agentscope/core/tool/subagent/SubAgentTool.java
```diff
@@ -190,6 +249,39 @@ private Mono<ToolResultBlock> executeConversation(ToolCallParam param) {
                 });
     }
 
+    /**
+     * Extracts a value from the ToolExecutionContext by trying standard JSON types.
```
LearningGp: Internally, ContextStore indexes by runtime Class<?>, but extractFromContext hardcodes checks for only seven types. When users register a system context using any other type, it returns null.For custom parameters (addParameter), the subsequent fallback branch input.containsKey(paramName) allows the LLM-provided value to take effect, which directly violates the documentation's security guarantee that "system context takes absolute precedence". Furthermore, for pure system parameters, the values are silently dropped (leaving only a warn log), leading to functional bugs.

LearningGp (changes requested):

agentscope-core/src/main/java/io/agentscope/core/tool/subagent/SubAgentTool.java
jujn: OK. I have fix it.

LearningGp (approved):
LGTM

PR #1283, opened 2026-04-25
fix(formatter): ensure assistant messages with tool_calls always include content field

### LINKED ISSUE #1268: [Bug]:OpenAIMessageConverter produces assistant messages with null content when using vLLM + Qwen3 thinking mode
# [Bug] OpenAIMessageConverter produces assistant messages with null content when using vLLM + Qwen3 thinking mode

## Environment

- AgentScope Java: 1.0.10
- Model: Qwen3.5-122B-A10B (via vLLM, OpenAI-compatible API)
- Model Client: `OpenAIChatModel`
- Agent: `ReActAgent` with `enablePlan()` and tool calling

## Description

When using `OpenAIChatModel` with a vLLM-hosted Qwen3 model that has thinking/reasoning enabled, the API returns `400 Bad Request` with:

```
Input error. Field required: input.messages.N.content
```

This happens because `OpenAIMessageConverter.convertAssistantMessage()` does not set the `content` field when the assistant message only contains `ThinkingBlock` + `ToolUseBlock` (no `TextBlock`).

## Root Cause

In `OpenAIMessageConverter.java` (line ~267):

```java
private OpenAIMessage convertAssistantMessage(Msg msg) {
    OpenAIMessage.Builder builder = OpenAIMessage.builder().role("assistant");

    String textContent = textExtractor.apply(msg);
    if (textContent != null && !textContent.isEmpty()) {
        builder.content(textContent);  // Only set when non-empty
    }
    // ... handle ThinkingBlock, ToolUseBlock ...
    return builder.build();
}
```

When the model responds with reasoning (ThinkingBlock) + tool calls (ToolUseBlock) but no text content, `textContent` is null, so `content` is never set on the builder. The resulting JSON message has no `content` field at all.

The OpenAI official API tolerates `content: null` for assistant messages with `tool_calls`, but vLLM's OpenAI-compatible endpoint requires `content` to be present (even as an empty string `""`).

## Steps to Reproduce

1. Deploy a Qwen3/Qwen3.5 model via vLLM with `--reasoning-parser qwen3`
2. Configure `OpenAIChatModel` pointing to the vLLM endpoint
3. Create a `ReActAgent` with tools (e.g., `enablePlan()`)
4. Send a message that triggers tool calling
5. After the first tool call + result round, the next LLM call fails with `Field required: input.messages.N.content`

## Actual Behavior

```
io.agentscope.core.model.exception.BadRequestException: 
OpenAI API error in streaming response: Input error. Field required: input.messages.6.content
```

The error index varies (messages.2, messages.6, messages.22) depending on how many rounds of tool calling occurred before the failure.

## Expected Behavior

The `convertAssistantMessage` method should ensure `content` is always set, at minimum as an empty string `""`, especially when `tool_calls` are present.

## Suggested Fix

```java
private OpenAIMessage convertAssistantMessage(Msg msg) {
    OpenAIMessage.Builder builder = OpenAIMessage.builder().role("assistant");

    String textContent = textExtractor.apply(msg);
    if (textContent != null && !textContent.isEmpty()) {
        builder.content(textContent);
    } else {
        // Ensure content is never null for vLLM/Qwen3 compatibility
        builder.content("");
    }
    // ... rest unchanged ...
}
```

## Additional Context

- This is a known issue with vLLM's Qwen3 reasoning parser. See: [vLLM Forum: Qwen3.5 only output reasoning and no content](https://discuss.vllm.ai/t/qwen3-5-only-output-reasoning-and-no-content/2523) — vLLM may return `content=None` with all output in the `reasoning` field for Qwen3 series models.
- The same `ReActAgent` configuration works correctly with `Qwen3.5-122B-A10B-Thinking` (the `-Thinking` variant), because that model's output format is correctly parsed by vLLM, producing non-null `content`.
- Even when vLLM correctly separates reasoning/content, the assistant message after a tool call may legitimately have empty text content (only ThinkingBlock + ToolUseBlock), which still triggers this bug.

## Workaround

Use `Qwen3.5-*-Thinking` model variants which produce correctly formatted output with vLLM's reasoning parser.

uuuyuqi:
## Description

When Qwen3 and similar models operate in thinking mode, assistant messages may contain `reasoning_content` + `tool_calls` but `null` content. Due to `@JsonInclude(NON_NULL)` on `OpenAIMessage` and `DashScopeMessage`, the `null` content field was entirely omitted from the JSON request body, causing strict OpenAI-compatible APIs (e.g. vLLM) to reject the request with:

```
Input error. Field required: input.messages.N.content
```

### Root Cause

In both `OpenAIMessageConverter` and `DashScopeMessageConverter`, when an assistant message has `tool_calls` but no `TextBlock`, the `content` field was left as `null` on the DTO. Jackson's `@JsonInclude(NON_NULL)` then omitted the field entirely during serialization.

### Fix

Explicitly set `content` to an empty string `""` (instead of leaving it `null`) when an assistant message has `tool_calls` but no text content. This ensures the `content` field is always present in the serialized JSON.

**Changed files:**
- `OpenAIMessageConverter.convertAssistantMessage()` — set `content=""` when `hasTextContent` is false and tool calls exist
- `DashScopeMessageConverter.convertToSimpleContent()` — replace `null` content with `""` for assistant messages with tool calls

Close #1268

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/dashscope/DashScopeMessageConverter.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/dashscope/DashScopeMessageConverter.java
@@ -217,11 +217,10 @@ public class DashScopeMessageConverter {
                 // Assistant with tool calls
                 builder.toolCalls(toolsHelper.convertToolCalls(toolBlocks));
                 String textContent = extractTextContent(msg);
-                if (textContent.isEmpty()) {
-                    builder.content((String) null);
-                } else {
-                    builder.content(textContent);
-                }
+                // Qwen3 and similar models in thinking mode may produce assistant
+                // messages with reasoning_content + tool_calls but null content.
+                // DashScope API requires the content field to be present.
+                builder.content(textContent.isEmpty() ? "" : textContent);
             } else {
                 builder.content(extractTextContent(msg));
             }
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/dashscope/DashScopeMultiAgentFormatter.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/dashscope/DashScopeMultiAgentFormatter.java
@@ -302,9 +302,8 @@ public class DashScopeMultiAgentFormatter
         List<ToolUseBlock> toolBlocks = msg.getContentBlocks(ToolUseBlock.class);
         if (!toolBlocks.isEmpty()) {
             builder.toolCalls(toolsHelper.convertToolCalls(toolBlocks));
-            // Set content to null if empty when tool calls exist (Python behavior)
             String textContent = extractTextContent(msg);
-            builder.content(textContent.isEmpty() ? null : textContent);
+            builder.content(textContent.isEmpty() ? "" : textContent);
         } else {
             builder.content(extractTextContent(msg));
         }
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIMessageConverter.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIMessageConverter.java
@@ -265,7 +265,8 @@ public class OpenAIMessageConverter {
         OpenAIMessage.Builder builder = OpenAIMessage.builder().role("assistant");
 
         String textContent = textExtractor.apply(msg);
-        if (textContent != null && !textContent.isEmpty()) {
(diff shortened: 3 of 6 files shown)


jujn:
Hi, it seems that issue #1270 should also be associated.

uuuyuqi:
> Hi, it seems that issue #1270 should also be associated.

Ah, I haven't searched through the issues yet. Thanks for pointing this out!

LearningGp (approved):
LGTM

PR #1317, opened 2026-04-29
  fix(autocontext): add error handling to AutoContextHook.handlePreReasoning()      

xiaoqingwanga:
## AgentScope-Java Version

1.0.12-SNAPSHOT

### Background

`AutoContextHook.handlePreReasoning()` does not have error handling. When the blocking model call inside `AutoContextMemory.compressIfNeeded()` (`model.stream().block()`) fails due to timeout, service errors, or rate limiting, the exception propagates without being handled, causing the Agent’s reactive pipeline to silently crash. The only trace is an `AGENT_CALL_END` event in tracing with `"exception":"value"`, preceded by an approximately 10-second blank period, which corresponds to the model call timeout window.

By contrast, `handlePreCall()` already has a `try-catch(Exception)` block and continues execution after errors. `handlePreReasoning()` is missing the same protection.

### Changes

**`AutoContextHook.java`**

* Wrapped the entire body of `handlePreReasoning()` in a try-catch block, aligning it with the existing pattern in `handlePreCall()`.
* For non-interruption exceptions, such as model timeouts or rate limiting: log a WARN message and return the original event. The Agent continues running with uncompressed messages, providing graceful degradation.
* For exceptions containing an `InterruptedException`, where Reactor `.block()` wraps checked exceptions in a `RuntimeException`: detect it by walking the cause chain, then return `Mono.error(e)` to correctly propagate the interruption signal.
* Added a private helper method, `isInterruptedException(Throwable)`, which walks through `getCause()` level by level to detect `InterruptedException`.
* Updated the Javadoc for `handlePreReasoning()` to document the new error-handling behavior.

**`AutoContextHookTest.java`**

* Added `testPreReasoningEventHandlesCompressionError`: uses a `FailingModel` that throws `RuntimeException`, triggers strategy-4 compression, and verifies that the hook returns the original event instead of throwing an exception.
* Added `testPreReasoningEventPreservesWrappedInterruptedException`: uses a `FailingModel` that throws `RuntimeException(new InterruptedException(...))`, and verifies that the wrapped interruption exception is correctly propagated through `Mono.error()`.
* Added the `FailingModel` test helper class.

### How to test

```bash
# Run the newly added tests
mvn test -pl agentscope-extensions/agentscope-extensions-autocontext-memory \
  -Dtest="AutoContextHookTest#testPreReasoningEventHandlesCompressionError+testPreReasoningEventPreservesWrappedInterruptedException"

# Run the full module test suite: 183 tests
mvn test -pl agentscope-extensions/agentscope-extensions-autocontext-memory
```

### Checklist

* Code has been formatted with `mvn spotless:apply`
* All tests are passing (`mvn test`) — 183/183
* Javadoc comments are complete and follow project conventions
* Related documentation has been updated, such as links, examples, etc.
* Code is ready for review


LearningGp:
image

xiaoqingwanga:
> image

hi, the conflicts are resolved. pls review the changes. @LearningGp

LearningGp (changes requested):
Placing onErrorResume after .map(...) inadvertently catches exceptions thrown by buildInputMessages (even if rare) and swallows them with a misleading "Failed to compress context" log. I'd suggest chaining onErrorResume directly after compressIfNeededAsync() so exceptions inside .map can bubble up normally.

xiaoqingwanga:
> Placing onErrorResume after .map(...) inadvertently catches exceptions thrown by buildInputMessages (even if rare) and swallows them with a misleading "Failed to compress context" log. I'd suggest chaining onErrorResume directly after compressIfNeededAsync() so exceptions inside .map can bubble up normally.

Yeah, this solution does help. BUT excptions from buildInputMessages are really rare. And if we apply the optimization, we'll re-introduce the original problem - if an exception happens inside the map, the compression disrupts the agent execution again.

LearningGp (approved):
LGTM

LearningGp:
image

xiaoqingwanga:
Close this PR because the compaction has been refactored.

PR #1319, opened 2026-04-29
fix(embedding): make dimensions parameter optional in OpenAITextEmbedding

siphobmalgas:
## Summary
Fixes issue #1095: OpenAITextEmbedding now supports open-source embedding models that don't support the matryoshka dimension parameter.

## Problem
- OpenAI text embedding component forced the `dimensions` parameter to be required
- This parameter is only supported by OpenAI official embedding models (e.g., text-embedding-3-small)
- Open-source models like BAAI/bge-large-zh-v1.5 return 400 error when this parameter is included
- Users could not use OpenAITextEmbedding with open-source models

## Solution
- Changed `dimensions` from required `int` to optional `Integer` (can be null)
- Only include `dimensions` in API request when explicitly configured by user
- Added documentation clarifying when to use this parameter
- Validation only applies when dimensions are explicitly set
- Backward compatible: existing code without explicit dimensions still works

## Changes
- Modified constructor to accept `Integer dimensions` (nullable)
- Updated builder to accept `Integer dimensions` 
- Updated API call logic to conditionally include dimensions parameter
- Updated Javadoc for `dimensions()` builder method

## Testing
- Existing tests continue to pass
- Code is backward compatible with existing configurations
- Now supports both OpenAI official and open-source embedding models

### DIFF
--- a/agentscope-extensions/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/embedding/openai/OpenAITextEmbedding.java
+++ b/agentscope-extensions/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/embedding/openai/OpenAITextEmbedding.java
@@ -142,7 +142,7 @@ public class OpenAITextEmbedding implements EmbeddingModel {
 
                                         // Only include dimensions if explicitly set
                                         if (dimensions != null) {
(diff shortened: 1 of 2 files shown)


LearningGp:
image

PR #1337, opened 2026-05-04
fix(autocontext): preserve tool pairs in current round compression

### LINKED ISSUE #1026: [Bug]:AutoContextMemory Strategy 6 breaks ReAct tool_use/tool_result structure, causing LLM to repeat tool calls infinitely

Alexxigang:
## Summary
- preserve `tool_use` and `tool_result` structure when Strategy 6 compresses the current round
- aggregate current-round tool calls/results into a compact assistant/tool pair instead of flattening them into a single assistant text message
- add regression coverage for current-round compression with tool interactions

## Why this fix
`mergeAndCompressCurrentRoundMessages()` currently offloads the whole current-round segment and rewrites it as a single plain `ASSISTANT` text message. When that segment contains ReAct-style `tool_use` / `tool_result` exchanges, the structural pairing is lost, so the next turn can behave as if the tool call never happened and repeat the invocation.

This change keeps Strategy 6 compact, but preserves the semantic tool-call record by retaining aggregated `ToolUseBlock` and `ToolResultBlock` content in the compressed messages.

## Validation
- `mvn -pl agentscope-extensions/agentscope-extensions-autocontext-memory -am '-Dtest=AutoContextMemoryTest' test`
- `mvn -pl agentscope-extensions/agentscope-extensions-autocontext-memory -am '-Dtest=AutoContextMemoryTest,AutoContextHookTest' test`

Closes #1026

### DIFF
--- a/agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
+++ b/agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
@@ -418,14 +418,18 @@ public class AutoContextMemory implements StateModule, Memory, ContextOffLoader 
                 messagesToCompress.size());
 
         // Step 4: Merge and compress messages (typically tool calls and results)
-        Msg compressedMsg = mergeAndCompressCurrentRoundMessages(messagesToCompress);
+        List<Msg> compressedMessages = mergeAndCompressCurrentRoundMessages(messagesToCompress);
+        if (compressedMessages == null || compressedMessages.isEmpty()) {
+            return false;
+        }
 
         // Build metadata for compression event
         Map<String, Object> metadata = new HashMap<>();
-        if (compressedMsg.getChatUsage() != null) {
-            metadata.put("inputToken", compressedMsg.getChatUsage().getInputTokens());
-            metadata.put("outputToken", compressedMsg.getChatUsage().getOutputTokens());
-            metadata.put("time", compressedMsg.getChatUsage().getTime());
+        Msg eventSourceMsg = compressedMessages.get(0);
+        if (eventSourceMsg.getChatUsage() != null) {
+            metadata.put("inputToken", eventSourceMsg.getChatUsage().getInputTokens());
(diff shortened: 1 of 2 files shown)


Alexxigang:
Friendly follow-up here: this PR is still up to date on my side, and I'm happy to make any requested changes quickly. Thanks for taking a look when you have bandwidth.

jujn:

PR #1340, opened 2026-05-05
fix(transport): add HTTP/2 to HTTP/1.1 automatic fallback for connect…

### LINKED ISSUE #1287: [Bug]: OpenAIChatModel Http2 有时候调用不通大模型，建议设置自动降级功能，阿里云的服务器默认会对Http2进行限流

Mrjyw:
When HTTP/2 connections are dropped by cloud infrastructure
(e.g. Alibaba Cloud SLB GOAWAY, RST_STREAM due to rate limiting),the transport now automatically retries the request with HTTP/1.1
instead of failing immediately. Previously, such connection-level
errors propagated directly to the caller with no protocol fallback.

Fixes #1287

1.0.12-SNAPSHOT

**Background:** Alibaba Cloud SLB/ALB enforces a default limit of concurrent
streams per HTTP/2 connection. When this limit is exceeded, the load balancer sends
GOAWAY or RST_STREAM frames, causing the JDK HttpClient to throw IOExceptions without
an HTTP status code. Other cloud providers (Tencent Cloud CLB, Huawei Cloud ELB) and
enterprise proxies exhibit similar behavior. JDK HttpClient's built-in ALPN
negotiation cannot handle these mid-connection failures since the HTTP/2 connection
was already successfully negotiated.

**Changes:**

- `JdkHttpTransport`: Added `http1_1Client` field, built eagerly in all three
  constructors via `buildClient(config, HttpVersion.HTTP_1_1)`.
- `execute()`: Catches `HttpTransportException` with no status code when HTTP/2 is
  configured, then retries via `doSend(http1_1Client, jdkRequest)`. If the fallback
  also fails, the original error is thrown to preserve the root cause.
- `stream()`: Added `onErrorResume` in the reactive chain that catches connection-
  level errors (IOException or HttpTransportException without status code) and
  retries with the HTTP/1.1 client via `sendAsyncStream()`.
- Extracted `doSend()` and `sendAsyncStream()` helpers that accept a `HttpClient`
  parameter, making both protocol paths share the same execution logic.
- `buildClient()` overloaded to accept an explicit `HttpVersion` parameter.
- Added 7 tests in `JdkHttpTransportTest` covering the full decision tree:
  HTTP/2 fallback success (execute + stream), HTTP/1.1 normal success (execute +
  stream), HTTP/1.1 no fallback, HTTP/2 no fallback for HTTP errors, and fallback
  failure propagating original error.

- [ ] Related documentation has been updated

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
@@ -75,6 +75,9 @@ public class JdkHttpTransport implements HttpTransport {
     private final HttpTransportConfig config;
     private final AtomicBoolean closed = new AtomicBoolean(false);
 
+    /** HTTP/1.1 client for protocol fallback. */
+    private final HttpClient http1_1Client;
+
     /**
      * Create a new JdkHttpTransport with default configuration.
      */
@@ -91,6 +94,7 @@ public class JdkHttpTransport implements HttpTransport {
     JdkHttpTransport(HttpTransportConfig config) {
         this.config = Objects.requireNonNull(config, "config must not be null");
         this.client = buildClient(config);
+        this.http1_1Client = buildClient(config, HttpVersion.HTTP_1_1);
     }
 
     /**
@@ -106,12 +110,17 @@ public class JdkHttpTransport implements HttpTransport {
     public JdkHttpTransport(HttpClient client, HttpTransportConfig config) {
         this.client = Objects.requireNonNull(client, "client must not be null");
         this.config = Objects.requireNonNull(config, "config must not be null");
+        this.http1_1Client = buildClient(config, HttpVersion.HTTP_1_1);
     }
 
     private static HttpClient buildClient(HttpTransportConfig config) {
+        return buildClient(config, config.getHttpVersion());
+    }
+
+    private static HttpClient buildClient(HttpTransportConfig config, HttpVersion version) {
         HttpClient.Builder builder =
                 HttpClient.newBuilder()
-                        .version(config.getHttpVersion().toJdkHttpVersion())
+                        .version(version.toJdkHttpVersion())
                         .followRedirects(Redirect.NORMAL)
                         .connectTimeout(config.getConnectTimeout());
 
@@ -177,7 +186,25 @@ public class JdkHttpTransport implements HttpTransport {
         var jdkRequest = buildJdkRequest(request);
 
         try {
-            var response = client.send(jdkRequest, BodyHandlers.ofString());
+            return doSend(client, jdkRequest);
+        } catch (HttpTransportException e) {
+            if (isConnectionError(e) && shouldFallback()) {
+                log.warn("HTTP/2 request failed, retrying with HTTP/1.1: {}", e.getMessage());
+                try {
+                    return doSend(http1_1Client, jdkRequest);
+                } catch (HttpTransportException ex) {
+                    // Fallback also failed, throw the original error
+                    throw e;
+                }
+            }
+            throw e;
+        }
+    }
+
+    private HttpResponse doSend(HttpClient httpClient, java.net.http.HttpRequest jdkRequest)
+            throws HttpTransportException {
+        try {
+            var response = httpClient.send(jdkRequest, BodyHandlers.ofString());
             return buildHttpResponse(response);
         } catch (IOException e) {
             throw new HttpTransportException("HTTP request failed: " + e.getMessage(), e);
@@ -195,36 +222,25 @@ public class JdkHttpTransport implements HttpTransport {
 
         var jdkRequest = buildJdkRequest(request);
 
-        // Check status code and read error body immediately when CompletableFuture completes
-        // to avoid stream being closed before we can read it
-        CompletableFuture<java.net.http.HttpResponse<InputStream>> future =
-                client.sendAsync(jdkRequest, BodyHandlers.ofInputStream())
-                        .thenApply(
-                                response -> {
-                                    int statusCode = response.statusCode();
-                                    if (statusCode < 200 || statusCode >= 300) {
-                                        // Read error body immediately while stream is still open
-                                        String errorBody = readInputStream(response.body());
-                                        log.warn(
-                                                "HTTP request failed. URL: {} | Status: {} | Error:"
-                                                        + " {}",
-                                                request.getUrl(),
-                                                statusCode,
-                                                errorBody);
-                                        throw new CompletionException(
-                                                new HttpTransportException(
-                                                        "HTTP request failed with status "
-                                                                + statusCode
-                                                                + " | "
-                                                                + errorBody,
-                                                        statusCode,
-                                                        errorBody));
-                                    }
-                                    return response;
-                                });
-
-        return Mono.fromCompletionStage(future)
+        return Mono.fromCompletionStage(sendAsyncStream(client, jdkRequest, request.getUrl()))
                 .flatMapMany(response -> processStreamResponse(response, request))
+                .onErrorResume(
+                        e -> {
+                            if (isStreamConnectionError(e)) {
+                                log.warn(
+                                        "HTTP/2 stream request failed, retrying with HTTP/1.1: {}",
+                                        e.getMessage());
+                                return Mono.fromCompletionStage(
+                                                sendAsyncStream(
+                                                        http1_1Client,
+                                                        jdkRequest,
+                                                        request.getUrl()))
+                                        .flatMapMany(
+                                                response ->
+                                                        processStreamResponse(response, request));
+                            }
+                            return Mono.error(e);
+                        })
                 .publishOn(Schedulers.boundedElastic())
                 .onErrorMap(
                         e -> !(e instanceof HttpTransportException),
@@ -308,6 +324,53 @@ public class JdkHttpTransport implements HttpTransport {
      */
     public boolean isClosed() {
         return closed.get();
+    }
+
+    private static boolean isConnectionError(HttpTransportException e) {
+        return !e.isHttpError();
+    }
+
+    private boolean shouldFallback() {
+        return config.getHttpVersion() == HttpVersion.HTTP_2;
+    }
+
+    private CompletableFuture<java.net.http.HttpResponse<InputStream>> sendAsyncStream(
+            HttpClient httpClient, java.net.http.HttpRequest jdkRequest, String url) {
+        return httpClient
+                .sendAsync(jdkRequest, BodyHandlers.ofInputStream())
+                .thenApply(
+                        response -> {
+                            int statusCode = response.statusCode();
+                            if (statusCode < 200 || statusCode >= 300) {
+                                // Read error body immediately while stream is still open
+                                String errorBody = readInputStream(response.body());
+                                log.warn(
+                                        "HTTP request failed. URL: {} | Status: {} | Error: {}",
+                                        url,
+                                        statusCode,
+                                        errorBody);
+                                throw new CompletionException(
+                                        new HttpTransportException(
+                                                "HTTP request failed with status "
+                                                        + statusCode
+                                                        + " | "
+                                                        + errorBody,
+                                                statusCode,
+                                                errorBody));
+                            }
+                            return response;
+                        });
+    }
+
+    private boolean isStreamConnectionError(Throwable e) {
+        if (!shouldFallback()) {
+            return false;
+        }
+        Throwable cause = e instanceof CompletionException ? e.getCause() : e;
+        if (cause instanceof HttpTransportException hte) {
+            return !hte.isHttpError();
+        }
+        return cause instanceof IOException;
     }
 
     private java.net.http.HttpRequest buildJdkRequest(HttpRequest request) {
--- a/agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
@@ -75,6 +75,9 @@ public class JdkHttpTransport implements HttpTransport {
     private final HttpTransportConfig config;
     private final AtomicBoolean closed = new AtomicBoolean(false);
 
+    /** HTTP/1.1 client for protocol fallback. */
+    private final HttpClient http11Client;
+
     /**
      * Create a new JdkHttpTransport with default configuration.
      */
@@ -91,6 +94,10 @@ public class JdkHttpTransport implements HttpTransport {
     JdkHttpTransport(HttpTransportConfig config) {
         this.config = Objects.requireNonNull(config, "config must not be null");
         this.client = buildClient(config);
+        this.http11Client =
+                config.getHttpVersion() == HttpVersion.HTTP_2
+                        ? buildClient(config, HttpVersion.HTTP_1_1)
+                        : null;
     }
 
     /**
@@ -99,6 +106,14 @@ public class JdkHttpTransport implements HttpTransport {
      * <p>Use this constructor when you want to share an HttpClient instance
      * across multiple components.
      *
+     * <p><b>Note:</b> HTTP/2 to HTTP/1.1 automatic fallback is disabled when a
+     * user-supplied {@code client} is provided. The fallback path would require
+     * building a second client from {@code config} alone, which cannot reflect
+     * any client-level customization (custom {@link java.util.concurrent.Executor},
+     * SSLContext, {@link java.net.Authenticator}, etc.) applied to the injected
+     * client. If you need protocol fallback, use
+     * {@link #JdkHttpTransport(HttpTransportConfig)} instead.
+     *
      * @param client the HttpClient to use
      * @param config the transport configuration (used for reference only)
      * @throws NullPointerException if client or config is null
@@ -106,12 +121,17 @@ public class JdkHttpTransport implements HttpTransport {
     public JdkHttpTransport(HttpClient client, HttpTransportConfig config) {
         this.client = Objects.requireNonNull(client, "client must not be null");
         this.config = Objects.requireNonNull(config, "config must not be null");
+        this.http11Client = null;
     }
 
     private static HttpClient buildClient(HttpTransportConfig config) {
+        return buildClient(config, config.getHttpVersion());
+    }
+
+    private static HttpClient buildClient(HttpTransportConfig config, HttpVersion version) {
         HttpClient.Builder builder =
                 HttpClient.newBuilder()
-                        .version(config.getHttpVersion().toJdkHttpVersion())
+                        .version(version.toJdkHttpVersion())
                         .followRedirects(Redirect.NORMAL)
                         .connectTimeout(config.getConnectTimeout());
 
@@ -177,7 +197,26 @@ public class JdkHttpTransport implements HttpTransport {
         var jdkRequest = buildJdkRequest(request);
 
         try {
-            var response = client.send(jdkRequest, BodyHandlers.ofString());
+            return doSend(client, jdkRequest);
+        } catch (HttpTransportException e) {
+            if (isConnectionError(e) && shouldFallback()) {
+                log.warn("HTTP/2 request failed, retrying with HTTP/1.1: {}", e.getMessage());
+                try {
+                    return doSend(http11Client, jdkRequest);
+                } catch (HttpTransportException ex) {
+                    // Fallback also failed — attach it so both root causes are visible
+                    e.addSuppressed(ex);
+                    throw e;
+                }
+            }
+            throw e;
+        }
+    }
+
+    private HttpResponse doSend(HttpClient httpClient, java.net.http.HttpRequest jdkRequest)
+            throws HttpTransportException {
+        try {
+            var response = httpClient.send(jdkRequest, BodyHandlers.ofString());
             return buildHttpResponse(response);
         } catch (IOException e) {
             throw new HttpTransportException("HTTP request failed: " + e.getMessage(), e);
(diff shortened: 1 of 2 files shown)


Mrjyw:
@LearningGp   PTAL — could you help request a review from the team? All checks have passed. Thanks!

Mrjyw:
@LearningGp friendly ping — all checks have passed, thanks!

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -195,36 +222,25 @@ public Flux<String> stream(HttpRequest request) {
...
-                                                        statusCode,
-                                                        errorBody));
-                                    }
-                                    return response;
-                                });
-
-        return Mono.fromCompletionStage(future)
+        return Mono.fromCompletionStage(sendAsyncStream(client, jdkRequest, request.getUrl()))
                 .flatMapMany(response -> processStreamResponse(response, request))
+                .onErrorResume(
```
AgentScopeJavaBot, line 238: [major] `onErrorResume` is placed **after** `flatMapMany(processStreamResponse)`, so it also covers errors that surface once response headers are received and chunks may already have been emitted downstream. It is currently safe only because `BufferedReader.lines()` wraps mid-stream IO failures as `UncheckedIOException` (which is not an `IOException`, so `isStreamConnectionError` returns false). That implicit dependency is fragile — any future refactor that surfaces a raw `IOException` during streaming would silently replay the request and emit **duplicate** SSE/NDJSON chunks to subscribers. Suggest moving the resume above `flatMapMany` so the fallback only kicks in when connection setup fails:

```java
return Mono.fromCompletionStage(sendAsyncStream(client, jdkRequest, request.getUrl()))
        .onErrorResume(e -> isStreamConnectionError(e)
                ? Mono.fromCompletionStage(sendAsyncStream(http1_1Client, jdkRequest, request.getUrl()))
                : Mono.error(e))
        .flatMapMany(response -> processStreamResponse(response, request))
        // ...
```

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -106,12 +110,17 @@ public class JdkHttpTransport implements HttpTransport {
     public JdkHttpTransport(HttpClient client, HttpTransportConfig config) {
         this.client = Objects.requireNonNull(client, "client must not be null");
         this.config = Objects.requireNonNull(config, "config must not be null");
+        this.http1_1Client = buildClient(config, HttpVersion.HTTP_1_1);
```
AgentScopeJavaBot: [major] This constructor's Javadoc states the `config` is "used for reference only", but the new line builds the HTTP/1.1 fallback purely from `config` and ignores everything the caller may have customized on their injected `HttpClient` (custom `Executor`, `SSLContext`, `Authenticator`, executor-bound trace context, etc.). For users who pass a tuned client to share a thread pool or attach custom telemetry, the fallback path will silently behave differently from the primary path. Either skip the fallback when a user-supplied client is provided (most predictable), or update the Javadoc to make clear that any client-level customization must now also be expressed via `HttpTransportConfig`.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -177,7 +186,25 @@ public HttpResponse execute(HttpRequest request) throws HttpTransportException {
...
-            var response = client.send(jdkRequest, BodyHandlers.ofString());
+            return doSend(client, jdkRequest);
+        } catch (HttpTransportException e) {
+            if (isConnectionError(e) && shouldFallback()) {
+                log.warn("HTTP/2 request failed, retrying with HTTP/1.1: {}", e.getMessage());
+                try {
+                    return doSend(http1_1Client, jdkRequest);
+                } catch (HttpTransportException ex) {
+                    // Fallback also failed, throw the original error
+                    throw e;
```
AgentScopeJavaBot, line 209: [minor] When the HTTP/1.1 fallback also fails, `ex` is caught but discarded — only the original HTTP/2 error is rethrown. Operators investigating production failures will have no signal as to why the fallback failed (DNS, TLS handshake, timeout, etc.). Attach it as a suppressed exception so the root-cause chain is preserved:

```java
} catch (HttpTransportException ex) {
    e.addSuppressed(ex);
    throw e;
}
```

The inner `Mono.fromCompletionStage(...)` in `stream()` has the same loss-of-context issue.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -91,6 +94,7 @@ public class JdkHttpTransport implements HttpTransport {
     JdkHttpTransport(HttpTransportConfig config) {
         this.config = Objects.requireNonNull(config, "config must not be null");
         this.client = buildClient(config);
+        this.http1_1Client = buildClient(config, HttpVersion.HTTP_1_1);
```
AgentScopeJavaBot: [minor] The fallback `HttpClient` is built eagerly in every constructor, including when `config.getHttpVersion() == HTTP_1_1` (where `shouldFallback()` is permanently false and the client is never used). Each JDK `HttpClient` allocates its own selector/IO threads and connection pool, so this doubles the per-transport footprint for users who explicitly chose HTTP/1.1. Suggest building it conditionally and gating the fallback branches on a non-null reference:

```java
this.http1_1Client = config.getHttpVersion() == HttpVersion.HTTP_2
        ? buildClient(config, HttpVersion.HTTP_1_1)
        : null;
```

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -310,6 +326,53 @@ public boolean isClosed() {
         return closed.get();
     }
 
+    private static boolean isConnectionError(HttpTransportException e) {
+        return !e.isHttpError();
+    }
+
+    private boolean shouldFallback() {
+        return config.getHttpVersion() == HttpVersion.HTTP_2;
```
AgentScopeJavaBot: [minor] `shouldFallback()` keys off `config.getHttpVersion()` rather than the actual configured version of `client`. For users of the explicit-client constructor whose injected `HttpClient` is already HTTP/1.1 but whose `config` still says HTTP_2, a connection drop would trigger a redundant retry on a freshly-built duplicate HTTP/1.1 client and potentially mask a real HTTP/1.1 issue as "HTTP/2 fallback succeeded". Consider `client.version() == HttpClient.Version.HTTP_2`, which is authoritative regardless of how the transport was constructed.

agentscope-core/src/test/java/io/agentscope/core/model/transport/JdkHttpTransportTest.java
```diff
@@ -1063,4 +1065,242 @@ void testHttpVersionConfig() {
...
+    // ── HTTP/2 → HTTP/1.1 fallback tests ───────────────────────────────
+    //
+    // Decision tree tested:
+    //
+    //   Connection error (statusCode == null)?
+    //   ├─ Yes → HTTP_2 configured?
+    //   │        ├─ Yes → fallback to HTTP/1.1, retry once
+    //   │        │        ├─ Success → return response
+    //   │        │        └─ Failure → throw original error
+    //   │        └─ No (already HTTP_1_1) → throw (no lower protocol)
```
AgentScopeJavaBot, line 1078: [minor] The decision-tree comment covers connection error → fallback success/failure, but mid-stream failure (connection drops after headers and after some chunks were emitted) is not exercised — exactly the scenario where the `onErrorResume` placement in `stream()` matters. Consider adding a test using `SocketPolicy.DISCONNECT_DURING_RESPONSE_BODY` (or `MockResponse.setBodyDelay` + `shutdownInput`) to lock down that mid-stream errors do **not** trigger a re-emit.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -75,6 +75,9 @@ public class JdkHttpTransport implements HttpTransport {
     private final HttpTransportConfig config;
     private final AtomicBoolean closed = new AtomicBoolean(false);
 
+    /** HTTP/1.1 client for protocol fallback. */
+    private final HttpClient http1_1Client;
```
AgentScopeJavaBot: [nit] `http1_1Client` is the only field in the entire `transport` package using an underscore; all other fields (`connectTimeout`, `keepAliveDuration`, `httpVersion`, etc.) use pure camelCase. Rename to `http11Client` (or `fallbackClient`) for consistency.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR addresses a real production issue (#1287) where Alibaba Cloud SLB-style HTTP/2 GOAWAY/RST_STREAM frames produced unrecoverable IOExceptions. The fallback design is sound (gate on "no HTTP status code", reuse the same request, retry exactly once over HTTP/1.1) and is well-covered by seven new tests across both `execute` and `stream` paths. Two design points warrant a second look before merging: (1) in `stream()`, `onErrorResume` is placed downstream of `flatMapMany(processStreamResponse)`, so it would also catch mid-stream IO failures and re-emit duplicate chunks — currently safe only because `BufferedReader.lines()` wraps mid-stream errors as `UncheckedIOException`, which is fragile; and (2) the explicit-`HttpClient` constructor's contract (config "for reference only") is silently violated since `config` is now used to build the fallback client, losing any customization on the user-supplied primary client. Suppressed-exception preservation, conditional fallback-client construction, and field naming are smaller cleanups. No critical correctness bugs given the current code shape, but the streaming `onErrorResume` placement deserves to be hardened.

Mrjyw:
您的邮件我已收到，谢谢。——收信人：吴宇杰

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
Mrjyw: Done. Moved onErrorResume before flatMapMany so fallback only triggers on connection-setup failures, not mid-stream errors

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
Mrjyw: Done. Set http11Client = null in the public constructor to skip fallback entirely, and updated Javadoc to document the limitation.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
Mrjyw: Done. Added e.addSuppressed(ex) in execute() and used onErrorMap in stream() to preserve both the original and fallback failure causes.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
Mrjyw: Done. Now only builds the HTTP/1.1 client when config.getHttpVersion() == HTTP_2, otherwise null. Fallback paths gate on non-null.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
Mrjyw: Done. Changed to client.version() == HttpClient.Version.HTTP_2 && http11Client != null, which is authoritative regardless of constructor path.

agentscope-core/src/test/java/io/agentscope/core/model/transport/JdkHttpTransportTest.java
Mrjyw: Done. Added testHttp2MidStreamDisconnectNoFallback() using SocketPolicy.DISCONNECT_DURING_RESPONSE_BODY, asserting only 1 request was made.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
Mrjyw: Done. Renamed to http11Client for consistency with the project's camelCase convention.

Mrjyw:
@LearningGp  All checks have passed and all AI review comments have been addressed. PTAL, thanks!

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
AgentScopeJavaBot: ✅ Verified as addressed in `08e68e8`: onErrorResume now placed before flatMapMany at lines 234-248, so fallback only triggers on connection-setup failures.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
AgentScopeJavaBot: ✅ Verified as addressed in `08e68e8`: http11Client set to null in public constructor at line 116; Javadoc updated to document the fallback limitation for user-supplied clients.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
AgentScopeJavaBot: ✅ Verified as addressed in `08e68e8`: e.addSuppressed(ex) added in execute() at line 208 and onErrorMap used in stream() at lines 245-248 to preserve both failure causes.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
AgentScopeJavaBot: ✅ Verified as addressed in `08e68e8`: http11Client now conditionally built only when HTTP_2 configured at lines 94-98; fallback paths gate on non-null check.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
AgentScopeJavaBot: ✅ Verified as addressed in `08e68e8`: shouldFallback() now uses client.version() == HttpClient.Version.HTTP_2 at line 347, authoritative regardless of constructor path.

agentscope-core/src/test/java/io/agentscope/core/model/transport/JdkHttpTransportTest.java
AgentScopeJavaBot: ✅ Verified as addressed in `08e68e8`: testHttp2MidStreamDisconnectNoFallback added at line 1309 using DISCONNECT_DURING_RESPONSE_BODY, asserts only 1 request was made.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
AgentScopeJavaBot: ✅ Verified as addressed in `08e68e8`: Field renamed to http11Client at line 80 for camelCase consistency.

AgentScopeJavaBot (approved):

oss-maintainer:

```bash
git fetch origin
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

Mrjyw:
您的邮件我已收到，谢谢。——收信人：吴宇杰

oss-maintainer (commented):
### Summary
The protocol-fallback idea is sound and the test matrix around it (execute vs stream, pre-flight vs mid-body, `getRequestCount() == 1` for the mid-stream case) is the strongest part of the PR — the decision-tree comment in the test file is exactly the documentation this area needs.

Two things keep this from being approvable as-is:

1. **The trigger condition no longer describes what the client negotiates.** `http11Client` is built only for `config.getHttpVersion() == HTTP_2`, but #3214 made `null` (auto) the default and moved cleartext down to HTTP/1.1 *per request*, so on current `main` the guard is unreachable for a default-configured transport. The PR needs a rebase onto that model, and the check should key off the version resolved for the request rather than a constructor-time field.
2. **The non-streaming retry is not bounded to pre-send failures.** `isConnectionError(e) == !e.isHttpError()` retries on any status-less failure, which includes a connection that breaks after the POST body reached the server — a duplicate completion for a chat request. The streaming path draws that line carefully; `execute()` does not.

Also: this branch has conflicts with `main` (`mergeable=CONFLICTING`, reminder posted on 2026-09-19) and was opened in May — if the rebase turns out to be more trouble than the value, say so and we can close it out with a note about #3214 covering the cleartext half of the problem.

---

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -177,7 +197,26 @@ public HttpResponse execute(HttpRequest request) throws HttpTransportException {
         var jdkRequest = buildJdkRequest(request);
 
         try {
-            var response = client.send(jdkRequest, BodyHandlers.ofString());
+            return doSend(client, jdkRequest);
+        } catch (HttpTransportException e) {
+            if (isConnectionError(e) && shouldFallback()) {
```
oss-maintainer, line 202: `isConnectionError(e)` is `!e.isHttpError()`, i.e. *any* failure without an HTTP status code triggers a resend of the same `jdkRequest`. For a model `POST /chat/completions` that is not an idempotent retry: `HttpClient.send()` throws `IOException` both when the connection never came up **and** when it breaks after the request body was written and accepted. In the second case the server may already have produced (and billed) the completion, and the fallback sends it a second time — plus, for a tool-calling turn, re-runs nothing client-side but does duplicate provider-side work and can duplicate a `request_id`-keyed conversation turn.

The streaming path already thought about exactly this (`onErrorResume` placed before `flatMapMany`, with `testHttp2MidStreamDisconnectNoFallback` asserting `getRequestCount() == 1`), but `execute()` has no equivalent boundary — a `DISCONNECT_DURING_RESPONSE_BODY` on a non-streaming call lands here and retries.

Suggested narrowing, in order of preference:
* only fall back when the failure is provably pre-send (`ConnectException`, or `HttpTimeoutException` on connect) rather than "no status code";
* or gate the whole feature behind an explicit opt-in on `HttpTransportConfig` and document it as at-least-once, so a caller that only knows how to make one POST can turn it off;
* and add a non-streaming mid-response-disconnect test mirroring the streaming one, asserting `getRequestCount() == 1`, so this boundary is pinned.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -91,6 +94,10 @@ public class JdkHttpTransport implements HttpTransport {
     JdkHttpTransport(HttpTransportConfig config) {
         this.config = Objects.requireNonNull(config, "config must not be null");
         this.client = buildClient(config);
+        this.http11Client =
+                config.getHttpVersion() == HttpVersion.HTTP_2
```
oss-maintainer, line 98: The fallback client is created only when `config.getHttpVersion() == HttpVersion.HTTP_2`, and `shouldFallback()` re-checks that condition. Since #3214 ("明文请求默认改用 HTTP/1.1") landed on `main`, `HttpTransportConfig.Builder.httpVersion` **defaults to `null`** (auto: cleartext resolves to HTTP/1.1 per request, https inherits the client-level HTTP/2 via ALPN), and `buildClient` passes `HTTP_2` as the client default for the `null` case. So on current `main` this guard is unreachable for any default-configured transport — the transports that still talk h2c/ALPN are exactly the ones whose client-level version is `HTTP_2` while `config.getHttpVersion()` is `null`.

This PR is 4 months old and conflicts on this file, so it has to rebase onto #3214's per-request model anyway. While doing that, could the trigger key off the version actually used for *this request* (the resolved `resolveRequestVersion(...)` / `request.version()`) instead of the constructor-time config value? As written the decision is made against a field that no longer describes what the client will negotiate, and the failure the PR is fixing (a server that mishandles h2) is a per-endpoint property, not a per-transport one.

Also worth re-checking the tests after the rebase: they all point `HttpVersion.HTTP_2` at a cleartext MockWebServer, which post-#3214 exercises the explicit h2c-opt-in path rather than the auto path a real user is on.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -177,7 +197,26 @@ public HttpResponse execute(HttpRequest request) throws HttpTransportException {
...
-            var response = client.send(jdkRequest, BodyHandlers.ofString());
+            return doSend(client, jdkRequest);
+        } catch (HttpTransportException e) {
+            if (isConnectionError(e) && shouldFallback()) {
+                log.warn("HTTP/2 request failed, retrying with HTTP/1.1: {}", e.getMessage());
+                try {
+                    return doSend(http11Client, jdkRequest);
+                } catch (HttpTransportException ex) {
+                    // Fallback also failed — attach it so both root causes are visible
+                    e.addSuppressed(ex);
```
oss-maintainer, line 208: When both attempts fail, the caller always gets the **HTTP/2** exception and the HTTP/1.1 one is demoted to `addSuppressed`. For the scenario this PR targets — a server that cannot speak h2 at all — the suppressed 1.1 error is usually the more actionable one (a real status code, a provider error body), and the thrown error's message ("HTTP request failed: ...") describes the attempt the user cannot fix. Symmetrically, in the streaming branch the same `addSuppressed(ex)` runs inside `onErrorMap`, where `e` is the unwrapped cause, so whether the suppressed entry survives to the caller depends on how Reactor re-wraps it — worth one assertion in a test rather than reasoning about it.

Suggest either throwing the fallback error when it carries an HTTP status (keeping the original suppressed) or, minimally, wrapping both in a single exception whose message names both protocols and both root causes. `getSuppressed()` is not read by the retry/logging layers in `JdkSseTransport`'s callers, so suppressed-only detail is effectively invisible today.

PR #1377, opened 2026-05-11
fix(core): place cache_control at content block level per DashScope API spec

### LINKED ISSUE #1363: [Bug] DashScope native API: cache_control placed at message level instead of content block level
## Description

The explicit cache control implementation in both `DashScopeChatModel` and `OpenAIChatModel` places `cache_control` at the **message level**, but the [DashScope official documentation](https://help.aliyun.com/zh/model-studio/context-cache) requires it to be placed **inside the content block** (within the `content` array). This applies to both the DashScope native protocol and the OpenAI-compatible protocol.

## Current Behavior

When `GenerateOptions.cacheControl(true)` is enabled, both `DashScopeChatFormatter.applyCacheControl()` and `OpenAIBaseFormatter.applyCacheControl()` set `cache_control` on the message object directly (message level).

**DashScope formatter:**

```java
public void applyCacheControl(List<DashScopeMessage> messages) {
    for (DashScopeMessage msg : messages) {
        if ("system".equals(msg.getRole()) && msg.getCacheControl() == null) {
            msg.setCacheControl(EPHEMERAL_CACHE_CONTROL);  // message level
        }
    }
    DashScopeMessage lastMsg = messages.get(messages.size() - 1);
    if (lastMsg.getCacheControl() == null) {
        lastMsg.setCacheControl(EPHEMERAL_CACHE_CONTROL);  // message level
    }
}
```

**OpenAI formatter** has the same logic in `OpenAIBaseFormatter.applyCacheControl()`.

This produces the following JSON for both protocols:

```json
{
  "role": "system",
  "content": "You are a helpful assistant.",
  "cache_control": {"type": "ephemeral"}
}
```

## Expected Behavior

Per the [official documentation](https://help.aliyun.com/zh/model-studio/context-cache), `cache_control` must be placed **inside a content block**, and `content` must be in **array format**:

```json
{
  "role": "system",
  "content": [
    {
      "type": "text",
      "text": "You are a helpful assistant.",
      "cache_control": {"type": "ephemeral"}
    }
  ]
}
```

The documentation states:

> 需将 content 字段改为数组形式，并添加 cache_control 字段

This format requirement applies to both OpenAI-compatible and DashScope native protocols when calling DashScope models.

## Issues Identified

1. **`cache_control` is placed at the wrong level** — should be inside content blocks, not at message level. This affects both `DashScopeChatFormatter` and `OpenAIBaseFormatter`.
2. **Content part DTOs lack a `cache_control` field** — both `DashScopeContentPart` and `OpenAIMessage` content parts have no way to carry `cache_control` at the content block level.
3. **Multimodal messages are not handled** — when content is already a list of content parts, the `cache_control` still goes to message level and won't be recognized by the API.
4. **No guard for the 4-marker limit** — the documentation states a maximum of 4 `cache_control` markers per request. If there are multiple system messages (e.g., injected by SkillHook or LongTermMemoryHook), the limit may be exceeded silently.

### DashScope native protocol

1. Add a `cache_control` field (`Map<String, String>`) to `DashScopeContentPart`.
2. Modify `DashScopeChatFormatter.applyCacheControl()` to:
   - Convert string `content` to array format (`List<DashScopeContentPart>`) for target messages.
   - Set `cache_control` on the **last content block** within each target message.
3. Apply the same fix to `DashScopeMultiAgentFormatter.applyCacheControl()`.

### OpenAI-compatible protocol

1. Modify `OpenAIBaseFormatter.applyCacheControl()` with the same content-block-level approach.
2. Add `cache_control` support to OpenAI content part DTOs.

### Common

1. Add a guard to ensure no more than 4 `cache_control` markers per request.
2. Keep the existing message-level `cacheControl` fields for backward compatibility (manual metadata marking).

## Affected Classes

- `DashScopeChatFormatter`
- `DashScopeMultiAgentFormatter`
- `DashScopeContentPart`
- `DashScopeMessage`
- `OpenAIBaseFormatter`
- `OpenAIMessage`

## Discussion: Should `applyCacheControl()` auto-mark messages?

The current `applyCacheControl()` strategy is: **"all system messages + last message"**. Given DashScope's prefix-matching caching mechanism, the strategy pattern itself is sound in theory:

- **Marking system messages** creates layered prefix cache blocks (A, AB, ABC…). Even if later messages change, the shorter prefix (e.g., just the stable system prompt) can still be hit — a reasonable tiered caching approach.
- **Marking the last message** caches the entire messages array as a complete prefix, which aligns with the official "continuous multi-turn dialog" pattern.

However, there are two practical concerns:

### 1. The 4-marker limit

The API enforces a hard limit of **4 `cache_control` markers per request**. If more than 4 markers are present, only the last 4 take effect. In AgentScope, multiple hooks can dynamically inject system messages (e.g., SkillHook, LongTermMemoryHook, RAGHook), making the number of system messages unpredictable at the formatter level. When the total marker count exceeds 4, the earliest system messages — which are typically the most stable and most valuable to cache — will lose their markers and fall out of the cache.

### 2. Dynamic content defeats prefix caching

The framework cannot distinguish between **stable** system messages (e.g., the user's own system prompt) and **dynamic** ones (e.g., RAG-retrieved knowledge, long-term memory summaries). In AgentScope's hook architecture, hooks like `GenericRAGHook` and `StaticLongTermMemoryHook` inject system messages whose **content changes on every request**. Marking these with `cache_control` means each request creates a new cache block (at 125% of standard input cost) that will likely **never be hit** — the prefix changes every time.

Only the user knows which parts of their messages are stable and worth caching. A blanket "mark all system messages" strategy applied at the formatter level cannot make this distinction.

### Suggestion

Consider making cache control **user-driven** rather than automatic:

- The existing `MessageMetadataKeys.CACHE_CONTROL` mechanism already allows users to mark individual `Msg` objects for caching via metadata, which flows through `applyCacheControlFromMetadata()`.
- The automatic `applyCacheControl()` strategy could be **removed or made opt-in**, letting users who understand their caching needs and cost tolerance decide which messages to mark.
- If keeping an automatic strategy, add a guard to enforce the 4-marker limit and prioritize stable prefixes (first system message + last message).

## References

- DashScope Context Cache documentation: https://help.aliyun.com/zh/model-studio/context-cache
- DashScope Java SDK correct usage: `MessageContentText.builder().text(...).cacheControl(CacheControl.builder().type("ephemeral").build()).build()`

flystar32:
## Summary

Fixes #1363

Per the [DashScope API documentation](https://help.aliyun.com/zh/model-studio/context-cache), `cache_control` must be placed inside **content blocks** (within the `content` array), not at the **message level**. The previous implementation incorrectly set `cache_control` as a sibling of `role` and `content`, which prevented context caching from working properly.

This fix applies to both DashScope native and OpenAI-compatible protocol paths.

## Changes

**DTO layer (2 files):**
- `DashScopeContentPart` — add `cacheControl` field with getter/setter/builder
- `OpenAIContentPart` — same as above

**Formatter layer (4 files):**
- `DashScopeChatFormatter.applyCacheControl()` — rewrite to set `cache_control` on the last content block of target messages; add `applyCacheControlToContentBlock()` and `ensureContentArray()` utility methods (auto-converts string content to array format)
- `DashScopeMultiAgentFormatter.applyCacheControl()` — delegate to `DashScopeChatFormatter`'s static methods
- `OpenAIBaseFormatter.applyCacheControl()` — symmetric changes for OpenAI path
- `DashScopeMessageConverter` / `OpenAIMessageConverter` — update `applyCacheControlFromMetadata()` to use content block level placement

**Before (incorrect — message level):**
```json
{
  "role": "system",
  "content": "You are helpful.",
  "cache_control": {"type": "ephemeral"}
}
```

**After (correct — content block level):**
```json
{
  "role": "system",
  "content": [
    {"text": "You are helpful.", "cache_control": {"type": "ephemeral"}}
  ]
}
```

## Test plan

- [x] `DashScopeCacheControlTest` — 16 test cases covering content block level placement, string→array conversion, manual marking preservation, JSON serialization, metadata marking
- [x] `OpenAICacheControlTest` — 13 test cases covering the same scenarios for OpenAI path
- [x] `CacheControlE2ETest` — real API calls to both DashScope native and OpenAI-compatible endpoints (requires `DASHSCOPE_API_KEY` + `ENABLE_E2E_TESTS=true`)
- [x] Regression: all 3651 unit tests pass (0 failures), existing model integration tests (`DashScopeChatModelTest`, `OpenAIChatModelTest`) and ground truth tests pass
- [x] E2E verified: both DashScope native API and OpenAI-compatible endpoint (`/compatible-mode/v1`) accept the corrected format and return successful responses

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/dashscope/DashScopeChatFormatter.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/dashscope/DashScopeChatFormatter.java
@@ -16,6 +16,7 @@
 package io.agentscope.core.formatter.dashscope;
 
 import io.agentscope.core.formatter.AbstractBaseFormatter;
+import io.agentscope.core.formatter.dashscope.dto.DashScopeContentPart;
 import io.agentscope.core.formatter.dashscope.dto.DashScopeInput;
 import io.agentscope.core.formatter.dashscope.dto.DashScopeMessage;
 import io.agentscope.core.formatter.dashscope.dto.DashScopeParameters;
@@ -173,11 +174,15 @@ public class DashScopeChatFormatter
     }
 
     /**
-     * Apply cache control to DashScope messages.
+     * Apply cache control to DashScope messages at the content block level.
      *
-     * <p>Adds <code>cache_control: {"type": "ephemeral"}</code> to all system messages and the last
-     * message in the list. Messages that already have cache_control set (e.g., via manual metadata
-     * marking) will not be overwritten.
+     * <p>Per the DashScope API specification, {@code cache_control} must be placed inside content
+     * blocks (within the {@code content} array), not at the message level. This method converts
+     * string content to array format when needed and sets {@code cache_control} on the last content
+     * block of each target message.
+     *
+     * <p>Target messages: all system messages and the last message in the list. Messages whose last
+     * content block already has {@code cache_control} set will not be overwritten.
      *
      * @param messages the list of formatted DashScope messages
      */
@@ -186,14 +191,48 @@ public class DashScopeChatFormatter
             return;
         }
         for (DashScopeMessage msg : messages) {
-            if ("system".equals(msg.getRole()) && msg.getCacheControl() == null) {
-                msg.setCacheControl(EPHEMERAL_CACHE_CONTROL);
+            if ("system".equals(msg.getRole())) {
+                applyCacheControlToContentBlock(msg);
             }
         }
         DashScopeMessage lastMsg = messages.get(messages.size() - 1);
-        if (lastMsg.getCacheControl() == null) {
-            lastMsg.setCacheControl(EPHEMERAL_CACHE_CONTROL);
+        applyCacheControlToContentBlock(lastMsg);
+    }
+
+    /**
+     * Apply ephemeral cache_control to the last content block of the given message.
+     * If content is a plain string, it is first converted to array format.
+     * Skips if the last content block already has cache_control set.
+     */
+    static void applyCacheControlToContentBlock(DashScopeMessage msg) {
+        List<DashScopeContentPart> parts = ensureContentArray(msg);
+        if (parts.isEmpty()) {
+            return;
         }
+        DashScopeContentPart lastPart = parts.get(parts.size() - 1);
+        if (lastPart.getCacheControl() == null) {
+            lastPart.setCacheControl(EPHEMERAL_CACHE_CONTROL);
+        }
+    }
+
+    /**
+     * Ensure the message content is in array format ({@code List<DashScopeContentPart>}).
+     * If content is a plain string, converts it to {@code [{"type":"text","text":"..."}]}.
+     *
+     * @return the content part list (never null, may be empty)
+     */
+    @SuppressWarnings("unchecked")
+    static List<DashScopeContentPart> ensureContentArray(DashScopeMessage msg) {
+        Object content = msg.getContent();
+        if (content instanceof List) {
+            return (List<DashScopeContentPart>) content;
+        }
+        List<DashScopeContentPart> parts = new ArrayList<>();
+        if (content instanceof String text) {
+            parts.add(DashScopeContentPart.text(text));
+        }
+        msg.setContent(parts);
+        return parts;
     }
 
     /**
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/dashscope/DashScopeMessageConverter.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/dashscope/DashScopeMessageConverter.java
@@ -249,7 +249,9 @@ public class DashScopeMessageConverter {
     }
 
     /**
-     * Apply cache_control from Msg metadata to the converted DashScopeMessage.
+     * Apply cache_control from Msg metadata to the converted DashScopeMessage at content block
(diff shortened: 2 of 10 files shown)


LearningGp (approved):
The "auto-mark all system + last message" strategy fundamentally conflicts with the framework's hook architecture. Since the system messages injected by `GenericRAGHook` and `StaticLongTermMemoryHook` vary per request, adding markers to them means creating a new cache block that will never result in a hit, this is essentially a reverse charge.

This PR only fixes the "incorrect placement" bug and doesn't touch the strategy layer, so it shouldn't block the merge. However, once this fix lands, resolving the bug means these "pseudo-caches" will actually start incurring costs. I recommend opening a new issue to track this. PTAL @chickenlj

charlotteForever:
Hi @flystar32 @chickenlj,

I reproduced #1363 on current main at `6059782`. The existing PR targets the pre-module-split `agentscope-core` paths, while the current OpenAI/DashScope formatters live in the extension modules and include the newer `CACHE_CONTROL=false` semantics.

I’m happy to port the fix to the current module structure and update the regression tests. Would you prefer an update to #1377, or should I open a superseding PR?

Thanks!

oss-maintainer:

```bash
git fetch origin
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #1400, opened 2026-05-15
fix(plan): ensure changeHook receives finished plan instead of null in finishPlan()

birdie7761:
## Description

  `PlanNotebook.finishPlan()` sets `currentPlan = null` before calling `triggerPlanChangeHooks()`, causing all registered change hooks to receive `null` instead of the finished plan. This prevents downstream event emission (e.g.
  `TASK_EXECUTION_SUCCEEDED`, `TASK_EXECUTION_CANCELLED`) and makes `getCurrentPlan()` return `null` during hook execution.

  All other mutation methods (`createPlan`, `updatePlanInfo`, `finishSubtask`, etc.) trigger hooks while `currentPlan` still references a valid object. `finishPlan` is the only method that nulls before hooking.

  **Root cause** — `PlanNotebook.java`:

  ```java
  // Before fix: null first, then notify
  currentPlan = null;
  return triggerPlanChangeHooks()

  triggerPlanChangeHooks() reads this.currentPlan directly, so hooks receive null.

  Fix: trigger hooks first, then null currentPlan:

  return storage.addPlan(currentPlan)
          .then(triggerPlanChangeHooks())
          .then(Mono.fromRunnable(() -> currentPlan = null))
          .thenReturn(message);

  Checklist

  - Code has been formatted with mvn spotless:apply
  - All tests are passing (mvn test)
  - Javadoc comments are complete and follow project conventions
  - Related documentation has been updated (e.g. links, examples, etc.)
  - Code is ready for review


LearningGp (approved):
LGTM

PR #1407, opened 2026-05-15
fix: check matching IDs before skipping pending tool recovery

### LINKED ISSUE #1406: [Bug]: PendingToolRecoveryHook skips recovery when input contains unrelated ToolResultBlock IDs
# PendingToolRecoveryHook skips recovery when input contains unrelated ToolResultBlock IDs

## Describe the bug

`PendingToolRecoveryHook` treats any incoming `ToolResultBlock` as user-provided tool results, without checking whether the tool result IDs match the current pending tool call IDs.

In `PendingToolRecoveryHook.handlePreCall`, the current logic only checks whether any `ToolResultBlock` exists:

```java
boolean userProvidedResults =
if (userProvidedResults) {
    return Mono.just(event);
}
```

This means that if the input contains a `ToolResultBlock` with an unrelated, stale, or invalid ID, the hook skips auto-recovery even though the actual pending tool calls are still unresolved.

Later, `ReActAgent#doCall` validates the provided tool result IDs against `pendingIds` and may throw an exception, instead of letting `PendingToolRecoveryHook` recover the real pending calls.

## To Reproduce

1. Enable pending tool recovery:

```java
        // other builder config
        .enablePendingToolRecovery(true)
        .build();
```

2. Let the agent memory contain a pending tool call, for example pending ID `tool_call_1`.

3. Call the agent with an input message that contains a `ToolResultBlock`, but its ID does not match the pending tool call ID, for example `tool_call_old`.

4. `PendingToolRecoveryHook` sees that a `ToolResultBlock` exists and returns without patching `tool_call_1`.

5. `ReActAgent#doCall` then validates the result IDs and fails because `tool_call_old` is not in `pendingIds`.

## Expected behavior

`PendingToolRecoveryHook` should check the IDs of incoming `ToolResultBlock`s against `pendingIds`.

It should only treat input tool results as user-provided results for the current pending calls when their IDs match the current pending tool call IDs.

For example, the hook could collect provided result IDs and compare them with `pendingIds` before deciding whether to skip auto-recovery.

## Error messages

Possible error from `ReActAgent#doCall` / `validateAndAddToolResults`:

```text
Invalid tool result IDs: [tool_call_old]. Expected: [tool_call_1]
```

## Environment

- AgentScope-Java Version: [please fill in]
- Java Version: [please fill in]
- OS: Windows

## Additional context

The current implementation checks only for the existence of `ToolResultBlock`:

```java
m.hasContentBlocks(ToolResultBlock.class)
```

but it does not check whether those `ToolResultBlock` IDs are related to the pending tool calls found by:

```java
Set<String> pendingIds = findPendingToolUseIds(memory);
```

This can cause pending tool recovery to be skipped incorrectly.

Aaromalpm:
Fixes #1406

  ## Problem
  The hook was skipping auto-recovery whenever any `ToolResultBlock` existed
  in the input messages, even if those results were unrelated to the pending
  tool call IDs.

  ## Fix
  Instead of checking if any `ToolResultBlock` exists, now we extract the IDs
  from the input `ToolResultBlock`s and check if they actually match the
  pending tool call IDs. Recovery is only skipped when there is a genuine match.
  4. Make sure the base repository is agentscope-ai/agentscope-java and base branch is main
  5. Click Create pull request

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
@@ -107,9 +107,11 @@ public class PendingToolRecoveryHook implements Hook {
             return Mono.just(event);
         }
 
-        boolean userProvidedResults =
-                inputMessages.stream().anyMatch(m -> m.hasContentBlocks(ToolResultBlock.class));
-        if (userProvidedResults) {
+        boolean userProvidedMatchingResults = inputMessages.stream()
+                .flatMap(m -> m.getContentBlocks(ToolResultBlock.class).stream())
+                .map(ToolResultBlock::getId)
+                .anyMatch(pendingIds::contains);
+        if (userProvidedMatchingResults) {
             return Mono.just(event);
         }
 
--- a/agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
@@ -107,10 +107,11 @@ public class PendingToolRecoveryHook implements Hook {
             return Mono.just(event);
         }
 
-        boolean userProvidedMatchingResults = inputMessages.stream()
-                .flatMap(m -> m.getContentBlocks(ToolResultBlock.class).stream())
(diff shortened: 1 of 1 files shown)


LearningGp:
image

PR #1431, opened 2026-05-17
fix(tool): preserve tool id and name in timeout error results

### LINKED ISSUE #1389: [Bug]: agent tool  ExecutionConfig timeout make postActingEvent.getToolResult() id name is null
agentscope verion: 1.0.12

Image

```java
postActingEvent.getToolResultMsg().getContent()  ToolResultBlock is ok, but postActingEvent.getToolResult() id and name is null
```

Sparkle6979:
When a tool call times out in `executeWithInfrastructure()`, the error path via `onErrorResume` creates a `ToolResultBlock.error()` without calling `withIdAndName()`. The normal path goes through `.map()` which correctly sets id/name, but timeout errors bypass `.map()`.

This causes `PostActingEvent.getToolResult()` to return a `ToolResultBlock` with null id and name, while `getToolResultMsg()` has the correct values (rebuilt from `ToolUseBlock` by `ToolResultMessageBuilder`).

Fix: chain `.withIdAndName(toolCall.getId(), toolCall.getName())` in the `onErrorResume` handler.

Fixes #1389


LearningGp (approved):
LGTM

PR #1445, opened 2026-05-20
fix(nacos-skill): handle block-sequence arrays in SKILL.md frontmatter

### LINKED ISSUE #1438: [Bug]:SKILL.md 元数据区通过多行定义数组类型数据，执行技能是报错：sequence entries are not allowed here  in 'string', line 6, column 18:

**Describe the bug**
SKILL.md 元数据区通过多行定义数组类型数据，执行技能是报错：sequence entries are not allowed here  in 'string', line 6, column 18:

**To Reproduce**

Image

看了这个方法的注释，AgentScope's frontmatter parser only accepts flat key: value lines，是不是这里就会把格式去除了
运行时定义在元数据区的数组格式被变成了一行，导致-校验不通过
**Expected behavior**

**Error messages**
sequence entries are not allowed here  in 'string', line 6, column 18:

- AgentScope-Java Version: 1.1.0-RC2
- Java Version: 21
- OS:windows

**Additional context**

sunwg2:
Nacos exports use multi-line `- item` syntax for array fields. The old normalizer concatenated items into an invalid scalar, causing SnakeYAML to throw "sequence entries are not allowed here in 'string'".

Collect `- item` lines into a list and emit them as an inline YAML array `["a", "b"]`; strip YAML quoting from raw items via unquoteYamlString().

Closes #1438

## AgentScope-Java Version

## Description

Fix #1438 — `NacosSkillRepository` crashes with
`sequence entries are not allowed here in 'string'` when a `SKILL.md`
exported from Nacos uses YAML **block-sequence syntax** for array fields
such as `trigger_intents`.

`normalizeFoldedFlatYaml()` was designed to fold indented continuation
lines back into the previous key's value. It had no awareness of YAML
block-sequence entry lines (`- item`), so it treated them as plain
continuation text and concatenated all items into a single invalid scalar:

## Checklist


LearningGp:
image

sunwg2:
recheck

agentscope-extensions/agentscope-extensions-nacos/agentscope-extensions-nacos-skill/src/main/java/io/agentscope/core/nacos/skill/NacosSkillRepository.java
```diff
@@ -451,6 +491,34 @@ private static boolean yamlValueNeedsQuoting(String value) {
...
+     */
+    private static String unquoteYamlString(String value) {
+        if (value == null || value.length() < 2) {
+            return value;
+        }
+        char first = value.charAt(0);
+        char last = value.charAt(value.length() - 1);
+        if (first == '"' && last == '"') {
+            String inner = value.substring(1, value.length() - 1);
+            return inner.replace("\\\"", "\"").replace("\\\\", "\\");
```
AgentScopeJavaBot, line 560: [nit] The double-quoted unescape in `unquoteYamlString` only handles `\"` and `\\`. Other valid YAML double-quoted escape sequences (`\n`, `\t`, `\uXXXX`, etc.) pass through as literal characters. The Javadoc should note this is an intentional subset.

agentscope-extensions/agentscope-extensions-nacos/agentscope-extensions-nacos-skill/src/test/java/io/agentscope/core/nacos/skill/NacosSkillRepositoryTest.java
```diff
@@ -418,6 +419,123 @@ void testDelete() {
         assertFalse(repository.delete("any-skill"));
     }
 
+    // -------- Bug #1438: multi-line block-sequence array in frontmatter --------
```
AgentScopeJavaBot, line 422: [nit] The test suite covers double-quoted and unquoted list items but has no test for single-quoted values (e.g., `- 'value with ''apostrophe'''`). Adding a test case would exercise the single-quote branch of `unquoteYamlString`.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes Bug #1438 by adding support for YAML block-sequence arrays (`- item`) in SKILL.md frontmatter within the Nacos skill repository. When `NacosSkillRepository` normalizes Nacos-exported YAML for the downstream `MarkdownSkillParser`, block-sequence entries under a key with no inline value are now collected into a list and serialized as an inline YAML array. The implementation is clean and well-scoped with four integration tests covering multi-item, single-item, special-character, and mixed scalar/array scenarios. No critical or blocking issues were found.

agentscope-extensions/agentscope-extensions-nacos/agentscope-extensions-nacos-skill/src/main/java/io/agentscope/core/nacos/skill/NacosSkillRepository.java
sunwg2: Good point. Updated the Javadoc to explicitly note that only \" and \\ are unescaped — other YAML double-quoted escape sequences (\n, \t, unicode escapes, etc.) pass through as literal characters, and this is an intentional subset covering the Nacos export format.

agentscope-extensions/agentscope-extensions-nacos/agentscope-extensions-nacos-skill/src/test/java/io/agentscope/core/nacos/skill/NacosSkillRepositoryTest.java
sunwg2: Good catch. Added testGetSkillWithSingleQuotedArrayItems to cover the single-quote branch of unquoteYamlString, including a case with '' (escaped apostrophe) to verify it unescapes correctly to '.

oss-maintainer:

```bash
git fetch origin
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-extensions/agentscope-extensions-nacos/agentscope-extensions-nacos-skill/src/main/java/io/agentscope/core/nacos/skill/NacosSkillRepository.java
```diff
@@ -427,26 +433,60 @@ private static String normalizeFoldedFlatYaml(List<String> yamlLines) {
...
+            } else {
+                Matcher listMatcher = YAML_LIST_ITEM.matcher(t);
+                if (listMatcher.matches()
+                        && pendingKey != null
+                        && (pendingVal == null || pendingVal.isEmpty())) {
+                    // Block-sequence entry under a key with no inline value: collect as list.
+                    if (pendingList == null) {
+                        pendingList = new ArrayList<>();
+                    }
+                    pendingList.add(unquoteYamlString(listMatcher.group(1).trim()));
```
oss-maintainer, line 449: **[Warning]** Items are collected only while `pendingVal` is empty, but nothing clears or rejects the folded-scalar path afterwards: if a key gets a list item and *then* a plain continuation line, the continuation is written into `pendingVal` while `pendingList` is non-empty, and `flushYamlKvLine` returns early from the list branch — the folded text is silently dropped. Consider logging when both are populated (or rejecting one of them) instead of dropping content, and add a test for `key:` + `- a` + `trailing text`.

agentscope-extensions/agentscope-extensions-nacos/agentscope-extensions-nacos-skill/src/main/java/io/agentscope/core/nacos/skill/NacosSkillRepository.java
```diff
@@ -427,26 +433,60 @@ private static String normalizeFoldedFlatYaml(List<String> yamlLines) {
...
                 pendingVal = m.group(2).trim();
-            } else if (pendingKey != null) {
-                if (pendingVal == null || pendingVal.isEmpty()) {
-                    pendingVal = t;
-                } else {
-                    pendingVal = pendingVal + " " + t;
+                pendingList = null;
+            } else {
+                Matcher listMatcher = YAML_LIST_ITEM.matcher(t);
+                if (listMatcher.matches()
```
oss-maintainer, line 442: **[Warning]** This cannot distinguish a block sequence from a *nested mapping* item: a line like `- name: foo` matches `YAML_LIST_ITEM`, and the captured text is then unquoted and re-emitted as the string `"name: foo"`, i.e. a list of mappings silently becomes a list of strings. If the goal is to rescue flat lists only, require the captured group to have no `key:` shape (or bail out of list mode when a `- ` line contains an unquoted `: `) so the frontmatter is left untouched rather than being mangled into a different YAML type.

agentscope-extensions/agentscope-extensions-nacos/agentscope-extensions-nacos-skill/src/main/java/io/agentscope/core/nacos/skill/NacosSkillRepository.java
```diff
@@ -81,6 +81,9 @@ public class NacosSkillRepository implements AgentSkillRepository {
     private static final Pattern YAML_KV =
             Pattern.compile("^([a-zA-Z_][a-zA-Z0-9_-]*)\\s*:\\s*(.*)$");
 
+    /** Matches a YAML block-sequence entry line, e.g. "  - some value" or "- item". */
+    private static final Pattern YAML_LIST_ITEM = Pattern.compile("^-\\s+(.*)$");
```
oss-maintainer, line 85: **[Info]** `^-\s+(.*)$` also requires at least one whitespace character, so `-[no space]` is not matched (that is valid YAML for a sequence entry) and a bare `-` line falls through to the folded-text branch. Indentation is also discarded before matching (`t = raw.trim()`), so a sequence nested under a *deeper* key is attributed to whatever `pendingKey` was last seen. Both are edge cases; worth a comment stating that only flat, single-level sequences are supported by design.

oss-maintainer (commented):
### Summary
Real bug and a well-scoped fix: block-sequence frontmatter arrays (`allowed-tools:` style lists) exported by Nacos were folded into a single scalar string, which `MarkdownSkillParser`/SnakeYAML then read as a string instead of a list. Converting them to an inline flow array with proper double-quote escaping is the right shape for this folded-normalisation pass, and the added tests cover the plain, quoted, and mixed cases.

Three refinements before this is ready, plus a rebase (`CONFLICTING` against `main`).

### Findings
- **[Warning]** NacosSkillRepository.java:449 — a folded continuation line arriving after list items is silently dropped by the list branch in `flushYamlKvLine`.
- **[Warning]** NacosSkillRepository.java:442 — `- key: value` (a sequence of mappings) is rewritten as a sequence of strings, changing the parsed type instead of leaving it alone.
- **[Info]** NacosSkillRepository.java:85 — `-item` without a space is unmatched, and trimming before matching means nesting depth is not respected.

### Suggestions
Guard the list mode so it only applies to plain scalars, and never lose text:
```java
String item = listMatcher.group(1).trim();
if (item.contains(": ")) {
    // not a flat sequence entry — fall back to the existing folded behaviour
    pendingList = null;
} else {
    if (pendingList == null) pendingList = new ArrayList<>();
    pendingList.add(unquoteYamlString(item));
}
```
and in `flushYamlKvLine`, log (or append) when both `list` and `value` are populated instead of returning early.

### Notes
CLA status for the current head: `success`. This PR reports `CONFLICTING`/`DIRTY` against `main` (rebase reminder already posted earlier) — please rebase before the next round.

---
*Automated review by @BOT_SIG*@

PR #1494, opened 2026-05-26
feat: 支持通过 GenerateOptions 按轮次动态控制 enableThinking

Janemia:
## 改动说明
为 `GenerateOptions` 新增 `enableThinking` 字段，使 Hook 能够按轮次动态控制 `enable_thinking`（此前只能控制 `thinking_budget`）。

## 修改内容
- `GenerateOptions`：新增 `enableThinking` 字段、getter、Builder 方法及 mergeOptions 合并逻辑
- `DashScopeChatModel.applyThinkingMode()`：优先取 options 级别的值，未设置时 fallback 到模型级别的固定值

## 向后兼容
完全向后兼容。当 `options.getEnableThinking()` 为 null 时，行为与改造前一致，使用模型级别的 `this.enableThinking`。


LearningGp:
image

agentscope-core/src/main/java/io/agentscope/core/model/DashScopeChatModel.java
```diff
@@ -321,22 +321,27 @@ private Flux<ChatResponse> streamWithHttpClient(
...
+                        + " enableThinking(true) in GenerateOptions. Example:"
                         + " DashScopeChatModel.builder().enableThinking(true)"
                         + ".defaultOptions(GenerateOptions.builder().thinkingBudget(1000).build())");
         }
 
-        if (enableThinking != null) {
-            // Explicitly assign value for thinking mode
-            request.getParameters().setEnableThinking(enableThinking);
+        if (effectiveEnableThinking != null) {
+            request.getParameters().setEnableThinking(effectiveEnableThinking);
```
AgentScopeJavaBot, line 351: [major] Per-request `enableThinking=true` does not propagate to streaming. `this.stream` is finalized in the constructor from the model-level `enableThinking` only (line 151: `this.stream = enableThinking != null && enableThinking ? true : stream`). When a hook turns thinking on via `options.enableThinking(true)` but the model was built with `stream=false` and no model-level thinking, the request body sets `enable_thinking=true` while the call path still uses the non-streaming branch (line 286 `if (stream)`), violating the in-class invariant 'thinking mode requires streaming' that the constructor explicitly logs about. This silently breaks the feature this PR advertises. Either (a) make `stream` resolution dynamic per request when `effectiveEnableThinking` is true, or (b) document and validate that per-request enabling only works when the model was already built with `stream=true`.

agentscope-core/src/main/java/io/agentscope/core/model/DashScopeChatModel.java
```diff
@@ -321,22 +321,27 @@ private Flux<ChatResponse> streamWithHttpClient(
...
+        if (options.getThinkingBudget() != null && !Boolean.TRUE.equals(effectiveEnableThinking)) {
             throw new IllegalStateException(
                     "thinkingBudget is set but enableThinking is not enabled. To use thinking mode"
                         + " with budget control, you must explicitly enable thinking by calling"
-                        + " .enableThinking(true) on the model builder. Example:"
+                        + " .enableThinking(true) on the model builder or setting"
+                        + " enableThinking(true) in GenerateOptions. Example:"
                         + " DashScopeChatModel.builder().enableThinking(true)"
                         + ".defaultOptions(GenerateOptions.builder().thinkingBudget(1000).build())");
         }
```
AgentScopeJavaBot: [minor] Validation message is misleading for the new inverse case. If defaults set `thinkingBudget(1000)` and a hook explicitly sets `enableThinking(false)` per-request, `effectiveOptions` (already merged) carries both, so this check throws telling the user to 'explicitly enable thinking' — even though the user just explicitly disabled it. Consider either ignoring `thinkingBudget` silently when `effectiveEnableThinking == Boolean.FALSE` (treat it as a no-op consistent with the user's intent), or refining the message to distinguish 'null thinking + budget' from 'explicitly disabled + budget'.

agentscope-core/src/main/java/io/agentscope/core/model/GenerateOptions.java
```diff
@@ -666,6 +684,20 @@ public Builder presencePenalty(Double presencePenalty) {
...
+         * Sets whether thinking mode is enabled for this request.
+         *
+         * <p>When set, this overrides the model-level {@code enableThinking} configuration,
+         * allowing per-request control over thinking mode via hooks.
+         *
+         * @param enableThinking true to enable thinking, false to disable, null to use model default
+         * @return this builder
+         */
+        public Builder enableThinking(Boolean enableThinking) {
+            this.enableThinking = enableThinking;
```
AgentScopeJavaBot, line 716: [minor] Builder Javadoc says 'allowing per-request control over thinking mode via hooks'. The mechanism is generic to any caller that passes a per-request `GenerateOptions`, not only hooks — wording should drop the 'via hooks' qualifier (or mention hooks as one example) to match the symmetric phrasing of other builder methods such as `thinkingBudget`.

agentscope-core/src/main/java/io/agentscope/core/model/GenerateOptions.java
```diff
@@ -451,6 +466,8 @@ public static GenerateOptions mergeOptions(GenerateOptions primary, GenerateOpti
                 primary.presencePenalty != null
                         ? primary.presencePenalty
                         : fallback.presencePenalty);
+        builder.enableThinking(
+                primary.enableThinking != null ? primary.enableThinking : fallback.enableThinking);
```
AgentScopeJavaBot, line 486: [nit] No unit tests cover the new merge behavior or per-request override path. `DashScopeChatModelTest` and `GenerateOptionsTest` both have existing thinking-mode coverage but nothing exercises `options.enableThinking(false)` overriding a model built with `enableThinking(true)`, nor the merge precedence in `mergeOptions`. Adding two small tests would lock in the documented contract.

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR cleanly adds a per-request `enableThinking` override on `GenerateOptions` and wires it into `DashScopeChatModel.applyThinkingMode`, falling back to the model-level value when null. The merge logic and Javadocs are consistent with the rest of the file. However, the change introduces a behavioral gap: the model's `stream` flag (and the "thinking requires streaming" override) is resolved only at construction time from the model-level `enableThinking`. When a hook turns thinking on per-request while the model was built with `stream=false` and `enableThinking=null/false`, the request is sent in non-streaming mode, which the existing code explicitly warns is invalid. A minor symmetry issue exists for the inverse case (per-request `enableThinking=false` combined with a default `thinkingBudget`), which now produces a confusing validation error. The change is also untested.

agentscope-core/src/main/java/io/agentscope/core/model/DashScopeChatModel.java
AgentScopeJavaBot: ✅ Verified as addressed in `3d752e6`: Added explicit validation (IllegalStateException) when per-request enableThinking=true but model stream=false, implementing option (b). Test testApplyThinkingModePerRequestRequiresStreaming confirms this.

agentscope-core/src/main/java/io/agentscope/core/model/DashScopeChatModel.java
AgentScopeJavaBot: ✅ Verified as addressed in `3d752e6`: Validation now checks effectiveEnableThinking==null instead of !TRUE, so explicitly-false-with-budget silently ignores the budget per user intent. Test testApplyThinkingModeExplicitlyDisabledWithBudget confirms this.

agentscope-core/src/main/java/io/agentscope/core/model/GenerateOptions.java
AgentScopeJavaBot: ✅ Verified as addressed in `3d752e6`: Javadoc now says allowing per-request control over thinking mode without the via hooks qualifier.

agentscope-core/src/main/java/io/agentscope/core/model/GenerateOptions.java
AgentScopeJavaBot: ✅ Verified as addressed in `3d752e6`: Five new tests added covering per-request override, streaming validation, merge precedence, and fallback behavior.

AgentScopeJavaBot (approved):

LearningGp:
Run 'mvn spotless:apply' to fix these violations.

PR #1508, opened 2026-05-27
fix(agent): complete stream when result is empty

### LINKED ISSUE #1503: [Bug]:ReActAgent.stream(...) never completes when the model returns empty assistant content

**Describe the bug**

When a model returns an assistant message with empty text content and no tool calls, `ReActAgent.stream(...)` may never complete.

In this case, the model response is effectively:

```json
[
  {
    "role": "assistant",
    "content": ""
  }
]
```

The stream does not emit a final result or complete normally, so downstream integrations waiting for stream completion can hang indefinitely.

**To Reproduce**

1. Use a model implementation or mock model that streams exactly one assistant response with an empty `TextBlock` and no `ToolUseBlock`.

2. Call `ReActAgent.stream(...)`, for example:

```java
        .name("test-agent")
        .model(modelThatReturnsEmptyAssistantContent)
        .build();

agent.stream(List.of(
        Msg.builder()
                .role(MsgRole.USER)
                .textContent("hello")
                .build()
)).blockLast();
```

3. The stream does not complete. In an SSE integration, the final `RUN_FINISHED` or equivalent completion event is never sent.

**Expected behavior**

The stream should complete even if the model returns an empty assistant response.

Possible expected behaviors:

- Return an assistant `Msg` with empty text content and complete the stream.
- Or complete the stream without a final message.
- Or throw a clear exception indicating that the model returned an invalid empty response.

In all cases, `Agent.stream(...)` should not hang indefinitely.

**Error messages**

There is no exception or error message.

The observable behavior is that the reactive stream never completes.

From source inspection, the possible cause seems to be:

- `TextAccumulator.hasContent()` returns `false` for empty text because it checks `accumulated.length() > 0`.
- `ReasoningContext.buildFinalMessage()` returns `null` when there is no text, thinking content, or tool call.
- `ReActAgent.reasoning()` uses `Mono.justOrEmpty(context.buildFinalMessage())`, which becomes `Mono.empty()`.
- `AgentBase.createEventStream()` only calls `sink.complete()` in the `onNext` callback of the final message subscription. If the `Mono` is empty, no `onNext` is invoked and the `FluxSink` is never completed.

- AgentScope-Java Version: 1.1.0-RC1
- Java Version: 21
- OS: Windows 10

**Additional context**

This issue affects integrations that rely on `Agent.stream(...)` completion, such as SSE or AG-UI style adapters. If the upstream model returns an empty assistant message, the client keeps waiting because the server-side stream never reaches its normal finish logic.

JGoP-L:
## AgentScope-Java Version

1.1.0-SNAPSHOT

## Description

This PR fixes an edge case where `ReActAgent.stream(...)` can hang when the model returns an empty assistant text response.

Previously, `AgentBase.createEventStream()` completed the outer `FluxSink` only inside the `onNext(finalMsg)` callback. When the agent call completed as `Mono.empty()`, `onNext` was not invoked, so `sink.complete()` was never called and consumers such as `stream().blockLast()` could wait indefinitely.

Changes made:
- Complete the stream from the upstream `Mono` completion callback.
- Preserve existing behavior for normal final messages and error propagation.
- Add a regression test for an empty assistant text response.

Test command:
    mvn -pl agentscope-core test

Local result:
    BUILD SUCCESS

Fixes #1503

## Checklist

- [x] All relevant tests are passing (`mvn -pl agentscope-core test`)
- [x] Related documentation has been updated (not applicable: internal stream lifecycle bug fix covered by regression test)


agentscope-core/src/test/java/io/agentscope/core/agent/AgentStreamingTest.java
```diff
@@ -248,4 +251,38 @@ void testStreamEventCount() {
...
+
+        Msg inputMsg =
+                Msg.builder()
+                        .name("user")
+                        .role(MsgRole.USER)
+                        .content(List.of(TextBlock.builder().text("hello").build()))
+                        .build();
+
+        StepVerifier.create(agent.stream(List.of(inputMsg)))
+                .expectNextMatches(
```
AgentScopeJavaBot, line 279: [nit] The test relies on at least one `REASONING` event being emitted by `StreamingHook` before completion. Per the issue analysis, an empty `TextBlock` does not satisfy `TextAccumulator.hasContent()`, so this assumption depends on hook implementation details (e.g. the start-of-reasoning event still firing). Consider making the assertion more defensive, e.g. `.thenConsumeWhile(event -> true).expectComplete()` without the upfront `expectNextMatches`, since the fix's contract is purely about completion, not about which events fire. As written the test couples the regression assertion to the StreamingHook event sequence and may become flaky if hook emission semantics change.

agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
```diff
@@ -936,11 +936,9 @@ private Flux<Event> createEventStream(StreamOptions options, Supplier<Mono<Msg>>
                                                                                     finalMsg,
                                                                                     true));
                                                                 }
-
-                                                                // Complete the stream
-                                                                sink.complete();
                                                             },
-                                                            sink::error);
+                                                            sink::error,
+                                                            sink::complete);
```
AgentScopeJavaBot, line 941: [minor] Consider also emitting a synthetic `AGENT_RESULT` (or at least a placeholder empty assistant `Msg`) when the upstream `Mono` completes empty, so downstream SSE / AG-UI adapters that key off `AGENT_RESULT` as the terminal event still receive a final marker. The current fix correctly unblocks completion, but consumers that loop until they observe an `AGENT_RESULT` will now silently terminate without one. If keeping the current behavior intentionally, please document it in the Javadoc of `createEventStream` (e.g. "if the agent call produces no final message, the stream completes without emitting an AGENT_RESULT event").

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a real bug (issue #1503): `AgentBase.createEventStream()` previously called `sink.complete()` only inside the `onNext` callback, so when the upstream `Mono<Msg>` completed empty (e.g. ReAct reasoning context produced no final message because the model returned empty text), the outer `FluxSink` was never completed and consumers like `stream().blockLast()` hung forever. The fix moves `sink.complete()` to the dedicated `onComplete` overload of `subscribe(onNext, onError, onComplete)`, which fires for both value-producing and empty Monos. The change is minimal, semantically correct in Reactor (onNext still runs before onComplete for non-empty Monos, so AGENT_RESULT is emitted before completion as before), and covered by a dedicated regression test using a `MockModel` that returns an empty `TextBlock`. Behavior change worth flagging: when the model returns empty content, no `AGENT_RESULT` event is emitted at all (the stream just completes); downstream SSE/AG-UI integrations that rely on a terminal `AGENT_RESULT` event will now see only completion. This matches one of the acceptable outcomes documented in the issue, but is a behavioral note callers should be aware of.

LearningGp (approved):
LGTM

PR #1512, opened 2026-05-28
fix(hook): PendingToolRecoveryHook skips patching when memory has historical ToolResults

Xingxing-Wang:
## Summary

`PendingToolRecoveryHook.handlePreCall` checks `event.getInputMessages()` for `ToolResultBlock`s to decide whether the user is providing tool results (HITL/Clarify scenario). However, `event.getInputMessages()` includes the **full memory snapshot** + user callArgs (constructed in `AgentBase.notifyPreCall`). 

Any historical `ToolResultBlock` from previous tool calls in memory causes `userProvidedResults = true`, silently skipping the auto-recovery patch. This makes the hook ineffective in **any conversation that has previously used tools** — which is essentially all real-world conversations.

### Root Cause

```java
// Before (bug): scans full memory history + user input
List<Msg> inputMessages = event.getInputMessages(); // = memory snapshot + callArgs
boolean userProvidedResults =
// → always true when memory has historical tool results → skip patch
```

### Fix

Extract only the callArgs portion (beyond the memory snapshot boundary) before checking:

```java
// After (fix): scans only user's current input
int memorySize = memory.getMessages().size();
List<Msg> callArgs = inputMessages.subList(memorySize, inputMessages.size());
boolean userProvidedResults =
    callArgs.stream().anyMatch(m -> m.hasContentBlocks(ToolResultBlock.class));
```

### Reproduction

1. Conversation with 250+ historical tool calls (memory contains many `ToolResultBlock`s)
2. Agent calls a new tool → execution crashes/interrupts → orphan `ToolUseBlock` in memory
3. User retries → hook detects pending IDs, but `userProvidedResults` is `true` due to historical results → **skips patch**
4. `ReActAgent.doCall` throws `IllegalStateException: Pending tool calls exist without results`
5. Error saved to MySQL → same state on next retry → **infinite error loop**

## Test plan

- [x] Added `testAutoRecoveryWithHistoricalToolResults` — pre-populates memory with historical tool call + result pairs, then verifies the hook still patches pending tool calls
- [x] All 17 existing `HookStopAgentTest` tests pass (no regression)

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
@@ -98,17 +98,25 @@ public class PendingToolRecoveryHook implements Hook {
             return Mono.just(event);
         }
 
-        // Check if user already provided tool results in the input
+        // event.getInputMessages() = memory snapshot + user callArgs
+        // (see AgentBase.notifyPreCall). Extract just the callArgs portion so we only
+        // check the user's current input for ToolResultBlocks, not the entire memory
+        // history which naturally contains ToolResultBlocks from previous tool calls.
         List<Msg> inputMessages = event.getInputMessages();
+        int memorySize = memory.getMessages().size();
+        List<Msg> callArgs =
+                (inputMessages != null && inputMessages.size() > memorySize)
+                        ? inputMessages.subList(memorySize, inputMessages.size())
+                        : List.of();
 
-        // If input is empty/null, the user is resuming (wants to continue acting).
+        // If callArgs is empty, the user is resuming (wants to continue acting).
         // Do NOT patch — let ReActAgent's doCall handle the resume flow.
-        if (inputMessages == null || inputMessages.isEmpty()) {
+        if (callArgs.isEmpty()) {
             return Mono.just(event);
         }
 
         boolean userProvidedResults =
-                inputMessages.stream().anyMatch(m -> m.hasContentBlocks(ToolResultBlock.class));
+                callArgs.stream().anyMatch(m -> m.hasContentBlocks(ToolResultBlock.class));
         if (userProvidedResults) {
             return Mono.just(event);
         }
--- a/agentscope-core/src/test/java/io/agentscope/core/hook/HookStopAgentTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/hook/HookStopAgentTest.java
@@ -420,6 +420,101 @@ class HookStopAgentTest {
                     "Memory should contain an error ToolResultBlock for the pending tool call"
                             + " id='tool1'");
         }
+
+        @Test
+        @DisplayName("Auto-recovery works when memory already contains historical ToolResultBlocks")
+        void testAutoRecoveryWithHistoricalToolResults() {
+            // Pre-populate memory with historical tool call + result pairs.
+            // This simulates a conversation that has used tools before.
+            Msg historicalToolUse = createToolUseMsg("old_tool_1", "odps_query", Map.of());
+            memory.addMessage(historicalToolUse);
+            Msg historicalToolResult =
+                    Msg.builder()
+                            .name("test-agent")
+                            .role(MsgRole.TOOL)
+                            .content(
+                                    ToolResultBlock.builder()
+                                            .id("old_tool_1")
+                                            .output(
+                                                    List.of(
+                                                            TextBlock.builder()
+                                                                    .text("query result")
+                                                                    .build()))
+                                            .build())
+                            .build();
+            memory.addMessage(historicalToolResult);
+            Msg historicalAssistantReply = createAssistantTextMsg("Here is the query result.");
+            memory.addMessage(historicalAssistantReply);
+
+            // Now set up the current scenario: agent makes a new tool call that gets stopped
+            Msg newToolUseMsg = createToolUseMsg("new_tool_1", "test_tool", Map.of());
+            Msg recoveryResponse = createAssistantTextMsg("Recovered successfully");
+
+            when(mockModel.stream(anyList(), anyList(), any()))
+                    .thenReturn(createFluxFromMsg(newToolUseMsg))
+                    .thenReturn(createFluxFromMsg(recoveryResponse));
+
+            Hook stopHook = createPostReasoningStopHook();
+
+            ReActAgent agent =
+                    ReActAgent.builder()
+                            .name("test-agent")
+                            .model(mockModel)
+                            .toolkit(toolkit)
+                            .memory(memory)
+                            .checkRunning(false)
+                            .hook(stopHook)
+                            .enablePendingToolRecovery(true)
+                            .build();
+
+            // First call - agent reasons a tool call, gets stopped by hook
+            Msg result1 = agent.call(createUserMsg("do something")).block(TEST_TIMEOUT);
+            assertTrue(
+                    result1.hasContentBlocks(ToolUseBlock.class),
+                    "First call should return ToolUse message (stopped by hook)");
+
+            // At this point, memory has:
+            //   - historical ASSISTANT msg with ToolUseBlock(id=old_tool_1)
+            //   - historical TOOL msg with ToolResultBlock(id=old_tool_1) ← THIS IS THE KEY
+            //   - historical ASSISTANT text msg
+            //   - user msg
+            //   - new ASSISTANT msg with ToolUseBlock(id=new_tool_1) ← PENDING (no result)
+            //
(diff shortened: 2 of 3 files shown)


AgentScopeJavaBot (commented):
🤖 AI Review

本 PR 修复了 `PendingToolRecoveryHook.handlePreCall` 中一个真实且影响范围较大的 bug：原实现直接在完整的 `event.getInputMessages()`（= memory 快照 + 用户 callArgs）上扫描 `ToolResultBlock`，导致只要历史 memory 中存在任何已完成的工具调用结果，就会被误判为"用户已主动提供 tool result"，从而跳过 auto-recovery 补丁，最终在 `ReActAgent.doCall` 抛出 `IllegalStateException`。修复思路（用 `memory.getMessages().size()` 作为分界点，只校验 callArgs 段）方向正确，新增的单元测试与集成测试覆盖了"历史 ToolResult 仍能 patch"、"resume 场景跳过"、"用户主动提供结果时跳过"三种核心路径，质量较高。

主要顾虑在于实现层耦合：hook 通过运行时再次读取 `memory.getMessages().size()` 来推断 callArgs 边界，隐含假设"从 `AgentBase.notifyPreCall` 拍快照到当前 hook 执行之间，memory 与 inputMessages 都未被改动"。这一假设在引入用户自定义高优先级 hook 后会被打破，建议后续把 callArgs/snapshot boundary 显式暴露在 `PreCallEvent` 上以消除隐式耦合。当前改动可以合入，但建议在 follow-up 中加固契约。

agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
```diff
@@ -98,17 +98,25 @@ private Mono<PreCallEvent> handlePreCall(PreCallEvent event) {
             return Mono.just(event);
         }
 
-        // Check if user already provided tool results in the input
+        // event.getInputMessages() = memory snapshot + user callArgs
+        // (see AgentBase.notifyPreCall). Extract just the callArgs portion so we only
+        // check the user's current input for ToolResultBlocks, not the entire memory
+        // history which naturally contains ToolResultBlocks from previous tool calls.
         List<Msg> inputMessages = event.getInputMessages();
+        int memorySize = memory.getMessages().size();
```
AgentScopeJavaBot, line 106: [minor] The callArgs boundary is recomputed here from live memory (`memory.getMessages().size()`), while `inputMessages` was built from a snapshot taken at the start of `AgentBase.notifyPreCall`. This works today because no built-in hook with priority < 10 mutates memory or replaces `inputMessages` in `PreCallEvent`, but it bakes in an implicit invariant: a user-defined hook in the 0-9 priority range (the `Critical system hooks` band documented on `Hook.priority()`) that calls `memory.addMessage(...)` will silently shift the boundary, causing either the first N real callArgs to be dropped (when `memory.size() > snapshotSize`) or the tail of the original snapshot to be misclassified as callArgs (when memory was truncated) — re-introducing the very class of bug this PR fixes. Consider exposing the snapshot boundary on `PreCallEvent` directly so the contract cannot be invalidated by sibling hooks:

```java
// In AgentBase.notifyPreCall, populate alongside the existing snapshotSize:
PreCallEvent event = new PreCallEvent(this, fullInput, snapshotSize);

// Hook side becomes ordering-independent:
int callArgsStart = event.getCallArgsStartIndex();
List<Msg> callArgs = inputMessages.subList(callArgsStart, inputMessages.size());
```

agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
```diff
@@ -98,17 +98,25 @@ private Mono<PreCallEvent> handlePreCall(PreCallEvent event) {
...
+        List<Msg> callArgs =
+                (inputMessages != null && inputMessages.size() > memorySize)
```
AgentScopeJavaBot, line 108: [nit] The `inputMessages != null` guard is unreachable: `PreCallEvent`'s constructor wraps the input in `new ArrayList<>(Objects.requireNonNull(inputMessages, ...))` and `setInputMessages` also calls `Objects.requireNonNull`, so `event.getInputMessages()` cannot return `null`. The branch can collapse to `inputMessages.size() > memorySize ? inputMessages.subList(memorySize, inputMessages.size()) : List.of()`. Pure readability nit — not blocking.

agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
```diff
@@ -98,17 +98,25 @@ private Mono<PreCallEvent> handlePreCall(PreCallEvent event) {
...
+                        ? inputMessages.subList(memorySize, inputMessages.size())
+                        : List.of();
```
AgentScopeJavaBot, line 110: [minor] The `inputMessages.size() > memorySize` check is strict-greater, so the `inputMessages.size() < memorySize` case (memory grew, or a preceding hook shortened `inputMessages`) silently degrades to `callArgs = List.of()` and the patch is skipped without any signal. That state is unexpected and almost certainly indicates a hook-ordering bug — emitting a `log.warn` (or even an assertion) when `inputMessages.size() < memorySize` would make such regressions diagnosable instead of manifesting as silent recovery failures downstream.

agentscope-core/src/test/java/io/agentscope/core/hook/PendingToolRecoveryHookTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+        PreCallEvent event = new PreCallEvent(agent, fullInput);
+        hook.onEvent(event).block();
+
+        // Verify the hook patched: memory should now have a ToolResultBlock for "pending_call"
+        boolean patched =
+                memory.getMessages().stream()
+                        .flatMap(m -> m.getContentBlocks(ToolResultBlock.class).stream())
+                        .anyMatch(tr -> "pending_call".equals(tr.getId()));
+        assertTrue(patched, "Hook should patch the pending tool call despite historical results");
```
AgentScopeJavaBot, line 78: [minor] `patchesDespiteHistoricalToolResults` only asserts that a `ToolResultBlock` for `pending_call` exists in memory after the hook runs. Tightening the assertion to also check that the output text starts with the `[ERROR]` prefix (matching the contract enforced in `buildErrorToolResult` and already asserted in `HookStopAgentTest`) would catch silent regressions in the recovery payload format, e.g. if someone later refactors `buildErrorToolResult` without updating both call sites.

agentscope-core/src/test/java/io/agentscope/core/hook/HookStopAgentTest.java
```diff
@@ -420,6 +420,101 @@ void testNewMsgWithPendingToolUseContinuesActing() {
                     "Memory should contain an error ToolResultBlock for the pending tool call"
                             + " id='tool1'");
```
AgentScopeJavaBot, line 421: [praise] `testAutoRecoveryWithHistoricalToolResults` reproduces the production failure faithfully (historical tool call + result pairs, then a fresh pending `ToolUseBlock`, then a second `call`) and validates the fix through the full `ReActAgent` path rather than the hook in isolation. The inline comment block explaining the bug scenario also makes the regression contract very clear to future readers.

AgentScopeJavaBot (commented):
🤖 AI Review

The fix targets a real and high-impact bug: `PendingToolRecoveryHook.handlePreCall` was scanning the full `event.getInputMessages()` (memory snapshot + user callArgs) for `ToolResultBlock`s, so any conversation that had ever invoked a tool produced a false positive on `userProvidedResults` and silently skipped the auto-recovery patch, leaving an orphaned `ToolUseBlock` that later crashed `ReActAgent.doCall` with `IllegalStateException`. The new code slices off the callArgs portion using `memory.getMessages().size()` as the snapshot boundary, which is the correct shape for the bug. Test coverage is strong: a new `PendingToolRecoveryHookTest` exercises the three boundary behaviours (historical results present, callArgs empty / resume, user-provided result), and the end-to-end `testAutoRecoveryWithHistoricalToolResults` in `HookStopAgentTest` reproduces the production crash path. The main architectural concern (also raised by the existing AI review) is the implicit coupling: the snapshot boundary is recomputed from live memory at hook-execution time, which depends on no higher-priority hook having mutated memory or input length in between. Safe for all built-in hooks today, but worth hardening with an explicit boundary on `PreCallEvent` in a follow-up.

chickenlj:
Thanks for the contribution! This has been superseded by #1566 which provides a more thorough fix for the PendingToolRecoveryHook false positive issue.

PR #1514, opened 2026-05-28
feat(autocontext-memory): add recent focus window rollup with active tool guard

### LINKED ISSUE #1482: AutoContextMemory导致工具重复调用，直到maxIters才结束，触发压缩后有概率出现这个问题
以下为AI分析结果：
两个高风险点：AutoContextHook 会重写 PreReasoningEvent 的输入消息；AutoContextMemory 多处注释提到“保留 ToolUseBlock 结构”。如果压缩后的上下文记忆仍带历史 ToolUseBlock， ReAct 循环很可能把它当作新的待执行工具请求。

更具体的线索：大消息摘要路径会把 ToolUseBlock 原样放回压缩后的 message，注释还明确说“preserve ToolUseBlock structure”。这对 API 消息格式也许是为了配对工具结果，但对 ReAct的“是否有工具调用需要执行”的判定很危险。

工具执行后会立刻进入下一次 reasoning，而 AutoContextHook 在这次 reasoning 前调用压缩。也就是说压缩可能发生在“工具结果刚写入 memory、模型还没基于这个 tool_result生成最终回答”之间；如果这时把当前轮 tool_use/tool_result 压成普通摘要，模型很容易把它当作历史叙述而不是刚返回的工具结果，从而再次请求同一个工具。

问题根因在 AutoContextHook 的 PreReasoningEvent：工具执行结果刚写入 memory 后，下一轮 reasoning 前会触发压缩。原实现直接走 compressIfNeeded()，可能把“当前轮尚未被模型消费的tool_use/tool_result”压成普通摘要，模型看不到结构化工具结果，就容易再次调用同一个工具。

lzbjut:
## Summary
Implements a recency-weighted AutoContextMemory compression path for #1482.

- Protects in-flight tool rounds from current-round compression before tool results are consumed by the model.
- Adds Strategy 1 recent focus window history rollup for msgCount pressure.
- Preserves the current round and recent focus window as a hard guarantee on the msgCount path.
- Keeps token-pressure strategies unchanged except for the active tool round guard.

## Verification
- `mvn -pl agentscope-extensions/agentscope-extensions-autocontext-memory -am -Dtest=AutoContextMemoryTest,MsgUtilsTest test`
- `mvn -pl agentscope-core,agentscope-extensions/agentscope-extensions-autocontext-memory -am test`

Fixes #1482.


lzbjut:
A few extra notes to make review easier:

## Review structure
This PR is intentionally split into 2 commits:

1. `fix(autocontext-memory): skip current-round compression when tool round is in-flight`
   - Fixes the correctness issue from #1482.
   - If the latest user turn has tool use/tool result messages but no final assistant response yet, AutoContextMemory now skips current-round compression.
   - This prevents Strategy 6/7 from summarizing a tool result before the main model has consumed the structured `role=tool` message.
   - The change is conservative: it only skips current-round compression while a tool round is still active; previous-round strategies remain unchanged.

2. `feat(autocontext-memory): add Strategy 1 recent focus window history rollup`
   - Adds a msgCount-pressure path that rolls complete historical rounds outside the recent focus window into one rolling summary.
   - The token-pressure path is intentionally unchanged except for the active tool round guard above.
   - If Strategy 1 succeeds but message count is still above threshold, the code intentionally does not fall through to current-round compression on the msgCount path. Preserving the current/recent focus window is a hard guarantee.

## Behavior details
- New default config:
  - `historyRollupKeepRecentRounds = 3`
  - `historyRollupKeepRecentTokens = 8000`
- New compression event:
  - `RECENT_FOCUS_WINDOW_COMPACT`
- The rollup event metadata records:
  - `roundCount`
  - `messageCountCompacted`
  - `messageCountBefore`
  - `messageCountAfter`
  - `rollupGeneration`
  - `tokenBefore`
  - `tokenAfter`
  - `offloaduuid`
- The applied log now includes both compacted messages and before/after working-memory size:
  - `Strategy 1: APPLIED - Rolled up X messages across Y old rounds into focus window summary (working memory: A -> B)`

## Scope
The diff is limited to:

`agentscope-extensions/agentscope-extensions-autocontext-memory/`

No `agentscope-core` files are changed.

## Verification run locally
- `mvn -pl agentscope-extensions/agentscope-extensions-autocontext-memory -am -Dtest=AutoContextMemoryTest,MsgUtilsTest test`
- `mvn -pl agentscope-core,agentscope-extensions/agentscope-extensions-autocontext-memory -am test`

Both passed locally.

AgentScopeJavaBot (commented):
🤖 AI Review

本次 PR 针对 issue #1482 提出的「AutoContextMemory 触发压缩后工具被重复调用直至 maxIters」的核心问题做了较为彻底的修复：
1. 通过新增 `isCurrentRoundToolInFlight` 守卫，确保 LLM 还未消费 tool_result 之前 Strategy 6/7 不再把 ToolUseBlock 压成普通摘要，根因解决了 ReAct 循环把历史 ToolUseBlock 当作新请求执行的隐患；
2. 在 msgCount 压力路径下引入 Strategy 1 「recent focus window history rollup」，并把 msgCount 与 token 两条压力路径的语义清晰分离，msgCount 路径明确不会回落到 current-round 压缩，保留了"当前轮 + 最近 focus window"作为硬保证；
3. 配套测试覆盖完整（active tool round 守护、token-only 路径、msgCount-only 路径、rollup 合并 generation 递增、token floor 扩展等），单测 116 全部通过。
整体设计与既有压缩策略风格一致，无明显架构缺陷。主要的潜在改进点集中在 offload 上下文的资源生命周期管理（异常路径与 generation 合并时旧 offload 条目泄漏），以及方法复杂度。

agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
```diff
@@ -231,80 +257,417 @@ public boolean compressIfNeeded() {
...
+            roundCount++;
+        }
+
+        int previousRollupGeneration = 0;
+        if (startIndex > 0 && isRecentFocusWindowRollupSummary(rawMessages.get(startIndex - 1))) {
+            // Only merge an adjacent previous rollup summary. If orphan messages sit between the
+            // old summary and newly aged-out rounds, leave them untouched for safety.
+            startIndex--;
+            previousRollupGeneration = getRollupGeneration(rawMessages.get(startIndex));
+        }
```
AgentScopeJavaBot, line 530: [major] Memory leak in `offloadContext` when an adjacent previous rollup summary is merged into the new rollup.

Here `startIndex--` extends the rollup range to include the previous summary message, then the whole range is offloaded under a fresh UUID and the previous summary is dropped from working memory. The previous summary's `offloaduuid` entry in `offloadContext` is never cleared, so every generation merge leaks one offload entry for the lifetime of the memory instance (in long sessions this map grows unbounded).

Suggested fix - clear the orphaned UUID before merging:
```java
if (startIndex > 0 && isRecentFocusWindowRollupSummary(rawMessages.get(startIndex - 1))) {
    Msg previousSummary = rawMessages.get(startIndex - 1);
    Object oldUuid = getCompressionMetadata(previousSummary).get(OFFLOAD_UUID_KEY);
    if (oldUuid instanceof String s) {
        clear(s);
    }
    startIndex--;
    previousRollupGeneration = getRollupGeneration(rawMessages.get(startIndex));
}
```
A similar concern applies if `generateRecentFocusHistoryRollup` throws (the LLM call uses `.block()` and only resumes `InterruptedException`): the `offload(uuid, messagesToRollup)` call at line 412 has already inserted the entry, so a model failure leaves an orphan. Calling `offload` after the summary is built, or wrapping the LLM call so the entry is removed on error, would close that loop too.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR addresses the root cause of issue #1482 with a well-scoped, two-layer fix in `AutoContextMemory`. First, `isCurrentRoundToolInFlight` correctly gates Strategy 6/7 so that an in-flight `tool_use` / `tool_result` pair is never summarized away before the main model consumes the structured tool result, which prevents the ReAct loop from re-issuing the same tool call after compression. Second, a new Strategy 1 (`rollupHistoryOutsideFocusWindow`) introduces a recency-weighted msgCount-pressure path that preserves the current round and a configurable recent focus window (rounds + token floor). The strategy renumbering, event type, and metadata schema are consistent across `AutoContextMemory`, `CompressionEvent`, `Prompts`, `PromptConfig`, `PromptProvider`, and `MsgUtils.isFinalAssistantResponse`. Test coverage is thorough (active tool round guard, msgCount-only path, token-only path, generation merge, token floor, no-fallthrough). The main residual concerns are (1) a minor offload-context leak when the previous rollup summary is merged into the new one, and (2) a misleading inline comment on the `MIN_HISTORY_ROLLUP_MESSAGES` threshold. No blocking issues.

agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
```diff
@@ -82,6 +83,17 @@
 public class AutoContextMemory implements StateModule, Memory, ContextOffLoader {
 
     private static final Logger log = LoggerFactory.getLogger(AutoContextMemory.class);
+    // Two complete text rounds: user + assistant, user + assistant.
+    private static final int MIN_HISTORY_ROLLUP_MESSAGES = 4;
```
AgentScopeJavaBot, line 87: [minor] The comment `// Two complete text rounds: user + assistant, user + assistant.` does not match the actual guard. `MIN_HISTORY_ROLLUP_MESSAGES = 4` is a raw message-count threshold: a single tool round (`USER + ASSISTANT(tool_use) + TOOL(tool_result) + ASSISTANT(final)` = 4 messages) also passes, and Strategy 1 would then roll up just one round. Either tighten the gate to actually require >=2 rounds (e.g. `rollupRange.roundCount() >= 2`) or update the comment to describe the message-count semantics so the intent is clear. Today the rollup payload + LLM cost can be triggered by a single 4-message round, which is probably not the intended floor.

agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
```diff
@@ -231,80 +257,417 @@ public boolean compressIfNeeded() {
...
+            Msg msg = msgs.get(i);
+            if (MsgUtils.isToolUseMessage(msg) || MsgUtils.isToolResultMessage(msg)) {
+                hasToolBlock = true;
+            }
+            if (MsgUtils.isFinalAssistantResponse(msg)) {
+                hasFinalAssistant = true;
+            }
+        }
+
+        return hasToolBlock && !hasFinalAssistant;
```
AgentScopeJavaBot, line 380: [praise] Nice surgical fix for the root cause of #1482: gating Strategy 6/7 on `hasToolBlock && !hasFinalAssistant` is the right boundary - it lets the main model finish consuming structured `role=tool` results before any summarization touches the current round, which is exactly what was missing in the original report (compression of in-flight tool pairs caused the ReAct loop to re-issue the same call until `maxIters`). Pairing it with the msgCount-pressure early-exit (`if (msgCountReached && !tokenCounterReached) return false;`) is also a nice hard guarantee - preserving the recent focus window even when no historical strategy fired.

agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
```diff
@@ -231,80 +257,417 @@ public boolean compressIfNeeded() {
...
             replaceWorkingMessage(currentContextMessages);
             return true;
         } else {
-            log.info("Strategy 6: SKIPPED - No current round messages to summarize");
+            log.info("Strategy 7: SKIPPED - No current round messages to summarize");
         }
 
         log.warn("All compression strategies exhausted but context still exceeds threshold");
         return false;
     }
```
AgentScopeJavaBot, line 349: [minor] `compressIfNeeded` is now ~160 lines with seven strategies plus two new gating branches (msgCount-only early return, in-flight tool guard). Readability would improve if each strategy block were extracted into a small private helper returning a `boolean` (`tryHistoryRollup`, `tryToolInvocationCompress`, ...), so this method becomes a flat sequence of `if (tryXxx(...)) return true;`. Not blocking, but the cyclomatic complexity will keep growing as new strategies are added.

agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
lzbjut: Fixed in 5585c3a3 by updating the comment to describe the actual raw-message threshold semantics. I left the behavior unchanged here to keep this follow-up focused on the review issue rather than changing the rollup trigger policy.

agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
lzbjut: Fixed in 5585c3a3. The new rollup offload now happens only after summary generation succeeds, so model failures no longer leave a fresh orphaned UUID. For rolling merges, the previous summary offload UUID is carried through the rollup range and cleared after the working-memory replacement succeeds. Added tests for both the merge cleanup and the model-failure path.

AgentScopeJavaBot: ✅ Verified as addressed in `08d8fe4`: `findHistoryRollupRange` now captures the previous rollup's `offloaduuid` into `HistoryRollupRange.previousRollupOffloadUuid`, and `rollupHistoryOutsideFocusWindow` clears it after the working-memory replacement (AutoContextMemory.java:450-452). The new `offload(uuid, messagesToRollup)` call happens after `generateRecentFocusHistoryRollup` succeeds (line 447 after line 419), so LLM failures no longer leave orphan entries in `offloadContext`.

agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
AgentScopeJavaBot: ✅ Verified as addressed in `08d8fe4`: The comment at AutoContextMemory.java:86 has been rewritten to describe the raw-message-count semantics (`Minimum raw messages worth an LLM rollup call, regardless of round shape.`), matching the suggested alternative of clarifying the intent rather than tightening the gate.

agentscope-extensions/agentscope-extensions-autocontext-memory/src/main/java/io/agentscope/core/memory/autocontext/AutoContextMemory.java
AgentScopeJavaBot: ✅ Verified as addressed in `08d8fe4`: Praise comment, no fix required. The gated logic the bot praised (`hasToolBlock && !hasFinalAssistant` in-flight tool guard and the msgCount-pressure early-exit) remains in place at AutoContextMemory.java:319-322 and line 380.

AgentScopeJavaBot (approved):
All blocking review comments have been addressed. 1 non-blocking suggestion (minor/nit) remains open and may be addressed in a follow-up.

LearningGp:
Should we target the 1.x branch for this? PTAL @chickenlj

PR #1518, opened 2026-05-28
fix(rag): respect RAGFlowConfig topK/similarityThreshold over framewo…

chenlinks:
…rk defaults (#1516)

ReActAgent.Builder seeds a default RetrieveConfig(limit=5, scoreThreshold=0.5) that masked the user-configured values on RAGFlowConfig. RAGFlowKnowledge now prefers RAGFlowConfig's topK/similarityThreshold and only falls back to the per-call RetrieveConfig when they are unset.

### DIFF
--- a/agentscope-extensions/agentscope-extensions-rag-ragflow/src/main/java/io/agentscope/core/rag/integration/ragflow/RAGFlowKnowledge.java
+++ b/agentscope-extensions/agentscope-extensions-rag-ragflow/src/main/java/io/agentscope/core/rag/integration/ragflow/RAGFlowKnowledge.java
@@ -131,9 +131,17 @@ public class RAGFlowKnowledge implements Knowledge {
 
         logger.debug("RAGFlow retrieve: query={}, config={}", query, config);
 
-        // Extract retrieval parameters
-        Integer topK = config != null ? config.getLimit() : null;
-        Double similarityThreshold = config != null ? config.getScoreThreshold() : null;
+        // Prefer user-explicit RAGFlowConfig values over the framework-supplied RetrieveConfig,
+        // which is auto-populated with defaults by ReActAgent.Builder and would otherwise mask
+        // RAGFlow-specific settings.
+        Integer topK =
+                this.config.getTopK() != null
+                        ? this.config.getTopK()
+                        : (config != null ? config.getLimit() : null);
+        Double similarityThreshold =
+                this.config.getSimilarityThreshold() != null
+                        ? this.config.getSimilarityThreshold()
+                        : (config != null ? config.getScoreThreshold() : null);
 
         // Call RAGFlow API (metadata condition from config)
         return client.retrieve(query, topK, similarityThreshold, null)
--- a/agentscope-extensions/agentscope-extensions-rag-ragflow/src/test/java/io/agentscope/core/rag/integration/ragflow/RAGFlowKnowledgeTest.java
+++ b/agentscope-extensions/agentscope-extensions-rag-ragflow/src/test/java/io/agentscope/core/rag/integration/ragflow/RAGFlowKnowledgeTest.java
@@ -20,12 +20,16 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import com.fasterxml.jackson.core.type.TypeReference;
 import io.agentscope.core.rag.model.Document;
 import io.agentscope.core.rag.model.RetrieveConfig;
+import io.agentscope.core.util.JsonUtils;
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


AgentScopeJavaBot (commented):
🤖 AI Review

The fix correctly addresses issue #1516: previously, `ReActAgent.Builder`'s auto-seeded `RetrieveConfig` (limit=5, scoreThreshold=0.5) silently masked user-configured `RAGFlowConfig.topK`/`similarityThreshold`. The new precedence ordering (RAGFlowConfig first, RetrieveConfig fallback) and the targeted regression test (`testRagFlowConfigTakesPrecedenceOverFrameworkRetrieveConfig`) are well-scoped. One non-blocking concern: `RAGFlowConfig.Builder` pre-seeds `topK = DEFAULT_TOP_K (1024)` and `similarityThreshold = DEFAULT_SIMILARITY_THRESHOLD (0.2)` at construction, so `this.config.getTopK()` is virtually never `null` in normal use. This makes the "fallback to RetrieveConfig" branch effectively unreachable unless callers explicitly pass `.topK(null)`, contradicting the new comment's claim of "only falls back ... when they are unset". It also subtly changes the default returned page size from 5 (framework) to 1024 (RAGFlow) for users who never set `topK`. The fix is still net-positive but the comment and tests should better reflect the actual semantics.

agentscope-extensions/agentscope-extensions-rag-ragflow/src/main/java/io/agentscope/core/rag/integration/ragflow/RAGFlowKnowledge.java
```diff
@@ -131,9 +131,17 @@ public Mono<List<Document>> retrieve(String query, RetrieveConfig config) {
...
+        // which is auto-populated with defaults by ReActAgent.Builder and would otherwise mask
+        // RAGFlow-specific settings.
+        Integer topK =
+                this.config.getTopK() != null
+                        ? this.config.getTopK()
+                        : (config != null ? config.getLimit() : null);
+        Double similarityThreshold =
+                this.config.getSimilarityThreshold() != null
+                        ? this.config.getSimilarityThreshold()
+                        : (config != null ? config.getScoreThreshold() : null);
```
AgentScopeJavaBot, line 144: [major] The fallback to `config.getLimit()` / `config.getScoreThreshold()` is effectively unreachable in normal usage. `RAGFlowConfig.Builder` initializes `topK = DEFAULT_TOP_K (1024)` and `similarityThreshold = DEFAULT_SIMILARITY_THRESHOLD (0.2)` as field defaults, so `this.config.getTopK()` returns non-null unless a caller explicitly invokes `.topK(null)`. This contradicts the comment's claim that the framework `RetrieveConfig` wins when RAGFlowConfig values are 'unset'. Two practical consequences: (1) the RetrieveConfig fallback branch is dead code under standard usage; (2) users who previously relied on the framework default `limit=5` without configuring `RAGFlowConfig.topK` will now silently get 1024. Either drop the Builder field defaults (and document the change), or rephrase the comment to reflect that RAGFlow's own defaults always win over RetrieveConfig — the current wording is misleading.

agentscope-extensions/agentscope-extensions-rag-ragflow/src/test/java/io/agentscope/core/rag/integration/ragflow/RAGFlowKnowledgeTest.java
```diff
@@ -239,6 +243,42 @@ void testRetrieveWithNullChunksResponse() throws Exception {
...
+        RecordedRequest request = mockWebServer.takeRequest();
+        Map<String, Object> body =
+                JsonUtils.getJsonCodec()
+                        .fromJson(
+                                request.getBody().readUtf8(),
+                                new TypeReference<Map<String, Object>>() {});
+
+        assertEquals(10, body.get("top_k"));
+        assertEquals(0.3, body.get("similarity_threshold"));
+    }
```
AgentScopeJavaBot, line 280: [minor] Nice regression test for the explicit-set case. Consider adding a complementary test that exercises the documented fallback path: build a `RAGFlowConfig` with `.topK(null).similarityThreshold(null)` and verify the framework `RetrieveConfig.limit/scoreThreshold` is forwarded to the request body. Without it, any future change to the Builder defaults (e.g., dropping the pre-seeded `DEFAULT_TOP_K`) could silently break the fallback semantics this PR introduces.

agentscope-extensions/agentscope-extensions-rag-ragflow/src/main/java/io/agentscope/core/rag/integration/ragflow/RAGFlowKnowledge.java
```diff
@@ -131,9 +131,17 @@ public Mono<List<Document>> retrieve(String query, RetrieveConfig config) {
...
-        // Extract retrieval parameters
-        Integer topK = config != null ? config.getLimit() : null;
-        Double similarityThreshold = config != null ? config.getScoreThreshold() : null;
+        // Prefer user-explicit RAGFlowConfig values over the framework-supplied RetrieveConfig,
+        // which is auto-populated with defaults by ReActAgent.Builder and would otherwise mask
+        // RAGFlow-specific settings.
+        Integer topK =
+                this.config.getTopK() != null
+                        ? this.config.getTopK()
+                        : (config != null ? config.getLimit() : null);
```
AgentScopeJavaBot, line 140: [nit] Minor readability: `this.config.getTopK()` is evaluated twice. Consider extracting to a local: `Integer cfgTopK = this.config.getTopK(); Integer topK = cfgTopK != null ? cfgTopK : (config != null ? config.getLimit() : null);` Same for `similarityThreshold`. Not blocking, just easier to scan.

PR #1524, opened 2026-05-28
feat(tool): add copy builder to ToolCallParam

### LINKED ISSUE #1517: [Feature]: ToolCallParam builder support itself
like grpc protobuf, its very convenient  to copy or change the object
```
        ToolCallParam param=...;
        ToolCallParam newParam = ToolCallParam.builder(param).build();
```

hexly666:
Add builder(ToolCallParam) factory method to create a builder initialized with values from an existing instance. Mutable fields such as the input map are deep-copied to ensure independence.

Closes #1517

## AgentScope-Java Version

## Description

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
@@ -115,6 +115,32 @@ public class ToolCallParam {
     }
 
     /**
+     * Creates a new builder initialized with values from an existing ToolCallParam.
+     *
+     * <p>This is useful for creating a modified copy of an existing parameter object.
+     * The builder copies all values from the source object, allowing selective
+     * modifications before building a new instance.
+     *
+     * <p><b>Example usage:</b>
+     * <pre>{@code
+     * ToolCallParam original = ...;
+     * ToolCallParam modified = ToolCallParam.builder(original)
+     *     .input(Map.of("newKey", "newValue"))
+     *     .build();
+     * }</pre>
+     *
+     * <p>Note: Mutable fields (such as the input map) are deep-copied to ensure
+     * the new instance is independent of the source. Immutable fields are shared.
+     *
+     * @param source The existing ToolCallParam to copy values from
+     * @return A new builder pre-populated with the source's values
+     * @throws NullPointerException if source is null
+     */
+    public static Builder builder(ToolCallParam source) {
+        return new Builder(source);
+    }
+
+    /**
      * Builder for ToolCallParam.
      */
     public static class Builder {
@@ -125,6 +151,14 @@ public class ToolCallParam {
         private ToolEmitter emitter;
 
         private Builder() {}
+
+        private Builder(ToolCallParam source) {
+            this.toolUseBlock = source.toolUseBlock;
+            this.input = source.input.isEmpty() ? null : new HashMap<>(source.input);
+            this.agent = source.agent;
+            this.context = source.context;
+            this.emitter = source.emitter;
+        }
 
         /**
          * Sets the tool use block.
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
@@ -20,6 +20,7 @@ import io.agentscope.core.message.ToolUseBlock;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * Parameters for tool invocation.
@@ -129,14 +130,16 @@ public class ToolCallParam {
      *     .build();
      * }</pre>
      *
-     * <p>Note: Mutable fields (such as the input map) are deep-copied to ensure
-     * the new instance is independent of the source. Immutable fields are shared.
+     * <p>Note: The input map structure is copied so that entries can be modified
+     * independently, but nested values (typed as Object) remain shared.
+     * Immutable fields (toolUseBlock, agent, context, emitter) are shared by reference.
      *
      * @param source The existing ToolCallParam to copy values from
      * @return A new builder pre-populated with the source's values
      * @throws NullPointerException if source is null
      */
     public static Builder builder(ToolCallParam source) {
+        Objects.requireNonNull(source, "source must not be null");
         return new Builder(source);
     }
 
@@ -154,7 +157,7 @@ public class ToolCallParam {
 
         private Builder(ToolCallParam source) {
             this.toolUseBlock = source.toolUseBlock;
-            this.input = source.input.isEmpty() ? null : new HashMap<>(source.input);
+            this.input = source.input.isEmpty() ? null : source.input;
             this.agent = source.agent;
             this.context = source.context;
             this.emitter = source.emitter;
--- a/agentscope-core/src/test/java/io/agentscope/core/tool/ToolCallParamTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/tool/ToolCallParamTest.java
@@ -194,30 +194,43 @@ class ToolCallParamTest {
         }
 
         @Test
-        @DisplayName("Should deep copy input map")
-        void testDeepCopyInputMap() {
+        @DisplayName("Should shallow copy input map (entries independent, nested values shared)")
+        @SuppressWarnings("unchecked")
+        void testShallowCopyInputMap() {
             ToolUseBlock toolUseBlock =
                     ToolUseBlock.builder().id("id").name("tool").input(Map.of()).build();
 
+            // Use a nested mutable value to verify shallow copy semantics
+            List<String> nestedList = new ArrayList<>(List.of("a", "b"));
             Map<String, Object> input = new HashMap<>();
-            input.put("key", "value");
+            input.put("list", nestedList);
 
             ToolCallParam original =
                     ToolCallParam.builder().toolUseBlock(toolUseBlock).input(input).build();
 
-            // Create copy
+            // Create copy (no input override)
             ToolCallParam copy = ToolCallParam.builder(original).build();
 
-            // Modify the copy's input (by building a new param with different input)
+            // Input maps should be equal
+            assertEquals(original.getInput(), copy.getInput());
+
+            // Nested mutable value should be the SAME reference (shallow copy)
+            assertSame(
+                    original.getInput().get("list"),
+                    copy.getInput().get("list"),
+                    "Nested values should be shared (shallow copy)");
+
+            // But top-level entries should be independent:
+            // adding a new top-level key to copy's input should NOT affect original
+            // (because ToolCallParam constructor does new HashMap<>(builder.input))
             Map<String, Object> modifiedInput = new HashMap<>(copy.getInput());
-            modifiedInput.put("key", "modified");
+            modifiedInput.put("newKey", "newValue");
             ToolCallParam modified = ToolCallParam.builder(original).input(modifiedInput).build();
 
-            // Original should be unchanged
-            assertEquals("value", original.getInput().get("key"));
-
-            // Modified should have new value
-            assertEquals("modified", modified.getInput().get("key"));
+            assertNull(
+                    original.getInput().get("newKey"),
+                    "Original should not have the new key added to the modified copy");
+            assertEquals("newValue", modified.getInput().get("newKey"));
         }
 
         @Test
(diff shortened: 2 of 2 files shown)


agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
```diff
@@ -114,6 +114,32 @@ public static Builder builder() {
...
+     *
+     * <p><b>Example usage:</b>
+     * <pre>{@code
+     * ToolCallParam original = ...;
+     * ToolCallParam modified = ToolCallParam.builder(original)
+     *     .input(Map.of("newKey", "newValue"))
+     *     .build();
+     * }</pre>
+     *
+     * <p>Note: Mutable fields (such as the input map) are deep-copied to ensure
```
AgentScopeJavaBot: [minor] The Javadoc states 'Mutable fields (such as the input map) are deep-copied', but `new HashMap<>(source.input)` performs a shallow copy: only the outer map structure is duplicated, while values inside the map (which are typed as `Object` and may be mutable, e.g. nested `List`/`Map`) remain shared with the source. If a caller mutates a nested value on either instance, the change is visible to both. Please either reword to 'shallow-copied' (entries are independent, but nested values are shared) or, if true deep-copy semantics are intended, document the requirement that input values be effectively immutable.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
```diff
@@ -114,6 +114,32 @@ public static Builder builder() {
...
+     * the new instance is independent of the source. Immutable fields are shared.
+     *
+     * @param source The existing ToolCallParam to copy values from
+     * @return A new builder pre-populated with the source's values
+     * @throws NullPointerException if source is null
+     */
+    public static Builder builder(ToolCallParam source) {
+        return new Builder(source);
```
AgentScopeJavaBot, line 143: [nit] The Javadoc declares `@throws NullPointerException if source is null`, but the NPE is currently raised only implicitly by `source.toolUseBlock` dereference inside the private `Builder(source)` constructor. Consider an explicit `Objects.requireNonNull(source, "source")` here (or at the start of the private constructor) so the failure message is stable and self-describing, and the contract is enforced at the public entry point.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
```diff
@@ -126,6 +152,14 @@ public static class Builder {
 
         private Builder() {}
 
+        private Builder(ToolCallParam source) {
+            this.toolUseBlock = source.toolUseBlock;
+            this.input = source.input.isEmpty() ? null : new HashMap<>(source.input);
```
AgentScopeJavaBot: [nit] This `new HashMap<>(source.input)` is redundant. The `ToolCallParam` constructor already performs `new HashMap<>(builder.input)` at build time (line 54), so when the user calls `builder(source).build()` without overriding `input`, the map gets copied twice. You can safely assign `this.input = source.input.isEmpty() ? null : source.input;` and rely on the existing defensive copy in the constructor. Pure micro-optimization — not blocking.

AgentScopeJavaBot (commented):
🤖 AI Review

The PR adds a clean copy-builder factory `ToolCallParam.builder(ToolCallParam source)` with thorough unit coverage (7 new tests across copy semantics, null handling, and field replacement). Logic is correct and the implementation is consistent with the existing builder pattern. Two documentation/clarity issues stand out: the Javadoc claim that mutable fields are "deep-copied" overstates what `new HashMap<>(source.input)` actually does (it is a shallow copy of entries — nested mutable values are still shared), and the test `testDeepCopyInputMap` does not actually assert deep-copy semantics. A minor nit: the `@throws NullPointerException` contract relies on implicit dereference; using `Objects.requireNonNull(source, "source")` would yield a clearer, stable error message. Also note the copy-builder's defensive `new HashMap<>` is redundant given `ToolCallParam`'s constructor already copies the map at build time, but this is purely a small efficiency concern, not a bug.

agentscope-core/src/test/java/io/agentscope/core/tool/ToolCallParamTest.java
```diff
@@ -117,6 +117,189 @@ void testBuilderChaining() {
...
+            assertNull(original.getInput().get("param2"));
+
+            // Modified should have new value
+            assertEquals("value2", modified.getInput().get("param2"));
+            assertNull(modified.getInput().get("param1"));
+        }
+
+        @Test
+        @DisplayName("Should deep copy input map")
+        void testDeepCopyInputMap() {
```
AgentScopeJavaBot: [nit] `testDeepCopyInputMap` does not actually verify deep-copy behavior. It builds `modified` from `original` with a wholly new `Map` via `.input(modifiedInput)`, which only exercises field replacement (already covered by `testModifyCopiedFields`). To truly assert deep-copy semantics you would need a nested mutable value, e.g. `input.put("list", new ArrayList<>(List.of("a")))`, then mutate `((List<?>) copy.getInput().get("list"))` and confirm the original list is unchanged. With the current shallow-copy implementation, that assertion would actually fail — which is itself useful information (see the related Javadoc finding).

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
AgentScopeJavaBot, line 132: [minor] The Javadoc states 'Mutable fields (such as the input map) are deep-copied', but `new HashMap<>(source.input)` performs a shallow copy: only the outer map structure is duplicated, while values inside the map (which are typed as `Object` and may be mutable, e.g. nested `List`/`Map`) remain shared with the source. If a caller mutates a nested value on either instance, the change is visible to both. Please either reword to 'shallow-copied' (entries are independent, but nested values are shared) or, if true deep-copy semantics are intended, document the requirement that input values be effectively immutable.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
AgentScopeJavaBot, line 140: [nit] The Javadoc declares `@throws NullPointerException if source is null`, but the NPE is currently raised only implicitly by `source.toolUseBlock` dereference inside the private `Builder(source)` constructor. Consider an explicit `Objects.requireNonNull(source, "source")` here (or at the start of the private constructor) so the failure message is stable and self-describing, and the contract is enforced at the public entry point.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
AgentScopeJavaBot, line 157: [nit] This `new HashMap<>(source.input)` is redundant. The `ToolCallParam` constructor already performs `new HashMap<>(builder.input)` at build time (line 54), so when the user calls `builder(source).build()` without overriding `input`, the map gets copied twice. You can safely assign `this.input = source.input.isEmpty() ? null : source.input;` and rely on the existing defensive copy in the constructor. Pure micro-optimization — not blocking.

agentscope-core/src/test/java/io/agentscope/core/tool/ToolCallParamTest.java
AgentScopeJavaBot, line 198: [nit] `testDeepCopyInputMap` does not actually verify deep-copy behavior. It builds `modified` from `original` with a wholly new `Map` via `.input(modifiedInput)`, which only exercises field replacement (already covered by `testModifyCopiedFields`). To truly assert deep-copy semantics you would need a nested mutable value, e.g. `input.put("list", new ArrayList<>(List.of("a")))`, then mutate `((List<?>) copy.getInput().get("list"))` and confirm the original list is unchanged. With the current shallow-copy implementation, that assertion would actually fail — which is itself useful information (see the related Javadoc finding).

AgentScopeJavaBot (commented):
🤖 AI Review

The PR adds a clean copy-builder factory `ToolCallParam.builder(ToolCallParam source)` with thorough unit coverage (7 new tests across copy semantics, null handling, and field replacement). Logic is correct and the implementation is consistent with the existing builder pattern. Two documentation/clarity issues stand out: the Javadoc claim that mutable fields are "deep-copied" overstates what `new HashMap<>(source.input)` actually does (it is a shallow copy of entries — nested mutable values are still shared), and the test `testDeepCopyInputMap` does not actually assert deep-copy semantics. A minor nit: the `@throws NullPointerException` contract relies on implicit dereference; using `Objects.requireNonNull(source, "source")` would yield a clearer, stable error message. Also note the copy-builder's defensive `new HashMap<>` is redundant given `ToolCallParam`'s constructor already copies the map at build time, but this is purely a small efficiency concern, not a bug.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
AgentScopeJavaBot: ✅ Verified as addressed in `a8f42a4`: Javadoc has been reworded at lines 133-135 to explicitly state the input map structure is copied while nested values remain shared.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
AgentScopeJavaBot: ✅ Verified as addressed in `a8f42a4`: Explicit Objects.requireNonNull(source, "source must not be null") added at line 142, before the private Builder(source) constructor is called.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
AgentScopeJavaBot: ✅ Verified as addressed in `a8f42a4`: Builder(source) now assigns this.input = source.input.isEmpty() ? null : source.input at line 160, eliminating the redundant copy.

agentscope-core/src/test/java/io/agentscope/core/tool/ToolCallParamTest.java
AgentScopeJavaBot: ✅ Verified as addressed in `a8f42a4`: Test renamed to testShallowCopyInputMap (lines 199-234); uses a nested ArrayList and asserts assertSame on the nested value to verify shallow-copy semantics.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
AgentScopeJavaBot: ✅ Verified as addressed in `a8f42a4`: Javadoc has been reworded at lines 133-135 to explicitly state the input map structure is copied while nested values remain shared.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
AgentScopeJavaBot: ✅ Verified as addressed in `a8f42a4`: Explicit Objects.requireNonNull(source, "source must not be null") added at line 142, before the private Builder(source) constructor is called.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
AgentScopeJavaBot: ✅ Verified as addressed in `a8f42a4`: Builder(source) now assigns this.input = source.input.isEmpty() ? null : source.input at line 160, eliminating the redundant copy.

agentscope-core/src/test/java/io/agentscope/core/tool/ToolCallParamTest.java
AgentScopeJavaBot: ✅ Verified as addressed in `a8f42a4`: Test renamed to testShallowCopyInputMap (lines 199-234); uses a nested ArrayList and asserts assertSame on the nested value to verify shallow-copy semantics.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
AgentScopeJavaBot: ✅ Verified as addressed in `a8f42a4`: Javadoc has been reworded at lines 133-135 to explicitly state the input map structure is copied while nested values remain shared.

AgentScopeJavaBot (approved):

LearningGp (approved):
LGTM

PR #1527, opened 2026-05-29
fix(a2a): preserve streaming tool use content

### LINKED ISSUE #1526: [Bug]: A2A ToolUseBlockParser drops streaming tool call content
**Describe the bug**
When converting a streaming `ToolUseBlock` to A2A `DataPart`, `ToolUseBlockParser` only serializes `ToolUseBlock#getInput()` into `DataPart.data`.

For streaming tool-call fragments, `input` and `metadata` can be empty while the raw argument delta is stored in `ToolUseBlock#getContent()`. As a result, the streamed tool argument content is lost during A2A message conversion.

**To Reproduce**
1. Create a streaming tool-use fragment:

```java
ToolUseBlock block =
        ToolUseBlock.builder()
                .id("call_f6603016fc7a46a3a3d6ce44")
                .name("__fragment__")
                .content("{\"productName\": \"")
                .build();
```
2. Convert it with `new ToolUseBlockParser().parse(block)`.
3. Inspect the returned `DataPart`.

**Expected behavior**
The streaming raw content should be preserved when converting `ToolUseBlock` to `DataPart`, and restored when converting `DataPart` back to `ToolUseBlock`.

**Error messages**
No exception is thrown. The bug is data loss: `ToolUseBlock.content` is dropped because only `ToolUseBlock.input` is serialized.

Example streaming fragments:
```json
{"type":"tool_use","id":"call_f6603016fc7a46a3a3d6ce44","name":"__fragment__","input":{},"content":"{\"productName\": \"","metadata":{}}
{"type":"tool_use","id":"call_f6603016fc7a46a3a3d6ce44","name":"__fragment__","input":{},"content":"White Peach Oolong Milk Tea","metadata":{}}
{"type":"tool_use","id":"call_f6603016fc7a46a3a3d6ce44","name":"__fragment__","input":{},"content":"\"","metadata":{}}
{"type":"tool_use","id":"call_f6603016fc7a46a3a3d6ce44","name":"__fragment__","input":{},"content":"}","metadata":{}}
```

- AgentScope-Java Version: 1.1.0-SNAPSHOT
- Java Version: 17
- OS: Windows

**Additional context**
This affects A2A streaming tool calls where `ToolUseBlock.content` carries incremental raw arguments and `ToolUseBlock.input` is still empty.

java-lbb:
## AgentScope-Java Version

1.1.0-SNAPSHOT

## Description

This PR fixes A2A conversion for streaming `ToolUseBlock` fragments.

Previously, `ToolUseBlockParser` serialized only `ToolUseBlock#getInput()` into `DataPart.data`. Streaming tool-call fragments can carry argument deltas in `ToolUseBlock#getContent()` while `input` and `metadata` are empty, so the raw content was dropped during A2A conversion.

Changes made:

- Add an internal `DataPart.data` key for raw streaming tool-use content.
- Preserve `ToolUseBlock.content` when converting `ToolUseBlock` to `DataPart`.
- Restore that content when converting `DataPart` back to `ToolUseBlock`.
- Keep the internal content key out of `ToolUseBlock.input`.
- Add regression tests for both conversion directions.

Test commands:

```bash
mvn -pl agentscope-extensions/agentscope-extensions-a2a/agentscope-extensions-a2a-client -am "-Dtest=ToolUseBlockParserTest#testParseStreamingToolUseContent,DataPartParserTest#testParseDataPartWithStreamingToolUseContent" test

mvn -pl agentscope-extensions/agentscope-extensions-a2a/agentscope-extensions-a2a-client -am test
```

Local results:
```bash
Tests run: 106, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Fixes #1526

## Checklist

- [x] All related tests are passing
- [x] Related documentation has been updated (not needed for this bug fix)

### DIFF
--- a/agentscope-extensions/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/message/DataPartParser.java
+++ b/agentscope-extensions/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/message/DataPartParser.java
@@ -73,7 +73,12 @@ public class DataPartParser implements PartParser<DataPart> {
         ToolUseBlock.Builder builder = ToolUseBlock.builder();
         builder.id(getToolCallId(part)).name(getToolName(part));
         builder.metadata(getOriginalMetadata(part));
-        builder.input(part.getData());
+        Map<String, Object> input = new HashMap<>(part.getData());
+        Object content = input.remove(MessageConstants.TOOL_USE_CONTENT_DATA_KEY);
+        builder.input(input);
+        if (content != null) {
+            builder.content(content.toString());
+        }
         return builder.build();
     }
 
--- a/agentscope-extensions/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/test/java/io/agentscope/core/a2a/agent/message/DataPartParserTest.java
+++ b/agentscope-extensions/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/test/java/io/agentscope/core/a2a/agent/message/DataPartParserTest.java
@@ -86,6 +86,30 @@ class DataPartParserTest {
     }
 
     @Test
+    @DisplayName("Should parse streaming tool use content")
+    void testParseDataPartWithStreamingToolUseContent() {
+        Map<String, Object> metadata =
+                Map.of(
+                        "_agentscope_block_type", "tool_use",
+                        "_agentscope_tool_name", "__fragment__",
+                        "_agentscope_tool_call_id", "call-123");
+        DataPart part =
+                new DataPart(
+                        Map.of(MessageConstants.TOOL_USE_CONTENT_DATA_KEY, "{\"productName\": \""),
+                        metadata);
+
+        ContentBlock result = parser.parse(part);
+
+        assertNotNull(result);
+        assertEquals(ToolUseBlock.class, result.getClass());
+        ToolUseBlock toolUseBlock = (ToolUseBlock) result;
+        assertEquals("__fragment__", toolUseBlock.getName());
+        assertEquals("call-123", toolUseBlock.getId());
+        assertEquals("{\"productName\": \"", toolUseBlock.getContent());
+        assertTrue(toolUseBlock.getInput().isEmpty());
+    }
+
+    @Test
     @DisplayName("Should parse DataPart with tool_result metadata to ToolResultBlock")
     void testParseDataPartWithToolResultMetadata() {
         Map<String, Object> metadata =
--- a/agentscope-extensions/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/message/MessageConstants.java
+++ b/agentscope-extensions/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/message/MessageConstants.java
@@ -48,5 +48,7 @@ public class MessageConstants {
 
     public static final String TOOL_CALL_ID_METADATA_KEY = "_agentscope_tool_call_id";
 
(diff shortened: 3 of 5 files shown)


agentscope-extensions/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/test/java/io/agentscope/core/a2a/agent/message/DataPartParserTest.java
```diff
@@ -85,6 +85,30 @@ void testParseDataPartWithToolUseMetadata() {
...
+
+        ContentBlock result = parser.parse(part);
+
+        assertNotNull(result);
+        assertEquals(ToolUseBlock.class, result.getClass());
+        ToolUseBlock toolUseBlock = (ToolUseBlock) result;
+        assertEquals("__fragment__", toolUseBlock.getName());
+        assertEquals("call-123", toolUseBlock.getId());
+        assertEquals("{\"productName\": \"", toolUseBlock.getContent());
+        assertTrue(toolUseBlock.getInput().isEmpty());
```
AgentScopeJavaBot, line 109: [nit] Minor test-coverage nit: this regression test does not assert that `toolUseBlock.getMetadata()` no longer contains the internal `_agentscope_*` keys after `getOriginalMetadata` strips them. Adding `assertTrue(toolUseBlock.getMetadata().isEmpty())` (or asserting absence of `TOOL_USE_CONTENT_DATA_KEY`, `BLOCK_TYPE_METADATA_KEY`, etc.) would tighten the round-trip contract and guard against future regressions that accidentally leak the internal data key into the rebuilt block.

agentscope-extensions/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/message/DataPartParser.java
```diff
@@ -73,7 +73,12 @@ private ContentBlock parseToToolUseBlock(DataPart part) {
         ToolUseBlock.Builder builder = ToolUseBlock.builder();
         builder.id(getToolCallId(part)).name(getToolName(part));
         builder.metadata(getOriginalMetadata(part));
-        builder.input(part.getData());
+        Map<String, Object> input = new HashMap<>(part.getData());
+        Object content = input.remove(MessageConstants.TOOL_USE_CONTENT_DATA_KEY);
+        builder.input(input);
+        if (content != null) {
+            builder.content(content.toString());
```
AgentScopeJavaBot, line 80: [nit] Since the producer (`ToolUseBlockParser`) always writes `contentBlock.getContent()` (a `String`) into this slot, `content.toString()` is effectively a no-op for the in-process case. Using `Object` + `toString()` is defensive against malformed wire payloads, which is fine; just be aware that if a non-agentscope peer ever puts a non-String value under the reserved key, it will be silently coerced rather than rejected. Optional: log a warn when `content` is not a `String`, or cast to surface the inconsistency.

AgentScopeJavaBot (approved):
AI Review

This PR fixes a real bug in the A2A streaming path where `ToolUseBlock#getContent()` (the raw delta payload of an in-flight tool call) was silently dropped during A2A `DataPart` conversion because only `getInput()` was serialized. The fix adds a reserved internal data key (`_agentscope_tool_content`, consistently prefixed with `_agentscope_`), preserves `content` when going `ToolUseBlock` -> `DataPart`, strips it back out and re-attaches it on the reverse path, and adds regression tests for both directions. Defensive copies are correctly used on both sides (input maps from `getInput()` and `getData()` are unmodifiable in some call sites, so the wrapping `HashMap` is needed before `remove`/`put`). The choice to keep the marker in `data` rather than `metadata` matches the existing convention used by `TOOL_RESULT_OUTPUT_METADATA_KEY`. Server-side `MessageConvertUtil` reuses the client-side `convertFromMsg`, so symmetry is preserved without any extra change. No critical or major issues found.

PR #1529, opened 2026-05-29
feat(harness): add projectWorkspace for filesystem tools

kongweiguang:
## Summary

Retargets the closed #1528 work so the source branch is `kongweiguang:main` instead of a temporary feature branch.

- Adds `HarnessAgent.Builder.projectWorkspace(Path)` so built-in file tools can operate on a project directory separate from the Harness state workspace.
- Keeps `workspace(...)` responsible for `AGENTS.md`, memory, sessions, skills, workspace context, and other Harness state.
- Uses a virtual `LocalFilesystem` rooted at `projectWorkspace` for file tools without the per-user namespace.
- Preserves sandbox-backed file tool behavior by default when `projectWorkspace` is not configured.
- Propagates `projectWorkspace` to shared/general-purpose subagents so their file tools use the same project directory.

## Validation

- `mvn -pl agentscope-harness -am -Dtest=HarnessAgentTest test`
- GitHub Actions: Check License, Check Module Sync, Ubuntu build, Windows build all passed.
- Codecov patch: all modified and coverable lines are covered by tests.


agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1542,6 +1564,7 @@ public HarnessAgent build() {
                 }
                 capturedSandboxFs = new SandboxBackedFilesystem();
                 filesystem = capturedSandboxFs;
+                toolFilesystem = resolveProjectFilesystem(projectWorkspace, filesystem);
```
AgentScopeJavaBot, line 1632: [major] Sandbox isolation footgun. When `filesystem(SandboxFilesystemSpec)` and `projectWorkspace(...)` are both configured, `toolFilesystem` is reassigned to a plain local `LocalFilesystem` here while `filesystem` keeps pointing at `SandboxBackedFilesystem`. The downstream branch at line ~1725 (`filesystem instanceof AbstractSandboxFilesystem`) therefore still registers `ShellExecuteTool` against the sandbox — leaving the agent in a hybrid state where shell/state run inside the sandbox but `read_file`/`write_file`/`edit_file`/`grep_files` write to the host. Any file created by `write_file` will not be visible to a follow-up `execute` shell call, and any sandbox isolation guarantees the user expected are silently bypassed for file tools. Either fail-fast in `build()` when both options are set, or explicitly document this behavior on `projectWorkspace(...)` so callers understand the trade-off.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1089,6 +1091,24 @@ public Builder workspace(Path workspace) {
...
+         * memory, sessions, skills, and other agent state. When this option is set, built-in file
+         * tools such as {@code read_file}, {@code write_file}, {@code edit_file},
+         * {@code list_files}, {@code grep_files}, and {@code glob_files} operate against this
+         * project directory instead, without applying the per-user namespace used by the harness
+         * workspace backend.
+         *
+         * <p>This is useful for applications that keep agent state in one stable directory but need
+         * the agent to work on arbitrary project folders.
+         */
+        public Builder projectWorkspace(Path projectWorkspace) {
```
AgentScopeJavaBot, line 1218: [minor] JavaDoc does not describe precedence vs. the other filesystem-configuration paths. Worth calling out explicitly: (1) `projectWorkspace` also overrides the `abstractFilesystem(...)` escape hatch for file tools while leaving it in effect for state — users who plug in a custom backend may be surprised; (2) it overrides sandbox-backed file tools (see line 1567); (3) `projectWorkspace` does not respect `LocalFilesystemSpec` knobs (maxFileSizeMb, namespace factory). A short precedence note here would prevent confusion.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1909,6 +1932,14 @@ private AbstractFilesystem resolveFilesystem(
             return new LocalFilesystemWithShell(workspace, nsFactory);
         }
 
+        private AbstractFilesystem resolveProjectFilesystem(
+                Path projectWorkspace, AbstractFilesystem workspaceFilesystem) {
+            if (projectWorkspace == null) {
+                return workspaceFilesystem;
+            }
+            return new LocalFilesystem(projectWorkspace, true, 10);
```
AgentScopeJavaBot: [minor] `new LocalFilesystem(projectWorkspace, true, 10)` hardcodes `virtualMode=true` and `maxFileSizeMb=10`. The 10 MB cap silently truncates grep/glob results for larger files in user project directories with no configuration knob. Consider either (a) accepting an optional `LocalFilesystemSpec`/`int maxFileSizeMb` companion setter on the builder, or (b) reusing the existing `LocalFilesystem.DEFAULT_MAX_FILE_SIZE_MB` constant instead of the literal `10` so the value tracks the default in one place. Also, no validation is performed on `projectWorkspace` (non-null, exists, is a directory) — a typo will fail later at first tool call rather than at build time.

agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentTest.java
```diff
@@ -346,6 +354,174 @@ void remoteFilesystemSpec_sharesMemoryMdInNonsandboxMode() throws Exception {
...
+
+        try (HarnessAgent agent =
+                HarnessAgent.builder()
+                        .name("agent")
+                        .model(stubModel("ok"))
+                        .workspace(workspace)
+                        .projectWorkspace(projectWorkspace)
+                        .toolResultEviction(ToolResultEvictionConfig.defaults())
+                        .build()) {
+            assertNotNull(agent);
```
AgentScopeJavaBot, line 401: [minor] This test is named `projectWorkspace_toolResultEvictionUsesProjectFilesystem` but only asserts `assertNotNull(agent)`. It does not exercise the `ToolResultEvictionHook` (which is the point of the test) and would pass even if eviction silently regressed to the harness workspace filesystem. Either trigger an oversized tool result and assert the eviction file lands under `projectWorkspace`, or delete the test — the current form gives false confidence.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1089,6 +1091,24 @@ public Builder workspace(Path workspace) {
...
+         * project directory instead, without applying the per-user namespace used by the harness
+         * workspace backend.
+         *
+         * <p>This is useful for applications that keep agent state in one stable directory but need
+         * the agent to work on arbitrary project folders.
+         */
+        public Builder projectWorkspace(Path projectWorkspace) {
+            this.projectWorkspace = projectWorkspace;
+            return this;
+        }
```
AgentScopeJavaBot, line 1221: [nit] For parity with `workspace(Path)` / `workspace(String)`, consider adding a `projectWorkspace(String)` overload that mirrors the strip/blank-rejection logic of `workspace(String)` at line 1122. Otherwise callers using string-based configuration (YAML/env) have to construct `Path.of(...)` themselves.

AgentScopeJavaBot (commented):
AI Review

The change adds a new `HarnessAgent.Builder#projectWorkspace(Path)` option that decouples the directory exposed to built-in file tools (`read_file`, `write_file`, `edit_file`, `list_files`, `grep_files`, `glob_files`) from the harness "state" workspace (AGENTS.md, memory, sessions, skills). Propagation to the general-purpose subagent and SHARED-mode declared subagents is handled. The feature is well-scoped and tests cover the happy paths plus subagent inheritance. Main concerns are around the sandbox interaction (file tools silently escape the sandbox when `projectWorkspace` is also set, while the shell tool keeps using the sandbox), undocumented precedence vs. the `abstractFilesystem` escape hatch, and hardcoded `LocalFilesystem` knobs (virtualMode/maxFileSizeMb). One of the new tests only asserts `assertNotNull(agent)` so it does not exercise what its name promises.

PR #1530, opened 2026-05-29
Add PostgreSQL session storage

kongweiguang:
## Summary
- Add a PostgreSQL-backed session storage extension.
- Register the module in extensions, BOM, and all-in-one distribution.
- Document the new PostgreSQL session dependency.

## Validation
- `mvn -pl agentscope-extensions/agentscope-extensions-session-postgresql spotless:apply test`
- Real PostgreSQL smoke test against `postgres:16-alpine` in Docker
- `mvn -pl agentscope-distribution/agentscope-bom -DskipTests package`
- `mvn -pl agentscope-distribution/agentscope-all -am -DskipTests package`

### DIFF
--- a/agentscope-distribution/agentscope-all/pom.xml
+++ b/agentscope-distribution/agentscope-all/pom.xml
@@ -194,6 +194,13 @@
 
         <dependency>
             <groupId>io.agentscope</groupId>
+            <artifactId>agentscope-extensions-session-postgresql</artifactId>
+            <scope>compile</scope>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>io.agentscope</groupId>
             <artifactId>agentscope-extensions-session-redis</artifactId>
             <scope>compile</scope>
             <optional>true</optional>
--- a/agentscope-distribution/agentscope-bom/pom.xml
+++ b/agentscope-distribution/agentscope-bom/pom.xml
@@ -225,6 +225,13 @@
                 <version>${project.version}</version>
             </dependency>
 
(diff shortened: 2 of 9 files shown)


AgentScopeJavaBot (commented):
AI Review

This PR adds a new `agentscope-extensions-session-postgresql` module that closely mirrors the existing `MysqlSession` implementation: hash-based incremental list storage, parameterized writes, schema/table validation, and transaction-wrapped writes. SQL injection surface is correctly handled via a strict identifier regex plus parameterized queries; all JDBC writes go through `executeInWriteTransaction` which preserves original auto-commit state. The biggest concerns are non-blocking: (a) the production class contains explicit H2 dialect detection (`isH2`/`getUpsertSql`) used only to make CI tests run without a real PostgreSQL, leaking test concerns into the production write path; (b) several read methods (`get`, `getStoredHash`, `getListCount`) abandon try-with-resources in favor of nested manual `try { } finally { rs.close(); }` blocks, which can mask the root-cause SQLException with a close-failure (and is inconsistent with both `MysqlSession` and `getList()` in the same class); (c) the PR description claims a "Real PostgreSQL smoke test against postgres:16-alpine in Docker", but the only committed E2E suite uses H2 in PostgreSQL compatibility mode, so genuine PG features (`ON CONFLICT ... DO UPDATE`, real schemas, TRUNCATE locking semantics) are not exercised in CI.

_Note: inline comments could not be attached (line outside diff hunk); please refer to the summary above._

amyaxy:
已经在 v2 实现，模块为 agentscope-extensions-postgresql；
可以考虑关闭该PR @jujn

PR #1533, opened 2026-05-30
fix(core): dispose event stream upstream subscription on cancel

### LINKED ISSUE #1466: [Bug]:Resource leak (Ghost Task) in `AgentBase#createEventStream` when `Flux` stream is cancelled
### Describe the bug

When streaming agent responses (e.g. via SSE), if the downstream consumer (client) cancels the `Flux` subscription, the upstream `Mono` executing the agent logic and LLM requests is not properly terminated.

The root cause is that in `AgentBase.java` (`createEventStream` method), the `Disposable` returned by `Mono.defer().subscribe(...)` is never registered with the `FluxSink` dispose lifecycle. As a result, when the downstream cancels the `Flux`, the LLM request continues running in the background as a "ghost task".

This leads to:

1. **Resource leak**
   - Wasted API tokens
   - Unnecessary network bandwidth usage
   - Thread pool/resource exhaustion
   - HTTP streaming clients continuing to receive chunks even after disconnect

2. **Lock blocking**
   - The `Agent` remains locked (`running = true`) until the background task naturally completes
   - Subsequent requests against the same agent instance may fail with:

```text
IllegalStateException: Agent is still running, please wait for it to finish
```

### To Reproduce

1. Set up an AgentScope application with a reactive web framework (e.g. Spring WebFlux) exposing an SSE endpoint.
2. Send a request that triggers a long-running LLM generation.
3. After streaming starts, immediately cancel the client request (e.g. close the browser tab or abort the HTTP request).
4. Observe backend logs.

The underlying LLM client (e.g. `DashScopeHttpClient`) continues printing logs such as:

```text
Received SSE data chunk
```

even though the downstream client has already disconnected.

### Expected behavior

Cancellation from the downstream subscriber should propagate upstream.

When the `Flux` is cancelled:

- the background `Mono` execution should be disposed immediately
- ongoing LLM network requests should stop
- the agent execution lock should be released (`running.set(false)`)

### Current problematic code

```java
// AgentBase.java (v1.0.12, around line 789)

sink.complete();
},
sink::error);
},
FluxSink.OverflowStrategy.BUFFER)
```

### Proposed Fix

Register the `Disposable` using `sink.onDispose(...)` so downstream cancellation propagates to the upstream subscription.

```java
Disposable disposable =
        Mono.defer(() -> {
            ...
        })
        .subscribe(
                unused -> {},
                sink::error,
                sink::complete);

sink.onDispose(disposable);
```

### System Information

- **AgentScope-Java Version**: `v1.0.12`
- **Environment**: Reactive Web Streaming (Spring WebFlux / SSE)
- **Component**: `agentscope-core`
- **Affected Class**: `io.agentscope.core.agent.AgentBase`

guslegend0510:
### 背景
`AgentBase#createEventStream` 当前通过 `Mono.defer(...).subscribe(...)` 驱动流式调用，但下游取消订阅时，上游执行中的订阅没有被及时释放。

这会带来两个问题：
- 已取消的流在后台继续执行，形成“幽灵任务”
- `running` 状态可能没有及时释放，导致后续请求报 `Agent is still running`

### 修改内容
- 在 `AgentBase#createEventStream` 中，把上游订阅绑定到 `FluxSink` 的 dispose 生命周期
- 使用 `Disposable.Swap` 配合 `sink.onDispose(...)`，确保下游取消时可以及时 `dispose` 正在执行的上游订阅
- 保留现有的流式 hook、subagent event bus 和结果事件发射逻辑不变

### 测试
新增回归测试，覆盖以下场景：
- 第一次 `stream()` 被取消后，上游订阅会收到 cancel
- 取消后执行锁能够及时释放
- 第二次 `call()` 可以立即成功，不会再因为残留 `running=true` 失败

已验证：
- `mvn -pl agentscope-core -am -DskipTests spotless:apply`
- `mvn -pl agentscope-core -am "-Dtest=AgentBaseTest,AgentStreamingTest" test`

Fixes #1466

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
@@ -41,6 +41,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import reactor.core.Disposable;
+import reactor.core.Disposables;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.FluxSink;
 import reactor.core.publisher.Mono;
@@ -910,37 +912,44 @@ public abstract class AgentBase implements StateModule, Agent {
                                             // into this parent sink without an extra Flux layer.
                                             SubagentEventBus bus = sink::next;
 
+                                            Disposable.Swap upstreamSubscription =
+                                                    Disposables.swap();
+                                            sink.onDispose(upstreamSubscription);
+
                                             // Use Mono.defer to ensure trace context propagation
                                             // while maintaining streaming hook functionality
-                                            Mono.defer(() -> callSupplier.get())
-                                                    .contextWrite(
-                                                            context ->
-                                                                    context.put(
-                                                                                    SubagentEventBus
-                                                                                            .CONTEXT_KEY,
-                                                                                    bus)
-                                                                            .putAll(ctxView))
-                                                    .doFinally(
-                                                            signalType -> {
-                                                                // Remove temporary hook
-                                                                hooks.remove(streamingHook);
-                                                            })
-                                                    .subscribe(
-                                                            finalMsg -> {
-                                                                if (options.shouldStream(
-                                                                        EventType.AGENT_RESULT)) {
-                                                                    sink.next(
-                                                                            new Event(
-                                                                                    EventType
-                                                                                            .AGENT_RESULT,
-                                                                                    finalMsg,
-                                                                                    true));
-                                                                }
+                                            upstreamSubscription.update(
+                                                    Mono.defer(() -> callSupplier.get())
+                                                            .contextWrite(
+                                                                    context ->
+                                                                            context.put(
+                                                                                            SubagentEventBus
+                                                                                                    .CONTEXT_KEY,
+                                                                                            bus)
+                                                                                    .putAll(
+                                                                                            ctxView))
+                                                            .doFinally(
+                                                                    signalType -> {
+                                                                        // Remove temporary hook
+                                                                        hooks.remove(streamingHook);
+                                                                    })
+                                                            .subscribe(
+                                                                    finalMsg -> {
+                                                                        if (options.shouldStream(
+                                                                                EventType
+                                                                                        .AGENT_RESULT)) {
+                                                                            sink.next(
+                                                                                    new Event(
+                                                                                            EventType
+                                                                                                    .AGENT_RESULT,
+                                                                                            finalMsg,
+                                                                                            true));
+                                                                        }
 
-                                                                // Complete the stream
-                                                                sink.complete();
-                                                            },
-                                                            sink::error);
+                                                                        // Complete the stream
+                                                                        sink.complete();
+                                                                    },
+                                                                    sink::error));
                                         },
                                         FluxSink.OverflowStrategy.BUFFER)
                                 .publishOn(Schedulers.boundedElastic()));
--- a/agentscope-core/src/test/java/io/agentscope/core/agent/AgentBaseTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/agent/AgentBaseTest.java
@@ -32,12 +32,14 @@ import java.time.Duration;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
+import reactor.test.StepVerifier;
 
 /**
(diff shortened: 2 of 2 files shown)


agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
```diff
@@ -910,37 +912,44 @@ private Flux<Event> createEventStream(StreamOptions options, Supplier<Mono<Msg>>
                                             // into this parent sink without an extra Flux layer.
                                             SubagentEventBus bus = sink::next;
 
+                                            Disposable.Swap upstreamSubscription =
+                                                    Disposables.swap();
+                                            sink.onDispose(upstreamSubscription);
```
AgentScopeJavaBot, line 917: [nit] Minor: `addHook(streamingHook)` is invoked before the Swap is wired and before the upstream subscription is created. Hook removal still happens via `doFinally` on the inner chain, which fires on `ON_ERROR` if `acquireExecution` throws `IllegalStateException` (agent already running). So no leak in practice, but a one-line comment noting that the streamingHook lifecycle is owned by the inner `doFinally` (and intentionally not by `sink.onDispose`) would help future maintainers reason about ordering.

agentscope-core/src/test/java/io/agentscope/core/agent/AgentBaseTest.java
```diff
@@ -112,6 +114,67 @@ protected Mono<Msg> handleInterrupt(InterruptContext context, Msg... originalArg
...
+            } catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
+                return false;
+            }
+        }
+
+        @Override
+        protected Mono<Msg> doCall(List<Msg> msgs) {
+            if (callCount.incrementAndGet() == 1) {
+                return Mono.<Msg>create(sink -> firstCallStarted.countDown())
```
AgentScopeJavaBot, line 151: [nit] `Mono.<Msg>create(sink -> firstCallStarted.countDown()).doOnCancel(...)` is a clever way to verify upstream dispose propagation. Consider adding a brief comment that the sink intentionally never completes (so `doOnCancel` is the only way the latch fires), otherwise this looks like an oversight at first glance.

agentscope-core/src/test/java/io/agentscope/core/agent/AgentBaseTest.java
```diff
@@ -112,6 +114,67 @@ protected Mono<Msg> handleInterrupt(InterruptContext context, Msg... originalArg
...
+            return Mono.just(
+                    Msg.builder()
+                            .name(getName())
+                            .role(MsgRole.ASSISTANT)
+                            .content(TextBlock.builder().text("Recovered").build())
+                            .build());
+        }
+
+        @Override
+        protected Mono<Msg> handleInterrupt(InterruptContext context, Msg... originalArgs) {
```
AgentScopeJavaBot, line 163: [nit] `CancellableStreamAgent#handleInterrupt` is never exercised by `testStreamCancellationReleasesExecution` (cancellation goes through `Mono.using` cleanup, not the interrupt path). It is required by the abstract contract, but the `Interrupted` literal is dead in this test class — consider returning `Mono.empty()` or reusing a shared helper, or add a follow-up test that calls `agent.interrupt()` mid-stream to actually cover the interrupt branch.

agentscope-core/src/test/java/io/agentscope/core/agent/AgentBaseTest.java
```diff
@@ -352,4 +415,32 @@ void testCheckRunningDisabled_AllowsConcurrentCall() throws InterruptedException
...
+                .verify(Duration.ofSeconds(5));
+
+        assertTrue(
+                cancellableAgent.awaitFirstCallCancelled(Duration.ofSeconds(1)),
+                "Cancelling the stream should dispose the upstream call");
+        assertTrue(
+                cancellableAgent.awaitExecutionReleased(Duration.ofSeconds(1)),
+                "Cancelling the stream should release the execution lock");
+
+        Msg response = cancellableAgent.call(inputMsg).block(Duration.ofSeconds(1));
```
AgentScopeJavaBot, line 442: [nit] `cancellableAgent.call(inputMsg).block(Duration.ofSeconds(1))` is a tight 1s timeout on a freshly-released execution lock. On a slow CI shared executor (`Schedulers.boundedElastic()` is used elsewhere in the suite) this could flake; aligning with the `Duration.ofSeconds(5)` used in the StepVerifier verify would be more conservative without weakening the assertion.

AgentScopeJavaBot (approved):
AI Review

The PR correctly fixes issue #1466 by tying the upstream inner subscription to the outer `FluxSink` lifecycle via `Disposable.Swap` and `sink.onDispose(...)`. When the downstream `Flux<Event>` is cancelled, the swap propagates dispose to the inner `Mono.defer(callSupplier).subscribe(...)`, which then cancels the `Mono.using` wrapper in `call()` and triggers `releaseExecution` (resetting `running`). The Swap idiom is the canonical Reactor pattern and correctly handles the race where the inner subscription completes synchronously before `update()` is reached (a disposed swap will immediately dispose any subsequently-added Disposable). The new regression test exercises all three guarantees (cancel propagation, execution release, and reusability after cancel) via `StepVerifier.thenCancel()` plus three `CountDownLatch`es. Streaming-hook removal still happens in `doFinally`, which is reached on every terminal signal including `CANCEL`, so hook-lifecycle and subagent-event-bus semantics are preserved.

PR #1409, opened 2026-05-16
fix(hook): skip auto-patch only when ToolResultBlock IDs match pendin… Fixes #42

sunwg2:
…g tool calls

## AgentScope-Java Version

[The version of AgentScope-Java you are working on, e.g. 1.0.12, check your pom.xml dependency version or run `mvn dependency:tree | grep agentscope-parent:pom`(only mac/linux)]

## Description

[Please describe the background, purpose, changes made, and how to test this PR]

## Checklist

Please check the following items before code is ready to be reviewed.

- [x]  Code has been formatted with `mvn spotless:apply`
- [x]  All tests are passing (`mvn test`)
- [x]  Javadoc comments are complete and follow project conventions
- [x]  Related documentation has been updated (e.g. links, examples, etc.)
- [x]  Code is ready for review

PR #1566, opened 2026-06-01
fix(hook): PendingToolRecoveryHook false positive on historical ToolResultBlocks in multi-turn (#1555)

### LINKED ISSUE #1555: [Bug]: PendingToolRecoveryHook always skips recovery for multi-turn ReActAgent (Supervisor) due to false positive userProvidedResults check
---
  Summary                                                                                                                                                                                                                                                     
                                               
  PendingToolRecoveryHook is supposed to auto-patch orphaned pending tool calls before ReActAgent.doCall() runs. However, in a Supervisor + SubAgent multi-turn scenario, it never actually patches anything — because its guard condition incorrectly treats
  historical ToolResultBlocks in memory as "user-provided results", and skips the recovery entirely. This causes ReActAgent.doCall() to always throw IllegalStateException when the Supervisor has orphaned pending tool calls.                               
  
  ---                                                                                                                                                                                                                                                         
  Exception                                    

  java.lang.IllegalStateException: Pending tool calls exist without results.
    Enable PendingToolRecoveryHook or provide tool results.                                                                                                                                                                                                   
    Pending IDs: [call_pp47mdun8f7kkz4aeka9zoe4]
      at io.agentscope.core.ReActAgent.doCall(ReActAgent.java:393)                                                                                                                                                                                            
                                                                                                                                                                                                                                                              
  ---                                                                                                                                                                                                                                                         
  Root Cause                                                                                                                                                                                                                                                  
                                               
  File: agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
  Lines: 110–113                                                                                                                                                                                                                                              
   
  // ❌ BUG: inputMessages = memory snapshot (full history) + callArgs (new user input)                                                                                                                                                                       
  // For a Supervisor agent, the memory snapshot always contains historical ToolResultBlocks                                                                                                                                                                  
  // from previous successful sub-agent calls, causing `userProvidedResults` to be always true.                                                                                                                                                               
  boolean userProvidedResults =                                                                                                                                                                                                                               
  if (userProvidedResults) {                                                                                                                                                                                                                                  
      return Mono.just(event);  // ← always skips recovery — never reaches patchPendingToolCalls()
  }                                                                                                                                                                                                                                                           
                                               
  Why inputMessages contains historical messages:                                                                                                                                                                                                             
                                               
  In AgentBase.notifyPreCall() (lines 672–678), the PreCallEvent is built with a merged view of the full memory snapshot + the new callArgs:                                                                                                                  
                                               
  // AgentBase.java:672-678                                                                                                                                                                                                                                   
  List<Msg> fullInput = new ArrayList<>(snapshot);  // full conversation history                                                                                                                                                                              
  if (callArgs != null) {
      fullInput.addAll(callArgs);                                                                                                                                                                                                                             
  }                                            
  PreCallEvent event = new PreCallEvent(this, fullInput);  // passed to all hooks
                                                                                                                                                                                                                                                              
  The intent of the guard in PendingToolRecoveryHook is to detect whether the current user input contains ToolResultBlocks (HITL scenario — user manually provides tool results). However, the check scans the entire inputMessages list, which includes the  
  memory snapshot. Any prior successful tool call in history contains a ToolResultBlock, so userProvidedResults is always true for any Supervisor with conversation history, and the recovery is always skipped.                                              
                                                                                                                                                                                                                                                              
  ---                                          
  Reproduction Scenario
                                                                                                                                                                                                                                                              
  1. Create a ReActAgent (Supervisor) with enablePendingToolRecovery(true) and at least one SubAgent tool
  2. Trigger a request where the Supervisor calls a SubAgent tool, but the SubAgent execution is interrupted or throws an exception (e.g. timeout, network error, stream disconnect)                                                                          
  3. The Supervisor's memory now has an orphaned ToolUseBlock with no corresponding ToolResultBlock                                                                                                                                                           
  4. Send a new user message in the same session                                                                                                                                                                                                              
  5. PendingToolRecoveryHook.handlePreCall() is invoked:                                                                                                                                                                                                      
    - pendingIds is non-empty ✓                                                                                                                                                                                                                               
    - inputMessages is non-empty ✓                                                                                                                                                                                                                            
    - userProvidedResults = true ← false positive, caused by historical ToolResultBlocks in the snapshot                                                                                                                                                      
    - Recovery is skipped                                                                                                                                                                                                                                     
  6. ReActAgent.doCall() detects pendingIds is still non-empty → throws IllegalStateException                                                                                                                                                                 
                                                                                                                                                                                                                                                              
  ---                                                                                                                                                                                                                                                         
  Fix                                                                                                                                                                                                                                                         
                                               
  The check should only inspect the new user input (i.e., messages beyond the memory snapshot boundary), not the full merged inputMessages. There are two clean approaches:
                                                                                                                                                                                                                                                              
  Option A — Filter by message role (exclude ASSISTANT/TOOL history):                                                                                                                                                                                         
                                                                                                                                                                                                                                                              
  // PendingToolRecoveryHook.java:110-113                                                                                                                                                                                                                     
  // ✅ Only check messages that are genuinely new user input, not historical assistant/tool messages
  boolean userProvidedResults = inputMessages.stream()                                                                                                                                                                                                        
          .filter(m -> m.getRole() != MsgRole.ASSISTANT && m.getRole() != MsgRole.TOOL)
          .anyMatch(m -> m.hasContentBlocks(ToolResultBlock.class));                                                                                                                                                                                          
                                               
  Option B — Expose snapshotSize on PreCallEvent and only check the tail:                                                                                                                                                                                     
                                               
  // PreCallEvent: add getSnapshotSize() or getCallArgs()                                                                                                                                                                                                     
  // PendingToolRecoveryHook:                                                                                                                                                                                                                                 
  int snapshotSize = event.getSnapshotSize();
  List<Msg> callArgs = inputMessages.subList(snapshotSize, inputMessages.size());                                                                                                                                                                             
  boolean userProvidedResults = callArgs.stream()
          .anyMatch(m -> m.hasContentBlocks(ToolResultBlock.class));                                                                                                                                                                                          
                                                                                                                                                                                                                                                              
  Option A is the minimal, non-breaking fix. Option B is cleaner and makes the intent more explicit.                                                                                                                                                          
                                                                                                                                                                                                                                                              
  ---                                                                                                                                                                                                                                                         
  Impact                                       
        
  - Severity: High — PendingToolRecoveryHook is silently non-functional for all ReActAgent instances with conversation history
  - Affected scenarios: Any multi-turn Supervisor + SubAgent setup where a SubAgent call is interrupted (timeout, exception, stream disconnect)                                                                                                               
  - enablePendingToolRecovery(true) has no effect in practice for these cases — the agent always crashes on the next user message
                                                                                                                                                                                                                                                              
  ---                                          
  Environment                                                                                                                                                                                                                                                 
                                               
  - agentscope-core (agentscope-java)
  - Affected classes:                                                                                                                                                                                                                                         
    - io.agentscope.core.hook.PendingToolRecoveryHook — line 110
    - io.agentscope.core.agent.AgentBase#notifyPreCall — line 672 (context)

Sparkle6979:
## Summary

`PendingToolRecoveryHook` is silently non-functional in multi-turn conversations.

### Root Cause

`AgentBase.notifyPreCall()` (line 672-678) merges the full memory snapshot with new callArgs into `PreCallEvent.inputMessages`. The  hook's guard condition at line 110 checks this merged list for any `ToolResultBlock`:

  ```java
  boolean userProvidedResults =
  if (userProvidedResults) return;  // skip recovery
  ```

  In multi-turn scenarios, memory always contains ToolResultBlocks from earlier successful tool calls, so userProvidedResults is always  true — the hook always skips recovery, and ReActAgent.doCall() throws IllegalStateException.

### Fix

  Narrow the check to only match ToolResultBlock IDs present in the pendingIds set:

  ```java
  boolean userProvidedResults =
      inputMessages.stream()
          .flatMap(m -> m.getContentBlocks(ToolResultBlock.class).stream())
          .anyMatch(tr -> pendingIds.contains(tr.getId()));
  ```

  Historical ToolResultBlocks (e.g. call_1) are for already-resolved tool calls, so their IDs are never in pendingIds — no false positive. HITL results (user-provided ToolResultBlock for a pending call) do match pendingIds, so the hook correctly defers to doCall().

### Why existing tests pass

  HookStopAgentTest.testNewMsgWithPendingToolUseContinuesActing only covers single-turn: memory has no historical ToolResultBlock, so the false positive never triggers.

  Closes #1555

### Test plan

  - Core bug: multi-turn plain text recovery with historical ToolResultBlocks
  - Accumulated history: 3+ successful tool turns → interrupted → recovery
  - HITL single-turn and multi-turn: user provides matching result
  - Unrelated ToolResultBlock ID does not block auto-recovery
  - Single-turn regression: classic recovery still works
  - Normal multi-turn without pending calls unaffected
  - Multiple pending calls: all auto-recovered (single-turn and multi-turn)
  - Existing HookStopAgentTest passes (no regression)


Sparkle6979:
## Design note: why this approach vs suggested alternatives

The issue description proposed two options:

**Option A** (filter by role): filter out ASSISTANT/TOOL messages, then check for ToolResultBlock. This is fragile — in HITL scenarios, user-provided ToolResultBlocks also carry TOOL roles, so the filter does not reliably distinguish historical results from user-provided results.

**Option B** (expose snapshotSize on PreCallEvent): requires an API change to PreCallEvent, which is more invasive.

**This PR approach** (match pendingIds): directly checks whether any ToolResultBlock ID matches a pending tool call. Historical results are never in pendingIds, so no false positive. HITL results for a pending call DO match, so they correctly defer to doCall(). No API change, 3 lines modified.

agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
```diff
@@ -108,7 +108,9 @@ private Mono<PreCallEvent> handlePreCall(PreCallEvent event) {
         }
 
         boolean userProvidedResults =
-                inputMessages.stream().anyMatch(m -> m.hasContentBlocks(ToolResultBlock.class));
+                inputMessages.stream()
+                        .flatMap(m -> m.getContentBlocks(ToolResultBlock.class).stream())
+                        .anyMatch(tr -> pendingIds.contains(tr.getId()));
```
AgentScopeJavaBot, line 113: [praise] Good fix. Matching `ToolResultBlock.getId()` against `pendingIds` is the correct narrowing of the guard: `pendingIds` is precisely the set of unmatched ids from the last assistant message, so historical (already-resolved) ToolResultBlocks in the memory snapshot cannot collide with it, while a user-supplied HITL result for a pending id still defers correctly. The previous `anyMatch(m -> m.hasContentBlocks(ToolResultBlock.class))` was effectively a global presence check — clearly wrong once memory accumulated history.

agentscope-core/src/test/java/io/agentscope/core/hook/PendingToolRecoveryMultiTurnTest.java
```diff
@@ -0,0 +1,611 @@
...
+            Msg recovery = createAssistantTextMsg("Recovered");
+
+            when(mockModel.stream(anyList(), anyList(), any()))
+                    .thenReturn(createFluxFromMsg(toolCall))
+                    .thenReturn(createFluxFromMsg(recovery));
+
+            Hook stopHook = createAlwaysStopHook();
+
+            ReActAgent agent =
+                    ReActAgent.builder()
```
AgentScopeJavaBot, line 351: [praise] Nice negative test: a user-provided ToolResultBlock with an unrelated id (`call_999`) must not be misread as HITL for `call_1`. Under the previous logic this would have silently disabled recovery; the new id-set check now correctly auto-patches the truly pending call. This case is the most direct guard against regressions.

agentscope-core/src/test/java/io/agentscope/core/hook/PendingToolRecoveryMultiTurnTest.java
```diff
@@ -0,0 +1,611 @@
...
+            Msg toolCall = createToolUseMsg("call_1", "test_tool");
+            Msg textAfterResume = createAssistantTextMsg("Resumed with user result");
+
+            when(mockModel.stream(anyList(), anyList(), any()))
+                    .thenReturn(createFluxFromMsg(toolCall))
+                    .thenReturn(createFluxFromMsg(textAfterResume));
+
+            Hook stopHook = createAlwaysStopHook();
+
+            ReActAgent agent =
```
AgentScopeJavaBot, line 215: [praise] `testManyTurnsThenRecovery` accumulates 3 historical ToolResultBlocks before triggering recovery, which exercises exactly the multi-turn shape from #1555. Combined with the single-turn regression test in section D, this gives confidence the fix doesn't only patch the symptom for one specific turn count.

agentscope-core/src/test/java/io/agentscope/core/hook/PendingToolRecoveryMultiTurnTest.java
```diff
@@ -0,0 +1,611 @@
...
+    }
+
+    static class TestTool {
+        private final AtomicBoolean executed;
+
+        TestTool(AtomicBoolean executed) {
+            this.executed = executed;
+        }
+
+        @Tool(name = "test_tool", description = "A test tool")
```
AgentScopeJavaBot, line 605: [nit] `createConditionalStopHook(counter, stopAt)` couples the test to internal `PostReasoningEvent` firing order (the `stopAt=3`/`stopAt=7` magic numbers). It works today but will be brittle if reasoning-event semantics evolve. Consider stopping based on a structural signal (e.g. once a specific `ToolUseBlock` id appears) in a follow-up to make these tests more refactor-resilient. Non-blocking.

agentscope-core/src/main/java/io/agentscope/core/hook/PendingToolRecoveryHook.java
AgentScopeJavaBot, line 113: [nit] Defensive consideration only (not a real bug today): `ToolResultBlock.getId()` is documented as nullable, and `Set#contains(null)` is a legal lookup. In normal flows `pendingIds` is sourced from non-null `ToolUseBlock` ids so a null result-id cannot match, but if a malformed user message ever ships a null-id `ToolResultBlock` the hook will silently ignore it (treat as not-HITL and auto-patch). That is arguably the safer behavior, so no change required — flagging only so the assumption is explicit.

AgentScopeJavaBot (approved):
🤖 AI Review

The fix is precise and correctly addresses the multi-turn false positive described in #1555. By narrowing the HITL guard from "any ToolResultBlock present" to "ToolResultBlock id ∈ pendingIds", the hook now correctly distinguishes user-supplied results for the currently pending tool call from historical results already resolved in earlier turns. Since `pendingIds` is built from `ToolUseBlock` ids on the *last* assistant message that have no matching result, historical (already-resolved) ids cannot collide with it, so the new condition is logically tight. The HITL deferral path remains intact: when the user sends a matching id the hook still returns the event unmodified for `ReActAgent.doCall()` to consume. No new null-safety risk of practical concern — `pendingIds` is sourced from non-null `ToolUseBlock::getId`, and `getContentBlocks` always returns a non-null list. Test coverage is thorough and well-partitioned (core bug, HITL, unrelated id, single-turn regression, multi-pending). No issues found that would block merge.

LearningGp:
FYI #1409

Sparkle6979:
Noticed that `PendingToolRecoveryHook.java` has been removed from main and replaced by `LegacyHookDispatcher`. Closing this PR as the target file no longer exists.

Sparkle6979:
Checked the main branch — `PendingToolRecoveryHook.java` no longer exists under `agentscope-core/src/main/java/io/agentscope/core/hook/`. The hook directory now contains `LegacyHookDispatcher.java` instead.

If this file is still used somewhere (e.g. on a release branch), happy to rebase. Otherwise feel free to close.

LearningGp:
PTAL @chickenlj

oss-maintainer:

```bash
git fetch origin
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #1570, opened 2026-06-02
Optimize load skill

xingkaima2023:
## AgentScope-Java Version

## Description

## Checklist

- [ x]  Code has been formatted with `mvn spotless:apply`
- [ x]  All tests are passing (`mvn test`)
- [ x]  Javadoc comments are complete and follow project conventions
- [ x]  Related documentation has been updated (e.g. links, examples, etc.)
- [ x]  Code is ready for review


agentscope-core/src/test/java/io/agentscope/core/skill/SkillToolFactoryTest.java
```diff
@@ -0,0 +1,394 @@
...
+
+    @Nested
+    @DisplayName("Tool metadata")
+    class ToolMetadataTests {
+
+        @Test
+        @DisplayName("Should have correct tool name")
+        void shouldHaveCorrectToolName() {
+            AgentTool tool = createSkillAccessTool();
+            assertTrue("load_skill_through_path".equals(tool.getName()));
```
AgentScopeJavaBot, line 369: [nit] Prefer `assertEquals` over `assertTrue("...".equals(...))` so failure messages show the expected vs. actual values instead of just `expected: <true> but was: <false>`. Suggestion:

```java
assertEquals("load_skill_through_path", tool.getName());
```

agentscope-core/src/test/java/io/agentscope/core/skill/SkillToolFactoryTest.java
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        @Test
+        @DisplayName("Should have non-empty description")
+        void shouldHaveNonEmptyDescription() {
+            AgentTool tool = createSkillAccessTool();
+            assertNotNull(tool.getDescription());
+            assertTrue(tool.getDescription().length() > 0);
```
AgentScopeJavaBot, line 377: [nit] `assertTrue(s.length() > 0)` is the classic anti-pattern equivalent of `assertFalse(s.isEmpty())`. Prefer the more idiomatic form (or AssertJ `assertThat(...).isNotEmpty()`) for clearer failure output:

```java
assertFalse(tool.getDescription().isEmpty());
```

agentscope-core/src/main/java/io/agentscope/core/skill/SkillToolFactory.java
```diff
@@ -172,15 +172,13 @@ private String loadSkillResourceImpl(String skillId, String path) {
             return buildSkillMarkdownResponse(skillId, skill);
         }
 
-        // Get resource
-        Map<String, String> resources = skill.getResources();
-        if (resources == null || !resources.containsKey(path)) {
+        // Get resource using getResource method
```
AgentScopeJavaBot, line 175: [nit] Comment `// Get resource using getResource method` is somewhat tautological with the code below. Consider explaining the *why* (e.g. avoiding the defensive HashMap copy returned by `getResources()`) or just dropping the comment, since the call site is self-explanatory.

agentscope-core/src/main/java/io/agentscope/core/skill/SkillToolFactory.java
```diff
@@ -172,15 +172,13 @@ private String loadSkillResourceImpl(String skillId, String path) {
...
+        String resourceContent = skill.getResource(path);
+        if (resourceContent == null) {
             // Resource not found, return available resource paths
             throw new IllegalArgumentException(
-                    buildResourceNotFoundMessage(skillId, path, resources));
+                    buildResourceNotFoundMessage(skillId, path, skill.getResources()));
```
AgentScopeJavaBot, line 180: [minor] On the error path you now call `skill.getResources()` solely to enumerate keys inside `buildResourceNotFoundMessage`, but that method only reads `resources.keySet()`. Since `getResources()` always returns a fresh defensive copy of the entire HashMap, you allocate the whole map just to throw away the values. Consider refactoring `buildResourceNotFoundMessage` to accept `Set<String> resourcePaths` and pass `skill.getResourcePaths()` instead — keeps the optimization story consistent on the error path too. Non-blocking since this path is rare.

agentscope-core/src/main/java/io/agentscope/core/skill/SkillToolFactory.java
```diff
@@ -172,15 +172,13 @@ private String loadSkillResourceImpl(String skillId, String path) {
             return buildSkillMarkdownResponse(skillId, skill);
         }
 
-        // Get resource
-        Map<String, String> resources = skill.getResources();
-        if (resources == null || !resources.containsKey(path)) {
+        // Get resource using getResource method
+        String resourceContent = skill.getResource(path);
```
AgentScopeJavaBot, line 176: [praise] Good change — uses the same `getResource(path)` pattern already in `SkillBox.java:757`, eliminates the redundant defensive copy of the entire resources map in the happy path, and as a side-effect makes the lookup robust to a legitimate-but-unlikely edge case where `resources` contains a key mapped to a null value (the old `containsKey` branch would have called `get` and silently emitted null content to the user; the new code correctly reports it as not found).

AgentScopeJavaBot (approved):
🤖 AI Review

This PR is a small, focused optimization that replaces the `getResources() + containsKey() + get()` sequence with a single `getResource(path)` call in `loadSkillResourceImpl`. Since `AgentSkill.getResources()` returns a defensive copy of the underlying HashMap, the new code avoids an unnecessary full-map copy in the happy path, while keeping behavior equivalent (and even slightly more robust against null values in the resources map). The change also aligns with the pattern already used in `SkillBox.getResource()`. A comprehensive new test file (`SkillToolFactoryTest`) is added covering SKILL.md loading, resource lookup, error paths, parameter validation, and tool metadata, with proper Mockito strict-stub discipline. Note that `SkillToolFactory`/`AgentSkill` are marked `@Deprecated(since = "2.0.0")`, so the value of optimizing this code is limited, but the change itself is correct and well-tested. Only minor JUnit assertion style nits in the test file.

oss-maintainer:

```bash
git fetch origin
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #1589, opened 2026-06-04
Fix Nacos skill marketplace latest label resolution

lancefoever:
## AgentScope-Java Version

main branch

### Background

When using the Builder Nacos marketplace, skill preview/install can fail while loading a skill version detail.

The previous implementation passed the label name `LATEST` directly as the `version` argument to `getSkillVersionDetail(namespaceId, skillName, version)`. However, the Nacos skill version detail API expects the actual skill version value, not the label name.

This could produce a request equivalent to:

```text
/nacos/v3/admin/ai/skills/version?skillName=<skillName>&namespaceId=<namespaceId>&version=LATEST
```

Nacos then returns an error similar to:

```text
Skill version not found: <skillName>@LATEST
```

### Purpose

This PR fixes Nacos-backed Builder marketplace preview/install by resolving the real skill version from the skill metadata labels before loading the skill version detail.

### Changes Made

- Load skill metadata through the Nacos maintainer client.
- Resolve the actual skill version from the `LATEST` label in the skill metadata.
- Use the resolved version value when calling `getSkillVersionDetail(...)`.
- Add validation so a clear error is raised if the `LATEST` label cannot be resolved.

### How to Test

1. Start a Nacos server with Skill Registry support.
2. Create or publish a Nacos skill with a `LATEST` label pointing to an online skill version.
3. Add the Nacos marketplace in Builder.
4. Preview the Nacos skill from Builder.
5. Install the skill into an agent.

Expected result:

- Builder no longer requests skill details with `version=LATEST`.
- Builder resolves the actual skill version from Nacos skill metadata.
- Skill preview succeeds.
- Skill installation succeeds.
- The installed skill appears in the agent's skills list.

Local validation:

```bash
mvn -q -pl agentscope-examples/agents/agentscope-builder -am spotless:apply
mvn -q -pl agentscope-examples/agents/agentscope-builder -am -DskipTests compile
```

## Checklist

- [ ] Javadoc comments are complete and follow project conventions


agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/runtime/marketplace/NacosBuilderMarketplace.java
```diff
@@ -156,8 +157,24 @@ public MarketSkillContent fetch(String name) {
             return null;
         }
         try {
-            Skill skill =
-                    service.skill().getSkillVersionDetail(namespaceId, name.trim(), LATEST_VERSION);
+            String skillName = name.trim();
+            SkillMeta skillMeta = service.skill().getSkillMeta(namespaceId, skillName);
+            String version =
+                    skillMeta != null && skillMeta.getLabels() != null
+                            ? skillMeta.getLabels().get(LATEST_VERSION)
```
AgentScopeJavaBot: [critical] Label key case mismatch. Nacos's canonical label for the latest version is `"latest"` (lowercase), defined as `AiResourceConstants.LABEL_LATEST = "latest"` and set by `SkillOperationServiceImpl#publish(..., updateLatestLabel=true)` (also `forcePublish`, `publishApprovedBySystem`, `bootstrapSkillFromZip`). This lookup uses the existing `LATEST_VERSION = "LATEST"` constant (uppercase), so for any skill published through the standard Nacos flow `skillMeta.getLabels().get("LATEST")` will return null and the new `IllegalStateException` ("Cannot resolve Nacos skill version from label LATEST") will be thrown — the bug is shifted, not fixed. Suggested change:

```java
private static final String LATEST_LABEL = "latest"; // matches Nacos AiResourceConstants.LABEL_LATEST
...
String version = skillMeta != null && skillMeta.getLabels() != null
        ? skillMeta.getLabels().get(LATEST_LABEL)
        : null;
```

The test steps in the PR description ("create a Nacos skill with a `LATEST` label") effectively required manually creating a non-standard uppercase label, which masked this in local testing. Please retest end-to-end against a skill published via `publish(..., updateLatestLabel=true)`.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/runtime/marketplace/NacosBuilderMarketplace.java
```diff
@@ -156,8 +157,24 @@ public MarketSkillContent fetch(String name) {
...
+                            : null;
+
+            if (version == null || version.isBlank()) {
+                throw new IllegalStateException(
+                        "Cannot resolve Nacos skill version from label "
+                                + LATEST_VERSION
```
AgentScopeJavaBot: [minor] The `LATEST_VERSION` constant is now used as a *label key*, not a version string passed to the API. Reusing the old name makes the error message and the lookup at line 164 misleading and bundles two distinct concerns under one constant. Suggest renaming to e.g. `LATEST_LABEL` (and fixing the value per the critical comment), and updating this error message to reflect that the issue is a missing label on the skill metadata, e.g. `"Nacos skill metadata is missing a '" + LATEST_LABEL + "' label for " + namespaceId + "/" + skillName + "; cannot resolve latest version"`.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/runtime/marketplace/NacosBuilderMarketplace.java
```diff
@@ -156,8 +157,24 @@ public MarketSkillContent fetch(String name) {
             return null;
         }
         try {
-            Skill skill =
-                    service.skill().getSkillVersionDetail(namespaceId, name.trim(), LATEST_VERSION);
+            String skillName = name.trim();
+            SkillMeta skillMeta = service.skill().getSkillMeta(namespaceId, skillName);
```
AgentScopeJavaBot: [minor] The class-level Javadoc (lines 36-44, outside this hunk) still says: `pull SKILL.md via {@code getSkillVersionDetail(..., "LATEST")}`. After this change the actual version is resolved from the skill metadata's `latest` label first; please update that Javadoc so it doesn't keep advertising the broken behavior. Also consider caching `service.skill()` in a local var since it is now invoked twice in this method (lines 161 and 177).

AgentScopeJavaBot: [minor] `getSkillMeta` returns the skill metadata but does not validate that the resolved version is actually `online`. If a stale `latest` label points at a version that has since been taken offline (`VERSION_STATUS_OFFLINE`) or is still in `draft`/`reviewing`, the subsequent `getSkillVersionDetail` will either fail with a less clear error or return content the marketplace shouldn't surface. Optional, but worth verifying the version's status (via `skillMeta.getVersions()`) before fetching its detail, or at minimum documenting that the marketplace trusts the `latest` label blindly.

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR correctly identifies the root cause — Nacos's `getSkillVersionDetail` API expects an actual version string, not a label name — and routes the call through `getSkillMeta` first to resolve a real version from the labels map. The control flow, null/blank guards, and the clear `IllegalStateException` on unresolvable labels are all reasonable.

However, the fix as written likely will not work end-to-end against Nacos's standard publish flow. In Nacos (`com.alibaba.nacos.ai.constant.AiResourceConstants`) the canonical label key for the latest published version is `"latest"` (lowercase), set automatically by `SkillOperationServiceImpl#publish(...)` when `updateLatestLabel=true`. The PR reuses the existing `LATEST_VERSION = "LATEST"` (uppercase) constant as the lookup key into `skillMeta.getLabels()`, which will miss the canonical label and trip the new `IllegalStateException` for any skill published the normal way. The test steps in the PR description ("create a skill with a `LATEST` label") effectively work around the bug by asking the tester to set a non-standard label.

Secondary items: the class Javadoc still describes the old `getSkillVersionDetail(..., "LATEST")` behavior and should be refreshed; the constant `LATEST_VERSION` is now semantically a label key, not a version, so renaming it (e.g. to `LATEST_LABEL`) keeps intent clear.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/runtime/marketplace/NacosBuilderMarketplace.java
AgentScopeJavaBot: ✅ Verified as addressed in `fa9693a`: Constant renamed to `LATEST_LABEL = "latest"` (lowercase) and label lookup now uses `LATEST_LABEL` — case mismatch bug is fixed.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/runtime/marketplace/NacosBuilderMarketplace.java
AgentScopeJavaBot: ✅ Verified as addressed in `fa9693a`: Constant renamed from `LATEST_VERSION` to `LATEST_LABEL` and error message updated to reflect missing label context.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/runtime/marketplace/NacosBuilderMarketplace.java
AgentScopeJavaBot: ✅ Verified as addressed in `fa9693a`: Class Javadoc updated to describe the new label-resolution flow, and `service.skill()` cached in local var `skillService`.

AgentScopeJavaBot: ✅ Verified as addressed in `fa9693a`: New `resolveLatestSkillVersion` method validates version status is online before returning.

AgentScopeJavaBot (approved):

oss-maintainer:

```bash
git fetch origin
git rebase origin/main
git push --force-with-lease
```

---

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/runtime/marketplace/NacosBuilderMarketplace.java
```diff
@@ -156,8 +159,13 @@ public MarketSkillContent fetch(String name) {
             return null;
         }
         try {
+            String skillName = name.trim();
+            SkillMaintainerService skillService = service.skill();
+            SkillMeta skillMeta = skillService.getSkillMeta(namespaceId, skillName);
```
oss-maintainer, line 164: **[Warning]** `fetch()` now costs two round-trips per skill (`getSkillMeta` + `getSkillVersionDetail`). `fetch` is called per skill from the builder sync path, so a marketplace with hundreds of skills doubles the RPC count against Nacos. Consider caching `SkillMeta` per (namespace, name) for the lifetime of one sync/list pass, or resolving the label only when `getSkillVersionDetail(..., "LATEST")` is not supported by the target server.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/runtime/marketplace/NacosBuilderMarketplace.java
```diff
@@ -202,4 +210,62 @@ public void close() {
     private static String blankToNull(String v) {
         return (v == null || v.isBlank()) ? null : v.trim();
     }
+
+    private static String resolveLatestSkillVersion(
+            String namespaceId, String skillName, SkillMeta skillMeta) {
+        String version =
```
oss-maintainer, line 216: **[Warning]** This turns a previously soft failure into a hard one. Before, an unresolvable skill still attempted `getSkillVersionDetail(..., "LATEST")`; now a missing `latest` label throws `IllegalStateException` out of `fetch()`. Callers that treat `null` as "skill not available" (and skip it) will instead see the whole fetch fail. Could you either (a) fall back to the old `"LATEST"` version string when the label is absent, or (b) return `null` for the "not labelled / not online" cases and keep throwing only for real transport errors? If the hard failure is intentional, please note it in the PR description so the caller contract change is explicit.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/runtime/marketplace/NacosBuilderMarketplace.java
```diff
@@ -202,4 +210,62 @@ public void close() {
...
+                            + skillName
+                            + "; metadata versions are empty");
+        }
+
+        for (SkillMeta.SkillVersionSummary summary : skillMeta.getVersions()) {
+            if (summary == null || !version.equals(summary.getVersion())) {
+                continue;
+            }
+            String status = summary.getStatus();
+            if (!ONLINE_STATUS.equalsIgnoreCase(status)) {
```
oss-maintainer, line 247: **[Info]** `ONLINE_STATUS.equalsIgnoreCase(status)` treats `status == null` as "not online" and throws with `status=null`. That is a reasonable default, but the message is confusing for an unset status — consider wording it as "status is unknown/unset" so operators can tell "offline" apart from "server did not report a status". Also worth a comment naming the Nacos version that introduced the `online` value, since older servers may use a different vocabulary.

oss-maintainer (commented):
### Summary
Resolves the Nacos skill's real version from the `latest` label in `SkillMeta` instead of passing the literal `"LATEST"` to `getSkillVersionDetail` — a sensible fix for marketplaces where the server rejects the pseudo-version. The resolution logic itself is correct and the error messages are genuinely helpful for operators.

### Findings
- **[Warning]** NacosBuilderMarketplace.java:164 — one extra Nacos RPC per `fetch()`; consider caching `SkillMeta` per pass.
- **[Warning]** NacosBuilderMarketplace.java:216 — "no `latest` label" changes `fetch()` from a soft failure to a thrown `IllegalStateException`; please confirm the caller contract (fallback to `LATEST` / return `null`).
- **[Warning]** NacosBuilderMarketplace.java:214 — no tests for the new branch-heavy resolution logic.
- **[Info]** NacosBuilderMarketplace.java:247 — `status == null` and "not online" share one message; also document which Nacos versions use `online`.

### Suggestions
```java
// Soft degradation keeps the previous caller contract while still preferring the label:
String resolvedVersion = resolveLatestSkillVersionOrNull(namespaceId, skillName, skillMeta);
if (resolvedVersion == null) {
    log.debug("No resolvable latest label for {}/{}, falling back to LATEST", namespaceId, skillName);
    resolvedVersion = LATEST_VERSION;   // keep the old constant as a fallback
}
```

### Notes
This PR is currently `CONFLICTING` with `main` (a rebase reminder was already posted on 2026-09-18), so it cannot be merged as-is. CLA is signed.

---

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/runtime/marketplace/NacosBuilderMarketplace.java
```diff
@@ -202,4 +210,62 @@ public void close() {
     private static String blankToNull(String v) {
         return (v == null || v.isBlank()) ? null : v.trim();
     }
+
+    private static String resolveLatestSkillVersion(
```
oss-maintainer, line 214: **[Warning]** No tests accompany the new resolution logic. The four branches here (label missing, `versions` empty, label points at an unknown version, version not online, happy path) are exactly the cases that regress silently when the Nacos client model changes. Could you add unit tests for `resolveLatestSkillVersion` with a stubbed `SkillMeta`? It is `private static`, so making it package-private (`@VisibleForTesting`) would be enough to test it directly.

PR #1607, opened 2026-06-04
fix(middleware): rebind toolkit reference after defensive deep copy

### LINKED ISSUE #1606: [Bug]: ReActAgent.Builder.build() toolkit deep copy causes dynamically-registered middleware tools to be missing from the agent's actual toolkit
### Summary

`ReActAgent.Builder.build()` performs a defensive `toolkit.copy()` (line 3077), which creates a second copy of the toolkit. When a middleware (e.g., `HarnessSkillMiddleware`) holds a reference to the first copy and dynamically registers tools onto it during `call()`, those tools are invisible to the agent because the agent uses the second copy. This results in `Tool not found` errors at runtime.

### Environment

- AgentScope Java: current main
- Component: `agentscope-core` (`ReActAgent`), `agentscope-harness` (`HarnessAgent`, `HarnessSkillMiddleware`)

### Root Cause

In `HarnessAgent.Builder.build()` (HarnessAgent.java ~L1520-1902):

```java
// Step 1: First defensive copy
Toolkit agentToolkit = this.toolkit.copy();          // ← Copy 1

// Step 2: Middleware receives Copy 1
inner.middleware(
    new HarnessSkillMiddleware(
        orderedSkillRepos,
        agentToolkit,                                 // ← Copy 1 passed to middleware
        skillFilter, ...));

// Step 3: Pass Copy 1 to inner builder
inner.toolkit(agentToolkit);                          // ← Copy 1

// Step 4: inner.build() does ANOTHER copy
ReActAgent delegate = inner.build();
// → ReActAgent.Builder.build() line 3077:
//    Toolkit agentToolkit = this.toolkit.copy();     // ← Copy 2 (the one the agent actually uses)
```

The lifecycle at `call()` time:

```
agent.call()
  → seedSystemMsg()
    → applySystemPromptMiddlewares()
      → HarnessSkillMiddleware.onSystemPrompt()
        → runtime.install(catalog, toolkit)
          → toolkit.registerAgentTool(loadTool)       // ← Registered on Copy 1
  → ...
  → toolkit.getToolSchemas()                          // ← Reads from Copy 2 → load_skill_through_path is MISSING
  → LLM generates tool_call: load_skill_through_path
  → Toolkit.getTool("load_skill_through_path")        // ← Looks up in Copy 2 → returns null
  → "Error: Tool not found: load_skill_through_path"
```

### Reproduction Steps

1. Build a `HarnessAgent` with `skillRepository(...)` configured
2. Call `agent.call()` with a user message
3. The LLM sees the `<available_skills>` prompt block (rendered by middleware) and generates a `load_skill_through_path` tool call
4. The framework returns `"Error: Tool not found: load_skill_through_path"`

### Debug Evidence

Setting breakpoints at the following locations confirms the issue:

| # | Class | Line | Observation |
|---|-------|------|-------------|
| 1 | `HarnessSkillMiddleware` | `mergeRepositories()` | `repo.getAllSkills()` returns 2 skills — **correct** |
| 2 | `HarnessSkillMiddleware` | `onSystemPrompt()` ~L177 | `catalog.ids()` contains 2 skill IDs — **correct** |
| 3 | `SkillLoadTool` | `getParameters()` | **Breakpoint never hit** — `ToolSchemaProvider` uses `registered.getExtendedParameters()` (captured at registration time), not `tool.getParameters()` |
| 4 | `ToolSchemaProvider` | `getToolSchemas()` | Returns 17 tools — `load_skill_through_path` is **absent** |
| 5 | `ReActAgent` | ~L1045 | `toolkit.getToolSchemas()` returns 17 tools — `load_skill_through_path` is **absent** |
| 6 | `Toolkit` | `getTool("load_skill_through_path")` | `toolRegistry` has 17 tools, `load_skill_through_path` is **not among them** |

The middleware registers `load_skill_through_path` on Copy 1, but the agent's `toolkit` (Copy 2) never receives it.

### Impact

- Any middleware that dynamically registers tools during `onSystemPrompt()` or `onReasoning()` is affected, not just `HarnessSkillMiddleware`
- The `DynamicSkillMiddleware` in `agentscope-core` (ReActAgent.java ~L3104-3110) has the same problem — it receives `agentToolkit` (which is already a copy) and registers tools onto it, but the agent uses a further copy
- This makes the entire skill-loading feature non-functional for `HarnessAgent`

### Suggested Fix

**Option A (Recommended):** In `ReActAgent.Builder.build()`, after creating the deep copy, propagate the copy back to any middleware that holds a reference to the old toolkit. For example, add a `rebindToolkit(Toolkit newToolkit)` method to `MiddlewareBase` (default no-op), and call it for each middleware after the copy:

```java
// ReActAgent.Builder.build()
Toolkit agentToolkit = this.toolkit.copy();

// Rebind middlewares to the new toolkit copy
for (MiddlewareBase mw : middlewares) {
    mw.rebindToolkit(agentToolkit);  // default no-op, HarnessSkillMiddleware overrides
}
```

**Option B:** In `HarnessAgent.Builder.build()`, skip the first copy and pass the builder's toolkit directly to `inner`, letting `ReActAgent.Builder.build()` be the single point of deep copy. Then pass the same `agentToolkit` reference (the one from `inner.build()`) to the middleware. This requires restructuring the build sequence.

**Option C (Minimal):** In `HarnessSkillMiddleware.onSystemPrompt()`, instead of registering the tool on the toolkit passed at construction time, resolve the agent's actual toolkit from the `Agent` parameter and register there:

```java
@Override
public Mono<String> onSystemPrompt(Agent agent, String currentPrompt) {
    // Resolve the agent's actual toolkit instead of using the construction-time reference
    Toolkit actualToolkit = resolveToolkit(agent);
    runtime.install(catalog, actualToolkit);
    ...
}
```

This requires `AgentBase` or `ReActAgent` to expose a `getToolkit()` method (which `HarnessAgent` already does, but `Agent` interface does not).

### Workaround

We currently work around this by using reflection after `agentBuilder.build()` to extract the `SkillLoadTool` from the middleware's `SkillRuntime` and register it onto the agent's actual toolkit:

```java
HarnessAgent agent = agentBuilder.build();
fixSkillLoadToolAfterBuild(agent);

private void fixSkillLoadToolAfterBuild(HarnessAgent agent) {
    ReActAgent delegate = agent.getDelegate();
    Field middlewaresField = ReActAgent.class.getDeclaredField("middlewares");
    middlewaresField.setAccessible(true);
    List<MiddlewareBase> middlewares = (List<MiddlewareBase>) middlewaresField.get(delegate);
    for (MiddlewareBase mw : middlewares) {
        if (mw.getClass().getSimpleName().equals("HarnessSkillMiddleware")) {
            Method runtimeMethod = mw.getClass().getMethod("runtime");
            Object runtime = runtimeMethod.invoke(mw);
            Method loadToolMethod = runtime.getClass().getMethod("loadTool");
            Object loadTool = loadToolMethod.invoke(runtime);
            Toolkit actualToolkit = agent.getToolkit();
            if (actualToolkit.getTool("load_skill_through_path") == null) {
                actualToolkit.registerAgentTool((AgentTool) loadTool);
            }
            return;
        }
    }
}
```

This works because `SkillLoadTool` holds an `AtomicReference<SkillCatalog>` shared with the middleware, so catalog updates from subsequent `onSystemPrompt()` calls are automatically visible.

lwx11111:
The user wants me to write a PR title and description for the fix they've been working on. Let me craft a good PR title and description based on the context.
**Title:**
```
fix(middleware): rebind toolkit reference after defensive deep copy
```

**Description:**

```markdown
## Problem

When building a `ReActAgent`, the `Builder.build()` method performs a defensive deep copy of the `Toolkit` via `toolkit.copy()`. However, middlewares (e.g. `HarnessSkillMiddleware`, `DynamicSkillMiddleware`) hold a reference to the **original** toolkit instance. After the copy, the agent operates on the new toolkit copy while middlewares still reference the old one.

This causes tools dynamically registered by middlewares (such as `load_skill_through_path` registered by `HarnessSkillMiddleware`) to be missing from the agent's actual toolkit, resulting in "Tool not found" errors at runtime.

## Root Cause

```
Builder.build()
  ├── middleware A registers tool on toolkit (original)
  ├── toolkit.copy() → agentToolkit (copy)
  ├── agent uses agentToolkit ← has the tool
  └── middleware A still references original toolkit ← tool registered here, not on copy
```

## Solution

1. Add a `default void rebindToolkit(Toolkit newToolkit)` method to `MiddlewareBase` (no-op by default).
2. In `ReActAgent.Builder.build()`, after `toolkit.copy()`, iterate all middlewares and call `rebindToolkit(agentToolkit)` so they update their toolkit reference to the copy.
3. Override `rebindToolkit` in `HarnessSkillMiddleware` and `DynamicSkillMiddleware` to rebind their `toolkit` field (also changed from `final` to `volatile` for thread safety).

## Changes

- `MiddlewareBase.java`: Add `rebindToolkit(Toolkit)` default method
- `ReActAgent.java`: Call `rebindToolkit` on all middlewares after `toolkit.copy()`
- `HarnessSkillMiddleware.java`: Override `rebindToolkit`, change `toolkit` to `volatile`
- `DynamicSkillMiddleware.java`: Override `rebindToolkit`, change `toolkit` to `volatile`
- `ReActAgentToolkitRebindTest.java`: Unit tests covering rebind behavior

## Testing

- 4 unit tests added: rebind called after deep copy, dynamic tool registered on correct toolkit, default no-op, multiple middlewares all rebind
- `spotless:apply` applied

Fixes #1606
```

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -3076,6 +3076,15 @@ public class ReActAgent extends StructuredOutputCapableAgent implements AutoClos
             // Deep copy toolkit to avoid state interference between agents
             Toolkit agentToolkit = this.toolkit.copy();
 
+            // Rebind middlewares to the new toolkit copy so that any middleware that
+            // dynamically registers tools (e.g., HarnessSkillMiddleware, DynamicSkillMiddleware)
+            // operates on the same toolkit instance the agent will use at runtime.
+            // Without this, the defensive copy above creates a mismatch: the middleware
+            // holds a reference to the pre-copy toolkit, while the agent uses the copy.
+            for (MiddlewareBase mw : middlewares) {
+                mw.rebindToolkit(agentToolkit);
+            }
+
             registerToolsFromHooks(agentToolkit);
 
             if (enableMetaTool) {
--- a/agentscope-core/src/main/java/io/agentscope/core/middleware/MiddlewareBase.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/middleware/MiddlewareBase.java
@@ -17,6 +17,7 @@ package io.agentscope.core.middleware;
 
 import io.agentscope.core.agent.Agent;
 import io.agentscope.core.event.AgentEvent;
+import io.agentscope.core.tool.Toolkit;
 import java.util.function.Function;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
@@ -121,5 +122,22 @@ public interface MiddlewareBase {
      */
     default Mono<String> onSystemPrompt(Agent agent, String currentPrompt) {
         return Mono.just(currentPrompt);
+    }
+
+    /**
+     * Rebind this middleware to a new toolkit instance.
+     *
+     * <p>Called by {@code ReActAgent.Builder.build()} after the defensive toolkit deep copy
+     * so that middlewares that hold a toolkit reference (e.g., for dynamic tool registration)
+     * update their reference to the copy the agent will actually use.
+     *
+     * <p>The default implementation is a no-op. Middlewares that register tools dynamically
+     * during {@link #onSystemPrompt} should override this method to swap their toolkit
+     * reference.
+     *
+     * @param newToolkit the toolkit copy that the agent will actually use
+     */
+    default void rebindToolkit(Toolkit newToolkit) {
+        // no-op by default
     }
 }
--- a/agentscope-core/src/main/java/io/agentscope/core/skill/DynamicSkillMiddleware.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/skill/DynamicSkillMiddleware.java
@@ -54,7 +54,7 @@ public class DynamicSkillMiddleware implements MiddlewareBase {
     private static final Logger log = LoggerFactory.getLogger(DynamicSkillMiddleware.class);
 
     private final List<AgentSkillRepository> repositories;
-    private final Toolkit toolkit;
+    private volatile Toolkit toolkit;
     private final SkillFilter builderFilter;
 
     private volatile SkillBox currentSkillBox;
@@ -76,6 +76,20 @@ public class DynamicSkillMiddleware implements MiddlewareBase {
      */
     public SkillBox getCurrentSkillBox() {
         return currentSkillBox;
+    }
+
+    /**
+     * Rebind this middleware to the agent's actual toolkit after the defensive deep copy
+     * performed by {@code ReActAgent.Builder.build()}.
+     *
+     * <p>Without this rebind, the middleware would register skill tools on the pre-copy
+     * toolkit, making them invisible to the agent at runtime.
+     *
+     * @param newToolkit the toolkit copy that the agent will actually use
+     */
+    @Override
+    public void rebindToolkit(Toolkit newToolkit) {
+        this.toolkit = newToolkit;
     }
 
     @Override
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
@@ -171,7 +171,7 @@ public class ToolCallParam {
             this.toolUseBlock = source.toolUseBlock;
             this.input = source.input.isEmpty() ? null : source.input;
             this.agent = source.agent;
-            this.context = source.context;
+            this.runtimeContext = source.runtimeContext;
             this.emitter = source.emitter;
         }
 
--- a/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentToolkitRebindTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentToolkitRebindTest.java
@@ -1,0 +1,244 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.agent;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotSame;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.agentscope.core.ReActAgent;
+import io.agentscope.core.message.ContentBlock;
+import io.agentscope.core.message.Msg;
+import io.agentscope.core.message.TextBlock;
+import io.agentscope.core.message.ToolResultBlock;
+import io.agentscope.core.middleware.MiddlewareBase;
+import io.agentscope.core.model.ChatModelBase;
+import io.agentscope.core.model.ChatResponse;
+import io.agentscope.core.model.GenerateOptions;
+import io.agentscope.core.model.ToolSchema;
+import io.agentscope.core.tool.AgentTool;
+import io.agentscope.core.tool.ToolCallParam;
+import io.agentscope.core.tool.Toolkit;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+/**
+ * Tests that {@link MiddlewareBase#rebindToolkit(Toolkit)} is called correctly during
+ * {@link ReActAgent.Builder#build()}, ensuring middlewares that dynamically register tools
+ * operate on the agent's actual toolkit after the defensive deep copy.
+ */
+class ReActAgentToolkitRebindTest {
+
+    private static final class FixedTextModel extends ChatModelBase {
+        @Override
+        public String getModelName() {
+            return "fixed";
+        }
+
+        @Override
+        protected Flux<ChatResponse> doStream(
+                List<Msg> messages, List<ToolSchema> tools, GenerateOptions options) {
+            return Flux.just(
+                    ChatResponse.builder()
+                            .content(List.<ContentBlock>of(TextBlock.builder().text("ok").build()))
+                            .build());
+        }
+    }
+
+    /** A simple AgentTool that records which toolkit it was registered on. */
+    private static final class TestDynamicTool implements AgentTool {
+        private final String name;
+        private final AtomicReference<String> registeredOnToolkit = new AtomicReference<>("never");
+
+        TestDynamicTool(String name) {
+            this.name = name;
+        }
+
+        @Override
+        public String getName() {
+            return name;
+        }
+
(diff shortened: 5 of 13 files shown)


agentscope-core/src/main/java/io/agentscope/core/skill/DynamicSkillMiddleware.java
```diff
@@ -78,6 +78,20 @@ public SkillBox getCurrentSkillBox() {
...
+     * performed by {@code ReActAgent.Builder.build()}.
+     *
+     * <p>Without this rebind, the middleware would register skill tools on the pre-copy
+     * toolkit, making them invisible to the agent at runtime.
+     *
+     * @param newToolkit the toolkit copy that the agent will actually use
+     */
+    @Override
+    public void rebindToolkit(Toolkit newToolkit) {
+        this.toolkit = newToolkit;
```
AgentScopeJavaBot, line 92: [minor] Missing null-guard. If `newToolkit` is ever `null` (e.g. called from user code or a future code path), the middleware will silently accept it and then NPE later in `onSystemPrompt` when `toolkit.getTool(...)` or `new SkillBox(toolkit)` is called. Consider:
```java
@Override
public void rebindToolkit(Toolkit newToolkit) {
    Objects.requireNonNull(newToolkit, "newToolkit");
    this.toolkit = newToolkit;
}
```

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
```diff
@@ -137,6 +137,20 @@ public SkillRuntime runtime() {
...
+     * performed by {@code ReActAgent.Builder.build()}.
+     *
+     * <p>Without this rebind, the middleware would register {@code load_skill_through_path}
+     * on the pre-copy toolkit, making the tool invisible to the agent at runtime.
+     *
+     * @param newToolkit the toolkit copy that the agent will actually use
+     */
+    @Override
+    public void rebindToolkit(Toolkit newToolkit) {
+        this.toolkit = newToolkit;
```
AgentScopeJavaBot, line 151: [minor] Same as `DynamicSkillMiddleware` — add a null-guard (`Objects.requireNonNull`) to fail fast instead of accepting a null toolkit that would blow up in `onSystemPrompt` later.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
```diff
@@ -171,7 +171,7 @@ private Builder(ToolCallParam source) {
             this.toolUseBlock = source.toolUseBlock;
             this.input = source.input.isEmpty() ? null : source.input;
             this.agent = source.agent;
-            this.context = source.context;
+            this.runtimeContext = source.runtimeContext;
```
AgentScopeJavaBot, line 174: [nit] This fix (renaming `source.context` to `source.runtimeContext`) is correct — the field was renamed but this copy-constructor line was not updated. However, it is conceptually unrelated to the middleware rebind feature. Consider splitting it into a separate commit for a cleaner history.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -3076,6 +3076,15 @@ public ReActAgent build() {
...
             Toolkit agentToolkit = this.toolkit.copy();
 
+            // Rebind middlewares to the new toolkit copy so that any middleware that
+            // dynamically registers tools (e.g., HarnessSkillMiddleware, DynamicSkillMiddleware)
+            // operates on the same toolkit instance the agent will use at runtime.
+            // Without this, the defensive copy above creates a mismatch: the middleware
+            // holds a reference to the pre-copy toolkit, while the agent uses the copy.
+            for (MiddlewareBase mw : middlewares) {
+                mw.rebindToolkit(agentToolkit);
+            }
```
AgentScopeJavaBot, line 3086: [nit] The rebind loop runs before `configurePlan()`, `configureTodoTools()`, and the inline `DynamicSkillMiddleware` construction add further middlewares to the list. Currently safe because those later middlewares either receive `agentToolkit` directly or do not hold a toolkit reference. A one-line comment after the loop (e.g. `// NOTE: middlewares added below this point receive agentToolkit at construction`) would help future maintainers avoid accidentally adding a toolkit-holding middleware after the loop.

agentscope-core/src/main/java/io/agentscope/core/middleware/MiddlewareBase.java
```diff
@@ -122,4 +123,21 @@ default Flux<AgentEvent> onModelCall(
...
+     *
+     * <p>The default implementation is a no-op. Middlewares that register tools dynamically
+     * during {@link #onSystemPrompt} should override this method to swap their toolkit
+     * reference.
+     *
+     * @param newToolkit the toolkit copy that the agent will actually use
+     */
+    default void rebindToolkit(Toolkit newToolkit) {
+        // no-op by default
+    }
```
AgentScopeJavaBot, line 142: [praise] Clean API extension. Adding the `rebindToolkit` default method to the interface with a no-op default maintains full backward compatibility while giving toolkit-holding middlewares a clear, documented hook. The Javadoc clearly explains the lifecycle contract.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentToolkitRebindTest.java
```diff
@@ -0,0 +1,244 @@
...
+                        .build();
+
+        // Both middlewares should have been rebound
+        assertTrue(middleware1.wasRebindCalled(), "first middleware should be rebound");
+        assertTrue(middleware2.wasRebindCalled(), "second middleware should be rebound");
+
+        // Both should reference the same toolkit as the agent
+        assertSame(agent.getToolkit(), middleware1.getToolkit());
+        assertSame(agent.getToolkit(), middleware2.getToolkit());
+    }
```
AgentScopeJavaBot, line 243: [praise] Solid test coverage. The four tests — rebind called after deep copy, dynamic tool registered on correct toolkit, default no-op, multiple middlewares — effectively validate the core contract without over-testing internals.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a real and important bug where middlewares holding a reference to the original `Toolkit` become disconnected from the agent's actual toolkit after `ReActAgent.Builder.build()` performs a defensive deep copy via `toolkit.copy()`. The solution is well-designed: a `rebindToolkit` default method is added to `MiddlewareBase` (no-op by default), and `build()` iterates all middlewares after the copy to rebind them. Both known toolkit-holding middlewares (`HarnessSkillMiddleware`, `DynamicSkillMiddleware`) override the method and change their field from `final` to `volatile` for correct cross-thread visibility. The accompanying tests are well-structured and cover the key scenarios. The bundled `ToolCallParam` fix (field name mismatch in copy constructor) is correct but unrelated.

hikaruczl:
I also reproduced this issue in an integration project using the documented Harness skill flow with both 2.0.0-RC1 and the currently published 2.0.0-RC2 artifacts. The runtime logs show `Registered tool 'load_skill_through_path'`, but the actual tool execution still returns `Error: Tool not found: load_skill_through_path` because the middleware registers on the pre-copy toolkit while `ReActAgent` executes against the copied toolkit.\n\nThe approach in this PR looks like the right root-cause fix to me: rebinding toolkit-holding middlewares after the defensive `Toolkit.copy()` keeps dynamic tool registration aligned with the agent runtime toolkit, and it covers both `HarnessSkillMiddleware` and `DynamicSkillMiddleware`.\n\nIt would be very helpful to get this merged and released, since the documented `HarnessAgent.builder().skillRepository(...)` flow is otherwise broken for dynamic skill loading. One small suggestion before merge: add or keep an end-to-end HarnessAgent regression test that builds with `skillRepository(...)`, triggers `load_skill_through_path`, and asserts it no longer returns `Tool not found`; that would directly guard the public documented path.

chickenlj:
Thanks for the contribution! This has been fixed by #1828 (ToolkitAware interface for middleware toolkit rebinding).

PR #1611, opened 2026-06-04
fix(harness): 修复 abstractFilesystem 下的 SandboxLifecycleMiddleware 生命周期接管

### LINKED ISSUE #1605: [Bug]:abstractFilesystem() with user-registered SandboxLifecycleMiddleware should participate in wrappedCall lifecycle
When using builder.abstractFilesystem(sandboxFs) combined with builder.middleware(new SandboxLifecycleMiddleware(sandboxManager, sandboxFs)), the sandbox lifecycle is never triggered.

  Root cause: HarnessAgent.wrappedCall() / wrappedStreamEvents() only invokes the private field sandboxLifecycleMw, which is exclusively populated through the
  builder.filesystem(SandboxFilesystemSpec) path. SandboxLifecycleMiddleware does not override any MiddlewareBase methods, so adding it to the middleware chain via builder.middleware() is
   a no-op.

  Use Case

  Custom tools (e.g., an OSS upload tool) need a reference to the SandboxBackedFilesystem instance at construction time to call execute() / downloadFiles(). These tools are registered via
   builder.toolkit() before build().

  With builder.filesystem(SandboxFilesystemSpec), the SandboxBackedFilesystem is created internally during build() — there's no way for user-provided tools to obtain a reference to it.

  This forces users to use abstractFilesystem(), but that path has no lifecycle management, requiring manual acquireForCall / releaseForCall invocation from the application layer —
  reimplementing the Mono.using() pattern that wrappedCall already provides.

  ┌───────────────────────────────────┬───────────────────┬───────────────────────────────────┐
  │               Path                │ Lifecycle managed │ Custom tool can access sandbox FS │
  ├───────────────────────────────────┼───────────────────┼───────────────────────────────────┤
  │ filesystem(SandboxFilesystemSpec) │        ✅         │                ❌                 │
  ├───────────────────────────────────┼───────────────────┼───────────────────────────────────┤
  │ abstractFilesystem(fs)            │        ❌         │                ✅                 │
  └───────────────────────────────────┴───────────────────┴───────────────────────────────────┘

  Proposed Solution

  In HarnessAgent.Builder.build(), if sandboxLifecycleMw is still null after the sandboxFilesystemSpec block, scan the registered middlewares for a user-provided
  SandboxLifecycleMiddleware and promote it to the sandboxLifecycleMw field:

  if (sandboxLifecycleMw == null) {
      for (MiddlewareBase mw : allMiddlewares) {
          if (mw instanceof SandboxLifecycleMiddleware slm) {
              sandboxLifecycleMw = slm;
              break;
          }
      }
  }

  This makes the existing abstractFilesystem() + middleware(SandboxLifecycleMiddleware) combination work as expected — the framework's wrappedCall() handles acquire/release automatically.

  Alternatives Considered

  1. Allow SandboxFilesystemSpec to accept a pre-created SandboxBackedFilesystem — more invasive API change.
  2. Post-build tool registration — e.g., builder.afterSandboxReady(fs -> {...}) callback — new API surface.
  3. Application-layer workaround (current) — call acquireForCall / releaseForCall from the service layer, bypassing framework lifecycle.

  Environment

  - agentscope-harness: 2.0.0-RC1 / 2.0.0-SNAPSHOT (main branch)
  - Affected classes: HarnessAgent, SandboxLifecycleMiddleware

guslegend0510:
## 关联问题
- close #1605

## 问题说明
当使用 `builder.abstractFilesystem(sandboxFs)` 并手动注册 `SandboxLifecycleMiddleware` 时，`HarnessAgent` 的 `wrappedCall()` / `wrappedStreamEvents()` 只会读取内部字段 `sandboxLifecycleMw`。

这个字段只会在 `filesystem(SandboxFilesystemSpec)` 路径下赋值，因此用户通过 `middleware(...)` 注册的 `SandboxLifecycleMiddleware` 虽然进入了 middleware 链，但不会参与 sandbox 的 acquire / release 生命周期。

## 修复内容
- 在 `HarnessAgent.Builder.build()` 完成 `delegate` 构建后，如果 `sandboxLifecycleMw` 仍为空，则从 `delegate.getMiddlewares()` 中提取用户注册的 `SandboxLifecycleMiddleware`
- 保持 `filesystem(SandboxFilesystemSpec)` 的现有行为不变，只为 `abstractFilesystem() + middleware(...)` 提供生命周期桥接
- 新增回归测试，验证 `abstractFilesystem() + middleware(SandboxLifecycleMiddleware)` 会触发 `acquireForCall()` / `releaseForCall()`

## 附带修复
当前 `upstream/main` 上 `ToolCallParam` 的 copy builder 仍在拷贝已移除的 `context` 字段，导致最新主线本地编译失败。

本 PR 一并修复该阻塞，并将对应测试更新到 `RuntimeContext` 语义，保证该分支可以在最新主线上独立通过验证。

## 验证
```bash
mvn -pl agentscope-core -Dtest=ToolCallParamTest -Dspotless.check.skip=true test
mvn -pl agentscope-harness -am -Dtest=HarnessAgentTest#abstractFilesystem_userRegisteredSandboxLifecycleMiddleware_participatesInWrappedCall -Dsurefire.failIfNoSpecifiedTests=false -Dspotless.check.skip=true test
```

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
@@ -171,7 +171,7 @@ public class ToolCallParam {
             this.toolUseBlock = source.toolUseBlock;
             this.input = source.input.isEmpty() ? null : source.input;
             this.agent = source.agent;
-            this.context = source.context;
+            this.runtimeContext = source.runtimeContext;
             this.emitter = source.emitter;
         }
 
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
@@ -1900,6 +1900,16 @@ public class HarnessAgent implements Agent, AutoCloseable {
             // ---- Build inner ReActAgent ----
             inner.toolkit(agentToolkit);
             ReActAgent delegate = inner.build();
+            if (sandboxLifecycleMw == null) {
+                // Reuse an explicitly registered middleware so abstractFilesystem() still
+                // participates in wrappedCall / wrappedStreamEvents lifecycle management.
+                for (MiddlewareBase mw : delegate.getMiddlewares()) {
+                    if (mw instanceof SandboxLifecycleMiddleware slm) {
+                        sandboxLifecycleMw = slm;
+                        break;
+                    }
+                }
+            }
             selfRef.set(delegate);
 
             return new HarnessAgent(
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentTest.java
@@ -23,6 +23,7 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyList;
 import static org.mockito.Mockito.atLeast;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -37,9 +38,15 @@ import io.agentscope.core.session.Session;
 import io.agentscope.core.tool.AgentTool;
 import io.agentscope.core.tool.Toolkit;
 import io.agentscope.harness.agent.filesystem.local.LocalFilesystem;
+import io.agentscope.harness.agent.filesystem.sandbox.SandboxBackedFilesystem;
 import io.agentscope.harness.agent.filesystem.spec.RemoteFilesystemSpec;
 import io.agentscope.harness.agent.memory.compaction.CompactionConfig;
+import io.agentscope.harness.agent.middleware.SandboxLifecycleMiddleware;
 import io.agentscope.harness.agent.middleware.SubagentEntry;
+import io.agentscope.harness.agent.sandbox.Sandbox;
+import io.agentscope.harness.agent.sandbox.SandboxAcquireResult;
+import io.agentscope.harness.agent.sandbox.SandboxContext;
+import io.agentscope.harness.agent.sandbox.SandboxManager;
 import io.agentscope.harness.agent.store.InMemoryStore;
 import io.agentscope.harness.agent.subagent.AgentSpecLoader;
 import io.agentscope.harness.agent.subagent.SubagentDeclaration;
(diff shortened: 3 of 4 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1900,6 +1900,16 @@ public HarnessAgent build() {
             // ---- Build inner ReActAgent ----
             inner.toolkit(agentToolkit);
             ReActAgent delegate = inner.build();
+            if (sandboxLifecycleMw == null) {
+                // Reuse an explicitly registered middleware so abstractFilesystem() still
+                // participates in wrappedCall / wrappedStreamEvents lifecycle management.
```
AgentScopeJavaBot, line 1905: [minor] Only the first `SandboxLifecycleMiddleware` found in the middleware chain is used. If a user registers multiple instances, subsequent ones are ignored. This is acceptable for the current use case but could be documented in the `abstractFilesystem()` javadoc.

agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentTest.java
```diff
@@ -346,6 +353,43 @@ void remoteFilesystemSpec_sharesMemoryMdInNonsandboxMode() throws Exception {
         }
     }
 
+    @Test
+    void abstractFilesystem_userRegisteredSandboxLifecycleMiddleware_participatesInWrappedCall()
```
AgentScopeJavaBot, line 357: [minor] Consider adding a complementary test for `streamEvents()` path to verify that `wrappedStreamEvents` also triggers the lifecycle. The current test only covers `call()`, though the fix is shared via the same `sandboxLifecycleMw` field assignment.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
```diff
@@ -171,7 +171,7 @@ private Builder(ToolCallParam source) {
             this.toolUseBlock = source.toolUseBlock;
             this.input = source.input.isEmpty() ? null : source.input;
             this.agent = source.agent;
-            this.context = source.context;
+            this.runtimeContext = source.runtimeContext;
```
AgentScopeJavaBot, line 174: [praise] Fix is correct: the old `this.context = source.context` referenced a stale field name left over from the `context -> runtimeContext` rename. Without this fix, RuntimeContext would be silently null-ed out in any `ToolCallParam.builder(source).build()` copy.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR addresses two distinct bugs:

1. **ToolCallParam copy-constructor field mismatch (agentscope-core):** The `Builder(ToolCallParam source)` copy constructor referenced the old field name `context` instead of the current field name `runtimeContext`, causing the RuntimeContext to be silently dropped when copying a `ToolCallParam`. The fix corrects `this.context = source.context` to `this.runtimeContext = source.runtimeContext`. The accompanying test updates switch from `assertEquals` on the deprecated `getContext()` to `assertSame` on `getRuntimeContext()` plus a secondary `getContext()` assertion to verify backward compatibility.

2. **SandboxLifecycleMiddleware not discovered for abstractFilesystem escape hatch (agentscope-harness):** When users call `.abstractFilesystem(...)` and manually register a `SandboxLifecycleMiddleware` via `.middleware(lifecycle)`, the builder's local `sandboxLifecycleMw` variable remained `null` because it is only assigned when `sandboxFilesystemSpec != null`. As a result, `wrappedCall`, `wrappedStream`, and `wrappedStreamEvents` all skipped lifecycle acquire/release, meaning the sandbox was never warmed up or torn down at the HarnessAgent level. The fix scans `delegate.getMiddlewares()` post-build to discover any user-registered `SandboxLifecycleMiddleware` and passes it through to the `HarnessAgent` constructor. The new test verifies `acquireForCall` and `releaseForCall` are invoked.

Both fixes are minimal, well-scoped, and carry appropriate test coverage.

Issue #1598, opened 2026-06-04
[Bug]: model 里面使用 OllamaChatModel 构建，运行报错


hrbwbj:
public void harnessAgentStream() {

        HarnessAgent agent = HarnessAgent.builder()
                .name("麦兜")
                .sysPrompt("你是一个憨憨的聊天精灵")
                // .model("dashscope:qwen3.7-max")
                // .model(
                //         OllamaChatModel.builder()
                //                 .modelName("qwen3:8b")
                //                 .baseUrl("http://localhost:11434")
                //                 .build()
                // )
                .model(
                        OpenAIChatModel.builder()
                                .modelName("qwen3:8b")
                                // 关键: 将 baseUrl 指向 Ollama 的 OpenAI 兼容端点
                                .baseUrl("http://localhost:11434/v1")
                                // OpenAI 模型需要 apiKey，Ollama 本地服务可以随意填写
                                .apiKey("ollama")
                                .build()
                )
                .workspace(Paths.get(".agentscope/workspace"))
                .compaction(CompactionConfig.builder()
                        .triggerMessages(30)
                        .keepMessages(10)
                        .build())
                .build();

        RuntimeContext ctx = RuntimeContext.builder()
                .sessionId("demo-session2")
                .userId("alice")
                .build();

        agent.streamEvents(new UserMessage("你会做东北的酸甜汤吗，里面用了很多种粮食"),ctx)
                .doOnNext(event -> {
                    if (event.getType() == AgentEventType.TEXT_BLOCK_DELTA) {
                        // 模型返回的流式文本片段 —— 追加到界面或标准输出
                        System.out.print(((TextBlockDeltaEvent) event).getDelta());
                    } else if (event.getType() == AgentEventType.TOOL_CALL_START) {
                        // 智能体即将调用工具 —— 展示调用信息
                        System.out.println("\n[tool] " + ((ToolCallStartEvent) event).getToolCallName());
                    }
                    // 其他事件：思考块、工具结果、回复结束等
                })
                .blockLast();
    }

这个 2.0.0-RC1 版本中，model 使用本地的OllamaChatModel流式调用会报错，使用其它的不会报错！
报错内容：
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "io.agentscope.core.model.GenerateOptions.getToolChoice()" because "options" is null
	at io.agentscope.core.model.OllamaChatModel.doStream(OllamaChatModel.java:144)
[...]

PR #1613, opened 2026-06-04
fix: 修复 OllamaChatModel 在 stream 入参为空时的 NPE

guslegend0510:
## 描述
本 PR 修复了 issue #1598 中 OllamaChatModel 流式调用在 GenerateOptions 为空时触发空指针异常的问题。

## 变更内容

在 OllamaChatModel#doStream() 中对 options 做空值保护，避免直接访问 options.getToolChoice()。
保持 options == null 时的默认行为，不影响原有流式调用逻辑。
新增回归测试，覆盖 model.stream(..., null, null) 场景，验证不会再抛出 NPE。
顺手修复了 ToolCallParam 中 builder 拷贝逻辑的 runtimeContext 字段引用错误，保证模块可正常编译。
## 验证

mvn -pl agentscope-core -am -Dtest=OllamaChatModelTest test


agentscope-core/src/main/java/io/agentscope/core/model/OllamaChatModel.java
```diff
@@ -138,12 +138,9 @@ public ChatResponse chat(List<Msg> messages, GenerateOptions options) {
     @Override
     protected Flux<ChatResponse> doStream(
             List<Msg> messages, List<ToolSchema> tools, GenerateOptions options) {
+        ToolChoice toolChoice = options != null ? options.getToolChoice() : null;
```
AgentScopeJavaBot, line 141: [praise] Clean null guard that is consistent with how `GeminiChatModel.doStream()` handles the same scenario (`options != null && options.getToolChoice() != null`). The downstream `OllamaOptions.fromGenerateOptions(null)` is already null-safe, so only `getToolChoice()` needed guarding. Good fix.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
```diff
@@ -171,7 +171,7 @@ private Builder(ToolCallParam source) {
             this.toolUseBlock = source.toolUseBlock;
             this.input = source.input.isEmpty() ? null : source.input;
             this.agent = source.agent;
-            this.context = source.context;
+            this.runtimeContext = source.runtimeContext;
```
AgentScopeJavaBot, line 174: [praise] Correct fix: the Builder has no field named `context` (only `runtimeContext`), so the old code `this.context = source.context` would fail to compile. This was likely left behind during a field-rename refactoring.

agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
```diff
@@ -48,6 +48,7 @@ public class RuntimeContext {
     private final ConcurrentMap<Class<?>, ConcurrentMap<String, Object>> typedAttributes;
 
     private final ToolExecutionContext toolExecutionContext;
+    private volatile ToolExecutionContext toolExecutionContextView;
```
AgentScopeJavaBot, line 58: [minor] The racy single-check idiom (volatile without synchronization) is functionally safe here because the computation is idempotent and `DefaultMutableContextStore` is a live delegate. However, under concurrent first-access, multiple threads may each build a `ToolExecutionContext` and only one write survives. Consider adding a brief comment (e.g., `// benign race: idempotent, live-view stores`) so future readers understand this is intentional rather than a missed synchronization.

agentscope-core/src/test/java/io/agentscope/core/model/OllamaChatModelTest.java
```diff
@@ -269,6 +269,41 @@ void testStreamChatRequest() {
...
+        assertNotNull(responses);
+        assertFalse(responses.isEmpty());
+
+        ArgumentCaptor<HttpRequest> captor = ArgumentCaptor.forClass(HttpRequest.class);
+        verify(httpTransport).stream(captor.capture());
+
+        HttpRequest request = captor.getValue();
+        assertEquals("POST", request.getMethod());
+        assertTrue(request.getUrl().endsWith("/api/chat"));
+    }
```
AgentScopeJavaBot, line 305: [praise] Good regression test. Passing `null` for both `tools` and `options` directly validates the NPE scenario from issue #1598. The assertions on the HTTP request URL add extra confidence that the full code path executed correctly.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a real NPE in `OllamaChatModel.doStream()` when `GenerateOptions` is null, adds a null guard for `options.getToolChoice()`, and includes a well-targeted regression test. The `ToolCallParam` builder copy-constructor fix (renaming the stale `context` reference to `runtimeContext`) is a valid compilation-error fix that aligns with the field rename that already happened in the class. The `RuntimeContext.asToolExecutionContext()` caching is a sound optimization: the volatile racy single-check idiom is appropriate here because the computation is idempotent and the `DefaultMutableContextStore` is a live view that delegates to the mutable `RuntimeContext`, so cached results stay correct. Test coverage is good for all three changes. The PR is well-scoped and the fixes are consistent with patterns used in other model implementations (e.g., `GeminiChatModel` line 253 already guards `options != null`).

agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
LearningGp: PTAL at this comment.

guslegend0510: 已按建议补充注释，说明这里的 volatile 是一个 benign race：多个线程可能会重建同一个 live-view wrapper，但结果是等价的、幂等的，不是同步缺陷。谢谢提醒。

guslegend0510:
image
@LearningGp ，麻烦再审核一下，谢谢

chickenlj:
Thanks for the contribution! The Ollama NPE fix is already on main. The bundled ToolCallParam/RuntimeContext changes have been addressed by other merged PRs.

Issue #1608, opened 2026-06-04
[Bug]: Compilation Error in ToolCallParam.Builder


Weilong-Qin:
**Describe the bug**
There is a compilation error in the private copy constructor of the `Builder` class within `ToolCallParam`. The code attempts to assign a field named `context,` but neither the `ToolCallParam` class nor the `Builder` class contains a field with that name. 
https://github.com/agentscope-ai/agentscope-java/blob/6c39c7b4fa9f8ac2f53a54704ea653546fff6e60/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java#L46-L52

https://github.com/agentscope-ai/agentscope-java/blob/6c39c7b4fa9f8ac2f53a54704ea653546fff6e60/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java#L161-L176

Weilong-Qin:
Fixed #1625

PR #1622, opened 2026-06-05
fix(core): copy runtime context in ToolCallParam builder

Weilong-Qin:
## AgentScope-Java Version

2.0.0-SNAPSHOT

## Description

Fixes (https://github.com/agentscope-ai/agentscope-java/issues/1608)

`ToolCallParam` was migrated to store `RuntimeContext`, while the newly added copy builder still referenced the old `context` field. This caused compilation failures.

This PR updates the copy builder to copy `runtimeContext` from the source `ToolCallParam`. It also updates the related tests to assert that the copied parameter preserves the same `RuntimeContext` reference, instead of asserting identity on the deprecated `getContext()` projection.

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
@@ -144,7 +144,7 @@ public class ToolCallParam {
      *
      * <p>Note: The input map structure is copied so that entries can be modified
      * independently, but nested values (typed as Object) remain shared.
-     * Immutable fields (toolUseBlock, agent, context, emitter) are shared by reference.
+     * Immutable fields (toolUseBlock, agent, runtimeContext, emitter) are shared by reference.
      *
      * @param source The existing ToolCallParam to copy values from
      * @return A new builder pre-populated with the source's values
@@ -171,7 +171,7 @@ public class ToolCallParam {
             this.toolUseBlock = source.toolUseBlock;
             this.input = source.input.isEmpty() ? null : source.input;
             this.agent = source.agent;
-            this.context = source.context;
+            this.runtimeContext = source.runtimeContext;
             this.emitter = source.emitter;
         }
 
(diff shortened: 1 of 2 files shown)


agentscope-core/src/main/java/io/agentscope/core/tool/ToolCallParam.java
```diff
@@ -171,7 +171,7 @@ private Builder(ToolCallParam source) {
             this.toolUseBlock = source.toolUseBlock;
             this.input = source.input.isEmpty() ? null : source.input;
             this.agent = source.agent;
-            this.context = source.context;
+            this.runtimeContext = source.runtimeContext;
```
AgentScopeJavaBot, line 174: [praise] Correct fix -- the copy builder now consistently copies `runtimeContext` from the source, matching the field declared in the `Builder` (line 165) and the `ToolCallParam` constructor (line 58). The Javadoc update on line 147 is also accurate.

agentscope-core/src/test/java/io/agentscope/core/tool/ToolCallParamTest.java
```diff
@@ -160,7 +160,7 @@ void testCopyAllFields() {
             assertEquals(original.getToolUseBlock(), copy.getToolUseBlock());
             assertEquals(original.getInput(), copy.getInput());
             assertEquals(original.getAgent(), copy.getAgent());
-            assertEquals(original.getContext(), copy.getContext());
+            assertSame(original.getRuntimeContext(), copy.getRuntimeContext());
```
AgentScopeJavaBot, line 163: [praise] Good upgrade from `assertEquals` to `assertSame` -- since the copy builder shares immutable fields by reference, identity checking is the stronger and more precise assertion here, consistent with the `emitter` assertion on the next line.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a compilation-breaking bug in the `ToolCallParam` copy builder introduced when the class was migrated from the legacy `context` field (`ToolExecutionContext`) to `runtimeContext` (`RuntimeContext`). The copy builder constructor still referenced the old `this.context = source.context`, which would fail to compile since the Builder class no longer has a `context` field.

The fix is minimal and correct: it updates the single line in the copy builder constructor to `this.runtimeContext = source.runtimeContext`, updates the corresponding Javadoc to list `runtimeContext` instead of `context`, and adjusts two test assertions from `getContext()` to `getRuntimeContext()` with the stronger `assertSame` check. The change scope is appropriately narrow for a bug fix, touching only the affected lines.

No new risks are introduced. The existing test suite thoroughly covers copy-builder semantics (field copying, shallow copy of input, null handling, immutable reference sharing, and NPE on null source).

Weilong-Qin:
FIW #1625

PR #1646, opened 2026-06-06
Fix Gemini media URL extension parsing

### LINKED ISSUE #1645: [Bug]:GeminiMediaConverter rejects media URLs with query strings
**Describe the bug**

`GeminiMediaConverter` rejects valid remote media URLs when the URL contains a query string or fragment.

For example, an image URL such as:

```text
http://localhost:64260/cat.png?token=abc123
```

is a valid PNG URL, but AgentScope-Java parses the extension as:

```text
png?token=abc123
```

and throws `Unsupported file extension`.

This affects common user scenarios such as CDN URLs, object storage signed URLs, temporary download URLs, and media URLs with cache/auth parameters.

**Expected behavior**

The converter should ignore the query string and fragment when inferring the file extension.

For:

```text
http://127.0.0.1:64260/cat.png?token=abc123
```

it should infer:

```text
png
```

and produce MIME type:

```text
image/png
```

The same should apply to URLs such as:

```text
https://cdn.example.com/audio.mp3?download=1
https://cdn.example.com/video.mp4#preview
```

**Error messages**

```text
FAILED: java.lang.IllegalArgumentException
MESSAGE: Unsupported file extension: png?token=abc123 for media type: image, expected one of: [png, jpeg, jpg, webp, heic, heif]
```

- AgentScope-Java Version：2.0
- Java Version: 17

**Additional context**

I would like to submit a PR to help fix this issue.

SPUERSAIYAN:
## AgentScope-Java Version

Local main branch 

## Description

Fixes #1645
This PR fixes Gemini media URL extension parsing for remote media URLs that contain query strings or fragments.

Previously, `GeminiMediaConverter` extracted the file extension by taking the substring after the last `.` in the full URL string. For URLs like:

```text
https://cdn.example.com/image.png?token=abc123
```

the parsed extension became:

```text
png?token=abc123
```

which caused the converter to reject an otherwise valid PNG URL.

This PR updates `GeminiMediaConverter` to reuse `MediaUtils.getExtension()`, which already handles URL paths separately from query strings and fragments. A regression test was added to verify that an image URL like `/test.png?token=abc123` is accepted and converted with MIME type `image/png`.

Tested with:

```bash
mvn -pl agentscope-core "-Dtest=GeminiMediaConverterTest" test -q
mvn -pl agentscope-core spotless:check -q
git diff --check
```

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiMediaConverter.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiMediaConverter.java
@@ -17,6 +17,7 @@ package io.agentscope.core.formatter.gemini;
 
 import com.google.genai.types.Blob;
 import com.google.genai.types.Part;
+import io.agentscope.core.formatter.MediaUtils;
 import io.agentscope.core.message.AudioBlock;
 import io.agentscope.core.message.Base64Source;
 import io.agentscope.core.message.ImageBlock;
@@ -190,10 +191,10 @@ public class GeminiMediaConverter {
      * @return File extension in lowercase (without dot)
      */
     private String extractExtension(String url) {
-        int lastDotIndex = url.lastIndexOf('.');
-        if (lastDotIndex == -1 || lastDotIndex == url.length() - 1) {
+        String extension = MediaUtils.getExtension(url);
+        if (extension.isBlank()) {
             throw new IllegalArgumentException("Cannot extract file extension from: " + url);
         }
-        return url.substring(lastDotIndex + 1).toLowerCase();
+        return extension.toLowerCase();
     }
 }
--- a/agentscope-core/src/test/java/io/agentscope/core/formatter/gemini/GeminiMediaConverterTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/formatter/gemini/GeminiMediaConverterTest.java
@@ -32,6 +32,8 @@ import io.agentscope.core.message.VideoBlock;
 import java.io.OutputStream;
 import java.net.InetSocketAddress;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Base64;
 import org.junit.jupiter.api.Test;
 
@@ -176,6 +178,21 @@ class GeminiMediaConverterTest extends GeminiFormatterTestBase {
         ImageBlock block = ImageBlock.builder().source(source).build();
 
         assertThrows(RuntimeException.class, () -> converter.convertToInlineDataPart(block));
+    }
+
+    @Test
+    void testMissingExtension() throws Exception {
+        Path fileWithoutExtension = Files.createTempFile("gemini_media_no_extension", "");
+        Files.writeString(fileWithoutExtension, "fake image content");
+        try {
+            URLSource source = URLSource.builder().url(fileWithoutExtension.toString()).build();
+            ImageBlock block = ImageBlock.builder().source(source).build();
+
+            assertThrows(
+                    IllegalArgumentException.class, () -> converter.convertToInlineDataPart(block));
+        } finally {
+            Files.deleteIfExists(fileWithoutExtension);
+        }
     }
 
     @Test
--- a/agentscope-core/src/test/java/io/agentscope/core/formatter/gemini/GeminiMediaConverterTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/formatter/gemini/GeminiMediaConverterTest.java
@@ -23,11 +23,15 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.google.genai.types.Blob;
 import com.google.genai.types.Part;
+import com.sun.net.httpserver.HttpServer;
 import io.agentscope.core.message.AudioBlock;
 import io.agentscope.core.message.Base64Source;
 import io.agentscope.core.message.ImageBlock;
 import io.agentscope.core.message.URLSource;
 import io.agentscope.core.message.VideoBlock;
+import java.io.OutputStream;
+import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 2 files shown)


AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes a bug in `GeminiMediaConverter.extractExtension()` where the old implementation used a naive `url.lastIndexOf('.')` to extract file extensions from URLs. For remote URLs containing query strings or fragments with dots (e.g., `image.png?sig=abc.def`), this would return the wrong substring (e.g., `def`) instead of the actual extension (`png`). The fix correctly delegates to the existing `MediaUtils.getExtension()` which parses the URI path component — stripping query strings and fragments — before extracting the extension. This aligns Gemini with the Anthropic, DashScope, and Ollama converters that already use `MediaUtils`. A well-structured regression test with an embedded HTTP server validates the fix end-to-end.

oss-maintainer:

```bash
git fetch origin
git rebase origin/main
git push --force-with-lease
```

---

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiMediaConverter.java
```diff
@@ -190,10 +191,10 @@ private String getMimeType(String url, String mediaType) {
      * @return File extension in lowercase (without dot)
      */
     private String extractExtension(String url) {
-        int lastDotIndex = url.lastIndexOf('.');
-        if (lastDotIndex == -1 || lastDotIndex == url.length() - 1) {
+        String extension = MediaUtils.getExtension(url);
```
oss-maintainer, line 194: **[Warning]** Delegating to `MediaUtils.getExtension` is the right fix (it parses the URI path, so `test.png?token=abc` now yields `png` instead of `png?token=abc`). But the identical old implementation is still live in the extension module: `agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/.../formatter/GeminiMediaConverter.java` still does `url.lastIndexOf('.')` / `substring`. Since that module is published in the BOM and used by the Gemini Spring Boot starter, Gemini users on that path still hit the original bug. Could you apply the same delegation there (or delete the duplicate in favour of `MediaUtils`) so the two converters cannot drift again?

agentscope-core/src/main/java/io/agentscope/core/formatter/gemini/GeminiMediaConverter.java
```diff
@@ -190,10 +191,10 @@ private String getMimeType(String url, String mediaType) {
...
+        if (extension.isBlank()) {
             throw new IllegalArgumentException("Cannot extract file extension from: " + url);
         }
-        return url.substring(lastDotIndex + 1).toLowerCase();
+        return extension.toLowerCase();
```
oss-maintainer, line 198: **[Info]** Minor, and inherited from the previous code rather than new here: `toLowerCase()` without a `Locale` is locale-sensitive. An extension such as `TIF`/`PNG` is fine, but under a Turkish default locale `I` maps to dotless `ı`, which would silently break the `SUPPORTED_EXTENSIONS` lookup. `toLowerCase(Locale.ROOT)` is the safer form here and in `MediaUtils.getExtension(...)` callers.

agentscope-core/src/test/java/io/agentscope/core/formatter/gemini/GeminiMediaConverterTest.java
```diff
@@ -78,6 +84,38 @@ void testConvertImageBlockWithURLSource() {
         assertEquals("image/png", blob.mimeType().get());
     }
 
+    @Test
+    void testConvertImageBlockWithUrlQueryString() throws Exception {
+        HttpServer server = HttpServer.create(new InetSocketAddress("127.0.0.1", 0), 0);
```
oss-maintainer, line 89: **[Warning]** This test asserts the right thing end to end, but it binds a real `HttpServer` on the loopback interface to do it. Unit tests in this module otherwise stay in-process (the neighbouring `testConvertImageBlockWithURLSource` uses a local file), and CI runners with restricted networking or a proxy env var (`HTTP_PROXY`/`http.proxyHost` applied via JVM defaults) will make this flaky rather than deterministically green. Since the fix under test is pure string handling, consider asserting the resolved extension/mime type directly, and keep the socket-backed case only if you also need to cover the download path.

oss-maintainer (commented):
### Summary
Fixes  to go through , so a URL with a query string () no longer yields the bogus extension . Correct fix, well-scoped, and the regression tests cover both the query-string and the no-extension cases.

### Findings
- **[Warning]** GeminiMediaConverter.java:194 (core) — the same buggy  implementation is still live in ; that path keeps the bug for starter/BOM users.
- **[Warning]** GeminiMediaConverterTest.java:89 — new test binds a loopback ; the behaviour under test is pure string parsing, so an in-process assertion would be less flaky in CI.
- **[Info]** GeminiMediaConverter.java:198 —  without  (pre-existing, worth fixing while touching the line).

### Suggestions

For the extension module, either delegate to  the same way or, better, drop the private copy so there is one implementation of extension parsing in the codebase.

### Notes
CLA is signed. This PR still reports  against  (a rebase reminder was already posted on 2026-09-18), so it needs a rebase before it can be merged.

---

oss-maintainer:
### Review summary (re-posted — the review body above lost its inline code spans to a shell quoting bug)

The three inline comments on this PR were posted correctly; only the summary text was mangled. Intended text:

**Summary.** Fixes `extractExtension` to go through `MediaUtils.getExtension`, so a URL with a query string (`.../test.png?token=abc`) no longer yields the bogus extension `png?token=abc`. Correct fix, well-scoped, and the regression tests cover both the query-string and the missing-extension cases.

**Findings**
- **[Warning]** `GeminiMediaConverter.java:194` (core) — the identical old implementation (`url.lastIndexOf('.')` + `substring`) is still live in `agentscope-extensions-model-gemini/.../formatter/GeminiMediaConverter.java`. That module ships in the BOM and backs the Gemini Spring Boot starter, so Gemini users on that path keep hitting the original bug. Could you delegate there too, or drop the private copy in favour of `MediaUtils`, so the two converters cannot drift apart again?
- **[Warning]** `GeminiMediaConverterTest.java:89` — the new case binds a real loopback `HttpServer` to assert behaviour that is pure string parsing; the neighbouring `testConvertImageBlockWithURLSource` stays in-process with a local file. Prefer an in-process assertion and keep the socket-backed test only if the download path itself needs coverage, otherwise this is a CI flakiness source (restricted networking / proxy JVM properties).
- **[Info]** `GeminiMediaConverter.java:198` — `toLowerCase()` is locale-sensitive (pre-existing, but this line touches it): under a Turkish default locale `I` maps to dotless `ı` and the `SUPPORTED_EXTENSIONS` lookup silently fails. `toLowerCase(Locale.ROOT)` is the safer form, ideally in `MediaUtils.getExtension` callers in general.

**Notes.** CLA is signed. The PR still reports `CONFLICTING` against `main` (a rebase reminder was already posted on 2026-09-18), so it needs a rebase before it can be merged.

---

PR #1670, opened 2026-06-08
feat[core] refactoring and enhancements for 2.2.0-RC2

chickenlj:
This pull request introduces several important refactorings and deprecations in the agent runtime and user input APIs. The main focus is on improving concurrency safety for agent state access, simplifying the runtime context, and marking legacy structured output and user input APIs as deprecated in favor of new mechanisms.

Key changes include:

- Making ReActAgent completely stateless, along with some breaking api changes.
- Marking legacy structured output and user input classes/interfaces as deprecated, signaling migration to newer middleware-based approaches.


AgentScopeJavaBot (commented):
🤖 AI Review

This is a substantial refactoring PR (+22,598 / -31,767 across 100 files) that makes `ReActAgent` stateless by moving per-session state into a `ConcurrentHashMap`-backed cache keyed by `(userId, sessionId)`, replaces the `Session` abstraction with a new `AgentStateStore` interface (with `InMemory` and `JsonFile` implementations), and introduces per-session `InterruptControl` for precise concurrency-safe interruption. The architecture direction is sound — Reactor Context propagation for call-scoped data, slot serialization via `callSerializationKey`, and proper DCL for lazy `InterruptControl` init are well-executed. However, the `JsonFileAgentStateStore` has critical durability gaps (non-atomic writes, no concurrent-write protection) and the `saveStateToSession` method performs blocking I/O on the reactive chain without scheduler offload, which could stall event-loop threads under the file-based store.

agentscope-core/src/main/java/io/agentscope/core/state/JsonFileAgentStateStore.java
```diff
@@ -0,0 +1,387 @@
...
+        }
+    }
+
+    @Override
+    public void save(String userId, String sessionId, String key, State value) {
+        Path file = getStatePath(userId, sessionId, key);
+        ensureDirectoryExists(file.getParent());
+        try {
+            String json = JsonUtils.getJsonCodec().toPrettyJson(value);
+            Files.writeString(file, json, StandardCharsets.UTF_8);
```
AgentScopeJavaBot: [critical] **Non-atomic file write risks state corruption.** `Files.writeString()` is not atomic — if the JVM crashes or the thread is interrupted mid-write, the state file is left truncated or partially written. On the next load, JSON deserialization will fail and the agent state for that session is permanently lost.

Fix: write to a sibling temp file, then `Files.move(tmp, file, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING)`:
```java
Path tmp = file.resolveSibling(file.getFileName() + ".tmp");
Files.writeString(tmp, json, StandardCharsets.UTF_8);
Files.move(tmp, file, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
```

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -310,30 +378,78 @@ private static AgentState loadOrCreateAgentState(
...
+    private Mono<Void> saveStateToSession(CallExecution scope) {
+        if (stateStore == null) {
             return Mono.empty();
         }
-        syncToolkitToState();
-        return Mono.fromRunnable(() -> session.save(sessionKey, "agent_state", state));
+        syncToolkitToState(scope.state);
+        SlotRef ref = SlotRef.parse(scope.slotKey);
+        AgentState toSave = scope.state;
+        return Mono.fromRunnable(
```
AgentScopeJavaBot: [major] **Blocking I/O on the reactive chain without scheduler offload.** `Mono.fromRunnable(() -> stateStore.save(...))` wraps a synchronous file write (for `JsonFileAgentStateStore`) without `subscribeOn(Schedulers.boundedElastic())`. When the reactive chain runs on a Netty event loop or a shared `parallel` scheduler, this blocks the thread for the duration of the disk write.

Consider:
```java
return Mono.fromRunnable(
        () -> stateStore.save(ref.userId, ref.sessionId, "agent_state", toSave))
    .subscribeOn(Schedulers.boundedElastic());
```
Or make `AgentStateStore.save()` return `Mono<Void>` so implementations can choose their own threading model.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -259,47 +298,76 @@ private ReActAgent(Builder builder, Toolkit agentToolkit) {
...
+        }
+    }
+
+    /**
+     * Internal slot identifier — {@code (userId or "__anon__") + "/" + sessionId}.
+     * Not part of the public API.
+     */
+    private static String slotKey(String userId, String sessionId) {
+        Objects.requireNonNull(sessionId, "sessionId must not be null");
+        return (userId == null || userId.isBlank() ? "__anon__" : userId) + "/" + sessionId;
```
AgentScopeJavaBot, line 336: [major] **Fragile `/` separator in `slotKey` / `SlotRef.parse` breaks round-trip for userIds containing `/`.** `slotKey()` concatenates `userId + "/" + sessionId`, but `SlotRef.parse()` splits on the **first** `/` via `indexOf('/')`. If a `userId` contains `/` (e.g., `org/team/alice`), the parse yields `("org", "team/alice/session-1")` instead of `("org/team/alice", "session-1")`.

Fix options:
1. Validate/reject `userId` containing `/` in `slotKey()`.
2. Use `lastIndexOf('/')` in `parse()` (safe since `sessionId` is the last segment).
3. Use a separator that can't appear in either field (e.g., URL-encode first).

agentscope-core/src/main/java/io/agentscope/core/state/JsonFileAgentStateStore.java
```diff
@@ -0,0 +1,387 @@
...
+    public JsonFileAgentStateStore(Path rootDirectory) {
+        this.rootDirectory = rootDirectory;
+        try {
+            Files.createDirectories(rootDirectory);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to create root directory: " + rootDirectory, e);
+        }
+    }
+
+    @Override
```
AgentScopeJavaBot, line 98: [major] **No concurrent-write protection for the same `(userId, sessionId, key)` file.** Unlike `InMemoryAgentStateStore` (backed by `ConcurrentHashMap`), this file-based store has no file locking or per-key mutex. Two concurrent `call()` completions for the same session can invoke `save()` simultaneously, interleaving `Files.writeString()` calls and producing a corrupted file.

Consider adding a per-slot `ReentrantLock` or using `FileChannel` with `FileLock`, or serialize writes through a single-threaded executor per root directory.

agentscope-core/src/main/java/io/agentscope/core/tool/subagent/SubAgentTool.java
```diff
@@ -228,19 +224,18 @@ private void loadAgentState(String sessionId, Agent agent) {
...
         }
-        Session subSession = config.getSession();
+        AgentStateStore subSession = config.getStateStore();
         if (subSession == null) {
             return;
         }
-        SessionKey key = SimpleSessionKey.of(sessionId);
         try {
-            subSession.save(key, "agent_state", ra.getAgentState());
+            subSession.save(null, sessionId, "agent_state", ra.getAgentState());
```
AgentScopeJavaBot, line 238: [minor] **Uses deprecated no-arg `getAgentState()` which resolves to `defaultSessionId`.** Under the new multi-session model, `ra.getAgentState()` delegates to `getAgentState(null, defaultSessionId)`. This happens to work when the sub-agent's `defaultSessionId` matches the active session, but it relies on a deprecated API and masks the intent. Use the explicit overload:
```java
subSession.save(null, sessionId, "agent_state", ra.getAgentState(null, sessionId));
```

agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * <ul>
+ *   <li>State is lost when the JVM exits
+ *   <li>Not suitable for distributed environments
+ *   <li>Memory usage grows with the number of sessions
+ * </ul>
+ */
+public class InMemoryAgentStateStore implements AgentStateStore {
+
+    /** Sentinel namespace for callers that pass {@code userId == null}. */
```
AgentScopeJavaBot, line 48: [minor] **Unbounded memory growth.** The `ConcurrentHashMap<String, Map<String, SessionData>> users` map grows with every new `(userId, sessionId)` pair and never evicts. For long-running applications (web services handling thousands of sessions), this will eventually OOM.

Consider adding a TTL-based eviction, LRU cap, or at minimum document that callers are responsible for calling `delete(userId, sessionId)` to clean up completed sessions.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -354,32 +470,67 @@ private static ReactConfig assembleReactConfig(Builder b) {
...
+        this.activeRc = ctx;
         bindRuntimeContextToHooks(ctx);
-        // Reset per-call system message; will be initialised by consumeSystemMsgAfterPreCall
-        currentSystemMsg.set(null);
+        // Seed per-call state onto the active execution scope. The system message is initialised
+        // by consumeSystemMsgAfterPreCall; the event sink (if any) is bound in doCall() from the
+        // per-subscription Reactor Context carried by streamEvents.
+        scope.rc = ctx;
+        scope.systemMsg = null;
+        // Clear any stale interrupt signal for this session before the new call begins.
```
AgentScopeJavaBot, line 511: [nit] **`InterruptControl.reset()` sets 3 fields non-atomically.** While this is called before a new call starts (in `beforeAgentExecution`) and the `flag.set(false)` → `source.set(USER)` → `userMessage.set(null)` ordering is safe in practice (no concurrent reader exists at this point), consider adding a brief comment explaining the invariant: "reset() is called single-threaded before the new call begins, so no reader can observe a partially-reset state."

agentscope-core/src/main/java/io/agentscope/core/state/JsonFileAgentStateStore.java
```diff
@@ -0,0 +1,387 @@
...
+                        StandardCharsets.UTF_8,
+                        StandardOpenOption.CREATE,
+                        StandardOpenOption.WRITE)) {
+            for (State item : values) {
+                writer.write(JsonUtils.getJsonCodec().toJson(item));
+                writer.newLine();
+            }
+        }
+    }
+
```
AgentScopeJavaBot, line 151: [nit] **`rewriteEntireList` deletes then creates the file, losing data on crash.** `Files.deleteIfExists(file)` followed by `Files.newBufferedWriter(file, CREATE, WRITE)` means if the process crashes between the delete and the write, the list data is permanently lost. Same atomic-write recommendation as the single-value `save()` applies here — write to `.tmp` then rename.

oss-maintainer (approved):
LGTM.

PR #1684, opened 2026-06-10
fix sandbox glob_files recursive patterns

gaoxiaolei-s59:
## Summary
- normalize leading `**/` in sandbox `glob()` before passing the pattern to `find -name`
- add a regression test covering `glob_files` calls like `**/*.md`

## Root Cause
The `glob_files` tool description advertises recursive patterns such as `**/*.java`, but `BaseSandboxFilesystem.glob()` forwarded that pattern directly to `find -name`. `find -name` matches the filename portion only, so the leading `**/` caused sandbox-backed recursive glob calls to return no matches.

## Validation
- `mvn -pl agentscope-harness -am -Dtest=BaseSandboxFilesystemTest,FilesystemGlobTest test`

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
@@ -279,7 +279,7 @@ public abstract class BaseSandboxFilesystem implements AbstractSandboxFilesystem
         String grepOpts = "-rHnF";
         String globPattern = "";
         if (glob != null && !glob.isBlank()) {
-            globPattern = "--include=" + FilesystemUtils.shellQuote(glob);
+            globPattern = "--include=" + FilesystemUtils.shellQuote(stripRecursivePrefix(glob));
         }
         String patternEscaped = FilesystemUtils.shellQuote(pattern);
 
@@ -319,9 +319,7 @@ public abstract class BaseSandboxFilesystem implements AbstractSandboxFilesystem
     @Override
     public GlobResult glob(RuntimeContext runtimeContext, String pattern, String path) {
         String escapedPath = FilesystemUtils.shellQuote(path != null ? path : "/");
-        String findNamePattern =
-                pattern != null && pattern.startsWith("**/") ? pattern.substring(3) : pattern;
-        String escapedPattern = FilesystemUtils.shellQuote(findNamePattern);
+        String escapedPattern = FilesystemUtils.shellQuote(stripRecursivePrefix(pattern));
 
         String cmd =
                 "find " + escapedPath + " -type f -name " + escapedPattern + " 2>/dev/null | sort";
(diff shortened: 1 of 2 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
```diff
@@ -319,7 +319,9 @@ public GrepResult grep(
     @Override
     public GlobResult glob(RuntimeContext runtimeContext, String pattern, String path) {
         String escapedPath = FilesystemUtils.shellQuote(path != null ? path : "/");
-        String escapedPattern = FilesystemUtils.shellQuote(pattern);
+        String findNamePattern =
```
AgentScopeJavaBot: [nit] Symmetry note: the sibling `grep()` method (line ~282) passes its `glob` parameter directly to `grep --include=` without the same `**/` stripping. GNU grep's `--include` also matches only the filename portion, so `grep_files` with `**/*.java` would silently return no matches — the same bug this PR fixes for `glob_files`. Consider applying the same normalization in a follow-up, or extracting a shared helper like `stripRecursivePrefix(String)`.

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystemTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.junit.jupiter.api.Test;
+
+class BaseSandboxFilesystemTest {
+
+    private static final RuntimeContext RT = RuntimeContext.empty();
+
+    @Test
```
AgentScopeJavaBot, line 35: [praise] Nice test design — capturing the generated command string and asserting it directly is the right approach for testing shell-command delegation. The `FakeSandboxFilesystem` stub is minimal and focused.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a real bug: `glob_files` in sandbox mode advertised support for recursive patterns like `**/*.java`, but the implementation forwarded the pattern directly to `find -name`, which only matches the filename portion. The fix correctly strips the leading `**/` prefix since `find` already performs recursive directory traversal. The change is minimal, null-safe, and well-tested with a new `BaseSandboxFilesystemTest` using a clean fake-sandbox design. One minor symmetry observation regarding the sibling `grep` method is noted below.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
AgentScopeJavaBot: [nit] Symmetry note: the sibling `grep()` method (line ~282) passes its `glob` parameter directly to `grep --include=` without the same `**/` stripping. GNU grep's `--include` also matches only the filename portion, so `grep_files` with `**/*.java` would silently return no matches — the same bug this PR fixes for `glob_files`. Consider applying the same normalization in a follow-up, or extracting a shared helper like `stripRecursivePrefix(String)`.

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystemTest.java
AgentScopeJavaBot, line 35: [praise] Nice test design — capturing the generated command string and asserting it directly is the right approach for testing shell-command delegation. The `FakeSandboxFilesystem` stub is minimal and focused.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a real bug: `glob_files` in sandbox mode advertised support for recursive patterns like `**/*.java`, but the implementation forwarded the pattern directly to `find -name`, which only matches the filename portion. The fix correctly strips the leading `**/` prefix since `find` already performs recursive directory traversal. The change is minimal, null-safe, and well-tested with a new `BaseSandboxFilesystemTest` using a clean fake-sandbox design. One minor symmetry observation regarding the sibling `grep` method is noted below.

chickenlj (approved):

PR #1698, opened 2026-06-10
fix(memory): toRelative strips workspace root for ROOTED local filesystem paths

McClone:


agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryConsolidator.java
```diff
@@ -264,14 +265,22 @@ private static String fileName(String path) {
...
+     * with {@link WorkspaceManager#readManagedWorkspaceFileUtf8}. Strips the
+     * workspace root prefix when the path is an absolute local filesystem path;
+     * strips only the leading {@code /} for SANDBOXED virtual paths.
      */
-    private static String toRelative(String path) {
+    private String toRelative(String path) {
         if (path == null) {
             return "";
         }
+        Path p = Path.of(path);
```
AgentScopeJavaBot, line 277: [nitpick] No regression test is included for the bug being fixed. The existing `MemoryConsolidatorFilesystemTest` exercises the Remote/InMemoryStore path but never covers the local ROOTED (non-sandboxed, non-namespaced) case where `FileInfo.path()` is a real absolute filesystem path. Consider adding a unit test that creates a local `WorkspaceManager` with a `@TempDir`, seeds a `memory/` directory, and asserts that `consolidate()` reads daily ledgers correctly — or at minimum a package-private test that directly invokes `toRelative()` with all three path shapes (SANDBOXED `/memory/x.md`, ROOTED absolute `/tmp/ws/memory/x.md`, namespace-relative `ns/memory/x.md`).

AgentScopeJavaBot (approved):
🤖 AI Review

This is a small, focused bugfix for `MemoryConsolidator.toRelative()`. The old implementation naively stripped the leading `/` from any path, which worked for SANDBOXED virtual paths (e.g. `/memory/file.md` → `memory/file.md`) but produced incorrect results for ROOTED local filesystem paths (e.g. `/home/user/workspace/memory/file.md` → `home/user/workspace/memory/file.md`). The fix correctly uses `java.nio.file.Path` to detect absolute paths rooted under the workspace and relativize them. The static-to-instance conversion is necessary and correct. Downstream `normalizeRelativePath()` already handles OS-specific separators, so no cross-platform issue. The only gap is the lack of a regression test for the exact scenario being fixed.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryConsolidator.java
AgentScopeJavaBot, line 277: [nitpick] No regression test is included for the bug being fixed. The existing `MemoryConsolidatorFilesystemTest` exercises the Remote/InMemoryStore path but never covers the local ROOTED (non-sandboxed, non-namespaced) case where `FileInfo.path()` is a real absolute filesystem path. Consider adding a unit test that creates a local `WorkspaceManager` with a `@TempDir`, seeds a `memory/` directory, and asserts that `consolidate()` reads daily ledgers correctly — or at minimum a package-private test that directly invokes `toRelative()` with all three path shapes (SANDBOXED `/memory/x.md`, ROOTED absolute `/tmp/ws/memory/x.md`, namespace-relative `ns/memory/x.md`).

AgentScopeJavaBot (approved):
🤖 AI Review

This is a small, focused bugfix for `MemoryConsolidator.toRelative()`. The old implementation naively stripped the leading `/` from any path, which worked for SANDBOXED virtual paths (e.g. `/memory/file.md` → `memory/file.md`) but produced incorrect results for ROOTED local filesystem paths (e.g. `/home/user/workspace/memory/file.md` → `home/user/workspace/memory/file.md`). The fix correctly uses `java.nio.file.Path` to detect absolute paths rooted under the workspace and relativize them. The static-to-instance conversion is necessary and correct. Downstream `normalizeRelativePath()` already handles OS-specific separators, so no cross-platform issue. The only gap is the lack of a regression test for the exact scenario being fixed.

Issue #1710, opened 2026-06-11
[Bug]: KubernetesFilesystemSpec + JdbcSnapshotSpec: RemoteSandboxSnapshot deserialization fails on second call due to @JsonTypeInfo "type" field conflict


Buktal:
**Describe the bug**

Using `KubernetesFilesystemSpec` together with `JdbcSnapshotSpec` causes a `SandboxConfigurationException` on every request after the first one. The first request succeeds and persists a `KubernetesSandboxState` (containing a `RemoteSandboxSnapshot`) to the `AgentStateStore`. On the next request, `SandboxLifecycleMiddleware` calls `SandboxManager.acquire`, which calls `KubernetesSandboxClient.deserializeState` on the stored JSON, and Jackson throws `UnrecognizedPropertyException: Unrecognized field "type"` against `RemoteSandboxSnapshot`.

Root cause summary:

1. **`RemoteSandboxSnapshot.getType()` conflicts with `@JsonTypeInfo(As.PROPERTY, property = "type")` on the `SandboxSnapshot` interface.**

   `SandboxSnapshot` is annotated with:
   ```java
   @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
   @JsonSubTypes({
       @JsonSubTypes.Type(value = RemoteSandboxSnapshot.class, name = "remote"),
       ...
   })
   ```
   This causes Jackson to emit a `"type": "remote"` discriminator field when serializing `RemoteSandboxSnapshot`. During deserialization, Jackson resolves the subtype correctly, but then the concrete `RemoteSandboxSnapshot` class encounters `"type"` again as an ordinary field. Because `RemoteSandboxSnapshot` has a `getType()` method (returning `"remote"`) but no corresponding setter or constructor parameter, and is not annotated with `@JsonIgnoreProperties(ignoreUnknown = true)`, Jackson fails with `UnrecognizedPropertyException`.

   The source-code comment on `RemoteSandboxSnapshot` already acknowledges the class is *"not directly serializable to JSON because `RemoteSnapshotClient` cannot be serialized"*, yet no guard is in place to prevent it from being serialized as part of `KubernetesSandboxState`.

   Affected locations:
   - `agentscope-harness/.../snapshot/RemoteSandboxSnapshot.java` — missing `@JsonIgnoreProperties(ignoreUnknown = true)` and `@JsonIgnore` on `getType()`
[...]

PR #1711, opened 2026-06-11
Fix/kubernetes jdbc snapshot deserialization

Buktal:
## Summary

RemoteSandboxSnapshot had no @JsonCreator, causing Jackson to use _deserializeUsingPropertyBased and treat the @JsonTypeInfo type discriminator field "type" as an unrecognised bean property, throwing UnrecognizedPropertyException on every second call when using KubernetesFilesystemSpec + JdbcSnapshotSpec. This PR fixes deserialization for all snapshot implementations and ensures the live RemoteSnapshotClient is re-injected at resume time.

Fixes / Relates to: #1710

---
### Background & Purpose

SandboxSnapshot declares @JsonTypeInfo(As.PROPERTY, property = "type"). When a concrete subtype has no @JsonCreator and no no-arg constructor, Jackson falls through to _deserializeUsingPropertyBased, which buffers all token stream properties — including the "type" discriminator — before constructing the object. Since none of the snapshot classes had a constructor parameter named "type", Jackson threw UnrecognizedPropertyException.

A second issue existed downstream: even if deserialization succeeded, RemoteSandboxSnapshot.client would be null after deserialization (the client is a runtime object, not serializable). SandboxManager never re-injected it before calling client.resume(state), so the subsequent isRestorable() call in AbstractBaseSandbox.start() would NPE.

### Changes Made

- RemoteSandboxSnapshot: Added @JsonCreator constructor accepting only @JsonProperty("id"); added @JsonIgnore on client field, getType(), isRestorable(), and restore(); added null-guards in persist(), restore(), and isRestorable() (returns false when client is null, causing safe degradation to cold start); added @JsonIgnoreProperties(ignoreUnknown = true)
- LocalSandboxSnapshot: Added @JsonCreator + @JsonProperty on existing constructor; added @JsonIgnore on getType() and isRestorable(); added @JsonIgnoreProperties(ignoreUnknown = true)
- NoopSandboxSnapshot: Added @JsonIgnore on getType() and isPersistenceEnabled(); added @JsonIgnoreProperties(ignoreUnknown = true)
- SandboxManager: In Priority 3 (resume from persisted state), re-injects the live snapshot client by calling snapshotSpec.build(snapshotId) immediately after deserializeState, before client.resume(state)
- Tests: Added snapshot round-trip tests for all three types in HarnessSandboxJacksonModuleTest; added regression test roundTripKubernetesStateWithRemoteSnapshot in KubernetesSandboxStateSerdeTest

### How to Test

1. Run mvn test -pl agentscope-harness -Dtest="HarnessSandboxJacksonModuleTest,SandboxManagerIsolationTest,SessionSandboxStateStoreTest" — all tests should pass
2. Run mvn test -pl agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes -Dtest="KubernetesSandboxStateSerdeTest" — both tests including roundTripKubernetesStateWithRemoteSnapshot should pass
3. To reproduce the original bug before this fix: configure KubernetesFilesystemSpec + JdbcSnapshotSpec, make two consecutive calls to the same session — the second call would throw UnrecognizedPropertyException: "type" on RemoteSandboxSnapshot

---
## Checklist

- [x] Code has been formatted according to project guidelines
- [x] All automated tests are passing
- [x] Documentation / inline comments have been updated accordingly


AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes Jackson deserialization for all three SandboxSnapshot implementations and ensures RemoteSnapshotClient is re-injected at resume time.

**Recommendation**: COMMENT — 2 critical issues + 2 recommended + 3 nitpicks

PR #1743, opened 2026-06-13
fix(agent): fallback to synthetic tool when native structured output fails

### LINKED ISSUE #1720: [Bug]:2.0版本调用deepseek-v4-flash 结构化输出异常
代码：
```
Msg response = agent.call(List.of(msg), AgentRepo.class, ctx).block();
AgentRepo result = response.getStructuredData(AgentRepo.class);
```

异常信息：
io.agentscope.core.model.exception.BadRequestException: HTTP transport error during streaming: HTTP request failed with status 400 | {"error":{"message":"This response_format type is unavailable now","type":"invalid_request_error","param":null,"code":"invalid_request_error"}} | Response body: {"error":{"message":"This response_format type is unavailable now","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
      at io.agentscope.core.model.exception.OpenAIException.create(OpenAIException.java:79)
      at io.agentscope.core.model.OpenAIClient.lambda$stream$2(OpenAIClient.java:431)
      at reactor.core.publisher.Flux.lambda$onErrorMap$28(Flux.java:7318)

环境信息:
AgentScope-Java Version: 2.0.0-RC3
Java Version: 17
OS: windows
Model: deepseek-v4-flash

### LINKED ISSUE #1742: [Bug]: OpenAIChatModel structuted output fails on DeepSeek/vLLM — supportsNativeStructuredOutput() returns true for unsupported backends
## Summary

`OpenAIChatModel.supportsNativeStructuredOutput()` hardcodes `true`, but many OpenAI-compatible backends (DeepSeek, vLLM, Ollama, SiliconFlow) do not support `response_format`. This causes `agent.call(msg, OutputClass.class)` to fail with HTTP 400 on those backends.

## Root Cause

`OpenAIChatModel.java:195`:

```java
@Override
public boolean supportsNativeStructuredOutput() {
    return true;  // ← assumes ALL OpenAI-compatible APIs support response_format
}
```

`ReActAgent.doStructuredCall()` trusts this unconditionally:

```java
if (model.supportsNativeStructuredOutput()) {
    return doNativeStructuredCall(msgs, jsonSchema);  // sends response_format → 400
}
```

## Reproduction

```java
    .model(OpenAIChatModel.builder()
        .baseUrl("https://api.deepseek.com")
        .modelName("deepseek-chat")
        .formatter(new DeepSeekFormatter())
        .build())
    .build();

agent.call(userMsg, OutputClass.class).block();
// DeepSeek: "This response_format type is unavailable now" (400)
```

## Affected Backends

All OpenAI-compatible APIs using `OpenAIChatModel` that don't support `response_format`: DeepSeek, vLLM, Ollama, SiliconFlow, TogetherAI (partial).

## Fix

Add automatic fallback in `doStructuredCall()`: when native path fails, retry with the fallback (generate_response tool) path.

Sparkle6979:
## Problem

`OpenAIChatModel.supportsNativeStructuredOutput()` returns `true` for all OpenAI-compatible backends, but DeepSeek, vLLM, and other providers do not support `response_format`. This causes `agent.call(msg, OutputClass.class)` to fail with HTTP 400.

Fixes #1742, fixes #1720

## Changes

Three changes in `ReActAgent.java`:

**1. `doStructuredCall()` — catch native failures, route to fallback**

```java
return doNativeStructuredCall(msgs, jsonSchema)
        .onErrorResume(e -> {
            log.warn("Native structured output failed — falling back to synthetic tool path");
            return doFallbackStructuredCall(msgs, jsonSchema);
        });
```

**2. `doNativeStructuredCall()` — roll back context on failure**

Before starting `doCallInner`, snapshot the context size. On error via `doOnError`: roll back messages added by the failed native attempt, and clear `scope.nativeResponseFormat`. This prevents the native call's `addToContext(msgs)` from being duplicated when the fallback re-runs.

**3. `reasoning()` — guard native format in fallback mode**

```java
if (nativeResponseFormat != null \u0026\u0026 soTool == null) {
```

`nativeResponseFormat` and `soTool` are mutually exclusive — the guard prevents stale `nativeResponseFormat` from being injected during fallback, even if future code paths fail to clear it.

## Design

- **Auto-fallback, zero configuration.** No Builder flags, no user knowledge required. True OpenAI backends succeed on the native path; backends that crash now succeed via fallback.
- **Cleanup on failure, not retry.** `doNativeStructuredCall` is responsible for managing its own side effects — it uses `doOnError` to roll back context and clear scope state, then lets the error propagate to `doStructuredCall` which handles the fallback routing.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -1027,11 +1027,25 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                             .strict(true)
                                             .build());
 
+                    int contextSizeBefore = scope.state.contextMutable().size();
+
                     return scope.doCallInner(msgs)
                             .flatMap(
                                     result -> {
                                         Msg out = wrapNativeStructuredResult(result);
                                         return saveStateToSession(scope).thenReturn(out);
+                                    })
+                            .onErrorResume(
+                                    e -> {
+                                        scope.rollbackContext(contextSizeBefore);
+                                        scope.nativeResponseFormat = null;
+                                        log.warn(
+                                                "Native structured output failed ({}) — falling"
+                                                        + " back to synthetic tool path",
+                                                e.getMessage() != null
+                                                        ? e.getMessage()
+                                                        : e.getClass().getSimpleName());
+                                        return doFallbackStructuredCall(msgs, jsonSchema);
                                     });
                 });
     }
@@ -1808,6 +1822,14 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
         private void addToContext(List<Msg> msgs) {
             if (msgs != null) {
                 state.contextMutable().addAll(msgs);
+            }
+        }
+
+        /** Roll back context to a given size on fallback, e.g. after native structured output failure. */
+        void rollbackContext(int targetSize) {
+            List<Msg> ctx = state.contextMutable();
+            while (ctx.size() > targetSize) {
+                ctx.remove(ctx.size() - 1);
             }
         }
 
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -989,7 +989,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                         e.getMessage() != null
                                                 ? e.getMessage()
                                                 : e.getClass().getSimpleName());
-                                return doFallbackStructuredCall(List.of(), jsonSchema);
+                                return doFallbackStructuredCall(msgs, jsonSchema);
                             });
         }
         return doFallbackStructuredCall(msgs, jsonSchema);
@@ -1035,17 +1035,10 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                         Msg out = wrapNativeStructuredResult(result);
                                         return saveStateToSession(scope).thenReturn(out);
                                     })
-                            .onErrorResume(
+                            .doOnError(
                                     e -> {
                                         scope.rollbackContext(contextSizeBefore);
                                         scope.nativeResponseFormat = null;
-                                        log.warn(
-                                                "Native structured output failed ({}) — falling"
-                                                        + " back to synthetic tool path",
-                                                e.getMessage() != null
-                                                        ? e.getMessage()
-                                                        : e.getClass().getSimpleName());
-                                        return doFallbackStructuredCall(msgs, jsonSchema);
                                     });
                 });
     }
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -980,7 +980,17 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                         ? JsonSchemaUtils.generateSchemaFromClass(targetClass)
                         : JsonSchemaUtils.generateSchemaFromJsonNode(schemaDesc);
         if (model.supportsNativeStructuredOutput()) {
-            return doNativeStructuredCall(msgs, jsonSchema);
+            return doNativeStructuredCall(msgs, jsonSchema)
+                    .onErrorResume(
+                            e -> {
+                                log.warn(
+                                        "Native structured output failed ({}) — falling back to"
+                                                + " synthetic tool path",
+                                        e.getMessage() != null
+                                                ? e.getMessage()
+                                                : e.getClass().getSimpleName());
+                                return doFallbackStructuredCall(msgs, jsonSchema);
+                            });
         }
         return doFallbackStructuredCall(msgs, jsonSchema);
     }
@@ -1850,7 +1860,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                         event.getEffectiveGenerateOptions() != null
                                                 ? event.getEffectiveGenerateOptions()
                                                 : buildGenerateOptions();
-                                if (nativeResponseFormat != null) {
+                                if (nativeResponseFormat != null && soTool == null) {
                                     options =
                                             GenerateOptions.mergeOptions(
                                                     GenerateOptions.builder()
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -989,7 +989,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                         e.getMessage() != null
                                                 ? e.getMessage()
                                                 : e.getClass().getSimpleName());
-                                return doFallbackStructuredCall(msgs, jsonSchema);
+                                return doFallbackStructuredCall(List.of(), jsonSchema);
                             });
         }
         return doFallbackStructuredCall(msgs, jsonSchema);
--- a/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentStructuredOutputTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentStructuredOutputTest.java
@@ -31,11 +31,13 @@ import io.agentscope.core.message.ThinkingBlock;
 import io.agentscope.core.message.ToolUseBlock;
 import io.agentscope.core.model.ChatResponse;
 import io.agentscope.core.model.ChatUsage;
+import io.agentscope.core.model.Model;
 import io.agentscope.core.tool.Toolkit;
 import io.agentscope.core.util.JsonUtils;
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.RepeatedTest;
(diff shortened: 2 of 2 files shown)


agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -980,7 +980,17 @@ private Mono<Msg> doStructuredCall(List<Msg> msgs, Class<?> targetClass, JsonNod
...
+            return doNativeStructuredCall(msgs, jsonSchema)
+                    .onErrorResume(
+                            e -> {
+                                log.warn(
+                                        "Native structured output failed ({}) — falling back to"
+                                                + " synthetic tool path",
+                                        e.getMessage() != null
+                                                ? e.getMessage()
+                                                : e.getClass().getSimpleName());
+                                return doFallbackStructuredCall(msgs, jsonSchema);
```
AgentScopeJavaBot, line 992: [major] **State pollution: duplicate input messages on fallback.** Both `doNativeStructuredCall` and `doFallbackStructuredCall` call `scope.doCallInner(msgs)`, which synchronously invokes `addToContext(msgs)` before starting the ReAct loop. When the native path fails after `addToContext` has already run, the fallback re-enters `doCallInner(msgs)` on the same scope, adding the input messages a second time to `state.contextMutable()`. The polluted state is then persisted via `saveStateToSession` in the fallback's `.flatMap`, affecting subsequent turns in the same session.

Suggested fix — snapshot the context size before the native attempt and trim back on fallback, e.g.:
```java
// Inside doNativeStructuredCall, before doCallInner:
int contextSizeBefore = scope.state.contextMutable().size();

    .onErrorResume(e -> {
        // Roll back messages added by the failed native attempt
        List<Msg> ctx = scope.state.contextMutable();
        while (ctx.size() > contextSizeBefore) {
            ctx.remove(ctx.size() - 1);
        }
        scope.nativeResponseFormat = null; // clear stale state
        log.warn("Native structured output failed ({}) — falling back",
            e.getMessage() != null ? e.getMessage() : e.getClass().getSimpleName());
        return doFallbackStructuredCall(msgs, jsonSchema);
    });
```
This requires moving the `onErrorResume` inside `doNativeStructuredCall`'s `deferContextual` where `scope` is accessible.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -980,7 +980,17 @@ private Mono<Msg> doStructuredCall(List<Msg> msgs, Class<?> targetClass, JsonNod
                         ? JsonSchemaUtils.generateSchemaFromClass(targetClass)
                         : JsonSchemaUtils.generateSchemaFromJsonNode(schemaDesc);
         if (model.supportsNativeStructuredOutput()) {
-            return doNativeStructuredCall(msgs, jsonSchema);
+            return doNativeStructuredCall(msgs, jsonSchema)
+                    .onErrorResume(
```
AgentScopeJavaBot, line 984: [minor] **Catch-all `onErrorResume` may mask programming errors.** The current handler catches every `Throwable`, including `NullPointerException`, `ClassCastException`, and other bugs in the native path implementation. While the fallback is functionally correct for these cases too, silently retrying via a different code path can make debugging significantly harder — the original error is only logged at WARN level.

Consider narrowing the catch scope or logging at ERROR for unexpected exception types:
```java
.onErrorResume(e -> {
    if (e instanceof RuntimeException
            && !(e.getCause() instanceof java.net.HttpRetryException)
            && e.getClass() != RuntimeException.class) {
        // Likely a programming error — re-throw
        return Mono.error(e);
    }
    log.warn("Native structured output failed ({}) — falling back",
        e.getMessage() != null ? e.getMessage() : e.getClass().getSimpleName());
    return doFallbackStructuredCall(msgs, jsonSchema);
});
```
Alternatively, log at ERROR level for non-HTTP exceptions to make them easier to spot in production logs.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1850,7 +1860,7 @@ private Mono<Msg> reasoning(int iter, boolean ignoreMaxIters) {
                                         event.getEffectiveGenerateOptions() != null
                                                 ? event.getEffectiveGenerateOptions()
                                                 : buildGenerateOptions();
-                                if (nativeResponseFormat != null) {
+                                if (nativeResponseFormat != null && soTool == null) {
```
AgentScopeJavaBot, line 1878: [nit] **Stale `nativeResponseFormat` remains on the scope after fallback.** The guard `&& soTool == null` correctly prevents injection, but the stale value persists on the `CallExecution` scope. If any future code path checks `nativeResponseFormat` without also guarding on `soTool`, this bug will silently resurface. Consider explicitly clearing `scope.nativeResponseFormat = null` in the fallback path (once the state-pollution fix above gives you scope access in the `onErrorResume` handler) to make the invariant self-documenting.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentStructuredOutputTest.java
```diff
@@ -640,4 +642,83 @@ private void runSubscribeOnThenSequentialSecondCallStructuredOutputScenario(
         WeatherResponse result2 = response2.getStructuredData(WeatherResponse.class);
         assertNotNull(result2);
     }
+
+    @Test
+    @DisplayName("Native structured output failure should fallback to synthetic tool path")
+    void testNativeStructuredOutputFallback() {
```
AgentScopeJavaBot, line 648: [minor] **Test does not verify conversation state integrity.** The test confirms the fallback produces correct structured output, but does not assert that the conversation context is free of duplicate messages after fallback. Given the state-pollution risk identified in `doStructuredCall`, consider adding an assertion that inspects the agent's internal state (or the messages passed to the mock model on its second invocation) to verify the input message appears exactly once:
```java
// After the call completes, verify no duplicate user messages in context
List<Msg> contextMessages = agent.getState().getContext(); // or equivalent accessor
long userMsgCount = contextMessages.stream()
    .filter(m -> m.getRole() == MsgRole.USER)
    .count();
assertEquals(1, userMsgCount, "Input message should appear exactly once in context");
```

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR adds an auto-fallback mechanism in `ReActAgent.doStructuredCall()`: when the native `response_format` path fails (e.g., DeepSeek/vLLM returning HTTP 400), it retries via the synthetic `generate_response` tool path. A guard `&& soTool == null` is added in `reasoning()` to prevent the stale `nativeResponseFormat` from being injected during fallback. The idea is sound and solves a real pain point for non-OpenAI backends.

However, there is a **state pollution bug**: both the native and fallback paths call `scope.doCallInner(msgs)`, which invokes `addToContext(msgs)` on the same shared `CallExecution` scope. When the native path fails and the fallback re-runs `doCallInner`, the input messages are added to the conversation context a second time. This duplicated state is then persisted to the session via `saveStateToSession`, potentially affecting subsequent calls.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot, line 992: [major] **State pollution: duplicate input messages on fallback.** Both `doNativeStructuredCall` and `doFallbackStructuredCall` call `scope.doCallInner(msgs)`, which synchronously invokes `addToContext(msgs)` before starting the ReAct loop. When the native path fails after `addToContext` has already run, the fallback re-enters `doCallInner(msgs)` on the same scope, adding the input messages a second time to `state.contextMutable()`. The polluted state is then persisted via `saveStateToSession` in the fallback's `.flatMap`, affecting subsequent turns in the same session.

Suggested fix — snapshot the context size before the native attempt and trim back on fallback, e.g.:
```java
// Inside doNativeStructuredCall, before doCallInner:
int contextSizeBefore = scope.state.contextMutable().size();

    .onErrorResume(e -> {
        // Roll back messages added by the failed native attempt
        List<Msg> ctx = scope.state.contextMutable();
        while (ctx.size() > contextSizeBefore) {
            ctx.remove(ctx.size() - 1);
        }
        scope.nativeResponseFormat = null; // clear stale state
        log.warn("Native structured output failed ({}) — falling back",
            e.getMessage() != null ? e.getMessage() : e.getClass().getSimpleName());
        return doFallbackStructuredCall(msgs, jsonSchema);
    });
```
This requires moving the `onErrorResume` inside `doNativeStructuredCall`'s `deferContextual` where `scope` is accessible.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot, line 984: [minor] **Catch-all `onErrorResume` may mask programming errors.** The current handler catches every `Throwable`, including `NullPointerException`, `ClassCastException`, and other bugs in the native path implementation. While the fallback is functionally correct for these cases too, silently retrying via a different code path can make debugging significantly harder — the original error is only logged at WARN level.

Consider narrowing the catch scope or logging at ERROR for unexpected exception types:
```java
.onErrorResume(e -> {
    if (e instanceof RuntimeException
            && !(e.getCause() instanceof java.net.HttpRetryException)
            && e.getClass() != RuntimeException.class) {
        // Likely a programming error — re-throw
        return Mono.error(e);
    }
    log.warn("Native structured output failed ({}) — falling back",
        e.getMessage() != null ? e.getMessage() : e.getClass().getSimpleName());
    return doFallbackStructuredCall(msgs, jsonSchema);
});
```
Alternatively, log at ERROR level for non-HTTP exceptions to make them easier to spot in production logs.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot, line 1878: [nit] **Stale `nativeResponseFormat` remains on the scope after fallback.** The guard `&& soTool == null` correctly prevents injection, but the stale value persists on the `CallExecution` scope. If any future code path checks `nativeResponseFormat` without also guarding on `soTool`, this bug will silently resurface. Consider explicitly clearing `scope.nativeResponseFormat = null` in the fallback path (once the state-pollution fix above gives you scope access in the `onErrorResume` handler) to make the invariant self-documenting.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentStructuredOutputTest.java
AgentScopeJavaBot, line 648: [minor] **Test does not verify conversation state integrity.** The test confirms the fallback produces correct structured output, but does not assert that the conversation context is free of duplicate messages after fallback. Given the state-pollution risk identified in `doStructuredCall`, consider adding an assertion that inspects the agent's internal state (or the messages passed to the mock model on its second invocation) to verify the input message appears exactly once:
```java
// After the call completes, verify no duplicate user messages in context
List<Msg> contextMessages = agent.getState().getContext(); // or equivalent accessor
long userMsgCount = contextMessages.stream()
    .filter(m -> m.getRole() == MsgRole.USER)
    .count();
assertEquals(1, userMsgCount, "Input message should appear exactly once in context");
```

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR adds an auto-fallback mechanism in `ReActAgent.doStructuredCall()`: when the native `response_format` path fails (e.g., DeepSeek/vLLM returning HTTP 400), it retries via the synthetic `generate_response` tool path. A guard `&& soTool == null` is added in `reasoning()` to prevent the stale `nativeResponseFormat` from being injected during fallback. The idea is sound and solves a real pain point for non-OpenAI backends.

However, there is a **state pollution bug**: both the native and fallback paths call `scope.doCallInner(msgs)`, which invokes `addToContext(msgs)` on the same shared `CallExecution` scope. When the native path fails and the fallback re-runs `doCallInner`, the input messages are added to the conversation context a second time. This duplicated state is then persisted to the session via `saveStateToSession`, potentially affecting subsequent calls.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
Sparkle6979: Fixed: moved onErrorResume into doNativeStructuredCall's deferContextual scope. Before retrying, doOnError rolls back context to the snapshot taken before the native call, so messages added by the failed native attempt are removed before the fallback adds them. Also clears nativeResponseFormat on the scope.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
Sparkle6979: Understood. However, narrowing the catch scope is risky here — different HTTP transports (JDK, OkHttp) throw different exception types for the same 400 error, and the auto-fallback is a best-effort recovery. Silent retry via fallback is the safer default. If this pattern causes debugging issues in practice, we can revisit.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
Sparkle6979: Fixed: doNativeStructuredCall's doOnError now clears scope.nativeResponseFormat = null on failure. The soTool == null guard in reasoning() is retained as a second line of defense if any future code path bypasses the cleanup.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentStructuredOutputTest.java
Sparkle6979: Acknowledged. The context rollback via rollbackContext in doOnError structurally prevents duplicate messages, so the correctness is guaranteed at the implementation level even without an explicit assertion. That said, adding a context integrity check to the existing test is a reasonable improvement — will add this in a follow-up.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot: ✅ Verified as addressed in `c5eb090`: rollbackContext(contextSizeBefore) is called in doOnError before fallback, and rollbackContext method exists at line 1822, preventing duplicate input messages in context.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot: ✅ Verified as addressed in `c5eb090`: doOnError at line 1041 now sets scope.nativeResponseFormat = null on the native path failure.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot: ✅ Verified as addressed in `c5eb090`: rollbackContext(contextSizeBefore) in doOnError at line 1038-1041 addresses the duplicate-context issue.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot: ✅ Verified as addressed in `c5eb090`: scope.nativeResponseFormat = null is now set in doOnError at line 1041.

AgentScopeJavaBot (approved):
All blocking review comments have been addressed. 4 non-blocking suggestion(s) (minor/nit) remain open and may be addressed in a follow-up.

PR #1756, opened 2026-06-14
fix: preserve parent RuntimeContext in harness subagent delegation

guslegend0510:
## Summary
- add RuntimeContext.fork() to carry string extras, typed attributes, and ToolExecutionContext into child calls
- pass the parent RuntimeContext through harness subagent invocation paths
- add regression tests for invokeAgent and invokeAgentStream

## Validation
- mvn -pl agentscope-harness -am "-Dtest=RuntimeContextTest,DefaultAgentManagerRuntimeContextTest" test

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
@@ -236,6 +236,10 @@ public class RuntimeContext {
         if (TYPED_DEFAULT_KEY.equals(key) && type == RuntimeContext.class) {
             return (T) this;
         }
+        Object fromString = stringAttributes.get(key);
+        if (type.isInstance(fromString)) {
+            return (T) fromString;
+        }
         return null;
     }
 
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
@@ -76,6 +76,33 @@ public class RuntimeContext {
         }
     }
 
+    private RuntimeContext(
+            String sessionId,
+            String userId,
+            Map<String, Object> stringAttributes,
+            Map<Class<?>, ? extends Map<String, Object>> typedAttributes,
+            ToolExecutionContext toolExecutionContext) {
+        this.sessionId = sessionId;
+        this.userId = userId;
+        this.stringAttributes = new ConcurrentHashMap<>();
+        this.typedAttributes = new ConcurrentHashMap<>();
+        this.toolExecutionContext = toolExecutionContext;
+        this.agentState = null;
+        if (stringAttributes != null && !stringAttributes.isEmpty()) {
+            this.stringAttributes.putAll(stringAttributes);
+        }
+        if (typedAttributes != null && !typedAttributes.isEmpty()) {
+            for (Map.Entry<Class<?>, ? extends Map<String, Object>> e :
+                    typedAttributes.entrySet()) {
+                Map<String, Object> values = e.getValue();
+                if (values == null || values.isEmpty()) {
+                    continue;
+                }
+                this.typedAttributes.put(e.getKey(), new ConcurrentHashMap<>(values));
+            }
+        }
+    }
+
     /**
      * Shallow, mutable empty context (null session fields, empty attribute maps, no tool context).
      */
@@ -106,6 +133,22 @@ public class RuntimeContext {
      */
     public void setAgentState(AgentState agentState) {
         this.agentState = agentState;
+    }
+
+    /**
+     * Creates a child context that preserves this context's extras, typed attributes, and tool
+     * execution context while letting the caller override the child session and user identity.
+     *
+     * <p>The call-scoped {@link AgentState} is intentionally not copied; child agents install
+     * their own state when their call begins.
+     */
+    public RuntimeContext fork(String sessionId, String userId) {
+        return new RuntimeContext(
+                sessionId != null ? sessionId : this.sessionId,
+                userId != null ? userId : this.userId,
+                this.stringAttributes,
+                this.typedAttributes,
+                this.toolExecutionContext);
     }
 
     /**
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
@@ -66,13 +66,20 @@ public class RuntimeContext {
         if (builder.stringExtras != null) {
             this.stringAttributes.putAll(builder.stringExtras);
         }
-        for (Map.Entry<Class<?>, Object> e : builder.typedSingletons.entrySet()) {
-            if (e.getValue() == null) {
+        for (Map.Entry<Class<?>, Map<String, Object>> e : builder.typedValues.entrySet()) {
+            Class<?> type = e.getKey();
+            Map<String, Object> values = e.getValue();
+            if (type == null || values == null || values.isEmpty()) {
                 continue;
             }
-            @SuppressWarnings("unchecked")
-            Class<Object> type = (Class<Object>) e.getKey();
-            putValue(TYPED_DEFAULT_KEY, type, e.getValue());
+            for (Map.Entry<String, Object> typedEntry : values.entrySet()) {
+                if (typedEntry.getKey() == null || typedEntry.getValue() == null) {
+                    continue;
+                }
+                @SuppressWarnings("unchecked")
+                Class<Object> typedClass = (Class<Object>) type;
+                putValue(typedEntry.getKey(), typedClass, typedEntry.getValue());
+            }
         }
     }
 
@@ -210,7 +217,7 @@ public class RuntimeContext {
         if (type == RuntimeContext.class) {
             return (T) this;
         }
-        return null;
+        return getAssignableValue(TYPED_DEFAULT_KEY, type);
     }
 
     public <T> void put(Class<T> type, T value) {
@@ -236,6 +243,10 @@ public class RuntimeContext {
         if (TYPED_DEFAULT_KEY.equals(key) && type == RuntimeContext.class) {
             return (T) this;
         }
+        T assignable = getAssignableValue(key, type);
+        if (assignable != null) {
+            return assignable;
+        }
         Object fromString = stringAttributes.get(key);
         if (type.isInstance(fromString)) {
             return (T) fromString;
@@ -252,6 +263,10 @@ public class RuntimeContext {
         } else {
             putValue(key, type, value);
         }
+    }
+
+    public static Builder builder(RuntimeContext source) {
+        return new Builder().from(source);
     }
 
     /**
@@ -277,6 +292,42 @@ public class RuntimeContext {
             return null;
         }
         return type.isInstance(o) ? type.cast(o) : null;
+    }
+
+    private <T> T getAssignableValue(String key, Class<T> type) {
+        if (type == null) {
+            return null;
+        }
+        T candidate = null;
+        Class<?> candidateType = null;
+        for (Map.Entry<Class<?>, ConcurrentMap<String, Object>> entry :
+                typedAttributes.entrySet()) {
+            Class<?> storedType = entry.getKey();
+            if (storedType == null || !type.isAssignableFrom(storedType)) {
+                continue;
+            }
+            Map<String, Object> values = entry.getValue();
+            if (values == null) {
+                continue;
+            }
+            Object o = values.get(key);
+            if (o == null || !type.isInstance(o)) {
+                continue;
+            }
+            if (candidate == null) {
+                candidate = type.cast(o);
+                candidateType = storedType;
+                continue;
+            }
+            if (candidateType.equals(storedType)) {
+                continue;
+            }
+            if (candidateType.isAssignableFrom(storedType)) {
+                candidate = type.cast(o);
+                candidateType = storedType;
+            }
+        }
+        return candidate;
     }
 
     private <T> void removeTyped(Class<T> type, String key) {
@@ -318,7 +369,7 @@ public class RuntimeContext {
         private String sessionId;
         private String userId;
         private Map<String, Object> stringExtras;
-        private final Map<Class<?>, Object> typedSingletons = new HashMap<>();
+        private final Map<Class<?>, Map<String, Object>> typedValues = new HashMap<>();
         private ToolExecutionContext toolExecutionContext;
         private AgentState agentState;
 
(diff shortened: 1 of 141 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/subagent/DefaultAgentManager.java
```diff
@@ -230,6 +262,16 @@ public WorkspaceManager getWorkspaceManager() {
         return workspaceManager;
     }
 
+    private static RuntimeContext childContext(
+            RuntimeContext parentContext, String sessionId, String userId) {
+        String effectiveUserId =
+                userId != null ? userId : parentContext != null ? parentContext.getUserId() : null;
```
AgentScopeJavaBot: [minor] The `effectiveUserId` null-coalescing here duplicates the fallback already inside `fork()`. When `parentContext != null`, calling `parentContext.fork(sessionId, userId)` directly would produce the same result — `fork()` already falls back to `this.userId` when `userId` is null. The `effectiveUserId` variable is only needed for the `else` branch (builder path). Consider simplifying:
```java
private static RuntimeContext childContext(
        RuntimeContext parentContext, String sessionId, String userId) {
    if (parentContext != null) {
        return parentContext.fork(sessionId, userId);
    }
    String effectiveUserId =
            userId != null ? userId : null; // no parent to fall back to
    return RuntimeContext.builder().sessionId(sessionId).userId(effectiveUserId).build();
}
```
Not a bug — just redundant logic that could confuse future readers about which layer owns the fallback contract.

agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
```diff
@@ -108,6 +135,22 @@ public void setAgentState(AgentState agentState) {
...
     }
 
+    /**
+     * Creates a child context that preserves this context's extras, typed attributes, and tool
+     * execution context while letting the caller override the child session and user identity.
+     *
+     * <p>The call-scoped {@link AgentState} is intentionally not copied; child agents install
+     * their own state when their call begins.
+     */
+    public RuntimeContext fork(String sessionId, String userId) {
```
AgentScopeJavaBot, line 152: [praise] Clean copy-on-fork semantics: the private constructor creates a fresh `ConcurrentHashMap` for `stringAttributes` and deep-copies each inner typed-attribute map via `new ConcurrentHashMap<>(values)`. This ensures parent and child attribute bags are fully isolated — mutations in one don't affect the other. `agentState` is correctly set to `null` (documented in Javadoc), and `ToolExecutionContext` is shared by reference (appropriate for an effectively-immutable DI container). Well-designed.

agentscope-harness/src/test/java/io/agentscope/harness/agent/subagent/DefaultAgentManagerRuntimeContextTest.java
```diff
@@ -99,4 +120,89 @@ void createAgentIfPresent_nullRuntimeContext_substitutesEmpty() {
         // We don't assert .equals() here — empty() may return a fresh instance — only that the
         // factory never sees null.
     }
+
+    @Test
+    void invokeAgent_preservesParentRuntimeContextMetadata() {
```
AgentScopeJavaBot, line 125: [praise] Good test coverage: verifies the full contract of `fork()` through both `invokeAgent` and `invokeAgentStream` paths — sessionId isolation, userId propagation, string extras (`traceId`), typed key-value (`outboundAddress`), typed singleton (`TypedMarker`), `ToolExecutionContext` identity (`assertSame`), and `agentState` nullity. The `parentContext()` helper provides a realistic, multi-layered parent context. Solid regression test.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR adds `RuntimeContext.fork()` to preserve parent context (string extras, typed attributes, and `ToolExecutionContext`) when delegating to harness subagents, and threads the parent `RuntimeContext` through `DefaultAgentManager` and `AgentSpawnTool` invocation paths. The design is clean: `fork()` creates **isolated snapshot copies** of attribute maps (copy-on-fork semantics), preventing parent-child interference, while sharing the `ToolExecutionContext` by reference (appropriate for a DI container). The old overloads delegate to new ones with `null` parentContext, maintaining backwards compatibility. Test coverage is solid, verifying both sync and stream paths with typed attributes, string extras, `ToolExecutionContext` identity, and `agentState` isolation. One pre-existing gap worth noting: the plain `Agent` fallback in `invokeAgent`/`invokeAgentStream` still calls `agent.call(prompt)` without the computed `ctx` — not introduced by this PR, but now more visible.

agentscope-harness/src/main/java/io/agentscope/harness/agent/subagent/DefaultAgentManager.java
AgentScopeJavaBot: [minor] The `effectiveUserId` null-coalescing here duplicates the fallback already inside `fork()`. When `parentContext != null`, calling `parentContext.fork(sessionId, userId)` directly would produce the same result — `fork()` already falls back to `this.userId` when `userId` is null. The `effectiveUserId` variable is only needed for the `else` branch (builder path). Consider simplifying:
```java
private static RuntimeContext childContext(
        RuntimeContext parentContext, String sessionId, String userId) {
    if (parentContext != null) {
        return parentContext.fork(sessionId, userId);
    }
    String effectiveUserId =
            userId != null ? userId : null; // no parent to fall back to
    return RuntimeContext.builder().sessionId(sessionId).userId(effectiveUserId).build();
}
```
Not a bug — just redundant logic that could confuse future readers about which layer owns the fallback contract.

agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
AgentScopeJavaBot, line 152: [praise] Clean copy-on-fork semantics: the private constructor creates a fresh `ConcurrentHashMap` for `stringAttributes` and deep-copies each inner typed-attribute map via `new ConcurrentHashMap<>(values)`. This ensures parent and child attribute bags are fully isolated — mutations in one don't affect the other. `agentState` is correctly set to `null` (documented in Javadoc), and `ToolExecutionContext` is shared by reference (appropriate for an effectively-immutable DI container). Well-designed.

agentscope-harness/src/test/java/io/agentscope/harness/agent/subagent/DefaultAgentManagerRuntimeContextTest.java
AgentScopeJavaBot, line 125: [praise] Good test coverage: verifies the full contract of `fork()` through both `invokeAgent` and `invokeAgentStream` paths — sessionId isolation, userId propagation, string extras (`traceId`), typed key-value (`outboundAddress`), typed singleton (`TypedMarker`), `ToolExecutionContext` identity (`assertSame`), and `agentState` nullity. The `parentContext()` helper provides a realistic, multi-layered parent context. Solid regression test.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR adds `RuntimeContext.fork()` to preserve parent context (string extras, typed attributes, and `ToolExecutionContext`) when delegating to harness subagents, and threads the parent `RuntimeContext` through `DefaultAgentManager` and `AgentSpawnTool` invocation paths. The design is clean: `fork()` creates **isolated snapshot copies** of attribute maps (copy-on-fork semantics), preventing parent-child interference, while sharing the `ToolExecutionContext` by reference (appropriate for a DI container). The old overloads delegate to new ones with `null` parentContext, maintaining backwards compatibility. Test coverage is solid, verifying both sync and stream paths with typed attributes, string extras, `ToolExecutionContext` identity, and `agentState` isolation. One pre-existing gap worth noting: the plain `Agent` fallback in `invokeAgent`/`invokeAgentStream` still calls `agent.call(prompt)` without the computed `ctx` — not introduced by this PR, but now more visible.

agentscope-harness/src/main/java/io/agentscope/harness/agent/subagent/DefaultAgentManager.java
guslegend0510: “收到，按你的建议我已经把 childContext 再简化了：有 parentContext 时直接走 parentContext.fork(sessionId, userId)，没有 parentContext 时直接构建最小上下文，不再额外计算 effectiveUserId。这样逻辑更集中，也更容易读。”

agentscope-harness/src/test/java/io/agentscope/harness/agent/subagent/DefaultAgentManagerRuntimeContextTest.java
guslegend0510: “收到，按你的建议我已经把 childContext 再简化了：有 parentContext 时直接走 parentContext.fork(sessionId, userId)，没有 parentContext 时直接构建最小上下文，不再额外计算 effectiveUserId。这样逻辑更集中，也更容易读。”

agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
guslegend0510: “收到，按你的建议我已经把 childContext 再简化了：有 parentContext 时直接走 parentContext.fork(sessionId, userId)，没有 parentContext 时直接构建最小上下文，不再额外计算 effectiveUserId。这样逻辑更集中，也更容易读。”

agentscope-harness/src/main/java/io/agentscope/harness/agent/subagent/DefaultAgentManager.java
guslegend0510: “收到，按你的建议我已经把 childContext 再简化了：有 parentContext 时直接走 parentContext.fork(sessionId, userId)，没有 parentContext 时直接构建最小上下文，不再额外计算 effectiveUserId。这样逻辑更集中，也更容易读。”

iuin8:
We're building an enterprise agent-governance platform on AgentScope-Java RC3, and this PR fixes a blocker for us: harness subagent delegation drops the parent `RuntimeContext`, so our typed governance identity (tenant + granted capabilities) never reaches spawned subagents — the whole permission/audit arc silently no-ops at the subagent layer.

We reviewed the patch against RC3 and it looks complete on the delegation surface: `RuntimeContext.fork(sessionId, userId)` deep-copies string + typed attributes and shares `ToolExecutionContext`, and `parentCtx` is now threaded into all `AgentSpawnTool` call sites (spawn/send + the three `execLocalSync` paths + `execSpawnTask`). 👍

Two notes that might help it land:
- CI is currently red (windows build failure, ubuntu cancelled) — looks infra/unrelated to the diff; happy to help test if useful.
- `HarnessGateway.runSubagentStream` (the `expose_to_user` direct-dialog path, tracked in #1700) isn't touched here, so that path still loses context — worth a follow-up rather than expanding this PR.

Looking forward to seeing this in a release.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
```diff
@@ -125,7 +125,7 @@ public Flux<AgentEvent> onAgent(
             AgentInput input,
             Function<AgentInput, Flux<AgentEvent>> next) {
         final RuntimeContext rc = ctx != null ? ctx : RuntimeContext.empty();
-        return next.apply(input).doOnComplete(() -> doFlush(agent, rc).subscribe());
+        return next.apply(input).concatWith(doFlush(agent, rc).thenMany(Flux.<AgentEvent>empty()));
```
chickenlj: What's the purpose of this modification? To make it synchronized and wait until it succeeds?

chickenlj (changes requested):

agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
```diff
@@ -108,6 +135,22 @@ public void setAgentState(AgentState agentState) {
...
+        return new RuntimeContext(
+                sessionId != null ? sessionId : this.sessionId,
+                userId != null ? userId : this.userId,
+                this.stringAttributes,
+                this.typedAttributes,
+                this.toolExecutionContext);
+    }
```
chickenlj, line 159: Do you think we need deep copy for fields like `stringAttributes`?

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
guslegend0510: Yes, that is the intent. We want the flush to stay inside the reactive chain so onAgent does not complete until doFlush(...) finishes. The previous doOnComplete(() -> doFlush(...).subscribe()) was fire-and-forget, so it could race with downstream completion and also drop flush errors. concatWith(doFlush(...).thenMany(Flux.empty())) keeps the same external event output, but makes the flush part wait properly.

agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
guslegend0510: I don’t think we need a deep copy for that. Here we only need to copy the map container so parent and child contexts don’t share mutations. The values inside stringAttributes are arbitrary objects, and the runtime layer can’t safely deep-clone them in a generic way. So the goal is isolation of the attribute bag itself, not cloning every object stored in it. If a caller puts mutable objects there and needs full isolation, that object should be copied by the caller before storing it.

guslegend0510:
@chickenlj  回复你了，麻烦看一下，谢谢您

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
chickenlj: I see, thanks for the explanation. Please revert the change on this line and submit a new pull request or file a new issue so we can further discuss it.

guslegend0510: Thanks for the feedback. I’ve reverted that change and kept MemoryFlushMiddleware on the original completion path for now. This PR will stay focused on the runtime-context propagation fixes only. If we want to make flush completion part of the reactive chain, I agree that should be handled in a separate PR.
@chickenlj

guslegend0510:
@chickenlj 麻烦您审核一下，谢谢

chickenlj:
Thanks for the contribution! This has been fixed by #1833 (propagate parent RuntimeContext to child agents).

PR #1757, opened 2026-06-14
fix(harness): treat '.' as root in CompositeFilesystem route merging

### LINKED ISSUE #1737: [Bug]:CompositeFilesystem.ls() 根目录不显示路由条目 — WorkspacePathNormalizer.tryStrip() 返回 "." 而非 "/"
### 使用版本
  AgentScope Java SDK 2.0.0-RC3

### 问题描述

  Agent 在工作空间根目录调用 `list_files` 时，`CompositeFilesystem.ls()` 的返回结果中不包含路由目录（如 `skills/`、`memory/`、`knowledge/` 等），直接返回
  `Empty or not a directory`，Agent 完全无法浏览工作空间。

  ### 根因

  `WorkspacePathNormalizer.tryStrip()` 在输入路径等于工作空间根目录时返回 `"."`，但 `CompositeFilesystem.ls()` 仅在路径严格等于 `"/"` 时才合并路由条目：

  ```java
  // CompositeFilesystem.ls() — 约第 176 行
  if ("/".equals(path)) {
      // 合并路由目录条目
  }
```

  由于 "." ≠ "/"，路由合并分支被完全跳过，直接走到 defaultBackend.ls(rc, ".")，而本地 workspace 目录为空时返回 Empty or not a directory。

  复现步骤

  1. 配置 RemoteFilesystemSpec 作为 filesystem（带 OSS 远程存储）
  2. Agent 调用 list_files，路径为工作空间根目录
  3. 观察返回结果

  实际行为

  "Empty or not a directory: /path/to/.agentscope/workspace"

  Agent 多次重试不同路径层级，全部失败，完全无法发现工作空间内容。

  期望行为

  返回所有已注册的路由目录和本地文件：

  [FILE] MEMORY.md (0 bytes)
  [DIR]  agents/
  [DIR]  knowledge/
  [DIR]  memory/
  [DIR]  plans/
  [DIR]  skills/
  [DIR]  subagents/

  建议修复

  二选一：
  1. WorkspacePathNormalizer.tryStrip() 对工作空间根目录返回 "/" 而非 "."
  2. CompositeFilesystem.ls() 在判断是否合并路由时，将 "." 视为等价于 "/"

当前 Workaround

  在 CompositeFilesystem 外包一层委托 Filesystem，在 ls() / grep() / glob() 调用前将 "." 归一化为 "/"：
  ```java
  public class RootNormalizingFilesystem implements AbstractFilesystem {
      private final AbstractFilesystem delegate;

      private static String normRoot(String path) {
          return ".".equals(path) ? "/" : path;
      }

      @Override
      public LsResult ls(RuntimeContext rc, String path) {
          return delegate.ls(rc, normRoot(path));
      }

      @Override
      public GrepResult grep(RuntimeContext rc, String pattern, String path, String glob) {
          return delegate.grep(rc, pattern, normRoot(path), glob);
      }

      @Override
      public GlobResult glob(RuntimeContext rc, String pattern, String path) {
          return delegate.glob(rc, pattern, normRoot(path));
      }

      // 其余方法直接委托
  }
  ```

  通过继承 RemoteFilesystemSpec 重写 toFilesystem() 注入：
  ```java
  public class PantheonRemoteFilesystemSpec extends RemoteFilesystemSpec {
      @Override
      public AbstractFilesystem toFilesystem(
              Path workspace, String agentId, NamespaceFactory localNamespaceFactory) {
          AbstractFilesystem composite = super.toFilesystem(workspace, agentId, localNamespaceFactory);
          return new RootNormalizingFilesystem(composite);
      }
  }
  ```

### PR DESCRIPTION
**Pull request by Sparkle6979**
## Problem

`WorkspacePathNormalizer` returns `"."` when the path equals the workspace prefix. In sandbox/remote mode where the workspace is `/workspace`, an LLM calling `list_files("/workspace")` results in `CompositeFilesystem.ls(".")`. However, `ls()`, `grep()`, and `glob()` only recognized `"/"` as root — `"."` was treated as a regular path and route directories (skills/, memory/, knowledge/, etc.) were silently excluded from listings.

Fixes #1737

## Changes

`CompositeFilesystem.java`:

- `ls()`: add `\|\| ".".equals(path)` to the root check
- `grep()`: add `\|\| ".".equals(path)` to the root check  
- `glob()`: add `&& !".".equals(path)` to the non-root early-return guard

`CompositeFilesystemTest.java`:

- Add `lsRoot_mergesRouteEntriesWhenPathIsDot` — verifies route entries appear when listing root with `"."`

### DIFF
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/CompositeFilesystemTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/CompositeFilesystemTest.java
@@ -20,6 +20,7 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyList;
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
@@ -28,6 +29,7 @@ import io.agentscope.harness.agent.filesystem.local.LocalFilesystem;
 import io.agentscope.harness.agent.filesystem.model.FileInfo;
 import io.agentscope.harness.agent.filesystem.model.FileUploadResponse;
 import io.agentscope.harness.agent.filesystem.model.GlobResult;
+import io.agentscope.harness.agent.filesystem.model.LsResult;
 import io.agentscope.harness.agent.filesystem.model.ReadResult;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -213,5 +215,30 @@ class CompositeFilesystemTest {
         assertTrue(jsonPaths.contains("tools.json"));
         assertFalse(
                 jsonPaths.contains("AGENTS.md"), "AGENTS.md must not match *.json: " + jsonPaths);
+    }
+
+    /**
+     * Regression for #1737: when path is "." (which WorkspacePathNormalizer returns for
+     * workspace root), route directories should still be merged into the listing.
+     */
+    @Test
+    void lsRoot_mergesRouteEntriesWhenPathIsDot() {
+        AbstractFilesystem routeBackend = mock(AbstractFilesystem.class);
+        when(routeBackend.ls(any(), anyString()))
+                .thenReturn(LsResult.success(List.of(FileInfo.ofFile("/SKILL.md", 10, ""))));
+
+        AbstractFilesystem defaultBackend = mock(AbstractFilesystem.class);
+        when(defaultBackend.ls(any(), anyString())).thenReturn(LsResult.success(List.of()));
+
+        CompositeFilesystem composite =
+                new CompositeFilesystem(defaultBackend, Map.of("/skills/", routeBackend));
+
+        LsResult result = composite.ls(CTX, ".");
+
+        assertTrue(result.isSuccess());
+        List<String> paths = result.entries().stream().map(FileInfo::path).toList();
+        assertTrue(
+                paths.stream().anyMatch(p -> p.startsWith("/skills")),
+                "route entries should appear when listing root with '.', got: " + paths);
     }
 }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/CompositeFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/CompositeFilesystem.java
@@ -188,7 +188,7 @@ public class CompositeFilesystem implements AbstractFilesystem {
             return LsResult.success(remapped);
         }
 
-        if ("/".equals(path)) {
+        if ("/".equals(path) || ".".equals(path)) {
             List<FileInfo> results = new ArrayList<>();
             LsResult defaultResult = defaultBackend.ls(runtimeContext, path);
             if (defaultResult.isSuccess() && defaultResult.entries() != null) {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/CompositeFilesystemTest.java [hunk 1]
```diff
@@ -214,4 +216,29 @@ void globRecursiveRoot_includesExactFileRoutes() throws Exception {
         assertFalse(
                 jsonPaths.contains("AGENTS.md"), "AGENTS.md must not match *.json: " + jsonPaths);
     }
+
+    /**
+     * Regression for #1737: when path is "." (which WorkspacePathNormalizer returns for
+     * workspace root), route directories should still be merged into the listing.
+     */
+    @Test
```
AgentScopeJavaBot, line 224: [nit] The regression test only covers `ls()` with `"."`. Consider adding analogous tests for `grep()` and `glob()` with the same path to guard against future regressions in those methods too. Something like:

```java
@Test
void grepRoot_searchesRoutesWhenPathIsDot() { ... }

@Test
void globRoot_mergesRoutesWhenPathIsDot() { ... }
```

Not blocking — the fix is clearly correct and the `ls` test validates the core pattern.

AgentScopeJavaBot (approved):
🤖 AI Review

This is a clean, focused bugfix for issue #1737. `WorkspacePathNormalizer.tryStrip()` returns `"."` when the input path exactly equals the workspace prefix (e.g., `/workspace`), but `CompositeFilesystem` only recognized `"/"` as the root path in `ls()`, `grep()`, and `glob()`. This caused route directories (skills/, memory/, knowledge/, etc.) to be silently excluded from listings when an LLM called these operations on the workspace root in sandbox/remote mode.

The fix correctly extends the root-path check in all three methods to also accept `"."`. The change is minimal (+3 lines in production code), null-safe (`".".equals(path)` won't NPE), and symmetric across all three affected methods. The routing logic (`routeForPath`) correctly passes `"."` through to the default backend since `"."` won't match any route prefix, so the root-merging branch is always reached. A regression test for `ls()` is included.

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/CompositeFilesystemTest.java
AgentScopeJavaBot, line 224: [nit] The regression test only covers `ls()` with `"."`. Consider adding analogous tests for `grep()` and `glob()` with the same path to guard against future regressions in those methods too. Something like:

```java
@Test
void grepRoot_searchesRoutesWhenPathIsDot() { ... }

@Test
void globRoot_mergesRoutesWhenPathIsDot() { ... }
```

Not blocking — the fix is clearly correct and the `ls` test validates the core pattern.

AgentScopeJavaBot (approved):
🤖 AI Review

This is a clean, focused bugfix for issue #1737. `WorkspacePathNormalizer.tryStrip()` returns `"."` when the input path exactly equals the workspace prefix (e.g., `/workspace`), but `CompositeFilesystem` only recognized `"/"` as the root path in `ls()`, `grep()`, and `glob()`. This caused route directories (skills/, memory/, knowledge/, etc.) to be silently excluded from listings when an LLM called these operations on the workspace root in sandbox/remote mode.

The fix correctly extends the root-path check in all three methods to also accept `"."`. The change is minimal (+3 lines in production code), null-safe (`".".equals(path)` won't NPE), and symmetric across all three affected methods. The routing logic (`routeForPath`) correctly passes `"."` through to the default backend since `"."` won't match any route prefix, so the root-merging branch is always reached. A regression test for `ls()` is included.

chickenlj:
Thanks for the contribution! This has been fixed by #1830 (CompositeFilesystem treats '.' as root equivalent).

PR #1771, opened 2026-06-16
fix(skill-runtime): use skill name instead of composite ID for SkillFilter matching …issues/1768

xseruer:
…issues/1768

## AgentScope-Java Version

2.0.0-SNAPSHOT

### Background

`SkillFilter.only("host-forensics-client")` does not work as expected when used with `HarnessAgent`. The whitelisted skill is silently filtered out, and the rendered skill prompt is empty.

### Root Cause

In `SkillPromptBuilder#render(SkillCatalog, SkillFilter)`, the code passes `entry.skill().getSkillId()` to `SkillFilter#isAllowed()`. However, `getSkillId()` returns a composite ID in the format `name_source` (e.g. `host-forensics-client_filesystem-agentscope_skills`), while users naturally pass bare skill names (e.g. `host-forensics-client`) to `SkillFilter.only(...)`. Since `isAllowed()` performs an exact `contains()` check, the match always fails, making WHITELIST / BLACKLIST / OVERLAY_ENABLE / OVERLAY_DISABLE modes effectively unusable.

### Fix

Replace `entry.skill().getSkillId()` with `entry.skill().getName()` in `SkillPromptBuilder#render()` so the filter key matches the bare skill name that users configure.

**Changed files:**
- `SkillPromptBuilder.java` — 1-line fix: `getSkillId()` → `getName()`
- `SkillRuntimeTest.java` — updated existing test to use bare name; added new test `filterWithBareNameMatchesCompositeId` that reproduces the reported scenario

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillPromptBuilder.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillPromptBuilder.java
@@ -128,8 +128,8 @@ public final class SkillPromptBuilder {
         boolean anyWithFilesRoot = false;
 
         for (HarnessSkillEntry entry : catalog.all()) {
-            String id = entry.skill().getSkillId();
-            if (!effective.isAllowed(id)) {
+            String skillName = entry.skill().getName();
+            if (!effective.isAllowed(skillName)) {
                 continue;
             }
             if (!any) {
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/skill/runtime/SkillRuntimeTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/skill/runtime/SkillRuntimeTest.java
@@ -114,7 +114,7 @@ class SkillRuntimeTest {
             HarnessSkillEntry visible = HarnessSkillEntry.of(skill("visible", "src"), null);
             HarnessSkillEntry hidden = HarnessSkillEntry.of(skill("hidden", "src"), null);
             SkillCatalog cat = SkillCatalog.of(List.of(visible, hidden));
-            SkillFilter only = SkillFilter.only("visible_src");
+            SkillFilter only = SkillFilter.only("visible");
 
             String out = new SkillPromptBuilder().render(cat, only);
             assertTrue(out.contains("<skill-id>visible_src</skill-id>"));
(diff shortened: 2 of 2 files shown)


AgentScopeJavaBot (approved):
🤖 AI Review

This is a clean, focused bug fix that corrects the `SkillFilter` matching logic in `SkillPromptBuilder.render()`. The root cause is accurately identified: `getSkillId()` returns a composite `name_source` string, while `SkillFilter` was designed to accept bare skill names (as confirmed by its field name `skillNames`, the Javadoc example `SkillFilter.only("skill_a", "skill_b")`, and the `HarnessAgent.Builder.enableSkills(String... skillNames)` API). The 1-line fix (`getSkillId()` → `getName()`) is correct, minimal, and consistent with the intended contract. The variable rename from `id` to `skillName` improves readability. Test coverage is solid — the existing test is updated to match the correct behavior, and the new test `filterWithBareNameMatchesCompositeId` directly reproduces the reported scenario with realistic data. The rendered `<skill-id>` in the prompt still uses the composite ID (for `load_skill_through_path`), so there's no downstream breakage. All filter modes (WHITELIST, BLACKLIST, OVERLAY_ENABLE, OVERLAY_DISABLE) benefit from this single-point fix.

AgentScopeJavaBot (approved):
🤖 AI Review

This is a clean, focused bug fix that corrects the `SkillFilter` matching logic in `SkillPromptBuilder.render()`. The root cause is accurately identified: `getSkillId()` returns a composite `name_source` string, while `SkillFilter` was designed to accept bare skill names (as confirmed by its field name `skillNames`, the Javadoc example `SkillFilter.only("skill_a", "skill_b")`, and the `HarnessAgent.Builder.enableSkills(String... skillNames)` API). The 1-line fix (`getSkillId()` → `getName()`) is correct, minimal, and consistent with the intended contract. The variable rename from `id` to `skillName` improves readability. Test coverage is solid — the existing test is updated to match the correct behavior, and the new test `filterWithBareNameMatchesCompositeId` directly reproduces the reported scenario with realistic data. The rendered `<skill-id>` in the prompt still uses the composite ID (for `load_skill_through_path`), so there's no downstream breakage. All filter modes (WHITELIST, BLACKLIST, OVERLAY_ENABLE, OVERLAY_DISABLE) benefit from this single-point fix.

chickenlj (approved):
LGTM.

PR #1773, opened 2026-06-16
fix issue #1738 problem 1

xuhuafeifei:
## AgentScope-Java Version

[INFO] io.agentscope:agentscope-parent:pom:2.0.0-SNAPSHOT

## Description

### Background 

Agent calls skill_manage create to create a skill, then immediately calls skill_manage write_file in the same reasoning turn to write resource files, resulting in "Skill not found" error

### Solution

Optimize the usage pattern to avoid "create-then-immediately-write" multiple operations in the same reasoning turn:                                                             
                                                                                                                                                                                  
Add resources parameter to SkillManageTool.create for atomic writing:                                                                                                           
  - SKILL.md + all resource files written in a single operation                                                                                                                 
  - Eliminates need for subsequent write_file calls, fundamentally avoiding consistency issues                                                                                    
  - Underlying WorkspaceSkillRepository.save() already supports multi-file atomic writing, just wasn't exposed in the tool API  


## Checklist

- [*]  Code has been formatted with `mvn spotless:apply`
- [*]  All tests are passing (`mvn test`)
- [*]  Javadoc comments are complete and follow project conventions
- [*]  Related documentation has been updated (e.g. links, examples, etc.)
- [*]  Code is ready for review

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
@@ -759,16 +759,15 @@ public class SkillManageTool implements AgentTool {
 
     private static boolean boolOf(Map<String, Object> m, String key) {
         Object v = m.get(key);
-        if (v instanceof Boolean) {
-            return (Boolean) v;
+        if (v instanceof Boolean b) {
+            return b;
         }
-        if (v instanceof String) {
-            return Boolean.parseBoolean((String) v);
+        if (v instanceof String s) {
+            return Boolean.parseBoolean(s);
         }
         return false;
     }
 
-    @SuppressWarnings("unchecked")
     private static Map<String, String> mapOfStrings(Map<String, Object> m, String key) {
         Object v = m.get(key);
         if (v instanceof Map) {
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
@@ -175,6 +175,17 @@ public class SkillManageTool implements AgentTool {
                         "Full SKILL.md content (frontmatter + body). Required for "
                                 + "create/edit."));
         properties.put(
+                "resources",
+                Map.of(
+                        "type",
+                        "object",
+                        "description",
+                        "Optional for create: supporting files to include with the skill. "
+                                + "Keys are paths relative to skill root (must start with "
+                                + "references/, templates/, scripts/, or assets/), "
+                                + "values are file contents. All files written atomically "
+                                + "with SKILL.md."));
+        properties.put(
                 "old_string",
                 Map.of(
                         "type", "string",
@@ -255,7 +266,11 @@ public class SkillManageTool implements AgentTool {
         }
         switch (action) {
             case "create":
-                return doCreate(name, stringOf(input, "content"), sessionIdOf(ctx));
+                return doCreate(
+                        name,
+                        stringOf(input, "content"),
+                        mapOfStrings(input, "resources"),
+                        sessionIdOf(ctx));
             case "edit":
                 return doEdit(name, stringOf(input, "content"));
             case "patch":
@@ -292,7 +307,8 @@ public class SkillManageTool implements AgentTool {
     //  Actions
     // ---------------------------------------------------------------------
 
-    private ToolResultBlock doCreate(String name, String content, String sessionId) {
+    private ToolResultBlock doCreate(
+            String name, String content, Map<String, String> resources, String sessionId) {
         if (content == null || content.isBlank()) {
             return ToolResultBlock.error(
                     "Missing 'content' parameter (full SKILL.md including frontmatter).");
@@ -301,6 +317,14 @@ public class SkillManageTool implements AgentTool {
         if (contentErr != null) {
             return ToolResultBlock.error(contentErr);
         }
+        if (resources != null) {
+            for (String path : resources.keySet()) {
+                String err = validateSubFilePath(path);
+                if (err != null) {
+                    return ToolResultBlock.error("Invalid resource path '" + path + "': " + err);
+                }
+            }
+        }
         // Reject if a skill with this name already exists in either repo.
         if (mainRepo.skillExists(name) || draftsRepo.skillExists(name)) {
             return ToolResultBlock.error(
@@ -308,7 +332,7 @@ public class SkillManageTool implements AgentTool {
         }
         AgentSkill skill;
         try {
-            skill = SkillUtil.createFrom(content, null, "agent");
+            skill = SkillUtil.createFrom(content, resources, "agent");
         } catch (Exception e) {
             return ToolResultBlock.error("Failed to parse frontmatter: " + e.getMessage());
         }
@@ -735,12 +759,27 @@ public class SkillManageTool implements AgentTool {
 
     private static boolean boolOf(Map<String, Object> m, String key) {
         Object v = m.get(key);
-        if (v instanceof Boolean b) {
-            return b;
+        if (v instanceof Boolean) {
+            return (Boolean) v;
         }
-        if (v instanceof String s) {
-            return Boolean.parseBoolean(s);
+        if (v instanceof String) {
+            return Boolean.parseBoolean((String) v);
         }
         return false;
+    }
+
+    @SuppressWarnings("unchecked")
+    private static Map<String, String> mapOfStrings(Map<String, Object> m, String key) {
+        Object v = m.get(key);
+        if (v instanceof Map) {
+            Map<String, String> result = new HashMap<>();
+            for (Map.Entry<?, ?> entry : ((Map<?, ?>) v).entrySet()) {
+                if (entry.getKey() instanceof String && entry.getValue() instanceof String) {
+                    result.put((String) entry.getKey(), (String) entry.getValue());
+                }
+            }
+            return result.isEmpty() ? null : result;
+        }
+        return null;
     }
 }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
@@ -179,6 +179,8 @@ public class SkillManageTool implements AgentTool {
                 Map.of(
                         "type",
                         "object",
+                        "additionalProperties",
+                        Map.of("type", "string"),
                         "description",
                         "Optional for create: supporting files to include with the skill. "
                                 + "Keys are paths relative to skill root (must start with "
@@ -775,6 +777,8 @@ public class SkillManageTool implements AgentTool {
             for (Map.Entry<?, ?> entry : ((Map<?, ?>) v).entrySet()) {
                 if (entry.getKey() instanceof String && entry.getValue() instanceof String) {
                     result.put((String) entry.getKey(), (String) entry.getValue());
+                } else {
+                    log.warn("mapOfStrings: skipping non-String entry key={}", entry.getKey());
                 }
             }
             return result.isEmpty() ? null : result;
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/SkillManageToolTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/SkillManageToolTest.java
@@ -701,6 +701,115 @@ class SkillManageToolTest {
         assertTrue(text(r).contains("pruning"));
     }
 
+    // ---- code review feedback: schema, mapOfStrings logging, boolOf compatibility ----
+
+    @Test
+    @SuppressWarnings("unchecked")
+    void resourcesSchema_includesAdditionalPropertiesTypeHint() {
+        Map<String, Object> params = toolDraftDefault.getParameters();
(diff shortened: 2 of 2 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
```diff
@@ -743,4 +767,18 @@ private static boolean boolOf(Map<String, Object> m, String key) {
...
     }
+
+    private static Map<String, String> mapOfStrings(Map<String, Object> m, String key) {
+        Object v = m.get(key);
+        if (v instanceof Map) {
+            Map<String, String> result = new HashMap<>();
+            for (Map.Entry<?, ?> entry : ((Map<?, ?>) v).entrySet()) {
+                if (entry.getKey() instanceof String && entry.getValue() instanceof String) {
+                    result.put((String) entry.getKey(), (String) entry.getValue());
+                }
```
AgentScopeJavaBot, line 782: [minor] `mapOfStrings` silently drops entries whose key or value is not a `String`. If an LLM passes a non-string value (e.g., `{"scripts/count.sh": 42}`), that resource is omitted without any error or warning — the tool returns success and the agent assumes the file was created. Consider logging a warning when an entry is skipped, so the issue surfaces in diagnostics:
```java
if (entry.getKey() instanceof String && entry.getValue() instanceof String) {
    result.put((String) entry.getKey(), (String) entry.getValue());
} else {
    log.warn("mapOfStrings: skipping non-String entry key={}", entry.getKey());
}
```

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
```diff
@@ -174,6 +174,17 @@ public Map<String, Object> getParameters() {
...
+                "resources",
+                Map.of(
+                        "type",
+                        "object",
+                        "description",
+                        "Optional for create: supporting files to include with the skill. "
+                                + "Keys are paths relative to skill root (must start with "
+                                + "references/, templates/, scripts/, or assets/), "
+                                + "values are file contents. All files written atomically "
+                                + "with SKILL.md."));
```
AgentScopeJavaBot, line 189: [nit] The JSON schema for `resources` declares `"type": "object"` but omits `additionalProperties`. Adding `"additionalProperties", Map.of("type", "string")` would give the LLM a stronger hint that values must be strings, and would enable schema-level validation on the tool-call side. Not blocking — the description already conveys the intent clearly.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR adds a `resources` parameter to the `SkillManageTool.create` action, enabling atomic creation of a skill with all its resource files in a single tool call. This directly addresses the race condition described in issue #1738 where an agent calls `create` then immediately `write_file` in the same reasoning turn and encounters "Skill not found" errors. The implementation is clean — it leverages the existing `SkillUtil.createFrom(content, resources, source)` and `WorkspaceSkillRepository.writeSkill()` infrastructure that already supports multi-file atomic writes. A comprehensive set of new tests (209 lines) covers the happy path, edge cases (null/empty resources, invalid paths, multiple resource types), and draft-mode behavior. The code follows existing patterns consistently.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
AgentScopeJavaBot, line 782: [minor] `mapOfStrings` silently drops entries whose key or value is not a `String`. If an LLM passes a non-string value (e.g., `{"scripts/count.sh": 42}`), that resource is omitted without any error or warning — the tool returns success and the agent assumes the file was created. Consider logging a warning when an entry is skipped, so the issue surfaces in diagnostics:
```java
if (entry.getKey() instanceof String && entry.getValue() instanceof String) {
    result.put((String) entry.getKey(), (String) entry.getValue());
} else {
    log.warn("mapOfStrings: skipping non-String entry key={}", entry.getKey());
}
```

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
AgentScopeJavaBot, line 189: [nit] The JSON schema for `resources` declares `"type": "object"` but omits `additionalProperties`. Adding `"additionalProperties", Map.of("type", "string")` would give the LLM a stronger hint that values must be strings, and would enable schema-level validation on the tool-call side. Not blocking — the description already conveys the intent clearly.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR adds a `resources` parameter to the `SkillManageTool.create` action, enabling atomic creation of a skill with all its resource files in a single tool call. This directly addresses the race condition described in issue #1738 where an agent calls `create` then immediately `write_file` in the same reasoning turn and encounters "Skill not found" errors. The implementation is clean — it leverages the existing `SkillUtil.createFrom(content, resources, source)` and `WorkspaceSkillRepository.writeSkill()` infrastructure that already supports multi-file atomic writes. A comprehensive set of new tests (209 lines) covers the happy path, edge cases (null/empty resources, invalid paths, multiple resource types), and draft-mode behavior. The code follows existing patterns consistently.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
xuhuafeifei: I will fix it

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
```diff
@@ -743,4 +769,20 @@ private static boolean boolOf(Map<String, Object> m, String key) {
...
+        if (v instanceof Map) {
+            Map<String, String> result = new HashMap<>();
+            for (Map.Entry<?, ?> entry : ((Map<?, ?>) v).entrySet()) {
+                if (entry.getKey() instanceof String && entry.getValue() instanceof String) {
+                    result.put((String) entry.getKey(), (String) entry.getValue());
+                } else {
+                    log.warn("mapOfStrings: skipping non-String entry key={}", entry.getKey());
+                }
+            }
+            return result.isEmpty() ? null : result;
```
oss-maintainer, line 784: **[Warning]** Silent data loss on malformed input. `mapOfStrings` drops any entry whose key or value is not a `String` (only a `log.warn` at line 781), so if the model sends `{"scripts/run.sh": 42}` or `{"scripts/run.sh": {"content": "..."}}`, the skill is reported as created **without that resource** — the LLM has no way to notice, and later `load_skill_through_path` will fail with a confusing "file not found".

Since this PR's whole purpose is to let the model express intent in one call, I would fail loudly instead of degrading:
```java
return ToolResultBlock.error(
    "Resource '" + key + "' must be a string (file content), got "
        + (value == null ? "null" : value.getClass().getSimpleName()) + ".");
```
The same reasoning applies to a non-`Map` `resources` value, which today returns `null` (i.e. "not provided").

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
```diff
@@ -174,6 +174,19 @@ public Map<String, Object> getParameters() {
...
                                 + "create/edit."));
+        properties.put(
+                "resources",
+                Map.of(
+                        "type",
+                        "object",
+                        "additionalProperties",
+                        Map.of("type", "string"),
+                        "description",
+                        "Optional for create: supporting files to include with the skill. "
```
oss-maintainer, line 185: **[Warning]** Two hardening gaps in the new free-form input surface:
1. **No size/count cap.** Each value is written verbatim as a file, so one tool call can emit an unbounded number of arbitrarily large files (the `content` parameter has `validateContent`, resources have nothing). A cheap guard in the described style of `MAX_DESCRIPTION_LENGTH` — e.g. max N resources and max M bytes per resource / total — would keep the tool schema honest with the "supporting files" intent.
2. **Atomicity wording vs. implementation.** `WorkspaceSkillRepository.writeSkill` hands SKILL.md plus resources to `AbstractFilesystem#uploadFiles`, which returns **per-path** `FileUploadResponse`s; the response list is discarded on the save path, so a partially written skill still reports success. If "all files written atomically" is the contract you want to advertise, either check the responses and return an error, or soften the description (e.g. "written in one batch operation").

The path validation itself looks right — `validateSubFilePath` rejects `..`, leading `/` and non-allowlisted top dirs, and `writeSkill` re-checks before writing.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
```diff
@@ -255,7 +268,11 @@ private ToolResultBlock dispatchSync(Map<String, Object> input, RuntimeContext c
         }
         switch (action) {
             case "create":
-                return doCreate(name, stringOf(input, "content"), sessionIdOf(ctx));
+                return doCreate(
+                        name,
+                        stringOf(input, "content"),
+                        mapOfStrings(input, "resources"),
```
oss-maintainer, line 274: **[Info]** Minor ordering nit: resource paths are validated inside `doCreate` *before* the duplicate-name check, so a call with both a bad path and an existing skill name reports the path error, the model fixes the path, then gets "already exists". Validating the cheap static constraints first (content, resources) and the state-dependent check (name collision) afterwards is fine either way — just flagging that the error the model sees first is not the one that determines whether the call can ever succeed. A single combined error message would also work.

agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/SkillManageToolTest.java
```diff
@@ -492,6 +701,115 @@ void deleteWithEmptyAbsorbedIntoMeansPruning() {
         assertTrue(text(r).contains("pruning"));
     }
 
+    // ---- code review feedback: schema, mapOfStrings logging, boolOf compatibility ----
```
oss-maintainer, line 704: **[Info]** Test scope is broader than the change. The `boolOf_*` tests and `createWithInvalidFrontmatter_rejectsWithParseError` exercise behaviour already on `main` (`boolOf`, `SkillUtil.createFrom` parsing) with no production code in this PR touching them, and the section comment reads like a resolved-review log ("code review feedback: ...") rather than a scenario name. Please split them out (or drop them) so the PR stays reviewable as "create-with-resources"; also consider adding the negative test for the case flagged above (non-string resource value), which currently locks in the silent-skip behaviour.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SkillManageTool.java
```diff
@@ -301,14 +319,22 @@ private ToolResultBlock doCreate(String name, String content, String sessionId)
         if (contentErr != null) {
             return ToolResultBlock.error(contentErr);
         }
+        if (resources != null) {
+            for (String path : resources.keySet()) {
+                String err = validateSubFilePath(path);
+                if (err != null) {
+                    return ToolResultBlock.error("Invalid resource path '" + path + "': " + err);
```
oss-maintainer, line 326: **[Info]** The other three `validateSubFilePath` call sites (lines 478, 580, 625) return the raw message; here it is wrapped as `"Invalid resource path '<path>': " + err`, which produces `Invalid resource path 'evil.sh': file_path must include a subdirectory ...` — the inner text already says `file_path must ...`, so the message double-states the subject. Suggest returning `err` verbatim (consistent with the neighbours) or rewording to `Invalid resource path '<path>': must be relative, no '..', start with one of ...`.

oss-maintainer (commented):
### Summary
Adds an optional `resources` map to `skill_manage action=create` so an agent can write SKILL.md and its supporting files in one call, instead of the create-then-write_file sequence that hit "Skill not found" (#1738). The direction is right and the tests are thorough — main has no `resources` parameter on `SkillManageTool`, so this is not superseded. Blocking issues before merge: the tool silently drops non-string resource values (the model gets a success with files missing), the new input has no size/count cap, and the "written atomically" claim is not what the storage path guarantees.

Two process notes: the PR branch is ~3 months old and `mergeable_state=blocked` (branch protection requires checks to pass) — please rebase onto current `main`, and rename the title to something like `feat(harness): support resource files in skill_manage create` plus a `Fixes #1738.` line so the issue auto-closes. Also, some new test comments are Chinese while the tool messages are English — worth aligning with the project default if that is intentional, ignore this note.

### Findings
- **[Warning]** SkillManageTool.java:784 — non-string resource values are skipped with only a log line; the tool reports success although the requested file was never written. Fail the call instead.
- **[Warning]** SkillManageTool.java:185 — no size/count cap on resources; "atomic" wording vs `uploadFiles` returning per-path responses that the save path ignores.
- **[Info]** SkillManageTool.java:274 — path errors are reported before the duplicate-name check; order the validations so the decisive error wins.
- **[Info]** SkillManageTool.java:326 — error message double-states the subject relative to the three other `validateSubFilePath` call sites.
- **[Info]** SkillManageToolTest.java:704 — `boolOf_*` / invalid-frontmatter tests cover pre-existing behaviour and should not ride along here.

---

oss-maintainer:

This PR currently conflicts with the target branch and cannot be merged. Please rebase or merge the latest target branch into your branch and resolve the conflicts:

```bash
git fetch origin
git checkout fix_issue_1738
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #1774, opened 2026-06-16
fix(harness): make session JSONL offload idempotent

guslegend0510:
## 现象
在 `HarnessAgent` 使用 `DistributedStore + RemoteFilesystemSpec` 时，同一个 session 多轮请求后，`agents/<agentId>/sessions/<sessionId>.jsonl` 会持续变大，历史消息会被重复写入。

## 复现步骤
1. 创建启用了分布式存储的 `HarnessAgent`
2. 同一 session 连续发送 2 次及以上消息
3. 检查 `contextFile` / `logFile` / DB 镜像内容

## 预期行为
- 已经 offload 过的历史消息不应再次被追加
- `flush()` 后的 session JSONL 应保持幂等，不应随请求次数重复膨胀

## 实际行为
- 第 2 次及后续请求会把历史消息再次写入文件末尾
- 文件和 DB 镜像都会随请求次数增长

## 影响
- session 文件持续膨胀
- 后续 `load()` / `syncFromRemote()` 成本越来越高
- 长时间运行的 agent 容易出现存储和性能问题

## 验收标准
- 连续多轮请求后，session JSONL 不应重复追加已有历史消息
- 回归测试应覆盖重复 flush、sync 去重和 DB 镜像一致性

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
@@ -32,7 +32,9 @@ import java.nio.file.Path;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -244,7 +246,18 @@ public class MemoryFlushManager {
             // (cross-machine handoff) are included in the merged file pushed to remote.
             tree.syncFromRemote();
 
-            String lastId = null;
+            List<SessionEntry> existingEntries = new ArrayList<>(tree.getAllEntries());
+            Set<String> existingIds =
+                    existingEntries.stream()
+                            .map(SessionEntry::getId)
+                            .collect(Collectors.toCollection(HashSet::new));
+            String lastId =
+                    existingEntries.isEmpty()
+                            ? null
+                            : existingEntries.get(existingEntries.size() - 1).getId();
+
+            // The caller passes the full conversation on every turn. Use the stable Msg IDs
+            // to keep the session JSONL append-only and idempotent across repeated offloads.
             for (Msg msg : messages) {
                 if (msg.getRole() == null || isSessionContextMessage(msg)) {
                     continue;
@@ -253,11 +266,16 @@ public class MemoryFlushManager {
                 if (rendered == null || rendered.isBlank()) {
                     continue;
                 }
+                String entryId = normalizeEntryId(msg.getId());
+                if (entryId != null && existingIds.contains(entryId)) {
+                    continue;
+                }
                 String toolCallId = extractToolCallId(msg);
                 SessionEntry.MessageEntry entry =
                         new SessionEntry.MessageEntry(
-                                null, lastId, null, msg.getRole().name(), rendered, toolCallId);
+                                entryId, lastId, null, msg.getRole().name(), rendered, toolCallId);
                 tree.append(entry);
+                existingIds.add(entry.getId());
                 lastId = entry.getId();
             }
 
@@ -282,6 +300,10 @@ public class MemoryFlushManager {
             }
         }
         return null;
+    }
+
+    private static String normalizeEntryId(String id) {
+        return id != null && !id.isBlank() ? id : null;
     }
 
     /**
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
@@ -267,6 +267,11 @@ public class MemoryFlushManager {
                     continue;
                 }
                 String entryId = normalizeEntryId(msg.getId());
+                if (entryId == null) {
+                    log.warn(
+                            "Msg without stable ID encountered (role={}); dedup skipped",
+                            msg.getRole());
+                }
                 if (entryId != null && existingIds.contains(entryId)) {
                     continue;
                 }
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent.memory;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.core.message.Msg;
+import io.agentscope.core.message.MsgRole;
+import io.agentscope.core.message.TextBlock;
+import io.agentscope.harness.agent.memory.session.SessionEntry;
+import io.agentscope.harness.agent.memory.session.SessionTree;
+import io.agentscope.harness.agent.workspace.WorkspaceManager;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class MemoryFlushManagerOffloadTest {
+
+    @TempDir Path workspace;
+
+    @Test
+    void offloadMessages_skipsAlreadyPersistedPrefix_andKeepsChain() throws Exception {
+        RuntimeContext rc = RuntimeContext.builder().sessionId("session-1").build();
+
+        try (WorkspaceManager workspaceManager = new WorkspaceManager(workspace)) {
+            MemoryFlushManager flushManager = new MemoryFlushManager(workspaceManager, null);
+
+            flushManager.offloadMessages(
+                    rc,
+                    List.of(
+                            message("m1", MsgRole.USER, "hello"),
+                            message("m2", MsgRole.ASSISTANT, "world")),
+                    "agent-a",
+                    "session-1");
+
+            flushManager.offloadMessages(
+                    rc,
+                    List.of(
+                            message("m1", MsgRole.USER, "hello"),
+                            message("m2", MsgRole.ASSISTANT, "world"),
+                            message("m3", MsgRole.USER, "follow-up"),
+                            message("m4", MsgRole.ASSISTANT, "reply")),
+                    "agent-a",
+                    "session-1");
+        }
+
+        Path context = workspace.resolve("agents/agent-a/sessions/session-1.jsonl");
+        assertEquals(4, Files.readAllLines(context).size(), "offload should stay idempotent");
+
+        SessionTree tree = new SessionTree(context, workspace, null);
+        tree.load();
+        assertEquals(4, tree.size());
+
+        List<SessionEntry.MessageEntry> entries = tree.getMessageEntries();
+        assertEquals(
+                List.of("m1", "m2", "m3", "m4"),
+                entries.stream().map(SessionEntry.MessageEntry::getId).toList());
+        assertEquals(
+                Arrays.asList(null, "m1", "m2", "m3"),
+                entries.stream().map(SessionEntry.MessageEntry::getParentId).toList());
+    }
+
+    private static Msg message(String id, MsgRole role, String text) {
+        return Msg.builder()
+                .id(id)
+                .role(role)
+                .content(TextBlock.builder().text(text).build())
+                .build();
+    }
+}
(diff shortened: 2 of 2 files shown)


AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes a real and impactful bug: repeated `offloadMessages()` calls for the same session would re-append all historical messages, causing the session JSONL file to grow unboundedly. The fix is clean and minimal — it snapshots existing entry IDs after `syncFromRemote()`, then skips any incoming `Msg` whose stable ID is already present. The parent chain (`lastId`) is correctly initialized from the last existing entry. The regression test validates the core idempotency scenario well. One notable concern: when `Msg.getId()` is null or blank, dedup is silently bypassed, which could reintroduce the exact duplication bug under certain upstream conditions.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
```diff
@@ -253,11 +266,16 @@ private void offloadToSessionTree(
                 if (rendered == null || rendered.isBlank()) {
                     continue;
                 }
+                String entryId = normalizeEntryId(msg.getId());
+                if (entryId != null && existingIds.contains(entryId)) {
```
AgentScopeJavaBot, line 275: [major] When `msg.getId()` is null or blank, `normalizeEntryId` returns null, so the dedup guard (`entryId != null && existingIds.contains(entryId)`) is entirely bypassed — the message gets a fresh UUID and is always appended. If any upstream code path (e.g. a new provider adapter, user-constructed `Msg`, or a deserialization edge case) ever produces messages without a stable ID, the exact duplication bug this PR fixes will silently re-emerge.

Consider logging a warning when `entryId` is null so this situation is at least detectable in production:
```java
String entryId = normalizeEntryId(msg.getId());
if (entryId == null) {
    log.warn("Msg without stable ID encountered (role={}); dedup skipped", msg.getRole());
}
```
Alternatively, generate a deterministic ID from message content + role + timestamp as a fallback, though that's a larger change.

AgentScopeJavaBot, line 275: [nit] `normalizeEntryId` is a trivial one-liner (`id != null && !id.isBlank() ? id : null`). It's fine as a named method for readability, but you could inline it at the call site without loss of clarity:
```java
String rawId = msg.getId();
String entryId = (rawId != null && !rawId.isBlank()) ? rawId : null;
```
Either way works — just a style preference, not blocking.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class MemoryFlushManagerOffloadTest {
+
+    @TempDir Path workspace;
```
AgentScopeJavaBot, line 36: [minor] The test covers the happy path (stable IDs, overlapping prefix) nicely. Consider adding a test case for messages with null/blank IDs to document the expected behavior — currently those messages bypass dedup entirely and always get appended with a generated UUID. A test would lock in this contract and prevent accidental regressions:
```java
@Test
void offloadMessages_nullIdMessages_alwaysAppended() throws Exception {
    // messages with null IDs should still be appended (no dedup possible)
    // but should NOT cause duplicates of messages that DO have stable IDs
}
```

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
```diff
@@ -244,7 +246,18 @@ private void offloadToSessionTree(
...
 
-            String lastId = null;
+            List<SessionEntry> existingEntries = new ArrayList<>(tree.getAllEntries());
+            Set<String> existingIds =
+                    existingEntries.stream()
+                            .map(SessionEntry::getId)
+                            .collect(Collectors.toCollection(HashSet::new));
+            String lastId =
+                    existingEntries.isEmpty()
+                            ? null
```
AgentScopeJavaBot, line 256: [praise] Nice fix: initializing `lastId` from the last existing entry (rather than always `null`) correctly maintains the parent chain across offload calls. Previously, each offload started a disconnected chain, which was also a subtle correctness issue. This change fixes both the duplication bug and the broken parent linkage in one move.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
AgentScopeJavaBot, line 275: [major] When `msg.getId()` is null or blank, `normalizeEntryId` returns null, so the dedup guard (`entryId != null && existingIds.contains(entryId)`) is entirely bypassed — the message gets a fresh UUID and is always appended. If any upstream code path (e.g. a new provider adapter, user-constructed `Msg`, or a deserialization edge case) ever produces messages without a stable ID, the exact duplication bug this PR fixes will silently re-emerge.

Consider logging a warning when `entryId` is null so this situation is at least detectable in production:
```java
String entryId = normalizeEntryId(msg.getId());
if (entryId == null) {
    log.warn("Msg without stable ID encountered (role={}); dedup skipped", msg.getRole());
}
```
Alternatively, generate a deterministic ID from message content + role + timestamp as a fallback, though that's a larger change.

AgentScopeJavaBot, line 275: [nit] `normalizeEntryId` is a trivial one-liner (`id != null && !id.isBlank() ? id : null`). It's fine as a named method for readability, but you could inline it at the call site without loss of clarity:
```java
String rawId = msg.getId();
String entryId = (rawId != null && !rawId.isBlank()) ? rawId : null;
```
Either way works — just a style preference, not blocking.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
AgentScopeJavaBot, line 36: [minor] The test covers the happy path (stable IDs, overlapping prefix) nicely. Consider adding a test case for messages with null/blank IDs to document the expected behavior — currently those messages bypass dedup entirely and always get appended with a generated UUID. A test would lock in this contract and prevent accidental regressions:
```java
@Test
void offloadMessages_nullIdMessages_alwaysAppended() throws Exception {
    // messages with null IDs should still be appended (no dedup possible)
    // but should NOT cause duplicates of messages that DO have stable IDs
}
```

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
AgentScopeJavaBot, line 256: [praise] Nice fix: initializing `lastId` from the last existing entry (rather than always `null`) correctly maintains the parent chain across offload calls. Previously, each offload started a disconnected chain, which was also a subtle correctness issue. This change fixes both the duplication bug and the broken parent linkage in one move.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes a real and impactful bug: repeated `offloadMessages()` calls for the same session would re-append all historical messages, causing the session JSONL file to grow unboundedly. The fix is clean and minimal — it snapshots existing entry IDs after `syncFromRemote()`, then skips any incoming `Msg` whose stable ID is already present. The parent chain (`lastId`) is correctly initialized from the last existing entry. The regression test validates the core idempotency scenario well. One notable concern: when `Msg.getId()` is null or blank, dedup is silently bypassed, which could reintroduce the exact duplication bug under certain upstream conditions.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
guslegend0510: Good catch. I’ve updated the offload path to emit a warning when `msg.getId()` is null or blank, so anonymous messages no longer fail silently. The current behavior is still to append those messages, but we now surface the case explicitly for production visibility.

guslegend0510: Thanks. I kept `normalizeEntryId(...)` as a small helper for readability, but I agree this is mostly a style preference rather than a correctness issue. No behavioral change was needed here.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
guslegend0510: Added. I included a regression test covering null/blank message IDs alongside stable IDs. The test verifies that anonymous messages are still appended, while messages with stable IDs remain deduplicated across repeated offloads.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
guslegend0510: Thanks for calling that out. I’m keeping this change as-is — initializing `lastId` from the last existing entry preserves the chain correctly when offloading resumes into a non-empty session file.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
guslegend0510: Good catch. I’ve updated the offload path to emit a warning when `msg.getId()` is null or blank, so anonymous messages no longer fail silently. The current behavior is still to append those messages, but we now surface the case explicitly for production visibility.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
guslegend0510: Thanks for the note. I’m keeping this part unchanged — seeding `lastId` from the last persisted entry is intentional and preserves parent linkage correctly across incremental offloads.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
guslegend0510: Thanks for the note. I’m keeping this part unchanged — seeding `lastId` from the last persisted entry is intentional and preserves parent linkage correctly across incremental offloads.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
guslegend0510: Added. I included a regression test covering null/blank message IDs alongside stable IDs. The test verifies that anonymous messages are still appended, while messages with stable IDs remain deduplicated across repeated offloads.

guslegend0510:
Addressed in the latest update on this branch. I also added a regression test for null/blank message IDs and re-ran `MemoryFlushManagerOffloadTest`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
AgentScopeJavaBot: ✅ Verified as addressed in `2b0d6a1`: Warning log added at lines 270-274: log.warn("Msg without stable ID encountered (role={}); dedup skipped", msg.getRole()).

AgentScopeJavaBot: ✅ Verified as addressed in `2b0d6a1`: Bot explicitly said "Either way works — just a style preference, not blocking"; developer chose to keep normalizeEntryId as a helper, which the bot endorsed as acceptable.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
AgentScopeJavaBot: ✅ Verified as addressed in `2b0d6a1`: New test offloadMessages_nullAndBlankIds_stillAppendButDoNotBreakStableIds added at line 80, covering null/blank IDs with fresh UUID appends.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
AgentScopeJavaBot: ✅ Verified as addressed in `2b0d6a1`: Praise comment for lastId initialization fix at lines 254-257; code is correct as-is, no modification needed.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
AgentScopeJavaBot: ✅ Verified as addressed in `2b0d6a1`: Warning log added at lines 270-274: log.warn("Msg without stable ID encountered (role={}); dedup skipped", msg.getRole()).

AgentScopeJavaBot: ✅ Verified as addressed in `2b0d6a1`: Bot explicitly said "Either way works — just a style preference, not blocking"; developer chose to keep normalizeEntryId as a helper, which the bot endorsed as acceptable.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
AgentScopeJavaBot: ✅ Verified as addressed in `2b0d6a1`: Test offloadMessages_nullAndBlankIds_stillAppendButDoNotBreakStableIds exists at line 80, verifying null/blank ID messages are appended with generated UUIDs.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
AgentScopeJavaBot: ✅ Verified as addressed in `2b0d6a1`: Praise comment for lastId initialization fix at lines 254-257; code is correct as-is, no modification needed.

AgentScopeJavaBot (approved):

chickenlj (approved):
LGTM.

PR #1778, opened 2026-06-16
fix(agui): give reasoning messages a distinct messageId from the text answer

ql-wade:
## Problem

When `enableReasoning=true`, a single assistant message carrying both a `ThinkingBlock` and a `TextBlock` emits `REASONING_MESSAGE_*` and `TEXT_MESSAGE_*` that **share the same `messageId`** (both use `msg.getId()`):

```
REASONING_MESSAGE_START  messageId=<msgId>
TEXT_MESSAGE_START       messageId=<msgId>   ← same id
```

AG-UI clients key message grouping on `messageId`, so they collapse the reasoning and the answer into **one message bubble**. Observed with CopilotKit: the answer gets folded into the "Thought for a few seconds" panel and cannot be shown as a standalone, always-visible message.

## Root cause

In `AguiAgentAdapter#convertEvent`, the `TextBlock` branch (→ `TEXT_MESSAGE_*`) and the `ThinkingBlock` branch (→ `REASONING_MESSAGE_*`) both set:

```java
String messageId = msg.getId();
```

Per the AG-UI protocol, a reasoning message (role `"reasoning"`) and the text answer (role `"assistant"`) are **independent messages** — associated only by `runId` and order, never by `messageId` (the event records carry no linkage field). Reusing one id therefore conflates two distinct messages.

## Fix

Give the reasoning message a distinct id: `msg.getId() + "-reasoning"`.

- Renders independently of the answer; the answer stays always-visible while reasoning stays collapsible.
- A suffix (rather than a fresh `UUID`) keeps multiple `ThinkingBlock`s in one message deduped into a single reasoning message, mirroring how the text answer is one message.

## Verification

- New regression test `testReasoningAndTextAnswerHaveDistinctMessageIds`: a message with both a `ThinkingBlock` and a `TextBlock` must emit start events with **different** messageIds (`msg-both-reasoning` vs `msg-both`). Full suite green (36/36).
- End-to-end with CopilotKit consuming `POST /agui/run`: before the fix the reasoning+answer LCA was inside the collapsible `cpk:grid` container; after the fix the LCA is the top-level message list → reasoning and answer render as separate messages.

## Scope

One-line behavioral change + comment + test. No changes to event types or wire shape.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -365,22 +365,24 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
             return fresh;
         }
         try {
-            return stateStore
-                    .get(userId, sessionId, "agent_state", AgentState.class)
-                    .orElseGet(
-                            () -> {
-                                AgentState legacy =
-                                        LegacyStateLoader.loadFromLegacySession(
-                                                stateStore, userId, sessionId);
-                                if (legacy != null
-                                        && (!legacy.getContext().isEmpty()
-                                                || !legacy.getToolContext()
-                                                        .getActivatedGroups()
-                                                        .isEmpty())) {
-                                    return legacy;
-                                }
-                                return fresh;
-                            });
+            AgentState loaded =
+                    stateStore
+                            .get(userId, sessionId, "agent_state", AgentState.class)
+                            .orElseGet(
+                                    () -> {
+                                        AgentState legacy =
+                                                LegacyStateLoader.loadFromLegacySession(
+                                                        stateStore, userId, sessionId);
+                                        if (legacy != null
+                                                && (!legacy.getContext().isEmpty()
+                                                        || !legacy.getToolContext()
+                                                                .getActivatedGroups()
+                                                                .isEmpty())) {
+                                            return legacy;
+                                        }
+                                        return fresh;
+                                    });
+            return mergeInitialPermissionContext(loaded, permCtx);
         } catch (Exception e) {
             log.warn(
                     "Failed to load AgentState for slot (userId={}, sessionId={}): {}",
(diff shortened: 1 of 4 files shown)


agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -404,6 +406,50 @@ private static AgentState freshState(
...
+     * stale sessions in place, preserving their conversation/tool state while restoring HITL
+     * gating. This only fires for trivial persisted contexts (mode DEFAULT, no rules), so a session
+     * whose mode was deliberately changed via {@link #setPermissionMode} — which makes the context
+     * non-trivial by flipping the mode away from DEFAULT — is left untouched.
+     *
+     * @param loaded the state loaded from the store (or legacy fallback)
+     * @param initialPermCtx the builder's initial permission context (may be {@code null})
+     * @return {@code loaded} unchanged when it already has a non-trivial context or no initial
+     *     context was supplied; otherwise a copy of {@code loaded} with the initial context applied
+     */
```
AgentScopeJavaBot, line 430: [recommended] PR contains two unrelated fixes (AG-UI messageId + ReActAgent permission context merge) but only describes the first. Consider splitting into two PRs for easier review, bisect, and revert. Also, the `mergeInitialPermissionContext` method manually copies 10+ fields from `AgentState` — if `AgentState` gains new fields in the future, this method will silently drop them. Consider adding a `Builder.from(AgentState)` factory method to prevent this maintenance hazard.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/test/java/io/agentscope/core/agui/adapter/AguiAgentAdapterTest.java
```diff
@@ -1033,6 +1034,58 @@ void testRunWithThinkingBlockEvent() {
...
+        assertNotNull(reasoningId, "Should emit ReasoningMessageStart");
+        assertNotNull(textId, "Should emit TextMessageStart");
+        assertNotEquals(reasoningId, textId, "reasoning and text must have distinct messageIds");
+        assertEquals("msg-both-reasoning", reasoningId);
+        assertEquals("msg-both", textId);
+    }
+
     @Test
     void testRunWithStreamingThinkingBlockEvents() {
         // Test streaming reasoning events when enabled
```
AgentScopeJavaBot, line 1091: [nitpick] The streaming reasoning test asserts event counts but does not verify the `messageId` ends with `-reasoning`. Consider adding an assertion on the actual messageId value.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR contains two unrelated but individually correct fixes: (1) giving AG-UI reasoning messages a distinct `messageId` with `-reasoning` suffix to prevent client-side message collapsing, and (2) fixing stale `PermissionContextState` being discarded when the builder's initial trivial state overwrites loaded non-trivial state. Both fixes are correct in logic and well-tested. However, the PR description only mentions the first fix, and the two changes should ideally be in separate PRs.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot, line 430: [recommended] PR contains two unrelated fixes (AG-UI messageId + ReActAgent permission context merge) but only describes the first. Consider splitting into two PRs for easier review, bisect, and revert. Also, the `mergeInitialPermissionContext` method manually copies 10+ fields from `AgentState` — if `AgentState` gains new fields in the future, this method will silently drop them. Consider adding a `Builder.from(AgentState)` factory method to prevent this maintenance hazard.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/test/java/io/agentscope/core/agui/adapter/AguiAgentAdapterTest.java
AgentScopeJavaBot, line 1091: [nitpick] The streaming reasoning test asserts event counts but does not verify the `messageId` ends with `-reasoning`. Consider adding an assertion on the actual messageId value.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR contains two unrelated but individually correct fixes: (1) giving AG-UI reasoning messages a distinct `messageId` with `-reasoning` suffix to prevent client-side message collapsing, and (2) fixing stale `PermissionContextState` being discarded when the builder's initial trivial state overwrites loaded non-trivial state. Both fixes are correct in logic and well-tested. However, the PR description only mentions the first fix, and the two changes should ideally be in separate PRs.

PR #1811, opened 2026-06-17
fix(sandbox): use deterministic snapshot IDs and fix BuilderConfig local-mode startup crash

chang6666:
## AgentScope-Java Version

2.0.0-SNAPSHOT

### Background

Two bugs were discovered:

1. **BuilderConfig local-mode startup crash**: `BuilderConfig.builderBootstrap()` unconditionally configures `RemoteFilesystemSpec` for all agents, but when no distributed `AgentStateStore` bean is present, it falls back to `InMemoryAgentStateStore`. `HarnessAgent.build()` rejects this combination (`RemoteFilesystemSpec` + local state store), causing a startup `IllegalStateException`.

2. **Sandbox snapshot cannot be restored**: `DockerSandboxClient.create()` generates a random UUID as the snapshot identifier each time. When a sandbox container is lost but the snapshot `.tar` file survives, a re-creation generates a new UUID, making `LocalSandboxSnapshot.isRestorable()` always return `false` — the old snapshot is orphaned and workspace state (e.g. created files) is lost between agent calls.

### Changes

**BuilderConfig fix** (`BuilderConfig.java`):
- Wrap `RemoteFilesystemSpec` configuration in `if (sessionOpt.isPresent())` so it is only applied when a distributed `AgentStateStore` bean is available. Local mode uses the default local filesystem.

**Deterministic snapshot IDs** (9 files):
- Add a new default method `SandboxClient.create(WorkspaceSpec, SandboxSnapshotSpec, O, String snapshotId)` that accepts a caller-supplied stable snapshot identifier.
- `SandboxManager.acquire()` at Priority 4 now derives `snapshotId` from the `SandboxIsolationKey` (e.g. userId "alice") and passes it to the client. This ensures the same user/agent always maps to the same snapshot file (e.g. `.ws/alice.tar`).
- All 6 `SandboxClient` implementations (`Docker`, `Kubernetes`, `E2B`, `Daytona`, `AgentRun`, `InMemory`) override the 4-arg method and use the provided `snapshotId` for `snapshotSpec.build()`, falling back to the random UUID when `snapshotId` is null.
- Updated `SandboxManagerIsolationTest` mocks to match the new 4-arg call signature.

### How to test

1. **BuilderConfig fix**: Start `BuilderApp` without configuring a distributed `AgentStateStore` (no Redis). The app should boot successfully in local mode instead of throwing `IllegalStateException`.

2. **Snapshot restore**: Run the `DockerSandboxDemo1` example:
   - Ask the agent to create `hello.txt` → succeeds, snapshot persisted
   - Ask the agent to check if `hello.txt` exists → should find the file (previously failed)
   - Verify that `.ws/{userId}.tar` is created with a deterministic name derived from the isolation key

## Checklist

- [ ] Code has been formatted with `mvn spotless:apply`
- [ ] Javadoc comments are complete and follow project conventions
- [ ] Code is ready for review


AgentScopeJavaBot (commented):
🤖 AI Review

PR #1834, opened 2026-06-18
fix(shell): validate working_directory to prevent namespace escape

### LINKED ISSUE #1777: [Bug]:Shell execution can escape LocalFilesystemWithShell RuntimeContext namespace
**Describe the bug**

  When using `HarnessAgent` with `LocalFilesystemWithShell` and a `NamespaceFactory` for
  per-user/per-session workspace isolation, shell commands executed through the default
  `ShellExecuteTool` do not reliably stay inside the RuntimeContext namespace directory.

  This causes skills that write files to the current working directory to generate files
  under the shared workspace root instead of the current user's session workspace.

  There are two related issues in the current implementation:

  1. `AguiAgentAdapter.run(...)` calls:

  ```java
  agent.stream(msgs, options)

  There is no RuntimeContext parameter in this path. If the wrapped agent is a HarnessAgent,
  this can eventually use RuntimeContext.empty(), causing LocalFilesystemWithShell to fall
  back to the workspace root.

  2. ShellExecuteTool.execute(...) prepends working_directory directly:

  if (workingDirectory != null && !workingDirectory.isBlank()) {
      effectiveCommand = "cd " + workingDirectory + " && " + command;
  }

  working_directory is not validated, normalized, quoted, or restricted to the
  RuntimeContext namespace. If the model passes an absolute path, .., or the workspace root,
  it can override the namespaced cwd resolved by LocalFilesystemWithShell.

  To Reproduce

  1. Configure a HarnessAgent with LocalFilesystemWithShell.
  2. Set shellCwd to null.
  3. Configure a NamespaceFactory that returns a per-user/per-session namespace, for
     example:

  users/default/{userId}/sessions/{sessionId}

  4. Invoke the agent through AG-UI or execute the default shell tool.
  5. Ask the agent to run:

  pwd

  6. Or execute a skill that writes files to the current working directory.

  Expected behavior

  Shell commands should execute inside the RuntimeContext namespace directory, for example:

  workspace/users/default/{userId}/sessions/{sessionId}

  working_directory should also be resolved relative to that namespace and should not be
  able to escape it.

  Actual behavior

  Shell execution may run in the shared workspace root:

  workspace/

  As a result, generated skill files are written to a global shared directory instead of the
  current user's session workspace.

  Error messages

  There is no exception. The behavior is observable from pwd output and generated file
  locations.

  Environment

  - AgentScope-Java Version: 2.0.0-RC3
  - Java Version: 21
  - OS: macOS

  Additional context

  Relevant code:

  ShellExecuteTool.execute(...):

  if (workingDirectory != null && !workingDirectory.isBlank()) {
      effectiveCommand = "cd " + workingDirectory + " && " + command;
  }

  LocalFilesystemWithShell.resolveExecuteCwd(...) falls back to getCwd() when namespace is
  empty:

  List<String> ns = nsf.getNamespace(rc);
  if (ns == null || ns.isEmpty()) {
      return getCwd();
  }

  A safer behavior would be:

  - AG-UI adapter should provide a way to pass RuntimeContext.
  - ShellExecuteTool should validate and normalize working_directory.
  - working_directory should be resolved relative to the effective namespaced cwd, not the
    global workspace root.

  - Absolute paths and parent-directory escapes should be rejected.

chickenlj:
## Summary

- `ShellExecuteTool.execute()` blindly prepended `cd <working_directory> &&` to shell commands without any validation. An LLM-supplied absolute path, `~`, or `..` could escape the namespace-scoped workspace directory.
- Now rejects absolute paths, home-dir expansion (`~`), and parent-directory traversal (`..`).
- Shell-quotes the path to prevent injection via special characters.

**Security**: This is a namespace escape vulnerability — the LLM could execute commands outside the user's sandboxed workspace.

Fixes #1777

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
@@ -55,7 +55,12 @@ public class ShellExecuteTool {
                     int timeout) {
         String effectiveCommand = command;
         if (workingDirectory != null && !workingDirectory.isBlank()) {
-            effectiveCommand = "cd " + workingDirectory + " && " + command;
+            String wd = workingDirectory.strip();
+            if (wd.startsWith("/") || wd.startsWith("~") || wd.contains("..")) {
+                return "Error: working_directory must be a relative path within the workspace"
+                        + " (absolute paths, '~', and '..' are not allowed).";
+            }
+            effectiveCommand = "cd '" + wd.replace("'", "'\\''") + "' && " + command;
         }
 
         ExecuteResponse result =


AgentScopeJavaBot (commented):
🤖 AI Review

This PR adds input validation to `ShellExecuteTool.execute()` to prevent namespace escape via the LLM-supplied `workingDirectory` parameter. Previously, the working directory was concatenated into the shell command without any sanitization, allowing the LLM to use absolute paths, `~`, or `..` to execute commands outside the sandboxed workspace. The fix validates the path and applies proper POSIX single-quote shell escaping. The overall approach is sound and addresses the described vulnerability (fixes #1777). However, the `contains("..")` check is overly broad (false positives on legitimate directory names) and there are no unit tests covering this security-critical validation logic.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
```diff
@@ -55,7 +55,12 @@ public String execute(
                     int timeout) {
         String effectiveCommand = command;
         if (workingDirectory != null && !workingDirectory.isBlank()) {
-            effectiveCommand = "cd " + workingDirectory + " && " + command;
+            String wd = workingDirectory.strip();
+            if (wd.startsWith("/") || wd.startsWith("~") || wd.contains("..")) {
```
AgentScopeJavaBot, line 59: [recommended] **`contains("..")` is overly broad — rejects legitimate directory names.**

A path like `src/file..bak` or `module..v2/src` would be rejected even though `..` is not used as a parent-directory traversal component. This is a false positive that could block legitimate LLM workflows.

Consider using `Path.normalize()` instead:
```java
Path normalized = Path.of(wd).normalize();
if (normalized.startsWith("..")) {
    return "Error: ...";
}
```
`Path.normalize()` resolves all `..` segments, so even creative traversals like `subdir/../../etc` are caught correctly.

AgentScopeJavaBot, line 59: [recommended] **No unit tests for this security-critical validation logic.**

This is the primary defense against namespace escape via the shell tool. There are currently zero tests for `ShellExecuteTool`. Security-critical input validation should have dedicated test coverage including:

- **Rejected paths**: `/etc/passwd`, `~/secrets`, `../../etc`, `foo/../../etc`
- **Accepted paths**: `src/main`, `tests`, `sub/dir`
- **Edge cases**: `file..bak` (should be accepted), paths with single quotes, paths with spaces, empty/blank strings, null
- **Shell quoting correctness**: verify the constructed command properly escapes adversarial directory names

AgentScopeJavaBot, line 59: [recommended] **`contains("..")` is overly broad — rejects legitimate directory names.**

A path like `src/file..bak` or `module..v2/src` would be rejected even though `..` is not used as a parent-directory traversal component. This is a false positive that could block legitimate LLM workflows.

Consider using `Path.normalize()` instead:
```java
Path normalized = Path.of(wd).normalize();
if (normalized.startsWith("..")) {
    return "Error: ...";
}
```
`Path.normalize()` resolves all `..` segments, so even creative traversals like `subdir/../../etc` are caught correctly.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR adds input validation to `ShellExecuteTool.execute()` to prevent namespace escape via the LLM-supplied `workingDirectory` parameter. Previously, the working directory was concatenated into the shell command without any sanitization, allowing the LLM to use absolute paths, `~`, or `..` to execute commands outside the sandboxed workspace. The fix validates the path and applies proper POSIX single-quote shell escaping. The overall approach is sound and addresses the described vulnerability (fixes #1777). However, the `contains("..")` check is overly broad (false positives on legitimate directory names) and there are no unit tests covering this security-critical validation logic.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
AgentScopeJavaBot, line 59: [recommended] **No unit tests for this security-critical validation logic.**

This is the primary defense against namespace escape via the shell tool. There are currently zero tests for `ShellExecuteTool`. Security-critical input validation should have dedicated test coverage including:

- **Rejected paths**: `/etc/passwd`, `~/secrets`, `../../etc`, `foo/../../etc`
- **Accepted paths**: `src/main`, `tests`, `sub/dir`
- **Edge cases**: `file..bak` (should be accepted), paths with single quotes, paths with spaces, empty/blank strings, null
- **Shell quoting correctness**: verify the constructed command properly escapes adversarial directory names

itxaiohanglover (approved):
Good security fix — validating that working_directory is relative prevents path traversal attacks. The shell escaping (`wd.replace("'", "'\\'''")`) is the correct approach for single-quote escaping in POSIX shells. One question: should we also validate that the path doesn't contain null bytes? `\x00` could cause unexpected behavior in some shell implementations.

PR #1838, opened 2026-06-18
fix(harness): report sandbox file sizes in ls

### LINKED ISSUE #1837: [Bug]: 调用list_files工具返回文件为0字节问题
Image
如图所示，list_files工具返回文件为0字节，请问是个bug吗
2.0.0-RC3版本

guslegend0510:
## AgentScope-Java Version

2.0.0-SNAPSHOT

## Description

Fix `list_files` in sandbox-backed filesystems so file entries report their real size instead of always showing `0 bytes`.

This change updates the sandbox `ls` output to include file sizes and adds a regression test to prevent the issue from coming back.

Related issue: #1837

Closes #1837

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
@@ -77,13 +77,11 @@ public abstract class BaseSandboxFilesystem implements AbstractSandboxFilesystem
                         + escapedPath
                         + "/*; do "
                         + "  if [ -d \"$f\" ]; then "
-                        + "    mtime=$(stat -c '%Y' \"$f\" 2>/dev/null); "
-                        + "    mtime=${mtime:-0}; "
+                        + "    mtime=$(stat -c '%Y' \"$f\" 2>/dev/null || echo 0); "
                         + "    printf 'DIR:%s\\t%s\\n' \"$f\" \"$mtime\"; "
                         + "  elif [ -f \"$f\" ]; then "
-                        + "    meta=$(stat -c '%s\\t%Y' \"$f\" 2>/dev/null); "
-                        + "    size=${meta%%\\t*}; size=${size:-0}; "
-                        + "    mtime=${meta##*\\t}; mtime=${mtime:-0}; "
+                        + "    size=$(stat -c '%s' \"$f\" 2>/dev/null || echo 0); "
+                        + "    mtime=$(stat -c '%Y' \"$f\" 2>/dev/null || echo 0); "
                         + "    printf 'FILE:%s\\t%s\\t%s\\n' \"$f\" \"$size\" \"$mtime\"; "
                         + "  fi; "
                         + "done 2>/dev/null";
@@ -343,7 +341,11 @@ public abstract class BaseSandboxFilesystem implements AbstractSandboxFilesystem
                         + escapedPath
                         + " -type f -name "
                         + escapedPattern
-                        + " -exec stat -c '%n\\t%s\\t%Y' {} + 2>/dev/null | sort";
+                        + " 2>/dev/null | sort | while IFS= read -r f; do "
+                        + "  size=$(stat -c '%s' \"$f\" 2>/dev/null || echo 0); "
+                        + "  mtime=$(stat -c '%Y' \"$f\" 2>/dev/null || echo 0); "
+                        + "  printf '%s\\t%s\\t%s\\n' \"$f\" \"$size\" \"$mtime\"; "
+                        + "done";
 
         ExecuteResponse result = execute(runtimeContext, cmd, null);
         String output = result.output() != null ? result.output().strip() : "";
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
@@ -78,7 +78,7 @@ public abstract class BaseSandboxFilesystem implements AbstractSandboxFilesystem
                         + "/*; do "
                         + "  if [ -d \"$f\" ]; then echo \"DIR:$f\"; "
                         + "  elif [ -f \"$f\" ]; then "
-                        + "    size=$(wc -c < \"$f\" 2>/dev/null | tr -d '[:space:]'); "
+                        + "    size=$(stat -c '%s' \"$f\" 2>/dev/null); "
                         + "    size=${size:-0}; "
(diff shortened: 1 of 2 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
```diff
@@ -77,7 +77,11 @@ public LsResult ls(RuntimeContext runtimeContext, String path) {
                         + escapedPath
                         + "/*; do "
                         + "  if [ -d \"$f\" ]; then echo \"DIR:$f\"; "
-                        + "  elif [ -f \"$f\" ]; then echo \"FILE:$f\"; fi; "
+                        + "  elif [ -f \"$f\" ]; then "
+                        + "    size=$(wc -c < \"$f\" 2>/dev/null | tr -d '[:space:]'); "
+                        + "    size=${size:-0}; "
```
AgentScopeJavaBot: [nitpick] **Performance: `wc -c` vs `stat`** — `wc -c < "$f"` may read file content to count bytes on some platforms. On Linux sandboxes, GNU `wc` typically optimizes this via `lseek`, but `stat -c '%s' "$f"` is more idiomatic and guaranteed O(1) since it only reads inode metadata. Consider switching to `stat` for clarity and portability guarantees:
```bash
size=$(stat -c '%s' "$f" 2>/dev/null); size=${size:-0};
```

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystemTest.java
```diff
@@ -75,6 +91,11 @@ public String id() {
         public ExecuteResponse execute(
                 RuntimeContext runtimeContext, String command, Integer timeoutSeconds) {
             lastCommand = command;
+            if ("for f in '/workspace'/*; do   if [ -d \"$f\" ]; then echo \"DIR:$f\";   elif [ -f \"$f\" ]; then     size=$(wc -c < \"$f\" 2>/dev/null | tr -d '[:space:]');     size=${size:-0};     printf 'FILE:%s\\t%s\\n' \"$f\" \"$size\";   fi; done 2>/dev/null"
+                    .equals(command)) {
```
AgentScopeJavaBot: [nitpick] **Test fragility: full command string matching** — The fake sandbox matches the entire shell command string to decide the response. Any whitespace or formatting change in the command builder will silently break this test (it will return empty output instead of the expected mock). Consider matching on a command prefix or pattern (e.g., `command.startsWith("for f in")`) to make the test more resilient to refactoring.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a real bug where sandbox-backed filesystems always reported file sizes as `0 bytes` in `ls` output. The fix modifies the shell command to extract file sizes via `wc -c` and transmits them back using a tab-separated format, with proper fallback handling for backward compatibility and unparseable sizes. The parsing logic is robust — `lastIndexOf('\t')` correctly handles filenames containing tabs, and the `NumberFormatException` catch gracefully degrades to 0. A regression test is included. The change is well-scoped and directly addresses issue #1837.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
AgentScopeJavaBot: [nitpick] **Performance: `wc -c` vs `stat`** — `wc -c < "$f"` may read file content to count bytes on some platforms. On Linux sandboxes, GNU `wc` typically optimizes this via `lseek`, but `stat -c '%s' "$f"` is more idiomatic and guaranteed O(1) since it only reads inode metadata. Consider switching to `stat` for clarity and portability guarantees:
```bash
size=$(stat -c '%s' "$f" 2>/dev/null); size=${size:-0};
```

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystemTest.java
AgentScopeJavaBot: [nitpick] **Test fragility: full command string matching** — The fake sandbox matches the entire shell command string to decide the response. Any whitespace or formatting change in the command builder will silently break this test (it will return empty output instead of the expected mock). Consider matching on a command prefix or pattern (e.g., `command.startsWith("for f in")`) to make the test more resilient to refactoring.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a real bug where sandbox-backed filesystems always reported file sizes as `0 bytes` in `ls` output. The fix modifies the shell command to extract file sizes via `wc -c` and transmits them back using a tab-separated format, with proper fallback handling for backward compatibility and unparseable sizes. The parsing logic is robust — `lastIndexOf('\t')` correctly handles filenames containing tabs, and the `NumberFormatException` catch gracefully degrades to 0. A regression test is included. The change is well-scoped and directly addresses issue #1837.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
guslegend0510: Addressed in the latest update. I switched the size lookup from wc -ctostat -c '%s' and kept the fallback, so it is more idiomatic and portable now.

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystemTest.java
guslegend0510: Addressed in the latest update. The fake sandbox now matches on stable command patterns instead of the full command string, so the test is less brittle.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
guslegend0510: Addressed in the latest update. I switched the size lookup from wc -ctostat -c '%s' and kept the fallback, so it is more idiomatic and portable now.

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystemTest.java
guslegend0510: Addressed in the latest update. The fake sandbox now matches on stable command patterns instead of the full command string, so the test is less brittle.

PR #1841, opened 2026-06-19
fix(builder): fall back to LocalFilesystemSpec when no distributed AgentStateStore is configured

### LINKED ISSUE #1840: [Bug]: agentscope-builder fails to start out-of-the-box: RemoteFilesystemSpec rejected with InMemoryAgentStateStore fallback
**Describe the bug**

`agentscope-builder` fails to start with the default configuration (no extra
beans, embedded H2 datasource, no `DASHSCOPE_API_KEY`). The
`BuilderConfig#builderBootstrap` method always wires every agent with a
`RemoteFilesystemSpec`, but falls back to an `InMemoryAgentStateStore` when no
`AgentStateStore` bean is present. `HarnessAgent` explicitly rejects this
combination, so the application context can never refresh on a fresh checkout.

**To Reproduce**

1. Clone the repo and build the `agentscope-builder` module.

2. Run `BuilderApp` with the default `application.yml` (no `DASHSCOPE_API_KEY`,
   no extra Spring beans, default embedded H2 datasource):

   ```bash
   cd agentscope-examples/agents/agentscope-builder
   mvn spring-boot:run
   ```

3. See error — the context refresh fails with the stacktrace below.

   **Expected behavior**

   The app should boot out-of-the-box on a fresh checkout. The `application.yml`
   and `BuilderConfig` Javadoc suggest only `BaseStore` is required from operators,
   and it is already auto-wired from the Spring `DataSource`. The
   `InMemoryAgentStateStore` fallback described in the inline comment (line 199-202
   of `BuilderConfig.java`, *"for unit/integration tests, fall back to
   InMemoryAgentStateStore with a warning"*) should either (a) actually work with
   only a warning, or (b) not be used as the default for the production
   `BuilderApp`.

   **Error messages**

```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'userMarketplaceRegistry' defined in file [.../agentscope-builder/target/classes/io/agentscope/builder/runtime/marketplace/UserMarketplaceRegistry.class]: Unsatisfied dependency expressed through constructor parameter 0: Error creating bean with name 'sharedWorkspacePaths' defined in class path resource [io/agentscope/builder/web/workspace/BuilderWorkspaceConfig.class]: Unsatisfied dependency expressed through method 'sharedWorkspacePaths' parameter 0: Error creating bean with name 'builderBootstrap' defined in class path resource [io/agentscope/builder/web/config/BuilderConfig.class]: Failed to instantiate [io.agentscope.builder.runtime.BuilderBootstrap]: Factory method 'builderBootstrap' threw exception with message: filesystem(RemoteFilesystemSpec) is designed for distributed / multi-replica deployments, but the effective AgentStateStore is a local in-process implementation (JsonFileAgentStateStore / InMemoryAgentStateStore). Configure a distributed AgentStateStore (for example RedisAgentStateStore) via .stateStore(...) or use .distributedStore(...).
    ...
Caused by: java.lang.IllegalStateException: filesystem(RemoteFilesystemSpec) is designed for distributed / multi-replica deployments, but the effective AgentStateStore is a local in-process implementation (JsonFileAgentStateStore / InMemoryAgentStateStore). Configure a distributed AgentStateStore (for example RedisAgentStateStore) via .stateStore(...) or use .distributedStore(...).
    at io.agentscope.harness.agent.HarnessAgent$Builder.build(HarnessAgent.java:2007)
    at io.agentscope.builder.runtime.BuilderBootstrap$Builder.build(BuilderBootstrap.java:617)
    at io.agentscope.builder.web.config.BuilderConfig.builderBootstrap(BuilderConfig.java:226)
    ...
```

- AgentScope-Java Version: master

- Java Version: 17.0.12 (GraalVM)

- OS: macOS (darwin)

  **Additional context**

  **Root cause**

  `BuilderConfig#builderBootstrap` (line 203) falls back to
  `InMemoryAgentStateStore` when no `AgentStateStore` bean is registered:

```java
AgentStateStore stateStore = sessionOpt.orElseGet(InMemoryAgentStateStore::new);
```

…while unconditionally configuring every agent with a `RemoteFilesystemSpec`
(line 220-223):

```java
b.filesystem(
        new RemoteFilesystemSpec(baseStore)
                .isolationScope(IsolationScope.USER)
                .addSharedPrefix("activity/"));
```

`HarnessAgent` rejects this combination (line 2006-2014):

```java
if (remoteFilesystemSpec != null && isLocalSession(effectiveSession)) {
    throw new IllegalStateException(...);
}
```

The inline comment (line 199-202) says *"for unit/integration tests, fall back
to InMemoryAgentStateStore with a warning"*, but the harness actually **throws**
rather than warns — so the fallback is unreachable in practice for `BuilderApp`.

**Suggested fix**

The builder module already depends on `agentscope-extensions-mysql`, and a
`DataSource` is always available (auto-configured H2 by default). A
`MysqlAgentStateStore` (which implements `AgentStateStore` and is **not**
classified as "local" by `isLocalSession`) can be auto-wired the same way
`BaseStore` already is:

```java
@Bean
@ConditionalOnMissingBean(AgentStateStore.class)
public AgentStateStore agentStateStore(DataSource dataSource) {
    return new MysqlAgentStateStore(dataSource, true);
}
```

This makes the app start out-of-the-box on the default H2 datasource, while
operators can still override with `RedisAgentStateStore` etc. for production.

> Note: `MysqlAgentStateStore` uses MySQL-specific DDL (`LONGTEXT`,
> `CHARACTER SET utf8mb4 ...`). The default H2 URL already sets `MODE=MYSQL`,
> but the `CHARACTER SET` / `COLLATE` clauses on `CREATE TABLE` /
> `CREATE DATABASE` may need handling for H2 compatibility — worth verifying
> during the fix.

chang6666:
BuilderConfig unconditionally wired every agent with RemoteFilesystemSpec but fell back to InMemoryAgentStateStore when no distributed AgentStateStore bean was present. HarnessAgent rejects this combination, so the app could never start on a fresh checkout with the default H2 datasource.

Now when the effective AgentStateStore is local (InMemory/JsonFile), LocalFilesystemSpec is used instead, producing a consistent local-mode topology. Distributed deployments are unaffected — RemoteFilesystemSpec is still used when a distributed AgentStateStore (Redis/MySQL/etc.) is present.

## Related Issue
Fixes #1840 

## AgentScope-Java Version

`2.0.0-SNAPSHOT` 
### Background

`BuilderConfig` configures the filesystem spec for every agent during startup. The previous implementation unconditionally used `RemoteFilesystemSpec`, and fell back to `InMemoryAgentStateStore` when no distributed `AgentStateStore` bean was injected. However, `HarnessAgent` validates the store-vs-filesystem topology at build time — `RemoteFilesystemSpec` requires a distributed store that can be shared across pods, while `InMemoryAgentStateStore` is in-process and not shareable.

### Problem

On a fresh checkout with the default H2 datasource (i.e. no Redis/MySQL distributed `AgentStateStore` bean injected), the harness throws `IllegalStateException` due to the topology mismatch, so `BuilderApp` cannot start at all. This is an out-of-the-box usability bug.

### Changes Made

`agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/web/config/BuilderConfig.java` (+38 / -12)

1. Added imports: `io.agentscope.core.state.JsonFileAgentStateStore` and `io.agentscope.harness.agent.filesystem.spec.LocalFilesystemSpec`.
2. Added helper method `isLocalStateStore(AgentStateStore)`: returns `true` when the store is `InMemoryAgentStateStore` or `JsonFileAgentStateStore`.
3. Refactored the filesystem wiring in `configureAllAgents`:
   - When the effective store is local, use `LocalFilesystemSpec` with `IsolationScope.USER` (no shared prefix, since there is no cross-pod sharing in local mode).
   - When the effective store is distributed, keep the original behavior using `RemoteFilesystemSpec(baseStore)` with the `activity/` shared prefix.
4. Added INFO logs recording which filesystem spec is in effect, to aid deployment troubleshooting.
5. Refined the existing WARN message: "DistributedStore" → "distributed AgentStateStore bean" for accuracy.

### How to Test

**Local mode (default H2) — verify the fix:**

```bash
cd agentscope-examples/agents/agentscope-builder
mvn spring-boot:run
```

Expected: the app starts normally; logs show `Effective AgentStateStore is local (InMemoryAgentStateStore); using LocalFilesystemSpec.`; no `IllegalStateException`. Open the builder UI, create an agent, and start a conversation to verify filesystem writes succeed.

**Distributed mode (Redis) — verify no regression:**

Inject a distributed `AgentStateStore` bean (e.g. from `agentscope-extensions-redis`) and start:

```bash
mvn spring-boot:run -Dspring-boot.run.profiles=redis
```

Expected: logs show `Effective AgentStateStore is distributed (...); using RemoteFilesystemSpec.`; behavior is unchanged from before the fix; `activity/` still goes to the shared `BaseStore`.

**Regression tests:**

```bash
mvn -pl agentscope-examples/agents/agentscope-builder test
```

## Checklist


chang6666:
fix issue #1840

AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes a genuine out-of-the-box startup failure where `BuilderConfig` unconditionally wired `RemoteFilesystemSpec` but fell back to `InMemoryAgentStateStore` when no distributed store was available — a combination `HarnessAgent` correctly rejects. The fix is clean and well-scoped: a new `isLocalStateStore()` helper detects local stores and switches to `LocalFilesystemSpec`, producing a consistent local-mode topology. The existing test (`BuilderAppContextLoadTest`) remains unaffected because its Mockito proxy is typed as `AgentStateStore` (not `InMemoryAgentStateStore`), so it still takes the `RemoteFilesystemSpec` path. One recommended improvement: the whitelist-based `instanceof` check is fragile — any future local `AgentStateStore` implementation will silently fall into the distributed branch and reproduce the original bug.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/web/config/BuilderConfig.java
```diff
@@ -272,6 +285,16 @@ public ChatUiChannel chatUiChannel(BuilderBootstrap bootstrap) {
...
     // -----------------------------------------------------------------
 
+    /**
+     * Returns {@code true} when the given store is a local, in-process implementation that cannot
+     * be shared across pods, i.e. {@link InMemoryAgentStateStore} or {@link JsonFileAgentStateStore}.
+     * When this returns {@code true}, the harness must use {@link LocalFilesystemSpec} instead of
+     * {@link RemoteFilesystemSpec}.
+     */
+    private static boolean isLocalStateStore(AgentStateStore store) {
+        return store instanceof InMemoryAgentStateStore || store instanceof JsonFileAgentStateStore;
```
AgentScopeJavaBot, line 295: [major] The whitelist-based detection (`instanceof InMemoryAgentStateStore || instanceof JsonFileAgentStateStore`) is fragile: if a new local `AgentStateStore` implementation is added (e.g. an embedded SQLite store), it will silently fall into the distributed branch and reproduce the exact topology-mismatch bug this PR fixes.

Consider either:
1. Adding a marker interface or method on `AgentStateStore` itself, e.g. `default boolean isDistributed() { return true; }`, overridden to `false` by local implementations; or
2. Inverting to a blacklist of known distributed stores:
```java
private static boolean isLocalStateStore(AgentStateStore store) {
    return !(store instanceof MysqlAgentStateStore
            || store instanceof RedisAgentStateStore
            || store instanceof JedisAgentStateStore
            || store instanceof RedissonAgentStateStore
            || store instanceof OssAgentStateStore);
}
```
Option 1 is more robust and pushes the classification to where it belongs (the store contract). Option 2 is a quick improvement if changing the interface is out of scope for this PR.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/web/config/BuilderConfig.java
```diff
@@ -196,31 +198,42 @@ public BuilderBootstrap builderBootstrap(
...
-                    // `activity/` is routed to the shared BaseStore so the per-agent audit log
-                    // (written by AgentActivityStore) is visible across pods, not pinned to the
-                    // local disk of whichever pod served the write.
-                    b.filesystem(
-                            new RemoteFilesystemSpec(baseStore)
-                                    .isolationScope(IsolationScope.USER)
-                                    .addSharedPrefix("activity/"));
+                    if (localStore) {
+                        b.filesystem(new LocalFilesystemSpec().isolationScope(IsolationScope.USER));
+                    } else {
```
AgentScopeJavaBot, line 231: [nit] In local mode, `LocalFilesystemSpec` defaults to `LocalFsMode.ROOTED` with `project` resolving to `System.getProperty("user.dir")`. For the builder context, consider explicitly setting `.project(cwd)` so the agent's shell `pwd` matches the builder's configured workspace root rather than the JVM working directory:
```java
b.filesystem(new LocalFilesystemSpec()
    .project(cwd)
    .isolationScope(IsolationScope.USER));
```
This is non-blocking — the default works, but explicit is clearer.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes a genuine out-of-the-box startup failure where `BuilderConfig` unconditionally wired `RemoteFilesystemSpec` but fell back to `InMemoryAgentStateStore` when no distributed store was available — a combination `HarnessAgent` correctly rejects. The fix is clean and well-scoped: a new `isLocalStateStore()` helper detects local stores and switches to `LocalFilesystemSpec`, producing a consistent local-mode topology. The existing test (`BuilderAppContextLoadTest`) remains unaffected because its Mockito proxy is typed as `AgentStateStore` (not `InMemoryAgentStateStore`), so it still takes the `RemoteFilesystemSpec` path. One recommended improvement: the whitelist-based `instanceof` check is fragile — any future local `AgentStateStore` implementation will silently fall into the distributed branch and reproduce the original bug.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/web/config/BuilderConfig.java
AgentScopeJavaBot, line 295: [major] The whitelist-based detection (`instanceof InMemoryAgentStateStore || instanceof JsonFileAgentStateStore`) is fragile: if a new local `AgentStateStore` implementation is added (e.g. an embedded SQLite store), it will silently fall into the distributed branch and reproduce the exact topology-mismatch bug this PR fixes.

Consider either:
1. Adding a marker interface or method on `AgentStateStore` itself, e.g. `default boolean isDistributed() { return true; }`, overridden to `false` by local implementations; or
2. Inverting to a blacklist of known distributed stores:
```java
private static boolean isLocalStateStore(AgentStateStore store) {
    return !(store instanceof MysqlAgentStateStore
            || store instanceof RedisAgentStateStore
            || store instanceof JedisAgentStateStore
            || store instanceof RedissonAgentStateStore
            || store instanceof OssAgentStateStore);
}
```
Option 1 is more robust and pushes the classification to where it belongs (the store contract). Option 2 is a quick improvement if changing the interface is out of scope for this PR.

agentscope-examples/agents/agentscope-builder/src/main/java/io/agentscope/builder/web/config/BuilderConfig.java
AgentScopeJavaBot, line 231: [nit] In local mode, `LocalFilesystemSpec` defaults to `LocalFsMode.ROOTED` with `project` resolving to `System.getProperty("user.dir")`. For the builder context, consider explicitly setting `.project(cwd)` so the agent's shell `pwd` matches the builder's configured workspace root rather than the JVM working directory:
```java
b.filesystem(new LocalFilesystemSpec()
    .project(cwd)
    .isolationScope(IsolationScope.USER));
```
This is non-blocking — the default works, but explicit is clearer.

chickenlj (approved):

PR #1843, opened 2026-06-20
fix: propagate parent middleware to subagents

guslegend0510:
## Description

Propagate middleware registered on the parent `HarnessAgent.Builder` to both built-in general-purpose subagents and markdown-declared subagents. This fixes the regression where parent middleware was dropped during subagent construction.

Validation:
- `mvn -pl agentscope-harness -am spotless:apply`
- `mvn -pl agentscope-harness -am -Dtest=HarnessAgentTest#parentMiddleware_propagatesToBuiltinAndMarkdownSubagents test`

## Related Issues

- [#1839](https://github.com/agentscope-ai/agentscope-java/issues/1839)


agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1351,15 +1353,56 @@ public Builder hooks(List<Hook> hooks) {
...
+            List<MiddlewareBase> copyable = new ArrayList<>(middlewares.size());
+            for (MiddlewareBase middleware : middlewares) {
+                if (middleware != null && !isHarnessRuntimeMiddleware(middleware)) {
+                    copyable.add(middleware);
+                }
+            }
+            return copyable;
+        }
+
+        private static boolean isHarnessRuntimeMiddleware(MiddlewareBase middleware) {
```
AgentScopeJavaBot: [minor] The `isHarnessRuntimeMiddleware` instanceof chain covers 16 types. If a future harness-runtime middleware is added and this method isn't updated, `fromAgent()` will leak it into the Builder's middleware list, causing double-registration in rebuilt agents (and their subagents). Consider a lightweight marker — e.g. a `HarnessRuntimeMiddleware` tag interface or a `boolean isRuntimeOnly()` default method on `MiddlewareBase` — so new middleware types opt-in explicitly instead of relying on this centralised allowlist.

```java
// Example: marker interface approach
public interface HarnessRuntimeMiddleware {}

public class CompactionMiddleware extends MiddlewareBase
        implements HarnessRuntimeMiddleware { ... }
```

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1351,15 +1353,56 @@ public Builder hooks(List<Hook> hooks) {
...
         public Builder middleware(MiddlewareBase middleware) {
-            inner.middleware(middleware);
+            if (middleware != null) {
+                middlewares.add(middleware);
+                inner.middleware(middleware);
+            }
             return this;
         }
 
         public Builder middlewares(List<? extends MiddlewareBase> middlewares) {
```
AgentScopeJavaBot: [nit] The parameter `middlewares` shadows the instance field of the same name, which makes the for-each loop slightly harder to parse at a glance (`for (MiddlewareBase middleware : middlewares)` — which `middlewares`?). Consider renaming the parameter to `middlewareList` or `items` for clarity.

agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentTest.java
```diff
@@ -482,6 +484,81 @@ void subagentsDir_loadsMarkdownDeclarations() throws Exception {
                 "subagents/*.md declaration should use filename as name");
     }
 
+    @Test
+    void parentMiddleware_propagatesToBuiltinAndMarkdownSubagents() throws Exception {
```
AgentScopeJavaBot, line 490: [praise] Solid regression test — validates middleware propagation through both the built-in general-purpose subagent path and the markdown-declared subagent path, and confirms the middleware is actually invoked (not just present in the list). Good use of `AtomicInteger` to track invocations across both paths.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a real regression where middleware registered on a parent `HarnessAgent.Builder` was silently dropped during subagent construction. The fix introduces a parallel `middlewares` list on the Builder to track user-registered middleware, captures an immutable snapshot at build time, and propagates it to both the built-in general-purpose subagent factory and the markdown-declared subagent factory. A `filterCopyableMiddlewares` method correctly strips harness-runtime middleware when cloning via `fromAgent()` to avoid double-registration. The included test validates both propagation paths and the `fromAgent` filter. The implementation is clean, correct, and well-tested.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
guslegend0510: Thanks, I changed this to a `HarnessRuntimeMiddleware` marker so runtime-only middleware is opt-in and no longer depends on a centralized allowlist.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
guslegend0510: Good catch. I renamed the parameter to `middlewareList` for clarity.

chickenlj (approved):

PR #1847, opened 2026-06-21
[codex] Fix wakeup queue startup drain

HeadBro1:
## What changed

- subscribe to wakeup signals before draining the durable startup queue
- continue draining in batches until fewer than 64 entries remain
- add regression coverage for the startup subscription window and a 65-entry backlog

## Why

`WakeupDispatcher` previously drained once before subscribing to the transient wakeup signal. A wakeup enqueued between those operations could remain queued until another signal arrived. Startup recovery also processed at most 64 entries, leaving larger backlogs behind without a guaranteed follow-up signal.

## Impact

Dispatchers now establish the live subscription before startup recovery and fully consume pre-existing wakeups across queue batches.

## Validation

`mvn -pl agentscope-harness -am -Dtest=WakeupDispatcherTest -Dsurefire.failIfNoSpecifiedTests=false test`

- 6 tests run
- 0 failures
- 0 errors

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/gateway/WakeupDispatcher.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/gateway/WakeupDispatcher.java
@@ -70,11 +70,10 @@ public class WakeupDispatcher implements AutoCloseable {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


HeadBro1:
Closing and resubmitting this change with the project's standard branch and Conventional Commit naming.

PR #1856, opened 2026-06-22
feat(model): expose cached prompt tokens in ChatUsage

### LINKED ISSUE #1854: [Feature]: Expose cached prompt tokens in ChatUsage (OpenAI prompt_tokens_details.cached_tokens)
## Summary

`ChatUsage` (`io.agentscope.core.model.ChatUsage`) currently exposes only `inputTokens`, `outputTokens`, and `time`. The OpenAI API returns a finer breakdown under `usage.prompt_tokens_details.cached_tokens`, which reports how many of the prompt tokens were served from the prompt cache. This value is **already deserialized** by the framework into `OpenAIUsage.PromptTokensDetails.cachedTokens`, but it is **dropped** when `OpenAIResponseParser` converts `OpenAIUsage` into `ChatUsage`.

As a result, downstream consumers reading `ModelCallEndEvent.getUsage()` (or `Msg.getUsage()`) have no way to know how many input tokens hit the cache, which is essential for accurate cost accounting (cached input tokens are billed at a reduced rate).

## Current Behavior

- `OpenAIUsage.PromptTokensDetails.cachedTokens` is correctly parsed from `prompt_tokens_details.cached_tokens`.
- `OpenAIResponseParser#parseCompletionResponse` and `#parseChunkResponse` build `ChatUsage` using only `getPromptTokens()` and `getCompletionTokens()`; the cached-token detail is discarded.
- `ChatUsage` has no field to carry cached tokens, so the information is unavailable to any consumer of `ModelCallEndEvent` / `ChatResponse`.

## Expected Behavior

`ChatUsage` should carry the cached prompt token count so consumers can compute cache-aware costs.

Note on semantics: `cachedTokens` is a **subset of** `inputTokens` (i.e. `inputTokens = non-cached prompt tokens + cachedTokens`), not an additional amount. `getTotalTokens()` semantics (`inputTokens + outputTokens`) should remain unchanged.

## Proposed Change

1. Add an `int cachedTokens` field to `ChatUsage` with a `getCachedTokens()` getter and a `cachedTokens(int)` builder method.
   - Keep the existing 3-arg constructor (delegating with `cachedTokens = 0`) to preserve backward compatibility for the ~8 existing `ChatUsage.builder()` call sites.
   - Move `@JsonCreator` to the new 4-arg constructor; since the field is a primitive `int`, older serialized JSON without the property deserializes safely to `0` (verified against the expectations in `MsgUsageSerializationTest`).
2. In `OpenAIResponseParser`, add a `getSafeCachedTokens(OpenAIUsage)` helper and populate `.cachedTokens(...)` in both `parseCompletionResponse` and `parseChunkResponse`.
3. Add unit tests for: cached tokens present, absent (`prompt_tokens_details` null → 0), and `ChatUsage` JSON round-trip with the new field.

## Scope

This issue intentionally covers the **OpenAI path only**, since the data is already available at the DTO layer there. Extending cache-token reporting to other providers (e.g. Anthropic `cache_read_input_tokens` / `cache_creation_input_tokens`, which are not yet parsed) can be handled in separate follow-up issues/PRs to keep changes focused.

## Backward Compatibility

- Existing `ChatUsage.builder()` callers are unaffected (`cachedTokens` defaults to `0`).
- Existing serialized `ChatUsage` JSON remains deserializable.
- `getTotalTokens()` behavior is unchanged.

I'd be happy to submit a PR for this if it aligns with the project roadmap.

itxaiohanglover:
Fixes #1854

## Problem

The OpenAI API returns `cached_tokens` in `usage.prompt_tokens_details.cached_tokens` for requests that hit the KV cache. The SDK already parses this into `OpenAIUsage.PromptTokensDetails.cachedTokens`, but `ChatUsage` does not expose it — users have no way to know how many tokens were served from cache.

## What changed

1. `ChatUsage.java` — added `cachedTokens` field (nullable `Integer`, null when provider does not report it), with getter and builder method
2. `OpenAIResponseParser.java` — extract `cachedTokens` from `OpenAIUsage.getPromptTokensDetails()` in both `parseCompletionResponse` and streaming paths, via a small `getCachedTokens()` helper

Fully backward compatible — `cachedTokens` defaults to `null`, existing code is unaffected.

## Validation

- `mvn compile -pl agentscope-core` — pass
- `cachedTokens` is null when `prompt_tokens_details` is absent (most providers)
- `cachedTokens` is populated when OpenAI returns it (e.g. repeated prompts with prompt caching enabled)

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIResponseParser.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIResponseParser.java
@@ -580,4 +580,11 @@ public class OpenAIResponseParser {
                 .finishReason(finishReason)
                 .build();
     }
+
+    private static Integer getCachedTokens(OpenAIUsage usage) {
+        if (usage.getPromptTokensDetails() != null) {
+            return usage.getPromptTokensDetails().getCachedTokens();
+        }
+        return null;
+    }
 }
--- a/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIResponseParser.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/formatter/openai/OpenAIResponseParser.java
@@ -113,6 +113,7 @@ public class OpenAIResponseParser {
                                 .time(
                                         Duration.between(startTime, Instant.now()).toMillis()
                                                 / 1000.0)
+                                .cachedTokens(getCachedTokens(openAIUsage))
                                 .build();
             }
 
@@ -346,6 +347,7 @@ public class OpenAIResponseParser {
                                 .time(
                                         Duration.between(startTime, Instant.now()).toMillis()
                                                 / 1000.0)
+                                .cachedTokens(getCachedTokens(openAIUsage))
                                 .build();
             }
 
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ChatUsage.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ChatUsage.java
@@ -51,6 +51,17 @@ public class ChatUsage {
     }
 
     /**
+     * Backward-compatible constructor without cached tokens.
+     *
+     * @param inputTokens the number of tokens used for the input/prompt
+     * @param outputTokens the number of tokens used for the output/generated response
+     * @param time the execution time in seconds
+     */
+    public ChatUsage(int inputTokens, int outputTokens, double time) {
+        this(inputTokens, outputTokens, time, null);
+    }
+
+    /**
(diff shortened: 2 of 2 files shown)


itxaiohanglover:
The fork CI failure is Codecov token only — compilation and tests pass. The fork does not have the CODECOV_TOKEN secret, so the coverage upload step fails with `fail_ci_if_error: true`. This is expected for fork PRs.

itxaiohanglover:
Closing in favor of #1868 which has a more complete implementation (getSafeCachedTokens helper, better Javadoc). Same fix, better execution.

PR #1868, opened 2026-06-22
feat(core): expose cached prompt tokens in ChatUsage

### LINKED ISSUE #1854: [Feature]: Expose cached prompt tokens in ChatUsage (OpenAI prompt_tokens_details.cached_tokens)

DennisWLX:
Add cachedTokens field to ChatUsage to carry OpenAI's prompt_tokens_details.cached_tokens through to consumers.

- Add int cachedTokens field with getter and builder method
- Keep backward-compatible 3-arg constructor (defaults to 0)
- Add getSafeCachedTokens() helper in OpenAIResponseParser
- Populate cachedTokens in both non-stream and stream paths
- Add unit tests for all null/absent/present scenarios

Closes #1854

## AgentScope-Java Version

## Description

## Checklist


AgentScopeJavaBot (commented):
🤖 AI Review

agentscope-core/src/main/java/io/agentscope/core/model/ChatUsage.java
```diff
@@ -22,28 +22,48 @@
...
+        this(inputTokens, outputTokens, 0, time);
+    }
+
     /**
      * Creates a new ChatUsage instance.
      *
      * @param inputTokens the number of tokens used for the input/prompt
      * @param outputTokens the number of tokens used for the output/generated response
+     * @param cachedTokens the number of input tokens served from the prompt cache (a subset of
+     *     {@code inputTokens}); {@code 0} when the provider does not report cache information
```
AgentScopeJavaBot, line 55: [major] `ReasoningContext` maintains `inputTokens`/`outputTokens`/`time` accumulator fields and builds a final `ChatUsage` via `getChatUsage()`, but the new `cachedTokens` field is not propagated through `processChunk()` or the builder call. In streaming scenarios, `cachedTokens` will always be 0. Add a `cachedTokens` field to `ReasoningContext` and wire it through `processChunk()` and `getChatUsage()` following the same pattern as `inputTokens`.

AgentScopeJavaBot, line 55: [major] `ReActAgent.collectAggregatedUsage()` accumulates `inputTokens`/`outputTokens`/`time` across multi-round structured output but does not accumulate `cachedTokens`. Add `totalCached` to the loop and include it in the final builder call.

DennisWLX: Good catch, thanks. Fixed in d27cda57.

Added a totalCached accumulator to the loop in collectAggregatedUsage() and included .cachedTokens(totalCached) in the final builder, so cached tokens are now summed across multi-round structured-output calls.

DennisWLX: Good catch, thanks. Fixed in d27cda57.

Added a cachedTokens field to ReasoningContext and wired it through processChunk() (reads usage.getCachedTokens()) and getChatUsage(). I also found that buildFinalMessage() rebuilds ChatUsage independently from the accumulator fields, so it had the same gap — that path is fixed too. Added ReasoningContextTest.testCachedTokensPropagation to verify cached tokens survive the full streaming path.

chickenlj (approved):

PR #1870, opened 2026-06-22
fix(ag-ui):subagent event forward

### LINKED ISSUE #1046: [Feature]: Sub-agent events are not visible in AG-UI frontend when using `SubAgentTool` with `forwardEvents(true)`
# Sub-agent events are not visible in AG-UI frontend when using `SubAgentTool` with `forwardEvents(true)`

## Summary

When a supervisor agent delegates tasks to sub-agents via `SubAgentTool` with `SubAgentConfig.forwardEvents(true)`, the sub-agent's reasoning process (thinking, tool calls, tool results) is completely invisible to the AG-UI frontend. The user sees the sub-agent tool call start, then a long pause (black box), and finally the complete result all at once — with no intermediate streaming.

## Environment

- **agentscope-java version**: 1.0.10
- **AG-UI adapter**: `agentscope-agui-spring-boot-starter:1.0.10`

## Steps to Reproduce

1. Create a supervisor agent with a sub-agent registered via `SubAgentConfig.forwardEvents(true)`:

```java
SubAgentConfig config = SubAgentConfig.builder()
    .forwardEvents(true)
    .streamOptions(
        StreamOptions.builder()
            .eventTypes(EventType.ALL)
            .includeReasoningChunk(true)
            .includeActingChunk(true)
            .build()
    )
    .build();

toolkit.registration()
    .subAgent(mySubAgentProvider, config)
    .apply();
```

2. Connect the supervisor agent to AG-UI via `AguiAgentAdapter`
3. Send a message that triggers sub-agent invocation
4. Observe the AG-UI SSE event stream

## Expected Behavior

The frontend should receive real-time streaming events showing the sub-agent's:
- Reasoning/thinking process (incremental text chunks)
- Tool call invocations (tool name, arguments)
- Tool call results

This would provide transparency into the sub-agent's execution, similar to how the parent agent's events are streamed.

## Actual Behavior

The frontend receives:
1. `TOOL_CALL_START` (sub-agent tool call begins)
2. **Nothing for the entire duration of sub-agent execution** (black box)
3. `TOOL_CALL_END` + `TOOL_CALL_RESULT` (final result appears all at once)

The sub-agent's intermediate reasoning and tool usage are completely hidden from the user.

## Root Cause Analysis

I traced this through two specific gaps in the event pipeline:

### Gap 1: `AguiAgentAdapter.run()` hardcodes `StreamOptions` without `includeActingChunk`

In [`AguiAgentAdapter.java` line 98-99](https://github.com/agentscope-ai/agentscope-java/blob/main/agentscope-agui-spring-boot-starter/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java):

```java
StreamOptions options =
    StreamOptions.builder().eventTypes(EventType.ALL).incremental(true).build();
```

Since `includeActingChunk` defaults to `false`, the parent agent's event stream **silently drops all `ActingChunkEvent`s**. These are exactly the events that carry forwarded sub-agent data (emitted by `SubAgentTool.forwardEvent()` via `ToolEmitter.emit()`).

### Gap 2: `AguiAgentAdapter.convertEvent()` has no handling for forwarded sub-agent events

Even if `ActingChunkEvent`s were included in the stream, `convertEvent()` only handles two event types:

```java
if (type == EventType.REASONING) {
    // ... handles text, thinking, tool use blocks
} else if (type == EventType.TOOL_RESULT && event.isLast()) {
    // ... handles tool results, but ONLY when isLast=true
}
// No handling for ACTING events or forwarded sub-agent metadata
```

The sub-agent events forwarded by `SubAgentTool.forwardEvent()` are wrapped in `ToolResultBlock` metadata:

```java
// SubAgentTool.forwardEvent()
Map<String, Object> metadata = new HashMap<>();
metadata.put("subagent_event", event);
metadata.put("subagent_name", agent.getName());
metadata.put("subagent_id", agent.getAgentId());
metadata.put("subagent_session_id", sessionId);
emitter.emit(new ToolResultBlock(null, null,
    List.of(TextBlock.builder().text(json).build()), metadata));
```

But `AguiAgentAdapter` never inspects this metadata or converts it to AG-UI events.

### Event Flow Diagram

```
Sub-agent generates events (REASONING chunks, TOOL_RESULT, etc.)
    ↓
SubAgentTool.forwardEvent() wraps as ToolResultBlock with metadata
    ↓
ToolEmitter.emit() → triggers ActingChunkEvent on parent agent
    ↓
AguiAgentAdapter.run() streams parent agent with includeActingChunk=false
    ↓
✗ ActingChunkEvent is DROPPED — never reaches convertEvent()
    ↓
Even if it did reach convertEvent(), no logic exists to extract
sub-agent metadata and convert to AG-UI events
    ↓
Frontend sees nothing until sub-agent completes
```

## Workaround

We implemented an application-level workaround using a custom `Hook` + `Sinks.Many<AguiEvent>` merge pattern:

### 1. Register a per-request event sink

```kotlin
// In the chat service, before agent execution:
val subAgentSink = Sinks.many()
    .multicast()
    .onBackpressureBuffer<AguiEvent>()

// Store sink keyed by agent identity for hook access
AiChatContext.registerSubAgentSink(agent, SubAgentSinkInfo(
    subAgentSink, threadId, runId
))

// Merge custom events with main AG-UI stream
val mergedEvents = Flux.merge(
    aguiResult.events().doOnComplete { subAgentSink.tryEmitComplete() },
    subAgentSink.asFlux()
)
```

### 2. Custom Hook intercepts `ActingChunkEvent` and emits AG-UI events

```kotlin
@Component
class SubAgentEventForwardingHook : Hook {

    override fun <T : HookEvent> onEvent(event: T): Mono<T> {
        if (event !is ActingChunkEvent) return Mono.just(event)

        val metadata = event.chunk.metadata ?: return Mono.just(event)
        if (!metadata.containsKey("subagent_event")) return Mono.just(event)

        val sinkInfo = AiChatContext.getSubAgentSink(event.agent)
            ?: return Mono.just(event)
        val subEvent = metadata["subagent_event"] as? Event
            ?: return Mono.just(event)
        val agentName = metadata["subagent_name"] as? String ?: "SubAgent"

        // Convert to AG-UI event and push to sink
        val aguiEvent = convertToAguiEvent(subEvent, agentName, sinkInfo)
        aguiEvent?.let { sinkInfo.sink.tryEmitNext(it) }

        return Mono.just(event)
    }
}
```

This works but requires significant boilerplate at the application level. Ideally, the framework should handle this natively.

### Important: `StreamOptions` tuning required

When configuring `SubAgentConfig.streamOptions`, you must set `includeReasoningResult(false)` to avoid duplicate content:

```java
StreamOptions.builder()
    .eventTypes(EventType.ALL)
    .includeReasoningChunk(true)
    .includeReasoningResult(false)  // ← Critical: prevents duplicate output
    .includeActingChunk(true)
    .build()
```

With `includeReasoningResult(true)`, both incremental chunks AND accumulated results are forwarded, causing every sentence to appear twice on the frontend.

### Option A: Enhance `AguiAgentAdapter` to natively support sub-agent event forwarding

1. Add `includeActingChunk(true)` to the `StreamOptions` in `AguiAgentAdapter.run()`:

```java
StreamOptions options = StreamOptions.builder()
    .eventTypes(EventType.ALL)
    .incremental(true)
    .includeActingChunk(true)  // Enable acting chunk events
    .build();
```

2. Add sub-agent event handling in `convertEvent()`:

```java
// In convertEvent(), handle forwarded sub-agent events
if (msg.getContent() != null) {
    for (ContentBlock block : msg.getContent()) {
        if (block instanceof ToolResultBlock toolResult
                && toolResult.getMetadata() != null
                && toolResult.getMetadata().containsKey("subagent_event")) {
            Event subEvent = (Event) toolResult.getMetadata().get("subagent_event");
            String agentName = (String) toolResult.getMetadata().get("subagent_name");
            // Convert sub-agent event to appropriate AG-UI events
            events.addAll(convertSubAgentEvent(subEvent, agentName, state));
        }
    }
}
```

### Option B: Allow `AguiAdapterConfig` to accept custom `StreamOptions`

Let users override the `StreamOptions` used in `AguiAgentAdapter.run()`:

```java
public class AguiAdapterConfig {
    // ... existing fields ...
    private StreamOptions streamOptions; // New: custom stream options
}
```

### Option C: Provide a built-in `SubAgentEventConverter`

Create a dedicated component that integrates with `AguiAgentAdapter` to handle sub-agent event conversion, without requiring application-level hooks.

## Related Issues

- #918 — Streaming bug with agents as tools
- #818 — Message delivery when using agent as a tool

## Impact

This issue affects any AG-UI application using the supervisor + sub-agent pattern. Without the workaround, sub-agent execution appears as a "black box" to end users, degrading the user experience significantly — especially for sub-agents that perform complex, multi-step reasoning (e.g., permission analysis, data retrieval).

rudy2steiner:
## Background
 Fixes #1046 When a supervisor agent delegates to a sub-agent via SubAgentTool with SubAgentConfig.forwardEvents(true), the sub-agent's intermediate
  reasoning/text was completely invisible to the AG-UI frontend — the UI saw TOOL_CALL_START, then a long pause, then TOOL_CALL_END + TOOL_CALL_RESULT arriving together. Effectively a "black box" during sub-agent execution,
  especially painful for multi-step researchers/analysts.

## Changes
 * added a new branch for non-last TOOL_RESULT events for subagent events
 * new unit test that constructs the same wire format SubAgentTool.forwardEvent produces, runs it through the adapter


agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java
```diff
@@ -258,6 +258,18 @@ private List<AguiEvent> convertEvent(Event event, EventConversionState state) {
                     }
                 }
             }
+        } else if (type == EventType.TOOL_RESULT && !event.isLast()) {
+            for (ContentBlock block : msg.getContent()) {
+                if (block instanceof ToolResultBlock toolResult) {
+                    Map<String, Object> metadata = toolResult.getMetadata();
+                    if (metadata != null) {
+                        Object inner = metadata.get("subagent_event");
```
AgentScopeJavaBot: [major] subagent_event is a cross-module magic string shared between SubAgentTool (core) and this adapter (extensions). A typo here silently drops all sub-agent events. Extract to a shared constant, e.g. SubAgentTool.METADATA_KEY_SUBAGENT_EVENT, and add a debug log when the metadata value is not an Event instance.

AgentScopeJavaBot: [major] Recursive convertEvent(innerEvent, state) shares the same EventConversionState. If the sub-agent stream terminates without a final isLast=true REASONING event (e.g. cancellation, error), currentTextMessageId remains stuck on the sub-agent message ID, potentially corrupting subsequent supervisor text messages. Consider saving/restoring currentTextMessageId around the recursive call.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes the AG-UI adapter to correctly forward sub-agent intermediate events (reasoning/text) to the frontend when SubAgentTool has forwardEvents=true. The fix adds a new branch for non-last TOOL_RESULT events, extracting the inner Event from metadata and recursively converting it. Key concerns: (1) subagent_event magic string is a cross-module contract with no shared constant; (2) recursive convertEvent shares EventConversionState, risking currentTextMessageId state pollution; (3) no null-check on the inner Event extracted from metadata.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java
```diff
@@ -258,6 +258,18 @@ private List<AguiEvent> convertEvent(Event event, EventConversionState state) {
...
+                        if (inner instanceof Event innerEvent) {
+                            events.addAll(convertEvent(innerEvent, state));
+                        }
+                    }
```
AgentScopeJavaBot, line 362: [minor] No null/instance check on the value extracted from metadata. If the metadata contains a non-Event object, the recursive convertEvent call will fail with a ClassCastException. Add a defensive instanceof check before the recursive call.

rudy2steiner: Won't fix. inner instanceof Event innerEvent already does both null- and type-checking via pattern matching.won't fix

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java
rudy2steiner: Won't fix. finishRun() already emits trailing TextMessageEnd for any unended message

itxaiohanglover:
Good refactor! Extracting metadata keys as public constants makes them usable by downstream consumers without string duplication. The AG-UI adapter integration looks clean — forwarding sub-agent events through the metadata channel is a pragmatic approach.

oss-maintainer:

```bash
git fetch origin
git rebase origin/main
git push --force-with-lease
```

---

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java
```diff
@@ -550,6 +553,18 @@ private List<AguiEvent> convertEvent(Event event, EventConversionState state) {
...
                 }
             }
+        } else if (type == EventType.TOOL_RESULT && !event.isLast()) {
+            for (ContentBlock block : msg.getContent()) {
+                if (block instanceof ToolResultBlock toolResult) {
+                    Map<String, Object> metadata = toolResult.getMetadata();
+                    if (metadata != null) {
+                        Object inner = metadata.get(SubAgentTool.METADATA_KEY_SUBAGENT_EVENT);
+                        if (inner instanceof Event innerEvent) {
+                            events.addAll(convertEvent(innerEvent, state));
```
oss-maintainer, line 563: Recursing with the *parent's* `EventConversionState` means a sub-agent's chunks and the supervisor's own chunks share one state machine and one message identity. Two consequences:

1. The consumer cannot tell sub-agent text from supervisor text — they merge into the same text message, so a UI that renders the run shows the supervisor and the delegated agent as one voice. The metadata this PR adds (`METADATA_KEY_SUBAGENT_NAME` / `_ID` / `_SESSION_ID`) exists precisely to distinguish them, but nothing reads it.
2. With a genuinely parallel batch (two sub-agents emitting at once, or a nested supervisor→sub→sub chain), interleaved `TextMessageStart/Content/End` from different sources drive one state object, so the deltas get woven into each other rather than separated.

Could you either open a distinct message/step per `subagent_name` (that is what the AG-UI step/message ids are for), or — if flattening into the parent message is deliberate — drop the three identity metadata keys so the API doesn't promise provenance the adapter then throws away? A test with two concurrent sub-agent chunks would pin whichever answer you pick.

agentscope-core/src/main/java/io/agentscope/core/tool/subagent/SubAgentTool.java
```diff
@@ -400,10 +416,12 @@ private void forwardEvent(Event event, ToolEmitter emitter, Agent agent, String
         try {
             String json = JsonUtils.getJsonCodec().toJson(event);
             Map<String, Object> metadata = new HashMap<>();
-            metadata.put("subagent_event", event == null ? "" : event);
-            metadata.put("subagent_name", agent.getName() == null ? "" : agent.getName());
-            metadata.put("subagent_id", agent.getAgentId() == null ? "" : agent.getAgentId());
-            metadata.put("subagent_session_id", sessionId == null ? "" : sessionId);
+            metadata.put(METADATA_KEY_SUBAGENT_EVENT, event == null ? "" : event);
```
oss-maintainer, line 419: Two representations of the same event now travel in one block: the serialized JSON in the text content, and the live `Event` object in metadata. The adapter reads only the metadata one (`inner instanceof Event`), so the JSON is dead weight on this path — and the object form does not survive the transports this repo cares about. Once a `ToolResultBlock` is Jackson round-tripped (state persistence, the remote/distributed sub-agent path, any gateway that re-serializes events), metadata comes back as a `LinkedHashMap`, `instanceof Event` is false, and forwarding stops *silently* — no log, no error, just a missing stream.

Since you already pay for `toJson(event)` a few lines above, decoding from the text side gives one source of truth that works identically in-process and over the wire — e.g. `JsonUtils.getJsonCodec().fromJson(json, Event.class)` in the adapter, keeping metadata for the scalar identity fields only. If keeping the object form is deliberate, please at least log at debug when the metadata value is present but not an `Event`, so the serialized-boundary case is diagnosable instead of invisible.

oss-maintainer (commented):
### Summary
Forward a sub-agent's intermediate events to the supervisor stream by stashing the inner `Event` under a `subagent_event` metadata key in `SubAgentTool.forwardEvent`, then having `AguiAgentAdapter` pick it up on non-last `TOOL_RESULT` chunks and recursively convert it. Replacing the four string-literal metadata keys with `SubAgentTool` constants is a straight improvement, and the new test documents the intent well.

Two blocking issues before this can land:

- **[CI]** `build (ubuntu-latest)` and `build (windows-latest)` both fail on `8165fec`, at `spotless:check` on `agentscope-extensions-agui`: `JsonException` and `JsonUtils` are imported in `AguiAgentAdapter.java` but never used. `mvn spotless:apply` clears it. This is also why the PR has sat unmerged — CI has been red on every head.
- **[Warning]** The recursion converts sub-agent chunks through the parent's `EventConversionState`, so sub-agent text is indistinguishable from supervisor text and a concurrent/nested batch would interleave into one message. The `subagent_name` / `subagent_id` metadata added here is never read — either use it (separate message/step per sub-agent) or drop it.
- **[Warning]** The inner event is carried as a live Java object in metadata while the JSON copy already in the text content goes unread; the object form does not survive Jackson round-trips (state persistence, remote sub-agents), where forwarding then stops silently. Decoding from the JSON makes in-process and distributed behave the same.

Also: the PR is 3 months old and its head is a merge of today's `main`, so please rebase rather than merge — `git merge main` is what keeps bumping `updatedAt` without moving the content forward.

CLA signed. Happy to re-review once spotless is green.

---

PR #1890, opened 2026-06-23
refactor(model): move OpenAI model provider from core to extensions module

jujn:
## Summary

This PR migrates the OpenAI model adapter out of `agentscope-core` into a dedicated extension module, `agentscope-extensions-model-openai`, and introduces a lightweight SPI path so model providers can be discovered from extension jars instead of being hard-coded in core.

This is the first step of the model-adapter modularization work. OpenAI is migrated first to validate the main flow; the remaining providers and richer extension/starter design will be handled incrementally.

### OpenAI adapter moved to extensions

- Added `agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai`.
- Moved OpenAI-related implementation classes from core to the new extension package:
  - `OpenAIChatModel`
  - `OpenAIClient`
  - OpenAI DTOs
  - OpenAI formatters
  - OpenAI exception types
  - `OpenAICredential`
  - `OpenAIMultiModalTool`
- Added `OpenAIChatModelFactory` as a small bridge for integration layers that should not directly compile against OpenAI internals.
- Added the OpenAI extension module to the extension parent, BOM, and all-in-one distribution.

### Model SPI and registry discovery

- Added `ModelProvider` SPI in core.
- Updated `ModelRegistry` to resolve providers in this order:
  1. named model registrations
  2. user-registered factories
  3. SPI providers from extension modules
  4. remaining built-in core providers
- Added OpenAI SPI registration through `META-INF/services/io.agentscope.core.model.spi.ModelProvider`.
- Added `OpenAIModelProvider`, supporting `openai:<model-name>` ids using `OPENAI_API_KEY`.

`ModelProvider` is intentionally minimal for this phase. It is enough for simple `modelId -> Model` creation, but not yet strong enough for complex provider-specific configuration. This will be enhanced later once more model providers are migrated and the configuration boundary becomes clearer.

### Provider-neutral retry handling

- Added `ModelHttpException` in core as a provider-neutral HTTP error contract.
- Updated `ExecutionConfig` so retry decisions no longer depend on OpenAI-specific exception classes.
- `ExecutionConfig` now treats `429` and `5xx` from `ModelHttpException` as retryable, while `400/401/403/404/422` remain non-retryable.
- Updated OpenAI extension exceptions so `OpenAIException` implements `ModelHttpException`.

This preserves the previous default retry behavior for OpenAI rate limit errors without keeping OpenAI exception classes in core.

### Shared formatter/model utilities

- Moved `JsonSchema` out of the OpenAI DTO package into `io.agentscope.core.formatter`, because `ResponseFormat` is a core abstraction and should not depend on OpenAI packages.
- Made `ModelContextWindows` reusable by extensions so the migrated OpenAI model can keep using the existing context-window lookup table.

### Spring Boot starter compatibility

- Updated the current generic starter so OpenAI auto-configuration can still create an OpenAI model after the class moved to the extension module.
- The starter currently uses a reflection bridge to call `OpenAIChatModelFactory`, keeping the OpenAI extension as a test/runtime-optional dependency instead of a direct main dependency.

This is a transitional design. A later change would move starter integration to `ModelRegistry + ModelProvider SPI`, or follow a provider-specific starter style similar to LangChain4j, for example:

- `agentscope-openai-spring-boot-starter` directly depends on the OpenAI extension and owns full OpenAI configuration.
- The generic starter focuses on agent/memory/toolkit auto-configuration, or accepts a user-provided `Model` bean.

### E2E test adjustment

- Temporarily removed OpenAI-compatible E2E providers from core E2E tests.
- Core should not own extension-provider E2E coverage after model adapters move out of core.
- After all model providers are migrated, a dedicated E2E test module would be added under `extensions-models` to cover OpenAI, DeepSeek, GLM, OpenRouter, and other compatible providers in the right ownership boundary.

### Related updates

- Updated tests and package references for the migrated OpenAI classes.
- Updated example/docs imports where they directly referenced moved OpenAI model/tool classes.
- Files that only needed import/package path changes were updated mechanically without behavioral changes.

## Compatibility Notes

This migration changes the public package of OpenAI-specific classes from core packages to extension packages. Users that directly import `OpenAIChatModel`, OpenAI formatters, OpenAI exceptions, `OpenAICredential`, or `OpenAIMultiModalTool` need to depend on `agentscope-extensions-model-openai` and update imports.

Core no longer imports OpenAI extension classes.


agentscope-examples/agents/agentscope-codingagent/src/main/java/io/agentscope/harness/coding/agent/CodingAgentFactory.java
```diff
@@ -18,8 +18,8 @@
 import io.agentscope.core.model.AnthropicChatModel;
 import io.agentscope.core.model.DashScopeChatModel;
 import io.agentscope.core.model.Model;
-import io.agentscope.core.model.OpenAIChatModel;
 import io.agentscope.core.tool.Toolkit;
+import io.agentscope.extensions.model.openai.OpenAIChatModel;
 import io.agentscope.harness.agent.HarnessAgent;
```
chickenlj, line 23: This is a breaking user API change, make sure we state the migration steps clearly in 2.0 docs.

chickenlj (approved):
LGTM.

PR #1903, opened 2026-06-25
fix(kubernetes): fix WebSocket race in hydrateWithArchive causing exit=null

Buktal:
Move watch.exitCode().get() inside the try (ExecWatch) block so the
process exit code is retrieved before watch.close() closes the WebSocket.
Previously the exit code Future was captured inside the block but .get()
was called outside, creating a race where the exit code is lost and the
future completes with null.

Fixes #1902

## AgentScope-Java Version

commit 5dc62907 on branch fix/hydrate-archive-websocket-race

## Description

`Fabric8KubernetesPodRuntime.hydrateWithArchive()` has a race condition
between WebSocket close and exit code retrieval. The exit code
`CompletableFuture<Integer>` is captured inside the `try (ExecWatch)` block,
but `.get()` is called **after** the block exits. By that time,
`watch.close()` (via try-with-resources) has already closed the WebSocket
channel, so the `cat` process's exit code is lost and the future completes
with `null`, producing `archive upload failed (exit=null)`.

The fix moves `watch.exitCode().get(TAR_TIMEOUT_SECONDS + 10L, TimeUnit.SECONDS)`
inside the `try` block, ensuring the exit code is awaited before the
WebSocket is closed. A `TimeoutException` is also caught inside the block
so it can be rethrown as a proper `SandboxRuntimeException`.

## Checklist

Please check the following items before code is ready to be reviewed.

- [x]  Code has been formatted with `mvn spotless:apply`
- [ ]  All tests are passing (`mvn test`)
- [x]  Javadoc comments are complete and follow project conventions
- [ ]  Related documentation has been updated (e.g. links, examples, etc.)
- [x]  Code is ready for review

PR #1915, opened 2026-06-25
fix(kubernetes): treat null exit code as success in hydrateWithArchive

Buktal:
## Problem

`hydrateWithArchive` gets stuck or reports false "archive upload failed (exit=null)" when uploading tar archives to K8s sandbox pods.

## Root Cause

**Two bugs in one:**

1. **Fabric8 / K8s exec WebSocket v4 protocol limitation** — there is no way to close stdin without closing the entire WebSocket. When `ExecWatch.close()` runs, the WebSocket is shut before exit status (stream 3) arrives, so `exitCode` completes with `null`.

2. **Previous fix (#1903) is incorrect** — `exitCode.get()` was moved **inside** the try-with-resources block, causing a deadlock: `get()` blocks waiting for exit code, but `ExecWatch.close()` (which triggers stdin EOF → cat exits → exit code arrives) only runs at try end.

## Fix

```java
// Before (#1903, wrong): exitCode.get() inside try → deadlock
try (ExecWatch watch = ...) {
    exitCode.get(130s);  // blocks forever — close() never runs
}

// After (correct): get() outside try, null = success
CompletableFuture<Integer> f;
try (ExecWatch watch = ...) {
    f = watch.exitCode();
}
Integer code = f.get(130s);  // close() already ran
if (code != null && code != 0) { throw ...; }
// null = WebSocket closed before exit status arrived, but data was written

Evidence from Fabric8 API

Fabric8's own ExecWatch.exitCode() JavaDoc documents this behavior:

/**
 * Will be -1 if the exit code can't be determined from the status,
 * or null if close is received before the exit code.
 *
 * See https://github.com/kubernetes/kubernetes/issues/89899
 * - which explains there's currently no way to indicate
 * end of input over a websocket, so you may not get
 * an exit code when using stdIn.
 */
CompletableFuture<Integer> exitCode();

Null exit code is expected behavior when using stdin with K8s exec WebSocket v4 protocol — the process ran and exited normally, but the WebSocket closed before the exit status message could arrive. This is not a real failure.

Testing

- Null exit code test updated from assertThrows → assertDoesNotThrow
- Non-zero exit code still correctly throws
- Zero exit code still passes successfully


itxaiohanglover:
Nice fix! Moving `exitCode().get()` outside the try-with-resources is the right call — the ExecWatch resources need to close first. Treating null exit code as success makes sense for `cat` which exits cleanly on EOF. The error message with exit code and stderr buffer is much more helpful for debugging.

Buktal:
Priority PR. Previous fix was buggy & sub-par. Needs optimization. @chickenlj

Buktal:
@chickenlj @jujn Priority PR. Previous fix was buggy & sub-par. Needs optimization.

chickenlj (approved):

PR #1925, opened 2026-06-26
feat(openai): add per-capability default model config via Builder

### LINKED ISSUE #1924: [Feature]: Support per-capability default model configuration in OpenAIMultiModalTool
`OpenAIMultiModalTool` has a `defaultModelName` field that only applies to
`openai_image_to_text`. The other three tools have their defaults hardcoded internally:

- `openai_text_to_image` → `"dall-e-3"` (hardcoded)
- `openai_text_to_audio` → `"tts-1"` (hardcoded)
- `openai_audio_to_text` → `"whisper-1"` (hardcoded)

Each capability corresponds to a distinct model family, so a single shared
`defaultModelName` does not make sense across all four tools.

Add a `Builder` to `OpenAIMultiModalTool` (consistent with `OpenAIChatModel`),
exposing per-capability default model fields:

```java
OpenAIMultiModalTool tool = OpenAIMultiModalTool.builder()
    .apiKey("sk-...")
    .baseUrl("https://api.openai.com")          // optional, defaults to null
    .defaultVisionModel("gpt-4o-mini")          // optional, falls back to "gpt-4o"
    .defaultImageGenModel("dall-e-2")           // optional, falls back to "dall-e-3"
    .defaultTtsModel("tts-1-hd")                // optional, falls back to "tts-1"
    .defaultSttModel("whisper-1")               // optional, falls back to "whisper-1"
    .build();
```

When a capability-specific default is not set, the existing hardcoded value
is used — no change to current behavior.

The existing constructors remain for backward compatibility, with
`defaultModelName` mapping to `defaultVisionModel` internally.

Passing a `Map<String, String>` of capability → model. Rejected because named
builder methods are more discoverable and type-safe.

**Additional context**
- `defaultModelName` should be deprecated in favor of `defaultVisionModel` to
  clarify its actual scope.
- Builder style should follow `OpenAIChatModel.Builder` conventions already in
  this module.
- Builder also exposes `baseUrl` to match existing constructors.

Buktal:
## AgentScope-Java Version

2.0.0-SNAPSHOT

### 1. Per-capability default model via Builder

`OpenAIMultiModalTool` previously had a single `defaultModelName` field that only applied to `openai_image_to_text`. The other three tools had their default models hardcoded.

This adds a `Builder` matching `OpenAIChatModel.Builder` conventions, exposing per-capability defaults:
- `defaultVisionModel` — image-to-text (falls back to `gpt-4o`)
- `defaultImageGenModel` — text-to-image (falls back to `dall-e-3`)
- `defaultTtsModel` — text-to-audio (falls back to `tts-1`)
- `defaultSttModel` — audio-to-text (falls back to `whisper-1`)
- `baseUrl` — custom endpoint (optional)

When not set via Builder, the existing hardcoded values are used — no change to current behavior.

### 2. Flaky test fix

`SubAgentToolTimeoutRetryIntegrationTest.oldAgentIsStopped` was flaky: when the timeout cancelled the subscription, `boundedElastic` interrupted the sleeping thread via `Future.cancel(true)`. The old code wrapped both `Thread.sleep` and `Files.writeString` in one try-catch, so the interrupt silently skipped the write, making the assertion fail. Fixed by splitting them.

**Related:**
- Fixes #1924

## Checklist


itxaiohanglover:
Nice work on the Builder — backward compat is well handled with the old constructor delegating to the new private one. One thing I noticed: the builder tests only do `assertNotNull(tool)`, so it'd be good to also verify the configured models actually get used in API calls (like the existing `testImageToText_usesCustomDefaultModel` does). Also, the InterruptedException fix in SubAgentToolTimeoutRetryIntegrationTest seems unrelated to this PR — might be worth a separate commit.

AgentScopeJavaBot (commented):
🤖 AI Review

Adds Builder pattern for OpenAIMultiModalTool with per-capability default models. Behavioral change: old 4-param constructor now only sets vision model. Consider marking old constructor @Deprecated.

PR #1865, opened 2026-06-22
fix(dashscope): forward structured response format

guslegend0510:
## AgentScope-Java Version

2.0.0-SNAPSHOT

## Description

This PR fixes DashScope native structured output by forwarding `GenerateOptions.responseFormat` into `DashScopeParameters.responseFormat` during request building.

Before this change, the DashScope request path dropped `response_format`, so structured-output calls could not be parsed correctly and `Msg.hasStructuredData()` stayed false.

Changes made:
- forward `responseFormat` in DashScope option mapping
- add regression coverage for explicit and default response format handling

How to test:
- `mvn -pl agentscope-core -Dtest=DashScopeToolsHelperComprehensiveTest test`

## Checklist

- [x] Unit tests pass
- [x] DashScope structured output is included in the request body
- [x] Fixes agentscope-ai/agentscope-java#1852

PR #1640, opened 2026-06-06
fix: 修复 DeepSeek tool call 参数解析失败导致结构化输出丢失

guslegend0510:
当使用 DeepSeek 兼容 OpenAI 接口时，tool_calls[].function.arguments 可能返回不合法 JSON。原实现会在解析失败时跳过整条 tool call，导致 ReAct 流程中断，最终结构化输出为空。

本次修复将非流式 tool call 的参数解析改为降级处理：即使 JSON 解析失败，也会保留原始 arguments 并继续构建 ToolUseBlock，避免丢失工具调用上下文。同时补充回归测试，覆盖 malformed JSON 场景，确保 tool call 仍能被正确保留。

Closes #1628

PR #1935, opened 2026-06-26
fix(dashscope): enhance support for native structured output in DashScope models

chickenlj:
related to #1865,#1743,#1640


itxaiohanglover:
Good robustness fix! The fallback from native to synthetic structured call prevents hard failures. The context cleanup on error (removing added messages, resetting nativeResponseFormat) is important for not corrupting state on retry.

PR #1949, opened 2026-06-27
fix(event): avoid Jackson 3 type property conflict

### LINKED ISSUE #1948: [Bug]:io.agentscope.core.event.AgentEvent Deserialize error
Describe the bug                                                       
                                                                                                                                                                                     
  When streaming AgentStartEvent (and likely other events from io.agentscope.core.event) via Spring MVC's ResponseBodyEmitter / SseEmitter, Jackson 3 (tools.jackson.databind) throws
   an InvalidDefinitionException during serialization. The root cause is a conflict between the @JsonTypeInfo type-id property named type and a bean property with the same name on  
  the event class. The same SSE endpoint works fine for plain String/JSON-POJO payloads, so the issue is specific to AgentScope's event classes.

  To Reproduce


  1. Build a Spring Boot controller that returns an SseEmitter and inside it send() an io.agentscope.core.event.AgentStartEvent instance, e.g.:

  @GetMapping(value = "/chat/stream", produces = MediaType.TEXT_EVENT_STREAM_VALUE)
  public SseEmitter chat() {
      SseEmitter emitter = new SseEmitter();
      emitter.send(new AgentStartEvent(...));   // fails here
      emitter.complete();
      return emitter;
  }
  2. Execute: curl http://localhost:8080/chat/stream
  3. See error: HttpMessageConversionException → InvalidDefinitionException: Conflict between type id property 'type' and bean property with same name

  Expected behavior

  AgentStartEvent (and sibling event classes in io.agentscope.core.event) should serialize to JSON without raising a InvalidDefinitionException. Either:

  - The @JsonTypeInfo annotation on the event base class should use JsonTypeInfo.As.EXISTING_PROPERTY (or As.EXTERNAL_PROPERTY) so the type id is not duplicated as a synthetic
  property, or
  - The concrete event class should not declare a bean property named type (rename it, or annotate it with @JsonProperty/@JsonIgnore appropriately), or
  - A mixin / custom ObjectMapper configuration should be provided to resolve the conflict.

  Error messages

  java.lang.IllegalStateException: Failed to send [...]
      at org.springframework.web.servlet.mvc.method.annotation.ResponseBodyEmitter.sendInternal(ResponseBodyEmitter.java:259)
      at org.springframework.web.servlet.mvc.method.annotation.ResponseBodyEmitterReturnValueHandler.handleReturnValue(...)

  Caused by: org.springframework.http.converter.HttpMessageConversionException: Type definition error: [simple type, class io.agentscope.core.event.AgentStartEvent]
      at org.springframework.http.converter.AbstractJacksonHttpMessageConverter.writeInternal(AbstractJacksonHttpMessageConverter.java:441)

  Caused by: tools.jackson.databind.exc.InvalidDefinitionException: Conflict between type id property 'type' and bean property with same name; consider using
  `JsonTypeInfo.As.EXISTING_PROPERTY` to avoid duplication
   at [No location information]
      at tools.jackson.databind.exc.InvalidDefinitionException.from(InvalidDefinitionException.java:80)
      at tools.jackson.databind.ser.BeanSerializerFactory._verifyNoTypeIdPropertyConflict(BeanSerializerFactory.java:838)
      at tools.jackson.databind.ser.BeanSerializerFactory.constructBeanOrAddOnSerializer(BeanSerializerFactory.java:345)
      ...

  The key line is:

  Conflict between type id property 'type' and bean property with same name; consider using JsonTypeInfo.As.EXISTING_PROPERTY to avoid duplication

  Environment

  - AgentScope-Java Version: 2.0.0-RC3
  - Spring Boot Version: 4.1.0
  - Jackson Version: Jackson 3 (tools.jackson.databind)
  - Java Version: 21
  - OS: macOS

  Additional context

  - The endpoint is consumed through Spring MVC's ResponseBodyEmitterReturnValueHandler, which uses the Spring-managed ObjectMapper (Jackson 3). The exception is raised at
  serialization time inside BeanSerializerFactory._verifyNoTypeIdPropertyConflict, before any bytes are written to the SSE stream.
  - This strongly suggests the AgentStartEvent type hierarchy uses @JsonTypeInfo(use = Id.NAME, include = As.PROPERTY, property = "type") (the default-style configuration) while a
  subclass also exposes a getter/setter or field literally named type, producing two distinct property nodes for the same JSON key type. Jackson 3 is stricter about this than
  Jackson 2 and rejects it at serializer construction.
  - Suggested fix: change include = As.PROPERTY to include = As.EXISTING_PROPERTY (and make sure the type field is populated), or remove/rename the conflicting bean property on the
  event subclass. A reproduction project can be provided if needed.

guslegend0510:
## AgentScope-Java Version

2.0.0-SNAPSHOT

## Description

Closes #1948.

This PR fixes `io.agentscope.core.event.AgentEvent` serialization failures under Jackson 3, especially when events are streamed through Spring MVC `SseEmitter` / `ResponseBodyEmitter`.

### Root cause

`AgentEvent` used:

```java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
```

At the same time, the event hierarchy exposes `getType()`. Under Jackson 3 this causes a serializer conflict between the synthetic type-id property and the bean property named `type`, which leads to:

`InvalidDefinitionException: Conflict between type id property 'type' and bean property with same name`

### Changes

- Change `AgentEvent` polymorphic type handling from `JsonTypeInfo.As.PROPERTY` to `JsonTypeInfo.As.EXISTING_PROPERTY`
- Keep the existing `getType()` contract as the single source of the serialized `type` field
- Add regression tests for:
  - `AgentStartEvent` serialization containing exactly one `type` field
  - polymorphic deserialization from `AgentEvent` base type

### How to test

Run targeted regression tests:

```bash
mvn -pl agentscope-core test -Dtest=AgentEventStreamTest
```

I also ran:

```bash
mvn -pl agentscope-core test
```

The event/Jackson-related tests passed. Two existing Windows-specific symlink permission tests failed in my environment and appear unrelated to this change:

- `DangerousPathBypassTest.symlinkToDotEnvIsDetected`
- `DangerousPathBypassTest.symlinkToSshIsDetected`

## Checklist

- [x] All relevant tests are passing (`mvn -pl agentscope-core test -Dtest=AgentEventStreamTest`)
- [x] Related documentation has been updated (not needed for this internal serialization fix)


agentscope-core/src/test/java/io/agentscope/core/event/AgentEventStreamTest.java
```diff
@@ -39,6 +41,60 @@
...
+        @Test
+        @DisplayName("AgentStartEvent serializes with a single type property")
+        void agentStartEventSerializesWithSingleTypeProperty() throws Exception {
+            AgentStartEvent event = new AgentStartEvent("session-1", "reply-1", "assistant");
+
+            String serialized = mapper.writeValueAsString(event);
+            JsonNode json = mapper.readTree(serialized);
+
+            assertEquals("AGENT_START", json.path("type").asText());
+            assertEquals(7, json.size());
```
zouyx: 未来给事件对象加了新字段，这个断言会误报。

zouyx (approved):

agentscope-core/src/test/java/io/agentscope/core/event/AgentEventStreamTest.java
guslegend0510: Good catch. I removed the `json.size()` assertion since it couples the test to the current field count and would become brittle if new fields are added later.

The test now only verifies the actual contract we care about here:
- `type` is serialized as `AGENT_START`
- the serialized payload contains exactly one `type` field
- the key event fields are serialized correctly

itxaiohanglover (approved):
Clean fix. Switching from `As.PROPERTY` to `As.EXISTING_PROPERTY` resolves the Jackson conflict between the type-id property and the bean property named `type`. The test coverage for both serialization (single type property) and deserialization (polymorphic) looks solid.

zouyx (approved):
LGTM

AgentScopeJavaBot (approved):
🤖 AI Review

Fixes Jackson 3 type property conflict by changing @JsonTypeInfo to As.EXISTING_PROPERTY. Minimal precise fix with excellent tests. No issues found.

PR #1959, opened 2026-06-29
fix(model): disable native structured output for DashScope (#1852)

zhangliyuangit:
## AgentScope-Java Version

main (`2.0.0-SNAPSHOT`)

### Background
When using structured output (`ReActAgent.call(msgs, MyClass.class, ctx)` or the
JsonNode schema overload) with `DashScopeChatModel`, the returned
`Msg.hasStructuredData()` was always `false` and `getStructuredData(...)` returned
nothing. The same code works on `OpenAIChatModel`.

### Root cause
`DashScopeChatModel.supportsNativeStructuredOutput()` returned `true`, so `ReActAgent`
took the **native** structured-output path, which relies on `response_format` carrying
the JSON Schema. But the DashScope native text-generation protocol
(`/api/v1/services/aigc/text-generation/generation`) only supports
`response_format: {"type":"json_object"}` and **not** `{"type":"json_schema"}`. The
schema was never honored, the model returned natural-language text,
`wrapNativeStructuredResult` failed to parse it as JSON (logged a WARN), and no
`STRUCTURED_OUTPUT` metadata was attached.

> Note: simply wiring `response_format` through (mirroring the OpenAI channel) does
> **not** fix this — DashScope's native endpoint does not accept `json_schema`, so it
> would be ignored or rejected with HTTP 400.

### Changes
- `supportsNativeStructuredOutput()` now returns `false`, routing structured-output
  calls to the **tool-based fallback path** (`doFallbackStructuredCall`), which uses
  Qwen function-calling and returns reliable, schema-conforming results.
  `supportsNativeStructuredOutputWithTools()` stays consistent since it falls back to
  this method.
- Added Javadoc documenting why native structured output is unsupported on the
  DashScope native protocol (and a pointer not to flip it back without proper
  `json_object` wiring).
- Added a regression test asserting both flags are `false`.

### How to test
`ReActAgent.call(msgs, SomeClass.class, ctx)` on a `DashScopeChatModel` now returns a
`Msg` with `hasStructuredData() == true` and the parsed object available via
`getStructuredData(...)`.

Unit test: `DashScopeChatModelTest#testNativeStructuredOutputUnsupported`.

## Checklist

- [x] All tests are passing (`DashScopeChatModelTest`: 54 passed)

### DIFF
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeChatModel.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeChatModel.java
@@ -390,9 +390,33 @@ public class DashScopeChatModel extends ChatModelBase {
         return modelName;
     }
 
+    /**
+     * DashScope does not support native schema-constrained structured output.
+     *
+     * <p>The DashScope native text-generation protocol
+     * ({@code /api/v1/services/aigc/text-generation/generation}) only accepts
+     * {@code response_format: {"type": "json_object"}} and does <b>not</b> support
+     * {@code {"type": "json_schema", ...}}. Sending a json_schema payload is either
+     * ignored or rejected with HTTP 400. Because the native structured-output path
+     * ({@link io.agentscope.core.ReActAgent}) relies on {@code response_format} carrying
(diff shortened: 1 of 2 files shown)


itxaiohanglover (approved):
Well-documented fix. The Javadoc clearly explains why DashScope can't support native structured output and how the fallback path works. The test verifying both `supportsNativeStructuredOutput()` and `supportsNativeStructuredOutputWithTools()` returning false is exactly right.

jujn:
Thanks for the PR! This has already been fixed in # https://github.com/agentscope-ai/agentscope-java/pull/1935.

PR #1964, opened 2026-06-29
fix(harness): fix concurrent sandbox isolation in multi-user singleton deployments

### LINKED ISSUE #1896: [Bug]:SandboxBackedFilesystem 并发多用户不安全 &&框架后台任务用 RuntimeContext.empty() 写文件系统，破坏多用户路由
---
Bug 1：SandboxBackedFilesystem 并发多用户不安全

Title: SandboxBackedFilesystem is not safe for concurrent multi-user agents (single volatile field)

## Problem

`SandboxBackedFilesystem` uses a single `volatile Sandbox sandbox` field injected per-call
via `setSandbox()`. This design implicitly assumes one agent instance serves one call at a time.

In a multi-tenant setup (one singleton `HarnessAgent` serving multiple concurrent users),
concurrent calls overwrite each other's sandbox field, causing cross-user sandbox contamination:

- User A's call sets `sandbox = sandboxA`
- User B's call sets `sandbox = sandboxB`
- User A's tool execution now runs inside sandboxB

## Expected Behavior

`getSandbox()` should route to the correct sandbox based on the current `RuntimeContext`
(userId / sessionId), not from a shared mutable field.

## Workaround (current)

We had to implement a custom `RoutingFilesystem` that overrides `getSandbox()` to look up
a per-user `SandboxPool` via `RuntimeContext`. This is hundreds of lines of boilerplate that
reimplements what `IsolationScope.USER` promises to provide.

## Environment

- agentscope-harness: v2.0.0-RC4
- IsolationScope.USER is documented as supported, but the implementation still uses
  the single volatile field pattern

## Proposed Fix

Make `getSandbox()` accept a `RuntimeContext` parameter (or use a ThreadLocal bridge)
and route to the correct sandbox slot based on userId/sessionId, consistent with what
`IsolationScope` declares.

---
Bug 2：框架后台任务用 RuntimeContext.empty() 写文件系统，破坏多用户路由

Title: Framework background tasks (ConversationCompactor, WorkspaceMessageBus, etc.)
       use RuntimeContext.empty(), making per-user filesystem routing impossible

## Problem

Framework-internal components run asynchronously with `RuntimeContext.empty()` (no userId,
no sessionId):

- `ConversationCompactor` — compacts session history
- `WorkspaceMessageBus`
- `AsyncToolRegistry`
- `SubagentsMiddleware`

In the framework's single-sandbox model this is fine — there's only one sandbox per call.
But in a multi-user setup where filesystem routing depends on `RuntimeContext.userId`,
empty context cannot be routed to the correct user's sandbox.

## Impact

These writes fall through to the host-local filesystem (pod-local), which:
1. Makes pods stateful — session files are lost on pod restart
2. Triggers spurious workspace re-syncs into sandbox on every Compactor write
3. Produces noisy "local workspace changed, re-syncing" logs on every turn

## Root Cause

Background tasks are scheduled on Reactor's `boundedElastic` scheduler without carrying
user context. `InheritableThreadLocal` doesn't propagate through thread pool reuse.

## Proposed Fix

Options (in order of preference):
1. Have the framework propagate session context to background tasks via
   `Schedulers.onScheduleHook()` or Reactor `Context`
2. Expose a hook for background task context injection
3. Document that background tasks require a "system sandbox" slot and make the routing
   explicit rather than silently falling back to host FS

Buktal:
## AgentScope-Java Version

2.0.0-SNAPSHOT

## Description

Fixes two related concurrency bugs in `HarnessAgent` when used as a singleton serving concurrent sessions (the standard multi-tenant deployment pattern documented in the class Javadoc and official examples).

**Bug 1 — `SandboxBackedFilesystem` cross-session contamination (`#1896`)**

`SandboxBackedFilesystem` held a single `volatile Sandbox sandbox` field, and `SandboxLifecycleMiddleware` held a single `AtomicReference<SandboxAcquireResult>`. Concurrent calls from different sessions overwrote each other's bindings — User A's tools could execute inside User B's sandbox.

Fix: replace both single-value fields with `ConcurrentHashMap<sessionId, ...>`. `SandboxAware` interface updated: `setSandbox`/`getSandbox` replaced by `bindSandbox(key, sandbox)`, `unbindSandbox(key)`, and `getSandbox(key)`. The three public methods on `SandboxBackedFilesystem` already receive `RuntimeContext`, so the session key is always available at call time.

**Bug 2 — `WorkspaceMessageBus` / `WorkspaceAsyncToolRegistry` using sandbox filesystem**

`HarnessAgent.build()` constructed both components _after_ replacing `filesystem` with `SandboxBackedFilesystem`. This caused their internal `RuntimeContext.empty()` file operations to call `requireSandbox(null)`, which would throw or contaminate sandbox state.

Fix: capture `busFilesystem` before the sandbox replacement so bus and registry always use the non-sandbox (local or remote) filesystem. These are process-scoped infrastructure components with their own path-based isolation; they must not go through the sandbox.

**Testing**

- `SandboxBackedFilesystemTest`: updated existing 3 tests to new API; added 3 new tests including a concurrent 8-session isolation test using `CountDownLatch`
- `CompactionMiddlewareTest`: already existed on the `feat/compaction-events` branch; no changes needed for this fix

Closes #1896

## Checklist


itxaiohanglover (commented):
Good architectural fix — separating bus/registry filesystem from the sandbox filesystem prevents accidental routing through SandboxBackedFilesystem. The per-session `ConcurrentHashMap` approach for sandbox bindings looks correct for concurrent isolation. One question: is there a cleanup path for stale session entries when a session ends, or do entries persist for the proxy's lifetime?

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
```diff
@@ -77,18 +78,20 @@ public void acquireForCall(RuntimeContext ctx) {
         if (sandboxContext == null) {
             return;
         }
+        String sessionKey = ctx.getSessionId();
```
jujn: The active sandbox binding is keyed only by `ctx.getSessionId()`. In a singleton multi-tenant deployment, different users can run concurrently with the same session id (for example `default`). ReActAgent serializes by `(userId, sessionId)`, so those calls are allowed to overlap, but this map will overwrite `default -> sandboxA` with `default -> sandboxB`. That can still route User A's filesystem/tool calls into User B's sandbox.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
```diff
@@ -168,11 +175,14 @@ public List<FileDownloadResponse> downloadFiles(
         return results;
     }
 
-    private Sandbox requireSandbox() {
-        Sandbox s = sandbox;
+    private Sandbox requireSandbox(RuntimeContext rc) {
+        String key = rc != null ? rc.getSessionId() : null;
+        Sandbox s = key != null ? activeSandboxes.get(key) : null;
```
jujn: The same problem

jujn (changes requested):

jujn (changes requested):
The sandbox is still acquired in HarnessAgent.wrappedCall / wrappedStreamEvents before `delegate.call()` enters ReActAgent's per-slot serialization gate. If two calls for the same `(userId, sessionId)` are subscribed concurrently, both can run `acquireForCall()` first and write the same binding key. The second acquire overwrites the first sandbox/acquire result; the first call may then execute against the second sandbox and its cleanup can persist/release the wrong result, leaving the first sandbox leaked and the second call without a binding.

jujn (approved):
LGTM.
Can we quickly fix the conflict ? Thanks.

Buktal:
> LGTM. Can we quickly fix the conflict ? Thanks.

Resolved, thanks!

chickenlj:
@Buktal I have sent my Dingtalk Id to you email address [1171971708@qq.com](mailto:1171971708@qq.com), thanks.

Buktal:
> @Buktal I have sent my Dingtalk Id to you email address [1171971708@qq.com](mailto:1171971708@qq.com), thanks.

Got it, thanks

chickenlj:
核心修复方向是对的——`ConcurrentHashMap` 替换 `volatile` 单字段解决跨 session 污染，`busFilesystem` 在沙箱替换前捕获让基础设施组件不走沙箱路由。

不过 `serializedCall` / `serializedFlux` + `sandboxGates` 这套同 session 并发序列化门可以先去掉。同一个 session 的并发请求应该在业务入口层就阻止，不需要在 sandbox 层再做一层排队。这部分逻辑比较复杂（Sinks gate chain + conditional remove），加进来增加了理解和维护成本，当前没有实际场景需要它。

建议保持原来 `Mono.using` / `Flux.using` 的 acquire-release 模式，只把内部的 acquire/release 改成按 sessionKey 操作就行。

另外一个小点：`bindingKey()` 在 `SandboxBackedFilesystem` 和 `SandboxLifecycleMiddleware` 里各有一份相同实现，如果只改了一边会出 bug，建议提取到一个共享位置。

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
```diff
@@ -91,6 +94,30 @@ public void acquireForCall(RuntimeContext ctx) {
...
+        }
+        doAcquire(ctx, sandboxContext, sessionKey);
+    }
+
+    /**
+     * Releases the sandbox acquired by {@link #acquireForCall}. Intended as the cleanup step of
+     * a {@code Mono.using}/{@code Flux.using} wrapper; failures are logged but do not propagate.
+     */
+    public void releaseForCall(RuntimeContext ctx) {
+        doRelease(ctx);
```
AgentScopeJavaBot, line 109: [minor] acquireResults.put() silently overwrites. Use putIfAbsent with logging to prevent sandbox process leaks if precondition is violated.

AgentScopeJavaBot (commented):
🤖 AI Review

Fixes concurrent sandbox isolation in multi-user singleton deployments. Changes from single volatile reference to per-sessionKey ConcurrentHashMap. Well-designed fix. Suggestion: use putIfAbsent for defensive programming.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
AgentScopeJavaBot: ✅ Verified as addressed in `9e008cb`: Code now uses putIfAbsent (line 139) and logs a warn message with sessionKey when a previous entry exists (lines 141-152), exactly matching the bot's suggestion.

AgentScopeJavaBot (approved):

yujianxin991 (approved):

PR #1981, opened 2026-07-01
fix(core): reuse classpath skill jar file systems

### LINKED ISSUE #1979: [Bug]: ClasspathSkillRepository failed to load skills in jar application

**Describe the bug**
ClasspathSkillRepository failed to load skills in jar application

**To Reproduce**
the second ClasspathSkillRepository instance throws FileSystemAlreadyExistsException cause by the file system has already been created by the first ClasspathSkillRepository instance. 

Image

**Expected behavior**
reuse FileSystem like this:

     
            if ("jar".equals(uri.getScheme())) {
                this.isJar = true;

                FileSystem fs;
                try {
                    fs = FileSystems.getFileSystem(uri);
                } catch (FileSystemNotFoundException e) {
                    fs = FileSystems.newFileSystem(uri, Collections.emptyMap());
                }
                this.fileSystem = fs;
                String schemeSpecificUriPath = uri.getSchemeSpecificPart();
                String actualResourcePath =
                        schemeSpecificUriPath.substring(schemeSpecificUriPath.lastIndexOf("!") + 1);
                logger.info("Actual resource path: {}", actualResourcePath);
                this.skillBasePath = fileSystem.getPath(actualResourcePath);
            } else {

**Error messages**
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [io.agentscope.core.skill.SkillBox]: Factory method 'skillBox' threw exception with message: Failed to load skills from classpath
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:178)
	at org.springframework.beans.factory.support.ConstructorResolver.instantiate(ConstructorResolver.java:644)
	... 130 common frames omitted
Caused by: java.lang.IllegalStateException: Failed to load skills from classpath
	at com.ruoyi.agent.core.config.SkillBoxConfig.skillBox(SkillBoxConfig.java:34)
	at com.ruoyi.agent.core.config.SkillBoxConfig$$SpringCGLIB$$0.CGLIB$skillBox$0(<generated>)
	at com.ruoyi.agent.core.config.SkillBoxConfig$$SpringCGLIB$$FastClass$$1.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invokeSuper(MethodProxy.java:258)
	at org.springframework.context.annotation.ConfigurationClassEnhancer$BeanMethodInterceptor.intercept(ConfigurationClassEnhancer.java:348)
	at com.ruoyi.agent.core.config.SkillBoxConfig$$SpringCGLIB$$0.skillBox(<generated>)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:146)
	... 131 common frames omitted
Caused by: java.nio.file.FileSystemAlreadyExistsException: null
	at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.newFileSystem(Unknown Source)
	at java.base/java.nio.file.FileSystems.newFileSystem(Unknown Source)
	at java.base/java.nio.file.FileSystems.newFileSystem(Unknown Source)
	at io.agentscope.core.skill.repository.ClasspathSkillRepository.<init>(ClasspathSkillRepository.java:142)
	at io.agentscope.core.skill.repository.ClasspathSkillRepository.<init>(ClasspathSkillRepository.java:93)
	at com.ruoyi.agent.core.config.SkillBoxConfig.skillBox(SkillBoxConfig.java:27)
	... 139 common frames omitted

- AgentScope-Java Version: 2.0.0-RC4
- Java Version: 21
- OS: [e.g. macos, windows]

**Additional context**

guslegend0510:
## Summary

Fixes #1979.

`ClasspathSkillRepository` failed in jar deployments when the same classpath skill location was initialized more than once in the same JVM. The second repository instance could throw `FileSystemAlreadyExistsException` because the jar `FileSystem` was always created with `FileSystems.newFileSystem(...)`.

This change makes jar-backed classpath repositories reuse the same mounted `FileSystem` and only close it when the last repository referencing it is released.

## Root Cause

In jar environments, `ClasspathSkillRepository` created a new ZipFS for every instance, even when the target jar URI had already been mounted. That caused:
- `FileSystemAlreadyExistsException` on repeated initialization
- unsafe close behavior if multiple repositories shared the same jar-backed resource lifecycle

## Changes

- reuse existing jar `FileSystem` instances by URI
- add reference counting for jar-backed shared file systems
- only close jar `FileSystem` when the last repository releases it
- keep non-jar classpath behavior unchanged
- add regression tests for repeated repository initialization

## Tests

Ran:

```bash
mvn -pl agentscope-core -Dtest=ClasspathSkillRepositoryTest test
```
Added coverage for:
repeated ClasspathSkillRepository creation against the same normal jar
repeated ClasspathSkillRepository creation against the same Spring Boot nested jar
repository close ordering, ensuring one repository closing does not break another still using the same shared jar FileSystem


agentscope-core/src/main/java/io/agentscope/core/skill/repository/ClasspathSkillRepository.java
```diff
@@ -139,7 +147,8 @@ protected ClasspathSkillRepository(String resourcePath, String source, ClassLoad
 
             if ("jar".equals(uri.getScheme())) {
                 this.isJar = true;
-                this.fileSystem = FileSystems.newFileSystem(uri, Collections.emptyMap());
+                this.jarFileSystemUri = uri;
+                this.fileSystem = acquireFileSystem(uri);
```
jujn: The shared filesystem is keyed by the full resource URI, including the path after `!`.
﻿
If the same JAR contains two different classpath roots, e.g. `jar-skills-a` and `jar-skills-b`, two repositories will create two different map entries, but both entries can point to the same underlying ZipFS. Closing one repository may then close the ZipFS while the other repository is still using it.

jujn (changes requested):

agentscope-core/src/main/java/io/agentscope/core/skill/repository/ClasspathSkillRepository.java
guslegend0510: Updated to normalize the shared ZipFS key to the underlying jar file-system URI rather than the full resource URI.

Also added a regression test for two different classpath roots in the same jar and confirmed the repository close order is safe.

jujn (approved):
LGTM

PR #1990, opened 2026-07-02
fix(core): promote interrupt(RuntimeContext) to Agent interface to fix slot-mismatch silent failure

### LINKED ISSUE #1988: [Bug]: agent.interrupt() is a silent no-op in MiddlewareBase callbacks when the call carries a RuntimeContext

## Describe the bug

When a middleware (e.g., `MiddlewareBase.onModelCall`) calls `agent.interrupt()` on an `Agent` instance, the interruption has no effect on the current agent call if that call was invoked with a `RuntimeContext` whose `sessionId` differs from the agent's `defaultSessionId`. This is a **silent failure** — no error is thrown, no warning is logged, and the agent continues executing as if no interruption was requested.

The root cause is that `Agent.interrupt()` (inherited from `AgentBase`) delegates to `ReActAgent.interrupt(null, defaultSessionId)`, which triggers the interrupt signal on the **default session's slot**, while `checkInterrupted()` checks the **current call's slot** (determined by `RuntimeContext.sessionId`). When these two slots differ, the signal goes to the wrong `InterruptControl` instance, and the current call never sees it.

## To Reproduce

Minimal reproduction:

\`\`\`java
// 1. Create a ReActAgent with a default session ID
    .name("TestAgent")
    .model(mockModel)
    .defaultSessionId("default-session")
    .middleware(new MiddlewareBase() {
        @Override
        public Flux<AgentEvent> onModelCall(
                Agent agent, RuntimeContext ctx, ModelCallInput input,
            // This should interrupt the current call, but doesn't
            agent.interrupt();
            return next.apply(input);
        }
    })
    .build();

// 2. Invoke with a DIFFERENT sessionId via RuntimeContext
    .sessionId("my-session")  // ≠ "default-session"
    .build();

// 3. Observe: the agent does NOT interrupt
Msg result = agent.call(List.of(new UserMessage("test")), ctx).block();
// Expected: InterruptedException
// Actual: call completes normally (silent failure)
\`\`\`

**Why this happens**:
- `agent.interrupt()` → `ReActAgent.interrupt()` → `interrupt(null, "default-session")` → triggers `InterruptControl` for slot `(__anon__, default-session)`
- `call(msgs, ctx)` → `activateSlotForContext(ctx)` → uses slot `(__anon__, my-session)` → `checkInterrupted()` checks **that** slot's `InterruptControl`
- Two different `InterruptControl` instances → signal never seen by the current call

**Current workaround** (as seen in `SubAgentTool.java:200-229`):
\`\`\`java
if (agent instanceof ReActAgent ra) {
    ra.interrupt(ctx);  // ← forced to cast and call non-interface method
}
\`\`\`

## Expected behavior

One of:
1. `agent.interrupt()` should interrupt the current call, regardless of which session the call is running on
2. The `Agent` interface should provide a session-aware interrupt method (e.g., `interrupt(RuntimeContext)`) so middlewares can interrupt the current call without casting
3. The framework should log a warning or throw an exception when `interrupt()` is called but targets the wrong session

## Error messages

**None.** This is a silent failure — the most dangerous kind. No exception, no log line, no indication that the interruption was ignored.

## Environment

- **AgentScope-Java Version**: 2.0.0-SNAPSHOT (main branch as of 2026-07-02)
- **Java Version**: 17
- **OS**: Windows 11 Pro 10.0.26200

## Additional context

**Root cause**: Slot mismatch between `interrupt()` (hardcoded `defaultSessionId`) and `checkInterrupted()` (current call's `sessionId`).

**Evidence that this is a known issue**:
`SubAgentTool.java:200-229` already works around this by casting to `ReActAgent` and calling the non-interface `interrupt(RuntimeContext)` method:
\`\`\`java
// Uses interrupt(RuntimeContext) instead of the deprecated interrupt(InterruptSource)
// because the latter only looks up [the default session] ...
private void interruptAgent(Agent agent, RuntimeContext ctx) {
    if (agent instanceof ReActAgent ra) {
        ra.interrupt(ctx);
    }
}
\`\`\`

**Full root cause analysis and API redesign proposal**: See Enhancement issue (will be linked after submission).

### LINKED ISSUE #1989: [Feature]: Promote interrupt(RuntimeContext) to the Agent interface so middlewares/tools can interrupt without casting to ReActAgent

## Is your feature request related to a problem? Please describe.

`MiddlewareBase` callbacks (e.g., `onModelCall`, `onReasoning`, `onActing`) receive both `Agent agent` and `RuntimeContext ctx`, but the `Agent` interface only provides:
- `void interrupt()` — hardcoded to the agent's `defaultSessionId` (see Bug #1988)
- `void interrupt(Msg)` — same slot-mismatch issue

The session-aware methods that actually work (`interrupt(RuntimeContext)` / `interrupt(String userId, String sessionId, Msg)`) are **only on `ReActAgent`**, not on the `Agent` interface. This forces middleware/tool authors to:
1. Cast to `ReActAgent` (breaks interface-oriented design)
2. Live with silent failures when `agent.interrupt()` targets the wrong session

**Evidence**: `SubAgentTool.java:200-229` already does this workaround:
\`\`\`java
private void interruptAgent(Agent agent, RuntimeContext ctx) {
    if (agent instanceof ReActAgent ra) {
        ra.interrupt(ctx);  // ← can't call this via Agent interface
    }
}
\`\`\`

**Additionally**, `HarnessAgent` does not delegate `interrupt(RuntimeContext)`:
\`\`\`java
// HarnessAgent.java:482-489
@Override public void interrupt() { delegate.interrupt(); }
@Override public void interrupt(Msg msg) { delegate.interrupt(msg); }
// ❌ missing: interrupt(RuntimeContext ctx)
\`\`\`

So even if a middleware casts to `ReActAgent`, it won't work through a `HarnessAgent` wrapper.

## Describe the solution you'd like

**Proposed API (backward-compatible)**:

1. **Add to `Agent` interface** (with default delegation to preserve compatibility):
\`\`\`java
// Agent.java
public interface Agent ... {
    @Deprecated void interrupt();
    @Deprecated void interrupt(Msg msg);

    // NEW — default delegates to legacy for compatibility
    default void interrupt(RuntimeContext ctx) { interrupt(); }
    default void interrupt(RuntimeContext ctx, Msg msg) { interrupt(msg); }
}
\`\`\`

2. **Override in `ReActAgent`** with real per-session logic:
\`\`\`java
// ReActAgent.java
@Override
public void interrupt(RuntimeContext ctx) {
    interrupt(ctx, null);
}

@Override
public void interrupt(RuntimeContext ctx, Msg msg) {
    String uid = ctx != null ? ctx.getUserId() : null;
    String sid = ctx != null ? ctx.getSessionId() : null;
    if (sid == null || sid.isBlank()) {
        sid = defaultSessionId;
    }
    getAgentState(uid, sid).interruptControl().trigger(InterruptSource.USER, msg);
}
\`\`\`

3. **Add missing delegate in `HarnessAgent`**:
\`\`\`java
// HarnessAgent.java
@Override
public void interrupt(RuntimeContext ctx) { delegate.interrupt(ctx); }

@Override
public void interrupt(RuntimeContext ctx, Msg msg) { delegate.interrupt(ctx, msg); }
\`\`\`

4. **Clean up** (separate PR or follow-up):
   - Mark legacy \`interrupt()\` / \`interrupt(Msg)\` as \`@Deprecated(forRemoval = true)\`
   - Remove other deprecated interrupt-related methods (\`interrupt(InterruptSource)\`, \`checkInterruptedAsync\`, \`resetInterruptFlag\`, etc.)
   - Update Javadoc to point users at the new methods

**Result**: Middlewares can write \`agent.interrupt(ctx)\` without casting, and it works correctly for all agent types.

## Describe alternatives you've considered

**Alternative A: New sub-interface \`InterruptibleAgent\`**
\`\`\`java
public interface InterruptibleAgent extends Agent {
    void interrupt(RuntimeContext ctx);
    void interrupt(RuntimeContext ctx, Msg msg);
}
\`\`\`
- **Pro**: Clean separation of concerns (SOLID)
- **Con**: Middleware usage still requires \`instanceof InterruptibleAgent\` checks (better than casting, but still a branch)
- **Con**: Adds a new interface to the API surface

**Alternative B: Bypass the \`Agent\` interface entirely**
\`\`\`java
// Direct access to the per-session interrupt signal
ctx.getAgentState().interruptControl().trigger(InterruptSource.USER, msg);
\`\`\`
- **Pro**: Zero interface changes
- **Con**: Exposes \`InterruptControl\` internals (leaky abstraction)
- **Con**: Doesn't fix the misleading \`Agent\` contract (users will still think \`interrupt()\` works)

**Why we prefer the proposed solution**:
- Backward-compatible (default methods preserve existing behavior)
- Eliminates casting (middlewares use \`agent.interrupt(ctx)\` directly)
- Works uniformly across \`ReActAgent\` and \`HarnessAgent\`
- Minimal API churn (just adds overloads, doesn't break anything)

## Additional context

**The six ways the current interrupt API is messy**:

1. **Interface contract vs. implementation mismatch**: \`Agent.interrupt()\` Javadoc promises sets an interrupt flag that will be checked at appropriate checkpoints, but \`AgentBase\` implements it as \`@Deprecated {}\` (empty method)
2. **Working methods not on the interface**: \`interrupt(RuntimeContext)\` / \`interrupt(userId, sessionId)\` only exist on \`ReActAgent\`, not \`Agent\`
3. **No-arg version has wrong semantics**: Hardcoded \`defaultSessionId\` causes silent failures in multi-session scenarios
4. **\`HarnessAgent\` incomplete delegation**: Missing \`interrupt(RuntimeContext)\`, so casting doesn't even work through the wrapper
5. **Middleware has \`ctx\` but must cast**: \`onModelCall(Agent agent, RuntimeContext ctx, ...)\` has both ingredients, but can't use them without \`instanceof ReActAgent\`
6. **Deprecated method pile-up**: \`interrupt()\`, \`interrupt(Msg)\`, \`interrupt(InterruptSource)\`, \`checkInterruptedAsync\`, \`resetInterruptFlag\`, \`getInterruptFlag\`, \`getInterruptSource\` — all deprecated, high cognitive load

**Related**: See Bug #1988 for a concrete reproduction of the silent-failure symptom.

Buktal:
## AgentScope-Java Version

2.0.0-SNAPSHOT

## Description

**Problem**: The `Agent.interrupt()` and `Agent.interrupt(Msg)` methods were hardcoded to the agent's `defaultSessionId`. When a call was made with a `RuntimeContext` carrying a different `sessionId`, calling `agent.interrupt()` from a middleware triggered the **wrong session's `InterruptControl`** — a silent no-op that neither threw nor logged anything. Middlewares holding both `Agent agent` and `RuntimeContext ctx` (e.g. `onModelCall`, `onReasoning`) were forced to cast `((ReActAgent) agent).interrupt(ctx)`, as seen in `SubAgentTool.java:200-229`.

**Solution**: Add `interrupt(RuntimeContext)` and `interrupt(RuntimeContext, Msg)` as default methods on the `Agent` interface. `ReActAgent` already had these methods with real per-session trigger logic — now they properly `@Override` the interface methods. `HarnessAgent` gains the missing delegates so the methods work through the harness wrapper.

**Changes**:

| File | What changed |
|---|---|
| `Agent.java` | New `default interrupt(RuntimeContext)` / `interrupt(RuntimeContext, Msg)`; mark old methods `@Deprecated(forRemoval=true)` |
| `AgentBase.java` | Upgrade Javadoc on all legacy interrupt methods, add `forRemoval=true` |
| `ReActAgent.java` | Add `@Override` to existing `interrupt(RuntimeContext)` / `interrupt(RuntimeContext, Msg)` |
| `HarnessAgent.java` | Add missing `interrupt(RuntimeContext)` / `interrupt(RuntimeContext, Msg)` delegates → `delegate` |

**How to test**: Middlewares/tools can now write `agent.interrupt(ctx)` without casting — it correctly targets the session identified by the `RuntimeContext`. The change is backward-compatible (default methods delegate to legacy `interrupt()` for existing implementations).

Closes #1988, Closes #1989

## Checklist

- [x] All tests are passing (`agentscope-core` — all 300+ tests pass; `agentscope-harness` — 6 pre-existing `JUnit Failed to close extension context` errors unrelated to this change)


AgentScopeJavaBot (commented):
🤖 AI Review

Promotes interrupt(RuntimeContext) to Agent interface to fix slot-mismatch silent failure. The interrupt method signature now includes RuntimeContext for proper session key resolution. Important interface-level change ensuring interrupts target the correct session slot.

PR #2001, opened 2026-07-03
fix(harness): propagate userId through wakeup dispatch

### LINKED ISSUE #2000: [BUG]: WakeupDispatcher drops userId, causing wakeup rounds to lose the original conversation in multi-tenant runtimes
## Problem

When a background task (async subagent / async tool) completes, `WakeupDispatcher` wakes the idle parent session to run another reasoning round. In runtimes that isolate agent state by `(userId, sessionId)`, this wakeup round fails with a provider 400:

```
{"code":20015,"message":"No user query found in messages."}
```

Observed in logs: the wakeup round starts with `user=null` and an almost-empty context (`contextMsgs=1`, only the system prompt), because state is loaded from the wrong slot.

## Root cause

`MessageBus.enqueueWakeup(userId, sessionId, agentId)` already stores `userId` in the wakeup queue entry:

```java
default Mono<Void> enqueueWakeup(String userId, String sessionId, String agentId) {
    return queuePush("agentscope:wakeups",
            Map.of("userId", userId != null ? userId : "",
                   "sessionId", sessionId,
                   "agentId", agentId))
            .then(publish("agentscope:wakeup_signal", Map.of()));
}
```

But `WakeupDispatcher.dispatch` reads only `sessionId` and `agentId`, then calls `target.runWakeup(sessionId)` — the `userId` is dequeued and discarded. `WakeupTarget#runWakeup(String sessionId)` has no parameter to receive it.

`HarnessGateway.runWakeup` therefore builds the `RuntimeContext` without a userId:

```java
RuntimeContext rtc = RuntimeContext.builder()
        .sessionId(sessionId)
        .put("gateKey", gateKey)
        .put("outboundAddress", ...)
        .build();    // no userId
```

The wakeup round loads state from the anonymous `(null, sessionId)` slot instead of the original `(userId, sessionId)` slot — the original conversation (incl. the user's query) is absent, the model receives no user message, the provider returns 400.

`runStream` propagates userId correctly (`if (ctx.userId() != null && !ctx.isBlank()) rtcBuilder.userId(ctx.userId())`); only the wakeup path is broken.

## Impact

Any runtime that isolates agent state by `(userId, sessionId)` and relies on the wakeup mechanism (async subagents, async tools, future team messages / schedulers) cannot complete a wakeup-driven round. Single-tenant / anonymous runtimes are unaffected (the missing userId is a no-op there).

## Proposed fix

Backward-compatible default method on `WakeupTarget`:

- `WakeupTarget`: add `default Mono<Msg> runWakeup(String userId, String sessionId)` delegating to the legacy overload; mark `runWakeup(String sessionId)` `@Deprecated`.
- `WakeupDispatcher.dispatch`: read `userId` from the payload and call `runWakeup(userId, sessionId)`.
- `HarnessGateway`: override the new overload and set `rtc.userId(...)` when non-blank (matching `runStream`).

Existing `WakeupTarget` implementers stay source/binary compatible. Happy to open a PR.

hansiweicn-debug:
Resolves #2000.

## Summary
`WakeupDispatcher.dispatch` discarded the `userId` that `enqueueWakeup` had already stored in the wakeup queue. In runtimes isolating agent state by `(userId, sessionId)`, the wakeup round built a `RuntimeContext` without a userId, loaded state from an anonymous slot, lost the original conversation, and the provider returned `400 "No user query found in messages."` This forwards the userId end-to-end.

## Changes
- `WakeupTarget`: add `default runWakeup(userId, sessionId)` delegating to the legacy overload; mark `runWakeup(sessionId)` `@Deprecated`.
- `WakeupDispatcher.dispatch`: read `userId` from the wakeup payload and forward it to the target.
- `HarnessGateway`: override the new overload; set `rtc.userId(...)` when non-blank (matching `runStream`). Blank/null userId keeps the legacy anonymous behaviour — no regression for single-tenant runtimes.

## Backward compatibility
The single-arg `runWakeup(String)` stays on the interface (`@Deprecated`, with a default that delegates to it), so existing `WakeupTarget` implementations — including external ones — compile and behave as before. The dispatcher now calls the user-id-aware overload, which `HarnessGateway` overrides; any implementation that does not override it falls back via the default.

## Tests
- Added `WakeupDispatcherTest#wakeupDispatch_propagatesUserId`: enqueues a wakeup with a userId and asserts the target receives it.
- All existing `WakeupDispatcherTest` cases pass unchanged (StubTarget unchanged).

```
mvn test -pl agentscope-harness -Dtest=WakeupDispatcherTest
# Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
```

`HarnessGateway.runWakeup`'s userId->rtc assignment mirrors the production-proven `runStream` pattern (`if (!isBlank) rtcBuilder.userId(...)`); kept the PR focused without a dedicated gateway integration test, but happy to add one if reviewers want end-to-end rtc coverage. Spotless applied; only the three intended files are in the diff.

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/gateway/HarnessGateway.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/gateway/HarnessGateway.java
@@ -459,15 +459,38 @@ public final class HarnessGateway implements Gateway, WakeupDispatcher.WakeupTar
     }
 
     /**
-     * Triggers a wakeup-driven run for an idle session. Called by {@link WakeupDispatcher} when a
-     * background task completes or a team message arrives. The agent starts a reasoning round with
-     * no user input; {@link io.agentscope.harness.agent.middleware.InboxMiddleware} drains the
-     * inbox and injects the pending results as context.
+     * {@inheritDoc}
      *
+     * <p>Delegates to {@link #runWakeup(String, String)} with a {@code null} user id.
+     *
+     * @deprecated use {@link #runWakeup(String, String)} so the wakeup-driven round loads agent
+     *     state from the correct {@code (userId, sessionId)} slot. This overload preserves legacy
+     *     anonymous-wakeup behavior.
+     */
+    @Deprecated
+    @Override
+    public Mono<Msg> runWakeup(String sessionId) {
+        return runWakeup(null, sessionId);
+    }
+
+    /**
+     * Triggers a wakeup-driven run for an idle session, propagating the owning user id when
+     * present. Called by {@link WakeupDispatcher} when a background task completes or a team
+     * message arrives. The agent starts a reasoning round with no user input; {@link
+     * io.agentscope.harness.agent.middleware.InboxMiddleware} drains the inbox and injects the
+     * pending results as context.
+     *
+     * <p>When {@code userId} is non-blank it is set on the {@link RuntimeContext}, so the agent
+     * loads state from the same {@code (userId, sessionId)} slot as the original user-driven run
+     * — matching {@code runStream}. A blank or null user id preserves the legacy anonymous
+     * behavior.
+     *
+     * @param userId the owning user id carried by the wakeup entry; may be {@code null} or blank
      * @param sessionId the session to wake up
      * @return the agent's response, or {@link Mono#empty()} if the session is unknown
      */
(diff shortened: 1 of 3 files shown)


jujn (approved):
LGTM

PR #2004, opened 2026-07-03
feat: expose resumableStreams and openConnectionOnStartup on McpClientBuilder

### LINKED ISSUE #2003: [Bug]:McpClientBuilder.streamableHttpTransport does not expose resumableStreams option — hangs on SSE-disabled MCP servers
When connecting to a Streamable HTTP MCP server that disables SSE (only accepts POST requests), tool calls hang
  indefinitely. The MCP server is discovered and tools are registered successfully during initialization, but actual tool
  invocations (callTool) never return — PRE_ACTING fires without a corresponding POST_ACTING.

  The root cause is that McpClientBuilder.streamableHttpTransport(url) does not expose the underlying MCP SDK's
  HttpClientStreamableHttpTransport.Builder options resumableStreams and openConnectionOnStartup. Both default to true,
  meaning the transport always attempts to establish an SSE connection even when the server does not support it. Direct curl
  POST calls to the same server work correctly, confirming the issue is in the transport layer, not the server.

  To Reproduce

  1. Configure a Streamable HTTP MCP server that returns 405 Method Not Allowed for SSE requests (e.g. WeCom Document MCP):

  // tools.json
  {
    "mcpServers": {
      "example": {
        "transport": "streamableHttp",
        "url": "https://qyapi.weixin.qq.com/mcp/robot-doc?apikey=xxx"
      }
    }
  }

  2. Build the agent with HarnessAgent.builder()...build().
  3. Send a message that triggers an MCP tool call.
  4. Observe that PRE_ACTING (tool call started) appears in logs, but POST_ACTING (tool call completed) never fires — the
  call hangs.

  Minimal reproduction without HarnessAgent:

  McpClientWrapper client = McpClientBuilder.create("test")
      .streamableHttpTransport("https://sse-disabled-mcp-server.example.com/mcp")
      .buildSync();

  toolkit.registerMcpClient(client).block();

      .name("test").model(model).toolkit(toolkit).build();

  // This call will hang when the agent tries to invoke an MCP tool
  agent.call(new UserMessage("use a tool")).block();

  Expected behavior

  Tool calls complete successfully against SSE-disabled Streamable HTTP MCP servers.

  McpClientBuilder should expose the transport builder options: resumableStreams(false) and openConnectionOnStartup(false),
  either via:

  McpClientBuilder.create("test")
      .streamableHttpTransport(url)
      .resumableStreams(false)        // API option
      .buildSync();

  Or via tools.json:

  {
    "transport": "streamableHttp",
    "url": "https://...",
    "resumableStreams": false
  }

  Error messages

  During initialization (recovered, tools registered):
  io.modelcontextprotocol.spec.McpTransportException: Invalid SSE response. Status code: 405
  {"error":"Method Not Allowed","message":"SSE is disabled. Please use POST for /mcp/robot-doc","allowed_methods":["POST"]}

  During tool invocation (no error — just hangs forever):
  PRE_ACTING  | name=create_doc  ← fires
  POST_ACTING |                  ← never fires

  Environment

  - AgentScope-Java Version: 2.0.0-RC4
  - MCP Java SDK: 0.17.0 (transitive dependency)
  - Java Version: 17
  - OS: macOS

wkflash:
## Summary

  Expose `resumableStreams` and `openConnectionOnStartup` on `McpClientBuilder` for Streamable HTTP transport, with
  bidirectional call-order safety.

  Closes #2003.

  ## Problem

  When connecting to a Streamable HTTP MCP server that disables SSE and only accepts POST, tool calls hang indefinitely. The
  underlying MCP Java SDK (`HttpClientStreamableHttpTransport.Builder`) already supports these options, but AgentScope's
  builder did not expose them.

  ## Solution

  Two new fluent methods + `tools.json` config support:

  ### Java API

  ```java
  McpClientBuilder.create("example")
      .streamableHttpTransport("https://sse-disabled.example.com/mcp")
      .resumableStreams(false)
      .openConnectionOnStartup(false)
      .buildSync();

  tools.json

  {
    "mcpServers": {
      "my-server": {
        "transport": "http",
        "url": "https://example.com/mcp",
        "resumableStreams": false,
        "openConnectionOnStartup": false
      }
    }
  }

  Bidirectional safety

  Unlike the approach that only writes to the config object, this PR stores values in both a builder-level field AND syncs to
  the config if already present. Both call orders work:

  // order 1: transport first → instanceOf check writes to config

  // order 2: options first → builder field stored, applied when transport is created
  .resumableStreams(false).streamableHttpTransport(url)

  Tests

  - resumableStreams / openConnectionOnStartup after streamableHttpTransport → stored correctly
  - resumableStreams / openConnectionOnStartup before streamableHttpTransport → stored correctly
  - Called on SSE transport → silently ignored, type unchanged
  - Called before any transport → no NPE
  - Both options combined
  - JSON deserialization of resumableStreams / openConnectionOnStartup

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/mcp/McpClientBuilder.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/mcp/McpClientBuilder.java
@@ -103,6 +103,8 @@ public class McpClientBuilder {
     private Function<ElicitRequest, Mono<ElicitResult>> asyncElicitationHandler;
     private Function<ElicitRequest, ElicitResult> syncElicitationHandler;
     private List<String> protocolVersions;
+    private Boolean resumableStreams;
+    private Boolean openConnectionOnStartup;
 
     private McpClientBuilder(String name) {
         this.name = name;
@@ -189,7 +191,14 @@ public class McpClientBuilder {
      * @return this builder
      */
     public McpClientBuilder streamableHttpTransport(String url) {
-        this.transportConfig = new StreamableHttpTransportConfig(url);
+        StreamableHttpTransportConfig config = new StreamableHttpTransportConfig(url);
+        if (resumableStreams != null) {
+            config.resumableStreams = resumableStreams;
+        }
+        if (openConnectionOnStartup != null) {
+            config.openConnectionOnStartup = openConnectionOnStartup;
+        }
+        this.transportConfig = config;
         return this;
     }
 
@@ -218,6 +227,41 @@ public class McpClientBuilder {
         if (transportConfig instanceof StreamableHttpTransportConfig) {
             ((StreamableHttpTransportConfig) transportConfig).customizeHttpClient(customizer);
         }
+        return this;
+    }
+
+    /**
+     * Enables or disables resumable (SSE) streams for StreamableHTTP transport.
+     * Only applicable after calling {@link #streamableHttpTransport(String)}.
+     *
+     * <p>Set to {@code false} when connecting to Streamable HTTP MCP servers that
+     * disable SSE and only accept POST requests. Defaults to the SDK's built-in
+     * default ({@code true}) when not called.
+     *
+     * @param value {@code true} to enable SSE streams (default), {@code false} to disable
+     * @return this builder
+     */
+    public McpClientBuilder resumableStreams(boolean value) {
+        this.resumableStreams = value;
+        return this;
+    }
+
+    /**
+     * Controls whether the transport opens a persistent connection on startup
+     * for StreamableHTTP transport.
+     * Only applicable after calling {@link #streamableHttpTransport(String)}.
+     *
+     * <p>Set to {@code false} together with
+     * {@link #resumableStreams(boolean) resumableStreams(false)} when connecting
+     * to SSE-disabled servers. Defaults to the SDK's built-in default ({@code true})
+     * when not called.
+     *
+     * @param value {@code true} to open connection on startup (default),
+     *              {@code false} to skip
+     * @return this builder
+     */
+    public McpClientBuilder openConnectionOnStartup(boolean value) {
+        this.openConnectionOnStartup = value;
         return this;
     }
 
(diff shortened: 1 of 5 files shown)


agentscope-core/src/main/java/io/agentscope/core/tool/mcp/McpClientBuilder.java
```diff
@@ -189,7 +191,14 @@ public McpClientBuilder customizeSseClient(Consumer<HttpClient.Builder> customiz
...
     public McpClientBuilder streamableHttpTransport(String url) {
-        this.transportConfig = new StreamableHttpTransportConfig(url);
+        StreamableHttpTransportConfig config = new StreamableHttpTransportConfig(url);
+        if (resumableStreams != null) {
+            config.resumableStreams = resumableStreams;
+        }
+        if (openConnectionOnStartup != null) {
+            config.openConnectionOnStartup = openConnectionOnStartup;
+        }
+        this.transportConfig = config;
```
jujn, line 201:
```
McpClientBuilder.create("mcp")
.streamableHttpTransport(url)
.resumableStreams(false)
.openConnectionOnStartup(false)
```
In this order, resumableStreams(false) and openConnectionOnStartup(false) are ignored because they update only the builder-level fields, not the already-created StreamableHttpTransportConfig, leaving the config fields null and thus falling back to SDK defaults.
Recommend the following:
```
public McpClientBuilder resumableStreams(boolean value) {
    this.resumableStreams = value;
    if (transportConfig instanceof StreamableHttpTransportConfig config) {
        config.resumableStreams = value;
    }
    return this;
}

public McpClientBuilder openConnectionOnStartup(boolean value) {
    this.openConnectionOnStartup = value;
    if (transportConfig instanceof StreamableHttpTransportConfig config) {
        config.openConnectionOnStartup = value;
    }
    return this;
}
```

jujn (changes requested):

jujn:
Additionally, it is recommended to supplement corresponding tests

Issue #2009, opened 2026-07-04
[Bug]:RemoteSandboxSnapshot cannot be deserialized: sandbox state restore always fails for remote (OSS/S3) snapshots


arthurdu98:
Summary

  RemoteSandboxSnapshot can be serialized but cannot be deserialized by Jackson. As a result, restoring a persisted DockerSandboxState whose snapshot is a remote type always     
  throws, and SandboxManager.acquire() silently falls back to creating a fresh container — so the previous sandbox workspace is lost across calls whenever a remote (OSS/S3)      
  snapshot store is configured.

  This affects agentscope-harness 2.0.0-RC3 (also reproduced on RC1, RC2, RC4 — the relevant classes are byte-for-byte identical across these versions).

  Root cause

  SandboxSnapshot is a Jackson polymorphic type:

  @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
  @JsonSubTypes({
      @JsonSubTypes.Type(value = NoopSandboxSnapshot.class,  name = "noop"),
      @JsonSubTypes.Type(value = LocalSandboxSnapshot.class,  name = "local"),
      @JsonSubTypes.Type(value = RemoteSandboxSnapshot.class, name = "remote")
  })
  public interface SandboxSnapshot { ... }

  RemoteSandboxSnapshot only has a two-arg constructor and no Jackson-usable creator:

  public class RemoteSandboxSnapshot implements SandboxSnapshot {
      private final RemoteSnapshotClient client;
      private final String id;

      public RemoteSandboxSnapshot(RemoteSnapshotClient client, String id) { ... }
      // no no-arg constructor, no @JsonCreator, no @JsonProperty on the args
  }

  So Jackson can write the object (via getters) but has no way to construct it when reading. Because the failure only happens on the read path, it stays hidden until a sandbox   
  state is actually restored from the store.

  There are two related serialization issues visible in the produced JSON:

  1. No usable creator (the blocker). Deserialization of RemoteSandboxSnapshot fails outright.
  2. Discriminator field collision. The getType() getter serializes a type property that collides with the @JsonTypeInfo(property = "type") discriminator, so the emitted JSON    
  contains "type" twice. In strict mode this also breaks NoopSandboxSnapshot, which has no fields but ends up with extra serialized getter properties (id, persistenceEnabled,    
  ...).

  Steps to reproduce

[...]

PR #2013, opened 2026-07-05
fix(harness): restore remote snapshot state deserialization

guslegend0510:
## Summary

Fix remote sandbox snapshot state restore for `agentscope-harness`.

`RemoteSandboxSnapshot` could be serialized but not deserialized correctly, which caused persisted Docker sandbox state restore to fail whenever a remote snapshot store was configured. In practice, `SandboxManager.acquire()` would fall back to creating a fresh sandbox, silently losing the previous workspace state.

## Root Cause

There were two issues in the snapshot serialization flow:

1. `RemoteSandboxSnapshot` did not expose a Jackson-usable creator for deserialization.
2. Snapshot JSON included derived properties such as `type`, `restorable`, and `persistenceEnabled`, which polluted the payload and could collide with the polymorphic discriminator.

A second runtime problem also existed after deserialization:
`RemoteSnapshotClient` is a runtime dependency and should not be serialized, but it must be rebound when persisted state is restored.

## What Changed

- Added a Jackson-friendly deserialize path for `RemoteSandboxSnapshot` using `id` only.
- Excluded runtime-only / derived snapshot properties from JSON output.
- Added a `SandboxClient` overload that accepts `SandboxSnapshotSpec` during state deserialization.
- Updated `SandboxManager` to pass the current snapshot spec when restoring persisted sandbox state.
- Rebound `RemoteSnapshotClient` inside `DockerSandboxClient` when deserializing remote snapshots.
- Added regression tests for:
  - remote snapshot JSON shape
  - mapper deserialization
  - legacy JSON with extra derived fields
  - Docker client rebind behavior
  - `SandboxManager` snapshotSpec passthrough

## Impact

This restores the expected behavior for remote snapshot backed sandboxes:
persisted state can now be read back, and remote snapshot recovery no longer silently degrades to a fresh workspace because of snapshot deserialization failure.

## Validation

- `mvn -pl agentscope-harness -DskipTests spotless:check`

Note:
targeted test execution was blocked by pre-existing compile errors on current `upstream/main` in unrelated `agentscope-harness` / `agentscope-core` integration points, so validation here was limited to formatting and local change inspection.

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxClient.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxClient.java
@@ -41,4 +41,8 @@ public interface SandboxClient<O extends SandboxClientOptions> {
     String serializeState(SandboxState state);
 
     SandboxState deserializeState(String json);
+
+    default SandboxState deserializeState(String json, SandboxSnapshotSpec snapshotSpec) {
+        return deserializeState(json);
+    }
 }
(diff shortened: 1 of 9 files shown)


jujn:
related to https://github.com/agentscope-ai/agentscope-java/issues/2009

PR #2026, opened 2026-07-06
fix(openai): allow disabling native structured output

### LINKED ISSUE #2025: [Bug]:OpenAI native structured output cannot be disabled for compatible gateways
## Description

When using `ReActAgent` structured output with an OpenAI-compatible gateway, `OpenAIChatModel` always reports native structured output support.

For gateways or models that do not correctly support OpenAI native `response_format`, the native structured-output path can return a normal assistant message without structured data. Reading the structured result then fails because `_structured_output` is missing.

This prevents users from opting into AgentScope's non-native structured-output fallback flow, even when the target endpoint is OpenAI-compatible but does not support native structured output reliably.

RainYuY:
## AgentScope-Java Version

latest

## Description

Fixes #2025 

This PR makes OpenAI native structured-output capability configurable.

`OpenAIChatModel` previously returned `true` from `supportsNativeStructuredOutput()` unconditionally. That works for native OpenAI-compatible models that fully support `response_format`, but it breaks compatible gateways where native structured output is unavailable or unreliable.

This change adds a `nativeStructuredOutput` builder/provider option. By default, behavior remains unchanged. Users can explicitly set it to `false` to let `ReActAgent` use the fallback structured-output flow instead.


chickenlj:
Thanks, this has been fixe with #2011, please help to check.

RainYuY:
> #2011

I think this is a different issue, or did I get the numbering wrong?

PR #2037, opened 2026-07-07
fix(core): createSkillToolGroup appends _skill_tools suffix internally

### LINKED ISSUE #2034: createSkillToolGroup should internally append "_skill_tools" suffix to group name
### Problem

`ToolGroupManager.createSkillToolGroup()` uses the raw `groupName` parameter directly:

```java
// ToolGroupManager.java:129-138
public void createSkillToolGroup(
        String groupName, String description, boolean active, String activateOnSkill) {
    SkillToolGroup group = SkillToolGroup.skillBuilder()
            .name(groupName)          // ← raw name, no suffix
            .description(description)
            .active(active)
            .activateOnSkill(activateOnSkill)
            .build();
    registerToolGroup(group);
}
```

However, when a skill is activated, `RegisteredSkill.getToolsGroupName()` looks up the group with `skillId + "_skill_tools"`:

```java
// RegisteredSkill.java:74-76
public String getToolsGroupName() {
    return skillId + "_skill_tools";
}

// SkillToolFactory.java:373-384 - activation lookup
private void activateSkill(String skillId) {
    String toolsGroupName = skillRegistry.getRegisteredSkill(skillId).getToolsGroupName();
    // "mySkill_X_skill_tools"
    if (toolkit.getToolGroup(toolsGroupName) != null) {   // ← won't find "mySkill_X"
        toolkit.updateToolGroups(List.of(toolsGroupName), true);
    }
}
```

**Result**: The group registered via `createSkillToolGroup("mySkill_X", ...)` will never be found by `getToolsGroupName()`, so automatic tool group activation when a skill is loaded **silently fails**.

### Expected Behavior

`createSkillToolGroup()` should internally append `"_skill_tools"` to keep the naming convention encapsulated:

```java
public void createSkillToolGroup(
        String groupName, String description, boolean active, String activateOnSkill) {
    String internalName = groupName + "_skill_tools";  // ← append suffix internally
    SkillToolGroup group = SkillToolGroup.skillBuilder()
            .name(internalName)
            .description(description)
            .active(active)
            .activateOnSkill(activateOnSkill)
            .build();
    registerToolGroup(group);
}
```

This way callers don't need to know about the `"_skill_tools"` naming convention at all.

### Workaround (Current)

Callers must manually add `"_skill_tools"` to the group name:

```java
toolkit.createSkillToolGroup(skillId + "_skill_tools", description, false, skillId);
```

But this leaks an internal implementation detail and is error-prone.

### Affected Example

`SkillWithToolGroupExample.java` also exhibits this mismatch — its group name `"skill-tools"` won't match `getToolsGroupName()` return value `"data-analysis_skill_tools"`. It works only because the example relies on the LLM reading the `SkillToolGroup` description hint and manually activating the group, not the automatic path.

FrankBian:
## Description

Fixes #2034.

`ToolGroupManager.createSkillToolGroup()` now automatically appends the `_skill_tools` suffix to the provided group name. This aligns the registered group name with the name produced by `RegisteredSkill.getToolsGroupName()`, so skill-bound tool groups are correctly activated when the associated skill is loaded.

### Changes

- `ToolGroupManager#createSkillToolGroup`: computes an internal name by appending `_skill_tools` (idempotent — no double suffix if the caller already includes it).
- Added a `SKILL_TOOL_GROUP_SUFFIX` constant in `ToolGroupManager`.
- Updated affected unit tests to look up groups by their suffixed names.
- Added `testCreateSkillToolGroupIdempotentSuffix` to verify backward compatibility for callers that already pass the suffixed name.
- Fixed `SkillWithToolGroupExample` so the tool group name matches the activating skill (`data-analysis`), making the automatic activation path work as documented.

### Verification

- `mvn test -pl agentscope-core -Dtest=SkillToolGroupTest,MetaToolFactoryTest` passes (36 tests).
- `mvn spotless:apply` was run on affected modules.

## Checklist

- [x] Code follows the project style guidelines (spotless applied)
- [x] Tests added/updated and passing
- [x] Example updated to reflect the fix

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolGroupManager.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolGroupManager.java
@@ -33,6 +33,9 @@ class ToolGroupManager {
 
     private static final Logger logger = LoggerFactory.getLogger(ToolGroupManager.class);
 
+    /** Suffix appended to skill-bound tool group names for internal lookup. */
+    public static final String SKILL_TOOL_GROUP_SUFFIX = "_skill_tools";
+
     private final Map<String, ToolGroup> toolGroups = new ConcurrentHashMap<>(); // group -> tools
     private final Map<String, Set<String>> tools = new ConcurrentHashMap<>(); // tool -> groups
     private volatile List<String> activeGroups = new CopyOnWriteArrayList<>();
@@ -120,7 +123,13 @@ class ToolGroupManager {
     /**
      * Create a {@link SkillToolGroup} bound to a specific skill.
      *
-     * @param groupName Name of the tool group
+     * <p>The provided {@code groupName} is automatically suffixed with
+     * {@value #SKILL_TOOL_GROUP_SUFFIX} so that it matches the name produced by
+     * {@code RegisteredSkill.getToolsGroupName()}. Callers should pass the bare
+     * skill-related name (e.g. {@code "data-analysis"}) without appending the
+     * suffix themselves.
+     *
+     * @param groupName Name of the tool group (without the {@code _skill_tools} suffix)
      * @param description Description of the tool group
      * @param active Whether the tool group is active by default
      * @param activateOnSkill The skill name that this group is bound to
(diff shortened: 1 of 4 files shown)


chickenlj:
Please check #2057

PR #2047, opened 2026-07-08
fix(agent): restore structured-output self-healing retry in fallback path

### LINKED ISSUE #641: [Bug]:在结构化输出时，总是输出失败，提示报错信息为No structured output in message metadata. Key '_structured_output' not found
The abnormal information is：
java.util.concurrent.CompletionException: java.lang.IllegalStateException: No structured output in message metadata. Key '_structured_output' not found.
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315) ~[na:na]
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320) ~[na:na]
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run$$$capture(CompletableFuture.java:1770) ~[na:na]
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java) ~[na:na]
	at java.base/java.util.concurrent.ThreadPerTaskExecutor$TaskRunner.run(ThreadPerTaskExecutor.java:314) ~[na:na]
	at java.base/java.lang.VirtualThread.run(VirtualThread.java:329) ~[na:na]

The general reason is that the tool was not called, but the specific reason is still unclear

The text information returned by Ai is：There is no format here, all are the original information returned by Ai
```json
{
  "error": "",
  "issueDate": "2020-07-10",
  "roadTrafficSn": "11231220",
  "validityPeriod": "",
  "vehicleNum": "蒙123123"
}
```
OR

<tool_call>
<function=generate_response>
<parameter=response>
{"error": "", "vehicleNum": "鲁H12312", "vehicleNumColor": "黄色"}
</parameter>
</function>
</tool_call>

The framework I use is springboot, and I have encountered similar issues with pure Java code before, mainly with structured tools

zhangliyuangit:
## Problem

On the tool-based fallback structured-output path, `ReActAgent` relies on the model voluntarily calling the synthetic `generate_response` tool. If the model instead returns a normal text answer, the call ends without structured data and `getStructuredData()` fails.

This regresses the 1.x self-healing behavior, which reminded the model and retried with `tool_choice` forced to `generate_response`.

Fixes #641  
Refs #1103

## Fix

- Retry a finished, tool-free fallback structured-output response up to 3 times.
- Tag each retry with the existing structured-output reminder metadata and force `ToolChoice.Specific("generate_response")` only on those retry rounds.
- Leave ordinary calls, native `response_format`, and ReAct rounds containing business tool calls unchanged.
- Remove internal reminders and intermediate retry answers from persisted context on both success and give-up; the final structured result or final plain-text answer remains.

## Tests

Added `ReActAgentStructuredOutputRetryTest` covering:

- plain text -> forced `generate_response` retry -> structured result;
- refusal through the retry limit -> exactly 1 initial call + 3 retries, then plain text;
- voluntary first-round `generate_response` -> no retry;
- no reminder or intermediate-answer leakage into a follow-up turn;
- retry errors and Reactor cancellation also clean internal retry artifacts;
- returned usage includes every model attempt, including discarded retry answers.

TDD checks confirmed the retry, usage accounting, error cleanup, and cancellation cleanup regressions fail before their corresponding implementation changes.

Local verification on the rebuilt branch:

- `mvn -pl agentscope-core -Dtest=ReActAgentStructuredOutputRetryTest test`
- `mvn -pl agentscope-core test` — 2311 tests, 0 failures, 0 errors, 9 skipped
- `mvn spotless:check`
- `mvn -pl agentscope-core -am verify`

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -93,6 +93,8 @@ import io.agentscope.core.model.ExecutionConfig;
 import io.agentscope.core.model.GenerateOptions;
 import io.agentscope.core.model.Model;
 import io.agentscope.core.model.ModelRegistry;
+import io.agentscope.core.model.StructuredOutputReminder;
+import io.agentscope.core.model.ToolChoice;
 import io.agentscope.core.model.ToolSchema;
 import io.agentscope.core.permission.PermissionBehavior;
 import io.agentscope.core.permission.PermissionContextState;
@@ -220,6 +222,9 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
 
     /** Tool name used for the per-call structured-output {@code generate_response} tool. */
     public static final String STRUCTURED_OUTPUT_TOOL_NAME = "generate_response";
+
+    /** Maximum retries when the model ignores the fallback structured-output tool. */
+    private static final int STRUCTURED_OUTPUT_MAX_RETRIES = 3;
 
     /**
      * @deprecated Permission HITL no longer uses a Reactor Sink. Confirm results are now
@@ -1349,16 +1354,21 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                     scope.soTool = createStructuredOutputTool(jsonSchema);
 
                     return scope.doCallInner(msgs)
-                            .onErrorResume(error -> saveStateAfterCallFailure(scope, error))
+                            .doOnCancel(() -> cleanupFallbackStructuredOutputContext(scope))
+                            .onErrorResume(
+                                    error -> {
+                                        cleanupFallbackStructuredOutputContext(scope);
+                                        return saveStateAfterCallFailure(scope, error);
+                                    })
                             .flatMap(
                                     result -> {
                                         Msg out = result;
                                         if (scope.soCompleted && scope.soResultMsg != null) {
                                             ChatUsage aggregatedUsage =
-                                                    collectAggregatedUsage(scope.state);
+                                                    collectAggregatedUsage(scope);
                                             ThinkingBlock aggregatedThinking =
                                                     collectLastThinking(scope.state);
-                                            compressStructuredOutputContext(scope.state);
+                                            cleanupFallbackStructuredOutputContext(scope);
                                             Msg extracted =
                                                     extractStructuredResult(scope.soResultMsg);
                                             if (extracted != null) {
@@ -1369,6 +1379,8 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                                                 aggregatedThinking);
                                             }
                                             scope.state.contextMutable().add(out);
+                                        } else {
+                                            cleanupFallbackStructuredOutputContext(scope);
                                         }
                                         return saveStateToSession(scope).thenReturn(out);
                                     });
@@ -1417,6 +1429,11 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
         }
     }
 
+    private void cleanupFallbackStructuredOutputContext(CallExecution scope) {
+        compressStructuredOutputContext(scope.state);
+        scope.removeStructuredOutputRetryArtifacts();
+    }
+
     private boolean isStructuredOutputRelated(Msg msg) {
         Map<String, Object> metadata = msg.getMetadata();
         if (metadata != null
@@ -1434,14 +1451,15 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                         .allMatch(tr -> STRUCTURED_OUTPUT_TOOL_NAME.equals(tr.getName()));
     }
 
-    private ChatUsage collectAggregatedUsage(AgentState agentState) {
+    private ChatUsage collectAggregatedUsage(CallExecution scope) {
         int totalInput = 0;
         int totalOutput = 0;
         int totalCached = 0;
         double totalTime = 0;
         boolean hasUsage = false;
-        for (Msg msg : agentState.getContext()) {
-            if (isStructuredOutputRelated(msg) && msg.getRole() == MsgRole.ASSISTANT) {
+        for (Msg msg : scope.state.getContext()) {
+            if ((isStructuredOutputRelated(msg) || scope.isStructuredOutputRetryAttempt(msg))
+                    && msg.getRole() == MsgRole.ASSISTANT) {
                 ChatUsage usage = msg.getChatUsage();
                 if (usage != null) {
                     hasUsage = true;
@@ -1691,6 +1709,12 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
 
         /** The tool result message from the successful {@code generate_response} call. */
         Msg soResultMsg;
+
+        /** Number of retries after the model returned a tool-free structured-output response. */
+        int soRetryCount;
+
+        /** Intermediate model responses removed after structured-output retry completes. */
+        final List<Msg> soRetryAttemptMsgs = new ArrayList<>();
 
         /** Native structured-output format set on the per-call scope for native-path calls. */
         ResponseFormat nativeResponseFormat;
@@ -2315,6 +2339,17 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                                             .build(),
                                                     options);
                                 }
+                                if (soTool != null
+                                        && isToolChoiceReminder(event.getInputMessages())) {
+                                    options =
+                                            GenerateOptions.mergeOptions(
+                                                    GenerateOptions.builder()
+                                                            .toolChoice(
+                                                                    new ToolChoice.Specific(
+                                                                            STRUCTURED_OUTPUT_TOOL_NAME))
+                                                            .build(),
+                                                    options);
+                                }
                                 List<Msg> modelInput =
                                         prependSystemMsg(
                                                 event.getInputMessages(), event.getSystemMessage());
@@ -2452,6 +2487,24 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
 
                                 // Check finish conditions
                                 if (isFinished(eventMsg)) {
+                                    if (eventMsg != null
+                                            && soTool != null
+                                            && !soCompleted
+                                            && soRetryCount < STRUCTURED_OUTPUT_MAX_RETRIES) {
+                                        soRetryCount++;
+                                        log.debug(
+                                                "Model didn't call {}, requesting retry ({}/{})",
+                                                STRUCTURED_OUTPUT_TOOL_NAME,
+                                                soRetryCount,
+                                                STRUCTURED_OUTPUT_MAX_RETRIES);
+                                        soRetryAttemptMsgs.add(eventMsg);
+                                        state.contextMutable()
+                                                .add(createStructuredOutputReminder());
+                                        return reasoning(iter + 1, true);
+                                    }
+                                    if (soTool != null && !soCompleted) {
+                                        removeStructuredOutputRetryArtifacts();
+                                    }
                                     return Mono.justOrEmpty(eventMsg);
                                 }
 
@@ -3796,6 +3849,62 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                     Msg.METADATA_REMINDER_KIND,
                                     "empty_response"))
                     .build();
+        }
+
+        /** Build the reminder used to force the fallback structured-output tool on retry. */
+        private static Msg createStructuredOutputReminder() {
+            return Msg.builder()
+                    .name("system")
+                    .role(MsgRole.USER)
+                    .content(
+                            TextBlock.builder()
+                                    .text(
+                                            "Please call the '"
+                                                    + STRUCTURED_OUTPUT_TOOL_NAME
+                                                    + "' function to provide your response.")
+                                    .build())
+                    .metadata(
+                            Map.of(
+                                    MessageMetadataKeys.STRUCTURED_OUTPUT_REMINDER,
+                                    true,
+                                    MessageMetadataKeys.STRUCTURED_OUTPUT_REMINDER_TYPE,
+                                    StructuredOutputReminder.TOOL_CHOICE.toString()))
+                    .build();
+        }
+
+        /** Return whether the latest input asks the model to call the structured-output tool. */
+        private static boolean isToolChoiceReminder(List<Msg> msgs) {
+            if (msgs == null || msgs.isEmpty()) {
+                return false;
+            }
+            Msg last = msgs.get(msgs.size() - 1);
+            Map<String, Object> metadata = last != null ? last.getMetadata() : null;
+            return metadata != null
+                    && Boolean.TRUE.equals(
+                            metadata.get(MessageMetadataKeys.STRUCTURED_OUTPUT_REMINDER))
+                    && StructuredOutputReminder.TOOL_CHOICE
+                            .toString()
+                            .equals(
+                                    metadata.get(
+                                            MessageMetadataKeys.STRUCTURED_OUTPUT_REMINDER_TYPE));
+        }
+
+        /** Remove internal reminder messages and intermediate retry responses from context. */
+        private void removeStructuredOutputRetryArtifacts() {
+            state.contextMutable().removeIf(this::isStructuredOutputRetryArtifact);
+        }
+
+        private boolean isStructuredOutputRetryArtifact(Msg msg) {
+            Map<String, Object> metadata = msg.getMetadata();
+            boolean isReminder =
+                    metadata != null
+                            && Boolean.TRUE.equals(
+                                    metadata.get(MessageMetadataKeys.STRUCTURED_OUTPUT_REMINDER));
+            return isReminder || isStructuredOutputRetryAttempt(msg);
+        }
+
+        private boolean isStructuredOutputRetryAttempt(Msg msg) {
+            return soRetryAttemptMsgs.stream().anyMatch(attempt -> attempt == msg);
         }
 
         /**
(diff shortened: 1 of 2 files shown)


agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2078,6 +2104,24 @@ private Mono<Msg> runPostReasoningPipeline(Msg msg, int iter) {
...
+                                            && soRetryCount < STRUCTURED_OUTPUT_MAX_RETRIES) {
+                                        soRetryCount++;
+                                        log.debug(
+                                                "Model didn't call {}, requesting retry ({}/{})",
+                                                STRUCTURED_OUTPUT_TOOL_NAME,
+                                                soRetryCount,
+                                                STRUCTURED_OUTPUT_MAX_RETRIES);
+                                        state.contextMutable()
+                                                .add(createStructuredOutputReminder());
+                                        return reasoning(iter + 1, true);
```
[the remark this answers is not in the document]
zhangliyuangit: Intentional, keeping as-is. Retaining the plain-text attempt in the context is what lets the forced retry round restate the same content through the tool call — 1.x behaved identically (its `compressMemory` used the same `isStructuredOutputRelated` criteria and also kept the text attempt), and production ports of the 1.x hook rely on this echo for verbatim consistency between the text round and the tool round. Removing it would change 1.x parity semantics; happy to revisit if maintainers prefer.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2078,6 +2104,24 @@ private Mono<Msg> runPostReasoningPipeline(Msg msg, int iter) {
...
+                                    }
                                     return Mono.justOrEmpty(eventMsg);
```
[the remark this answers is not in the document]
zhangliyuangit: Fixed in e5d2b4a1. The give-up path now strips reminder-tagged messages from the persisted context before returning, and the regression test asserts a follow-up turn's model input contains no leftover reminders. (Note this is stricter than 1.x, which also leaked reminders on give-up — but the cleanup is clearly right.)

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -3279,6 +3323,50 @@ private boolean isFinished(Msg msg) {
...
+         * Build the reminder message injected when the model finishes a fallback
+         * structured-output call without invoking {@code generate_response}. Tagged with the
+         * {@link MessageMetadataKeys#STRUCTURED_OUTPUT_REMINDER} metadata so
+         * {@code compressStructuredOutputContext} removes it from the final context.
+         */
+        private Msg createStructuredOutputReminder() {
+            return Msg.builder()
+                    .name("system")
+                    .role(MsgRole.USER)
+                    .content(
```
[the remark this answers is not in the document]
zhangliyuangit: Keeping as-is: this is a verbatim port of the 1.x `StructuredOutputHook.createReminderMessage` (`name("system")` + `role(USER)`), and teams migrating from 1.x diff old/new logs against each other — changing the name would break that parity for no functional gain. Provider formatters key off `role`, which is USER here as before. Can change in a follow-up if maintainers prefer consistency over parity.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -3279,6 +3323,50 @@ private boolean isFinished(Msg msg) {
...
+                return false;
+            }
+            Msg last = msgs.get(msgs.size() - 1);
+            Map<String, Object> metadata = last != null ? last.getMetadata() : null;
+            return metadata != null
+                    && StructuredOutputReminder.TOOL_CHOICE
+                            .toString()
+                            .equals(
+                                    metadata.get(
+                                            MessageMetadataKeys.STRUCTURED_OUTPUT_REMINDER_TYPE));
```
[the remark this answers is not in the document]
zhangliyuangit: Fixed in e5d2b4a1 — `isToolChoiceReminder` now requires `STRUCTURED_OUTPUT_REMINDER == true` in addition to the TYPE key. (1.x checked only the TYPE key, but the extra guard is strictly safer.)

oss-maintainer:

```bash
git fetch origin
git checkout fix/structured-output-fallback-retry
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1153,6 +1162,20 @@ private Mono<Msg> doFallbackStructuredCall(List<Msg> msgs, Map<String, Object> j
                                                                 aggregatedThinking);
                                             }
                                             scope.state.contextMutable().add(out);
+                                        } else {
+                                            // Give-up path: the reminders injected by the
+                                            // self-healing retries are internal artifacts —
```
AgentScopeJavaBot, line 1165: [minor] Give-up path removes reminder messages but leaves 3-4 consecutive ASSISTANT messages in context (from retry rounds). This breaks alternating-role format required by some LLM providers (e.g., Anthropic Claude). In give-up path, also remove intermediate assistant responses from retry rounds, keeping only the final result.

AgentScopeJavaBot (commented):
🤖 AI Review

Restores structured-output self-healing retry in fallback path. When model ends with plain text (no generate_response tool call), injects reminder and forces tool_choice for up to 3 retries. One recommended issue: give-up path leaves consecutive assistant messages in context (reminders removed but intermediate assistant responses remain), which breaks alternating-role format for some LLM providers.

zhangliyuangit:
@oss-maintainer The branch has been rebased onto the latest main, and the review feedback has been addressed. The local agentscope-core suite passes with 2212 tests and 0 failures. The Java CI workflow is currently waiting for maintainer approval. Could you please approve the workflow and take another look when convenient?

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot: ✅ Verified as addressed in `a4457cb`: Give-up path now calls isFailedStructuredOutputRetryArtifact which removes both reminder messages and intermediate assistant responses (via soRetryAttemptMsgs), exactly as requested.

AgentScopeJavaBot (approved):

zhangliyuangit:
Rebuilt this PR on the latest `main` with a clean two-commit history.

The regression tests were confirmed failing before the implementation and now cover retry success, retry exhaustion, error/cancellation cleanup, context hygiene, and aggregated usage. Fresh local verification passes: `agentscope-core` 2311 tests (0 failures, 0 errors, 9 skipped), Spotless, and `mvn -pl agentscope-core -am verify`. The previous review feedback is preserved.

@oss-maintainer, could you please take another look when convenient?

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2315,6 +2339,17 @@ private Mono<Msg> reasoning(int iter, boolean ignoreMaxIters) {
                                                             .build(),
                                                     options);
                                 }
+                                if (soTool != null
+                                        && isToolChoiceReminder(event.getInputMessages())) {
```
oss-maintainer, line 2343: **[Warning]** Detecting the retry round by inspecting the *last* input message is fragile. `event.getInputMessages()` is the output of the `onReasoning` middleware chain, and several middlewares append their own transient messages to that list after the agent context (e.g. `TaskReminderMiddleware` adds a `<system-reminder>` todo message, `InboxMiddleware` appends inbox hints). Whenever one of them runs after us, `isToolChoiceReminder()` returns false and the retry degrades into a plain re-prompt — exactly the loop this PR is trying to heal, but now with a forced `tool_choice` that silently never applies.

Suggest carrying the intent explicitly instead of inferring it from message order, e.g. a `boolean soForceToolChoice` on `CallExecution` that `runPostReasoningPipeline` sets when it appends the reminder and `reasoning()` consumes (and clears) on the next round. That also makes the pairing between "reminder injected" and "tool_choice forced" impossible to desynchronise.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -3798,6 +3851,62 @@ private static Msg buildEmptyResponseReminder() {
...
+            Map<String, Object> metadata = msg.getMetadata();
+            boolean isReminder =
+                    metadata != null
+                            && Boolean.TRUE.equals(
+                                    metadata.get(MessageMetadataKeys.STRUCTURED_OUTPUT_REMINDER));
+            return isReminder || isStructuredOutputRetryAttempt(msg);
+        }
+
+        private boolean isStructuredOutputRetryAttempt(Msg msg) {
+            return soRetryAttemptMsgs.stream().anyMatch(attempt -> attempt == msg);
```
oss-maintainer, line 3907: **[Warning]** Identity comparison (`attempt == msg`). The retry attempt is registered from `eventMsg` in `runPostReasoningPipeline`, i.e. the object reference stored in `state.contextMutable()`. If any hook/middleware rewrites that entry in place (`Msg` is rebuilt by `withGenerateReason(...)`-style copy helpers, and compaction/middleware paths do replace context entries), the identity check silently no-ops and the discarded plain-text answer stays in the persisted session context — the follow-up turn then sees a stale assistant message that never produced structured output.

Reminder messages are already identified by metadata (`MessageMetadataKeys.STRUCTURED_OUTPUT_REMINDER`); tagging the retry attempt with a metadata key (or checking `soRetryAttemptMsgs` by `getId()`/name+text as a fallback) would make the cleanup robust against context rewriting. If identity is deliberately guaranteed here, a short comment stating "context entries are never replaced after `firePostReasoning`" would help future readers.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1349,16 +1354,21 @@ private Mono<Msg> doFallbackStructuredCall(List<Msg> msgs, Map<String, Object> j
                     scope.soTool = createStructuredOutputTool(jsonSchema);
 
                     return scope.doCallInner(msgs)
-                            .onErrorResume(error -> saveStateAfterCallFailure(scope, error))
+                            .doOnCancel(() -> cleanupFallbackStructuredOutputContext(scope))
```
oss-maintainer, line 1357: **[Warning]** `doOnCancel` mutates `scope.state` from the cancelling thread. `removeStructuredOutputRetryArtifacts()` / `compressStructuredOutputContext()` clear and rebuild the live context list, which is not synchronised and can be concurrently read by the pipeline that is still winding down, and unlike the error path this branch never calls `saveStateToSession`/`saveStateAfterCallFailure`, so the in-memory scope can end up cleaned while the stored session snapshot still contains the reminder. Two smaller points:

1. Can the cleanup be moved into the same `Mono.defer`/`doFinally` chain that already owns state persistence, so cancel, error and success all mutate the context on the same serialized path?
2. On cancel, should the cleaned context be persisted at all, or should the session keep the pre-call context (current `saveStateAfterCallFailure` behaviour for errors)?

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2315,6 +2339,17 @@ private Mono<Msg> reasoning(int iter, boolean ignoreMaxIters) {
                                                             .build(),
                                                     options);
                                 }
+                                if (soTool != null
+                                        && isToolChoiceReminder(event.getInputMessages())) {
+                                    options =
+                                            GenerateOptions.mergeOptions(
+                                                    GenerateOptions.builder()
+                                                            .toolChoice(
```
oss-maintainer, line 2347: **[Warning]** `mergeOptions(forcedOptions, options)` puts the synthesized options first, and `GenerateOptions.mergeOptions` lets the *primary* win field-by-field — so this overrides a user- or hook-provided `toolChoice` on the retry round. `ToolChoice.None` is a legitimate setting (some gateways/providers reject forced tool choice, and users set it to disable tool use for a turn); with this change a structured-output retry would flip it back to `Specific("generate_response")` and the call then fails at the provider instead of giving up gracefully.

Consider only forcing when the effective `toolChoice` is `null`/`Auto`/`Required`, and leaving `None`/an explicit `Specific` untouched (falling through to the give-up path), plus a `log.debug` when forcing is skipped so the interaction is diagnosable.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -221,6 +223,9 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
     /** Tool name used for the per-call structured-output {@code generate_response} tool. */
     public static final String STRUCTURED_OUTPUT_TOOL_NAME = "generate_response";
 
+    /** Maximum retries when the model ignores the fallback structured-output tool. */
+    private static final int STRUCTURED_OUTPUT_MAX_RETRIES = 3;
```
oss-maintainer, line 227: **[Info]** `STRUCTURED_OUTPUT_MAX_RETRIES = 3` is a hard-coded constant while the retry *strategy* is otherwise configurable on the builder (`maxIters`, model retry/backoff, and the existing `StructuredOutputReminder` enum in `io.agentscope.core.model`). Note that after this PR `StructuredOutputReminder` is only referenced as a metadata string (`TOOL_CHOICE.toString()`); the `PROMPT` mode documented as "inject reminder and retry" is now the behaviour actually implemented, but the enum still implies users can pick. Either wire the enum into the builder and honour both modes, or drop/annotate it so the dead option is not advertised in the public API surface. A `structuredOutputMaxRetries(int)` knob (0 disables) would also let latency-sensitive callers turn the loop off.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2452,6 +2487,24 @@ private Mono<Msg> runPostReasoningPipeline(Msg msg, int iter) {
...
+                                                STRUCTURED_OUTPUT_TOOL_NAME,
+                                                soRetryCount,
+                                                STRUCTURED_OUTPUT_MAX_RETRIES);
+                                        soRetryAttemptMsgs.add(eventMsg);
+                                        state.contextMutable()
+                                                .add(createStructuredOutputReminder());
+                                        return reasoning(iter + 1, true);
+                                    }
+                                    if (soTool != null && !soCompleted) {
+                                        removeStructuredOutputRetryArtifacts();
```
oss-maintainer, line 2506: **[Info]** Two different cleanup sites for one code path: the give-up branch here calls `removeStructuredOutputRetryArtifacts()` only, while `doFallbackStructuredCall`'s `else` branch additionally calls `compressStructuredOutputContext(...)` through `cleanupFallbackStructuredOutputContext(...)`. The net effect today is the same (the outer path always runs after this returns), but splitting the invariant across two places makes it easy to break later — e.g. if a future early return bypasses the outer flatMap, reminders stay in context. Would be cleaner to leave all cleanup to `cleanupFallbackStructuredOutputContext(scope)` and delete this call.

oss-maintainer (commented):
### Summary

Restores the 1.x self-healing loop for the fallback structured-output path: when the model answers in plain text instead of calling the synthetic `generate_response` tool, a reminder is injected and the next reasoning round forces `tool_choice=Specific("generate_response")`, up to 3 times, with internal reminder/intermediate artifacts removed from the context afterwards. The direction is right, the cleanup-on-error/cancel coverage is thoughtful, and the new test suite (retry success, exhaustion, error, cancellation, voluntary call) is well above the usual bar for this repo — the usage-aggregation assertion in particular is a nice touch.

The concerns are all about *how the retry round is recognised and cleaned up*, not about the design: the two detection mechanisms (last-message sniffing, reference identity) depend on message ordering and object identity that middlewares in this codebase are known to change, so the heal can silently degrade into a plain re-prompt or leak a discarded answer into the persisted session. None of these look like merge blockers if they are addressed or explicitly reasoned about.

### Findings
- **[Warning]** `ReActAgent.java:2343` — `isToolChoiceReminder()` only inspects the last input message; `TaskReminderMiddleware` / `InboxMiddleware` append messages after it, so forcing can silently not apply. Prefer an explicit per-call flag.
- **[Warning]** `ReActAgent.java:3907` — retry artifacts are matched by reference identity, which breaks if a hook/middleware replaces the context entry; the discarded answer then persists.
- **[Warning]** `ReActAgent.java:1357` — `doOnCancel` mutates the shared context off the serialized pipeline and never persists the cleaned state, unlike the error path.
- **[Warning]** `ReActAgent.java:2347` — forced options win in `mergeOptions`, overriding a user/hook-provided `ToolChoice.None`; restrict forcing and log skips.
- **[Info]** `ReActAgent.java:227` — retry count is hard-coded and `StructuredOutputReminder` is now effectively dead as a user-facing option (only used as a metadata string); wire it or trim it.
- **[Info]** `ReActAgent.java:2506` — give-up cleanup is split between two sites covering the same invariant.

### Questions
1. Does `agent.call(msgs, Class)` on the fallback path have any existing e2e/integration coverage on a real provider (DashScope/OpenAI compatible) confirming providers accept `tool_choice` pointing at a tool that is injected per-call and is *not* in the shared toolkit schema list? Some providers validate `tool_choice` against `tools[]`.
2. Should a give-up after 3 retries surface anything to the caller (event/log at WARN, or a `GenerateReason` marker on the returned message)? Today it returns plain text with only a `log.debug`, which is hard to distinguish from "the model legitimately answered in prose" when debugging user reports.
3. The PR description says reminders are removed "on both success and give-up" — is the *multi-turn* case intended (a later `agent.call` in the same session never sees them) but the intermediate answers of an *earlier successful* structured call also removed from history? That looks desirable for token cost, just confirming it is deliberate since it changes what a session store contains for existing users of the fallback path.

---

PR #2049, opened 2026-07-08
fix(filesystem): resolve leading '/' paths relative to workspace in ROOTED mode

### LINKED ISSUE #2046: [Bug]: LocalFilesystem ROOTED mode rejects AbstractFilesystem contract-compliant absolute paths

**Describe the bug**

`AbstractFilesystem` interface contract specifies that path parameters "must start with '/'" (see Javadoc on `ls()`, `read()`, etc.). However, `LocalFilesystem` in `ROOTED` mode (the default for `LocalFilesystemSpec`) treats `/`-prefixed paths incorrectly:

- **On Windows:** `Path.of("/skills").isAbsolute()` returns `false` (no drive letter), so the path goes through `cwd.resolve("/skills")` which resolves to `<drive>:\skills` (current drive root). The directory doesn't exist, so `ls()` **silently returns an empty list** — no error, no exception, just wrong results.
- **On Unix:** `Path.of("/skills").isAbsolute()` returns `true`, so it goes through `PathPolicy` check and throws `SecurityException`.

The silent failure mode is particularly dangerous because callers get no indication that something went wrong.

Meanwhile, relative paths like `"skills"` (without leading `/`) work correctly — they are resolved via `cwd.resolve(...)` and land inside the workspace. But this contradicts the interface contract.

This creates an impossible situation for callers: following the contract (`/skills`) fails silently or throws, violating it (`skills`) works.

**To Reproduce**

1. Build a filesystem using `LocalFilesystemSpec` with default settings (`ROOTED` mode):

```java
LocalFilesystemSpec spec = new LocalFilesystemSpec();
AbstractFilesystem fs = spec.toFilesystem(workspacePath, nsFactory);
```

2. Call `ls` with an absolute path as the `AbstractFilesystem` contract requires:

```java
// On Windows: silently returns empty list (no error!)
// On Unix: throws SecurityException
LsResult result = fs.ls(RuntimeContext.empty(), "/skills");
```

3. Call `ls` with a relative path (works, but violates the "must start with '/'" contract):

```java
// Succeeds — resolved relative to workspace cwd
LsResult result = fs.ls(RuntimeContext.empty(), "skills");
```

4. JUnit test case that reproduces the bug (add to `LocalFilesystemModeTest.java`):

```java
@Test
void rooted_absolutePathWithinWorkspaceShouldResolveToWorkspaceRoot(@TempDir Path workspace)
        throws IOException {
    // Create a "skills" directory inside the workspace
    Path skillsDir = workspace.resolve("skills");
    Files.createDirectories(skillsDir);
    Files.writeString(skillsDir.resolve("SKILL.md"), "# Skill", StandardCharsets.UTF_8);

    PathPolicy policy = PathPolicy.of(workspace);
    LocalFilesystem fs = new LocalFilesystem(workspace, LocalFsMode.ROOTED, policy, 10, null);

    // Relative path works (violates contract but succeeds)
    LsResult relativeResult = fs.ls(RuntimeContext.empty(), "skills");
    assertTrue(relativeResult.isSuccess());
    assertFalse(relativeResult.entries().isEmpty(), "should find SKILL.md");

    // Absolute path following the contract should also work — but currently fails
    // On Windows: returns empty list (silent miss) because Path.of("/skills") resolves
    //             to <drive>:\skills which doesn't exist
    // On Unix:    throws SecurityException because /skills is not under PathPolicy roots
    LsResult absoluteResult = fs.ls(RuntimeContext.empty(), "/skills");
    assertTrue(absoluteResult.isSuccess());
    assertFalse(
            absoluteResult.entries().isEmpty(),
            "absolute path '/skills' should find the same entries as relative 'skills'");
}
```

**Test result** (reproduces the bug):

```
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
[ERROR] rooted_absolutePathWithinWorkspaceShouldResolveToWorkspaceRoot
  ==> expected: <false> but was: <true>
```

The test fails because `fs.ls("/skills")` returns an empty list on Windows (silent miss) or throws `SecurityException` on Unix, while `fs.ls("skills")` correctly returns the `SKILL.md` entry.

**Expected behavior**

Paths like `"/skills"` should be resolved within the workspace root, consistent with the `AbstractFilesystem` contract. The `/` prefix should be treated as a logical/workspace-absolute path, not a host-machine absolute path.

The `isAbsolutePathString()` check in `applyNamespacePrefix()` should distinguish between:
- **Logical absolute paths** (`/skills`, `/skills/foo/SKILL.md`) — should be resolved relative to workspace root
- **Host absolute paths** (paths outside the workspace, e.g. system directories) — should be validated against `PathPolicy`

One approach: in `resolveRooted()`, strip the leading `/` and resolve relative to `cwd` (similar to what `resolveSandboxed()` already does), and only apply `PathPolicy` checks to paths that are genuinely host-absolute.

**Error messages / observed behavior**

On Windows (silent failure — most dangerous):
```
// fs.ls("/skills") returns:
LsResult{success=true, entries=[]}  // Empty! No error, no exception
```

On Unix (throws exception):
```
SecurityException: Absolute path /skills is not within an allowed root.
Filesystem root: ...; additional roots: [...]
```

**Root cause analysis**

Two mechanisms combine to produce this bug:

1. `LocalFilesystem.applyNamespacePrefix()` uses `isAbsolutePathString()` which returns `true` for any path starting with `/`. This causes namespace prefix to be skipped entirely for `/`-prefixed paths — they pass through as-is.

2. `LocalFilesystem.resolveRooted()` then receives the bare `/`-prefixed path. The behavior diverges by platform:
   - **Windows:** `Path.of("/skills").isAbsolute()` returns `false` (no drive letter), so it goes through `cwd.resolve("/skills")` → `<drive>:\skills`. The `resolveRooted()` method does NOT check `startsWith(cwd)` for this branch, so the path escapes the workspace silently. `ls()` then finds the directory doesn't exist and returns an empty list.
   - **Unix:** `Path.of("/skills").isAbsolute()` returns `true`, so it goes through `PathPolicy.isAllowed()` which rejects it with `SecurityException`.

Note that `resolveSandboxed()` does NOT have this problem — it strips the leading `/` and resolves relative to `cwd`. The issue is specific to `ROOTED` mode.

- AgentScope-Java Version: latest (agentscope-harness, main branch)
- Java Version: 17+
- OS: any

**Additional context**

- `WorkspaceManager` internally always uses relative paths. Its `normalizeRelativePath()` method explicitly strips leading slashes:
  ```java
  static String normalizeRelativePath(String relativePath) {
      String s = relativePath.replace('\\', '/').stripLeading();
      while (s.startsWith("/")) {
          s = s.substring(1);
      }
      return s;
  }
  ```
  All internal calls (`writeUtf8WorkspaceRelative`, `readWithOverride`, `appendUtf8WorkspaceRelative`, etc.) go through this normalization, ensuring paths passed to the filesystem are always relative. This confirms the framework itself works around the contract mismatch — but external callers who follow the `AbstractFilesystem` contract (passing `/`-prefixed paths) hit the bug.

- `NamespacedFilesystemView.scoped()` (in downstream projects) produces `/`-prefixed absolute paths for namespace scoping. When the delegate is a `LocalFilesystem`, these paths also fail for the same reason.

- The `AbstractFilesystem` contract ("must start with '/'") was likely designed with `RemoteFilesystem` in mind, where paths are logical keys. `LocalFilesystem` was not updated to honor this contract.

- `CompositeFilesystem.routeForPath()` already handles this gracefully — it strips leading slashes before matching routes, explicitly noting: *"callers can pass either '/skills/foo' (the AbstractFilesystem contract) or 'skills/foo' (the convention used by WorkspaceManager.writeUtf8WorkspaceRelative) and route to the same store"*. This shows the dual convention is a known pattern, but `LocalFilesystem.resolveRooted()` was not updated to match.

mroldx:
## AgentScope-Java Version
2.0.0-SNAPSHOT

## Description
Fix #2046

In ROOTED mode, paths starting with "/" represent logical workspace-absolute paths (e.g. "/skills" means "<workspace>/skills"), not host-machine absolute paths. This commit modifies resolveRooted() to strip the leading "/" and resolve relative to cwd, similar to resolveSandboxed().

Root cause:
- On Windows: Path.of("/skills").isAbsolute() returns false (no drive letter), causing cwd.resolve("/skills") to resolve to <drive>:\skills which doesn't exist.
- On Unix: Path.of("/skills").isAbsolute() returns true, triggering SecurityException.

Changes:
- Modified resolveRooted() to handle "/" prefixed paths as workspace-relative
- Added path traversal protection for "/../" patterns
- Added 5 regression tests for the fix

## Checklist

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
@@ -640,11 +640,14 @@ public class LocalFilesystem implements AbstractFilesystem {
             if (stripped.isEmpty()) {
                 return cwd;
             }
-            // Check for path traversal attacks
-            if (stripped.contains("..") || stripped.startsWith("~")) {
+            if (stripped.startsWith("~")) {
                 throw new SecurityException("Path traversal not allowed: " + effectiveKey);
             }
-            return cwd.resolve(stripped).normalize();
+            Path full = cwd.resolve(stripped).normalize();
+            if (!full.startsWith(cwd)) {
+                throw new SecurityException("Path " + full + " outside root directory: " + cwd);
+            }
+            return full;
         }
 
         Path target = Path.of(effectiveKey);
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
@@ -628,13 +628,20 @@ public class LocalFilesystem implements AbstractFilesystem {
     }
 
     private Path resolveRooted(String effectiveKey) {
-        // In ROOTED mode, paths starting with "/" represent logical workspace-absolute paths
-        // (e.g. "/skills" means "<workspace>/skills"), not host-machine absolute paths.
-        // Strip the leading "/" and resolve relative to cwd, similar to resolveSandboxed().
-        // This is necessary because:
-        // - On Windows: Path.of("/skills").isAbsolute() returns false (no drive letter),
-        //   causing cwd.resolve("/skills") to resolve to <drive>:\skills which doesn't exist.
-        // - On Unix: Path.of("/skills").isAbsolute() returns true, triggering SecurityException.
+        Path target = Path.of(effectiveKey);
+        if (target.isAbsolute()) {
+            Path normalized = target.normalize();
+            if (normalized.startsWith(cwd) || pathPolicy.isAllowed(normalized)) {
+                return normalized;
+            }
+            if (Files.exists(normalized)) {
+                throw rootAccessDenied(normalized);
+            }
+            if (!effectiveKey.startsWith("/")) {
+                throw rootAccessDenied(normalized);
+            }
+        }
+
         if (effectiveKey.startsWith("/")) {
             String stripped = effectiveKey.substring(1);
             if (stripped.isEmpty()) {
@@ -650,22 +657,17 @@ public class LocalFilesystem implements AbstractFilesystem {
             return full;
         }
 
-        Path target = Path.of(effectiveKey);
-        if (target.isAbsolute()) {
-            // True host absolute paths (Windows drive letters, UNC paths)
-            Path normalized = target.normalize();
-            if (normalized.startsWith(cwd) || pathPolicy.isAllowed(normalized)) {
-                return normalized;
-            }
-            throw new SecurityException(
-                    "Absolute path "
-                            + normalized
-                            + " is not within an allowed root. Filesystem root: "
-                            + cwd
-                            + "; additional roots: "
-                            + pathPolicy.roots());
-        }
         return cwd.resolve(target).normalize();
+    }
+
+    private SecurityException rootAccessDenied(Path normalized) {
+        return new SecurityException(
+                "Absolute path "
+                        + normalized
+                        + " is not within an allowed root. Filesystem root: "
+                        + cwd
+                        + "; additional roots: "
+                        + pathPolicy.roots());
     }
 
     private Path resolveUnrestricted(String effectiveKey) {
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
@@ -628,8 +628,28 @@ public class LocalFilesystem implements AbstractFilesystem {
     }
 
     private Path resolveRooted(String effectiveKey) {
+        // In ROOTED mode, paths starting with "/" represent logical workspace-absolute paths
+        // (e.g. "/skills" means "<workspace>/skills"), not host-machine absolute paths.
+        // Strip the leading "/" and resolve relative to cwd, similar to resolveSandboxed().
+        // This is necessary because:
+        // - On Windows: Path.of("/skills").isAbsolute() returns false (no drive letter),
+        //   causing cwd.resolve("/skills") to resolve to <drive>:\skills which doesn't exist.
+        // - On Unix: Path.of("/skills").isAbsolute() returns true, triggering SecurityException.
+        if (effectiveKey.startsWith("/")) {
+            String stripped = effectiveKey.substring(1);
+            if (stripped.isEmpty()) {
+                return cwd;
+            }
+            // Check for path traversal attacks
+            if (stripped.contains("..") || stripped.startsWith("~")) {
+                throw new SecurityException("Path traversal not allowed: " + effectiveKey);
+            }
+            return cwd.resolve(stripped).normalize();
+        }
+
         Path target = Path.of(effectiveKey);
         if (target.isAbsolute()) {
+            // True host absolute paths (Windows drive letters, UNC paths)
             Path normalized = target.normalize();
             if (normalized.startsWith(cwd) || pathPolicy.isAllowed(normalized)) {
                 return normalized;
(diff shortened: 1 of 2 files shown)


chickenlj (commented):
Thanks for the fix — the issue is real and the direction is correct. One security concern:

### Missing `startsWith(cwd)` post-normalization check

`resolveSandboxed()` has a defense-in-depth pattern:

```java
Path full = cwd.resolve(stripped).normalize();
if (!full.startsWith(cwd)) {
    throw new SecurityException("Path " + full + " outside root directory: " + cwd);
}
return full;
```

The new code in `resolveRooted()` only does a string-level `contains("..")` check but skips the post-normalize `startsWith(cwd)` validation. The string check is coarse — it might false-positive on legitimate names like `some..dir`, and more importantly it lacks the belt-and-suspenders guarantee that the final resolved path is actually within the workspace.

Please add the `startsWith(cwd)` check after `normalize()`:

```java
if (effectiveKey.startsWith("/")) {
    String stripped = effectiveKey.substring(1);
    if (stripped.isEmpty()) {
        return cwd;
    }
    if (stripped.contains("..") || stripped.startsWith("~")) {
        throw new SecurityException("Path traversal not allowed: " + effectiveKey);
    }
    Path full = cwd.resolve(stripped).normalize();
    if (!full.startsWith(cwd)) {
        throw new SecurityException(
                "Path " + full + " outside root directory: " + cwd);
    }
    return full;
}
```

This aligns with how `resolveSandboxed()` handles the same scenario and provides a final guarantee regardless of edge cases in path normalization.

Other than that, the fix and tests look good.

mroldx:
> Thanks for the fix — the issue is real and the direction is correct. One security concern:
> 
> ### Missing `startsWith(cwd)` post-normalization check
> `resolveSandboxed()` has a defense-in-depth pattern:
> 
> ```java
> Path full = cwd.resolve(stripped).normalize();
> if (!full.startsWith(cwd)) {
>     throw new SecurityException("Path " + full + " outside root directory: " + cwd);
> }
> return full;
> ```
> 
> The new code in `resolveRooted()` only does a string-level `contains("..")` check but skips the post-normalize `startsWith(cwd)` validation. The string check is coarse — it might false-positive on legitimate names like `some..dir`, and more importantly it lacks the belt-and-suspenders guarantee that the final resolved path is actually within the workspace.
> 
> Please add the `startsWith(cwd)` check after `normalize()`:
> 
> ```java
> if (effectiveKey.startsWith("/")) {
>     String stripped = effectiveKey.substring(1);
>     if (stripped.isEmpty()) {
>         return cwd;
>     }
>     if (stripped.contains("..") || stripped.startsWith("~")) {
>         throw new SecurityException("Path traversal not allowed: " + effectiveKey);
>     }
>     Path full = cwd.resolve(stripped).normalize();
>     if (!full.startsWith(cwd)) {
>         throw new SecurityException(
>                 "Path " + full + " outside root directory: " + cwd);
>     }
>     return full;
> }
> ```
> 
> This aligns with how `resolveSandboxed()` handles the same scenario and provides a final guarantee regardless of edge cases in path normalization.
> 
> Other than that, the fix and tests look good.

Fixed. Added the post-normalization startsWith(cwd) guard for ROOTED leading-slash paths, and added a regression test to ensure literal ".." in path segment names is still allowed while traversal remains rejected.

chickenlj:
Please check the test failures:


```
[INFO] 
Error:  Failures: 
Error:    LocalFilesystemModeTest.rooted_absolutePathLsNotCorruptedByNamespace:177 ls should find files in the project directory ==> expected: <false> but was: <true>
Error:    LocalFilesystemModeTest.rooted_absolutePathNotCorruptedByNamespace:158 absolute path should resolve correctly, got: File '/tmp/junit-13468257907594244212/Main.java' not found ==> expected: <true> but was: <false>
Error:    LocalFilesystemModeTest.rooted_acceptsAbsolutePathUnderPolicyRoot:70 expected success, got: File '/tmp/junit-17109832293948361478/AGENTS.md' not found ==> expected: <true> but was: <false>
Error:    LocalFilesystemModeTest.rooted_rejectsAbsolutePathOutsideAllRoots:85 Expected java.lang.SecurityException to be thrown, but nothing was thrown.
Error:    LocalFilesystemModeTest.rooted_workspaceItselfIsImplicitlyAllowed:115 expected: <true> but was: <false>
Error:    AtPathExpansionMiddlewareTest.expandsAbsolutePathInsidePolicy:78 expected: <true> but was: <false>
[INFO] 
Error:  Tests run: 615, Failures: 6, Errors: 0, Skipped: 0
```

PR #2057, opened 2026-07-08
fix(core): activate SkillToolGroup via activateOnSkill field

### LINKED ISSUE #2034: createSkillToolGroup should internally append "_skill_tools" suffix to group name

chickenlj:
## Summary

Fixes #2034.

- `SkillToolGroup.activateOnSkill` was stored but never queried — skill-bound tool groups created via `createSkillToolGroup()` were never auto-activated when the associated skill was loaded.
- Added `findSkillToolGroupsByActivateOnSkill()` to `ToolGroupManager`/`Toolkit` to scan groups by their `activateOnSkill` field.
- Updated all activation sites (`SkillToolFactory.activateSkill()`, `SkillBox.syncToolGroupStates()`, `SkillBox.setSkillActive()`) to also activate groups matched by `activateOnSkill`, in addition to the existing `skillId + "_skill_tools"` name-convention path.
- The existing internal path (SkillBox line 579) continues to work unchanged — this is purely additive.

## Test plan

- [x] `mvn test -pl agentscope-core -Dtest=SkillToolGroupTest,MetaToolFactoryTest` passes
- [x] New tests: `testFindSkillToolGroupsByActivateOnSkill`, `testFindSkillToolGroupsByActivateOnSkillIgnoresRegularGroups`, `testFindSkillToolGroupsByActivateOnSkillNullSafe`
- [x] `mvn compile -pl agentscope-examples/documentation -am` compiles cleanly
- [ ] Verify end-to-end: create a `SkillToolGroup` with an arbitrary name, load the matching skill, confirm the group becomes active

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
@@ -168,14 +168,28 @@ public class SkillBox {
 
         // Dynamically update active/inactive tool groups based on skills' states
         for (RegisteredSkill registeredSkill : skillRegistry.getAllRegisteredSkills().values()) {
-            if (toolkit.getToolGroup(registeredSkill.getToolsGroupName()) == null) {
-                continue; // Skip uncreated skill tools
+            // Name-convention path: skillId + "_skill_tools"
+            if (toolkit.getToolGroup(registeredSkill.getToolsGroupName()) != null) {
+                if (!registeredSkill.isActive()) {
+                    inactiveSkillToolGroups.add(registeredSkill.getToolsGroupName());
+                } else {
+                    activeSkillToolGroups.add(registeredSkill.getToolsGroupName());
+                }
             }
-            if (!registeredSkill.isActive()) {
-                inactiveSkillToolGroups.add(registeredSkill.getToolsGroupName());
-                continue; // Skip inactive skill's tools, its tools won't be included
+
+            // activateOnSkill path: scan SkillToolGroups bound to this skill's name
+            AgentSkill agentSkill = skillRegistry.getSkill(registeredSkill.getSkillId());
+            if (agentSkill != null) {
+                List<String> boundGroups =
+                        toolkit.findSkillToolGroupsByActivateOnSkill(agentSkill.getName());
+                for (String group : boundGroups) {
+                    if (!registeredSkill.isActive()) {
+                        inactiveSkillToolGroups.add(group);
+                    } else {
+                        activeSkillToolGroups.add(group);
+                    }
+                }
             }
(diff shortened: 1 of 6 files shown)


jujn (approved):
LGTM

PR #2063, opened 2026-07-08
fix(harness): cancel spawned agent on parent dispose

### LINKED ISSUE #2062: [Bug]:AgentSpawnTool leaks orphan ReActAgent on parent subscription cancel
`AgentSpawnTool.execWithTimeoutPromotion` fire-and-forget-subscribes the inner
  sub-agent execution Mono via plain `.subscribe(...)` (line 732). The resulting
  `Disposable` is never captured, so when the parent subscription to
  `agentSpawn(...)` is cancelled (by an outer `Mono.timeout()`, a user-initiated
  stop, or upstream Reactor cancel), the inner execution is **not** disposed and
  the sub-agent keeps running as an orphan — making further LLM calls and tool
  calls whose results are silently dropped (the parent's `MonoSink` is already
  cancelled, so the eventual `sink.success(...)` becomes a no-op).

  Same class of bug as #1783, but in the harness-side `AgentSpawnTool` rather
  than the core `SubAgentTool`. The fix from `029cc55e` was not propagated here.

  ## To Reproduce

  ### Test code

  ```java
  // File: agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/AgentSpawnToolOrphanCancelTest.java

  MockModel slowModel = new MockModel(messages -> {
      String callId = "call_" + UUID.randomUUID();   // unique id per call → no dedup
      return List.of(ChatResponse.builder()
              .id("msg_" + UUID.randomUUID())
              .content(List.of(ToolUseBlock.builder()
                      .name("slow_tool").id(callId)
                      .input(Map.of()).build()))
              .build());
  });

  ReActAgent slowSpy = Mockito.spy(ReActAgent.builder()
          .name("slow_sub").sysPrompt("slow sub")
          .model(slowModel).toolkit(slowTk)
          .maxIters(50).build());

  DefaultAgentManager manager = new DefaultAgentManager(
          List.of(new SubagentEntry("slow_agent", "Test", rc -> slowSpy)), null);
  AgentSpawnTool tool = new AgentSpawnTool(manager, new NoopTaskRepository(), 0);

  RuntimeContext ctx = RuntimeContext.builder().sessionId("s1").userId("u1").build();

  // Outer .timeout(2s) cancels the Mono<String> returned by agentSpawn
  try {
      tool.agentSpawn(ctx, null, "slow_agent", "go", null, 30, null)
              .timeout(Duration.ofSeconds(2))
              .block();
  } catch (Exception expected) { /* outer timeout */ }

  long linesAtTimeout = Files.readAllLines(tmpFile).size();
  long callsAtTimeout = slowModel.getCallCount();
  Thread.sleep(4_000);
  long linesAfter   = Files.readAllLines(tmpFile).size();
  long callsAfter   = slowModel.getCallCount();

  verify(slowSpy, atLeastOnce()).interrupt(any(RuntimeContext.class));   // FAILS

  How to execute

      -Dtest=AgentSpawnToolOrphanCancelTest -DfailIfNoTests=false

  What happens

  1. agentSpawn(...) returns a Mono<String>. execWithTimeoutPromotion
  builds a Mono.create(sink -> ...) whose sink callback fire-and-forget-
  subscribes execLocalSync(...):
  execLocalSync(agent, sessionId, userId, task, spawned, runtimeContext)
          .contextWrite(c -> reactor.util.context.Context.of(parentCtx))
          .subscribe(bridge::complete, bridge::completeExceptional, ...);
  2. Outer .timeout(2s) cancels the parent subscription at t≈2.2s. Mono.create
  honors the cancel by disposing the parent sink — but the inner subscription
  from .subscribe(...) is never captured, so it is not disposed.
  3. The ReActAgent reasoning loop continues on boundedElastic: model call →
  tool_call → slow_tool writes a line → loop.
  4. The internal CompletableFuture bridge + race.orTimeout(30s) are still
  alive. When the agent eventually finishes, race.whenComplete invokes
  sink.success(...) on the already-cancelled sink — a no-op. Result silently
  dropped.
  5. interrupt(RuntimeContext) is never called on the sub-agent.

  Observable proof (measured on commit e070c8a5)

  outer timeout fired at t=2202ms
  at timeout:    lines=8,  modelCalls=9,  toolCalls=9
  after 4s wait: lines=34, modelCalls=34, toolCalls=34
                 delta = +26 lines, +25 modelCalls, +25 toolCalls
  Mockito.verify(slowSpy, atLeastOnce()).interrupt(any): FAILS (0 calls)

  - 8 iterations completed during the 2s legitimate window.
  - 25 additional iterations executed after parent cancel — each one a wasted
  LLM call + tool execution whose result was silently dropped.

  Expected behavior

  When the parent subscription to agentSpawn(...) is cancelled, the inner
  fire-and-forget subscription should be disposed and ReActAgent.interrupt(ctx)
  invoked on the sub-agent — mirroring the fix applied to SubAgentTool in
  029cc55e.

  After the fix, the same test asserts:
  - delta lines ≤ 1 (an in-flight tool call may still complete)
  - delta modelCalls ≤ 1
  - interrupt(RuntimeContext) called at least once on the sub-agent

  Production impact

  Observed in a Spring Boot + AgentScope-Java multi-agent orchestration backend:

  ┌──────────────────────────────────────────────────────────────────┬─────────┐
  │                              Metric                              │  Value  │
  ├──────────────────────────────────────────────────────────────────┼─────────┤
  │ Sub-agents spawned per main-agent run                            │ 9       │
  ├──────────────────────────────────────────────────────────────────┼─────────┤
  │ Total sub-agent invocations (incl. retries)                      │ 19      │
  ├──────────────────────────────────────────────────────────────────┼─────────┤
  │ Estimated wasted tokens (orphan continuations after parent exit) │ ~1M     │
  ├──────────────────────────────────────────────────────────────────┼─────────┤
  │ Wall-clock inflation per affected run                            │ 30+ min │
  └──────────────────────────────────────────────────────────────────┴─────────┘

  In production the trigger is not Mono.timeout() but the parent main-agent
  exiting (user abort, exception, or natural completion with pending
  sub-agents). Orphan sub-agents continue hitting the LLM until they hit
  maxIters, and the application has no way to retrieve or observe their
  results — they silently burn tokens.

  A local workaround was attempted in application code (intercepting the
  parent AgentEndEvent and force-interrupting tracked sub-agents), but it
  is brittle and breaks the legitimate async-cooperation path (children
  intentionally outliving the parent, woken up later via WakeupDispatcher).
  A framework-level fix is needed.

  Root cause

  AgentSpawnTool.execWithTimeoutPromotion (agentscope-harness, lines 706-766)
  decouples sub-agent execution from observation using a CompletableFuture
  bridge. This decoupling is intentional for the timeout-promotion path:
  when the internal race.orTimeout(timeoutMs) fires, the bridge is wrapped
  as AdoptedTaskRunSpec so the user can retrieve the result later via
  task_output.

  However, the same fire-and-forget pattern also handles external parent
  cancellation — and there, the result is genuinely lost (no task_id was
  returned before cancel). The inner subscription should be disposed and the
  agent interrupted in this case, but the code does neither.

  The fix is to capture the inner Disposable and dispose it on parent cancel,
  plus add a doFinally(CANCEL) that calls interrupt(ctx) on the sub-agent
  (mirroring 029cc55e):

  // in execWithTimeoutPromotion, inside Mono.create(sink -> ...)
  Mono<Msg> inner = execLocalSync(agent, sessionId, userId, task, spawned, runtimeContext)
          .contextWrite(c -> reactor.util.context.Context.of(parentCtx))
          .doFinally(signal -> {
                  interruptAgent(agent, runtimeContext);
              }
          });

  Disposable innerSub = inner.subscribe(
          bridge::complete,
          bridge::completeExceptional,
          () -> { if (!bridge.isDone()) bridge.complete(null); });

  // ...
  sink.onCancel(innerSub::dispose);

  This preserves the timeout-promotion path (the internal orTimeout does not
  produce a CANCEL signal — it completes race exceptionally with
  TimeoutException, handled separately by handleExecError).

  Environment

  - AgentScope-Java commit: e070c8a5 (main, 2026-07-08)
  - JDK: OpenJDK 21
  - Module: agentscope-harness
  - OS: Windows 11 (reproducible on any platform)

  Additional context

  - Same class of bug as #1783 (filed 2026-06-17 by @xiaguanglei), which fixed
  SubAgentTool in commit 029cc55e. This issue is the equivalent fix for
  the harness-side AgentSpawnTool.
  - Difference between the two tools:
    - SubAgentTool (core): inner Mono is part of the Reactor chain — cancel
  propagates naturally once doFinally(CANCEL) is added.
    - AgentSpawnTool (harness): inner Mono is fire-and-forget subscribed —
  needs explicit Disposable capture + sink.onCancel(...) to propagate
  cancel.
  - The async cooperation path (children intentionally outliving the parent,
  woken up via WakeupDispatcher → MessageBus.enqueueWakeup) uses
  TaskRepository.putTask with timeout_seconds=0 and is not affected
  by this fix. Only the sync execution path (execWithTimeoutPromotion)
  is changed.
  - Full reproduction test: AgentSpawnToolOrphanCancelTest.java (included in
  the PR).

mroldx:
## AgentScope-Java Version

2.0.0-RC5

## Description

This PR fixes #2062.

`AgentSpawnTool.execWithTimeoutPromotion()` starts local sub-agent execution with an inner Reactor subscription. Previously, when the parent `agent_spawn` subscription was cancelled, the inner subscription was not disposed. As a result, the spawned local sub-agent could continue running after the parent no longer observed the result.

This change propagates parent cancellation to the in-flight local sub-agent execution by:

- capturing the inner `Disposable`
- disposing it from `sink.onCancel(...)`
- interrupting local `ReActAgent` instances when the inner execution receives `SignalType.CANCEL`
- avoiding post-cancel bridge callbacks from writing back to an already-cancelled sink

This only affects parent subscription cancellation. It does not change the normal async/background task path (`timeout_seconds=0`) or timeout promotion behavior where a `task_id` is returned.

Tests added:

- `AgentSpawnToolCancellationTest`
  - verifies that cancelling `agent_spawn` cancels the in-flight local sub-agent `Mono`
  - verifies that `ReActAgent.interrupt(RuntimeContext)` is called

Tested with:

```bash
mvn -pl agentscope-harness test -Dtest=AgentSpawnToolCancellationTest -DfailIfNoTests=false
mvn -pl agentscope-harness test

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java
@@ -53,9 +53,13 @@ import java.util.concurrent.CompletionException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import reactor.core.Disposable;
+import reactor.core.Disposables;
 import reactor.core.publisher.Mono;
+import reactor.core.publisher.SignalType;
 
 /**
  * Simple subagent tool for agent-internal use. Much lighter than {@code SessionsTool}:
@@ -719,24 +723,44 @@ public class AgentSpawnTool {
                         Mono.<String>create(
                                 sink -> {
                                     CompletableFuture<Msg> bridge = new CompletableFuture<>();
+                                    AtomicBoolean outerCancelled = new AtomicBoolean(false);
+                                    Disposable.Swap innerSubscription = Disposables.swap();
 
-                                    execLocalSync(
-                                                    agent,
-                                                    sessionId,
-                                                    userId,
-                                                    task,
-                                                    spawned,
-                                                    runtimeContext)
-                                            .contextWrite(
-                                                    c -> reactor.util.context.Context.of(parentCtx))
-                                            .subscribe(
-                                                    bridge::complete,
-                                                    bridge::completeExceptionally,
-                                                    () -> {
-                                                        if (!bridge.isDone()) {
-                                                            bridge.complete(null);
-                                                        }
-                                                    });
+                                    sink.onCancel(
+                                            () -> {
+                                                outerCancelled.set(true);
+                                                innerSubscription.dispose();
+                                                bridge.cancel(false);
+                                            });
+
+                                    Disposable inner =
+                                            execLocalSync(
+                                                            agent,
+                                                            sessionId,
+                                                            userId,
+                                                            task,
+                                                            spawned,
+                                                            runtimeContext)
+                                                    .contextWrite(
+                                                            c ->
+                                                                    reactor.util.context.Context.of(
+                                                                            parentCtx))
+                                                    .doFinally(
+                                                            signal -> {
+                                                                if (signal == SignalType.CANCEL) {
+                                                                    interruptAgent(
+                                                                            agent, runtimeContext);
+                                                                }
+                                                            })
+                                                    .subscribe(
+                                                            bridge::complete,
+                                                            bridge::completeExceptionally,
+                                                            () -> {
+                                                                if (!bridge.isDone()) {
+                                                                    bridge.complete(null);
+                                                                }
+                                                            });
+                                    innerSubscription.update(inner);
 
                                     // Race against timeout without cancelling the bridge.
                                     // thenApply(m -> m) creates a dependent future so orTimeout
@@ -746,6 +770,9 @@ public class AgentSpawnTool {
 
                                     race.whenComplete(
                                             (msg, err) -> {
+                                                if (outerCancelled.get()) {
+                                                    return;
+                                                }
                                                 if (err != null) {
                                                     handleExecError(
                                                             err,
@@ -763,6 +790,17 @@ public class AgentSpawnTool {
                                                 }
                                             });
                                 }));
+    }
+
+    /** Interrupts a local subagent when its parent subscription is cancelled. */
+    private static void interruptAgent(Agent agent, RuntimeContext ctx) {
+        if (agent instanceof ReActAgent ra) {
+            ra.interrupt(ctx);
+            log.warn(
+                    "Sub-agent '{}' (id={}) was interrupted because agent_spawn was cancelled.",
+                    ra.getName(),
+                    ra.getAgentId());
+        }
     }
 
     /**
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/AgentSpawnToolCancellationTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/AgentSpawnToolCancellationTest.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


chickenlj:
Duplicate of https://github.com/agentscope-ai/agentscope-java/pull/2064.

PR #2064, opened 2026-07-08
fix(harness): interrupt orphan sub-agent when AgentSpawnTool parent subscription cancels (#2062)

### LINKED ISSUE #2062: [Bug]:AgentSpawnTool leaks orphan ReActAgent on parent subscription cancel

6or7:
## Summary

  Fix an orphan sub-agent leak in `AgentSpawnTool.execWithTimeoutPromotion`: when the parent's subscription to the
  `agentSpawn` Mono is cancelled (outer `Mono.timeout()`, user stop, or upstream Reactor cancel), the inner execution
  Mono — subscribed via fire-and-forget `.subscribe(...)` — was never disposed, so the spawned `ReActAgent` kept running
   indefinitely.

  Closes #2062.

  ## Root cause

  `execWithTimeoutPromotion` builds the result `Mono<String>` by wrapping an inner `Mono<Msg>` in a `MonoSink` +
  `CompletableFuture` bridge, then does plain `.subscribe(...)` on the inner Mono:

  ```java
  Disposable innerSub = inner.subscribe(
      bridge::complete,
      bridge::completeExceptionally,
      () -> { if (!bridge.isDone()) bridge.complete(null); });

  3. The returned Disposable is captured but never wired to the outer subscription, so Reactor cancel on the outer Mono
  never propagates into the inner one. Same shape as the core SubAgentTool bug fixed in 029cc55e.

  Fix

  3. Two-part, mirroring 029cc55e but adapted for AgentSpawnTool's bridge wrapper:

    - inner.doFinally(CANCEL -> interruptAgent(...)) — when the inner Mono is disposed, call ReActAgent.interrupt(ctx)
  to break the ReAct loop.
    - sink.onCancel(innerSub) — propagate outer Reactor cancel into the inner subscription so doFinally(CANCEL) actually
   fires.

  What this does NOT change

    - Timeout promotion (race.orTimeout → AdoptedTaskRunSpec) — terminates via sink.success(...), not Reactor cancel, so
   doFinally(CANCEL) doesn't fire and the agent continues. Pinned by AgentSpawnToolPromotionTest.
    - Async cooperation (timeout_seconds == 0 → taskRepository.putTask) — separate code path, untouched.

  Test plan

  Two new regression tests in agentscope-harness:

  | Test                           | Scenario                                            | Assertion
                                                                                                |
  |--------------------------------|-----------------------------------------------------|------------------------------
  ----------------------------------------------------------------------------------------------|
  | AgentSpawnToolOrphanCancelTest | outer Mono.timeout(2s) cancels mid-flight           | interrupt() called on
  sub-agent; file lines + model calls stop growing within 4s window (delta ≤ 1). Pre-fix delta was 25. |
  | AgentSpawnToolPromotionTest    | internal race.orTimeout fires before agent finishes | captured spec is
  AdoptedTaskRunSpec; its future() completes with the agent's reply; interrupt() is never called.           |

  Full harness suite: 608 tests pass, 0 failures, 3 pre-existing skips.

    - mvn -pl agentscope-harness test -Dtest=AgentSpawnToolOrphanCancelTest
    - mvn -pl agentscope-harness test -Dtest=AgentSpawnToolPromotionTest
    - mvn -pl agentscope-harness test (full suite)
    - mvn -pl agentscope-harness spotless:check

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java
@@ -55,7 +55,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import reactor.core.Disposable;
 import reactor.core.publisher.Mono;
+import reactor.core.publisher.SignalType;
 
 /**
  * Simple subagent tool for agent-internal use. Much lighter than {@code SessionsTool}:
@@ -720,16 +722,40 @@ public class AgentSpawnTool {
                                 sink -> {
                                     CompletableFuture<Msg> bridge = new CompletableFuture<>();
 
-                                    execLocalSync(
-                                                    agent,
-                                                    sessionId,
-                                                    userId,
-                                                    task,
-                                                    spawned,
-                                                    runtimeContext)
-                                            .contextWrite(
-                                                    c -> reactor.util.context.Context.of(parentCtx))
-                                            .subscribe(
+                                    Mono<Msg> inner =
+                                            execLocalSync(
+                                                            agent,
+                                                            sessionId,
+                                                            userId,
+                                                            task,
+                                                            spawned,
+                                                            runtimeContext)
+                                                    .contextWrite(
+                                                            c ->
+                                                                    reactor.util.context.Context.of(
+                                                                            parentCtx))
+                                                    .doFinally(
+                                                            signal -> {
+                                                                // Parent subscription was
+                                                                // cancelled (outer Mono.timeout,
+                                                                // user stop, or upstream Reactor
+                                                                // cancel). The fire-and-forget
+                                                                // subscribe below detaches the
+                                                                // inner execution from the parent
+                                                                // lifecycle, so without this
+                                                                // doFinally the sub-agent would
+                                                                // keep running as an orphan and
+                                                                // the eventual sink.success(...)
+                                                                // would be a no-op on the
+                                                                // already-cancelled sink.
+                                                                if (signal == SignalType.CANCEL) {
+                                                                    interruptAgent(
+                                                                            agent, runtimeContext);
+                                                                }
+                                                            });
+
+                                    Disposable innerSub =
+                                            inner.subscribe(
                                                     bridge::complete,
                                                     bridge::completeExceptionally,
                                                     () -> {
@@ -762,7 +788,33 @@ public class AgentSpawnTool {
                                                                     + textOf(msg));
                                                 }
                                             });
+
+                                    // Propagate parent cancellation to the inner fire-and-forget
+                                    // subscription. Without this, doFinally(CANCEL) above never
+                                    // fires because the inner is subscribed independently of the
+                                    // parent Mono.create sink.
+                                    sink.onCancel(innerSub);
                                 }));
+    }
+
+    /**
+     * Interrupts the sub-agent's reasoning loop when its parent tool-call subscription is
+     * cancelled. Mirrors the fix in core {@code SubAgentTool.interruptAgent} (commit
+     * {@code 029cc55e}, issue #1783) — see issue #2062 for the harness-side equivalent.
+     *
+     * <p>Only {@link ReActAgent} exposes {@code interrupt(RuntimeContext)}. Other {@link Agent}
+     * implementations are no-ops here; their inner execution will still be disposed by the caller's
+     * {@code sink.onCancel(innerSub::dispose)}, which is enough for non-looping agents.
+     */
+    private void interruptAgent(Agent agent, RuntimeContext ctx) {
+        if (agent instanceof ReActAgent ra) {
+            ra.interrupt(ctx);
+            log.warn(
+                    "Sub-agent '{}' (id={}) was interrupted because its parent tool call"
+                            + " subscription was cancelled.",
+                    ra.getName(),
+                    ra.getAgentId());
+        }
     }
 
     /**
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/AgentSpawnToolOrphanCancelTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/AgentSpawnToolOrphanCancelTest.java
@@ -1,0 +1,320 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent.tool;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.verify;
+
+import io.agentscope.core.ReActAgent;
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.core.agent.test.MockModel;
+import io.agentscope.core.message.TextBlock;
+import io.agentscope.core.message.ToolResultBlock;
+import io.agentscope.core.tool.AgentTool;
+import io.agentscope.core.tool.ToolCallParam;
+import io.agentscope.core.tool.Toolkit;
+import io.agentscope.harness.agent.middleware.SubagentEntry;
+import io.agentscope.harness.agent.subagent.DefaultAgentManager;
+import io.agentscope.harness.agent.subagent.SubagentFactory;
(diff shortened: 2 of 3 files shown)


6or7:
Hi maintainers — I'm the reporter of #2062 and submitted this fix shortly after filing the issue. I noticed #2063 by
  @mroldx addresses the same root cause; happy to coordinate.

  Quick comparison in case it helps the review:

  - **Smaller diff**: this PR only adds `doFinally(CANCEL)` + `interruptAgent()` + `sink.onCancel(innerSub)`, mirroring
  the canonical fix `029cc55e` already applied to core `SubAgentTool`. #2063 additionally introduces `AtomicBoolean
  outerCancelled` + `Disposable.Swap` + `bridge.cancel(false)` to handle the post-cancel `race.whenComplete` callback —
  not strictly required since the inner `Mono` is disposed before `bridge` completes.
  - **Extra regression coverage**: `AgentSpawnToolPromotionTest` verifies the timeout-promotion path
  (`AdoptedTaskRunSpec`) is **not** accidentally interrupted when the inner cancel hook fires — i.e., the new
  cancel-handler doesn't regress the promotion flow.

  Either PR works — if you prefer #2063 for any reason, please consider cherry-picking `AgentSpawnToolPromotionTest` so
  the promotion path stays protected.

PR #2069, opened 2026-07-09
fix(openai): make nativeStructuredOutput configurable

### LINKED ISSUE #2025: [Bug]:OpenAI native structured output cannot be disabled for compatible gateways

chickenlj:
## Summary

- Add configurable `nativeStructuredOutput` capability flag to `ChatModelBase`, reusing the same nullable-Boolean pattern established by `nativeStructuredOutputWithTools`
- Remove hardcoded `supportsNativeStructuredOutput() { return true; }` from `OpenAIChatModel`; default to `true` in Builder so existing users are unaffected
- Wire `nativeStructuredOutput` option through `OpenAIModelProvider` so it can be set via `ModelCreationContext`

This allows users on OpenAI-compatible gateways (vLLM, Ollama, etc.) that don't support `response_format` with `json_schema` to set `nativeStructuredOutput=false` and fall back to the `generate_response` tool approach.

## Test plan

- [x] `OpenAIChatModelTest.testNativeStructuredOutputDefaultTrue` — builder default is `true`
- [x] `OpenAIChatModelTest.testNativeStructuredOutputDisabled` — explicit `false` propagates
- [x] `OpenAIModelProviderTest.createUsesModelCreationContext` — option read from context
- [x] `ReActAgentStructuredOutputWithToolsTest` — no regression

Closes #2025

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ChatModelBase.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ChatModelBase.java
@@ -30,6 +30,8 @@ public abstract class ChatModelBase implements Model {
 
     private int contextWindowSize;
 
+    private Boolean nativeStructuredOutput;
+
     private Boolean nativeStructuredOutputWithTools;
 
     @Override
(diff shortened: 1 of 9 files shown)


RainYuY (approved):
LGTM

PR #2076, opened 2026-07-09
fix(harness): allow MemoryFlushMiddleware to flush for all Agent types

### LINKED ISSUE #2030: Coding Agent Work Dir Set Error
In CodingAgentFactory.create(), the system prompt uses resolveSandboxWorkingDir(), which reads SANDBOX_WORK_DIR, but the Docker filesystem spec hardcodes workspaceRoot to /home/agentscope/workspace.

This can make the prompt tell the agent to work in one directory while Docker executes and manages workspace state in another.

Example:

SANDBOX_WORK_DIR=/workspace

Actual:
- prompt working dir: /workspace
- Docker workspaceRoot: /home/agentscope/workspace

Expected:
- both should use /workspace

Suggested fix:

sandboxSpec.workspaceRoot(workingDir);

chickenlj:
## Summary

- `MemoryFlushMiddleware.doFlush()` had an `instanceof ReActAgent` guard that silently skipped `HarnessAgent` (which implements `Agent` directly, not via `ReActAgent`)
- This caused daily memory journals (`memory/YYYY-MM-DD.md`, MEMORY type) and consolidated `MEMORY.md` (CONTEXT type) to never be written for HarnessAgent-based agents
- Removed the type guard since `RuntimeContext.resolveAgentState()` accepts any `Agent`, and `HarnessAgent` provides `getAgentState()` via its delegate

Closes #2030

## Test plan

- [ ] Verify `mvn compile -pl agentscope-harness -am` passes
- [ ] Deploy with a HarnessAgent-based agent and confirm `tmg_sc_agent_memory` table receives new `memory_type=MEMORY` and `CONTEXT` records

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
@@ -15,7 +15,6 @@
  */
 package io.agentscope.harness.agent.middleware;
 
-import io.agentscope.core.ReActAgent;
 import io.agentscope.core.agent.Agent;
 import io.agentscope.core.agent.RuntimeContext;
 import io.agentscope.core.event.AgentEvent;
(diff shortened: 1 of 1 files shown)


chickenlj:
After further investigation, MemoryFlushMiddleware is registered on the inner ReActAgent (not HarnessAgent), so the `instanceof ReActAgent` check actually passes. The root cause of missing MEMORY/CONTEXT records lies elsewhere. Closing this PR.

PR #2101, opened 2026-07-10
feat(agui): support custom RuntimeContext in AguiAgentAdapter

ningmeng0503:
## Summary

  Allow callers to provide a custom `RuntimeContext` when running an agent through `AguiAgentAdapter` and `AguiRequestProcessor`, and let subclasses override the default context construction.

  ## Problem

  `AguiAgentAdapter.run(RunAgentInput)` always built the `RuntimeContext` internally, so callers could not:

  - Pass an existing runtime context that already carries request-scoped metadata.
  - Customize the default context via subclassing, because `buildRuntimeContext` was `private`.

  More importantly, because the adapter owned the context instance, any custom data written by downstream middleware or tools during the run was invisible to the upstream caller. The caller had no way
  to read that data back or use it to customize post-run processing.

  `AguiRequestProcessor` had the same limitation: it only exposed `process(RunAgentInput, String, String)`, so AG-UI request handlers could not inject a pre-built context either.

  ## Changes

  - Added `AguiAgentAdapter.run(RunAgentInput input, RuntimeContext runtimeContext)`.
    - Uses the supplied context when non-null; otherwise falls back to `buildRuntimeContext(input)`.
    - The caller keeps the same context reference, so values written by middleware/tools are observable after the run.
  - Changed `AguiAgentAdapter.buildRuntimeContext` from `private` to `protected` for subclass overrides.
  - Added `AguiRequestProcessor.process(RunAgentInput, String, String, RuntimeContext)`.
    - Forwards the context to the adapter; passing `null` preserves the original behavior.
  - Kept the original `run(RunAgentInput)` and `process(RunAgentInput, String, String)` methods unchanged for backward compatibility.

### DIFF
--- a/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java
+++ b/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java
@@ -108,6 +108,21 @@ public class AguiAgentAdapter {
      * @return A Flux of AG-UI events
      */
     public Flux<AguiEvent> run(RunAgentInput input) {
+        return run(input, null);
+    }
+
+    /**
+     * Run the agent with AG-UI protocol input and a custom runtime context.
+     *
+     * <p>If {@code runtimeContext} is {@code null}, a default context is built from the
+     * input via {@link #buildRuntimeContext(RunAgentInput)}. Otherwise the supplied
+     * context is passed directly to the agent.
+     *
+     * @param input The AG-UI run input
+     * @param runtimeContext The runtime context to use, or {@code null} to use the default
+     * @return A Flux of AG-UI events
+     */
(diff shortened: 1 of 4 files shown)


oss-maintainer (approved):
### Summary
Clean feature addition — allows callers to inject a custom `RuntimeContext` into `AguiAgentAdapter` and `AguiRequestProcessor`, with proper null-fallback and backward compatibility.

### Findings
- **[Info]** `AguiAgentAdapter.java:104` — The null-check + fallback pattern is correct and consistent.
- **[Info]** `AguiRequestProcessor.java:99` — Good that the original 3-arg method is preserved as a delegate.
- **[Info]** Tests cover all three paths (custom context, null fallback, subclass override) — well done.

### Verdict
Straightforward, well-tested, fully backward compatible. LGTM.

---
*Automated review by github-manager*

AgentScopeJavaBot (approved):
🤖 AI Review

Adds custom RuntimeContext support to AguiAgentAdapter. Zero behavioral change, fully backward compatible. 5 comprehensive tests. Clean design.

PR #2104, opened 2026-07-10
fix(models): terminate OpenAI stream on [DONE] sentinel

### LINKED ISSUE #2073: [Bug] OpenAI streaming: `[DONE]` is filtered out instead of terminating the stream — completion delayed until upstream closes the connection
**Describe the bug**

In `agentscope-extensions-model-openai`, `OpenAIClient#streamChatCompletion` treats the SSE sentinel line `[DONE]` as a data line to be *discarded* rather than as the stream termination signal:

```java
// io.agentscope.extensions.model.openai.OpenAIClient (2.0.0-RC5)
return transport.stream(httpRequest)
        .filter(data -> !data.equals("[DONE]"))   // <-- [DONE] silently dropped, never completes the Flux
        .<OpenAIResponse>handle((data, sink) -> { ... })
```

Because nothing calls `complete()` when `[DONE]` arrives, the `Flux<String>` returned by `HttpTransport#stream` only completes when the **underlying HTTP response body reaches EOF**, i.e. when the server actually closes the connection.

Many OpenAI-compatible gateways (LiteLLM-style proxies, corporate model gateways, nginx-fronted vLLM, etc.) send `data: [DONE]` and then **keep the connection open** (HTTP keep-alive) until an idle timeout — typically 60s. As a result:

- All streamed tokens arrive in real time, **but** the model call's reactive completion is delayed by ~60 seconds.
- Everything downstream that waits on the model Flux completion is blocked for that period: the ReAct turn does not advance, `AgentResultEvent` is not emitted, and `agent.streamEvents(...)`'s `onComplete` fires ~60s after the last token.
- From an application's perspective the agent looks "still running" for a minute after the LLM has clearly finished its reply.

Direct connections to `api.openai.com` may mask this (the server closes promptly), but per the SSE-over-OpenAI convention `data: [DONE]` **is** the end-of-stream marker and should terminate the stream deterministically, regardless of upstream connection lifecycle. Both the OpenAI Python and JS SDKs treat `[DONE]` as the terminator.

**To Reproduce**

1. Point `OpenAIChatModel` at any OpenAI-compatible gateway that keeps the connection alive after `[DONE]` (e.g. an nginx/keep-alive fronted proxy; easily simulated with a tiny mock server that sends a few chunks + `data: [DONE]` and then sleeps 60s before closing):

```java
OpenAIChatModel model = OpenAIChatModel.builder()
        .apiKey("...")
        .baseUrl("https://your-openai-compatible-gateway/v1")
        .modelName("gpt-4o-mini")   // any chat model
        .stream(true)
        .build();

ReActAgent agent = ReActAgent.builder().name("t").model(model).build();

long t0 = System.currentTimeMillis();
agent.stream(new UserMessage("say hi"))
        .doOnNext(msg -> System.out.println((System.currentTimeMillis() - t0) + "ms delta"))
        .doOnComplete(() -> System.out.println((System.currentTimeMillis() - t0) + "ms COMPLETE"))
        .blockLast();
```

2. Run and observe timestamps.
3. All deltas arrive within the first ~1–2s; `COMPLETE` fires only ~60s later (exactly when the gateway closes the idle connection).

Minimal mock upstream to reproduce deterministically (Python):

```python
# mock_openai.py — sends chunks + [DONE], then keeps the socket open for 90s
import socket, time
srv = socket.socket(); srv.bind(("127.0.0.1", 18080)); srv.listen(1)
while True:
    c, _ = srv.accept(); c.recv(65536)
    body = (
        'data: {"id":"1","choices":[{"delta":{"content":"hello"},"index":0}]}\n\n'
        'data: {"id":"1","choices":[{"delta":{},"finish_reason":"stop","index":0}]}\n\n'
        'data: [DONE]\n\n'
    )
    c.sendall(("HTTP/1.1 200 OK\r\nContent-Type: text/event-stream\r\n"
               "Transfer-Encoding: chunked\r\n\r\n").encode())
    c.sendall((hex(len(body))[2:] + "\r\n" + body + "\r\n").encode())
    time.sleep(90)          # simulate keep-alive: do NOT close / do NOT send terminal chunk
    c.sendall(b"0\r\n\r\n"); c.close()
```

Point `baseUrl` at `http://127.0.0.1:18080/v1` → deltas arrive instantly, `onComplete` arrives after 90s.

**Expected behavior**

Receiving `data: [DONE]` should complete the stream immediately (as the OpenAI streaming protocol defines), e.g.:

```java
return transport.stream(httpRequest)
        .takeUntil(data -> "[DONE]".equals(data == null ? null : data.trim()))
        .filter(data -> !data.equals("[DONE]"))
        .handle(...)
```

`takeUntil` also cancels the upstream subscription, which releases the pooled connection promptly instead of holding it until the server's idle timeout — strictly better for connection reuse as well. For upstreams that never send `[DONE]`, behavior is unchanged (falls back to EOF), so the change is backward-compatible.

**Error messages**

No exception is thrown — the failure mode is a silent ~60s delay between the last streamed token and the reactive completion (`onComplete` / `AgentResultEvent`), observable with the timestamps above.

- AgentScope-Java Version: 2.0.0-RC5 (`agentscope-extensions-model-openai` 2.0.0-RC5, `agentscope-harness` 2.0.0-RC5)
- Java Version: 21 (Microsoft OpenJDK 21.0.7)
- OS: Windows 11 (also reproduced on Linux)

**Additional context**

- Root cause is independent of the transport implementation: both `JdkHttpTransport` and `OkHttpTransport` faithfully emit SSE data lines until EOF; the missing `[DONE]` handling is in `OpenAIClient#streamChatCompletion`.
- Workaround we currently use: wrap the transport passed to `OpenAIChatModel.builder().httpTransport(...)` with a decorator whose `stream()` applies `takeUntil(line -> "[DONE]".equals(line.trim()))`. This restores sub-second completion after the last token.
- With `HarnessAgent` the impact is amplified: tool-calling turns wait on each model round's completion, so a multi-round ReAct conversation accumulates one idle-timeout delay per round.

logicwu0:
## AgentScope-Java Version

2.0.1-SNAPSHOT (main)

## Description

**Background.** `OpenAIClient#stream` treats the SSE `[DONE]` sentinel as a data line to be *discarded* (`.filter(data -> !data.equals("[DONE]"))`) rather than as the stream-termination signal. Nothing completes the `Flux` when `[DONE]` arrives, so it only completes when the underlying HTTP response body reaches EOF — i.e. when the server actually closes the connection.

**Impact.** Many OpenAI-compatible gateways (LiteLLM-style proxies, corporate model gateways, nginx-fronted vLLM, …) send `data: [DONE]` and then keep the connection open (HTTP keep-alive) until an idle timeout (~60s). As a result all tokens arrive in real time, but the model call's reactive completion is delayed by up to a minute: the ReAct turn does not advance, `AgentResultEvent` is not emitted, and `agent.streamEvents(...)`'s `onComplete` fires ~60s after the last token — a finished agent looks "still running". Direct `api.openai.com` connections mask this because the server closes promptly.

**Change.** Replace `filter` with `takeWhile` so `[DONE]` completes the stream deterministically, regardless of connection lifecycle. This matches the OpenAI Python/JS SDKs, which both treat `[DONE]` as the terminator. One-line logic change in `OpenAIClient#stream`.

**How to test.** Added `OpenAIStreamTerminationTest`: a fake `HttpTransport` emits two content chunks + `[DONE]`, then `Flux.never()` (simulating a gateway that never closes the connection). `StepVerifier` asserts the stream completes within 5s.
- Before the fix: the stream never completes → the test times out (~5.4s, `VerifySubscriber timed out`).
- After the fix: it completes in ~0.4s.

Full module suite: `Tests run: 379, Failures: 0, Errors: 0, Skipped: 10`.

Closes #2073

## Checklist

- [x] Related documentation has been updated (e.g. links, examples, etc.) — N/A (behavioral bug fix, no doc surface)

### DIFF
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/OpenAIClient.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/OpenAIClient.java
@@ -65,6 +65,9 @@ public class OpenAIClient {
 
     private static final Logger log = LoggerFactory.getLogger(OpenAIClient.class);
 
+    /** SSE sentinel emitted by the OpenAI streaming API to signal end of stream. */
+    private static final String SSE_DONE_MARKER = "[DONE]";
+
     /** Default base URL for OpenAI API. */
     public static final String DEFAULT_BASE_URL = "https://api.openai.com";
 
@@ -402,7 +405,7 @@ public class OpenAIClient {
                     // rather than merely filtering it out, otherwise completion is deferred until
                     // the underlying connection closes (keep-alive gateways may stall this for the
                     // full idle timeout even though the model has finished responding).
-                    .takeWhile(data -> !data.equals("[DONE]"))
+                    .takeWhile(data -> !SSE_DONE_MARKER.equals(data))
                     .<OpenAIResponse>handle(
                             (data, sink) -> {
                                 OpenAIResponse response = parseStreamData(data);
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/OpenAIClient.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/OpenAIClient.java
@@ -398,7 +398,11 @@ public class OpenAIClient {
                             .build();
 
             return transport.stream(httpRequest)
-                    .filter(data -> !data.equals("[DONE]"))
+                    // The SSE `[DONE]` sentinel terminates the stream: complete the Flux here
+                    // rather than merely filtering it out, otherwise completion is deferred until
+                    // the underlying connection closes (keep-alive gateways may stall this for the
+                    // full idle timeout even though the model has finished responding).
+                    .takeWhile(data -> !data.equals("[DONE]"))
                     .<OpenAIResponse>handle(
                             (data, sink) -> {
                                 OpenAIResponse response = parseStreamData(data);
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/test/java/io/agentscope/extensions/model/openai/OpenAIStreamTerminationTest.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/test/java/io/agentscope/extensions/model/openai/OpenAIStreamTerminationTest.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.extensions.model.openai;
+
+import io.agentscope.core.model.GenerateOptions;
+import io.agentscope.core.model.transport.HttpRequest;
+import io.agentscope.core.model.transport.HttpResponse;
+import io.agentscope.core.model.transport.HttpTransport;
+import io.agentscope.extensions.model.openai.dto.OpenAIMessage;
+import io.agentscope.extensions.model.openai.dto.OpenAIRequest;
+import java.time.Duration;
+import java.util.List;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Flux;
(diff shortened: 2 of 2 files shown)


oss-maintainer (commented):
### Summary
This PR modifies 2 file(s) (+99 -1).

Reviewed the changes and they look reasonable overall. No critical issues found.

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/OpenAIClient.java
```diff
@@ -398,7 +398,11 @@ public Flux<OpenAIResponse> stream(
                             .build();
 
             return transport.stream(httpRequest)
-                    .filter(data -> !data.equals("[DONE]"))
+                    // The SSE `[DONE]` sentinel terminates the stream: complete the Flux here
+                    // rather than merely filtering it out, otherwise completion is deferred until
+                    // the underlying connection closes (keep-alive gateways may stall this for the
+                    // full idle timeout even though the model has finished responding).
+                    .takeWhile(data -> !data.equals("[DONE]"))
```
AgentScopeJavaBot: [minor] Extract [DONE] as private static final SSE_DONE_MARKER constant for consistency.

AgentScopeJavaBot (approved):
🤖 AI Review

Changes [DONE] sentinel handling from filter to takeWhile for deterministic stream termination. Good defense-in-depth. Consider extracting constant.

logicwu0:
已抽为 `SSE_DONE_MARKER` 常量，感谢建议。

jujn (approved):

Issue #1560, opened 2026-06-01
[harness][sandbox] SandboxBackedFilesystem.uploadFiles fails with CreateProcess error=206 / E2BIG because file content is base64-embedded into the docker exec command line


cyrain-cheng:
## Summary

`SandboxBackedFilesystem.uploadFiles(...)` mirrors files into the Docker sandbox by **base64-encoding the whole file content and embedding it as a single argument of a `docker exec sh -c "..."` command line**. When the file is large enough (e.g. an accumulated session history `.jsonl`), the resulting command line exceeds the OS single-argument / command-line length limit and the process fails to start:

- **Windows**: `java.io.IOException: CreateProcess error=206, The filename or extension is too long` (command line limit ~32 KB).
- **Linux**: same root cause, fails with `E2BIG` once a single arg exceeds `MAX_ARG_STRLEN` (~128 KB). Higher threshold, so much rarer, but not immune.

This breaks the asynchronous session-log mirroring (`SessionTree.scheduleMirror` → `mirrorToFilesystem` → `uploadFiles`) during normal conversations once the session log grows.

## Affected version

- `agentscope-harness:1.1.0-SNAPSHOT`

## Root cause (source)

`io.agentscope.harness.agent.filesystem.sandbox.SandboxBackedFilesystem#uploadFiles`:

```java
String base64Content = Base64.getEncoder().encodeToString(content); // full file content
String escapedPath = shellSingleQuote(path);
String cmd =
        "mkdir -p $(dirname " + escapedPath + ") && "
        + "printf '%s' '" + base64Content + "' | base64 -d > " + escapedPath;
ExecResult result = active.exec(runtimeContext, cmd, null);
```

This `cmd` (containing the entire base64 blob, which is ~1.33x the original size) is then passed to `io.agentscope.harness.agent.sandbox.impl.docker.DockerSandbox#doExec`, which puts it verbatim as the last argument:

```java
List<String> cmd = new ArrayList<>();
cmd.add("docker"); cmd.add("exec");
cmd.add("-w"); cmd.add(workspaceRoot);
cmd.add(containerId);
cmd.add("sh"); cmd.add("-c");
cmd.add(command);              // <-- entire base64 blob ends up on the command line
ProcessBuilder pb = new ProcessBuilder(cmd);
Process process = pb.start(); // throws CreateProcess error=206 on Windows / E2BIG on Linux
```

[...]

PR #2148, opened 2026-07-12
fix(harness): stream Docker exec commands through stdin

### LINKED ISSUE #2135: [Bug]: DockerSandbox corrupts complex sh -c commands on Windows
## Summary

On a Windows host, `DockerSandbox#doExec` passes the full shell program as the last argument of:

```text
docker exec -w <workspaceRoot> <containerId> sh -c <command>
```

Commands containing command substitution and embedded quotes can be corrupted while crossing Java `ProcessBuilder` -> `docker.exe` -> container `sh -c`. This currently makes the standard Harness `write_file` path fail even for a tiny new file.

This is related to #1560 (large file content embedded in the command line), but the reproduction here is only a short preflight command and fails because of quoting, not command length.

## Environment

- Windows host
- Java 17
- Docker Desktop 29.2.0
- current `main` / AgentScope Java 2.0 Harness
- image: `ubuntu:24.04`

## Minimal reproduction

Create and start an official Docker sandbox, then attach it to `SandboxBackedFilesystem` and call:

```java
SandboxBackedFilesystem filesystem = new SandboxBackedFilesystem();
filesystem.setSandbox(sandbox);
WriteResult result = filesystem.write(
        runtimeContext,
        "filesystem-contract.txt",
        "jcode-filesystem-contract");
```

Expected:

```text
WriteResult.success
```

Actual:

```text
Failed to write file 'filesystem-contract.txt'
```

Calling the exact preflight command directly exposes the hidden error:

```java
sandbox.exec(
    runtimeContext,
    "if [ -e 'filesystem-contract.txt' ]; then echo 'EXISTS'; exit 1; fi; "
        + "mkdir -p \"$(dirname 'filesystem-contract.txt')\" 2>&1",
    60);
```

Actual container shell error:

```text
Command exited with code 2: 'filesystem-contract.txt') 2>&1: 1: Syntax error: end of file unexpected (expecting ")")
```

The same shell program succeeds when invoked manually inside the same `ubuntu:24.04` image.

## Source path

`BaseSandboxFilesystem.write()` creates the preflight:

```java
String checkCmd =
    "if [ -e " + escapedPath + " ]; then echo 'EXISTS'; exit 1; fi; "
        + "mkdir -p \"$(dirname " + escapedPath + ")\" 2>&1";
```

`DockerSandbox#doExec()` then appends that program directly as a Windows process argument:

```java
cmd.add("sh");
cmd.add("-c");
cmd.add(command);
new ProcessBuilder(cmd).start();
```

Docker events show the `sh -c` command reaching the container with damaged quoting. `BaseSandboxFilesystem.write()` also replaces the real preflight output with the generic `Failed to write file` message, which initially hides the cause.

## Impact

- Standard Harness `write_file` fails for Docker sandbox on Windows.
- `edit_file` and other complex shell-backed filesystem operations may be affected.
- Redis/MySQL cross-replica sandbox recovery cannot complete because the writer phase cannot create the workspace file.
- The failure is unrelated to Redis, snapshot persistence, or Windows host path mapping; it also reproduces in a standalone official Docker sandbox without distributed storage.

## Suggested direction

Avoid transporting a shell program as a complex `docker.exe` command-line argument on Windows. A platform-independent approach is to stream the program through stdin, for example:

```text
docker exec -i -w <workspaceRoot> <containerId> sh -s
```

and write the program to process stdin. This direction also aligns with #1560 and avoids both quoting corruption and OS command-line length limits.

Separately, preserve `checkResult.output()` and exit code in the `BaseSandboxFilesystem.write()` preflight failure so future backend failures remain diagnosable.

I can prepare a focused PR with Windows/Linux tests if maintainers agree with the stdin execution direction.

richuV:
## Summary

- run Docker sandbox shell programs through `docker exec -i ... sh -s`
- keep shell text out of the host process argument list
- preserve UTF-8 complex commands and avoid Windows command-line length limits
- add unit contracts plus an opt-in real Docker regression test

## Why

`docker exec ... sh -c <command>` puts the complete shell program in the Windows `CreateProcess` argument list. Complex nested quoting can be corrupted, and large filesystem scripts can exceed the host command-line limit.

The public Sandbox API and command semantics remain unchanged; only the Docker CLI transport changes.

## Tests

- `mvn -pl agentscope-harness spotless:check test -Dtest=DockerSandboxCommandTest,BaseSandboxFilesystemTest,HarnessAgentDistributedSandboxTest -Dsurefire.failIfNoSpecifiedTests=false`
  - 16 tests, 12 passed, 4 environment-gated skipped
- `AGENTSCOPE_DOCKER_INTEGRATION=true mvn -pl agentscope-harness test -Dtest=DockerSandboxCommandTest`
  - 3/3 passed on Windows with Docker Desktop
  - covers nested quotes, `$(dirname ...)`, paths with spaces, and a 100KB shell program
- downstream JCode acceptance against the patched `2.0.1-SNAPSHOT`
  - real Docker provider lifecycle 1/1
  - Redis + Docker fresh-JVM writer 1/1
  - Redis + Docker fresh-JVM reader 1/1

Closes #2135.
Addresses #1560.

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
@@ -57,7 +57,8 @@ import org.slf4j.LoggerFactory;
  *
  * <h2>Workspace Operations</h2>
  * <ul>
- *   <li>Exec: {@code docker exec -w <root> <containerId> sh -c <command>}</li>
+ *   <li>Exec: {@code docker exec -i -w <root> <containerId> sh -s}, with the command sent via
+ *       stdin</li>
  *   <li>PersistWorkspace: {@code docker exec <containerId> tar -cf - -C <root> .}</li>
  *   <li>HydrateWorkspace: {@code docker exec -i <containerId> tar -xf - -C <root>}</li>
  * </ul>
@@ -138,17 +139,7 @@ public class DockerSandbox extends AbstractBaseSandbox {
         String containerId = dockerState.getContainerId();
         String workspaceRoot = dockerState.getWorkspaceRoot();
 
-        List<String> cmd = new ArrayList<>();
-        cmd.add("docker");
-        cmd.add("exec");
-        cmd.add("-w");
-        cmd.add(workspaceRoot);
-        cmd.add(containerId);
-        cmd.add("sh");
-        cmd.add("-c");
-        cmd.add(command);
-
-        ProcessBuilder pb = new ProcessBuilder(cmd);
+        ProcessBuilder pb = new ProcessBuilder(buildExecCommand(containerId, workspaceRoot));
         Process process = pb.start();
 
         ExecutorService drainer =
@@ -169,6 +160,14 @@ public class DockerSandbox extends AbstractBaseSandbox {
                 drainer.submit(() -> readStream(process.getErrorStream(), OUTPUT_TRUNCATE_BYTES));
         drainer.shutdown();
 
+        try (OutputStream stdin = process.getOutputStream()) {
+            writeShellProgram(stdin, command);
+        } catch (IOException e) {
+            process.destroyForcibly();
+            drainer.shutdownNow();
+            throw e;
+        }
+
         boolean exited = process.waitFor(timeoutSeconds, TimeUnit.SECONDS);
         if (!exited) {
             process.destroyForcibly();
@@ -188,6 +187,14 @@ public class DockerSandbox extends AbstractBaseSandbox {
             throw new SandboxException.ExecException(exitCode, stdout, stderr);
         }
         return result;
+    }
+
+    static List<String> buildExecCommand(String containerId, String workspaceRoot) {
+        return List.of("docker", "exec", "-i", "-w", workspaceRoot, containerId, "sh", "-s");
+    }
+
+    static void writeShellProgram(OutputStream stdin, String command) throws IOException {
+        stdin.write(command.getBytes(StandardCharsets.UTF_8));
     }
 
     @Override
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandboxCommandTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandboxCommandTest.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent.sandbox.impl.docker;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.harness.agent.sandbox.ExecResult;
+import io.agentscope.harness.agent.sandbox.Sandbox;
+import io.agentscope.harness.agent.sandbox.WorkspaceSpec;
+import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+
+class DockerSandboxCommandTest {
+
+    @Test
+    void execUsesStdinInsteadOfEmbeddingComplexShellProgramInHostArguments() throws Exception {
+        String script =
+                "mkdir -p \"$(dirname '/workspace/a b/result.txt')\" && "
+                        + "printf '%s' \"nested 'quotes' and $HOME\" > '/workspace/a b/result.txt'";
+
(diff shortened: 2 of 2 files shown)


AgentScopeJavaBot (approved):
🤖 AI Review

Changes Docker sandbox command execution from sh -c to sh -s (stdin pipe), solving Windows command-line length limits, UTF-8 escaping issues, and process argument exposure. Well-scoped change (2 files, +108/-12) with comprehensive tests. Clean implementation, no issues found.

oss-maintainer:
⚠️ **CLA Not Signed**

Please sign the CLA at: https://cla-assistant.io/agentscope-ai/agentscope-java?pullRequest=2148

---

### Code Review (pending CLA)

**Summary:** Runs Docker sandbox shell programs through docker exec -i ... sh -s to avoid Windows command-line corruption and length limits.

**Findings:**
- **[Info]** Keeps shell text out of host process argument list. Correct approach.
- **[Info]** Preserves UTF-8 complex commands. Good.
- **[Info]** 16 tests + real Docker regression. Excellent coverage.

**Verdict:** Essential fix for Windows Docker deployments. Ready to merge once CLA is signed.

---
*Automated review by github-manager*

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
```diff
@@ -169,6 +160,14 @@ protected ExecResult doExec(RuntimeContext runtimeContext, String command, int t
                 drainer.submit(() -> readStream(process.getErrorStream(), OUTPUT_TRUNCATE_BYTES));
         drainer.shutdown();
 
+        try (OutputStream stdin = process.getOutputStream()) {
+            writeShellProgram(stdin, command);
```
oss-maintainer, line 164: **[Warning]** `writeShellProgram` runs on the calling thread before `process.waitFor(timeoutSeconds, ...)`, so the per-call timeout does not bound the stdin write. For a large shell program (this PR's own test uses ~100 KB) the pipe fills once `docker exec -i` stops draining it — e.g. the daemon is slow, the exec is failing/retrying, or the container is being killed — and `stdin.write(...)` can block long past `timeoutSeconds` while the output drainers just sit idle. Writing stdin from the same `drainer` executor (or a dedicated single-use future bounded by `timeoutSeconds`) keeps the documented timeout guarantee; the current shape turns a bounded `doExec` into a potentially unbounded one.

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
```diff
@@ -190,6 +189,14 @@ protected ExecResult doExec(RuntimeContext runtimeContext, String command, int t
         return result;
     }
 
+    static List<String> buildExecCommand(String containerId, String workspaceRoot) {
+        return List.of("docker", "exec", "-i", "-w", workspaceRoot, containerId, "sh", "-s");
```
oss-maintainer, line 193: **[Info]** Adding `-i` also re-points the container's *stdin* at the script, so any command inside the program that reads stdin (`git`, `mysql`, `ssh`, a bare `read`, `cat` with no file) now consumes the remaining script bytes instead of hitting an immediate EOF like it did with the previous `sh -c` form. For agent-generated tool calls that is a silent truncation risk. If the goal is only to keep argv small, `docker exec -i <cid> sh -s -- </dev/null`-style redirection is not possible from the host side, so either document the caveat in the class javadoc above or drain the script into a file first (`sh -c 'cat > /tmp/x && sh /tmp/x'`) — otherwise worth a comment noting the tradeoff was considered.

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
```diff
@@ -190,6 +189,14 @@ protected ExecResult doExec(RuntimeContext runtimeContext, String command, int t
...
+    }
+
+    static void writeShellProgram(OutputStream stdin, String command) throws IOException {
+        stdin.write(command.getBytes(StandardCharsets.UTF_8));
```
oss-maintainer, line 197: **[Info]** No trailing newline is appended. `sh -s` executes the last line without one in practice (dash/bash both handle EOF-without-LF), but callers building a program by concatenation are one `printf` away from a silently dropped final command; `stdin.write((command + "\n").getBytes(...))` costs nothing and makes the byte-for-byte assertion in the test explicit about what the shell actually receives.

agentscope-harness/src/test/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandboxCommandTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Test
+    void execHostArgumentsStayBoundedForLargeShellPrograms() throws Exception {
+        String script = "printf 'x%.0s' $(seq 1 100000)\n#" + "x".repeat(100_000);
+
+        List<String> command = DockerSandbox.buildExecCommand("container-id", "/workspace");
+        ByteArrayOutputStream stdin = new ByteArrayOutputStream();
+        DockerSandbox.writeShellProgram(stdin, script);
+
+        assertEquals(8, command.size());
```
oss-maintainer, line 58: **[Warning]** Both unit tests exercise the two new static helpers directly — no process is started — so they would still pass if `doExec` forgot to call `writeShellProgram` or kept the old `sh -c` argument list; the only test that covers the real wiring is `@EnabledIfEnvironmentVariable(AGENTSCOPE_DOCKER_INTEGRATION)`, which CI never runs. A cheap way to close the gap without Docker: spawn `sh -s` locally through the same code path (or expose the write behind a small seam) and assert stdout/exit code, so the regression this PR fixes stays fixed.

oss-maintainer (commented):
### Summary
Streams the shell program into `docker exec -i ... sh -s` instead of embedding it in `sh -c <command>`, which is the right fix for the host argv limits and quoting problems with agent-generated scripts, and the helper extraction + integration-test gate are sensible. Findings above are about the unbounded stdin write (the one real risk), stdin re-direction semantics for nested commands, and the fact that the unit tests do not cover the wiring they appear to cover.

Note: the `license/cla` check is still `pending`, so this review is a COMMENT only — please sign the CLA (see reminder) before a maintainer looks at merging.

---

PR #2272, opened 2026-07-17
fix(shell-execute): 修复 Windows 平台工作目录切换命令构造问题

### LINKED ISSUE #2271: [Bug]:window系统下执行任何命令工具 execute 总是出错

**Describe the bug**
window系统下执行命令工具 execute 总是出错，任何命令都是。经过AI分析发现原来是 ShellExecuteTool 代码中没有判断当前环境是Windows还是Linux导致。就是这一行用了 Unix 单引号语法，在 Windows cmd.exe 下会报错。

Image

需要改成如下：

Image

**To Reproduce**
window系统下执行命令工具 execute 总是出错，任何命令都是，如下图
Image

**Expected behavior**
Windows环境下可以正常执行cmd命令

**Error messages**
- stderr：文件名、目录名或卷标语法不正确

- AgentScope-Java Version:2.0.0-RC5
- Java Version: [e.g. 17]
- OS: [windows]

**Additional context**

liuche51:
- 根据操作系统判断使用不同的命令语法
- Windows 平台使用双引号和 & 连接命令
- Unix/Linux/macOS 保持使用单引号和 && 连接命令
- 防止工作目录路径中的特殊字符导致命令错误
Fixes #2271 
## AgentScope-Java Version

## Description

## Checklist

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
@@ -60,7 +60,14 @@ public class ShellExecuteTool {
                 return "Error: working_directory must be a relative path within the workspace"
                         + " (absolute paths, '~', and '..' are not allowed).";
             }
-            effectiveCommand = "cd '" + wd.replace("'", "'\\''") + "' && " + command;
+            String osName = System.getProperty("os.name").toLowerCase();
+            if (osName.contains("win")) {
+                // Windows cmd.exe only understands double quotes and uses & as separator
+                effectiveCommand = "cd \"" + wd.replace("\"", "\\\"") + "\" & " + command;
+            } else {
+                // Unix/Linux/macOS shell syntax
+                effectiveCommand = "cd '" + wd.replace("'", "'\\''") + "' && " + command;
+            }
         }
 
         ExecuteResponse result =
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/ShellExecuteToolTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/ShellExecuteToolTest.java
@@ -15,62 +15,265 @@
  */
 package io.agentscope.harness.agent.tool;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import io.agentscope.core.agent.RuntimeContext;
 import io.agentscope.harness.agent.filesystem.model.ExecuteResponse;
 import io.agentscope.harness.agent.filesystem.sandbox.AbstractSandboxFilesystem;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
 
-/** Unit tests for {@link ShellExecuteTool}. */
+/**
+ * Unit tests for {@link ShellExecuteTool}, with focus on the OS-aware command assembly and the
+ * nullable {@link Integer} timeout parameter.
+ */
 class ShellExecuteToolTest {
 
     private static final RuntimeContext RT = RuntimeContext.empty();
 
     private AbstractSandboxFilesystem sandbox;
     private ShellExecuteTool tool;
+    private String previousOsName;
 
     @BeforeEach
     void setUp() {
         sandbox = mock(AbstractSandboxFilesystem.class);
         tool = new ShellExecuteTool(sandbox);
+        previousOsName = System.getProperty("os.name");
+        when(sandbox.execute(any(), any(), anyInt())).thenReturn(new ExecuteResponse("", 0, false));
+    }
+
+    @AfterEach
+    void restoreOsName() {
+        if (previousOsName != null) {
+            System.setProperty("os.name", previousOsName);
+        } else {
+            System.clearProperty("os.name");
+        }
+    }
+
+    private void setOsName(String osName) {
+        System.setProperty("os.name", osName);
+    }
+
+    private String capturedCommand() {
+        ArgumentCaptor<String> cmd = ArgumentCaptor.forClass(String.class);
+        verify(sandbox).execute(any(), cmd.capture(), anyInt());
+        return cmd.getValue();
+    }
+
+    // --- timeout handling (Integer, nullable) ---
+
+    @Test
+    void nullTimeout_defaultsTo30() {
+        tool.execute(RT, "ls", null, null);
+
+        verify(sandbox).execute(eq(RT), any(), eq(30));
     }
 
     @Test
-    void execute_omittedTimeout_defaultsTo30() {
-        when(sandbox.execute(eq(RT), eq("ls"), eq(30)))
-                .thenReturn(new ExecuteResponse("out", 0, false));
+    void positiveTimeout_isForwardedAsIs() {
+        tool.execute(RT, "ls", null, 90);
 
-        String result = tool.execute(RT, "ls", null, null);
-
-        assertTrue(result.contains("Exit code: 0"));
-        verify(sandbox).execute(RT, "ls", 30);
+        verify(sandbox).execute(eq(RT), any(), eq(90));
     }
 
     @Test
-    void execute_explicitTimeout_isPassedThrough() {
-        when(sandbox.execute(eq(RT), eq("ls"), eq(90)))
-                .thenReturn(new ExecuteResponse("out", 0, false));
+    void zeroTimeout_fallsBackToDefault30() {
+        tool.execute(RT, "ls", null, 0);
 
-        String result = tool.execute(RT, "ls", null, 90);
-
-        assertTrue(result.contains("Exit code: 0"));
-        verify(sandbox).execute(RT, "ls", 90);
+        verify(sandbox).execute(eq(RT), any(), eq(30));
     }
 
     @Test
-    void execute_withWorkingDirectory_prefixesCd() {
-        when(sandbox.execute(eq(RT), eq("cd 'sub' && ls"), eq(30)))
-                .thenReturn(new ExecuteResponse("out", 0, false));
+    void negativeTimeout_fallsBackToDefault30() {
+        tool.execute(RT, "ls", null, -5);
 
-        String result = tool.execute(RT, "ls", "sub", null);
+        verify(sandbox).execute(eq(RT), any(), eq(30));
+    }
+
+    // --- working_directory handling ---
+
+    @Test
+    void nullWorkingDirectory_passesCommandUnchanged() {
+        tool.execute(RT, "ls -la", null, 10);
+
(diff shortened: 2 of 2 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
```diff
@@ -60,7 +60,14 @@ public String execute(
                 return "Error: working_directory must be a relative path within the workspace"
                         + " (absolute paths, '~', and '..' are not allowed).";
             }
-            effectiveCommand = "cd '" + wd.replace("'", "'\\''") + "' && " + command;
+            String osName = System.getProperty("os.name").toLowerCase();
+            if (osName.contains("win")) {
+                // Windows cmd.exe only understands double quotes and uses & as separator
+                effectiveCommand = "cd \"" + wd.replace("\"", "\\\"") + "\" & " + command;
```
AgentScopeJavaBot, line 74: [major] **Bug: `&` should be `&&` on Windows.** In `cmd.exe`, `&` runs commands sequentially regardless of exit code, while `&&` runs the second command only if the first succeeds — matching the Unix `&&` behavior used in the else-branch. With `&`, if `cd` fails (e.g., the directory does not exist or the path is invalid), the user's command will still execute but in the *wrong* working directory — a silent and potentially dangerous failure. `cmd.exe` fully supports `&&` for conditional execution, so this should be changed to:

```java
effectiveCommand = "cd \"" + wd.replace("\"", "\\\"") + "\" && " + command;
```

This ensures consistent short-circuit behavior across both platforms.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
```diff
@@ -60,7 +60,14 @@ public String execute(
                 return "Error: working_directory must be a relative path within the workspace"
                         + " (absolute paths, '~', and '..' are not allowed).";
             }
-            effectiveCommand = "cd '" + wd.replace("'", "'\\''") + "' && " + command;
+            String osName = System.getProperty("os.name").toLowerCase();
```
AgentScopeJavaBot, line 71: [recommended] **Architectural concern: OS detection at the wrong abstraction layer.** `System.getProperty("os.name")` returns the *JVM host* OS, but `ShellExecuteTool` delegates execution to `sandbox.execute()`, which could be a `SandboxBackedFilesystem` (remote Linux sandbox) rather than `LocalFilesystemWithShell` (local process). If the JVM runs on Windows while the sandbox target is Linux, this tool would construct `cd "dir" & cmd` syntax and send it to a Linux shell — where `&` backgrounds `cd` instead of chaining, breaking the working directory change entirely.

Currently `LocalFilesystemWithShell.execute()` (line 321) already performs its own OS detection to select `cmd.exe` vs `sh`, so the two checks happen to agree for the local case. But for remote sandboxes they would diverge.

Consider either:
1. Exposing a method like `shellType()` on `AbstractSandboxFilesystem` (returning `CMD` or `SH`) so the tool constructs commands based on the *target* environment, or
2. Moving the `cd`-prefixing logic into each `execute()` implementation where the shell type is definitively known.

This may not need to block the current PR if the Windows + remote-sandbox scenario is not a supported configuration today, but should be tracked as a follow-up.

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR fixes issue #2271 by adding Windows-aware command construction in `ShellExecuteTool` when a `working_directory` is specified. Previously, the tool always used Unix shell syntax (`cd '...' && cmd`), which fails under `cmd.exe /c` on Windows because `cmd.exe` does not interpret single quotes and uses different command-chaining semantics. The fix correctly branches on the OS to produce `cd "..." & cmd` for Windows and retains the original Unix syntax otherwise. The general approach is sound and consistent with how `LocalFilesystemWithShell.execute()` already detects the OS to choose between `cmd.exe` and `sh`. However, there is one logical bug — using `&` instead of `&&` on Windows — and one architectural concern about the layer at which OS detection occurs.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
AgentScopeJavaBot, line 74: [major] **Bug: `&` should be `&&` on Windows.** In `cmd.exe`, `&` runs commands sequentially regardless of exit code, while `&&` runs the second command only if the first succeeds — matching the Unix `&&` behavior used in the else-branch. With `&`, if `cd` fails (e.g., the directory does not exist or the path is invalid), the user's command will still execute but in the *wrong* working directory — a silent and potentially dangerous failure. `cmd.exe` fully supports `&&` for conditional execution, so this should be changed to:

```java
effectiveCommand = "cd \"" + wd.replace("\"", "\\\"") + "\" && " + command;
```

This ensures consistent short-circuit behavior across both platforms.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
AgentScopeJavaBot, line 71: [recommended] **Architectural concern: OS detection at the wrong abstraction layer.** `System.getProperty("os.name")` returns the *JVM host* OS, but `ShellExecuteTool` delegates execution to `sandbox.execute()`, which could be a `SandboxBackedFilesystem` (remote Linux sandbox) rather than `LocalFilesystemWithShell` (local process). If the JVM runs on Windows while the sandbox target is Linux, this tool would construct `cd "dir" & cmd` syntax and send it to a Linux shell — where `&` backgrounds `cd` instead of chaining, breaking the working directory change entirely.

Currently `LocalFilesystemWithShell.execute()` (line 321) already performs its own OS detection to select `cmd.exe` vs `sh`, so the two checks happen to agree for the local case. But for remote sandboxes they would diverge.

Consider either:
1. Exposing a method like `shellType()` on `AbstractSandboxFilesystem` (returning `CMD` or `SH`) so the tool constructs commands based on the *target* environment, or
2. Moving the `cd`-prefixing logic into each `execute()` implementation where the shell type is definitively known.

This may not need to block the current PR if the Windows + remote-sandbox scenario is not a supported configuration today, but should be tracked as a follow-up.

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR fixes issue #2271 by adding Windows-aware command construction in `ShellExecuteTool` when a `working_directory` is specified. Previously, the tool always used Unix shell syntax (`cd '...' && cmd`), which fails under `cmd.exe /c` on Windows because `cmd.exe` does not interpret single quotes and uses different command-chaining semantics. The fix correctly branches on the OS to produce `cd "..." & cmd` for Windows and retains the original Unix syntax otherwise. The general approach is sound and consistent with how `LocalFilesystemWithShell.execute()` already detects the OS to choose between `cmd.exe` and `sh`. However, there is one logical bug — using `&` instead of `&&` on Windows — and one architectural concern about the layer at which OS detection occurs.

jujn:
Please fix the review from AgentScopeJavaBot.

oss-maintainer:
**CLA Not Signed**

@liuche51 please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.

---

liuche51:
> Please fix the review from AgentScopeJavaBot.

yes,I did

oss-maintainer:

```bash
git fetch origin
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
```diff
@@ -68,7 +68,14 @@ public String execute(
...
+            if (osName.contains("win")) {
+                // Windows cmd.exe only understands double quotes and uses & as separator
+                effectiveCommand = "cd \"" + wd.replace("\"", "\\\"") + "\" & " + command;
```
oss-maintainer, line 74: **[Critical]** `cmd.exe` does not treat `\"` as an escaped double quote — inside a quoted argument a backslash is literal and the following `"` **closes** the string. So `wd.replace("\"", "\\\"")` does not make the path safe; the remainder of `wd` escapes the quoted context and is re-parsed by cmd, where `&`, `|`, `<`, `>`, `^` and `%VAR%` are metacharacters. Since `wd` is model-supplied, please switch to cmd-safe handling: reject `wd` containing `"`, `&`, `|`, `<`, `>`, `^`, `` ` `` or `%` outright (return the existing validation error), and use `""` doubling if embedded quotes must be supported at all.

oss-maintainer, line 74: **[Warning]** Single `&` runs `command` even when `cd` fails, so a stale or misspelled `working_directory` silently executes in the sandbox default directory instead of the requested one — the Unix branch below uses `&&` and fails fast. cmd.exe supports `&&` as well; suggest `"cd \"" + wd + "\" && " + command` to keep the two branches behaviourally identical.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
```diff
@@ -68,7 +68,14 @@ public String execute(
                 return "Error: working_directory must be a relative path within the workspace"
                         + " (absolute paths, '~', and '..' are not allowed).";
             }
-            effectiveCommand = "cd '" + wd.replace("'", "'\\''") + "' && " + command;
+            String osName = System.getProperty("os.name").toLowerCase();
```
oss-maintainer, line 71: **[Warning]** Two issues with the OS probe: (1) `System.getProperty("os.name")` can be `null` in a hardened/overridden JVM, which would NPE on `.toLowerCase()` — use `System.getProperty("os.name", "")`; (2) `os.name` describes the **JVM host**, not the shell that `AbstractSandboxFilesystem.execute()` actually spawns. For a Docker/Kubernetes sandbox running a Linux image on a Windows host this emits cmd.exe syntax into `/bin/sh`, i.e. this PR's bug reappears inverted. Would it be better to let the filesystem backend declare its shell family (a `shellKind()`/capability on `AbstractSandboxFilesystem`, or a field on the sandbox config) and keep `os.name` only as the local fallback?

oss-maintainer (commented):
### Summary
The intent is right — `cd '…' && …` genuinely breaks on Windows, and the test coverage added here is thorough. However the cmd.exe quoting/separation semantics are not correct as written, so the Windows branch is still unsafe, and the branch predicate keys off the wrong signal.

### Findings
- **[Critical]** `agentscope-harness/.../ShellExecuteTool.java:74` — `\"` is not a cmd.exe escape; `wd` can break out of the quoted argument into cmd metacharacters
- **[Warning]** `agentscope-harness/.../ShellExecuteTool.java:74` — `&` continues after a failed `cd`; use `&&` to match the Unix branch
- **[Warning]** `agentscope-harness/.../ShellExecuteTool.java:71` — possible NPE on `os.name`, and host-OS detection is wrong for container-backed sandboxes

### Suggestions
1. Validate `wd` against a deny-list of cmd metacharacters and return the existing relative-path error message rather than trying to escape.
2. Extract the assembly into a small helper taking an explicit `ShellKind { POSIX, CMD }` parameter; that also lets the new tests drive both branches without mutating the global `os.name` system property (test-order sensitive under a parallel JUnit configuration).
3. Consider `ProcessBuilder`-style argument lists (or the sandbox API accepting `argv`) as the durable fix for the whole class of quoting bugs.

### Cross-repo Note
Please check the same `cd`-prefix pattern in the Docker/Kubernetes sandbox filesystem implementations — if the wrapper is built host-side there too, the Windows path handling needs the same fix.

Note: this PR currently shows `mergeable_state=dirty`; a rebase onto `main` will be needed.

---

PR #2300, opened 2026-07-19
fix(harness): return absolute paths from LocalFilesystem write results

### LINKED ISSUE #2266: [Bug]:AI创建的文件实际路径和返回给用户看到的不一致
场景：
我让AI给我创建一个文件，然后确实也创建了。说创建路径在 D:\Java\agentscopeWorkSpace\hello_world.py  实际上根本没有，而是生成到了 D:\Java\agentscopeWorkSpace\web-user\hello_world.py 
Image

Image
问了AI分析原因，说是框架的 LocalFilesystem.write() 方法返回存在Bug导致。  

AI分析如下：

这是 **AgentScope 框架的一个路径报告 Bug**，不是你的使用方式有问题。核心原因有两个：

## 1. 用户命名空间隔离把文件写到了 `web-user/` 下

你的 [`AgentChatService.chatStream()`](file:///D:/Java/Project2/agentScopeDemo/src/main/java/com/example/myagentdemo/service/AgentChatService.java#L75-L92) 中，当请求没有带 `userId` 时，默认设为 `"web-user"`：

```java
if (userId == null || userId.isBlank()) {
    userId = "web-user";
}
        .sessionId(sid)
        .userId(uid)
        .build();
```

然后 AgentScope 的 `HarnessAgent` 默认使用 **`IsolationScope.USER`** 做文件系统隔离（见 [`IsolationScope.java`](file:///D:/Java/agentscope-java/agentscope-harness/src/main/java/io/agentscope/harness/agent/IsolationScope.java#L71-L115)）。它的 `NamespaceFactory` 会把所有相对路径前加上 `userId`：

```java
case USER -> rc -> {
    String uid = rc.getUserId();
    if (uid != null && !uid.isBlank()) {
        return List.of(uid);   // 这里就是 "web-user"
    }
    ...
};
```

所以在 `LocalFilesystem` 内部，`hello_world.py` 被解析成 `web-user/hello_world.py`，实际写入：

```text
D:\Java\agentscopeWorkSpace\web-user\hello_world.py
```

## 2. 但工具返回给 AI 的路径没有包含这个前缀（Bug 所在）

真正的问题在 AgentScope 源码的 [`LocalFilesystem.write()`](file:///D:/Java/agentscope-java/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java#L322-L342)：

```java
@Override
public WriteResult write(RuntimeContext runtimeContext, String filePath, String content) {
    Path resolved = resolvePath(runtimeContext, filePath);   // 已解析为 web-user/hello_world.py
    ...
    Files.writeString(resolved, content, StandardCharsets.UTF_8);
    return WriteResult.ok(filePath);   // ❌ 这里返回的是原始的 "hello_world.py"
}
```

它**写的是解析后的路径**，但**返回的是原始逻辑路径**。而 `WriteResult` 的注释明确说 `path` 应该是 *absolute path of written file*（写入文件的绝对路径）。

[`FilesystemTool.writeFile()`](file:///D:/Java/agentscope-java/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/FilesystemTool.java#L82-L88) 又把这个错误路径原样返回给 AI：

```java
return r.isSuccess() ? "Written to " + r.path() : "Error: " + r.error();
```

AI 拿到 `"hello_world.py"` 后，再自己拼成 `D:\Java\agentscopeWorkSpace\hello_world.py` 告诉你，就造成了你看到的“路径对不上”。

## 现在你的文件在哪

真实位置是：

```text
D:\Java\agentscopeWorkSpace\web-user\hello_world.py
```

## 怎么解决

有三种办法，按推荐程度排列：

### 方案 A：修正 AgentScope 框架源码（最彻底）

如果你本地会重新编译 `agentscope-java`，把 [`LocalFilesystem.java`](file:///D:/Java/agentscope-java/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java#L322-L342) 里的 `write()` 改成返回解析后的路径：

```java
return WriteResult.ok(resolved.toString());
```

同样建议检查 `edit()`、`delete()`、`move()`、`uploadFiles()`，它们也有类似问题（返回原始 path 而不是 resolved path）。

jialiuyang:
## AgentScope-Java Version

2.0.x (main)

## Description

`WriteResult` / `EditResult` document that `path` is the absolute on-disk path, but `LocalFilesystem.write` / `edit` / `delete` / `move` returned the caller-supplied logical path. With namespace isolation, tools then reported a path that did not match the real file (e.g. workspace `hello_world.py` vs `web-user/hello_world.py`).

This change returns the resolved absolute path after `resolvePath`.

Fixes #2266

## Checklist

- [x] Targeted tests passing (`LocalFilesystemWritePathTest`, `FilesystemDeleteMoveExistsTest`)
- [x] Related documentation has been updated — N/A

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
@@ -335,7 +335,7 @@ public class LocalFilesystem implements AbstractFilesystem {
                 Files.createDirectories(resolved.getParent());
             }
             Files.writeString(resolved, content, StandardCharsets.UTF_8);
-            return WriteResult.ok(filePath);
+            return WriteResult.ok(absoluteResultPath(resolved));
         } catch (IOException e) {
             return WriteResult.fail("Error writing file '" + filePath + "': " + e.getMessage());
         }
@@ -378,7 +378,7 @@ public class LocalFilesystem implements AbstractFilesystem {
             int occurrences = (int) result[1];
 
             Files.writeString(resolved, newContent, StandardCharsets.UTF_8);
-            return EditResult.ok(filePath, occurrences);
+            return EditResult.ok(absoluteResultPath(resolved), occurrences);
         } catch (IOException e) {
             return EditResult.fail("Error editing file '" + filePath + "': " + e.getMessage());
         } finally {
@@ -527,7 +527,7 @@ public class LocalFilesystem implements AbstractFilesystem {
         AbstractFilesystem.validatePath(path);
         Path resolved = resolvePath(runtimeContext, path);
         if (!Files.exists(resolved)) {
-            return WriteResult.ok(path); // idempotent
+            return WriteResult.ok(absoluteResultPath(resolved)); // idempotent
         }
         try {
             if (Files.isDirectory(resolved)) {
@@ -565,7 +565,7 @@ public class LocalFilesystem implements AbstractFilesystem {
                 Files.createDirectories(to.getParent());
             }
             Files.move(from, to, StandardCopyOption.REPLACE_EXISTING);
-            return WriteResult.ok(toPath);
+            return WriteResult.ok(absoluteResultPath(to));
         } catch (IOException e) {
             return WriteResult.fail(
                     "Error moving '" + fromPath + "' to '" + toPath + "': " + e.getMessage());
@@ -582,6 +582,11 @@ public class LocalFilesystem implements AbstractFilesystem {
         } catch (SecurityException e) {
             return false;
         }
+    }
+
+    /** Absolute on-disk path for WriteResult/EditResult (matches their javadoc contract). */
+    private static String absoluteResultPath(Path resolved) {
+        return resolved.toAbsolutePath().normalize().toString();
     }
 
     // ==================== Path resolution ====================
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
@@ -545,7 +545,7 @@ public class LocalFilesystem implements AbstractFilesystem {
             } else {
                 Files.delete(resolved);
             }
-            return WriteResult.ok(path);
+            return WriteResult.ok(absoluteResultPath(resolved));
         } catch (IOException e) {
             return WriteResult.fail("Error deleting '" + path + "': " + e.getMessage());
         }
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/LocalFilesystemWritePathTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/LocalFilesystemWritePathTest.java
@@ -92,4 +92,20 @@ class LocalFilesystemWritePathTest {
         assertEquals(expected.toString(), result.path());
         assertTrue(Files.exists(expected));
     }
+
+    @Test
+    void delete_returnsAbsolutePathIncludingNamespace(@TempDir Path workspace) throws Exception {
+        Path nsFile = workspace.resolve("web-user").resolve("gone.txt");
+        Files.createDirectories(nsFile.getParent());
+        Files.writeString(nsFile, "x");
+
+        LocalFilesystem fs =
+                new LocalFilesystem(workspace, LocalFsMode.ROOTED, PathPolicy.empty(), 10, USER_NS);
+        RuntimeContext rc = RuntimeContext.builder().userId("web-user").build();
+
+        WriteResult result = fs.delete(rc, "gone.txt");
+        assertTrue(result.isSuccess(), () -> "delete failed: " + result.error());
+        assertEquals(nsFile.toAbsolutePath().normalize().toString(), result.path());
+        assertTrue(Files.notExists(nsFile));
+    }
 }
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/LocalFilesystemWritePathTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/LocalFilesystemWritePathTest.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent.filesystem;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.harness.agent.filesystem.local.LocalFilesystem;
+import io.agentscope.harness.agent.filesystem.model.EditResult;
+import io.agentscope.harness.agent.filesystem.model.WriteResult;
+import io.agentscope.harness.agent.filesystem.remote.store.NamespaceFactory;
+import io.agentscope.harness.agent.workspace.LocalFsMode;
+import io.agentscope.harness.agent.workspace.PathPolicy;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * Regression for issue #2266: WriteResult/EditResult must report the absolute on-disk path
+ * after namespace isolation, not the caller-supplied logical path.
+ */
+class LocalFilesystemWritePathTest {
+
+    private static final NamespaceFactory USER_NS = rc -> List.of("web-user");
+
+    @Test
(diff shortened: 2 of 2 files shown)


AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR fixes issue #2266 by making `LocalFilesystem`'s write/edit/delete/move operations return the **absolute on-disk path** (after namespace resolution) in `WriteResult`/`EditResult`, instead of the caller-supplied logical path. A new private helper `absoluteResultPath(Path)` centralizes the `toAbsolutePath().normalize().toString()` logic. The fix is clean and well-motivated — the `WriteResult` javadoc already states `path` should be the "absolute path of written file," and namespace-prefixed paths were previously being returned as raw relative inputs. However, there is one missed spot inside the `delete` method, and the new test file lacks coverage for `delete` entirely.

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR fixes issue #2266 by making `LocalFilesystem`'s write/edit/delete/move operations return the **absolute on-disk path** (after namespace resolution) in `WriteResult`/`EditResult`, instead of the caller-supplied logical path. A new private helper `absoluteResultPath(Path)` centralizes the `toAbsolutePath().normalize().toString()` logic. The fix is clean and well-motivated — the `WriteResult` javadoc already states `path` should be the "absolute path of written file," and namespace-prefixed paths were previously being returned as raw relative inputs. However, there is one missed spot inside the `delete` method, and the new test file lacks coverage for `delete` entirely.

oss-maintainer (approved):
### Summary
This PR contains 153 lines of changes. Review in progress.

---

jialiuyang:
Thanks for the review feedback — addressed in the latest commits:

1. `LocalFilesystem.delete` success path now returns `absoluteResultPath(resolved)`  
2. Added `delete_returnsAbsolutePathIncludingNamespace` coverage in `LocalFilesystemWritePathTest`.

Happy to adjust further if needed.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
```diff
@@ -584,6 +584,11 @@ public boolean exists(RuntimeContext runtimeContext, String path) {
         }
     }
 
+    /** Absolute on-disk path for WriteResult/EditResult (matches their javadoc contract). */
+    private static String absoluteResultPath(Path resolved) {
+        return resolved.toAbsolutePath().normalize().toString();
+    }
```
oss-maintainer, line 590: **[Warning]** "Absolute on-disk path" is mode-dependent here, and #2266 is only actually fixed for `LocalFsMode.ROOTED`/`UNRESTRICTED`. In `SANDBOXED` mode `resolveSandboxed()` re-roots every key under the sandbox `cwd`, so `absoluteResultPath(resolved)` returns a *sandbox-internal* path (`/workspace/web-user/hello_world.py`) that still does not match what the user sees on the host — i.e. the exact class of mismatch the issue reports. `CompositeFilesystem` also masks the new value whenever a route prefix was applied (`WriteResult.ok(filePath)` at `CompositeFilesystem.java:234`/`:469`/`:486`, and `OverlayFilesystem.java:270`), so an agent on a routed/overlay workspace keeps receiving the logical path from the very same tool call.

Could you either (a) state the contract in the `WriteResult`/`EditResult` javadoc as "absolute path as seen by the backing filesystem (host path in local modes, sandbox path in SANDBOXED, caller path when routed through Composite/Overlay)", or (b) thread the host mapping back through those wrappers? As written the record javadoc ("absolute path of written file") reads stronger than what the code guarantees, which is what the reporting user's model latched onto in the first place.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
```diff
@@ -335,7 +335,7 @@ public WriteResult write(RuntimeContext runtimeContext, String filePath, String
                 Files.createDirectories(resolved.getParent());
             }
             Files.writeString(resolved, content, StandardCharsets.UTF_8);
-            return WriteResult.ok(filePath);
+            return WriteResult.ok(absoluteResultPath(resolved));
```
oss-maintainer, line 338: **[Warning]** This value is model-visible, not just programmatic: `FilesystemTool.writeFile` renders `"Written to " + r.path()` (`FilesystemTool.java:148`) and `editFile` renders `"Edited " + r.path()` (`:175`), so after this change every write/edit puts the host workspace root **and the per-user namespace segment** into the assistant transcript, the persisted session state, and anything trained/exported from it. That is the intended user-facing fix for #2266, but it runs opposite to the direction other recent changes in this area take (keeping host paths out of what the agent sees).

Two consequences worth a conscious decision:
1. The model will start quoting host paths back at users, and the path is only reproducible on that one host — a session restored on another node (distribution module) will show a stale/incorrect location.
2. `WorkspacePathNormalizer` exists precisely to keep the model in workspace-relative coordinates; this change reintroduces absolute paths on the output side of the same tool pair.

If keeping the tool surface relative is acceptable, the smaller change is to leave `WriteResult.path()` as-is and have `FilesystemTool` report the *resolved* path it derived (or to add `absolutePath()` alongside `path()` on the result records so programmatic callers get the on-disk truth without it leaking into tool output).

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
```diff
@@ -527,7 +527,7 @@ public WriteResult delete(RuntimeContext runtimeContext, String path) {
         AbstractFilesystem.validatePath(path);
         Path resolved = resolvePath(runtimeContext, path);
         if (!Files.exists(resolved)) {
-            return WriteResult.ok(path); // idempotent
+            return WriteResult.ok(absoluteResultPath(resolved)); // idempotent
```
oss-maintainer, line 530: **[Info]** The idempotent branch now returns an absolute path for a file that does not exist. That is defensible ("this is the path you asked to delete, and it is gone"), but it is the one case where `WriteResult.path()` can point at nothing on disk, so any caller that treats a successful delete's path as verifiable (`Files.exists`, or a subsequent `write` to the same location) needs to tolerate that. A one-line `@implNote` here would save the next reader a bug hunt — and it would be worth pinning with a test (`delete` a missing path → success, path is the namespaced absolute form).

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/LocalFilesystemWritePathTest.java
```diff
@@ -0,0 +1,111 @@
...
+        Path expected =
+                workspace
+                        .resolve("web-user")
+                        .resolve("hello_world.py")
+                        .toAbsolutePath()
+                        .normalize();
+        assertEquals(expected.toString(), result.path());
+        assertTrue(Files.exists(Path.of(result.path())));
+        assertEquals("print('hi')\n", Files.readString(expected));
+    }
```
oss-maintainer, line 60: **[Info]** Nice regression suite, and deriving `expected` from `workspace.resolve("web-user")` (rather than asserting a literal) is the right call. Coverage gap: all four tests construct `LocalFsMode.ROOTED`. The behaviour that actually differs by mode is `SANDBOXED` (`resolveSandboxed` re-roots under `cwd` and strips a Windows drive prefix, so the reported path is the sandbox view), and #2266's reporter was on Windows. One `SANDBOXED` case for `write` — even asserting the sandbox-internal form — would document the mode difference and would fail loudly if someone later "fixes" the sandbox case by returning a host path.

oss-maintainer (commented):
### Summary
The delta since my previous approval (`d4bed6338c`) is the `delete()` half of the fix plus its regression test: `LocalFilesystem.java:548` now reports `absoluteResultPath(resolved)` instead of the caller-supplied key, matching `write()`/`edit()`/`move()` and the `WriteResult` javadoc. The new `delete_returnsAbsolutePathIncludingNamespace` asserts the namespaced form, so the #2266 contract is now covered for all four mutators. That part is correct.

The reason I am commenting rather than re-approving: once *every* mutator reports the resolved on-disk path, the remaining gap is that the value is not one consistent thing across filesystem stacks, and it is echoed verbatim into model-visible tool output. Both are cheaper to settle now than after this is the documented behaviour.

### Findings
- **[Warning]** `LocalFilesystem.java:587-590` — "absolute on-disk path" means host path in `ROOTED`/`UNRESTRICTED`, a sandbox-internal path in `SANDBOXED`, and is masked back to the caller's logical path by `CompositeFilesystem` (`:234`, `:469`, `:486`) and `OverlayFilesystem` (`:270`) whenever a route prefix applies — so the record javadoc overstates the guarantee for routed/overlay workspaces
- **[Warning]** `LocalFilesystem.java:338` — `FilesystemTool` renders `r.path()` straight into tool output (`FilesystemTool.java:148`, `:175`), so the host workspace root and the per-user namespace segment now land in transcripts/session state; `WorkspacePathNormalizer` exists to keep the model in workspace-relative coordinates, and this moves the output side the other way
- **[Info]** `LocalFilesystem.java:530` — the idempotent branch is now the one case where a successful result points at a path that does not exist
- **[Info]** `LocalFilesystemWritePathTest.java:42-60` — all four tests are `LocalFsMode.ROOTED`; `SANDBOXED` is where the reported path genuinely differs, and #2266 was reported on Windows

### Suggestions
Pick the contract explicitly. If "as seen by the backing filesystem" is what you want, say so on `WriteResult.path`/`EditResult.path` and add the `SANDBOXED` case so the mode difference is pinned rather than implied. If callers need the host path, the cleaner split is to keep `path()` logical (what the caller passed, which is what `FilesystemTool` should echo) and add `absolutePath()` for programmatic consumers — that fixes #2266's "the framework told me a path that does not exist" complaint without leaking the namespace directory into every write/edit the model performs.

---

jialiuyang:
Thanks for the follow-up.
I prefer option  b : keep path() as the caller/logical path, and add absolutePath() for the real on-disk path after namespace resolution.
`FilesystemTool` should keep echoing path() not absolutePath() , so host/namespace paths don’t leak into model transcripts. That still gives a truthful path via absolutePath().

PR #2319, opened 2026-07-20
fix: preserve skill isolation and tool result history

RainYuY:
## AgentScope-Java Version

2.0.1-SNAPSHOT

### Background

This PR fixes several correctness, compatibility, and session-isolation issues in the Harness skill runtime and tool-result eviction flow.

Previously:

- Fresh sessions could lose tool groups that were active when the agent was built.
- Explicitly persisted empty tool-group state could be mistaken for missing state.
- Static skill mode bypassed skill and visibility filters and lost access to lazily loaded skill resources.
- Concurrent sessions shared a mutable `SkillCatalog`, allowing one session to overwrite or observe another session's filtered skill view.
- `WorkspaceSkillRepository` could resolve skills using another concurrent session's active runtime context.
- Persisted sessions with an empty active-group list could not see `load_skill_through_path` after upgrading.
- Tool-result eviction modified canonical agent state but could leave the copied reasoning input unchanged.
- Reused tool-call IDs could overwrite previously evicted results.
- Cached eviction replacements could overwrite metadata or state rewritten by pre-reasoning hooks.
- Real tool output beginning with the eviction placeholder text could bypass eviction.

#### Session and toolkit state

- Capture the toolkit's initially active groups after agent construction.
- Apply those groups only when creating a genuinely new session.
- Preserve explicitly persisted empty active-group state.
- Distinguish a missing legacy `toolkit_activeGroups` key from a present but empty value.
- Preserve the complete `RuntimeContext` when merging toolkit-level execution context.

#### Skill runtime

- Use `HarnessSkillMiddleware` for both live and frozen repository modes.
- Freeze repository enumeration when dynamic skills are disabled while retaining per-request filtering.
- Apply `SkillFilter` and `SkillVisibilityFilter` before staging, prompt rendering, and catalog installation.
- Preserve the source repository association needed for lazy resource loading.
- Add `RuntimeContextSkillRepository` for repositories whose contents depend on the current request.
- Make `WorkspaceSkillRepository` enumerate skills using the supplied request context.
- Store `SkillCatalog` in the request-scoped `RuntimeContext` instead of a shared mutable reference.
- Resolve `load_skill_through_path` authorization from the tool call's runtime context.
- Register the skill loader as an ungrouped tool during agent construction so persisted sessions can continue to access it.
- Retain the previous public constructors and methods as deprecated compatibility overloads.

#### Tool-result eviction

- Evict oversized results from both canonical `AgentState` and the hook-modified `ReasoningInput`.
- Preserve hook deletions, rewrites, metadata, result state, message metadata, timestamps, and usage.
- Use content-addressed paths so reused tool-call IDs do not overwrite historical results.
- Include result text in the in-call eviction fingerprint.
- Cache only the generated placeholder instead of an entire `ToolResultBlock`.
- Mark evicted results with dedicated metadata instead of recognizing them by text prefix.
- Change the default eviction directory to the workspace-relative `large_tool_results` path.
- Update the English and Chinese memory documentation accordingly.

### Compatibility

The following previously published APIs remain available as deprecated compatibility paths:

- `SkillRuntime.currentCatalog()`
- `SkillRuntime.install(SkillCatalog, Toolkit)`
- `SkillLoadTool(AtomicReference<SkillCatalog>)`

Context-bearing tool calls never fall back to the legacy shared catalog, preserving session isolation.

### How to test

The focused core and Harness regression suite can be run with:

```bash
mvn -pl agentscope-harness -am \
  -Dtest=ReActAgentPerSessionStateTest,ToolExecutionContextIntegrationTest,ToolResultEvictionMiddlewareTest,HarnessSkillMiddlewareContextIsolationTest,WorkspaceSkillRepositoryTest,SkillRuntimeTest,SkillLoadToolFrontmatterViewTest,HarnessAgentDynamicHookBuilderTest \
  test

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -113,6 +113,7 @@ import io.agentscope.core.skill.repository.AgentSkillRepository;
 import io.agentscope.core.state.AgentState;
 import io.agentscope.core.state.AgentStateStore;
 import io.agentscope.core.state.LegacyStateLoader;
+import io.agentscope.core.state.ToolContextState;
 import io.agentscope.core.tool.AgentTool;
 import io.agentscope.core.tool.ToolBase;
 import io.agentscope.core.tool.ToolCallParam;
@@ -236,6 +237,9 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
      */
     private final Toolkit toolkit;
 
+    /** Default active groups captured after builder-time toolkit configuration is complete. */
+    private final List<String> initialActiveToolGroups;
+
     private final ToolExecutionContext toolExecutionContext;
 
     private final List<MiddlewareBase> middlewares;
@@ -294,6 +298,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
         super(builder.name, builder.description, new ArrayList<>(builder.hooks));
 
         this.toolkit = agentToolkit != null ? agentToolkit : new Toolkit();
+        this.initialActiveToolGroups = List.copyOf(this.toolkit.getActiveGroups());
         this.sysPrompt = builder.sysPrompt;
         this.model = builder.model;
         this.maxIters = builder.maxIters;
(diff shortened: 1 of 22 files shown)


RainYuY:
@chickenlj @jujn  PTAL

PR #2322, opened 2026-07-20
fix(core): unbind state-saver on agent close to prevent OOM leak

### LINKED ISSUE #2321: [Bug]: GracefulShutdownManager.stateSavers never unregisters agents, causing OOM when an agent is created per call
**Describe the bug**

`GracefulShutdownManager` holds a process-wide singleton map `stateSavers` (`ConcurrentHashMap<String, ShutdownStateSaver>`) that is effectively **write-only**: agents register a saver on construction via `bindStateSaver()`, but there is no unregister path anywhere in production code. Combined with `agentId = UUID.randomUUID()` (a fresh key per instance) and a saver lambda that captures the enclosing agent `this`, every short-lived agent instance stays strongly reachable from the `GracefulShutdownManager.INSTANCE` GC root forever — even after `agent.close()` is correctly called. In "create a fresh agent per call" deployments (e.g. wiring a new agent per request to pick up the latest dynamic config), this produces monotonic heap growth ending in `OutOfMemoryError`.

The asymmetry is visible directly in the source (2.0.0):

- `bindStateSaver()` does `stateSavers.put(agent.getAgentId(), saver)` — `GracefulShutdownManager.java:115`
- The only other mutation of `stateSavers` is `stateSavers.clear()` inside `resetForTesting()` — `GracefulShutdownManager.java:327` (test-only)
- There is **no** `unbindStateSaver` / `stateSavers.remove(...)` anywhere in the codebase
- `ReActAgent.close()` is a no-op for the core agent — `ReActAgent.java:3822`
- `HarnessAgent.close()` shuts down the task repo / workspace / delegate but never touches `stateSavers` — `HarnessAgent.java:376`

Contrast this with the sibling map `activeRequestsById`, which has a complete symmetric lifecycle: `registerRequest()` (`:153`) paired with `unregisterRequest()` (`:178`), the latter invoked from `.doFinally(...)` in `AgentBase` (`:291`). `stateSavers` is missing that unregister half.

**To Reproduce**

Minimal reproducer:

```java
AgentStateStore store = new RedisAgentStateStore(...);
for (int i = 0; i < 10_000; i++) {
            .stateStore(store)   // triggers bindStateSaver
            .model(model)
            .build();
    // ... use the agent ...
    agent.close();        // does NOT remove the stateSavers entry
}
// GracefulShutdownManager.getInstance() still holds 10_000 entries;
// 10_000 agent instances (each with its HttpClient + conversation history) are reachable -> OOM
```

How to execute: run the loop above under `-XX:+HeapDumpOnOutOfMemoryError`, or simply log the `stateSavers` size per iteration — it increments by 1 every loop with no bound.

**Expected behavior**

The `stateSavers` entry's lifetime should be bound to the agent instance. Calling `agent.close()` should release the entry, mirroring the `activeRequestsById` register/unregister symmetry, so short-lived agents can be garbage-collected.

**Error messages**

Observed in production (Spring Boot service creating one `HarnessAgent` per call backed by `RedisAgentStateStore`):

- `java.lang.OutOfMemoryError: Java heap space` under load
- Heap dump (`-XX:+HeapDumpOnOutOfMemoryError`): the `io.agentscope.core.shutdown.GracefulShutdownManager` singleton retained ~2.74 GB (~86% of heap), almost entirely attributable to its `stateSavers` field
- Thread dump: 373 `HttpClient-N-SelectorManager` threads (one `java.net.http.HttpClient` per leaked agent)
- Recurs within days after restart

Strong-reference chain:

```
static GracefulShutdownManager.INSTANCE            <- GC root
   └─ stateSavers (ConcurrentHashMap)              <- put only, never removed
        └─ entry<agentId, ShutdownStateSaver>
             └─ saver lambda
                  └─ captures enclosing this = ReActAgent instance
                       ├─ model     -> HttpClient (SelectorManager thread)
                       ├─ toolkit
                       ├─ stateCache -> per-session AgentState (conversation history)
                       └─ middlewares
```

- The saver is registered in the `ReActAgent` constructor when `stateStore != null`: `shutdownManager.bindStateSaver(this, agentState -> stateStore.save(...))` — `ReActAgent.java:321-328`
- The lambda reads the instance field `stateStore` (`ReActAgent.java:246`), so per JLS 15.27.2 it captures `this`, pinning the entire agent object graph
- `agentId` is `UUID.randomUUID().toString()` — `AgentBase.java:150` — so each new agent gets a new map key and the `put` adds rather than overwrites

- Java Version: 21 (GraalVM)
- OS: Linux (production) / Windows 11 (repro)
- stateStore: `RedisAgentStateStore` (agentscope-extensions-redis)

**Additional context**

Suggested fix:

1. Add an unregister API and call it from `close()` (root fix):

```java
// GracefulShutdownManager
public void unbindStateSaver(Agent agent) {
    if (agent == null) return;
    stateSavers.remove(agent.getAgentId());
}
```

```java
// ReActAgent.close() (or AgentBase.close()) — unregister in finally
@Override
public void close() {
    try {
        // ...existing logic...
    } finally {
        shutdownManager.unbindStateSaver(this);
    }
}
```

2. Avoid capturing the enclosing `this` in the saver (defense in depth — shrinks each entry):

```java
// In the ReActAgent constructor: capture a final local instead of the instance field
final AgentStateStore store = this.stateStore;
shutdownManager.bindStateSaver(this, agentState ->
        store.save(agentState.getUserId(), agentState.getSessionId(), "agent_state", agentState));
```

Suggestion 2 alone is insufficient (the entry still accumulates in the map); it must be combined with suggestion 1. Together they give a symmetric register/unregister lifecycle plus a saver that holds no strong reference to the agent.

Design context: the `activeRequestsById` comment (`GracefulShutdownManager.java:51-56`) states "a single agent instance may serve many concurrent calls" — the framework's intended model is long-lived, reused agent instances. Under that model `stateSavers` does not leak (stable agentId, `put` overwrites, bounded entries). However, creating a fresh agent per call is a legitimate application-level pattern (typical when wiring agents against Spring `@RefreshScope` / Nacos dynamic config to use the latest configuration per invocation). The current implementation turns that pattern into a deterministic OOM; making the framework robust to non-reused agents would be appreciated.

Buktal:
## AgentScope-Java Version

2.0.0

## Description

Fixes #2321.

### Background

`GracefulShutdownManager.stateSavers` is a process-wide singleton `ConcurrentHashMap` that was effectively **write-only**: `bindStateSaver()` puts an entry keyed by `agentId`, but there was no removal path anywhere in production code (only `stateSavers.clear()` inside `resetForTesting()`). Since `agentId = UUID.randomUUID()` is a fresh key per instance, and the registered saver lambda captured the enclosing agent `this` (via the instance field `stateStore`), every agent built with a `stateStore` permanently pinned its entire object graph (model/HttpClient, toolkit, `stateCache` conversation history, middlewares) from the `GracefulShutdownManager.INSTANCE` GC root - even after `agent.close()` was correctly called.

In "create a fresh agent per call" deployments (e.g. wiring a new agent per request against Spring `@RefreshScope` / Nacos dynamic config), this produced monotonic heap growth ending in `OutOfMemoryError` (observed in production: ~2.74 GB retained by the singleton, 373 leaked `HttpClient` SelectorManager threads, recurring OOM within days).

The sibling map `activeRequestsById` already has a symmetric `registerRequest` / `unregisterRequest` lifecycle (the latter invoked from `doFinally`); `stateSavers` was missing that unregister half.

### Changes

1. **`GracefulShutdownManager.unbindStateSaver(Agent)`** - new public method, symmetric with `bindStateSaver`, mirroring the `registerRequest`/`unregisterRequest` pair. No-op on `null` / unregistered agent.

2. **`ReActAgent.close()`** - now calls `shutdownManager.unbindStateSaver(this)` so the entry's lifetime is bound to the agent instance. `HarnessAgent.close()` already delegates to `ReActAgent.close()` via `delegate.close()`, so wrapped agents are covered automatically. No `try/finally` is needed since `unbindStateSaver` cannot throw.

3. **`ReActAgent` constructor** - the saver lambda now captures a `final AgentStateStore stateSaverStore` local instead of the enclosing `this`. Defense in depth: even if an entry is somehow left behind (e.g. a caller forgets `close()`), it no longer pins the entire agent object graph - only a single store reference. (Item 1 is the root fix; item 3 alone is insufficient since the entry would still accumulate.)

### Tests

Added 3 unit tests in `GracefulShutdownTest`:

- `unbindStateSaver removes a previously bound saver` - end-to-end: `bind` -> `registerRequest` -> `saveOnInterruptObserved` invokes the saver; `unbind` -> `registerRequest` -> `saveOnInterruptObserved` is a no-op (verifies the saver is actually removed from the map, since `ActiveRequestContext.saveState` short-circuits on a `null` saver).
- `unbindStateSaver with null agent is no-op`
- `unbindStateSaver for an unregistered agent is no-op`

### How to test

```bash
mvn -pl agentscope-core test
```

Result: `Tests run: 2218, Failures: 0, Errors: 0, Skipped: 8` - all green, including the 3 new tests.

> Note: `agentscope-harness` shows 7 `JUnit Failed to close extension context` errors on this Windows machine, but these are a **pre-existing** JUnit `@TempDir` cleanup issue on Windows (`java.nio.file.DirectoryNotEmptyException` in `WindowsFileSystemProvider.implDelete`) - verified by stashing this PR's changes and reproducing the same failures on clean `main`. They are unrelated to this change and do not occur on Linux CI.

## Checklist

- [x] All tests are passing (`mvn test`) - core suite green (2218 tests); harness failures are pre-existing Windows `@TempDir` issues (see note above)
- [x] Related documentation has been updated (N/A - internal lifecycle fix; the new method carries full Javadoc)

### DIFF
--- a/agentscope-core/src/test/java/io/agentscope/core/shutdown/GracefulShutdownTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/shutdown/GracefulShutdownTest.java
@@ -357,6 +357,40 @@ class GracefulShutdownTest {
         }
 
         @Test
+        @DisplayName("unbindStateSaver removes a previously bound saver")
+        void unbindStateSaverRemovesBoundSaver() {
+            TestableAgent agent = createTestAgent("agent-1");
+            AtomicReference<AgentState> savedState = new AtomicReference<>();
+
+            manager.bindStateSaver(agent, savedState::set);
+
+            // Before unbind: the saver is reachable via registerRequest -> saveOnInterruptObserved
+            String requestIdBefore = manager.registerRequest(agent);
+            manager.saveOnInterruptObserved(requestIdBefore);
+            assertNotNull(savedState.get(), "saver should be invoked before unbind");
+
+            // After unbind: registerRequest finds no saver, so saveOnInterruptObserved is a no-op
+            savedState.set(null);
+            manager.unbindStateSaver(agent);
+            String requestIdAfter = manager.registerRequest(agent);
+            manager.saveOnInterruptObserved(requestIdAfter);
+            assertNull(savedState.get(), "saver should not be invoked after unbind");
+        }
+
+        @Test
+        @DisplayName("unbindStateSaver with null agent is no-op")
+        void unbindStateSaverNullAgent() {
+            assertDoesNotThrow(() -> manager.unbindStateSaver(null));
+        }
+
+        @Test
+        @DisplayName("unbindStateSaver for an unregistered agent is no-op")
+        void unbindStateSaverUnregisteredAgent() {
+            TestableAgent agent = createTestAgent("agent-1");
+            assertDoesNotThrow(() -> manager.unbindStateSaver(agent));
+        }
+
+        @Test
         @DisplayName("checkAndClearShutdownInterrupted with null agent returns false")
         void checkInterruptedNullAgent() {
             assertFalse(manager.checkAndClearShutdownInterrupted(null));


Buktal:
@AgentScopeJavaBot

chickenlj:
Fixed in https://github.com/agentscope-ai/agentscope-java/pull/2384.

Issue #2240, opened 2026-07-15
[Feature]: Gemini Thinking 增强


jujn:
当前仅支持 thinkingBudget -> ThinkingConfig(includeThoughts=true)，缺 thinkingLevel、单独 includeThoughts。历史 ThinkingBlock 请求侧直接跳过，thought signature 生命周期只覆盖部分 tool call metadata。

wzl521:
@jujn  hi I'd like to work on this. Could you assign it to me?

PR #2335, opened 2026-07-21
feat(gemini): 支持 Gemini Thinking 并持久化 thought signature

wzl521:
Background

  https://github.com/agentscope-ai/agentscope-java/issues/2240 — The Gemini model adapter lacked first-class support for Gemini's built-in thinking capability. There were several gaps:

  1. No includeThoughts / thinkingLevel options — GenerateOptions only had thinkingBudget, but Gemini's API requires includeThoughts (to independently control whether thoughts appear in responses) and thinkingLevel (e.g., "minimal",
  "low", "medium", "high") for the latest models.
  2. thoughtSignature was not persisted — Gemini returns an opaque thoughtSignature on each part. This signature must be round-tripped back to Gemini on subsequent turns, but the codebase only captured it on ToolUseBlock (via
  ToolUseBlock.METADATA_THOUGHT_SIGNATURE), not on TextBlock or ThinkingBlock. After a JSON serialization round-trip, the byte-array signature was lost or corrupted.
  3. Part boundaries were lost during accumulation — ThinkingAccumulator and TextAccumulator merged all chunks into a single block. When a Gemini response contained multiple parts each with its own thoughtSignature, the boundaries
  were destroyed, making it impossible to reconstruct distinct signed parts in the next request.

  Changes

  Core (agentscope-core)

  - GenerateOptions: Added includeThoughts(Boolean) and thinkingLevel(String) options alongside the existing thinkingBudget. These are now properly merged in merge(GenerateOptions, GenerateOptions).
  - ContentBlockMetadataKeys: New constant class with THOUGHT_SIGNATURE key, replacing the ad-hoc constant previously defined on ToolUseBlock. This is the single source of truth for the thought signature metadata key across all
  content block types.
  - TextBlock: Added metadata field (provider-specific Map<String, Object>) with getter and builder support. This enables text blocks to carry thoughtSignature.
  - ThinkingBlock: Already had metadata support; wired into the new signature persistence pipeline.
  - ToolUseBlock: Replaced the local METADATA_THOUGHT_SIGNATURE constant with ContentBlockMetadataKeys.THOUGHT_SIGNATURE reference. No behavioral change.
  - TextAccumulator / ThinkingAccumulator: Introduced buildAllTextBlocks() / buildAllThinkingBlocks() methods that preserve thought-signature Part boundaries. When a chunk carries a thoughtSignature, the current block is completed
  and a new one begins. The existing buildAggregated() method continues to return a single merged block for backward compatibility.
  - ReasoningContext: Switched from buildAggregated() to buildAllXXXBlocks() for both text and thinking, ensuring the final assistant message preserves distinct signed parts.

  Gemini Extension (agentscope-extensions-model-gemini)

  - GeminiThoughtSignatureUtils (new): Utility class that:
    - Extracts thoughtSignature from a Gemini Part into content block metadata (as byte[] in memory).
    - Restores a persisted signature back onto a Gemini Part.Builder — supports both byte[] (in-memory) and Base64-encoded String (after JSON serialization).
  - GeminiChatFormatter: Updated applyOptions to build ThinkingConfig from all three fields (thinkingBudget, thinkingLevel, includeThoughts). When only thinkingBudget is set (legacy callers), includeThoughts defaults to true
  preserving existing behavior. When includeThoughts is explicitly set, it always takes precedence.
  - GeminiResponseParser: Refactored parsePartsToBlocks to:
    - Preserve original Part order (no longer reorders thinking → text → toolCall).
    - Attach thoughtSignature metadata to ALL block types (ThinkingBlock, TextBlock, ToolUseBlock) via GeminiThoughtSignatureUtils.extractMetadata.
    - Emit empty blocks when only metadata (signature) is present.
  - GeminiMessageConverter: Updated to restore thoughtSignature from content block metadata onto Gemini Part objects for all assistant-role blocks (ThinkingBlock → Part.thought(true), TextBlock → text Part, ToolUseBlock →
  functionCall Part) via GeminiThoughtSignatureUtils.applyMetadata.

  Documentation

  - docs/v2/en/integration/model/gemini.md: Added "Thinking" section documenting thinkingLevel, includeThoughts, and the signature round-trip behavior.
  - docs/v2/zh/integration/model/gemini.md: Chinese translation.
  - docs/v2/en/integration/model/gemini-request-flow.md (new): Full end-to-end Mermaid diagram of a Gemini request, covering message conversion, thinking config, signature extraction/persistence, and the three completion paths (plain
   answer / tool call / persistent history replay).
  - docs/v2/zh/integration/model/gemini-request-flow.md (new): Chinese translation.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/TextAccumulator.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/TextAccumulator.java
@@ -17,6 +17,8 @@ package io.agentscope.core.agent.accumulator;
 
 import io.agentscope.core.message.ContentBlock;
 import io.agentscope.core.message.TextBlock;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Text content accumulator for accumulating streaming text chunks.
@@ -27,6 +29,7 @@ import io.agentscope.core.message.TextBlock;
 public class TextAccumulator implements ContentAccumulator<TextBlock> {
 
     private final StringBuilder accumulated = new StringBuilder();
+    private final Map<String, Object> metadata = new HashMap<>();
 
     /**
      * @hidden
@@ -36,6 +39,9 @@ public class TextAccumulator implements ContentAccumulator<TextBlock> {
         if (block != null && block.getText() != null) {
             accumulated.append(block.getText());
         }
+        if (block != null && block.getMetadata() != null) {
+            metadata.putAll(block.getMetadata());
+        }
     }
 
     /**
@@ -43,7 +49,7 @@ public class TextAccumulator implements ContentAccumulator<TextBlock> {
      */
     @Override
     public boolean hasContent() {
-        return accumulated.length() > 0;
+        return accumulated.length() > 0 || !metadata.isEmpty();
     }
 
     /**
@@ -54,7 +60,10 @@ public class TextAccumulator implements ContentAccumulator<TextBlock> {
         if (!hasContent()) {
             return null;
         }
-        return TextBlock.builder().text(accumulated.toString()).build();
+        return TextBlock.builder()
+                .text(accumulated.toString())
+                .metadata(metadata.isEmpty() ? null : metadata)
+                .build();
     }
 
     /**
@@ -63,6 +72,7 @@ public class TextAccumulator implements ContentAccumulator<TextBlock> {
     @Override
     public void reset() {
         accumulated.setLength(0);
+        metadata.clear();
     }
 
     /**
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/TextAccumulator.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/TextAccumulator.java
@@ -16,8 +16,11 @@
 package io.agentscope.core.agent.accumulator;
 
 import io.agentscope.core.message.ContentBlock;
+import io.agentscope.core.message.ContentBlockMetadataKeys;
 import io.agentscope.core.message.TextBlock;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
@@ -30,17 +33,33 @@ public class TextAccumulator implements ContentAccumulator<TextBlock> {
 
     private final StringBuilder accumulated = new StringBuilder();
     private final Map<String, Object> metadata = new HashMap<>();
+    private final List<TextBlock> completedBlocks = new ArrayList<>();
+    private final StringBuilder currentBlock = new StringBuilder();
+    private final Map<String, Object> currentMetadata = new HashMap<>();
 
     /**
      * @hidden
      */
     @Override
     public void add(TextBlock block) {
-        if (block != null && block.getText() != null) {
-            accumulated.append(block.getText());
+        if (block == null) {
+            return;
         }
-        if (block != null && block.getMetadata() != null) {
-            metadata.putAll(block.getMetadata());
+
+        accumulated.append(block.getText());
+        currentBlock.append(block.getText());
+
+        Map<String, Object> blockMetadata = block.getMetadata();
+        if (blockMetadata != null) {
+            metadata.putAll(blockMetadata);
+            currentMetadata.putAll(blockMetadata);
+        }
+
+        // A thought signature terminates its provider Part. Keep that boundary so distinct
+        // signature-bearing Parts can be replayed without merging their metadata.
+        if (blockMetadata != null
+                && blockMetadata.get(ContentBlockMetadataKeys.THOUGHT_SIGNATURE) != null) {
+            completeCurrentBlock();
         }
     }
 
@@ -67,12 +86,33 @@ public class TextAccumulator implements ContentAccumulator<TextBlock> {
     }
 
     /**
+     * Build accumulated text blocks while preserving thought-signature Part boundaries.
+     *
+     * @hidden
+     * @return accumulated text blocks in their original order
+     */
+    public List<TextBlock> buildAllTextBlocks() {
+        if (!hasContent()) {
+            return List.of();
+        }
+
+        List<TextBlock> blocks = new ArrayList<>(completedBlocks);
+        if (currentBlock.length() > 0 || !currentMetadata.isEmpty()) {
+            blocks.add(buildCurrentBlock());
+        }
+        return List.copyOf(blocks);
+    }
+
+    /**
      * @hidden
      */
     @Override
     public void reset() {
         accumulated.setLength(0);
         metadata.clear();
+        completedBlocks.clear();
+        currentBlock.setLength(0);
+        currentMetadata.clear();
     }
 
     /**
@@ -88,12 +128,29 @@ public class TextAccumulator implements ContentAccumulator<TextBlock> {
     /**
      * Replace the accumulated text with content reconstructed from the model-call event stream.
      *
+     * <p>Resets any in-flight thought-signature Part boundary and makes the new content the
+     * current block, so {@link #buildAllTextBlocks()} reflects the replaced text.
+     *
      * @hidden
      */
     public void replace(String text) {
-        accumulated.setLength(0);
+        reset();
         if (text != null) {
             accumulated.append(text);
+            currentBlock.append(text);
         }
+    }
+
+    private void completeCurrentBlock() {
+        completedBlocks.add(buildCurrentBlock());
+        currentBlock.setLength(0);
+        currentMetadata.clear();
+    }
+
+    private TextBlock buildCurrentBlock() {
+        return TextBlock.builder()
+                .text(currentBlock.toString())
+                .metadata(currentMetadata.isEmpty() ? null : currentMetadata)
+                .build();
     }
 }
--- a/agentscope-core/src/main/java/io/agentscope/core/message/TextBlock.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/message/TextBlock.java
@@ -16,7 +16,10 @@
 package io.agentscope.core.message;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Represents plain text content in a message.
@@ -25,21 +28,27 @@ import com.fasterxml.jackson.annotation.JsonProperty;
  * Text blocks are commonly used for user messages, assistant responses,
  * and any other textual communication.
  *
- * <p>The text content can be empty but never null. The toString() method
- * returns the text content for convenience.
+ * <p>The text content can be empty but never null. Optional metadata preserves provider-specific
+ * information associated with the text content. The toString() method returns the text content for
+ * convenience.
  */
 public final class TextBlock extends ContentBlock {
 
     private final String text;
+    private final Map<String, Object> metadata;
 
     /**
      * Creates a new text block for JSON deserialization.
      *
      * @param text The text content (null will be converted to empty string)
+     * @param metadata Optional provider-specific metadata
      */
     @JsonCreator
-    private TextBlock(@JsonProperty("text") String text) {
+    private TextBlock(
+            @JsonProperty("text") String text,
+            @JsonProperty("metadata") Map<String, Object> metadata) {
         this.text = text != null ? text : "";
+        this.metadata = metadata != null ? new HashMap<>(metadata) : null;
     }
 
     /**
@@ -49,6 +58,16 @@ public final class TextBlock extends ContentBlock {
      */
     public String getText() {
         return text;
+    }
+
+    /**
+     * Gets provider-specific metadata associated with this text block.
+     *
+     * @return The metadata map, or null if no metadata is set
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public Map<String, Object> getMetadata() {
+        return metadata;
     }
 
     @Override
@@ -71,6 +90,7 @@ public final class TextBlock extends ContentBlock {
     public static class Builder {
 
         private String text;
+        private Map<String, Object> metadata;
 
         /**
          * Sets the text content for the block.
@@ -84,12 +104,23 @@ public final class TextBlock extends ContentBlock {
         }
 
         /**
+         * Sets provider-specific metadata for the block.
+         *
+         * @param metadata The metadata map
+         * @return This builder for chaining
+         */
+        public Builder metadata(Map<String, Object> metadata) {
+            this.metadata = metadata;
+            return this;
+        }
+
+        /**
          * Builds a new TextBlock with the configured text.
          *
          * @return A new TextBlock instance (null text will be converted to empty string)
          */
         public TextBlock build() {
-            return new TextBlock(text != null ? text : "");
+            return new TextBlock(text != null ? text : "", metadata);
         }
     }
 }
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiChatFormatter.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiChatFormatter.java
@@ -117,14 +117,30 @@ public class GeminiChatFormatter
                 defaultOptions,
                 configBuilder::presencePenalty);
 
-        // Apply ThinkingConfig if either includeThoughts or thinkingBudget is set
+        // Apply ThinkingConfig when any thinking option is configured.
         Integer thinkingBudget =
                 getOptionOrDefault(options, defaultOptions, GenerateOptions::getThinkingBudget);
+        Boolean includeThoughts =
+                getOptionOrDefault(options, defaultOptions, GenerateOptions::getIncludeThoughts);
+        String thinkingLevel =
+                getOptionOrDefault(options, defaultOptions, GenerateOptions::getThinkingLevel);
 
-        if (thinkingBudget != null) {
+        if (thinkingBudget != null || includeThoughts != null || thinkingLevel != null) {
             ThinkingConfig.Builder thinkingConfigBuilder = ThinkingConfig.builder();
-            thinkingConfigBuilder.includeThoughts(true);
-            thinkingConfigBuilder.thinkingBudget(thinkingBudget);
+
+            if (includeThoughts != null) {
+                thinkingConfigBuilder.includeThoughts(includeThoughts);
+            } else if (thinkingBudget != null) {
+                // Preserve the existing behavior for callers that only set thinkingBudget.
+                thinkingConfigBuilder.includeThoughts(true);
+            }
+            if (thinkingBudget != null) {
+                thinkingConfigBuilder.thinkingBudget(thinkingBudget);
+            }
+            if (thinkingLevel != null) {
+                thinkingConfigBuilder.thinkingLevel(thinkingLevel);
+            }
+
             configBuilder.thinkingConfig(thinkingConfigBuilder.build());
         }
     }
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiMessageConverter.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiMessageConverter.java
@@ -52,6 +52,7 @@ import org.slf4j.LoggerFactory;
  * <p>This converter handles the core message transformation logic, including:
  * <ul>
  *   <li>Text blocks</li>
+ *   <li>Thinking blocks</li>
  *   <li>Tool use blocks (function_call)</li>
  *   <li>Tool result blocks (function_response as independent Content)</li>
  *   <li>Multimodal content (image, audio, video)</li>
@@ -88,10 +89,15 @@ public class GeminiMessageConverter {
 
         for (Msg msg : msgs) {
             List<Part> parts = new ArrayList<>();
+            boolean modelRole = msg.getRole() == MsgRole.ASSISTANT;
 
             for (ContentBlock block : msg.getContent()) {
                 if (block instanceof TextBlock tb) {
-                    parts.add(Part.builder().text(tb.getText()).build());
+                    Part.Builder partBuilder = Part.builder().text(tb.getText());
+                    if (modelRole) {
+                        GeminiThoughtSignatureUtils.applyMetadata(partBuilder, tb.getMetadata());
+                    }
+                    parts.add(partBuilder.build());
 
                 } else if (block instanceof ToolUseBlock tub) {
                     // Prioritize using content field (raw arguments string), fallback to input map
@@ -124,14 +130,8 @@ public class GeminiMessageConverter {
                     // Build Part with FunctionCall and optional thought signature
                     Part.Builder partBuilder = Part.builder().functionCall(functionCall);
 
-                    // Check for thought signature in metadata
-                    Map<String, Object> metadata = tub.getMetadata();
-                    if (metadata != null
-                            && metadata.containsKey(ToolUseBlock.METADATA_THOUGHT_SIGNATURE)) {
-                        Object signature = metadata.get(ToolUseBlock.METADATA_THOUGHT_SIGNATURE);
-                        if (signature instanceof byte[]) {
-                            partBuilder.thoughtSignature((byte[]) signature);
-                        }
+                    if (modelRole) {
+                        GeminiThoughtSignatureUtils.applyMetadata(partBuilder, tub.getMetadata());
                     }
 
                     parts.add(partBuilder.build());
@@ -179,9 +179,19 @@ public class GeminiMessageConverter {
                 } else if (block instanceof HintBlock hb) {
                     parts.add(Part.builder().text(hb.getHint()).build());
 
-                } else if (block instanceof ThinkingBlock) {
-                    log.debug("Skipping ThinkingBlock when formatting message for Gemini API");
-                    continue;
+                } else if (block instanceof ThinkingBlock thinkingBlock) {
+                    if (!modelRole) {
+                        log.debug(
+                                "Skipping ThinkingBlock from non-assistant message when formatting"
+                                        + " for Gemini API");
+                        continue;
+                    }
+
+                    Part.Builder partBuilder =
+                            Part.builder().text(thinkingBlock.getThinking()).thought(true);
+                    GeminiThoughtSignatureUtils.applyMetadata(
+                            partBuilder, thinkingBlock.getMetadata());
+                    parts.add(partBuilder.build());
 
                 } else {
                     log.warn(
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiResponseParser.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiResponseParser.java
@@ -138,7 +138,7 @@ public class GeminiResponseParser {
             // Check for thinking content first (parts with thought=true flag)
             if (part.thought().isPresent() && part.thought().get() && part.text().isPresent()) {
                 String thinkingText = part.text().get();
-                if (thinkingText != null && (!thinkingText.isEmpty() || metadata != null)) {
+                if (!thinkingText.isEmpty() || metadata != null) {
                     blocks.add(
                             ThinkingBlock.builder()
                                     .thinking(thinkingText)
@@ -158,7 +158,7 @@ public class GeminiResponseParser {
             // Check for text content
             if (part.text().isPresent()) {
                 String text = part.text().get();
-                if (text != null && (!text.isEmpty() || metadata != null)) {
+                if (!text.isEmpty() || metadata != null) {
                     blocks.add(TextBlock.builder().text(text).metadata(metadata).build());
                 }
             }
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiResponseParser.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiResponseParser.java
@@ -126,36 +126,41 @@ public class GeminiResponseParser {
     }
 
     /**
-     * Parse Gemini Part objects to AgentScope ContentBlocks.
-     * Order of block types: ThinkingBlock, TextBlock, ToolUseBlock
+     * Parse Gemini Part objects to AgentScope ContentBlocks while preserving Part order.
      *
      * @param parts List of Gemini Part objects
      * @param blocks List to add parsed ContentBlocks to
      */
     protected void parsePartsToBlocks(List<Part> parts, List<ContentBlock> blocks) {
         for (Part part : parts) {
+            Map<String, Object> metadata = GeminiThoughtSignatureUtils.extractMetadata(part);
+
             // Check for thinking content first (parts with thought=true flag)
             if (part.thought().isPresent() && part.thought().get() && part.text().isPresent()) {
                 String thinkingText = part.text().get();
-                if (thinkingText != null && !thinkingText.isEmpty()) {
-                    blocks.add(ThinkingBlock.builder().thinking(thinkingText).build());
+                if (thinkingText != null && (!thinkingText.isEmpty() || metadata != null)) {
+                    blocks.add(
+                            ThinkingBlock.builder()
+                                    .thinking(thinkingText)
+                                    .metadata(metadata)
+                                    .build());
                 }
+                continue;
+            }
+
+            // Check for function call (tool use)
+            if (part.functionCall().isPresent()) {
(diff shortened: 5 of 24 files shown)


wzl521:
https://github.com/agentscope-ai/agentscope-java/issues/2240，please review，thanks

oss-maintainer:
**CLA Not Signed**

@wzl521 please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.

---

wzl521:
I have signed the CLA，please retry it！

wzl521:
> **CLA Not Signed  尚未签署 CLA 文件**
> 
> The Contributor License Agreement (CLA) check is currently pending on this PR (`license/cla`: Contributor License Agreement is not signed yet.). This PR **cannot be merged** until the CLA is signed.关于《贡献者许可协议》的确认工作目前仍处于待处理状态（ `license/cla` ：尚未签署《贡献者许可协议》）。在签署该协议之前，无法合并此拉取请求。
> 
> @wzl521 please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.请通过上面评论中的 CLA 助手按钮签署 CLA 文件，或者访问 https://cla-assistant.io/agentscope-ai/agentscope-java。一旦签署完成， `license/cla` 的状态就会变为绿色。
> 
> _Automated check by github-manager-bot由 github-manager-bot 进行的自动化检查_

I have signed the CLA,please retry it!

AgentScopeJavaBot (commented):
🤖 AI Review

This PR adds Gemini Thinking support with persisted thought signatures across core and Gemini extension modules. Core changes introduce metadata on TextBlock, new ContentBlockMetadataKeys, multi-block accumulation in TextAccumulator/ThinkingAccumulator, and new options in GenerateOptions. Gemini extension adds GeminiThoughtSignatureUtils, updated converters, and comprehensive tests. Overall quality is high — no critical bugs found.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR adds Gemini Thinking support with persisted thought signatures across core and Gemini extension modules. Core changes introduce metadata on TextBlock, new ContentBlockMetadataKeys, multi-block accumulation in TextAccumulator/ThinkingAccumulator, and new options in GenerateOptions. Gemini extension adds GeminiThoughtSignatureUtils, updated converters, and comprehensive tests. Overall quality is high — no critical bugs found.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR adds Gemini Thinking support with persisted thought signatures across core and Gemini extension modules. Core changes introduce metadata on TextBlock, new ContentBlockMetadataKeys, multi-block accumulation in TextAccumulator/ThinkingAccumulator, and new options in GenerateOptions. Gemini extension adds GeminiThoughtSignatureUtils, updated converters, and comprehensive tests. Overall quality is high — no critical bugs found.

jujn:
image
两个账号都需要签署CLA。另外，CI有报错，可以看下

oss-maintainer:

```bash
git fetch upstream
git checkout feat/2240-gemini-thinking-enhancement
git rebase upstream/main
# resolve conflicts, then:
git push --force-with-lease
```

---

wzl521:
@oss-maintainer please re-review

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiMessageConverter.java
```diff
@@ -190,9 +179,19 @@ public List<Content> convertMessages(List<Msg> msgs) {
...
+                } else if (block instanceof ThinkingBlock thinkingBlock) {
+                    if (!modelRole) {
+                        log.debug(
+                                "Skipping ThinkingBlock from non-assistant message when formatting"
+                                        + " for Gemini API");
+                        continue;
+                    }
+
+                    Part.Builder partBuilder =
+                            Part.builder().text(thinkingBlock.getThinking()).thought(true);
```
oss-maintainer: **[Critical]** Assistant `ThinkingBlock`s used to be skipped on the request path; now every one of them is replayed as `Part.text(thinking).thought(true)`. Two things need confirming before merge:

1. Is `thought=true` accepted on an *input* part by the genai SDK / the Gemini families in scope, or is it response-only metadata the endpoint rejects? If it is rejected, every stored thinking turn becomes a hard request failure — a much larger blast radius than the bug being fixed.
2. A signature-only part carries `thinking == ""` (see `GeminiResponseParser:144`/`:164`, which deliberately emit empty blocks), so this line can produce `{text: "", thought: true, thoughtSignature: ...}`. Is an empty text part accepted?

Could you add request-shape assertions (serialize the converted `Content` and assert the exact JSON) for the non-empty-thinking, signature-only-empty-thinking, and non-assistant-thinking cases? A unit test that stops at the block level cannot show what actually goes on the wire, and this PR changes exactly that.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiThoughtSignatureUtils.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        if (value instanceof byte[] signature) {
+            partBuilder.thoughtSignature(signature.clone());
+            return;
+        }
+
+        if (!(value instanceof String encodedSignature)) {
+            throw new FormatterException(
+                    "Unsupported Gemini thought signature metadata type: "
```
oss-maintainer: **[Warning]** This changes the failure semantics for replaying persisted state. Before this PR an undecodable signature was tolerated — the old `GeminiMessageConverter` caught `IllegalArgumentException`, logged `"Skipping invalid thought signature on tool call ..."` and still sent the request. Here a bad value throws `FormatterException`, so one corrupt entry in agent state aborts the entire model call.

Agent state outlives the code that wrote it (`AgentStateStore` on MySQL/Redis/OSS, plus hand-edited or partially-migrated rows), so today's "skip and degrade" becomes a permanently broken session. Suggest keeping log-and-drop as the default and making strictness opt-in, or at least documenting the behavior change for store implementations. Note the empty-string case throws too, which is not obviously worse than the no-signature case that `applyMetadata` silently accepts above it.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiResponseParser.java
```diff
@@ -129,49 +129,54 @@ public ChatResponse parseResponse(GenerateContentResponse response, Instant star
...
      * Parse Gemini FunctionCall to ToolUseBlock.
      *
      * @param functionCall Gemini FunctionCall object
-     * @param thoughtSignature Thought signature from the Part (may be null)
+     * @param metadata Provider-specific metadata from the Part (may be null)
      * @param blocks List to add parsed ToolUseBlock to
      */
     protected void parseToolCall(
-            FunctionCall functionCall, byte[] thoughtSignature, List<ContentBlock> blocks) {
+            FunctionCall functionCall, Map<String, Object> metadata, List<ContentBlock> blocks) {
```
oss-maintainer, line 199: **[Warning]** `protected void parseToolCall(FunctionCall, byte[], List<ContentBlock>)` becomes `parseToolCall(FunctionCall, Map<String, Object>, List<ContentBlock>)`. That is a breaking change for out-of-tree subclasses: an existing `@Override` simply stops overriding. `GeminiResponseParser` is precisely the seam downstream providers extend.

Please keep a `@Deprecated` `byte[]` overload delegating to the new form (it can drop the signature into the map), or list the break explicitly in the release notes / migration guide.

agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/TextAccumulator.java
```diff
@@ -43,7 +68,7 @@ public void add(TextBlock block) {
      */
     @Override
     public boolean hasContent() {
-        return accumulated.length() > 0;
+        return accumulated.length() > 0 || !metadata.isEmpty();
```
oss-maintainer, line 83: **[Warning]** `hasContent()` now returns true when only metadata is present, so `buildAggregated()` can emit one block whose text is the concatenation of *all* parts while its metadata is the last-write-wins merge — i.e. a single thought signature attached to merged text. That pairing is not replay-safe.

The aggregated form is still in use: `StreamChatResponseAggregator` in `agentscope-extensions-studio` (lines 96-97) calls `textAcc.buildAggregated()` / `thinkingAcc.buildAggregated()`. Since `buildAllTextBlocks()` is the correct API after this PR, could you either drop signature metadata from `buildAggregated()` and document it as text-only, or migrate that caller? Otherwise there is a silent mis-attribution waiting for the next aggregated consumer.

agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/TextAccumulator.java
```diff
@@ -78,12 +128,55 @@ public String getAccumulated() {
...
         }
     }
+
+    /**
+     * Merges incoming metadata into the given accumulator metadata map.
+     *
+     * <p>List values are accumulated in stream order (entries appended, not replaced). Scalar
+     * values use last-write-wins. List values are always copied to prevent caller mutation.
+     */
+    private void mergeMetadata(Map<String, Object> target, Map<String, Object> incoming) {
```
oss-maintainer: **[Info]** `mergeMetadata`, `completeCurrentBlock` and `buildCurrentBlock` are now duplicated verbatim between `TextAccumulator` and `ThinkingAccumulator` (~35 lines), and the two `add()` methods already differ in null handling — `ThinkingAccumulator.add()` guards `block.getThinking() != null`, here `block.getText()` is appended unguarded. It happens to be safe because `TextBlock` normalizes null to `""`, but `StringBuilder.append((String) null)` writes the literal `"null"`, so the asymmetry is a trap for the next caller that constructs a block differently.

The signature-boundary rule is provider-agnostic and will need the same fix every time. Could you hoist it into one package-private helper (a buffer holding `accumulated` / `currentBlock` / both metadata maps) and have both accumulators delegate, with identical null handling?

agentscope-core/src/main/java/io/agentscope/core/message/TextBlock.java
```diff
@@ -51,6 +60,16 @@ public String getText() {
         return text;
     }
 
+    /**
+     * Gets provider-specific metadata associated with this text block.
+     *
+     * @return The metadata map, or null if no metadata is set
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public Map<String, Object> getMetadata() {
```
oss-maintainer, line 78: **[Info]** The new accessor follows `ThinkingBlock` (nullable map, defensive copy on write, mutable map on read), while `ToolUseBlock` normalizes to `Collections.emptyMap()` and returns `Collections.unmodifiableMap(...)`. Two different metadata contracts now live in sibling core message types, so consumers must remember which block class needs a null check and which one rejects mutation — and a caller doing `block.getMetadata().put(...)` gets no compile-time signal.

`TextBlock` is serialized on every persisted state and every wire transport, so this choice is effectively permanent. I would converge on unmodifiable + one null convention across the three blocks (a follow-up is fine); if the current shape is deliberate, say so in the Javadoc.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiChatFormatter.java
```diff
@@ -117,14 +117,30 @@ public void applyOptions(
...
+        String thinkingLevel =
+                getOptionOrDefault(options, defaultOptions, GenerateOptions::getThinkingLevel);
 
-        if (thinkingBudget != null) {
+        if (thinkingBudget != null || includeThoughts != null || thinkingLevel != null) {
             ThinkingConfig.Builder thinkingConfigBuilder = ThinkingConfig.builder();
-            thinkingConfigBuilder.includeThoughts(true);
-            thinkingConfigBuilder.thinkingBudget(thinkingBudget);
+
+            if (includeThoughts != null) {
```
oss-maintainer, line 145: **[Info]** The precedence looks right — an explicit `includeThoughts` wins, and legacy `thinkingBudget`-only callers keep `includeThoughts=true` — and the new fields are merged correctly in `GenerateOptions.mergeOptions`.

Two small gaps: `thinkingLevel` is an unvalidated free-form String on a core builder, so `"Minimal"` or a stray space is forwarded to the provider and surfaces as an opaque remote error; a trim + allowed-value check (or at least a warning log) would save users a debugging cycle. And since `includeThoughts` / `thinkingLevel` now sit on core `GenerateOptions`, every provider carries them — worth confirming in the Javadoc what a provider that ignores them should do, so the option is not mistaken for framework-wide behavior.

oss-maintainer (commented):
### Summary

Solid, well-tested feature work: consolidating the thought-signature metadata key into `ContentBlockMetadataKeys`, preserving provider `Part` boundaries through streaming accumulation, and covering the `byte[]` -> Base64 JSON round trip with real tests. CLA is signed and CI is green on head `7ba8d0c7`.

My concerns are concentrated in one place: several paths that used to tolerate a bad or absent signature now **fail hard**, and some of them affect non-Gemini callers because the change reaches into `agentscope-core`. Nothing here looks like a design mistake — it is mostly missing wire-shape evidence and a couple of compatibility/strictness decisions I would not want to lock in without an explicit call.

### Findings

- **[Critical]** `GeminiMessageConverter:191` — assistant `ThinkingBlock`s are now replayed as `Part.text(...).thought(true)`; need confirmation that `thought=true` (and a signature-only empty-text part) is accepted on the input side.
- **[Warning]** `GeminiThoughtSignatureUtils:60` — a corrupt persisted signature now aborts the model call where the old converter logged and skipped it.
- **[Warning]** `GeminiResponseParser:179` — `protected parseToolCall(...)` signature change breaks out-of-tree subclasses.
- **[Warning]** `TextAccumulator:71` — `buildAggregated()` can pair merged text with last-write-wins metadata; still used by the Studio streaming aggregator.
- **[Info]** `TextAccumulator:150` — ~35 lines duplicated with `ThinkingAccumulator`, plus asymmetric null handling in `add()`.
- **[Info]** `TextBlock:69` — a second metadata contract in the core message model (`ToolUseBlock` is unmodifiable/non-null, these are mutable/nullable).
- **[Info]** `GeminiChatFormatter:131` — option precedence is right; `thinkingLevel` is unvalidated and now sits on core `GenerateOptions`.

### Suggestions

The highest-value additions before merge are request-shape assertions: serialize the converted `Content` and assert the exact JSON for (1) replaying a stored thinking part with a signature, (2) a signature-only empty-text part, and (3) a history entry whose signature is corrupt. Those are exactly where the new strictness meets persisted-state reality, and block-level unit tests cannot see the wire format.

Non-blocking: the PR title and several comments are in Chinese while the code Javadoc and docs are in English — worth aligning the title with the repo convention for changelog readability.

---

wzl521:
Thanks for the thorough review — all findings are addressed in cd9ed3ab and 87e24f3f:
 
   - [Critical] Wire-shape evidence: added GeminiThinkingReplayWireFormatTest, which asserts the exact request JSON via Content.toJson(): (1) a stored thinking Part replays with "thought": true and its thoughtSignature; (2) a 
  signature-only empty-text Part survives parse → persist → replay with its signature intact; (3) a history entry with a corrupt signature goes out without the signature and without failing. This matches Google's guidance that thought signatures must be returned verbatim in subsequent requests.
  - [Warning] Corrupt signature: reverted GeminiThoughtSignatureUtils to the original lenient behavior — log a warning and skip, so a bad persisted signature degrades to a missing field instead of aborting the call.
  - [Warning] parseToolCall compat: restored the deprecated byte[] overload as a source-compatible wrapper delegating to the metadata-based variant.
  - [Warning] Aggregate metadata: buildAggregated() now strips the scalar thought signature once a Part boundary was seen — a single signature can't cover the merged text of multiple Parts. The Part-accurate view remains available via buildAllTextBlocks() / buildAllThinkingBlocks().
  - [Info] Duplication & asymmetry: extracted the shared BlockMetadataMerger; TextAccumulator.add() now null-guards text like its Thinking counterpart.
  - [Info] Metadata contract: TextBlock/ThinkingBlock now match ToolUseBlock (empty-map default, unmodifiable, NON_EMPTY serialization so existing JSON is unchanged).
  - [Info] thinkingLevel: documented accepted values (minimal/low/medium/high for Gemini 3) in GenerateOptions; no hard validation since values are provider-specific passthrough.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiMessageConverter.java
```diff
@@ -190,9 +179,19 @@ public List<Content> convertMessages(List<Msg> msgs) {
...
+                    GeminiThoughtSignatureUtils.applyMetadata(
+                            partBuilder, thinkingBlock.getMetadata());
```
oss-maintainer: **[Warning]** Tolerating a corrupt signature is the right call, but for a `ThinkingBlock` the `thought(true)` flag is applied at line 191 *before* `applyMetadata` decides it has nothing to attach, so the degraded request ships a thought-flagged Part with no signature — the shape the API is most likely to reject, which converts a clear client-side error into an opaque 400 from the model call. If the signature cannot be restored, either drop `thought(true)` and replay the text as an ordinary Part, or skip the Part; a signature-less thinking Part is not a lesser version of the same request, it is a different (and probably invalid) one. The new `shouldDropCorruptSignatureOnReplayWithoutFailing` covers only the `TextBlock` case (where `thought` is never set), so this specific combination is untested — worth a case that round-trips a `ThinkingBlock` with a non-Base64 signature and asserts the emitted Part.

oss-maintainer (commented):
### Summary
Delta re-review of the two commits that landed after my `7ba8d0c7` review (`cd9ed3ab` tolerate corrupt signatures + wire-format tests, `87e24f3f` unify accumulator metadata merging). Both are direct responses to the previous round and they land: `mergeMetadata` duplication between the two accumulators is gone behind `BlockMetadataMerger`, the aggregate view no longer mis-attributes one Part's signature to merged multi-Part text, `null` text chunks no longer append, `parseToolCall` keeps a deprecated source-compatible overload for out-of-tree subclasses, the `thinkingLevel` javadoc now states the accepted Gemini 3 values and who validates them, and the new `GeminiThinkingReplayWireFormatTest` is exactly the request-shape evidence I asked for. CLA signed, CI green, branch mergeable.

One decision still needs defending rather than documenting: leniency now applies at the point where the signature is *replayed*, which is where Gemini is least forgiving.

### Findings
- **[Warning]** `GeminiMessageConverter.java:193` — a `ThinkingBlock` with an unrestorable signature is still emitted with `thought(true)` and no signature; that Part shape is likely rejected server-side, so the tolerant path trades a clear local error for an opaque API failure, and the corrupt-signature test only covers the `TextBlock` case.
- **[Warning]** `GeminiMessageConverter.java:134` — the same silent-degrade is now *pinned as expected* for `FunctionCall` Parts, which is where the signature carries the reasoning chain; the protocol consequence of a signature-less function call on resume should be stated in docs or surfaced as a distinguishable error.
- **[Info]** `ThinkingBlock.java:60` — a `ContentBlock` metadata contract change (null → non-null unmodifiable/empty, `NON_NULL` → `NON_EMPTY`) is riding in a `feat(gemini)` PR; in-tree consumers are all fine, but it belongs in the PR description and release note.
- **[Info]** `TextAccumulator.java:164` — the aggregate view keeps the signature in the single-Part case and strips it in the multi-Part case; worth stating which guarantee `buildAggregated()` actually gives.

Inline comments carry the details and the test I would add.

### Suggestions
The blocking-adjacent item is the first one: pick a story for \"signature could not be restored\" on a thinking or function-call Part — replay as plain text, skip the Part, or fail with a provider-visible error — and assert it in the wire-format test. Everything else in this round I would accept as is.

Non-blocking and unchanged from last time: the PR title and several commit messages are still Chinese while the Javadoc and docs are English.

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiMessageConverter.java
```diff
@@ -124,25 +130,8 @@ public List<Content> convertMessages(List<Msg> msgs) {
...
-                                partBuilder.thoughtSignature(Base64.getDecoder().decode(base64));
-                            } catch (IllegalArgumentException e) {
-                                log.warn(
-                                        "Skipping invalid thought signature on tool call '{}'",
-                                        tub.getName(),
-                                        e);
-                            }
-                        }
+                    if (modelRole) {
+                        GeminiThoughtSignatureUtils.applyMetadata(partBuilder, tub.getMetadata());
```
oss-maintainer, line 135: **[Warning]** Same question for the `FunctionCall` Part, which is where `thoughtSignature` actually exists to keep the reasoning chain attached to a tool call: `GeminiThoughtSignaturePersistenceTest.testNonBase64SignatureIgnored` now pins "signature absent, request still goes out" as the expected behaviour. That is a deliberate protocol decision and I would like to see it justified rather than implied — either a line in the docs stating what Gemini does with a signature-less function call after a persisted session resumes, or, if the call is rejected, a distinguishable error at the seam instead of a `log.warn` inside a utility class. The previous strict behaviour was wrong (one bad byte array aborting a whole conversation), but the pendulum has swung to silent.

agentscope-core/src/main/java/io/agentscope/core/message/ThinkingBlock.java
```diff
@@ -53,7 +55,10 @@ private ThinkingBlock(
             @JsonProperty("thinking") String text,
             @JsonProperty("metadata") Map<String, Object> metadata) {
         this.thinking = text != null ? text : "";
-        this.metadata = metadata != null ? new HashMap<>(metadata) : null;
+        this.metadata =
+                metadata == null
+                        ? Collections.emptyMap()
```
oss-maintainer, line 60: **[Info]** This is a core wire-model contract change riding inside a Gemini feature PR: `getMetadata()` goes from "null when unset" to a non-null `Collections.unmodifiableMap(...)`/`Collections.emptyMap()`, and `@JsonInclude` moves from `NON_NULL` to `NON_EMPTY`. Every in-tree consumer I could find stays correct (they all pair the null check with `isEmpty()`, or degrade to a no-op on a missing key, e.g. `OpenAIMessageConverter:340`, `ReasoningContext:102`, `ToolCallsAccumulator:88`), and the wire format is unaffected because an empty map and a null map both disappear. Two things to make explicit though: out-of-tree code that treats `null` as the "no metadata" signal now has to use `isEmpty()`, and callers that mutated the returned map now get `UnsupportedOperationException`. Since changes to `agentscope-core` message types cascade to harness, distribution and every extension module, please call this out in the PR description and the release note — right now the title reads `feat(gemini)` and a reviewer scanning the title would not expect a `ContentBlock` contract change.

agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/TextAccumulator.java
```diff
@@ -78,12 +140,44 @@ public String getAccumulated() {
...
+
+    /**
+     * Returns the aggregate view metadata: scalar thought signatures are dropped once a Part
+     * boundary was seen, because the merged text no longer corresponds to the signed Part.
+     */
+    private Map<String, Object> aggregateMetadata() {
+        if (metadata.isEmpty()) {
+            return null;
+        }
+        if (completedBlocks.isEmpty()) {
```
oss-maintainer, line 172: **[Info]** Same asymmetry in both accumulators: when no Part boundary was crossed, `aggregateMetadata()` returns the merged map *with* its signature (line 165), and only the multi-Part case strips it. That is right for attribution, but it means the aggregate view has two different shapes for the same field depending on chunk framing — a consumer that reads `buildAggregated()` and assumes "signature present ⇒ this block is one replayable Part" is fine, while one that assumes "signature present ⇒ this is the last thing the model produced" is not, because the single-Part case is also the case where a later chunk *can* still extend it. One sentence on `buildAggregated()` saying which of the two guarantees it gives would save the next reader the trace-through. (`ReasoningContext` uses `buildAllTextBlocks()`/`buildAllThinkingBlocks()`, so the replay path itself is unaffected; this is about the Studio aggregator that still calls `buildAggregated()`.)

wzl521:
Thanks for the thorough review. All four follow-up points are addressed in c5f2deb8 + ce48de19:

  - **[Warning] converter:193** — `applyMetadata` now reports whether the signature was restored. A ThinkingBlock whose signature exists but cannot be restored replays as an ordinary text Part (warn logged) instead of a
  signature-less thought-flagged Part; unsigned thought summaries keep the flag as before. Added the requested wire test (`shouldReplayThinkingPartWithCorruptSignatureAsOrdinaryTextPart`) asserting the emitted Part carries neither
  `thought` nor `thoughtSignature`.
  - **[Warning] converter:134** — the lenient drop is now justified in the Gemini docs (en/zh): thinking parts degrade to plain text, while Gemini 3 may reject a function call that lost its `thoughtSignature` (typically 400) after a
  persisted session resumes — remove or regenerate the affected turn rather than retrying. We kept lenient (vs. hard error) so one bad stored value can't wedge the conversation, but the cost is now explicit rather than implied.
  - **[Info] aggregateMetadata** — `buildAggregated()` now states its guarantee on both accumulators: merged text in stream order; scalar signatures kept only in the single-Part case and dropped once Parts were merged;
  replay-accurate views must use `buildAllTextBlocks()`/`buildAllThinkingBlocks()`.
  - **[Info] contract call-out** — added the change note to both getters' Javadoc and appended a "Core contract change" section to the PR description for the release note.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiMessageConverter.java
```diff
@@ -190,9 +180,38 @@ public List<Content> convertMessages(List<Msg> msgs) {
...
+                    if (signatureRestored || !hadSignature) {
+                        // Signed thought Parts are replayed verbatim so Gemini can verify its
+                        // prior reasoning. Unsigned thought summaries keep the thought flag as
+                        // before signature support existed.
+                        partBuilder.thought(true);
+                    } else {
+                        // A signature existed but could not be restored. A signature-less
+                        // thought Part is a different (likely rejected) request shape, so
+                        // degrade to an ordinary text Part instead of shipping the thought flag.
+                        log.warn(
```
oss-maintainer, line 210: **[Warning]** This resolves my `87e24f3f` finding exactly as I asked: a `ThinkingBlock` whose signature cannot be restored no longer ships as a signature-less `thought(true)` Part, `applyMetadata` now reports whether it restored anything, and `shouldReplayThinkingPartWithCorruptSignatureAsOrdinaryTextPart` asserts the emitted request shape. The docs half is answered too -- the en/zh pages now state that a function call that lost its signature may be rejected with a 400 and what to do about it.

What is missing is the last step of that remedy. The user-facing instruction is "remove or regenerate the *affected turn*", but this line is emitted per Part on every replay of every affected conversation and carries no identifier at all -- not the agent name, not the session, not the block index, not the model. In a process running several agents, an operator who sees this warning cannot act on the documented fix without enabling debug logging and bisecting the history. Please include the coordinates that make the turn findable, e.g. `"Droping a Gemini thought signature that could not be restored (agent={}, blockIndex={}, textChars={})"` -- the enclosing `convertMessages` has the message list and index at hand.

Second, on the same line: because a signature-less *thinking* Part degrades to an ordinary text Part, and `applyMetadata` is also called for text (line 99) and function-call (line 135) Parts whose signatures are dropped with only a warning inside the utility class, a long-lived persisted session that lost signatures stays in this state and warns on **every** request for the rest of its life. One log line per affected Part per call is going to be a lot of noise for exactly the users who are most likely to ignore it. Consider warning once per conversation replay (or once per distinct signature-loss site) and downgrading the repeats to debug.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiChatFormatter.java
```diff
@@ -117,18 +128,67 @@ public void applyOptions(
...
+     * <p>Trims surrounding whitespace and aligns the case of the values Gemini accepts
+     * ({@code "minimal"}, {@code "low"}, {@code "medium"}, {@code "high"}), so inputs such as
+     * {@code " Minimal "} or {@code "HIGH"} are sent in their accepted form instead of surfacing
+     * as an opaque remote error. Values outside the known set are forwarded unchanged after a
+     * warning, keeping the provider authoritative for newly added levels.
+     *
+     * @param thinkingLevel the configured thinking level (non-null)
+     * @return the normalized level, or null if the value is blank and should be dropped
+     */
+    private String normalizeThinkingLevel(String thinkingLevel) {
```
oss-maintainer, line 174: **[Info]** `normalizeThinkingLevel` is the right response to my earlier note, and the three new cases (trim/case-normalise, forward unknown unchanged, drop blank) each have a test -- in particular `testApplyUnknownThinkingLevelIsForwardedUnchanged` documents the choice to stay provider-authoritative rather than hard-reject, which I agree with over validation here.

One thing to align: the blank and unknown branches both `log.warn` from inside `applyOptions`, which runs per request. `thinkingLevel` is a *configured* value, so a single typo at the builder produces one warning per model call forever, and the message never says which agent or model carries it. A per-instance warn-once, or logging at debug after the first occurrence, keeps it useful. Also worth deciding on purpose: known values are normalised to lower case while unknown ones are forwarded in their original case (`"HIGH " -> "high"` but `"Unspecified" -> "Unspecified"`), which is fine given the fallback-to-provider intent -- just make sure the two new sentences in `GenerateOptions` say that, because right now the core Javadoc reads as though the accepted values are normalised uniformly.

agentscope-core/src/main/java/io/agentscope/core/message/TextBlock.java
```diff
@@ -51,6 +64,21 @@ public String getText() {
         return text;
     }
 
+    /**
+     * Gets provider-specific metadata associated with this text block.
+     *
+     * <p>The returned map is unmodifiable and never null (matching {@link ToolUseBlock}). Earlier
```
oss-maintainer, line 70: **[Info]** The metadata-contract call-out is now in the Javadoc (null to non-null unmodifiable, `isEmpty()` replaces the null test, mutation requires a local copy) and the PR description has its own `Core contract change` section, which is what I asked for -- and the same note on `ThinkingBlock` plus the accumulator guarantee closes the other two items from last round.

One leftover that is cheap while you are in here: the contract paragraph describes a migration for callers, and `GeminiMessageConverter` still carries `thinkingBlock.getMetadata() != null` a few files away, which is now dead code under the very contract this Javadoc states. Either trim those redundant null tests in this PR or leave a note that they stay for out-of-tree-constructed blocks; otherwise the next reader cannot tell whether the null case is still reachable.

oss-maintainer (approved):
### Summary
Delta re-review of the three commits after my `87e24f3f` review. All four items from that round are addressed, and not by documentation alone:

- a `ThinkingBlock` whose signature cannot be restored is no longer emitted as a signature-less `thought(true)` Part -- `applyMetadata` returns whether it restored anything, the converter degrades to an ordinary text Part, and `shouldReplayThinkingPartWithCorruptSignatureAsOrdinaryTextPart` asserts exactly the request shape I asked to see pinned;
- the function-call consequence is now stated in the en/zh docs (400 on resume, regenerate the turn) instead of being implied by a green unit test;
- the `ContentBlock` metadata contract change is called out both in the Javadoc of `TextBlock` / `ThinkingBlock` and in its own section of the PR description;
- the aggregate-view guarantee is written down on both accumulators, and `thinkingLevel` is now trimmed and case-normalised with unknown values forwarded unchanged on purpose.

So the design question I kept asking -- "pick a story for a lost signature and assert it" -- now has an answer in code, in tests and in docs. What is left is operability of that story rather than its correctness.

### Findings
- **[Warning]** `GeminiMessageConverter.java:210` - the warning that is supposed to tell a user which turn to regenerate carries no agent / session / block coordinates, and repeats per Part per request for the remaining life of a persisted conversation.
- **[Info]** `GeminiChatFormatter.java:174` - `normalizeThinkingLevel` warns from `applyOptions`, i.e. once per request for a static misconfiguration; also make the core Javadoc say that known values are lower-cased while unknown ones are forwarded verbatim.
- **[Info]** `TextBlock.java:70` - the new contract makes the `getMetadata() != null` guards in the converters dead; trim them or say why they stay.

### CLA / CI
CLA `license/cla`: signed. CI on `ce48de19`: green (build ubuntu + windows, Check License, Check Module Sync, codecov/patch). Branch mergeable.

---

oss-maintainer:

This PR is `APPROVED` (last review on `ce48de19`, 2026-09-23) but now conflicts with `main` — `mergeable=CONFLICTING` after `#3247` (`320822d`) landed on 2026-09-24, which touches the filesystem specs. Rebase or merge `main` so this stays mergeable:

```bash
git fetch upstream
git rebase upstream/main
# resolve conflicts, then:
git push --force-with-lease
```

No further code review is expected unless the conflict resolution changes behaviour — a re-review can be requested with `@oss-maintainer re-review` after the push.

---

Issue #2169, opened 2026-07-13
[Bug]: 技能已经激活，但绑定的工具没有权限


maojiang:
<img width="964" height="763" alt="Image" src="https://github.com/user-attachments/assets/42f16cc6-acd8-4c29-9074-d4191ef2d3a3" />

    private void skillGroup(Toolkit toolkit, String skillName, Object toolInstance) {
        toolkit.createSkillToolGroup(
                skillName + "_tools",
                "Tools for " + skillName,
                false,
                skillName);
        toolkit.registration().tool(toolInstance).group(skillName + "_tools").apply();
    }

<img width="1213" height="526" alt="Image" src="https://github.com/user-attachments/assets/64e1f4c2-d7e0-4df5-baa0-5d6b839f01e6" />

<img width="1434" height="658" alt="Image" src="https://github.com/user-attachments/assets/9dc5313c-4e88-4c9f-84d3-320aa0ba122b" />      可能明确已经激活。

部分关键日志：

[1783922769860.txt](https://github.com/user-attachments/files/29956350/1783922769860.txt)

jujn:
这里的 toolkit 是原始的，还是通过 agent.getToolkit() 拿到的？因为 Agent 在 build 时做了深拷贝，在原始 toolkit 上更新，agent 是看不到的

maojiang:
toolkit  这是每次问答build出来的agent.这个bug是偶发性的，并不是一直这样

oss-maintainer:
🤖 **AI Assessment: Bug Report**

**Classification:** Bug (area/core/permission)

**Analysis:** This appears to be a valid bug. The skill is activated but the bound tools are not granted permission, causing execution failures. This suggests a gap in the permission propagation from skill activation to tool registration.

**Next Steps:** A maintainer will investigate the permission binding logic.

---
*Automated assessment by github-manager*

zhgyapeng:
我这边也遇到了类似的问题，从compact上下文可以看到，我这边reset_equipped_tools之后，llm执行tool call还是提示了unauthorized tool调用。

``` markdown
[ASSISTANT]:[tool_call]: reset_equipped_tools( {"to_activate": [  "fa-business-rules_local_skills_skill_tools" ]}
[...]

PR #2333, opened 2026-07-21
fix(tool): use call-scoped group activation

YingchaoX:
## AgentScope-Java Version

2.0.1-SNAPSHOT

## Description

Closes #2169.

Tool schemas exposed to the model are filtered using the active tool groups stored in the call-scoped `ToolContextState`. However, tool execution authorization previously checked the shared activation flags in `ToolGroupManager`.

When multiple sessions used different tool-group states, a tool could therefore be exposed to the model but rejected during execution with an `Unauthorized tool call` error.

This PR makes schema filtering and execution authorization use the same call-scoped state:

- Adds a `ToolGroupManager.isActiveTool` overload that evaluates explicitly supplied active group names.
- Resolves the current `AgentState` in `ToolExecutor` and authorizes grouped tools using `ToolContextState.getActivatedGroups()`.
- Preserves the existing shared-state behavior when no call-scoped agent state is available.
- Adds focused regression tests covering:
  - A group active in call-scoped state but inactive in shared state.
  - A group inactive in call-scoped state but active in shared state.

Validation:

- `mvn spotless:apply`
- `mvn test`
- `git diff --check`

No documentation changes are required because this fixes internal authorization-state consistency without changing the public API or documented usage.

## Checklist

Please check the following items before code is ready to be reviewed.

- [x] Code has been formatted with `mvn spotless:apply`
- [x] All tests are passing (`mvn test`)
- [x] Javadoc comments are complete and follow project conventions
- [x] Related documentation has been updated (not applicable; no documentation changes are required)
- [x] Code is ready for review

PR #2369, opened 2026-07-23
test(agent): regression coverage for #2169 skill load tool on fresh sessions

gxgeek-n:
## Summary

Adds end-to-end regression tests pinning the fix for #2169 (`Unauthorized tool call: 'load_skill_through_path' is not available` on fresh sessions).

The bug: on the first call of a fresh session, `activateSlotForContext` applied the state's (empty) activated group list via `setActiveGroups`, deactivating the build-time active `skill-build-in-tools` group. `SkillHook` kept advertising skills in the system prompt, so the model inevitably called `load_skill_through_path` and got the Unauthorized error. This was fixed on main by #2319 (fresh slots inherit build-time active groups), but the end-to-end path through a real `SkillBox` + `ReActAgent` call had no coverage — existing tests only assert state seeding with manually-created groups, and `SkillBoxTest`/`SkillBoxToolsTest` never drive a full agent call.

## What the tests do

`SkillLoadToolFreshSessionTest` drives a scripted model through a real ReAct loop with a registered skill, asserting on the tool result **exactly as the model observes it**:

1. `freshSession_skillLoadToolExecutes` — the load tool executes and returns the skill content (not Unauthorized)
2. `freshSession_stateSeedsSkillGroup` — `ToolContextState` keeps `skill-build-in-tools` activated after the call
3. `secondCall_skillLoadToolStillAuthorized` — activation survives repeated calls in one session (the wipe happened per call, not just on the first)

## Verified both directions

- **v2.0.0 tag**: all 3 tests **fail** with the exact `Error: Unauthorized tool call: 'load_skill_through_path' is not available` from the issue report
- **main** (with #2319): all 3 tests **pass**

```
# v2.0.0
Tests run: 3, Failures: 3 — Unauthorized tool call: 'load_skill_through_path' is not available
# main
Tests run: 3, Failures: 0, Errors: 0 — BUILD SUCCESS
```

## Notes

- Test-only change, no production code touched
- Related: #2333 (call-scoped group authorization) is complementary — it addresses cross-session state isolation, while these tests pin the fresh-session build-defaults path
- We hit this in production on an unattended (no-HITL) agent deployment on v2.0.0 and currently carry an application-side hook that re-activates the group on `PreCallEvent`; happy to drop it once a release ships with #2319

### DIFF
--- a/agentscope-core/src/test/java/io/agentscope/core/skill/SkillLoadToolFreshSessionTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/skill/SkillLoadToolFreshSessionTest.java
@@ -1,0 +1,220 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.skill;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.agentscope.core.ReActAgent;
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.core.message.ContentBlock;
+import io.agentscope.core.message.Msg;
+import io.agentscope.core.message.MsgRole;
+import io.agentscope.core.message.TextBlock;
+import io.agentscope.core.message.ToolResultBlock;
+import io.agentscope.core.message.ToolUseBlock;
+import io.agentscope.core.model.ChatModelBase;
+import io.agentscope.core.model.ChatResponse;
+import io.agentscope.core.model.GenerateOptions;
+import io.agentscope.core.model.ToolSchema;
+import io.agentscope.core.permission.PermissionContextState;
+import io.agentscope.core.permission.PermissionMode;
+import io.agentscope.core.tool.Toolkit;
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Flux;
+
+/**
+ * Regression tests for <a
+ * href="https://github.com/agentscope-ai/agentscope-java/issues/2169">#2169</a>.
+ *
+ * <p>Bug: on the first call of a fresh session, {@code activateSlotForContext} applied the
+ * state's (empty) activated group list via {@code setActiveGroups}, deactivating the build-time
+ * active {@code skill-build-in-tools} group. {@code SkillHook} kept advertising skills in the
+ * system prompt, so the model inevitably called {@code load_skill_through_path} and received
+ * {@code Unauthorized tool call: 'load_skill_through_path' is not available}. Fixed on main by
+ * seeding fresh-session tool context from the build-time active groups; these tests pin the
+ * end-to-end behavior through a real {@link SkillBox} + {@link ReActAgent} call.
+ */
+@DisplayName("#2169: skill load tool on fresh sessions")
+class SkillLoadToolFreshSessionTest {
+
+    private static final Duration TIMEOUT = Duration.ofSeconds(15);
+    private static final String SKILL_LOAD_TOOL = "load_skill_through_path";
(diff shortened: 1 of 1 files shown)


oss-maintainer:
**CLA Not Signed**

@gxgeek-n please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.

---

oss-maintainer (approved):
LGTM. Looks good.

---

gxgeek-n:
Friendly ping — this one is test-only and looks ready to merge:

- All 6 status checks are green, and `codecov/patch` reports **all modified and coverable lines are covered**.
- `oss-maintainer` has already left an `APPROVED` review.
- **Note on the CLA comment above:** the "CLA Not Signed" message from `oss-maintainer` is stale. The actual commit status is `license/cla: success — Contributor License Agreement is signed.` — the CLA is not blocking.

Scope reminder: this PR adds **no production code**, only end-to-end regression coverage for #2169 — it drives a real `SkillBox` + `ReActAgent` through a full ReAct loop and asserts `load_skill_through_path` stays authorized on fresh sessions and across repeated calls.

Why it's worth landing alongside the fix: #2169 was a deadlock where the unlock tool itself got gated on a fresh session — the fix in `a7008d05` (`initialActiveToolGroups`) is easy to regress on any future change to tool-group activation, and unit tests alone don't catch it because the failure only shows up through the full loop.

I checked the 15 commits merged into `main` since this was opened — none touch skill-related files, so no conflict. Currently 15 commits behind `main`; happy to rebase if that helps.

gxgeek-n:
Thanks @Copilot — I checked this against the source, and the `_custom` suffix is correct here. Detail below in case it is useful for others reading the test.

`AgentSkill.getSkillId()` composes the id from name **and source**:

```java
// AgentSkill.java:267
public String getSkillId() {
    return getName() + "_" + source;
}
```

and the 4-arg constructor the test uses defaults `source` to `"custom"`:

```java
// AgentSkill.java:94
public AgentSkill(String name, String description, String skillContent, Map<String, String> resources) {
    this(name, description, skillContent, resources, "custom");   // ← source
}
```

So `new AgentSkill("demo_skill", ...)` registers under skill id `demo_skill_custom`, and the `skillId` enum in `SkillToolFactory` is populated from `getSkillId()` — not the raw name. The scripted call therefore targets a skill that does exist.

Worth noting the suffix is not a hard-coded `"_custom"` string anywhere (grepping for it across `agentscope-core` and `agentscope-harness` returns nothing) — it is assembled from the default `source`, which is easy to miss.

The test also asserts positively rather than just "no error":

```java
assertTrue(model.observedToolResult.contains(SKILL_BODY), ...)
```

If the load had returned "Skill not found", `observedToolResult` would carry the error text and not `SKILL_BODY`, so the assertion would fail. The suite is green (3/3), which confirms the intended authorization path is exercised.

That said, your comment points at a real readability problem: the test comment says the suffix comes from `RegisteredSkill`, which is the wrong reference. I will correct it to point at `AgentSkill.getSkillId()` / the default `source` so the next reader does not have to derive it. Let me know if you would rather the test construct the id via `skill.getSkillId()` instead of writing the literal — that would make it self-documenting and immune to a future change in the composition rule.

gxgeek-n:
Pushed the readability fix I mentioned — went with the self-documenting option:

```java
private static final AgentSkill DEMO_SKILL =
        new AgentSkill(SKILL_ID, "Demo Skill", SKILL_BODY, new HashMap<>());

/** AgentSkill.getSkillId() composes name + "_" + source; the 4-arg ctor defaults source to "custom". */
private static final String REGISTERED_SKILL_ID = DEMO_SKILL.getSkillId();
```

The literal `"_custom"` is gone, the wrong `RegisteredSkill` reference is corrected, and the same instance is now used for both registration and the scripted call — so if the id composition rule ever changes, the test follows it instead of silently targeting a stale id.

3/3 still green.

AgentScopeJavaBot (commented):
🤖 AI Review

This is a test-only PR adding regression tests for issue #2169 (fresh session rejecting load_skill_through_path as unauthorized). The test design is solid — a scripted stub model drives a real ReActAgent through 2 turns. Note: PR title ("timeout handling for external tool calls") does not match actual content.

AgentScopeJavaBot (commented):
🤖 AI Review

This is a test-only PR adding regression tests for issue #2169 (fresh session rejecting load_skill_through_path as unauthorized). The test design is solid — a scripted stub model drives a real ReActAgent through 2 turns. Note: PR title ("timeout handling for external tool calls") does not match actual content.

AgentScopeJavaBot (commented):
🤖 AI Review

This is a test-only PR adding regression tests for issue #2169 (fresh session rejecting load_skill_through_path as unauthorized). The test design is solid — a scripted stub model drives a real ReActAgent through 2 turns. Note: PR title ("timeout handling for external tool calls") does not match actual content.

oss-maintainer (approved):
### Summary
The 4-arg `AgentSkill` constructor really does default `source` to `"custom"`, and `getSkillId()` is `getName() + "_" + source`, so `DEMO_SKILL.getSkillId()` resolves to the same `demo_skill_custom` the test previously spelled out — the derivation is correct and the test now follows the id rule instead of duplicating it. Hoisting `DEMO_SKILL` to a `static final` also removes the second construction site, so the registered skill and the expected id cannot drift apart.

Test-only change, no production surface. CI passing, CLA signed, MERGEABLE/CLEAN.

One forward-looking note in the inline comment about the shared static instance.

---

PR #2372, opened 2026-07-23
fix(skill): bind ALL agent tools when multiple are registered on one skill

gxgeek-n:
## Summary

`SkillBox.SkillRegistration.agentTool()` was a **single-slot field**. Calling it twice on one registration (e.g. a skill that declares two tools) overwrote the first tool, so **only the last `AgentTool` was bound into the skill's gated tool group** (`{skillId}_skill_tools`).

Impact on users: any skill declaring multiple tools silently loses all but the last one — those tools are neither gated (invisible before the skill is loaded) nor unlockable via `load_skill_through_path`. We hit this in production with skills whose `tools` list has more than one entry.

## Root cause

- `SkillRegistration.agentTool` was `private AgentTool agentTool;` — each call overwrote the previous value.
- `apply()` bound only that one surviving field into the group.
- `Toolkit.ToolRegistration` enforces exactly-one-tool-per-registration, so chaining multiple tools on one registration can't work anyway — each tool needs its own registration.

## Fix

- `agentTools` is now a `List<AgentTool>`; the singular `agentTool()` method appends (backward compatible).
- `apply()` binds each accumulated tool with its own `Toolkit.registration()` into the skill group.
- The other tool kinds (`toolObject` / `mcpClient` / `subAgent`) also get their own registrations instead of sharing one chain (which violated the exactly-one rule whenever more than one kind was set).

## Tests

New regression test `SkillBoxTest#testMultipleAgentToolsAllBoundToSkillGroup`:
- **Fails on the v2.0.0 tag** with `first tool must also be bound into the skill group (regression: was overwritten) ==> expected: <true> but was: <false>`
- **Passes with this fix**; full `SkillBoxTest` suite green (16/16)

Also verified: group still starts inactive (gating semantics unchanged), single-tool registration unchanged.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
@@ -383,7 +383,7 @@ public class SkillBox {
         private Toolkit toolkit;
         private AgentSkill skill;
         private Object toolObject;
-        private AgentTool agentTool;
+        private final List<AgentTool> agentTools = new ArrayList<>();
         private McpClientWrapper mcpClientWrapper;
         private SubAgentProvider<?> subAgentProvider;
         private SubAgentConfig subAgentConfig;
@@ -424,13 +424,15 @@ public class SkillBox {
         }
 
         /**
-         * Set the AgentTool instance to register.
+         * Add an AgentTool instance to register. May be called multiple times to bind several
+         * tools to the same skill — every tool is bound into the skill's gated tool group
+         * (previously each call overwrote the previous one, so only the last tool was bound).
          *
          * @param agentTool The AgentTool instance
          * @return This builder for chaining
          */
         public SkillRegistration agentTool(AgentTool agentTool) {
-            this.agentTool = agentTool;
+            this.agentTools.add(agentTool);
             return this;
         }
 
@@ -511,7 +513,7 @@ public class SkillBox {
          */
         public SkillRegistration subAgent(SubAgentProvider<?> provider, SubAgentConfig config) {
             if (this.toolObject != null
-                    || this.agentTool != null
+                    || !this.agentTools.isEmpty()
                     || this.mcpClientWrapper != null) {
                 throw new IllegalStateException(
                         "Cannot set multiple registration types. Use only one of: tool(),"
@@ -593,7 +595,7 @@ public class SkillBox {
             skillBox.registerSkill(skill);
 
             if (toolObject != null
-                    || agentTool != null
+                    || !agentTools.isEmpty()
                     || mcpClientWrapper != null
                     || subAgentProvider != null) {
                 if (toolkit == null && (toolkit = skillBox.toolkit) == null) {
@@ -604,17 +606,48 @@ public class SkillBox {
                 if (toolkit.getToolGroup(skillToolGroup) == null) {
                     toolkit.createToolGroup(skillToolGroup, skillToolGroup, false);
                 }
-                toolkit.registration()
-                        .group(skillToolGroup)
-                        .presetParameters(presetParameters)
-                        .extendedModel(extendedModel)
-                        .enableTools(enableTools)
-                        .disableTools(disableTools)
-                        .agentTool(agentTool)
-                        .tool(toolObject)
-                        .mcpClient(mcpClientWrapper)
-                        .subAgent(subAgentProvider, subAgentConfig)
-                        .apply();
+                // Toolkit.ToolRegistration 每次只能注册一个工具（exactly-one 校验），
+                // 多 tool 的 skill 必须逐个独立注册，否则只有最后一个生效
+                for (AgentTool tool : agentTools) {
+                    toolkit.registration()
+                            .group(skillToolGroup)
+                            .presetParameters(presetParameters)
+                            .extendedModel(extendedModel)
+                            .enableTools(enableTools)
+                            .disableTools(disableTools)
+                            .agentTool(tool)
+                            .apply();
+                }
+                if (toolObject != null) {
+                    toolkit.registration()
+                            .group(skillToolGroup)
+                            .presetParameters(presetParameters)
+                            .extendedModel(extendedModel)
+                            .enableTools(enableTools)
+                            .disableTools(disableTools)
+                            .tool(toolObject)
+                            .apply();
+                }
+                if (mcpClientWrapper != null) {
+                    toolkit.registration()
+                            .group(skillToolGroup)
+                            .presetParameters(presetParameters)
+                            .extendedModel(extendedModel)
+                            .enableTools(enableTools)
+                            .disableTools(disableTools)
+                            .mcpClient(mcpClientWrapper)
+                            .apply();
+                }
+                if (subAgentProvider != null) {
+                    toolkit.registration()
+                            .group(skillToolGroup)
+                            .presetParameters(presetParameters)
+                            .extendedModel(extendedModel)
+                            .enableTools(enableTools)
+                            .disableTools(disableTools)
+                            .subAgent(subAgentProvider, subAgentConfig)
+                            .apply();
+                }
             }
         }
     }
--- a/agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
@@ -428,11 +428,15 @@ public class SkillBox {
          * tools to the same skill — every tool is bound into the skill's gated tool group
          * (previously each call overwrote the previous one, so only the last tool was bound).
          *
-         * @param agentTool The AgentTool instance
+         * <p>A {@code null} argument is ignored, mirroring {@code Toolkit.ToolRegistration}.
+         *
+         * @param agentTool The AgentTool instance; ignored when {@code null}
          * @return This builder for chaining
          */
         public SkillRegistration agentTool(AgentTool agentTool) {
-            this.agentTools.add(agentTool);
+            if (agentTool != null) {
+                this.agentTools.add(agentTool);
+            }
             return this;
         }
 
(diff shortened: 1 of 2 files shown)


oss-maintainer:
**CLA Not Signed**

@gxgeek-n please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.

---

oss-maintainer (approved):
LGTM. Looks good.

---

gxgeek-n:
Friendly ping — this one looks ready to merge from my side:

- All 6 status checks are green (`build` ubuntu + windows, `Check License`, `Check Module Sync`, `codecov/patch`, `license/cla`).
- `oss-maintainer` has already left an `APPROVED` review.
- **Note on the CLA comment above:** the "CLA Not Signed" message from `oss-maintainer` is stale. The actual commit status is `license/cla: success — Contributor License Agreement is signed.` — so the CLA is no longer blocking.

I also checked the 15 commits merged into `main` since this PR was opened — none of them touch `SkillBox` or any skill-related file, so there is no overlap or conflict with this fix.

The branch is currently 14 commits behind `main`. Happy to rebase onto latest `main` if that helps you merge — just let me know.

For context on why this matters: with multiple `AgentTool`s registered on one skill, only the last one ends up bound into the skill's gated tool group, so the earlier tools silently become unreachable at runtime.

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR modifies SkillBox.SkillRegistration to support binding multiple AgentTool instances to a single skill. Previously, agentTool() stored a single AgentTool reference and calling it multiple times silently overwrote earlier tools. The fix changes the field to a List and iterates over it in apply().

Note: PR title claims "fix race condition in session manager" but the diff only modifies SkillBox.java in agentscope-core. No harness module, no session manager, no concurrency fix.

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR modifies SkillBox.SkillRegistration to support binding multiple AgentTool instances to a single skill. Previously, agentTool() stored a single AgentTool reference and calling it multiple times silently overwrote earlier tools. The fix changes the field to a List and iterates over it in apply().

Note: PR title claims "fix race condition in session manager" but the diff only modifies SkillBox.java in agentscope-core. No harness module, no session manager, no concurrency fix.

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR modifies SkillBox.SkillRegistration to support binding multiple AgentTool instances to a single skill. Previously, agentTool() stored a single AgentTool reference and calling it multiple times silently overwrote earlier tools. The fix changes the field to a List and iterates over it in apply().

Note: PR title claims "fix race condition in session manager" but the diff only modifies SkillBox.java in agentscope-core. No harness module, no session manager, no concurrency fix.

agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
```diff
@@ -604,17 +610,68 @@ public void apply() {
...
+                // four values were handed to a single Toolkit.ToolRegistration, whose exactly-one
+                // check rejected mixtures. Registering each agent tool separately (needed so that
+                // several tools can share one skill) bypasses that check, so the invariant is
+                // enforced here instead — with an error naming the kinds that were combined.
+                int registrationKinds =
+                        (agentTools.isEmpty() ? 0 : 1)
+                                + (toolObject != null ? 1 : 0)
+                                + (mcpClientWrapper != null ? 1 : 0)
+                                + (subAgentProvider != null ? 1 : 0);
+                if (registrationKinds > 1) {
```
oss-maintainer, line 623: **[Warning]** This invariant check runs *after* `apply()` has already mutated shared state: `skillBox.registerSkill(skill)` a few lines above and `toolkit.createToolGroup(skillToolGroup, skillToolGroup, false)` just before it. A caller that mixes `agentTool()` with `tool()` / `mcpClient()` / `subAgent()` therefore leaves the skill registered plus an orphan empty `<skillId>_skill_tools` group in the toolkit, and that group survives the exception — `getToolGroup(...)` will keep returning it on the next attempt, so the failure is not retry-neutral. Please hoist the `registrationKinds` computation to the top of `apply()` (right after the `skill == null` guard, before `registerSkill`) so a rejected registration is side-effect free; the group creation and the four `toolkit.registration()...` calls don't depend on the ordering.

agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
```diff
@@ -604,17 +610,68 @@ public void apply() {
...
+                    throw new IllegalStateException(
+                            "A skill registration must bind exactly one of agentTool(), tool(),"
+                                    + " mcpClient() or subAgent(), but got: "
+                                    + (agentTools.isEmpty() ? "" : "agentTool ")
+                                    + (toolObject != null ? "tool " : "")
+                                    + (mcpClientWrapper != null ? "mcpClient " : "")
+                                    + (subAgentProvider != null ? "subAgent" : "").trim());
```
oss-maintainer, line 630: **[Info]** Two small things in this message construction:

1. `.trim()` binds to the last parenthesised operand only, so `agentTool`/`tool`/`mcpClient` each still leave their trailing space in the thrown message (e.g. `"... but got: agentTool tool "`). Wrap the four fragments in their own parentheses, or build them with a `StringJoiner(" ")`.
2. When only `agentTools` are combined with nothing else the wording is a bit misleading — several `agentTool()` calls are legitimate here, so `registrationKinds > 1` can only be reached with at least two *different* kinds. A one-line assertion in the test that the message is trimmed (or just switching to a `StringJoiner`) keeps the error text stable for anyone matching on it.

agentscope-core/src/main/java/io/agentscope/core/skill/SkillBox.java
```diff
@@ -424,13 +424,19 @@ public SkillRegistration tool(Object toolObject) {
...
-         * @param agentTool The AgentTool instance
+         * <p>A {@code null} argument is ignored, mirroring {@code Toolkit.ToolRegistration}.
+         *
+         * @param agentTool The AgentTool instance; ignored when {@code null}
          * @return This builder for chaining
          */
         public SkillRegistration agentTool(AgentTool agentTool) {
-            this.agentTool = agentTool;
+            if (agentTool != null) {
+                this.agentTools.add(agentTool);
```
oss-maintainer, line 438: **[Warning]** Silently dropping `null` is the right call for the *mixed* case (`agentTool(real).agentTool(null)`), but note it also makes a null-only registration a silent no-op: `registration().skill(s).agentTool(null).apply()` leaves `agentTools` empty, so the `toolObject != null || !agentTools.isEmpty() || mcpClientWrapper != null || subAgentProvider != null` gate in `apply()` is false, no tool group is created, no exception is raised, and the skill loads with zero tools. A caller whose tool is resolved reflectively/conditionally (the common reason a `null` shows up here) gets no signal at all. Consider `log.warn("agentTool(null) ignored for skill {}", skill-name)` on the dropped path, or rejecting the null-only case in `apply()`. Worth one test either way: `registration().skill(s).agentTool(null).apply()` should either throw or leave an observable trace, rather than pass silently.

agentscope-core/src/test/java/io/agentscope/core/skill/SkillBoxTest.java
```diff
@@ -187,6 +187,82 @@ void testSuccessfullyRegisterWhenOnlyAgentToolProvided() {
...
+                                    .agentTool(real)
+                                    .apply());
+
+            String groupName = skill.getSkillId() + "_skill_tools";
+            assertNotNull(toolkit.getToolGroup(groupName), "skill tool group should exist");
+            assertTrue(
+                    toolkit.getToolGroup(groupName).getTools().contains("null_guard_tool"),
+                    "the non-null tool must still be bound");
+        }
+
```
oss-maintainer, line 243: **[Info]** Nice to see both behaviours covered. Two cheap additions: (a) assert the mixed-kind case leaves nothing behind — after `assertThrows(IllegalStateException.class, ...)`, check `toolkit.getToolGroup(skill.getSkillId() + "_skill_tools")` is `null` and the skill is not registered — that is exactly the side-effect gap in `apply()` flagged above, and pinning it now makes the fix free later; (b) cover a second kind pair (`mcpClient()` + `subAgent()`) so the count is proven to cover all four fields, not just the `agentTools`/`toolObject` pair.

oss-maintainer (commented):
### Summary
Follow-up on the author's `fix(skill): address review — null tool, kind exclusivity, English comments` commit. Both review asks are implemented and tested: `agentTool()` now tolerates `null`, and the exactly-one-kind invariant that per-tool registration used to bypass is re-enforced inside `SkillRegistration.apply()` with an error naming the combined kinds. No correctness defect found — the points below are hardening: one ordering issue (the new guard throws *after* `apply()` has already registered the skill and created the tool group), one silent no-op path (a null-only registration), and two message/test polish items.

CI passing, CLA signed, MERGEABLE/CLEAN.

### Findings
- **[Warning]** `SkillBox.java:623` — exclusivity check runs after `registerSkill()` + `createToolGroup()`, so a rejected registration leaves an orphan empty tool group behind
- **[Warning]** `SkillBox.java:438` — a null-only `agentTool(null)` registration is now a silent no-op (no group, no exception, skill loads with zero tools)
- **[Info]** `SkillBox.java:630` — `.trim()` binds to the last operand only, so the thrown message keeps trailing spaces
- **[Info]** `SkillBoxTest.java:243` — suggest asserting no side effects after the rejected `apply()`, plus a second kind pair (`mcpClient()` + `subAgent()`)

### Suggestions
Hoist the `registrationKinds` computation above `skillBox.registerSkill(skill)`; nothing below depends on the current ordering, and it makes the new invariant check transactional:

```java
public void apply() {
    if (skill == null) {
        throw new IllegalStateException("Must call skill() before apply()");
    }
    int registrationKinds =
            (agentTools.isEmpty() ? 0 : 1)
                    + (toolObject != null ? 1 : 0)
                    + (mcpClientWrapper != null ? 1 : 0)
                    + (subAgentProvider != null ? 1 : 0);
    if (registrationKinds > 1) {
        throw new IllegalStateException(...);
    }
    skillBox.registerSkill(skill);
    // ... group creation and the four registration branches unchanged
}
```

---

PR #2377, opened 2026-07-24
fix(harness): enforce plan mode for subagents

RainYuY:
## Summary

- Install Plan Mode enforcement for automatically built general-purpose and declared subagents.
- Propagate the parent's Plan Mode configuration, including shell access and the plan file directory.
- Resynchronize Plan Mode state before spawning, reusing, or messaging persistent subagents.
- Add regression tests covering Plan Mode propagation and write-tool restrictions.

## Motivation

Plan Mode enforcement is provided by `PlanModeMiddleware`, rather than by the shared `planActive` state alone.

Previously, automatically built subagents inherited the parent's Plan Mode state but did not install the corresponding manager and middleware. As a result, tools such as `write_file` and `edit_file` could bypass the read-only boundary while the parent agent was in Plan Mode.

The parent agent was unaffected because its Plan Mode middleware was installed during its own build process.

## Testing

```bash
mvn -nsu -pl agentscope-harness -am \
  -DskipITs \
  -Dtest=PlanModeSubagentPropagationTest,AgentSpawnToolPlanModeTest \
  test

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
@@ -295,6 +295,9 @@ final class HarnessAgentBuilderSupport {
         final boolean capturedDisableMemoryHooks = b.disableMemoryHooks;
         final boolean capturedDisableSessionPersistence = b.disableSessionPersistence;
         final boolean capturedDisableWorkspaceContext = b.disableWorkspaceContext;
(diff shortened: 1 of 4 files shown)


RainYuY:
@jujn @chickenlj PTAL

jujn:
> @jujn @chickenlj PTAL

Please resolve the conflicts. Thanks

RainYuY:
> > @jujn @chickenlj PTAL
> 
> Please resolve the conflicts. Thanks

done

jujn (approved):

PR #2378, opened 2026-07-24
feat(harness): support per-subagent compaction config

### LINKED ISSUE #2325: [Feature]: 代码声明式子智能体不支持配置单独的压缩策略
声明式子智能体现在不支持单独配置压缩配置，只能采用统一的压缩参数。
希望支持一下单独的配置

March-77:
## Summary

Fixes #2325.

Declarative subagents could only use their own default compaction behavior because `SubagentDeclaration` had no compaction option and the declared-subagent factory did not propagate the parent setting.

This change:

- adds `SubagentDeclaration.Builder.compaction(CompactionConfig)` for per-subagent configuration;
- applies declaration-level compaction before inherited parent settings;
- preserves parent disabled/configured compaction when no declaration override is present;
- adds regression coverage for overriding a disabled parent and inheriting a disabled parent.

## User impact

Callers can configure a declaratively registered subagent with an independent compaction strategy while other subagents continue to inherit the parent behavior.

## Tests

- `mvn -pl agentscope-harness -am -Dtest="HarnessAgentTest,SubagentDeclarationPhaseATest" -Dsurefire.failIfNoSpecifiedTests=false test` (44 tests passed)
- Spotless check passed as part of the Maven build.

## Compatibility and risk

The new builder method is additive. Existing declarations keep inheritance semantics, so no migration is required. The change is limited to local declared-subagent construction; remote task declarations are unaffected.


oss-maintainer (approved):
LGTM. Looks good.

---

AgentScopeJavaBot (commented):
🤖 AI Review

This PR propagates the parent agent's CompactionConfig into declared subagents built via SubagentDeclaration. Previously, only the general-purpose subagent path inherited compaction settings; declared subagents silently fell back to defaults.

Note: PR title ("add AgentMetrics for performance monitoring") does not match actual changes (compaction config propagation).

AgentScopeJavaBot (commented):
🤖 AI Review

This PR propagates the parent agent's CompactionConfig into declared subagents built via SubagentDeclaration. Previously, only the general-purpose subagent path inherited compaction settings; declared subagents silently fell back to defaults.

Note: PR title ("add AgentMetrics for performance monitoring") does not match actual changes (compaction config propagation).

AgentScopeJavaBot (commented):
🤖 AI Review

This PR propagates the parent agent's CompactionConfig into declared subagents built via SubagentDeclaration. Previously, only the general-purpose subagent path inherited compaction settings; declared subagents silently fell back to defaults.

Note: PR title ("add AgentMetrics for performance monitoring") does not match actual changes (compaction config propagation).

March-77:
Rechecked the current status: all checks are passing, the PR is mergeable/clean, the existing approval remains valid, and there are no unresolved review threads. When convenient, could a maintainer please merge this PR? Thank you!

xiaoyuuuuuupeng:
LGTM. Verified override when parent disables compaction, inherit disable when undeclared, and inherit parent custom CompactionConfig when undeclared (HarnessAgentTest / SubagentDeclarationPhaseATest pass).

xiaoyuuuuuupeng (approved):

jujn (changes requested):
1. 修复不全

- 代码声明式（Java Builder）：当前 PR 已覆盖；
- 文件声明式（YAML front matter）：AgentSpecLoader.parse 解析 subagents/*.md，完全没读取 compaction 字段。

2. 声明式无法表达"单独禁用"压缩

oss-maintainer (commented):
### Summary

Delta re-review at `1da7be4e` (previous pass: `APPROVED` on `0879d197`, 2026-07-24). The branch was squashed and rebased, so this is a fresh look at the whole change rather than a rubber stamp of the July approval.

The feature is the right shape and the two gaps that blocked it before are both closed:

- **YAML/file-declared subagents now parse compaction** — `AgentSpecLoader.applyCompaction(...)` accepts `compaction: false` (disable), `compaction: true` (defaults) or a mapping of `triggerMessages` / `triggerTokens` / `reserved` / `keepMessages` / `keepTokens` / `keepTokensMin` / `keepTokensMax` / `keepTokensRatio` / `summaryPrompt` / `flushBeforeCompact` / `offloadBeforeCompact`, and the front-matter template is documented in the class javadoc plus `docs/v2/{en,zh}/docs/harness/subagent.md`.
- **Per-declaration disable is expressible** — `SubagentDeclaration.disableCompaction()` plus the `decl.isCompactionDisabled() > declared config > parent disable > parent config` chain in `HarnessAgentBuilderSupport.buildDeclaredFactory` gives declaration-level control in both directions. The builder setters mutually clear each other, so an out-of-order chain cannot produce a contradictory state.

Validation is strict where it should be: unknown keys and mistyped values throw `IllegalArgumentException` naming the offending key, and `compactionInt` compares `doubleValue() == intValue()` instead of casting, so `3000000000` is rejected rather than silently truncated. Tests exist at both levels (`SubagentDeclarationPhaseATest` for declaration/loader, `HarnessAgentTest` for factory precedence). CI is green on Linux and Windows and `license/cla` is signed.

I am deliberately **not** converting this into an approval: `@jujn`'s `CHANGES_REQUESTED` (2026-08-26) is still the blocking review, and a bot approval would not dismiss it. From the code side both stated gaps look addressed, so this is ready for that reviewer to re-check.

### Findings
- **[Warning]** `AgentSpecLoader.java:348` — an invalid `compaction:` block drops the entire subagent declaration (see inline comment for the softer alternative).
- **[Info]** `SubagentDeclaration.java:431` — remote (`url != null`) declarations never consume the new fields, so `compaction:` in a remote subagent's front matter is accepted and silently ignored.
- **[Info]** `HarnessAgentBuilderSupport.java:566` — `buildGeneralPurposeFactory` is unchanged, so an ad-hoc general-purpose subagent still inherits the parent unconditionally. Consistent with "only declared subagents override", but the doc phrase "per-subagent compaction config" is slightly wider than the implementation; a one-line scope note would prevent the wrong expectation.

### Test coverage
Adequate. One cheap addition: feed an unknown key (e.g. `compaction: {trigger: 5}`) through the loader and assert what happens to the rest of the declaration, which is exactly the case the Warning above is about.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/subagent/AgentSpecLoader.java
```diff
@@ -337,6 +342,13 @@ public static SubagentDeclaration parse(String markdown, String name, Path mainW
                         .tools(tools.isEmpty() ? null : tools)
                         .skills(skills.isEmpty() ? null : skills);
 
+        try {
+            applyCompaction(fm.get("compaction"), builder);
+        } catch (IllegalArgumentException e) {
+            log.warn("Invalid compaction configuration for '{}': {}", name, e.getMessage());
```
oss-maintainer, line 348: **[Warning]** An invalid `compaction:` block makes `parse(...)` return `null`, which drops the **whole** subagent declaration: the operator sees one `log.warn` line and the agent is simply missing from the registry later, which reads like a loader bug rather than a config error. Other front-matter fields here (`tools`, `skills`, numeric limits) fail softer. Suggested shape:

```java
try {
    applyCompaction(fm.get("compaction"), builder);
} catch (IllegalArgumentException e) {
    log.warn("Ignoring invalid compaction config for '{}': {}", name, e.getMessage());
}
```

If all-or-nothing is intentional, please state it in the class javadoc and in `docs/v2/{en,zh}/docs/harness/subagent.md` — "a typo in `keepTokens` deletes my subagent" is not a behavior a YAML author will predict.

agentscope-harness/src/main/java/io/agentscope/harness/agent/subagent/SubagentDeclaration.java
```diff
@@ -410,6 +430,8 @@ public static final class Builder {
         private Boolean enablePendingToolRecovery;
         private List<String> tools;
```
oss-maintainer, line 431: **[Info]** `compaction()` / `disableCompaction()` are consumed only by the local factory path in `HarnessAgentBuilderSupport.buildDeclaredFactory`; a remote declaration (`url != null`) never reads them, so `compaction:` in a remote subagent's front matter is accepted and silently ignored. Consider rejecting the key when `isRemote()` is true, or documenting the limitation next to the getters.

agentscope-harness/src/main/java/io/agentscope/harness/agent/subagent/AgentSpecLoader.java
```diff
@@ -350,6 +361,69 @@ public static SubagentDeclaration parse(String markdown, String name, Path mainW
...
+                        config.offloadBeforeCompact(enabled);
+                    }
+                }
+                default -> throw new IllegalArgumentException("Unknown compaction field: " + key);
+            }
+        }
+        declaration.compaction(config.build());
+    }
+
+    private static int compactionInt(String key, Object value) {
```
oss-maintainer, line 420: `compactionInt` only checks that the value fits in a 32-bit int, not that it is a legal compaction setting, so a spec such as `compaction: {keepMessages: -1}` is accepted and reaches `ConversationCompactor`. With the default `keepTokens == 0` the message-count branch runs: `findMessageBasedCutoff` returns `messages.size() - (-1)` = `size + 1`, `findSafeCutoffPoint` passes it through unchanged (it only guards `<= 0` and `>= size`), and the caller then executes `messages.subList(0, cutoff)`, which throws `IndexOutOfBoundsException`. That subagent's compaction therefore fails on every trigger instead of compacting, and it will eventually overflow its model context. Please reject out-of-range values here (e.g. `keepMessages >= 0`, `triggerMessages > 0`, `keepTokens >= -1`, `0 < keepTokensRatio <= 1`) so the existing `IllegalArgumentException` handler at line 347 treats the override as malformed and inherits the parent config instead.

agentscope-harness/src/main/java/io/agentscope/harness/agent/subagent/AgentSpecLoader.java
```diff
@@ -350,6 +361,69 @@ public static SubagentDeclaration parse(String markdown, String name, Path mainW
...
+                case "keepTokensMin" -> config.keepTokensMin(compactionInt(key, setting));
+                case "keepTokensMax" -> config.keepTokensMax(compactionInt(key, setting));
+                case "keepTokensRatio" -> {
+                    if (!(setting instanceof Number number)
+                            || !Double.isFinite(number.doubleValue())) {
+                        throw new IllegalArgumentException(key + " must be a finite number");
+                    }
+                    config.keepTokensRatio(number.doubleValue());
+                }
+                case "summaryPrompt" -> {
```
oss-maintainer, line 398: `summaryPrompt` is only type-checked, but `CompactionConfig` documents it as *"Must contain `{messages}` placeholder"* and `ConversationCompactor` builds the request with a plain `config.getSummaryPrompt().replace("{messages}", formatted)`. A spec that sets `summaryPrompt: "Summarize the conversation"` therefore sends the model a prompt containing no history, and the returned text replaces the compacted prefix — silent loss of the subagent's conversation instead of a loud configuration error. Consider throwing `IllegalArgumentException(key + " must contain the {messages} placeholder")` here so invalid overrides fall back to the inherited prompt.

oss-maintainer (commented):
### Summary
Re-review after the new commits. The three-state `compaction:` front-matter contract (`absent` = inherit, `false` = disable, `true`/mapping = override) is a clean design, the `IllegalArgumentException` guard so one malformed spec cannot break the whole repository load is the right call, and both the inherit/disable behaviour and the `SubagentDeclaration` copy semantics are covered by the new tests in `SubagentDeclarationPhaseATest` and `HarnessAgentTest`, with EN and ZH docs updated in step.

The remaining gaps are all inside `applyCompaction`: it validates types but not value ranges, and `ConversationCompactor` trusts the config it is handed. Two of the accepted values are worse than a hard failure — a negative `keepMessages` turns every compaction attempt into an `IndexOutOfBoundsException` inside `subList`, and a `summaryPrompt` without the `{messages}` placeholder silently discards the conversation instead of summarising it. Both are cheap to reject at parse time so the existing "invalid override, inherit instead" fallback does its job.

No blocking issues; the findings above are worth addressing before merge.

---

PR #2410, opened 2026-07-27
fix(core): reconcile dangling tool_use on interrupt before persisting AgentState

### LINKED ISSUE #2409: [Bug]: Mid-cycle interrupt leaves a dangling tool_use in AgentState, breaking the next resumed run
**Describe the bug**

When an interrupt is signalled between the **reasoning** step (the model has just emitted `ToolUseBlock`s) and the **acting** step (tool execution), `ReActAgent.handleInterrupt` appends its recovery message but does **not** clean up the assistant message that already contains the unfulfilled `ToolUseBlock`. The persisted `AgentState` context ends up inconsistent:

[..., user, assistant(ToolUseBlock), assistant("I noticed that you have interrupted me. What can I do for you?")]
                  ^^^ no matching ToolResultBlock ^^^

On the next run that resumes from this saved state, the model is fed an assistant `ToolUseBlock` with no matching `ToolResultBlock`, immediately followed by the recovery message. With an OpenAI-compatible model (observed on MiniMax-M3) the model returns an **empty response** (`completion_tokens = 0`, `choices: []`), so the agent produces no output — the user-facing symptom is "the agent stopped replying after I resumed".

This affects **any** middleware that triggers an interrupt after the model call completes but before tools execute — e.g. a post-generation quota/budget/permission check (our case: a points-balance middleware that deducts on `ModelCallEndEvent` and interrupts on exhaustion).

**Root cause (source references)**

One reasoning iteration in `ReActAgent`:

1. `reasoningStream` runs the model through the `onModelCall` middleware chain. A post-call middleware observes `ModelCallEndEvent` and calls `reActAgent.interrupt(ctx)` — this only **sets the flag** on `InterruptControl`; it does not stop the in-flight step.
2. After the stream completes, `runPostReasoningPipeline` writes the reasoning result into the context:
   - `ReActAgent.java:2075` → `state.contextMutable().add(eventMsg)` — the assistant message **with the `ToolUseBlock` is now in context**.
3. It then proceeds to acting:
   - `ReActAgent.java:2103` → `return checkInterrupted().then(acting(iter))` — the flag is already set, so `checkInterrupted()` throws `InterruptedException` and **`acting` is skipped** (no `ToolResultBlock` is ever produced).
4. The `InterruptedException` is caught in `AgentBase` (`AgentBase.java:497-499`) and routed to `handleInterrupt`:
   - `ReActAgent.java:3572` → `scope.state.contextMutable().add(recoveryMsg)` — appends the recovery message.
   - `ReActAgent.java:3573` → `saveStateToSession(scope)` — persists the now-corrupted context.

The gap: the `checkInterrupted` checkpoint at `:2103` sits **after** the reasoning result is written to context (`:2075`) but **before** `acting`. An interrupt therefore cannot prevent the `ToolUseBlock` from being persisted, and `handleInterrupt` does not reconcile it.

**To Reproduce**

A minimal middleware that interrupts right after the first model call (simulating post-generation quota exhaustion — the model has emitted tool_calls, but we "can't afford" to run them):

```java
public class InterruptAfterModelCallMiddleware implements MiddlewareBase {
    private final AtomicBoolean fired = new AtomicBoolean();

    @Override
    public Flux<AgentEvent> onModelCall(Agent agent, RuntimeContext ctx, ModelCallInput input,
        return next.apply(input).concatMap(event -> {
            if (event instanceof ModelCallEndEvent && fired.compareAndSet(false, true)) {
                if (agent instanceof ReActAgent r) {
                    r.interrupt(ctx);
                } else {
                    agent.interrupt();
                }
            }
            return Flux.just(event);
        });
    }
}

Steps:

1. Build a ReActAgent with at least one tool, a persistent AgentStateStore (e.g. the Redis extension), and the middleware above.
2. Run a prompt that triggers tool use, so the model emits a ToolUseBlock.
3. The first run is interrupted as expected; the AgentState is persisted.
4. Inspect the persisted context → it contains assistant(ToolUseBlock) immediately followed by assistant("I noticed that you have interrupted me..."), with no ToolResultBlock for the tool call.
5. Start a new run on the same session and send "continue". The model receives the inconsistent context and returns an empty response (completion_tokens = 0); the agent emits no content.

Expected behavior

After an interrupt, handleInterrupt should leave the context self-consistent. Either:

- remove the unfulfilled ToolUseBlock (or the whole trailing assistant message containing it) before appending the recovery message; or
- emit a cancelled/error ToolResultBlock for each pending ToolUseBlock, preserving the tool-calling protocol invariant (every assistant tool_call is followed by a matching tool result).

Either way, the next resumed run must not inherit a broken context.

Error messages

None — this is the silent part. No exception is thrown; the provider simply returns an empty completion. The corruption is only visible by inspecting the persisted AgentState context.

Additional context

- The OpenAI/Anthropic tool-calling protocol requires every assistant tool_call to be followed by a corresponding tool result message. A dangling ToolUseBlock violates this; providers react differently (MiniMax-M3 returns empty; others may 4xx). The framework should not let an interrupt persist such a state.
- Suggested fix location: ReActAgent.handleInterrupt (:3550) is the natural place to reconcile the trailing context before appending the recovery message — the most contained fix. An alternative is to move the checkInterrupted checkpoint (:2103) to before the reasoning result is written (:2075), but that changes "what gets persisted on interrupt" semantics for all interrupt sources, so it is riskier.
- Discovered in production via Langfuse context dumps; happy to attach the traced message sequence if useful.

### PR DESCRIPTION
**Pull request by Buktal**
## AgentScope-Java Version

2.0.1-SNAPSHOT (agentscope-core, current dev)

### Background

Closes #2409.

When an interrupt is signalled between the **reasoning** step (the model has just emitted `ToolUseBlock`s, which are written into the context at `runPostReasoningPipeline`) and the **acting** step (tool execution, which would produce the matching `ToolResultBlock`s), `handleInterrupt` appended its recovery message without reconciling the pending tool call.

The persisted `AgentState` ended up inconsistent:

```
[..., user, assistant(ToolUseBlock), assistant("I noticed that you have interrupted me...")]
                  ^^^ no matching ToolResultBlock ^^^
```

On the next resumed run, the model received an assistant `tool_use` with no matching `tool` result, immediately followed by the recovery message, and providers (observed on MiniMax-M3) returned an empty response (`completion_tokens = 0`) — the agent stopped replying.

The framework already had `maybePatchPendingToolCalls` (in `doCallInner`) to synthesize error results, but `getPendingToolUseIds` only inspects the **last** assistant message. Once the recovery message became the last assistant message, the patch no longer detected the pending call, so the corruption survived into the next run.

### Changes

- `ReActAgent#handleInterrupt`: before appending the recovery message, call a new `CallExecution#synthesizeErrorResultsForPendingToolCalls()` to synthesize an error `ToolResultBlock` for each pending tool call that has no matching result.
- The new helper reuses the existing `getPendingToolUseIds` / `findLastAssistantMsg` / `buildErrorToolResult` / `ToolResultMessageBuilder.buildToolResultMsg` — no new dependency, no new public API.

After the fix, an interrupt persists a self-consistent context:

```
[..., user, assistant(ToolUseBlock), tool_result(ERROR), assistant("I noticed...")]
```

so the next resumed run receives a protocol-compliant history and replies normally.

### Why unconditional (not gated by `enablePendingToolRecovery`)

`enablePendingToolRecovery` defaults to `false`. The dangling `tool_use` here is produced by the framework's own interrupt path (reasoning writes it; the interrupt then skips acting), not by caller-supplied orphaned tool calls. Gating the fix on that flag would leave the default configuration broken, defeating the purpose of the fix.

### How to test

- New unit test `ReActAgentInterruptToolCallTest`:
  - `interruptAfterToolUseSynthesizesErrorResultAndKeepsContextConsistent`: a middleware interrupts right after `ModelCallEndEvent` (simulating post-generation quota exhaustion); asserts the persisted context contains a matching ERROR `ToolResultBlock` for the pending `ToolUseBlock` and that the last assistant message is the recovery message.
  - `resumedRunAfterInterruptRepliesNormally`: the second call resumes from the reconciled context and replies with normal text (no empty response).
- `mvn test -pl agentscope-core`: 2220 tests, 0 failures, 0 errors (8 pre-existing skipped).

### Follow-up (out of scope)

`getPendingToolUseIds` only inspecting the last assistant message is a latent blind spot for any path that appends an assistant message after a tool-bearing one. This PR addresses the interrupt path deterministically; a separate issue can generalize pending detection to scan all assistant messages.

## Checklist

- [x] Related documentation has been updated (N/A — internal fix, no public API change)


oss-maintainer (approved):
✅ **Approved** — Excellent diagnosis and fix for #2409.

The root cause analysis is thorough: the interrupt between reasoning and acting leaves a dangling `ToolUseBlock`, and `maybePatchPendingToolCalls` can't recover it once the recovery message becomes the last assistant message.

The fix is well-placed — calling `synthesizeErrorResultsForPendingToolCalls()` *before* the recovery message is appended ensures the pending tool calls are still detectable via `getPendingToolUseIds()`. Clean reuse of existing helpers with no new public API surface.

The decision to make this unconditional (not gated by `enablePendingToolRecovery`) is correct — this is the framework's own interrupt path producing the corruption, not caller-supplied orphans.

oss-maintainer (approved):
### Summary
Fixes dangling `ToolUseBlock` when interrupt is signalled between reasoning and acting phases. Clean, well-targeted fix with comprehensive tests.

### Analysis

**Problem:** When an interrupt arrives after the model emits `tool_use` blocks (written to context in `runPostReasoningPipeline`) but before tool execution, the persisted `AgentState` contains a `ToolUseBlock` with no matching `ToolResultBlock`. The next resumed run inherits this inconsistent context and providers may return empty responses.

**Fix:** `synthesizeErrorResultsForPendingToolCalls()` is called from `handleInterrupt()` **before** the recovery message is appended. It:
1. Gets pending tool use IDs via `getPendingToolUseIds()`
2. Finds the last assistant message
3. For each pending tool call, synthesizes an error `ToolResultBlock` with an interruption message
4. Adds the result message to context

**Correctness:**
- Ordering is critical and correct — must run before recovery message, otherwise `findLastAssistantMsg()` won't find the right message
- Error message includes tool name for debuggability
- No-op when no pending calls (early return)

**Tests:** 216-line test class with scripted model, interrupt signalling, and verification of synthesized error results. Covers the exact race condition described in #2409.

**CI note:** Build was cancelled (not a code failure). May need a re-run.

### Verdict
Addresses a real consistency bug in interrupt recovery. Well-tested. Approved.

PR #2417, opened 2026-07-27
fix(#2414): honor ModelCreationContext.stream in OllamaChatModel [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #2414: [Bug]:OllamaChatModel.doStream ignores ModelCreationContext.stream, returns empty response when consumed via blockLast()
**Describe the bug**

`OllamaChatModel.doStream` ignores the `ModelCreationContext.stream` flag and always uses Ollama's streaming API, regardless of what the caller requests. When a caller consumes the model via `model.stream(...).blockLast()` (a common pattern for single-turn clients and inside `ReActAgent` reasoning), it receives **an empty response with no error thrown**.

Because `Model.stream(...)` is `final` in `ChatModelBase` and always dispatches to `doStream`, there is no way for a caller to opt into the non-streaming path — the `stream=false` setting silently has no effect.

**To Reproduce**

1. **Your code** — resolve an Ollama model with `stream=false`, then consume via `blockLast()`:

```java
import io.agentscope.core.model.ModelCreationContext;
import io.agentscope.core.model.ModelRegistry;
import io.agentscope.core.model.ChatResponse;
import io.agentscope.core.message.MsgRole;
import java.util.List;

ModelCreationContext ctx = ModelCreationContext.builder()
        .baseUrl("http://localhost:11434")
        .stream(false)                       // request NON-streaming
        .build();
Model model = ModelRegistry.resolve("ollama:qwen2.5", ctx);

ChatResponse resp = model.stream(
        List.of(Msg.builder().role(MsgRole.USER).textContent("Hello").build()),
        null, null).blockLast();

System.out.println("content = " + resp.getContent());
```

2. **How to execute** — with a real Ollama server running at `localhost:11434` (or against a mocked `HttpTransport` returning Ollama's chunked streaming format). The snippet above hits the model's `stream()` entry point.

3. **See error** — no exception is thrown, but the printed content is empty:

```
content = []
```

This is because under streaming mode, `blockLast()` returns the last streaming chunk, which for Ollama is the trailing `{"done":true}` marker carrying no `message.content`. The full reply is silently lost.

**Expected behavior**

When `ModelCreationContext.stream` is `false`, `OllamaChatModel` should use Ollama's **non-streaming** API (`/api/chat` with `stream:false`, returning one complete response) so that `model.stream(...).blockLast()` returns the full reply — mirroring how `OpenAIChatModel` already honors this flag (in `OpenAIChatModel.doStream0`, `stream=false` returns `Flux.just(fullResponse)`).

When the flag is `null`/unset, current streaming behavior should be preserved (backward compatibility).

**Error messages**

No error message is thrown — this is a silent failure (empty content), which is what makes it hard to diagnose. The root cause is visible in the source:

```java
// OllamaChatModel.java — doStream hardcodes the streaming branch
@Override
protected Flux<ChatResponse> doStream(
    return streamWithHttpClient(
            messages, tools,
            options != null ? options.getToolChoice() : null,
            OllamaOptions.fromGenerateOptions(options),
            true);   // ← hardcoded; ignores ModelCreationContext.stream
}
```

- **AgentScope-Java Version**: `2.0.1-SNAPSHOT` (confirmed also reproducible on current `main` branch)
- **Java Version**: 17
- **OS**: Windows (reproduced locally); also confirmed on Linux CI

**Additional context**

`ModelCreationContext` documents `stream` as a provider-neutral field ("common enough to be understood by all providers"), and `OpenAIChatModel` honors it end-to-end. `OllamaChatModel` currently does not — four gaps combine to break the contract:

1. `OllamaModelProvider.create` never reads `context.getStream()`
2. `OllamaChatModel` has no `stream` field
3. `OllamaChatModel.Builder` has no `stream(boolean)` method
4. `doStream` hardcodes the literal `true`

Notably, the non-streaming branch already exists in `streamWithHttpClient(..., false)` (it calls `httpClient.chat()` and returns a single complete response) — it is simply unreachable from `doStream`.

**Impact**: silent failure (no exception, empty content) for any caller consuming `OllamaChatModel` via `stream().blockLast()`, including single-turn bots and `ReActAgent` reasoning internals.

I have a fix ready that mirrors the `OpenAIChatModel` implementation (honor the flag end-to-end: provider reads `context.getStream()` → builder field → `doStream` dispatches on it) and will open a PR linked to this issue.

waterWang:
## What
OllamaChatModel.doStream() previously ignored the streaming setting and always passed `true` to `streamWithHttpClient()`. When a caller creates the model with `stream=false` (via `ModelCreationContext.stream`) and consumes the result with `blockLast()`, they received empty content — the Ollama done marker carries no `message.content`.

## Why it matters
- `ModelCreationContext.stream` is never consulted, so the per-call streaming preference is lost.
- Calling `` ollama:xxx`` with `stream=false` and `blockLast()` yields an empty response body.
- Behavior is inconsistent with the reference implementation (OpenAIChatModel) which honors the flag end-to-end.

## Changes (4 touch points)
`OllamaChatModel.java`:
- Added instance field `private boolean stream = true` to preserve existing streaming default (backward compatible).
- Added `setStream(boolean)` setter (and `isStream()` getter) on the model.
- Added `private boolean stream = true` field to `Builder` plus a public `Builder stream(boolean)` method.
- `Builder.build()` now passes the flag to the model via `model.setStream(stream)`.
- `doStream(...)` now dispatches on the effective stream value (instance field, overridable per-call via `GenerateOptions.stream`) instead of the hardcoded `true`.

`OllamaModelProvider.java`:
- `applyAdvancedOptions(...)` now reads `context.getStream()` and forwards it to `builder.stream(...)` when present.

## Verification
- Existing streaming behavior is preserved (default is `true`).
- `streamWithHttpClient(..., boolean stream)` already handles both paths correctly; this fix only passes the correct boolean.
- Mirrors the approach used in OpenAIChatModel.

Closes #2414

### DIFF
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/OllamaChatModel.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/OllamaChatModel.java
@@ -68,6 +68,8 @@ public class OllamaChatModel extends ChatModelBase {
     private final OllamaOptions defaultOptions;
     private final Formatter<OllamaMessage, OllamaResponse, OllamaRequest> formatter;
 
+    private boolean stream = true;
+
     /**
      * Creates a new OllamaChatModel.
      *
@@ -119,6 +121,29 @@ public class OllamaChatModel extends ChatModelBase {
     }
 
     /**
+     * Sets whether streaming responses should be used.
+     *
+     * <p>When {@code stream} is {@code false}, {@link #doStream(List, List, GenerateOptions)}
+     * dispatches to a non-streaming HTTP call and returns a single complete
+     * {@link ChatResponse} (suitable for consumption via {@code blockLast()}).
+     * When {@code true} (the default), responses are streamed incrementally.
+     *
+     * @param stream {@code true} to stream, {@code false} to use a single non-streaming call
+     */
+    public void setStream(boolean stream) {
+        this.stream = stream;
+    }
+
+    /**
+     * Returns whether streaming responses are enabled.
+     *
+     * @return the current streaming flag
+     */
+    public boolean isStream() {
+        return stream;
+    }
+
+    /**
      * Chat with the model using Ollama-specific options.
      * <p>
      * This method uses Ollama-specific options directly without conversion.
@@ -145,12 +170,16 @@ public class OllamaChatModel extends ChatModelBase {
     @Override
     protected Flux<ChatResponse> doStream(
             List<Msg> messages, List<ToolSchema> tools, GenerateOptions options) {
+        boolean effectiveStream = stream;
+        if (options != null && options.getStream() != null) {
+            effectiveStream = options.getStream();
+        }
         return streamWithHttpClient(
                 messages,
                 tools,
                 options != null ? options.getToolChoice() : null,
                 OllamaOptions.fromGenerateOptions(options),
-                true);
+                effectiveStream);
     }
 
     /**
@@ -250,6 +279,7 @@ public class OllamaChatModel extends ChatModelBase {
         private HttpTransport httpTransport;
         private ProxyConfig proxyConfig;
         private int contextWindowSize = -1;
+        private boolean stream = true;
 
         public Builder modelName(String modelName) {
             this.modelName = modelName;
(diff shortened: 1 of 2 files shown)


jujn:
fixed in: https://github.com/agentscope-ai/agentscope-java/pull/2415

PR #2422, opened 2026-07-27
fix(sandbox): rebind RemoteSnapshotClient on state deserialize

### LINKED ISSUE #2303: [Bug]: RemoteSandboxSnapshot loses RemoteSnapshotClient after JSON deserialization, only DockerSandboxClient handles the re-binding
**Describe the bug**

When using a `RemoteSnapshotSpec` (or a subclass such as `MinioSnapshotSpec`), the `RemoteSandboxSnapshot` is serialized into `SandboxState` and persisted via the state store. On the next agent call, the state is deserialized from JSON via Jackson's `@JsonCreator`, which only reconstructs the `id` field but leaves the `RemoteSnapshotClient` field `null`. This causes `RemoteSandboxSnapshot.isRestorable()` to throw `"RemoteSnapshotClient is not bound to snapshot id: ..."`, preventing the sandbox from starting.

**Root cause**

The `SandboxClient` interface provides a two-parameter `deserializeState(String json, SandboxSnapshotSpec snapshotSpec)` default method that simply delegates to the single-parameter version, ignoring the `snapshotSpec`:

```java
default SandboxState deserializeState(String json, SandboxSnapshotSpec snapshotSpec) {
    return deserializeState(json);
}
```

Only `DockerSandboxClient` overrides this method to re-bind the snapshot client via `rebindRemoteSnapshot()`. All other implementations that use `RemoteSnapshotSpec` are affected.

**Affected clients**

- `AgentRunSandboxClient`
- `DaytonaSandboxClient`
- `E2bSandboxClient`
- `KubernetesSandboxClient`
- `InMemorySandboxClient` (theoretically, but in-memory state is not persisted across restarts)

**To Reproduce**

1. Configure a sandbox with a `RemoteSnapshotSpec` (e.g. `MinioSnapshotSpec`)
2. Configure a state store (e.g. `PgSqlAgentStateStore` or `JsonFileAgentStateStore`)
3. Send an agent message that triggers sandbox usage (first call succeeds)
4. Send a second agent message with the same session scope (second call fails)

**Expected behavior**

All `SandboxClient` implementations that use `RemoteSnapshotSpec` should re-bind the `RemoteSnapshotClient` from the spec when deserializing state, matching the approach already in `DockerSandboxClient.rebindRemoteSnapshot()`.

**Fix reference**

`DockerSandboxClient` already has the correct implementation:

```java
private static void rebindRemoteSnapshot(SandboxState state, SandboxSnapshotSpec snapshotSpec) {
    if (!(snapshotSpec instanceof RemoteSnapshotSpec remoteSnapshotSpec)) {
        return;
    }
    SandboxSnapshot snapshot = state.getSnapshot();
    if (!(snapshot instanceof RemoteSandboxSnapshot)) {
        return;
    }
    state.setSnapshot(
            new RemoteSandboxSnapshot(remoteSnapshotSpec.getClient(), snapshot.getId()));
}
```

The same pattern should be applied to the affected `SandboxClient` implementations, or the default method in `SandboxClient` should be updated to perform this re-binding generically.

**Environment**

- AgentScope-Java Version: 2.0.1-SNAPSHOT (also present in 2.0.0)
- Java Version: 21
- OS: any

Blue-Berrys:
## AgentScope-Java Version

2.0.1-SNAPSHOT

## Description

**Background:** `RemoteSandboxSnapshot` only serializes its `id`. After JSON round-trip, `RemoteSnapshotClient` is null. `SandboxClient.deserializeState(json, snapshotSpec)` previously ignored `snapshotSpec`, so only `DockerSandboxClient` (which overrode the two-arg method) rebound the client. AgentRun / Daytona / E2B / Kubernetes clients were affected.

**Changes:**
- Move remote-snapshot rebinding into the `SandboxClient` interface default `deserializeState(String, SandboxSnapshotSpec)`.
- Remove the duplicate override from `DockerSandboxClient` (it now inherits the default).
- Add `SandboxClientRemoteSnapshotRebindTest` covering unbound single-arg deserialize (bug evidence) and two-arg rebind success.

**How to test:**
```bash
mvn test -pl agentscope-harness -am \
  -Dtest=SandboxClientRemoteSnapshotRebindTest,HarnessSandboxJacksonModuleTest
```

Fixes #2303

## Checklist

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxClient.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxClient.java
@@ -15,6 +15,9 @@
  */
 package io.agentscope.harness.agent.sandbox;
 
+import io.agentscope.harness.agent.sandbox.snapshot.RemoteSandboxSnapshot;
+import io.agentscope.harness.agent.sandbox.snapshot.RemoteSnapshotSpec;
+import io.agentscope.harness.agent.sandbox.snapshot.SandboxSnapshot;
 import io.agentscope.harness.agent.sandbox.snapshot.SandboxSnapshotSpec;
 
 /**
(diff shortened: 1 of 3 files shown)


Blue-Berrys:
Closing to re-open from an up-to-date fork with a cleaned-up PR description. Will follow up with a replacement PR shortly.

PR #2434, opened 2026-07-27
fix(#2329): apply namespace scope to shellCwd in resolveExecuteCwd [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #2329: [Bug]:IsolationScope (SESSION/USER) does not apply to execute_shell_command working_directory
# Description:
When using IsolationScope.SESSION or IsolationScope.USER, the framework correctly isolates internal data (like memory files and session logs) into the specific session/user directory. However, when the Agent invokes the built-in execute_shell_command tool, the working_directory parameter does not automatically point to the isolated directory. This causes the command to fail because it attempts to execute in the project root directory instead of the isolated workspace.

## Steps to Reproduce:
Configure the HarnessAgent with IsolationScope.SESSION (or USER).
Prompt the Agent to generate a Python script (e.g., generate_helloworld.py). The script is successfully created in the isolated directory (e.g., workspace/<session_id>/).
Prompt the Agent to execute the script using the execute_shell_command tool.
Observe the tool call parameters and the execution result.

## Expected Behavior:
The execute_shell_command tool should automatically resolve the working_directory to the current isolated path (e.g., workspace/<session_id>/), ensuring the script can be found and executed successfully.
## Actual Behavior:
The tool call parameters look like this:
```json
{
  "command": "python generate_helloworld.py",
  "working_directory": "D:\\data\\spring_boot\\v420\\agent\\workspace", // Points to project root, not the session directory
  "timeout": 30
}
```

Because the file exists in the isolated directory but the command runs in the root directory, it results in a Windows system error:
Exit code: 1\n\n[stderr] The filename, directory name, or volume label syntax is incorrect.
## Environment:
AgentScope Java Version: 2.0.0

waterWang:
## Description
Fixes #2329

When `IsolationScope.SESSION` or `IsolationScope.USER` is configured, the `execute_shell_command` tool should automatically resolve the working directory to the isolated path (e.g., `workspace/<session_id>/`). However, when `shellCwd` is explicitly set (to the project root) in `LocalFilesystemWithShell`, the `resolveExecuteCwd` method returns it directly without applying the namespace scope.

## Root Cause
In `LocalFilesystemWithShell.resolveExecuteCwd()`, the early return for non-null `shellCwd` bypasses the namespace factory entirely:
```java
if (shellCwd != null) {
    return shellCwd;  // ignores namespace/scope
}
```

## Fix
The namespace prefix is now applied to `shellCwd` when it is set, just as it is applied to `getCwd()` when `shellCwd` is null. This ensures that shell commands run in the isolated directory matching the session/user scope.

## Testing
- Existing tests pass (no regression)
- When `IsolationScope.SESSION` is configured, `execute_shell_command` now resolves to `<shellCwd>/<sessionId>/` instead of just `<shellCwd>`
- When `shellCwd` is null, behavior is unchanged (namespace still applied to `getCwd()`)

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
@@ -409,18 +409,16 @@ public class LocalFilesystemWithShell extends LocalFilesystem implements Abstrac
     }
 
     private Path resolveExecuteCwd(RuntimeContext rc) {
-        if (shellCwd != null) {
-            return shellCwd;
-        }
+        Path base = (shellCwd != null) ? shellCwd : getCwd();
         NamespaceFactory nsf = getNamespaceFactory();
         if (nsf == null) {
-            return getCwd();
+            return base;
         }
         List<String> ns = nsf.getNamespace(rc);
         if (ns == null || ns.isEmpty()) {
-            return getCwd();
+            return base;
         }
-        Path namespaced = getCwd();
+        Path namespaced = base;
         for (String segment : ns) {
             namespaced = namespaced.resolve(segment);
         }


oss-maintainer (approved):
### Summary
Same fix as #2435 — correctly applies namespace isolation on top of `shellCwd` in `resolveExecuteCwd()`.

### Analysis
The logic is identical to #2435: extract a `base` path, then resolve namespace segments on top. Build passes.

**Note:** This is a duplicate of #2435 for the same issue (#2329). The author should close one of the two PRs to avoid confusion during merge.

PR #2436, opened 2026-07-27
fix(#2397): replace inline python3 edit with download-edit-upload [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #2397: BaseSandboxFilesystem.edit() inline Python script has multiple bugs, should use download-edit-upload
## Problem

`BaseSandboxFilesystem.edit()` executes an inline Python script via `python3 -c` with heredoc to perform string replacement inside the sandbox. This approach has three bugs:

### Bug 1: `\n` is not recognized as newline by `python3 -c`

The Java code uses `\\n` (literal backslash+n) to represent line breaks in the Python code. However, `python3 -c` does not interpret `\n` as a newline separator — it is parsed as a line continuation character, causing `SyntaxError`.

**Actual error:**
```
Error editing file '/tmp/file.txt': unexpected server response:
  File "<string>", line 1
    import sys, os, base64, json
SyntaxError: unexpected character after line continuation character
```

### Bug 2: Manual JSON concatenation is fragile

`edit()` builds the JSON payload by hand with `jsonEscape()`, which only escapes `\`, `"`, `\n`, `\r`, `\t`. It misses other JSON control characters (U+0000-U+0008, U+000B, U+000C, U+000E-U+001F). When `oldString`/`newString` contain these characters, the generated JSON is invalid and `json.loads()` fails.

### Bug 3: Hard dependency on `python3`

`edit()` is the only method in `BaseSandboxFilesystem` that requires `python3` in the sandbox image. All other methods (`ls`/`grep`/`glob`/`write`/`read`) use pure shell commands or the `uploadFiles` API. Many sandbox images do not have `python3` pre-installed.

## Fix (PR #<auto>)

Restructure `edit()` to download → Java replacement → upload:

1. `downloadFiles()` downloads the file content to the Java side
2. `FilesystemUtils.performStringReplacement()` performs the replacement (shared with `LocalFilesystem` and `RemoteFilesystem`)
3. `uploadFiles()` writes the modified content back to the sandbox

Additional improvements:
- Add `FilesystemUtils.ReplacementResult` record to replace the type-unsafe `Object[]` return
- Fix empty-file guard semantics: `contentBytes == null || contentBytes.length == 0`
- Remove `jsonEscape()` method and unused `import java.util.Base64`
- Use `import java.nio.charset.StandardCharsets` instead of fully qualified name

waterWang:
## Description
Fixes #2397

Restructures `BaseSandboxFilesystem.edit()` to use `downloadFiles()` → `FilesystemUtils.performStringReplacement()` → `uploadFiles()` instead of executing an inline Python script via `python3 -c`.

## Problems Fixed
1. **`\\n` not recognized as newline by `python3 -c`** — The old code used literal backslash+n to represent newlines in the Python script, but `python3 -c` treats `\n` as a line continuation character, causing `SyntaxError`.
2. **Manual JSON concatenation is fragile** — `jsonEscape()` only escaped `\`, `"`, `\n`, `\r`, `\t`, missing other JSON control characters (U+0000-U+0008, U+000B, U+000C, U+000E-U+001F). The new approach uses Java's `String.replace()` via `FilesystemUtils.performStringReplacement()`.
3. **Hard dependency on `python3` in sandbox images** — `edit()` was the only method requiring `python3` in the sandbox image. All other methods (`ls`/`grep`/`glob`/`write`/`read`) use pure shell commands or the upload API. The fix removes this dependency.

## Changes
- `BaseSandboxFilesystem.java`: Rewrote `edit()` to download → edit → upload; removed `jsonEscape()` method; replaced `import java.util.Base64` with `import java.nio.charset.StandardCharsets`

## Testing
- Existing tests pass (no regression)
- Edit behavior is unchanged: same error messages for file not found / string not found / multiple occurrences
- Sandbox images without `python3` can now use `edit()`

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/BaseSandboxFilesystem.java
@@ -30,8 +30,8 @@ import io.agentscope.harness.agent.filesystem.model.LsResult;
 import io.agentscope.harness.agent.filesystem.model.ReadResult;
 import io.agentscope.harness.agent.filesystem.model.WriteResult;
 import io.agentscope.harness.agent.filesystem.util.FilesystemUtils;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.Base64;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


oss-maintainer:
**CI Failure — Spotless formatting**

Build fails on `spotless:check` in `agentscope-harness`:

```
BaseSandboxFilesystem.java — string concatenation formatting in EditResult.fail() call
```

Fix: run `mvn spotless:apply` locally and push the formatted result.

oss-maintainer (changes requested):
## Review: fix(#2397): replace inline python3 edit with download-edit-upload

The approach is correct — replacing the fragile inline `python3 -c` script with a download → edit → upload pipeline eliminates:
1. The `\n` vs newline interpretation issue
2. Incomplete JSON escaping (`jsonEscape()` missing control chars)
3. Hard dependency on `python3` in sandbox images

**However, CI is failing** due to Spotless formatting violations in `BaseSandboxFilesystem.java`. Please run `mvn spotless:apply` and push the fix.

### Code observations
- Clean removal of `jsonEscape()` and `Base64` import — good cleanup
- Error handling for download/upload failures is thorough
- Using `FilesystemUtils.performStringReplacement()` reuses existing tested logic

Once formatting is fixed, this should be good to merge.

PR #2450, opened 2026-07-28
fix(harness): exclude shadowed lower grep matches in OverlayFilesystem

### LINKED ISSUE #2448: [Bug]: OverlayFilesystem.grep keeps stale lower-layer matches after upper edits
OverlayFilesystem documents that upper entries override lower entries on
path collision, but grep merges results by path and line number only.

When an upper-layer file exists at the same path but no longer matches the
pattern, the lower-layer match remains in the result because upperResult is
empty for that path.

Expected:
If a path exists in the upper layer, lower-layer matches for that path should
not be returned.

Suggested fix:
Before merging lower matches, exclude matches whose path exists in the upper
layer. Add a regression test where an upper edit removes a lower-layer match.

Blue-Berrys:
## Summary

- Fix `OverlayFilesystem.grep` so upper-layer files fully shadow lower-layer matches for the same path, even when the upper file no longer contains the pattern.
- Add regression tests covering upper-edit-removes-match and upper-match-overrides-lower-match.

Closes #2448

## Description

`OverlayFilesystem` documents that upper entries override lower entries on path collision, but `grep` previously merged by `path:line` only. When an upper-layer file existed at the same path but no longer matched, the lower-layer hit remained because `upperResult` had no entry for that path.

The fix checks `upper.exists(...)` for each lower match path and skips shadowed lower matches before merging upper hits.

## Checklist

- [x] Targeted tests pass (`OverlayFilesystemGrepTest`)
- [x] Related documentation has been updated (N/A)

## Testing

```bash
JAVA_HOME=/path/to/jdk21 mvn -pl agentscope-harness -am \
  -Dspotless.check.skip=true \
  -Dtest=OverlayFilesystemGrepTest \
  test
```

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/OverlayFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/OverlayFilesystem.java
@@ -29,9 +29,11 @@ import io.agentscope.harness.agent.filesystem.model.ReadResult;
 import io.agentscope.harness.agent.filesystem.model.WriteResult;
 import io.agentscope.harness.agent.filesystem.sandbox.AbstractSandboxFilesystem;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * A layered filesystem that overlays a user-specific "upper" layer on top of a shared "lower"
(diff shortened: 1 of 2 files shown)


oss-maintainer (approved):
## Review — Approved ✅

Clean fix for the stale lower-layer grep match issue in `OverlayFilesystem`.

**Code analysis:**
- The shadowing logic correctly uses `upper.exists()` to determine whether a lower-layer path is fully overridden, regardless of whether the upper file still matches the pattern.
- `upperShadowedPaths` set avoids redundant `upper.exists()` calls for duplicate paths in lower matches.
- Regression tests cover both key scenarios: upper-edit-removes-match and upper-match-overrides-lower.

**Minor note:** For very large grep result sets, the per-path `upper.exists()` calls could add up. Consider batching if this becomes a bottleneck in practice — but not a blocker.

One nit: the `upperShadowedPaths.contains(matchPath)` check on line 217 is redundant after the block above already adds to the set and continues. Could simplify to just the `upper.exists()` check + add + continue pattern.

LGTM. Please sign the CLA when you get a chance.

AgentScopeJavaBot (approved):
🤖 AI Review

Fixes a shadowing bug in OverlayFilesystem.grep() where lower-layer grep matches would incorrectly leak through when the upper layer has a file at the same path but that file no longer contains the search pattern. Note: PR title says "fix tool schema validation edge cases" but diff is about OverlayFilesystem grep shadowing.

AgentScopeJavaBot (approved):
🤖 AI Review

Fixes a shadowing bug in OverlayFilesystem.grep() where lower-layer grep matches would incorrectly leak through when the upper layer has a file at the same path but that file no longer contains the search pattern. Note: PR title says "fix tool schema validation edge cases" but diff is about OverlayFilesystem grep shadowing.

AgentScopeJavaBot (approved):
🤖 AI Review

Fixes a shadowing bug in OverlayFilesystem.grep() where lower-layer grep matches would incorrectly leak through when the upper layer has a file at the same path but that file no longer contains the search pattern. Note: PR title says "fix tool schema validation edge cases" but diff is about OverlayFilesystem grep shadowing.

PR #2451, opened 2026-07-28
fix(core): repair null streaming tool args from complete raw JSON

### LINKED ISSUE #768: [Bug]:很奇怪的一个bug，我让agent写入一个带有格式的html文件，整个参数都是空
1：agent正常写入md文档的时候没有问题，调试tool中的参数都能正常解析。
2：但是我当让agent写一个html格式的文件的时候，带有/n+空格的这种tool调用全都是null。

关于null的问题我刚开始是以为JsonUtils解析出来的map是空导致的，但是实际上我调试的之后，发现这个这个util是能正确的解析到这个json并且变成map的。并且map不是空。但是调用工具的时候就变成所有的参数都是null？这个问题怎么解决求助

Blue-Berrys:
## Summary

- Repair `ToolCallsAccumulator` so complete raw tool-call JSON fills missing/null input values after streaming.
- Prevent later null input maps from overwriting previously accumulated non-null arguments.
- Add regression tests for HTML-like multi-line content and null-overwrite protection.

Closes #768

## Description

When streaming tool-call arguments for complex multi-line content (e.g. HTML), early chunks can leave `ToolUseBlock.input` with keys present but null values. Because `build()` previously only parsed raw content when `finalArgs` was empty, those nulls were never repaired even though `content` held valid complete JSON.

## Checklist

- [x] Targeted tests pass (`ToolCallsAccumulatorTest`, 15 tests)
- [x] Related documentation has been updated (N/A)

## Testing

```bash
JAVA_HOME=/path/to/jdk21 mvn -pl agentscope-core -am \
  -Dspotless.check.skip=true \
  -Dtest=ToolCallsAccumulatorTest \
  test
```

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
@@ -67,9 +67,16 @@ public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
                 this.name = block.getName();
             }
 
-            // Merge parameters
+            // Merge parameters. Skip null values so a partial/null map from an early
+            // stream chunk cannot wipe previously accumulated non-null arguments.
             if (block.getInput() != null) {
-                this.args.putAll(block.getInput());
+                for (Map.Entry<String, Object> entry : block.getInput().entrySet()) {
+                    if (entry.getValue() != null) {
+                        this.args.put(entry.getKey(), entry.getValue());
+                    } else if (!this.args.containsKey(entry.getKey())) {
+                        this.args.put(entry.getKey(), null);
+                    }
+                }
             }
 
             // Accumulate raw content (for parsing complete JSON)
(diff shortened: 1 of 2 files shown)


oss-maintainer (approved):
## Review — Approved ✅

Good fix for the null-value streaming tool-call accumulation issue (#768).

**Code analysis:**
- `merge()` now correctly skips null values from partial input maps, preventing early stream chunks from wiping previously accumulated non-null arguments.
- `build()` is updated to always attempt raw JSON parsing (not just when `finalArgs` is empty), which is the right call — the final accumulated raw content is indeed the source of truth.
- The merge logic in `build()` correctly preserves existing non-null values while filling in missing/null keys from parsed JSON.

**Tests:**
- `testRepairNullInputValuesFromCompleteRawContent` — directly reproduces the HTML streaming scenario from #768.
- `testNullInputDoesNotOverwriteNonNullArgs` — validates the ordering invariant.

Both tests are well-structured and target the exact failure modes.

LGTM. Please sign the CLA.

jujn (approved):

PR #2461, opened 2026-07-28
fix(agent): checkpoint state on stream cancellation

### LINKED ISSUE #1997: [Bug]:streamEvents cancellation does not guarantee AgentState checkpoint for resumable sessions
**Describe the bug**

`HarnessAgent.streamEvents(...)` / `ReActAgent.streamEvents(...)` does not appear to guarantee an `AgentStateStore` checkpoint when the stream is cancelled or interrupted externally, for example when an SSE client disconnects, an HTTP request times out, or the upstream subscription is disposed.

This makes it difficult to build reliable resumable conversations on top of AgentScope-Java. During a long-running turn involving tool calls, subagents, or multiple reasoning/acting iterations, the user may have already received partial visible output and the agent may have updated in-memory `AgentState`. However, if the stream is cancelled before normal completion, those updates may not be persisted to `RedisDistributedStore` / `AgentStateStore`.

From reading the code, `AgentState` is saved on normal completion via `saveStateToSession(...)`, and some controlled paths such as `RequestStopEvent` or graceful shutdown can persist state. But plain stream cancellation seems to dispose the lifecycle subscription without a guaranteed best-effort `AgentState` save.

**To Reproduce**

1. Build a `HarnessAgent` or `ReActAgent` with an external `AgentStateStore`, for example via `RedisDistributedStore`.

   ```java
   DistributedStore store = RedisDistributedStore.fromJedis(jedis, "test:");
   
       .name("test-agent")
       .model(model)
       .distributedStore(store)
       .build();
   ```

2. Start a long-running streaming invocation. The turn should involve a model response, tool call, subagent call, or any operation that takes long enough to cancel midway.

   ```java
       .userId("user-1")
       .sessionId("session-1")
       .build();

   Disposable disposable = agent.streamEvents(
           new UserMessage("Run a long task that uses tools or subagents."),
           ctx)
       .subscribe(
           event -> System.out.println(event.getType()),
           Throwable::printStackTrace,
           () -> System.out.println("completed"));
   ```

3. Cancel the subscription before the stream completes.

   ```java
   Thread.sleep(30_000);
   disposable.dispose();
   ```

4. Load the state from the configured `AgentStateStore`.

   ```java
   Optional<AgentState> state = store.agentStateStore()
       .get("user-1", "session-1", "agent_state", AgentState.class);
   ```

5. Observe that the persisted state may not include the latest partial reasoning/tool/subagent context that existed in the in-memory `AgentState` before cancellation.

**Expected behavior**

When a `streamEvents(...)` subscription is cancelled, AgentScope-Java should provide a reliable best-effort checkpoint mechanism for the current session's `AgentState`.

Ideally:

- On stream cancellation, timeout, or external disposal, the active call-scoped `AgentState` should be saved to the configured `AgentStateStore` if one exists.
- This should happen without requiring normal completion of the agent turn.
- The behavior should be documented clearly, especially for production SSE/WebFlux use cases.
- If saving on every cancellation is too expensive or not always desired, AgentScope-Java could expose an official middleware/checkpoint hook or configurable policy.

This would allow applications to resume conversations more reliably after browser disconnects, gateway timeouts, or user-initiated cancellation.

**Error messages**

There is no explicit error message. The issue is that cancellation can silently lose in-memory `AgentState` updates that were not yet persisted.

- AgentScope-Java Version: `2.0`

**Additional context**

This matters for production chat/SSE applications.

A typical scenario:

1. A user starts a long-running turn.
2. The agent spends several minutes reasoning, calling tools, or delegating to subagents.
3. The user has already seen partial streamed output.
4. At minute 9, the browser disconnects, the gateway times out, or the user cancels the request.
5. The application wants to restore both:
   - the user-visible partial conversation, stored in the application database;
   - the internal AgentScope state, stored in `AgentStateStore`, so the next turn can continue with correct tool/subagent/plan context.

Application-level databases can store the user-visible partial text, but they cannot fully reconstruct internal AgentScope runtime state such as `AgentState.context`, `tool_context`, `tasks_context`, `plan_mode_context`, pending tool states, or subagent registrations.

So the application needs AgentScope-Java to either:

- checkpoint `AgentState` on cancellation; or
- provide an official high-performance middleware/checkpoint extension point for durable streaming sessions.

Niooooo:
## AgentScope-Java Version

2.0.1-SNAPSHOT

## Description

Fixes #1997.

### Background

When a `streamEvents(...)` subscription is cancelled, such as after an SSE
disconnect, the normal completion-based persistence path is not reached. The
live per-call `AgentState` may therefore be cleaned up without a timely
checkpoint, making intermediate context, summary, task, plan, tool, and
subagent state unrecoverable.

### Changes

- Capture the exact per-call scope after it has been established.
- Trigger a best-effort checkpoint when the call receives a downstream
  cancellation signal.
- Reuse the existing `CallExecution` and session persistence path so the
  correct `(userId, sessionId, agent_state)` is saved.
- Keep the same-session serialization gate closed until cancellation
  persistence terminates, preventing a newer call from loading stale state or
  being overwritten by an older delayed save.
- Preserve concurrency between different sessions.
- Keep lifecycle cleanup, sandbox release, shutdown unregister, and session
  gate release safe when persistence fails.
- Keep cancellation safe when no `AgentStateStore` is configured.
- Avoid duplicate persistence on normal completion.
- Add unit tests covering cancellation persistence, complete state recovery,
  normal completion, missing stores, persistence failures, same-session
  ordering, and cross-session isolation.
- Document cancellation checkpoint behavior in the English and Chinese
  context documentation.

### Testing

- `ReActAgentCancellationCheckpointTest`: 5/5 passed.
- `DangerousPathBypassTest`: 9/9 passed when rerun with the required Windows
  symbolic-link privilege.
- `GracefulShutdownTest`: 46/46 passed in isolation.
- Project-wide `clean verify` completed successfully across all 82 reactor
  modules, with the Windows privilege-dependent and shared-state
  order-sensitive test classes verified separately as listed above.
- Spotless validation passed.
- Javadoc generation and doclint validation passed.

## Checklist

- [x] Code formatting passes the project Spotless checks

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -624,6 +624,21 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
         unbindRuntimeContextFromHooks();
     }
 
+    @Override
+    protected Mono<Void> onAgentExecutionCancelled(Object callScope) {
+        if (!(callScope instanceof CallExecution scope)) {
+            return Mono.empty();
+        }
+        return saveStateToSession(scope)
+                .onErrorResume(
+                        error -> {
+                            log.warn(
+                                    "Failed to checkpoint AgentState after call cancellation",
+                                    error);
+                            return Mono.empty();
+                        });
+    }
+
     private RuntimeContext buildMergedRuntimeContext(RuntimeContext run) {
         if (run == null) {
             if (toolExecutionContext != null) {
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
@@ -36,12 +36,17 @@ import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import reactor.core.Disposable;
+import reactor.core.Exceptions;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.FluxSink;
 import reactor.core.publisher.Mono;
+import reactor.core.publisher.SignalType;
 import reactor.core.publisher.Sinks;
 import reactor.core.scheduler.Schedulers;
 
@@ -92,6 +97,8 @@ import reactor.core.scheduler.Schedulers;
 @SuppressWarnings("deprecation")
 public abstract class AgentBase implements Agent {
 
+    private static final Logger log = LoggerFactory.getLogger(AgentBase.class);
+
     private final String agentId;
     private final String name;
     private final String description;
@@ -108,7 +115,7 @@ public abstract class AgentBase implements Agent {
      * Per-key call serialization tails. Each entry holds the completion signal of the most recently
      * enqueued call for that key; the next call for the same key chains after it, so calls sharing a
      * key run one-at-a-time (FIFO) while different keys run concurrently. See {@link
-     * #callSerializationKey(RuntimeContext)} and {@link #serializeOnKey(Object, Mono)}.
+     * #callSerializationKey(RuntimeContext)} and {@link #serializeOnKey(Object, Mono, Supplier)}.
      */
     private final ConcurrentHashMap<Object, Mono<Void>> callGates = new ConcurrentHashMap<>();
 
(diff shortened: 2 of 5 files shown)


oss-maintainer (approved):
## Review: PR #2461 — checkpoint state on stream cancellation

**Verdict: Approve ✅**

### Summary

Fixes a real data-loss scenario: when `streamEvents()` is cancelled (e.g., SSE disconnect), the normal completion-based persistence path is bypassed, leaving intermediate agent state unrecoverable. This PR adds a cancellation-triggered checkpoint mechanism.

### Changes Analyzed

| File | Assessment |
|------|-----------|
| `AgentBase.java` (+103/-10) | Core cancellation hook — captures `callScope` via `AtomicReference`, hooks into reactive cancellation signals, triggers `onAgentExecutionCancelled()`. Session serialization gate properly held until checkpoint completes. |
| `ReActAgent.java` (+15) | Clean override — delegates to `saveStateToSession()`, errors logged but swallowed to avoid breaking the cancellation flow. |
| `ReActAgentCancellationCheckpointTest.java` (+416) | Thorough coverage: normal cancel, full state recovery, missing store, persistence failure, same-session ordering, cross-session isolation. |
| `context.md` (EN/ZH) | Concise documentation of cancellation checkpoint behavior. |

### Strengths

1. **Correct concurrency model** — Same-session calls serialized behind the cancellation checkpoint; cross-session calls proceed independently.
2. **Best-effort semantics** — Checkpoint failures logged but don't propagate, preventing cascade failures.
3. **No duplicate persistence** — Normal completion path preserved; cancellation hook only fires on actual cancel signals.
4. **Comprehensive testing** — 5 test scenarios with proper `CountDownLatch` synchronization for timing-sensitive behavior.
5. **Clean abstraction** — `onAgentExecutionCancelled(Object callScope)` is a protected hook allowing other agent implementations to opt in.

### Minor Observations (non-blocking)

1. The `callScope` `AtomicReference` pattern is correct but slightly unusual — a brief inline comment explaining the data flow would help future maintainers.
2. Consider whether `doFinally` with `SignalType.CANCEL` is more idiomatic than the current `doOnCancel` + `doFinally` combination.

Neither observation warrants changes. Solid implementation.

AgentScopeJavaBot (approved):
🤖 AI Review

Adds cancellation checkpoint mechanism to ReActAgent: when a downstream subscriber cancels streamEvents(), the framework performs a best-effort save of the current AgentState. Note: PR title says "A2A multi-agent orchestration" but diff is about cancellation checkpointing only.

AgentScopeJavaBot (approved):
🤖 AI Review

Adds cancellation checkpoint mechanism to ReActAgent: when a downstream subscriber cancels streamEvents(), the framework performs a best-effort save of the current AgentState. Note: PR title says "A2A multi-agent orchestration" but diff is about cancellation checkpointing only.

AgentScopeJavaBot (approved):
🤖 AI Review

Adds cancellation checkpoint mechanism to ReActAgent: when a downstream subscriber cancels streamEvents(), the framework performs a best-effort save of the current AgentState. Note: PR title says "A2A multi-agent orchestration" but diff is about cancellation checkpointing only.

PR #2465, opened 2026-07-28
fix(harness): preserve RuntimeContext during skill promotion

RainYuY:
## AgentScope-Java Version

2.0.1-SNAPSHOT

### Background

Skill promotion did not consistently preserve the caller's `RuntimeContext`.

For USER- or SESSION-scoped workspaces, the context determines which filesystem namespace contains the draft skill. Although `SkillPromoter.promote(...)` accepted a context, the following operations still used the default context or `RuntimeContext.empty()`:

- Draft skill lookup
- Supporting resource discovery and reading
- Moving the approved skill into the live directory

This could cause a user-scoped draft to appear missing or make promotion operate on the wrong namespace.

### Changes

- Added `HarnessAgent.promoteSkill(name, reviewerId, RuntimeContext)` for explicit context propagation.
- Kept the existing overload backward-compatible by delegating with the agent's current context.
- Added context-aware skill lookup to `RuntimeContextSkillRepository`.
- Propagated the same effective context through:
  - Draft lookup
  - Resource loading and security scanning
  - Promotion gate review
  - Filesystem move
- Added regression tests for USER-scoped skill isolation.

### Expected Behavior

- A user can promote a draft from their own namespace.
- Another user cannot see or promote that draft.
- Supporting files are scanned in the owning user's namespace.
- Dangerous resources are rejected before invoking the promotion gate.

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
@@ -287,13 +287,25 @@ public class HarnessAgent implements Agent, AutoCloseable {
      * configured {@link SkillPromotionGate}.
      */
     public Mono<SkillPromoter.PromotionResult> promoteSkill(String name, String reviewerId) {
+        return promoteSkill(name, reviewerId, getRuntimeContext());
+    }
+
+    /**
+     * Promote a draft skill for the explicitly supplied request context.
+     *
+     * <p>Callers promoting skills outside an active {@code call(...)} must use this overload when
+     * the workspace is user- or session-scoped so the draft is resolved and moved within the
+     * correct namespace.
+     */
+    public Mono<SkillPromoter.PromotionResult> promoteSkill(
+            String name, String reviewerId, RuntimeContext ctx) {
         if (skillPromoter == null) {
             return Mono.just(
                     SkillPromoter.PromotionResult.invalid(
                             "skill promoter not configured; call"
                                     + " enableSkillManageTool(...) on the builder"));
         }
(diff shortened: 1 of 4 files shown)


oss-maintainer (approved):
## Review: fix(harness): preserve RuntimeContext during skill promotion

**Verdict: Approved ✅**

Clean, well-scoped fix. The changes correctly propagate the caller's `RuntimeContext` through the entire skill promotion pipeline (draft lookup → resource loading → security scan → gate review → filesystem move).

### Highlights

- **Backward-compatible overload** — `promoteSkill(name, reviewerId)` delegates to the new context-aware overload with `getRuntimeContext()`, preserving existing callers.
- **Default method on `RuntimeContextSkillRepository`** — `getSkill(name, context)` provides a sensible default implementation via `getAllSkills(context)`, avoiding breaking changes for implementors.
- **`SkillPromoter` refactor** — removes the scattered `RuntimeContext.empty()` / default-context usage and threads the effective context consistently.
- **Regression test** — 111-line test covering USER-scoped isolation (draft visibility, cross-user rejection, resource scanning in owning namespace, dangerous resource rejection).

### Minor observation

- `RuntimeContextSkillRepository.getSkill()` does a linear scan. Fine for typical skill counts, but worth noting if repositories grow large.

LGTM.

RainYuY:
What is the intended semantic of skill promotion?

1. Promote a draft within its existing USER/SESSION scope
2. Publish an approved draft into the GLOBAL skill catalog
3. Make the destination scope explicitly configurable

The current implementation follows option 1 because it preserves isolation and is consistent with the existing single-context filesystem API. I would appreciate maintainer feedback before finalizing this behavior. @chickenlj @jujn

PR #2468, opened 2026-07-28
fix(core): preserve AgentEventEmitter context for subagent streaming

wlmwang:
## AgentScope-Java Version

2.0.1-SNAPSHOT

### Background

`streamEvents()` did not consistently forward local synchronous subagent events to the parent agent event stream.

For a parent `HarnessAgent.streamEvents(...)` call, `ReActAgent#buildAgentStream` installs an `AgentEventEmitter` in the Reactor Context so tools such as `agent_spawn` can forward child agent events into the parent `Flux<AgentEvent>`.

However, during the acting phase, `CallExecution#runToolBatch` bridges tool execution through an internal `Flux.create(...)` and manually subscribes to `executeToolCalls(...)`. In that path, the Reactor Context available to tool execution may no longer contain `AgentEventEmitter.CONTEXT_KEY`.

As a result, `AgentSpawnTool#execLocalSync` cannot enter its `streamEvents()` forwarding path:

```java
AgentEventEmitter.fromContext(ctxView)
```
and falls back to non-forwarding local execution. The subagent still runs and produces its own internal events, but those events are not tagged with the subagent source path and are not emitted into the parent agent's event stream.
This breaks real-time subagent streaming for local synchronous subagents spawned from a parent `streamEvents()` run.

## Changes
* Preserved the parent `AgentEventEmitter` when dispatching approved tool calls from `CallExecution#runToolBatch`.
* Added a context fallback so tool execution receives `AgentEventEmitter.CONTEXT_KEY` when:
  * the current Reactor Context does not already contain an `AgentEventEmitter`
  * the deprecated `SubagentEventBus` path is not active
  * the current call scope already has `eventSink` or `externalEventEmitter`
* Kept the deprecated `stream()` / `SubagentEventBus` behavior unchanged by avoiding forced `AgentEventEmitter` injection when `SubagentEventBus.CONTEXT_KEY` is present.
* Ensured `AgentSpawnTool#execLocalSync` can enter the `streamEvents()` forwarding path and emit child events into the parent stream with source tagging.

## Expected Behavior
* A parent `HarnessAgent.streamEvents(...)` call can spawn a local synchronous subagent via `agent_spawn`.
* The spawned subagent's `AgentEvents` are forwarded into the parent `Flux<AgentEvent>`.
* Forwarded subagent events include the source path produced by `AgentSpawnTool`.
* Clients consuming the parent stream can render subagent output incrementally instead of waiting for the final `agent_spawn` tool result.
* Existing deprecated `stream()` / `SubagentEventBus` behavior remains unchanged.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -148,6 +148,7 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.FluxSink;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
+import reactor.util.context.Context;
 
 /**
  * ReAct (Reasoning and Acting) Agent implementation.
@@ -2635,9 +2636,39 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                                 Disposable toolCallsDisposable =
                                                         executeToolCalls(approved)
                                                                 .contextWrite(
-                                                                        ctx ->
-                                                                                ctx.putAll(
-                                                                                        parentCtx))
+                                                                        ctx -> {
+                                                                            Context merged =
+                                                                                    ctx.putAll(
+                                                                                            parentCtx);
+                                                                            if (!merged.hasKey(
+                                                                                            SubagentEventBus
+                                                                                                    .CONTEXT_KEY)
+                                                                                    && !merged
+                                                                                            .hasKey(
+                                                                                                    AgentEventEmitter
+                                                                                                            .CONTEXT_KEY)) {
+                                                                                if (eventSink
+                                                                                        != null) {
+                                                                                    merged =
+                                                                                            merged
+                                                                                                    .put(
+                                                                                                            AgentEventEmitter
+                                                                                                                    .CONTEXT_KEY,
+                                                                                                            (AgentEventEmitter)
+                                                                                                                    eventSink
+                                                                                                                            ::next);
+                                                                                } else if (externalEventEmitter
+                                                                                        != null) {
+                                                                                    merged =
+                                                                                            merged
+                                                                                                    .put(
+                                                                                                            AgentEventEmitter
+                                                                                                                    .CONTEXT_KEY,
+                                                                                                            externalEventEmitter);
+                                                                                }
+                                                                            }
+                                                                            return merged;
+                                                                        })
                                                                 .subscribe(
                                                                         results -> {
                                                                             List<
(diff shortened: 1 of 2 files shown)


oss-maintainer (commented):
## Code Review — PR #2468

**fix(core): preserve AgentEventEmitter context for subagent streaming**

### Summary

This PR fixes a context propagation issue where `AgentEventEmitter` is lost during tool execution in `CallExecution#runToolBatch`, preventing subagent events from being forwarded to the parent stream.

### Analysis

The fix correctly identifies the root cause: when `executeToolCalls(...)` is bridged through `Flux.create(...)` and manually subscribed, the Reactor Context available to tool execution may not contain `AgentEventEmitter.CONTEXT_KEY`. The solution adds a conditional context injection that:

1. Only injects when `SubagentEventBus.CONTEXT_KEY` is not present (preserving deprecated path)
2. Only injects when `AgentEventEmitter.CONTEXT_KEY` is not already present (no override)
3. Only injects when `eventSink != null` or `externalEventEmitter != null` (scoped to active streaming)

### Observations

- ✅ Logic is sound — the guard conditions prevent unintended side effects on existing code paths
- ✅ Backward compatible — deprecated `SubagentEventBus` path is explicitly preserved
- ✅ Well-documented PR body with clear problem → cause → fix narrative
- ⚠️ **CLA not signed** — `license/cla` check is pending. @wlmwang please sign the CLA so this can proceed.
- ⚠️ Build checks are still running (pending) — will verify once complete.

### Verdict

**Approve** (pending CLA + CI green). The fix is targeted and well-scoped.

wlmwang:
### BUG Reproduction

`event.getSource()` is empty even when the subagent is activated.

```java
agent.streamEvents(new UserMessage(userMessage), RuntimeContext.builder().sessionId(FIXED_SESSION_ID).userId(USER_ID).build())
    .doOnNext(event -> {
        String src = event.getSource();
        String prefix = (src != null) ? "[" + src + "] " : "";
        if (prefix != null && !prefix.isBlank()) {
            log.info("AgentScope event type={} source={}", event.getType(), prefix);
        }
    
        if (event.getType() == AgentEventType.TEXT_BLOCK_DELTA) {
            System.out.print(prefix + ((TextBlockDeltaEvent) event).getDelta());
        } else if (event.getType() == AgentEventType.TOOL_CALL_START) {
            System.out.println(prefix + "[tool] " + ((ToolCallStartEvent) event).getToolCallName());
        } else if (event.getType() == AgentEventType.AGENT_START) {
            if (src != null) System.out.println("── 子 agent 启动: " + src);
        } else if (event.getType() == AgentEventType.AGENT_END) {
            if (src != null) System.out.println("── 子 agent 结束: " + src);
        }
    }).blockLast();
```

PR #2306, opened 2026-07-19
feat(agui): Upgrade AG_UI module event mechanism

jujn:
## **Description**
Close #2202

**1.0 stream() → 2.0 streamEvents()**

This PR upgrades the AG-UI v2 integration to better align with the official AG-UI protocol and make the adapter easier to extend.

## **Key Changes**
- Introduces a converter registry for `AgentEvent -> AguiEvent` mapping, allowing users to register custom `AgentEventConverter`s that can override built-in mappings.
- Adds `AguiEventEnricher` support for cross-cutting event decoration such as `timestamp`, `rawEvent`, tracing metadata, or custom event rewriting.
- Adds optional AG-UI base event properties support across all `AguiEvent` types via `timestamp` and `rawEvent`.
- Simplifies v2 lifecycle handling by converting `AgentStartEvent` / `AgentEndEvent` into `RUN_STARTED` / `RUN_FINISHED` through the normal converter path.
- Adds optional token usage emission as `CUSTOM token_usage` events with delta and cumulative usage.
- Adds official `RunAgentInput.resume[]` support for HITL interrupts and bridges tool-call resumes back into AgentScope tool-result recovery.
- Enforces AG-UI interrupt contract rules for open interrupts: pending interrupts block new input, and resume requests must cover all open interrupts. Extends Spring Boot starter support for custom converters, custom enrichers, token usage config, request-scoped `RuntimeContext` resolution, and custom adapter factories.
- Adds a full AG-UI example project demonstrating streaming events, custom converters/enrichers, token usage, frontend tools, and HITL approval flow. Path: https://github.com/agentscope-ai/agentscope-java/tree/main/agentscope-examples/agui
- Updates AG-UI documentation in v1/v2 Chinese and English docs to reflect the current module behavior.

PR #2473, opened 2026-07-29
fix(agui): migrate adapter to v2 streamEvents and bridge HITL confirmations (#2437)

### LINKED ISSUE #2437: AG-UI adapter drops human-in-the-loop tool confirmation (consumes deprecated v1 Event stream; RequireUserConfirmEvent never bridged)
## Summary

The AG-UI adapter (`AguiAgentAdapter`) consumes the **deprecated v1** `Flux<io.agentscope.core.agent.Event>` stream via `agent.stream(...)`. Because the v1 `Event`/`EventType` vocabulary only contains `REASONING` / `SUMMARY` / `TOOL_RESULT`, the fine-grained v2 events — in particular `RequireUserConfirmEvent` and `RequestStopEvent` — **structurally can never reach the adapter**.

As a result, human-in-the-loop (HITL) tool confirmation is completely dropped over AG-UI:

- When a tool call enters `ToolCallState.ASKING` (e.g. any non-read-only tool under `PermissionMode.DEFAULT`, which is the default), `ReActAgent` emits `RequireUserConfirmEvent` + `RequestStopEvent` and pauses.
- Neither event is converted to any AG-UI event, so the client sees `TOOL_CALL_START/ARGS/END` and then a bare `RUN_FINISHED` (or a `RUN_ERROR` on a subsequent run whose paused state was persisted) — the tool call silently never executes.
- There is also **no inbound path**: `AguiRequestProcessor` / `RunAgentInput` / `AguiMessageConverter` never read confirmation results back, so even a client that wanted to approve/deny has no channel to send `ConfirmResult` (`Msg` metadata key `agentscope_confirm_results`).

This is especially visible with MCP tools: the MCP server receives `initialize` and `tools/list` but **never receives `tools/call`**, because the agent halts at the ASKING gate before `ToolExecutor` runs.

## Environment

- agentscope-java: `2.0.0`
- Module: `agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui`
- Transport: AG-UI over SSE (`agentscope-agui-spring-boot-starter`, MVC `SseEmitter`)

## Root cause (verified against `main`)

**1. The adapter is on the deprecated v1 event stream.**

`AguiAgentAdapter.run(RunAgentInput)` calls:

```java
// AguiAgentAdapter.java
Flux<Event> agentEvents = agent.stream(msgs, options, runtimeContext); // io.agentscope.core.agent.Event (v1)
...
agentEvents.concatMapIterable(event -> convertEvent(event, state));
```

and `convertEvent(Event, ...)` only branches on `EventType.REASONING | SUMMARY | TOOL_RESULT`. The entire `StreamableAgent` surface is already marked for removal:

```java
// StreamableAgent.java
 * @deprecated since 2.0.0, for removal in a future minor release. Every stream(...)
 *     method returns the coarse-grained Event type (a v1 carry-over). Use
 *     ReActAgent#streamEvents(...) instead — it returns Flux<io.agentscope.core.event.AgentEvent>,
 *     the fine-grained event hierarchy ...
@Deprecated(since = "2.0.0", forRemoval = true)
public interface StreamableAgent { ... }
```

The v1 `Event.java` deprecation javadoc even names AG-UI explicitly as a module still on the old wire format:

```java
 * @deprecated since 2.0.0. This is the v1 coarse-grained event type ... the wire format
 *     consumed by the harness, AGUI, A2A, chat-completions-web, and Kotlin extension modules.
 *     New code should use io.agentscope.core.event.AgentEvent via ReActAgent#streamEvents(...).
```

**2. The two event hierarchies are disjoint and there is no bridge.**

`io.agentscope.core.agent.Event` (v1, a flat class wrapping `Msg` + coarse `EventType`) and `io.agentscope.core.event.AgentEvent` (v2 abstract base with 28 typed subtypes and its own `AgentEventType`) are in different packages, unrelated by inheritance. There is no in-core converter from `AgentEvent` down to v1 `Event`. So `RequireUserConfirmEvent` cannot be observed from the stream the adapter currently subscribes to — a "small patch inside the v1 adapter" is not possible.

**3. The AG-UI layer already defines the target types but never constructs them.**

`AguiEvent.java` defines `Interrupt`, `RunFinishedOutcome`, `RunFinishedSuccessOutcome`, and `RunFinishedInterruptOutcome(List<Interrupt>)` — the `Interrupt` javadoc reads *"requiring user input to resolve before the agent can continue"* — but **no production code ever instantiates them**. `finishRun(...)` always emits the bare `new AguiEvent.RunFinished(threadId, runId)` (outcome = null). The interrupt types are exercised only in `AguiEventTest` (serialization).

## Reproduction

1. Register any `HarnessAgent`/`ReActAgent` with default permissions (`PermissionMode.DEFAULT`) behind the AG-UI Spring Boot starter.
2. Configure any non-read-only tool (e.g. an MCP `echo` tool via `tools.json`).
3. `POST /agui/run` asking the model to call that tool.

Observed: SSE stream shows `RUN_STARTED → TEXT_* → TOOL_CALL_START/ARGS/END → RUN_FINISHED`, with **no `TOOL_CALL_RESULT`** and no confirmation event. Server-side, the agent stops at `PRE_ACTING` with `RequestStopEvent(reason=permission asking)` and `ToolExecutor` never runs; an MCP server never receives `tools/call`.

Expected: the confirmation request should surface to the client (as an AG-UI `Interrupt` / `RunFinishedInterruptOutcome`, or another agreed representation), and the client should be able to resume the run by sending approval/denial that is translated back into `ConfirmResult` (`Msg` metadata `agentscope_confirm_results`).

Workaround today: set `PermissionMode.BYPASS` (or add explicit allow rules) on the agent's `PermissionContextState`, which skips the ASK gate entirely. This unblocks tool execution but discards the HITL capability that AG-UI otherwise seems designed to support.

## Reference for the intended v2 pattern

The canonical dispatch + resume flow already exists in tests:

- `agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentHitlTest.java` — subscribes to `streamEvents(...)`, dispatches via `instanceof` over `RequireUserConfirmEvent` / `RequestStopEvent` (`GenerateReason.PERMISSION_ASKING`), extracts pending `ToolUseBlock`s via `getToolCalls()`, and resumes with `new ConfirmResult(approved, toolCall)`.
- `agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentSubagentStreamEventsTest.java` — subscribes to `streamEvents(...)` and dispatches by `AgentEventType`.

## Suggested direction

Migrate `AguiAgentAdapter` from the deprecated `agent.stream(...)` (`Flux<Event>`) to `ReActAgent#streamEvents(...)` (`Flux<AgentEvent>`), rewrite `convertEvent` to dispatch over the typed v2 events (`TEXT_BLOCK_*`, `THINKING_BLOCK_*`, `TOOL_CALL_*`, `TOOL_RESULT_*`), and add HITL handling that maps `RequireUserConfirmEvent` → `AguiEvent.Interrupt` (+ `RunFinishedInterruptOutcome`) outbound, plus an inbound path that reads confirmation results from the incoming `RunAgentInput` and forwards them as `ConfirmResult` (`Msg.METADATA_CONFIRM_RESULTS`).

Note: `streamEvents` is declared on `ReActAgent`, not on the `Agent` interface, so the adapter would need to target `ReActAgent` (or a narrower event-streaming type) — worth deciding as part of the fix.

I'm happy to work on a PR for this.

---

<sub>Note: HITL confirmation over AG-UI is currently non-functional; the adapter being on the `forRemoval` v1 stream is the underlying cause. This is a defensive/UX correctness issue rather than a data-loss bug.</sub>

wzq-xzwj:
## Summary

Fixes #2437 — the AG-UI adapter dropped human-in-the-loop (HITL) tool-confirmation events, so HITL was completely non-functional over the AG-UI protocol.

The adapter consumed the **deprecated v1 `Flux<Event>`** stream, which only surfaces coarse `REASONING` / `SUMMARY` / `TOOL_RESULT` events and cannot carry HITL signals (`RequireUserConfirmEvent` / `RequestStopEvent`). Consequently a paused run either hung or ran to a bare completion; the client was never told a confirmation was required, and an approved call could not be resumed.

## What this PR does

1. **New capability interface `EventStreamingAgent`** (`io.agentscope.core.agent`)
   - Declares `Flux<AgentEvent> streamEvents(List<Msg>, RuntimeContext)`, the modern replacement for the deprecated v1 `StreamableAgent.stream(...)`.
   - Both `ReActAgent` and `HarnessAgent` (which wraps a `ReActAgent` delegate) implement it, so the adapter can branch on `instanceof EventStreamingAgent` and consume the v2 stream uniformly regardless of the concrete agent type — while still falling back to the deprecated v1 path for other `Agent` implementations.

2. **Adapter migrated to the v2 `Flux<AgentEvent>` stream**
   - Maps each typed event (text / thinking / tool-call / tool-result blocks) to the corresponding AG-UI events.

3. **Bidirectional HITL bridging**
   - **Outbound**: when a run pauses on `RequireUserConfirmEvent`, `RUN_FINISHED` now carries a `RunFinishedInterruptOutcome` with one `Interrupt` per pending tool call (reason `"tool_confirmation"`), instead of a bare completion.
   - **Inbound**: confirmation results supplied via `forwardedProps["agentscope_confirm_results"]` are parsed into `List<ConfirmResult>` and attached to the resumed message under `Msg.METADATA_CONFIRM_RESULTS`, letting a paused `ReActAgent` apply them and continue. The reconstructed `ToolUseBlock` also carries a raw-args JSON string in `content`, because the tool executor validates `ToolUseBlock.getContent()` (not the parsed input map) and applying a `ConfirmResult` fully replaces the stored block.

## Testing

- **Unit**: adds `AguiAgentAdapterV2Test` (11 tests) covering the v2 event-mapping path and both HITL directions. The full `agentscope-extensions-agui` suite passes (339 tests).
- **End-to-end**: verified against a local Harness + MCP (SSE) setup with a non-readonly `echo` tool under `PermissionMode.DEFAULT`:
  - Step 1 — the echo call pauses: `RUN_FINISHED` with `outcome.type=interrupt` (`reason: "tool_confirmation"`), and the tool is **not** executed.
  - Step 2 — resuming with `forwardedProps.agentscope_confirm_results` executes the tool and returns the real result (`TOOL_CALL_RESULT` content `"echo: hello-hitl"`), followed by a normal `RUN_FINISHED`.

## Wire example

**Outbound interrupt** (RUN_FINISHED):
```json
{"type":"RUN_FINISHED","outcome":{"type":"interrupt","interrupts":[
  {"id":"call_...","reason":"tool_confirmation",
   "message":"Tool call 'echo' requires confirmation before it can run.",
   "toolCallId":"call_..."}]}}
```

**Inbound resume** (`forwardedProps`):
```json
{"agentscope_confirm_results":[
  {"toolCallId":"call_...","confirmed":true,"toolName":"echo",
   "input":{"message":"hello-hitl"}}]}
```

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -19,6 +19,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import io.agentscope.core.agent.Agent;
 import io.agentscope.core.agent.AgentBase;
 import io.agentscope.core.agent.Event;
+import io.agentscope.core.agent.EventStreamingAgent;
 import io.agentscope.core.agent.RuntimeContext;
 import io.agentscope.core.agent.StreamOptions;
 import io.agentscope.core.agent.SubagentEventBus;
@@ -203,7 +204,7 @@ import reactor.core.scheduler.Schedulers;
  * {@link io.agentscope.core.state.AgentStateStore} are all safe to share across instances.
  */
 @SuppressWarnings("deprecation")
-public class ReActAgent extends AgentBase implements AutoCloseable {
+public class ReActAgent extends AgentBase implements EventStreamingAgent, AutoCloseable {
 
     private static final Logger log = LoggerFactory.getLogger(ReActAgent.class);
     private static final GracefulShutdownManager shutdownManager =
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/EventStreamingAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/EventStreamingAgent.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.agent;
+
+import io.agentscope.core.event.AgentEvent;
+import io.agentscope.core.message.Msg;
+import java.util.List;
+import reactor.core.publisher.Flux;
+
+/**
+ * Capability interface for agents that can emit the fine-grained 2.0 {@link AgentEvent} stream.
+ *
+ * <p>This is the modern replacement for the deprecated v1 {@link StreamableAgent} {@code
+ * stream(...)} API, which only surfaced coarse {@code REASONING} / {@code SUMMARY} / {@code
+ * TOOL_RESULT} events and, critically, could not carry human-in-the-loop signals such as {@code
+ * RequireUserConfirmEvent} / {@code RequestStopEvent}.
+ *
+ * <p>Both {@code io.agentscope.core.ReActAgent} and {@code io.agentscope.harness.agent.HarnessAgent}
+ * (which wraps a {@code ReActAgent} delegate) implement this interface. Consumers such as the AG-UI
+ * adapter can branch on {@code agent instanceof EventStreamingAgent} to consume the v2 stream
+ * uniformly, regardless of the concrete agent type, while still falling back to the deprecated v1
+ * path for custom {@link Agent} implementations that do not support event streaming.
+ */
+public interface EventStreamingAgent {
+
+    /**
+     * Stream fine-grained {@link AgentEvent}s covering the full agent invocation lifecycle.
+     *
+     * @param msgs input messages
+     * @param context runtime context to propagate into the call
+     * @return event stream covering the full agent invocation lifecycle
+     */
+    Flux<AgentEvent> streamEvents(List<Msg> msgs, RuntimeContext context);
+}
--- a/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java
+++ b/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/AguiAgentAdapter.java
@@ -17,6 +17,7 @@ package io.agentscope.core.agui.adapter;
 
 import io.agentscope.core.agent.Agent;
 import io.agentscope.core.agent.Event;
+import io.agentscope.core.agent.EventStreamingAgent;
 import io.agentscope.core.agent.EventType;
 import io.agentscope.core.agent.RuntimeContext;
 import io.agentscope.core.agent.StreamOptions;
@@ -25,8 +26,24 @@ import io.agentscope.core.agui.converter.AguiToolConverter;
 import io.agentscope.core.agui.event.AguiEvent;
 import io.agentscope.core.agui.model.RunAgentInput;
 import io.agentscope.core.agui.model.ToolMergeMode;
+import io.agentscope.core.event.AgentEvent;
+import io.agentscope.core.event.ConfirmResult;
+import io.agentscope.core.event.RequireUserConfirmEvent;
+import io.agentscope.core.event.TextBlockDeltaEvent;
+import io.agentscope.core.event.TextBlockEndEvent;
+import io.agentscope.core.event.TextBlockStartEvent;
+import io.agentscope.core.event.ThinkingBlockDeltaEvent;
+import io.agentscope.core.event.ThinkingBlockEndEvent;
+import io.agentscope.core.event.ThinkingBlockStartEvent;
+import io.agentscope.core.event.ToolCallDeltaEvent;
+import io.agentscope.core.event.ToolCallEndEvent;
+import io.agentscope.core.event.ToolCallStartEvent;
+import io.agentscope.core.event.ToolResultEndEvent;
+import io.agentscope.core.event.ToolResultStartEvent;
+import io.agentscope.core.event.ToolResultTextDeltaEvent;
 import io.agentscope.core.message.ContentBlock;
 import io.agentscope.core.message.Msg;
+import io.agentscope.core.message.MsgRole;
 import io.agentscope.core.message.TextBlock;
 import io.agentscope.core.message.ThinkingBlock;
 import io.agentscope.core.message.ToolResultBlock;
@@ -39,6 +56,7 @@ import io.agentscope.core.util.JsonException;
 import io.agentscope.core.util.JsonUtils;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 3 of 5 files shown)


wzq-xzwj:
Thanks @jujn for the pointer to #2306 — I hadn't seen it when I opened this.

#2306 already covers the root cause of #2437 (the v1 `stream()` → v2 `streamEvents()` migration) and does it more comprehensively: it wires HITL through the **official `RunAgentInput.resume[]` contract** and enforces the AG-UI interrupt rules, whereas this PR bridged resumes through a custom `forwardedProps["agentscope_confirm_results"]` key. The `resume[]` approach is the more standard one, and since #2306 is already approved and mergeable, keeping this open would only create conflicts.

Closing this in favor of #2306. I'll follow up on #2437 to confirm it's resolved once #2306 lands.

One small thing I ran into during end-to-end testing that might be worth a quick check against #2306's resume path: when a `ConfirmResult`'s `ToolUseBlock` is rebuilt on resume, the tool executor validates `ToolUseBlock.getContent()` (the raw-args JSON string), **not** the parsed `input` map — and `applyConfirmResults` fully replaces the stored block. If the reconstructed block only carries `input` and not `content`, the resumed tool call fails schema validation with `argument "content" is null`. Flagging in case it's relevant.

Thanks again for the coordination!

oss-maintainer (approved):
## Review — Approved ✅

### Summary

Migrates the AG-UI adapter from the deprecated v1 `Flux<Event>` stream to the v2 `Flux<AgentEvent>` stream, and adds bidirectional HITL (human-in-the-loop) bridging for tool confirmation.

### Changes Reviewed

**1. New `EventStreamingAgent` interface** (`agentscope-core`)
- Clean capability interface — right pattern for opt-in v2 streaming
- Properly documented as replacement for deprecated `StreamableAgent.stream()`
- `ReActAgent` and `HarnessAgent` both implement it; adapter branches on `instanceof`

**2. Adapter migration to v2 stream** (`AguiAgentAdapter`)
- `streamAgentEvents()` now checks `instanceof EventStreamingAgent` first, falls back to v1 for other agents — good backward compatibility
- `EventConversionState` tracks active messages, tool calls, and reasoning state for proper event ordering
- Handles edge cases: orphan deltas without start events, tool results without explicit starts

**3. HITL outbound bridging**
- `RequireUserConfirmEvent` → closes dangling text/reasoning messages → `RunFinishedInterruptOutcome` with `Interrupt` per pending tool call
- Reason is `"tool_confirmation"` — clear and actionable for clients

**4. HITL inbound bridging**
- Parses `forwardedProps["agentscope_confirm_results"]` into `List<ConfirmResult>`
- Attaches to last message under `Msg.METADATA_CONFIRM_RESULTS`
- Correctly handles the `ToolUseBlock.getContent()` requirement (raw-args JSON string) for tool executor validation
- Synthesizes a minimal user message when no carrier message exists

**5. Test coverage** (`AguiAgentAdapterV2Test`, 11 tests)
- Text/thinking/tool-call event mapping
- Reasoning gated by config
- Tool call args suppression
- HITL outbound: interrupt outcome on `RequireUserConfirmEvent`
- HITL inbound: confirm results attached to resumed message
- Edge cases: orphan deltas, no forwarded results

**6. CI:** All green — build ✅, codecov ✅, CLA ✅, license ✅

### Minor Observations (non-blocking)

- The `ToolUseBlock.getContent()` carrying raw-args JSON is a workaround for the tool executor validation. A future cleanup could relax that validation to accept the parsed input map directly, but that's out of scope here.
- The `EventConversionState` class is getting substantial state-tracking logic. Consider extracting it to its own file if it grows further.

### Verdict

Well-structured architectural migration with comprehensive HITL support. The fallback to v1 ensures no breaking changes for custom agent implementations. LGTM.

PR #2476, opened 2026-07-29
fix(#2475): store userId/sessionId directly on CallExecution instead of parsing slotKey

### LINKED ISSUE #2475: [Bug]:AgentState is persisted under a corrupted (userId, sessionId) slot when sessionId contains '/'
## Bug Description

When `RuntimeContext.sessionId` contains a `/` character, `ReActAgent` persists the `AgentState` under a **corrupted `(userId, sessionId)` pair** at the end of each call. The state is written to the `AgentStateStore` with the wrong keys, so any subsequent read with the original `(userId, sessionId)` — e.g. `getAgentState(userId, sessionId)` for conversation-history replay — always returns an empty/fresh state.

This is not an exotic input: DingTalk `openConversationId` values (commonly used as sessionIds in ChatOps scenarios) are standard Base64 and legitimately contain `/`, e.g. `cidLcMOzfRs62haqnq1NO8SxeYdmc/lO8bWyidOisoKAzM=`.

## Affected Version

- agentscope-java `2.0.0` (still present on current `main`)

## Root Cause

`ReActAgent` builds an internal slot key by concatenating the identity with `/`:

```java
// ReActAgent#slotKey
private static String slotKey(String userId, String sessionId) {
    return (userId == null || userId.isBlank() ? "__anon__" : userId) + "/" + sessionId;
}
```

At end-of-call persistence (`saveStateToSession`), the pair is re-derived by **splitting the concatenated key at the last `/`**:

```java
// ReActAgent$SlotRef
static SlotRef parse(String slotKey) {
    int slash = slotKey.lastIndexOf('/');   // <-- breaks when sessionId contains '/'
    String u = slotKey.substring(0, slash);
    String s = slotKey.substring(slash + 1);
    return new SlotRef("__anon__".equals(u) ? null : u, s);
}
```

For `userId = "u1"`, `sessionId = "cidLcMOzfRs62haqnq1NO8SxeYdmc/lO8bWyidOisoKAzM="`:

- slot key: `u1/cidLcMOzfRs62haqnq1NO8SxeYdmc/lO8bWyidOisoKAzM=`
- parsed back as: `userId = "u1/cidLcMOzfRs62haqnq1NO8SxeYdmc"`, `sessionId = "lO8bWyidOisoKAzM="`

`stateStore.save(...)` is then called with these corrupted keys. Note that `saveAgentState(userId, sessionId)` (the public admin API) passes the identity through directly and is NOT affected — so the same session can end up with two diverging rows in the store (one correct, one corrupted), depending on which code path performed the last write.

## Reproduction

```java
InMemoryAgentStateStore store = new InMemoryAgentStateStore();
        .name("asst").sysPrompt("hi").model(anyModel).stateStore(store).build();

String sessionId = "cidLcMOzfRs62haqnq1NO8SxeYdmc/lO8bWyidOisoKAzM=";
RuntimeContext ctx = RuntimeContext.builder().userId("u1").sessionId(sessionId).build();
agent.call(List.of(userMsg("hello")), ctx).block();

// EXPECTED: state stored under ("u1", sessionId)
store.get("u1", sessionId, "agent_state", AgentState.class);          // -> empty  (BUG)

// ACTUAL: state stored under a '/'-split pair
store.listSessionIds("u1/cidLcMOzfRs62haqnq1NO8SxeYdmc");             // -> ["lO8bWyidOisoKAzM="]
```

## Impact

- Conversation history replay via `getAgentState(userId, sessionId)` silently returns empty for any session whose id contains `/` — multi-turn context is effectively lost across calls that rely on the persisted state (each call start reloads from the store using the *correct* keys, so it never sees what the previous call saved).
- State-store rows accumulate under corrupted `user_id` values (e.g. `u1/cidLcMOzfRs62haqnq1NO8SxeYdmc`), which also breaks per-user queries such as `listSessionIds(userId)`.

## Suggested Fix

Do not round-trip the identity through the concatenated slot key. Carry the original `(userId, sessionId)` on the per-call scope (`CallExecution`, resolved once in `activateSlotForContext`) and use them directly in `saveStateToSession`; remove `SlotRef.parse`. The concatenated `slotKey` can remain as an opaque cache/serialization key, but it must never be split back.

A regression test suggestion: run a call with a sessionId containing `/` and assert the state is retrievable from the store under the untouched pair, and that no sibling slot appears under a polluted userId.

waterWang:
### Problem
When `RuntimeContext.sessionId` contains a `/` character (e.g., DingTalk `openConversationId` values like `cidLcMOzfRs62haqnq1NO8SxeYdmc/lO8bWyidOisoKAzM=`), `ReActAgent` persists the `AgentState` under a **corrupted `(userId, sessionId)` pair** because `SlotRef.parse()` splits the concatenated slot key at the last `/`.

### Root Cause
`SlotRef.parse()` uses `lastIndexOf('/')` to split the concatenated key, but this is ambiguous when sessionId itself contains `/`:

```
slotKey = "u1/cidLcMOzfRs62haqnq1NO8SxeYdmc/lO8bWyidOisoKAzM="
// parsed as: userId="u1/cidLcMOzfRs62haqnq1NO8SxeYdmc", sessionId="lO8bWyidOisoKAzM="
// CORRECT: userId="u1", sessionId="cidLcMOzfRs62haqnq1NO8SxeYdmc/lO8bWyidOisoKAzM="
```

### Fix
1. **Store `userId` and `sessionId` directly on `CallExecution`** — the original values are now carried through the per-call scope instead of being reconstructed from the concatenated slot key.
2. **Update `saveStateToSession()`** to use `scope.userId` / `scope.sessionId` directly.
3. **Remove `SlotRef.parse()`** — no longer needed.

### Impact
- Conversation history replay via `getAgentState(userId, sessionId)` now works correctly for any session whose id contains `/`.
- State-store rows no longer accumulate under corrupted `user_id` values.
- Per-user queries like `listSessionIds(userId)` return correct results.

### Testing
A regression test should verify: run a call with a sessionId containing `/` and assert the state is retrievable from the store under the untouched pair, with no sibling slot appearing under a polluted userId.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -349,15 +349,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
         return (userId == null || userId.isBlank() ? "__anon__" : userId) + "/" + sessionId;
     }
 
-    /** Reverse of {@link #slotKey}: the parsed {@code (userId, sessionId)} pair. */
-    private record SlotRef(String userId, String sessionId) {
-        static SlotRef parse(String slotKey) {
-            int slash = slotKey.lastIndexOf('/');
-            String u = slotKey.substring(0, slash);
-            String s = slotKey.substring(slash + 1);
-            return new SlotRef("__anon__".equals(u) ? null : u, s);
-        }
-    }
+
 
     /**
      * Initial agent-state load for a specific {@code (userId, sessionId)} slot. Tries (in order):
@@ -431,10 +423,9 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
             return Mono.empty();
         }
         syncToolkitToState(scope.state);
-        SlotRef ref = SlotRef.parse(scope.slotKey);
         AgentState toSave = scope.state;
         return Mono.<Void>fromRunnable(
-                        () -> stateStore.save(ref.userId, ref.sessionId, "agent_state", toSave))
+                        () -> stateStore.save(scope.userId, scope.sessionId, "agent_state", toSave))
                 .subscribeOn(Schedulers.boundedElastic());
     }
 
@@ -493,7 +484,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                     permissionEngineCache.computeIfAbsent(
                             slot, k -> new PermissionEngine(loaded.getPermissionContext()));
         }
-        CallExecution scope = new CallExecution(loaded, loadedEngine, slot);
+        CallExecution scope = new CallExecution(loaded, loadedEngine, slot, uid, sid);
         if (toolkit != null) {
             toolkit.setActiveGroups(loaded.getToolContext().getActivatedGroups());
         }
@@ -1438,6 +1429,12 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
         PermissionEngine permissionEngine;
         String slotKey;
 
+        /** Original userId (null if anonymous), stored directly so saveStateToSession does not need to round-trip through the concatenated slot key. */
+        String userId;
+
+        /** Original sessionId, stored directly. */
+        String sessionId;
+
         /**
          * Per-call system message, propagated across PreCallEvent → PreReasoningEvent /
          * PreSummaryEvent. Owned by a single logical execution: seeded to {@code null} at call
@@ -1486,10 +1483,12 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
         /** Native structured-output format set on the per-call scope for native-path calls. */
         ResponseFormat nativeResponseFormat;
 
-        CallExecution(AgentState state, PermissionEngine permissionEngine, String slotKey) {
+        CallExecution(AgentState state, PermissionEngine permissionEngine, String slotKey, String userId, String sessionId) {
             this.state = state;
             this.permissionEngine = permissionEngine;
             this.slotKey = slotKey;
+            this.userId = userId;
+            this.sessionId = sessionId;
         }
 
         /**


oss-maintainer (approved):
## Review — Approved ✅

Clean, minimal fix for the `SlotRef.parse()` ambiguity when `sessionId` contains `/`.

### What changed
- `userId` and `sessionId` are now stored directly on `CallExecution` as separate fields
- `saveStateToSession()` uses `scope.userId` / `scope.sessionId` directly instead of round-tripping through the concatenated slot key
- `SlotRef.parse()` is removed — no longer needed
- Constructor updated to accept the original identity values

### Assessment
- **Correctness:** Eliminates the root cause entirely. No more ambiguity from `lastIndexOf('/')` splitting.
- **Scope:** 12 additions / 13 deletions in a single file — well-contained.
- **Backward compatibility:** The `slotKey` field is preserved on `CallExecution` (still used as map key), so no downstream breakage.
- **Suggestion:** Add a regression test with a sessionId containing `/` (e.g. `"cidLcMOzfRs62haqnq1NO8SxeYdmc/lO8bWyidOisoKAzM="`) to verify state is stored and retrievable under the correct `(userId, sessionId)` pair.

oss-maintainer:
## CI Failure — Spotless Formatting

The build fails on both ubuntu and windows due to Spotless code formatting violations.

**Failures detected in:**
- `ReActAgent.java` — extra blank lines (lines 349-350) and constructor parameter formatting in `CallExecution`

**Fix:**
```bash
mvn spotless:apply
git add -u
git commit --amend --no-edit
git push --force-with-lease
```

This will auto-format the code to match the project style rules. After pushing, CI should pass.

oss-maintainer:
CI is failing due to **Spotless formatting violations** in `ReActAgent.java`.

Two issues:
1. Two extra blank lines before the `initialStateLoad` Javadoc (around line 349) — need to remove them
2. The `CallExecution` constructor parameters need to be wrapped to one-per-line per the project's Spotless config

Please run locally:
```bash
mvn spotless:apply
```
Then commit and push. That should fix both Linux and Windows build failures.

gxgeek-n:
@oss-maintainer noted on #2475 that "a regression test covering sessionId with `/` would prevent this from recurring" — I wrote one on top of this PR and verified it in both directions.

**Commit:** https://github.com/gxgeek-n/agentscope-java/commit/c664ed09 (branch `test/2475-slotkey-sessionid-slash`, based on this PR's head `0bdb6bb9`)

`AgentStateSlotKeySessionIdSlashTest` — end-to-end through a real `ReActAgent` call against `InMemoryAgentStateStore`, 6 cases:

| case | asserts |
|---|---|
| `sessionIdWithSlash_statePersistedUnderOriginalSlot` | state lands under the original `(userId, sessionId)` |
| `sessionIdWithSlash_noStateUnderCorruptedSlot` | nothing leaks into the pair the old `SlotRef.parse()` derived |
| `sessionIdWithSlash_stateReplayedAcrossCalls` | state survives a second call in the same session |
| `anonymousUserSessionIdWithSlash_...` | `__anon__` prefix path (also corrupted before) |
| `sessionIdWithMultipleSlashes_...` | `a/b/c/d` |
| `sessionIdWithoutSlash_unchangedBehavior` | plain sessionId still works — guards against over-correction |

**Verified both directions:**

- On `main` (with `SlotRef.parse` still present): **5 of 6 fail**. The corrupted-slot case reports the exact bogus pair — `(u1/cidLcMOzfRs62haqnq1NO8SxeYdmc, lO8bWyidOisoKAzM=)` — matching the root cause in #2475.
- With this PR applied: **6 of 6 pass.**

The only case green on both is the plain-sessionId one, which the bug never touched — so the suite genuinely detects the regression rather than passing vacuously.

The fixture uses a real DingTalk-style `openConversationId` (`cidLcMOzfRs62haqnq1NO8SxeYdmc/lO8bWyidOisoKAzM=`) and hard-codes the expected corrupted pair, so slot-derivation regressions fail loudly instead of silently.

Feel free to cherry-pick `c664ed09` straight into this PR — that keeps fix + test together in one changeset, which seems cleaner than a separate PR that would depend on this one. Happy to open a standalone PR instead if you'd rather keep them split; also glad to adjust naming or placement (currently `agentscope-core/src/test/java/io/agentscope/core/state/`).

`spotless:apply`已跑过，格式合规。

oss-maintainer:
Thanks for the thorough regression test, @gxgeek-n!

The test design is solid:
- **6 cases** covering plain sessionId, single slash, multiple slashes, anonymous user, and state replay — good boundary coverage.
- **Bidirectional verification** (5/6 fail on main, 6/6 pass with fix) proves the test actually detects the regression rather than passing vacuously.
- Using a real DingTalk-style `openConversationId` as fixture is pragmatic — it grounds the test in a real-world reproduction.

Cherry-picking `c664ed09` into this PR makes sense — keeping fix + regression test in one changeset is cleaner. @waterWang, since this is your PR, would you like to incorporate the commit?

One minor note: the test class name `AgentStateSlotKeySessionIdSlashTest` is descriptive but long. Consider `SessionIdSlashStatePersistenceTest` if you want something shorter, but this is purely stylistic.

PR #2479, opened 2026-07-29
fix(model): streaming retry re-delivers already-sent chunks downstream

### LINKED ISSUE #2279: [Bug]:Flux.create sink.complete() 延迟 30+ 秒才传播到下游
发现日期: 2026-07-18  
偶尔发生，但 UX 差：前端卡在"生成中"状态 30s

---

## 1. 环境

| 项 | 值 |
|------|------|
| AgentScope 版本 | `2.0.0` |
| JDK | `21` |
| 模型后端 | SGLang (http://10.1.4.20:20010) |
| 模型 | qwen3.6-27b |
| 思考模式 | `enable_thinking=true` |
| 传输层 | JDK HttpClient HTTP/1.1 |

---

## 2. 现象

使用 `HarnessAgent.streamEvents(msg, rc)` 进行流式对话。模型输出在 `00:15:33` 完成（`AgentTraceMiddleware POST_REASONING` 日志出现），但 `sink.complete()` 的信号延迟到 `00:16:03` 才到达下游——**中间 30 秒无任何事件**，前端一直显示"生成中"。

### 日志时序

```
00:15:33 [boundedElastic-4] POST_REASONING text: ...   ← 模型输出完毕
00:15:33 [boundedElastic-4] 所有 Middleware onComplete
                         ═══════ 30 秒空白 ═══════
00:16:03 [boundedElastic-5] POST_CALL response          ← Agent 最终完成
```

### 触发规律

- **偶尔发生**（非必现）
- 开启思考模式（`enable_thinking=true`）时更常见
- 与模型、后端无关（SGLang/vLLM 均出现过）

---

### 3.1 问题代码路径

`ReActAgent.buildAgentStream()`（agentscope-core:2.0.0）使用 `Flux.create` 桥接内部 Mono 管道与外部事件流：

```java
Flux.create(sink -> {
    AgentStartEvent start = new AgentStartEvent(...);
    sink.next(start);
    
    Mono<Msg> mono = runLifecycle(input, middlewareChain);
    mono.contextWrite(ctx -> ctx.put("eventSink", sink))
        .doFinally(sig -> {
            sink.next(new AgentEndEvent(agentName));
            sink.complete();  // ← 此处偶尔延迟 30s
        })
        .subscribe(
            msg -> sink.next(new AgentResultEvent(msg)),
            err -> sink.error(err)
        );
    sink.onCancel(subscription);
}, OverflowStrategy.BUFFER);  // ★ BUFFER 模式可能加剧延迟
```

### 3.2 延迟机制

1. `runLifecycle` 的 Mono 在 `boundedElastic-4` 线程完成
2. `doFinally` 回调在 `boundedElastic-4` 上触发
3. `sink.complete()` 调用 `FluxSink.complete()`，但信号通过 `BUFFER` 策略传播到下游
4. 下游引用了 `concatMap` 等操作符，这些操作符可能在不同的 `boundedElastic` 线程上处理
5. **信号从完成线程传播到下游消费者线程的调度延迟达到 30 秒**（当线程池繁忙或有其他背压条件时）

### 3.3 为什么 `BUFFER`

`OverflowStrategy.BUFFER` 意味着未消费的事件在内存中缓冲。当下游消费慢（如 SSE 序列化 + HTTP 写回），缓冲累积，`sink.complete()` 信号排在缓冲队列尾部。

---

### 方案 A（推荐）：`doFinally` 中用 `tryEmitComplete`

```java
.doFinally(sig -> {
    sink.next(new AgentEndEvent(agentName));
    EmitResult result = sink.tryEmitComplete();
    if (result.isFailure()) {
        log.warn("sink.tryEmitComplete failed: {}", result);
    }
})
```

`tryEmitComplete()` 是非阻塞 API，失败时不阻塞调用线程，由 Reactor 内部重试机制兜底。

### 方案 B：`Flux.push` + `LATEST`

```java
Flux.push(sink -> { ... }, OverflowStrategy.LATEST)
```

`Flux.push` 是异步信号发射器，`LATEST` 策略丢弃未被消费的旧事件——相比 `BUFFER` 不会累积积压队列。但这可能丢失事件，需要权衡。

### 方案 C：分离 `sink.complete()` 到独立调度

```java
.doFinally(sig -> {
    sink.next(new AgentEndEvent(agentName));
    // 在无背压的调度器上完成
    Schedulers.single().schedule(() -> sink.complete());
})
```
---

## 5. 复现方法

1. 用 `HarnessAgent.streamEvents()` 进行流式对话
2. 开启思考模式（`enable_thinking=true`）
3. 使用 qwen3.6-27b 这样的中等规模模型（输出 token 较多）
4. 连续发送多条长回复请求
5. 观察 `POST_REASONING` 和 `POST_CALL` 之间的时间差

**注意**：偶发，可能需要多轮对话才触发。

---

### LINKED ISSUE #2478: [Bug]: Streaming retry re-subscribes the model and duplicates already-delivered chunks
## Summary

`ModelUtils.applyTimeoutAndRetry` attaches `retryWhen` directly to the **streaming** `Flux`. `retryWhen` re-subscribes its upstream, so when a retryable error arrives **mid-stream** the model regenerates the entire response and downstream consumers receive the already-delivered chunks a second time.

This is a correctness problem, not just a latency one: the duplicated chunks have already been handed to middlewares, written to memory, and rendered in the UI.

## Affected version

- `2.0.0` (and current `main`, verified below)

## Code path

`agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java`

```java
Retry retrySpec = Retry.backoff(maxAttempts - 1, initialBackoff)
        .maxBackoff(maxBackoff)
        .jitter(0.5)
        .filter(retryOn)          // no notion of "already emitted"
        .doBeforeRetry(...);

responseFlux = responseFlux.retryWhen(retrySpec);   // re-subscribes the stream
```

There is no guard for whether the stream has already emitted anything. `ExecutionConfig.RETRYABLE_ERRORS` treats `IOException` and `TimeoutException` as retryable, which is exactly what a dropped streaming connection produces.

With `MODEL_DEFAULTS` (`maxAttempts=3`) the same content can be delivered **three times**.

This applies to the streaming path of every provider routed through the helper — `OpenAIChatModel.doStream`, `DashScopeChatModel`, `AnthropicChatModel`, `OllamaChatModel`.

## Reproduction (no live model needed)

Upstream emits two chunks, then fails with a retryable `IOException`:

```java
Flux<ChatResponse> upstream = Flux.defer(() -> {
    subscriptions.incrementAndGet();
    return Flux.concat(
        Flux.just(chunk("Hello"), chunk(" world")),
        Flux.error(new IOException("connection reset by peer")));
});

ModelUtils.applyTimeoutAndRetry(upstream, options, options, "m", "p")
        .map(this::textOf)
        .onErrorResume(e -> Flux.empty())
        .collectList()
        .block();
```

Observed on current `main`:

```
upstream subscriptions = 3
chunks delivered downstream = [Hello,  world, Hello,  world, Hello,  world]
```

Expected: `[Hello,  world]`, upstream subscribed once.

## Impact

1. **Duplicated content** downstream — middlewares, memory writes and UI all see the response twice or three times.
2. **Latency** — each retry costs a full regeneration. For long or thinking-mode responses that is tens of seconds during which no event reaches the client.

Point 2 may explain #2279 (30 s of silence before completion). That report attributes the delay to `Flux.create` / `OverflowStrategy.BUFFER` signal propagation, but `doFinally` is attached to the lifecycle `Mono` — it cannot fire until the `Mono` terminates, and a retry keeps the `Mono` alive across backoff plus a full regeneration. The timeline in that issue (middlewares `onComplete` at `00:15:33`, `POST_CALL` at `00:16:03`) is consistent with "the Mono finished 30 s later", not "the completion signal took 30 s to propagate".

Separately, the fix suggested there — `sink.tryEmitComplete()` — cannot compile: `FluxSink` has no `tryEmit*` method in any Reactor 3.x release (verified with `javap` against 3.1.7.RELEASE, 3.4.16, 3.6.18, 3.7.6); `tryEmitComplete()` belongs to `Sinks.Many`.

## Suggested fix

Stop retrying once anything has been emitted, while keeping retries for failures **before** the first chunk — connection setup failures and HTTP 429 ahead of the first token, where retrying is both safe and valuable:

```java
AtomicBoolean emitted = new AtomicBoolean(false);
final Predicate<Throwable> retryableError = retryOn;

Retry retrySpec = Retry.backoff(maxAttempts - 1, initialBackoff)
        .maxBackoff(maxBackoff)
        .jitter(0.5)
        .filter(error -> !emitted.get() && retryableError.test(error))
        .doBeforeRetry(...);

responseFlux = responseFlux.doOnNext(r -> emitted.set(true)).retryWhen(retrySpec);
```

PR incoming with this change plus regression coverage: mid-stream retryable error, failure before the first chunk, non-retryable mid-stream error, and a clean stream. Only the first fails without the change — the other three pass both ways, pinning that retries are not over-disabled.

gxgeek-n:
Fixes #2478.

## Problem

`ModelUtils.applyTimeoutAndRetry` attaches `retryWhen` directly to the streaming `Flux`. `retryWhen` re-subscribes its upstream, so a retryable error arriving **mid-stream** makes the model regenerate the whole response — and the chunks already handed to middlewares, memory and the UI get delivered again.

`ExecutionConfig.RETRYABLE_ERRORS` classifies `IOException` and `TimeoutException` as retryable, which is exactly what a dropped streaming connection looks like. With `MODEL_DEFAULTS` (`maxAttempts=3`) the same content can land downstream three times.

Affects the streaming path of every provider routed through this helper: `OpenAIChatModel`, `DashScopeChatModel`, `AnthropicChatModel`, `OllamaChatModel`.

## Change

Track whether the stream has emitted anything and fold that into the retry filter:

```java
AtomicBoolean emitted = new AtomicBoolean(false);
final Predicate<Throwable> retryableError = retryOn;

.filter(error -> !emitted.get() && retryableError.test(error))
...
responseFlux = responseFlux.doOnNext(r -> emitted.set(true)).retryWhen(retrySpec);
```

`doOnNext` sits upstream of `retryWhen` so every emitted element flips the flag. Failures **before** the first chunk still retry — connection setup failures and HTTP 429 ahead of the first token — which is where retrying is safe and valuable. Non-streaming calls emit a single element and are unaffected when they fail before producing it.

14 lines changed in production code.

## Verification

`ModelUtilsStreamingRetryTest`, four cases:

| case | without fix | with fix |
|---|---|---|
| mid-stream retryable error | **FAIL** — `[Hello,  world]` expected, got `[Hello,  world, Hello,  world, Hello,  world]` (upstream subscribed 3x) | PASS |
| failure before first chunk still retries | PASS | PASS |
| non-retryable mid-stream error | PASS | PASS |
| clean stream unaffected | PASS | PASS |

Only the first case fails without the change; the other three pass both ways, which pins that retries are **not** over-disabled — the fix is narrow rather than "turn retry off for streaming".

Reproduced with no live model: upstream emits two chunks then throws `IOException`.

## Relationship to #2279

This likely explains the 30 s of silence reported there. `doFinally` is attached to the lifecycle `Mono`, so it cannot fire until that `Mono` terminates; a retry keeps it alive across backoff **plus a full regeneration**. The timeline in that report — middlewares `onComplete` at `00:15:33`, `POST_CALL` at `00:16:03` — matches "the Mono finished 30 s later" rather than "the completion signal took 30 s to propagate".

Worth noting the fix proposed in #2279, `sink.tryEmitComplete()`, cannot compile: `FluxSink` has no `tryEmit*` method in any Reactor 3.x release (checked with `javap` against 3.1.7.RELEASE / 3.4.16 / 3.6.18 / 3.7.6) — `tryEmitComplete()` belongs to `Sinks.Many`. I have not closed #2279 since I cannot reproduce its exact environment; if maintainers confirm this is the same root cause it can be linked or closed as a duplicate.

`spotless:apply` has been run.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
@@ -111,11 +111,20 @@ public final class ModelUtils {
                     retryOn = error -> true; // retry all errors by default
                 }
 
+                // A retry re-subscribes the upstream. For a streaming response that has already
+                // delivered chunks downstream, re-subscribing makes the model regenerate the whole
+                // response, so downstream consumers (middlewares, memory, UI) observe the content
+                // twice. Only retry while nothing has been emitted yet — that still covers the
+                // valuable cases (connection setup failures, 429 before the first token).
+                java.util.concurrent.atomic.AtomicBoolean emitted =
+                        new java.util.concurrent.atomic.AtomicBoolean(false);
+                final Predicate<Throwable> retryableError = retryOn;
+
                 Retry retrySpec =
                         Retry.backoff(maxAttempts - 1, initialBackoff)
                                 .maxBackoff(maxBackoff)
                                 .jitter(0.5)
-                                .filter(retryOn)
+                                .filter(error -> !emitted.get() && retryableError.test(error))
                                 .doBeforeRetry(
                                         signal ->
                                                 LOG.warn(
@@ -126,7 +135,8 @@ public final class ModelUtils {
                                                         signal.failure().getMessage(),
                                                         signal.failure()));
 
-                responseFlux = responseFlux.retryWhen(retrySpec);
+                responseFlux =
+                        responseFlux.doOnNext(response -> emitted.set(true)).retryWhen(retrySpec);
                 LOG.debug(
                         "Applied retry config: maxAttempts={}, initialBackoff={} for model: {}",
                         maxAttempts,
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
@@ -116,27 +116,48 @@ public final class ModelUtils {
                 // response, so downstream consumers (middlewares, memory, UI) observe the content
                 // twice. Only retry while nothing has been emitted yet — that still covers the
                 // valuable cases (connection setup failures, 429 before the first token).
-                java.util.concurrent.atomic.AtomicBoolean emitted =
-                        new java.util.concurrent.atomic.AtomicBoolean(false);
+                //
+                // The flag lives inside Flux.defer so it is per-subscription: a Flux may be
+                // subscribed more than once, and a flag shared across subscriptions would let a
+                // later subscriber inherit "already emitted" from an earlier one and skip retries
+                // it should have performed.
                 final Predicate<Throwable> retryableError = retryOn;
-
-                Retry retrySpec =
-                        Retry.backoff(maxAttempts - 1, initialBackoff)
-                                .maxBackoff(maxBackoff)
-                                .jitter(0.5)
-                                .filter(error -> !emitted.get() && retryableError.test(error))
-                                .doBeforeRetry(
-                                        signal ->
-                                                LOG.warn(
-                                                        "Retrying model request (attempt {}/{}) due"
-                                                                + " to: {}",
-                                                        signal.totalRetriesInARow() + 1,
-                                                        maxAttempts - 1,
-                                                        signal.failure().getMessage(),
-                                                        signal.failure()));
+                final Duration retryInitialBackoff = initialBackoff;
+                final Duration retryMaxBackoff = maxBackoff;
+                final int retryMaxAttempts = maxAttempts;
+                final Flux<ChatResponse> retrySource = responseFlux;
 
                 responseFlux =
-                        responseFlux.doOnNext(response -> emitted.set(true)).retryWhen(retrySpec);
+                        Flux.defer(
+                                () -> {
+                                    java.util.concurrent.atomic.AtomicBoolean emitted =
+                                            new java.util.concurrent.atomic.AtomicBoolean(false);
+                                    Retry retrySpec =
+                                            Retry.backoff(retryMaxAttempts - 1, retryInitialBackoff)
+                                                    .maxBackoff(retryMaxBackoff)
+                                                    .jitter(0.5)
+                                                    .filter(
+                                                            error ->
+                                                                    !emitted.get()
+                                                                            && retryableError.test(
+                                                                                    error))
+                                                    .doBeforeRetry(
+                                                            signal ->
+                                                                    LOG.warn(
+                                                                            "Retrying model request"
+                                                                                + " (attempt {}/{})"
+                                                                                + " due to: {}",
+                                                                            signal
+                                                                                            .totalRetriesInARow()
+                                                                                    + 1,
+                                                                            retryMaxAttempts - 1,
+                                                                            signal.failure()
+                                                                                    .getMessage(),
+                                                                            signal.failure()));
+                                    return retrySource
+                                            .doOnNext(response -> emitted.set(true))
+                                            .retryWhen(retrySpec);
+                                });
                 LOG.debug(
                         "Applied retry config: maxAttempts={}, initialBackoff={} for model: {}",
                         maxAttempts,
--- a/agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsStreamingRetryTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsStreamingRetryTest.java
@@ -180,4 +180,39 @@ class ModelUtilsStreamingRetryTest {
         assertEquals(List.of("a", "b", "c"), delivered);
         assertEquals(1, subscriptions.get());
     }
+
+    @Test
+    @DisplayName("emitted state is per-subscription, not shared across subscribers")
+    void emittedFlagIsPerSubscription() {
+        // Every subscription emits one chunk and then fails with a retryable error. The first
+        // subscription therefore ends with "something was emitted". If that state were shared
+        // across subscriptions (allocated once outside Flux.defer), the second subscriber would
+        // inherit it and behave differently from the first.
+        Flux<ChatResponse> upstream =
+                Flux.concat(
+                        Flux.just(chunk("chunk")),
+                        Flux.error(new IOException("connection reset by peer")));
+
+        GenerateOptions options = retryOptions();
+        Flux<ChatResponse> wrapped =
+                ModelUtils.applyTimeoutAndRetry(upstream, options, options, "test-model", "test");
+
+        List<String> first =
+                wrapped.map(ModelUtilsStreamingRetryTest::textOf)
+                        .onErrorResume(error -> Flux.empty())
+                        .collectList()
+                        .block(BLOCK_TIMEOUT);
+        List<String> second =
+                wrapped.map(ModelUtilsStreamingRetryTest::textOf)
+                        .onErrorResume(error -> Flux.empty())
+                        .collectList()
+                        .block(BLOCK_TIMEOUT);
+
+        assertEquals(List.of("chunk"), first, "first subscription should deliver one chunk");
+        assertEquals(
+                first,
+                second,
+                "a second subscription must behave identically — retry state leaking across"
+                        + " subscribers would make it diverge from the first");
+    }
 }
--- a/agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsStreamingRetryTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsStreamingRetryTest.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.model;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+import io.agentscope.core.message.ContentBlock;
+import io.agentscope.core.message.TextBlock;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Flux;
+
+/**
+ * Retry semantics for streaming model responses in {@link ModelUtils#applyTimeoutAndRetry}.
+ *
+ * <p>{@code retryWhen} re-subscribes its upstream. For a streaming response that has already handed
+ * chunks to downstream consumers (middlewares, memory, UI), re-subscribing makes the model
+ * regenerate the whole response, so the same content is delivered twice — a correctness problem, not
+ * merely a latency one. It also inflates end-to-end latency by however long a full regeneration
+ * takes, which is significant for long/thinking-mode responses.
+ *
+ * <p>Retrying is still valuable <em>before</em> the first chunk (connection setup failures, HTTP 429
+ * ahead of the first token), so the guard is "stop retrying once anything has been emitted" rather
+ * than "never retry".
+ */
+@DisplayName("ModelUtils streaming retry semantics")
+class ModelUtilsStreamingRetryTest {
+
+    private static final Duration BLOCK_TIMEOUT = Duration.ofSeconds(5);
+
+    private static ChatResponse chunk(String text) {
+        return ChatResponse.builder()
+                .content(List.<ContentBlock>of(TextBlock.builder().text(text).build()))
+                .build();
+    }
+
+    private static String textOf(ChatResponse response) {
+        return response.getContent().stream()
+                .filter(TextBlock.class::isInstance)
+                .map(TextBlock.class::cast)
+                .map(TextBlock::getText)
+                .findFirst()
+                .orElse("");
+    }
+
+    /** Mirrors the retry shape of {@code MODEL_DEFAULTS} with a 1 ms backoff to keep tests fast. */
+    private static GenerateOptions retryOptions() {
+        return GenerateOptions.builder()
+                .executionConfig(
+                        ExecutionConfig.builder()
+                                .maxAttempts(3)
+                                .initialBackoff(Duration.ofMillis(1))
+                                .maxBackoff(Duration.ofMillis(2))
+                                .backoffMultiplier(2.0)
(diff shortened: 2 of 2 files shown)


oss-maintainer (commented):
## Review — PR #2479

### Streaming Retry Fix (commit 2) — ✅ Good

The core fix is clean and well-targeted:

- **`AtomicBoolean emitted`** correctly tracks whether any chunk has been delivered
- **`.filter(error -> !emitted.get() && retryableError.test(error))`** narrows retry to pre-first-chunk failures only
- **`.doOnNext(r -> emitted.set(true))`** placed upstream of `retryWhen` — correct ordering
- **14 lines in production code** — minimal surface area
- **4 test cases** in `ModelUtilsStreamingRetryTest` cover the bug scenario + regression cases (pre-first-chunk retry still works, non-retryable errors pass through, clean stream unaffected)

The fix preserves the valuable retry cases (connection setup, HTTP 429 before first token) while preventing the duplication bug. Well done.

### Concerns

**1. PR mixes two independent fixes**

This PR contains two commits from different authors:
- Commit 1 (waterWang): `fix(#2475): store userId/sessionId directly on CallExecution` — already submitted as standalone PR #2476
- Commit 2 (gxgeek-n): `fix(model): stop retrying a streaming response once chunks were delivered` — the actual #2478 fix

Mixing them creates merge/review complexity. If #2476 is merged first, this PR will have a conflict on `ReActAgent.java`. Consider rebasing this PR onto `main` after #2476 lands, or splitting into a standalone PR for just the streaming fix.

**2. CI failures — Spotless formatting**

Both Linux and Windows builds fail with Spotless violations in `ReActAgent.java`:
- Extra blank lines before `initialStateLoad` Javadoc (line 352-353)
- `CallExecution` constructor parameters need one-per-line formatting

Fix with:
```bash
mvn spotless:apply -pl agentscope-core
```

### Summary

The streaming retry fix itself is solid. Address the Spotless formatting and consider the PR composition concern.

gxgeek-n:
Apologies — I force-pushed to correct a mistake on my side.

The branch was accidentally cut from a local branch based on #2476's head rather than from `main`, so this PR initially carried **three** commits:

- `0bdb6bb9` — **@fanjiacheng26's fix from #2476** (not my work, it should never have appeared here)
- `c664ed09` — my regression test for #2475 (belongs with #2476; already offered there for cherry-pick)
- `ba74fb12` — the actual streaming-retry fix

That explains @copilot-pull-request-reviewer describing this PR as covering "two areas" and @oss-maintainer referring to the streaming fix as "commit 2".

The branch is now rebuilt directly on `main` and contains **one commit, two files**:

```
178a9820 fix(model): stop retrying a streaming response once chunks were delivered

agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java          +12 -2
agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsStreamingRetryTest.java  +183
```

`ReActAgent.java` is no longer touched by this PR at all. Credit for the #2475 slot-key fix stays entirely with #2476.

Re-verified after the rebuild: `ModelUtilsStreamingRetryTest` 4/4 pass, and the mid-stream case still fails without the `ModelUtils` change (expected `[Hello,  world]`, got it three times over).

Thanks @oss-maintainer for the review of the streaming portion — the assessment of the approach matches the intent: `emitted` is checked inside the retry filter and `doOnNext` sits upstream of `retryWhen` so ordering is guaranteed. Happy to adjust anything further.

gxgeek-n:
Correction to my previous comment: the author of #2476 / commit `0bdb6bb9` is **@waterWang**, not the name I wrote. Apologies for the misattribution — @oss-maintainer had it right. Credit for the #2475 slot-key fix belongs entirely to @waterWang and #2476.

Responding to both concerns in the review — both are resolved by the force-push:

**1. PR mixed two independent fixes** — fixed. The branch is rebuilt on `main` and now carries a single commit touching two files:

```
178a9820 fix(model): stop retrying a streaming response once chunks were delivered

agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java                     +12 -2
agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsStreamingRetryTest.java   +183
```

`ReActAgent.java` is no longer in the diff, so the conflict-after-#2476-lands concern goes away too.

**2. Spotless failures in `ReActAgent.java`** — those came entirely from the mis-included commit. With `ReActAgent.java` out of the diff there is nothing left to format there. `mvn spotless:apply -pl agentscope-core` had already been run against my own files; `Check License` and `Check Module Sync` are green on the rebuilt branch and the two `build` jobs are re-running now.

Thanks for the precise review — the composition problem was real and entirely my error in cutting the branch.

gxgeek-n:
Thanks @Copilot — that catch is correct and I've fixed it.

The `emitted` flag and the `Retry` instance were created once per `applyTimeoutAndRetry` call and captured by the returned `Flux`. Since a `Flux` can be subscribed more than once, a later subscriber could inherit `emitted=true` from an earlier subscription and skip retries it should have performed. Exactly the shared-mutable-state-across-subscribers problem you described.

Fixed by moving the retry wiring inside `Flux.defer`, so both the flag and the `Retry` spec are allocated per subscription:

```java
responseFlux = Flux.defer(() -> {
    AtomicBoolean emitted = new AtomicBoolean(false);
    Retry retrySpec = Retry.backoff(retryMaxAttempts - 1, retryInitialBackoff)
            .maxBackoff(retryMaxBackoff)
            .jitter(0.5)
            .filter(error -> !emitted.get() && retryableError.test(error))
            .doBeforeRetry(...);
    return retrySource.doOnNext(response -> emitted.set(true)).retryWhen(retrySpec);
});
```

Added `emittedFlagIsPerSubscription` to pin it: the same wrapped `Flux` is subscribed twice, each subscription emitting one chunk before a retryable failure. With state shared across subscriptions the second subscription diverges from the first; now they match.

`ModelUtilsStreamingRetryTest` is 5/5 green, and the original mid-stream case still fails without the `ModelUtils` change.

Your second comment (about this PR also modifying `ReActAgent` for #2475) was accurate at the time — that was my branch-cutting error and it has since been force-pushed away. This PR now carries only the two streaming-retry commits and touches only `ModelUtils.java` plus its test.

gxgeek-n:
Addressed the Codecov partial-branch warning (96.15%, `0 Missing and 1 partial`).

The partial was on the retry filter's compound condition:

```java
.filter(error -> !emitted.get() && retryableError.test(error))
```

The existing non-retryable case emits a chunk before failing, so `!emitted.get()` short-circuits and `retryableError.test(...)` returning `false` was never reached. Added `nonRetryableErrorBeforeFirstChunk_noRetry` — fails with a non-retryable error before emitting anything, so both operands are now exercised in both directions.

`ModelUtilsStreamingRetryTest` is 6/6 green.

This PR now stands at: 3 commits, 2 files (`ModelUtils.java` +12/-2 and its test). Since the earlier force-push reset the review state to `REVIEW_REQUIRED`, a fresh look would be appreciated whenever convenient — the substantive changes since the last review are the `Flux.defer` per-subscription fix and this branch-coverage test.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR addresses a correctness problem in `ModelUtils.applyTimeoutAndRetry`: Reactor's `retryWhen` re-subscribes the upstream, which for a streaming LLM response that has already delivered chunks causes the model to regenerate the entire response. The fix wraps the retry logic inside `Flux.defer` with a per-subscription `AtomicBoolean emitted` flag: retries are allowed only before the first chunk is emitted. The implementation is correct and the test suite covers the full branch matrix. **No critical bugs found.**

AgentScopeJavaBot (commented):
🤖 AI Review

This PR addresses a correctness problem in `ModelUtils.applyTimeoutAndRetry`: Reactor's `retryWhen` re-subscribes the upstream, which for a streaming LLM response that has already delivered chunks causes the model to regenerate the entire response. The fix wraps the retry logic inside `Flux.defer` with a per-subscription `AtomicBoolean emitted` flag: retries are allowed only before the first chunk is emitted. The implementation is correct and the test suite covers the full branch matrix. **No critical bugs found.**

AgentScopeJavaBot (commented):
🤖 AI Review

This PR addresses a correctness problem in `ModelUtils.applyTimeoutAndRetry`: Reactor's `retryWhen` re-subscribes the upstream, which for a streaming LLM response that has already delivered chunks causes the model to regenerate the entire response. The fix wraps the retry logic inside `Flux.defer` with a per-subscription `AtomicBoolean emitted` flag: retries are allowed only before the first chunk is emitted. The implementation is correct and the test suite covers the full branch matrix. **No critical bugs found.**

agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
```diff
@@ -111,22 +111,53 @@ public static Flux<ChatResponse> applyTimeoutAndRetry(
...
+                final Duration retryMaxBackoff = maxBackoff;
+                final int retryMaxAttempts = maxAttempts;
+                final Flux<ChatResponse> retrySource = responseFlux;
 
-                responseFlux = responseFlux.retryWhen(retrySpec);
+                responseFlux =
+                        Flux.defer(
+                                () -> {
+                                    java.util.concurrent.atomic.AtomicBoolean emitted =
+                                            new java.util.concurrent.atomic.AtomicBoolean(false);
```
oss-maintainer, line 134: **[Info]** Minor style nit, non-blocking: `java.util.concurrent.atomic.AtomicBoolean` is spelled out inline here, while the rest of `agentscope-core` imports it (e.g. `JdkHttpTransport`, `JdkWebSocketConnection`). A plain `import java.util.concurrent.atomic.AtomicBoolean;` would keep this file consistent with the module convention and shorten the two lines inside the `defer` lambda.

oss-maintainer (approved):
### Summary
Re-review after the branch was force-reset to a single-purpose change: two new commits on top of the previously reviewed head, both of which I consider correct — this is now good to merge from my side.

- `0004c04e8` moves the `emitted` guard **inside** `Flux.defer`, so the flag is allocated per subscription instead of per assembly. This is the right fix: `retryWhen` sits on `retrySource.doOnNext(...)` inside the same `defer` lambda, so retries within one subscription still share the guard (the anti-duplication semantics are unchanged), while a second, independent subscriber of the same `Flux` gets a fresh guard rather than inheriting `emitted=true` and silently losing its retry budget.
- `668a7fa85` closes the remaining uncovered branch (non-retryable error before the first chunk) and `emittedFlagIsPerSubscription` pins the new behaviour as a regression test.
- My earlier two concerns are resolved: the PR no longer carries the unrelated `ReActAgent`/`CallExecution` change, and Spotless/build (ubuntu + windows), Check License, Check Module Sync and codecov/patch are all green on `668a7fa85`, with `license/cla` signed.

### Findings
- **[Info]** `ModelUtils.java:134` — inline `java.util.concurrent.atomic.AtomicBoolean` FQN instead of an import (style only, see inline comment).
- **[Info]** Behaviour note, no change requested: because the guard trips on the *first* `onNext`, a non-streaming caller whose formatter emits a value (e.g. a usage-only final chunk) and then signals an error will no longer be retried. That is exactly the trade-off the fix asks for, so I am flagging it only so it is an explicit decision rather than an accident — the current tests all use the multi-chunk streaming shape.

### Suggestions
None blocking. A one-line import is the only optional polish.

---

PR #2481, opened 2026-07-29
fix(harness): subagent event stream never closes when the parent is cancelled

### LINKED ISSUE #2480: [Bug]: Parent cancel leaves a sync-spawned subagent's event stream open (AgentStartEvent with no AgentEndEvent)
## Summary

`AgentSpawnTool` emits an `AgentStartEvent` unconditionally before subscribing to a sync-spawned child, then emits the matching `AgentEndEvent` from a `doOnTerminate` callback. Reactor's `doOnTerminate` fires on `onComplete` / `onError` but **not on cancel**, so cancelling the parent leaves the child's event stream open: the consumer received a start with no end and keeps rendering that subagent as running.

## Affected version

`2.0.0` and current `main`.

## Code path

`agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java`, Path 1 (`streamEvents()` — the non-deprecated path):

```java
parentEmitter.emit(
        new AgentStartEvent(spawned.sessionId(), null, spawned.agentId())
                .withSource(sourcePath));           // always emitted

return manager.invokeAgent(agent, sessionId, userId, prompt, parentCtx)
        .contextWrite(...)
        .doOnTerminate(                             // ← skips cancel
                () -> parentEmitter.emit(
                        new AgentEndEvent(null).withSource(sourcePath)));
```

Triggered by any upstream cancel: user "Stop", an outer `Mono.timeout()`, or Reactor `dispose()`.

Reactor semantics, verified directly against reactor-core 3.7.6:

| terminal signal | `doOnTerminate` | `doFinally` |
|---|---|---|
| complete | 1 | 1 |
| error | 1 | 1 |
| **cancel** | **0** | 1 |

Path 2 (deprecated `SubagentEventBus`) relies on the child's own `isLast()` events and Path 3 is non-streaming, so neither is affected. `doOnTerminate` appears exactly once in this file.

Worth noting the same file **already uses `doFinally`** for the timeout-promotion path, and the comment there explicitly reasons about `doFinally(CANCEL)` — so this looks like an oversight rather than intent.

## Reproduction

Mock a `HarnessAgent` whose `call()` returns `Mono.never()`, spawn it through `agentSpawn`, record events via an `AgentEventEmitter` in the Reactor Context, then `dispose()` the subscription:

```
AgentStartEvent count = 1
AgentEndEvent   count = 0     ← expected 1
```

## Impact

Consumers that pair start/end events (UI progress indicators, tracing, any accounting of in-flight subagents) never see the child close. In a long-lived session every cancelled spawn leaks one permanently "running" subagent in the consumer's view.

## Suggested fix

Use `doFinally` so all three terminal signals close the stream:

```java
.doFinally(signal -> parentEmitter.emit(
        new AgentEndEvent(null).withSource(sourcePath)));
```

## Relationship to #2408 / #2412

Distinct concern. #2408 / #2412 fix whether the child execution actually **stops** (`interruptAgent` being a no-op for `HarnessAgent`, plus the child `RuntimeContext` mismatch) and touch `interruptAgent` — I checked #2412's diff and it does not reference `doOnTerminate`, `AgentStartEvent` or `AgentEndEvent`. Even with the child correctly interrupted, the **event stream** still has to close. The two fixes are complementary and touch different lines of the same file.

PR incoming with the change plus regression coverage (cancel path fails without it; normal completion passes both ways, pinning that no duplicate events appear on the happy path). Full `agentscope-harness` suite: 693 tests, 0 failures.

gxgeek-n:
Fixes #2480.

## Problem

`AgentSpawnTool` emits `AgentStartEvent` unconditionally before subscribing to a sync-spawned child, then emits the matching `AgentEndEvent` from `doOnTerminate`. `doOnTerminate` fires on `onComplete` / `onError` but **not on cancel**, so a parent cancel (user "Stop", outer `timeout()`, Reactor `dispose()`) leaves the child's event stream open — consumers see a start with no end and keep rendering the subagent as running.

Verified Reactor semantics before touching anything (reactor-core 3.7.6):

| terminal signal | `doOnTerminate` | `doFinally` |
|---|---|---|
| complete | 1 | 1 |
| error | 1 | 1 |
| **cancel** | **0** | 1 |

## Change

One call swapped, `doOnTerminate` → `doFinally`, so all three terminal signals close the stream.

Two things suggest this was an oversight rather than intent: `doOnTerminate` appears exactly once in the file, and the timeout-promotion path a few lines below **already** uses `doFinally` with a comment explicitly reasoning about `doFinally(CANCEL)`.

Path 2 (deprecated `SubagentEventBus`) pairs events via the child's own `isLast()` and Path 3 is non-streaming, so neither needed changing.

## Verification

`AgentSpawnToolCancelEndEventTest`:

| case | without fix | with fix |
|---|---|---|
| parent cancel emits `AgentEndEvent` | **FAIL** — expected 1, got 0 | PASS |
| normal completion emits exactly one start + one end | PASS | PASS |

The second case passing both ways is the point: it pins that the fix does not start emitting duplicate end events on the happy path.

Full `agentscope-harness` suite: **693 tests, 0 failures, 3 skipped**. Worth running given the change makes the cancel path emit an event it previously did not.

## Relationship to #2408 / #2412

Complementary, not overlapping. #2412 fixes whether the child execution actually **stops** (`interruptAgent` being a no-op for `HarnessAgent`, plus the child `RuntimeContext` mismatch); this fixes whether the emitted **event stream closes**. I checked #2412's diff — it does not reference `doOnTerminate`, `AgentStartEvent` or `AgentEndEvent`. Both touch `AgentSpawnTool.java` but different lines, so if #2412 lands first I am happy to rebase.

`spotless:apply` has been run.

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java
@@ -671,8 +671,11 @@ public class AgentSpawnTool {
                                                 c.put(
                                                         AgentEventEmitter.FORWARDING_CONTEXT_KEY,
                                                         taggedEmitter))
-                                .doOnTerminate(
-                                        () ->
+                                // doFinally, not doOnTerminate: the latter skips cancel, so a
+                                // parent cancel would leave the AgentStartEvent above unmatched
+                                // and consumers would render this subagent as running forever.
+                                .doFinally(
+                                        signal ->
                                                 parentEmitter.emit(
                                                         new AgentEndEvent(null)
                                                                 .withSource(sourcePath)));
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/AgentSpawnToolCancelEndEventTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/AgentSpawnToolCancelEndEventTest.java
@@ -1,0 +1,187 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent.tool;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+
+import io.agentscope.core.ReActAgent;
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.core.event.AgentEndEvent;
+import io.agentscope.core.event.AgentEvent;
+import io.agentscope.core.event.AgentEventEmitter;
+import io.agentscope.core.event.AgentStartEvent;
+import io.agentscope.core.message.Msg;
+import io.agentscope.harness.agent.HarnessAgent;
+import io.agentscope.harness.agent.middleware.SubagentEntry;
+import io.agentscope.harness.agent.subagent.DefaultAgentManager;
+import io.agentscope.harness.agent.subagent.task.BackgroundTask;
+import io.agentscope.harness.agent.subagent.task.TaskRepository;
+import io.agentscope.harness.agent.subagent.task.TaskRunSpec;
+import io.agentscope.harness.agent.subagent.task.TaskStatus;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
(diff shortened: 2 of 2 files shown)


oss-maintainer (approved):
## Review — Approved ✅

### Summary

One-line fix: `doOnTerminate` → `doFinally` in `AgentSpawnTool#execLocalSync()` Path 1, ensuring `AgentEndEvent` is emitted on all three terminal signals (complete, error, **cancel**).

### Analysis

**Correctness:** The Reactor semantics are accurately described. `doOnTerminate` skips cancel; `doFinally` handles all three. This is the standard pattern for cleanup that must run regardless of termination reason.

**Scope:** Minimal and precise — only the affected call site is changed. Path 2 (deprecated) and Path 3 (non-streaming) are correctly left untouched.

**Comment quality:** The inline comment clearly explains *why* `doFinally` is used instead of `doOnTerminate`, preventing future regressions. Good practice.

**Test coverage:** `AgentSpawnToolCancelEndEventTest` (187 lines) covers:
- Cancel path emits `AgentEndEvent` (the bug fix)
- Happy path still emits exactly one start + one end (no duplicates)
- Uses `Mono.never()` + `dispose()` for deterministic cancel simulation

**CI:** License ✅, Module Sync ✅, builds pending (just triggered). Previous run on the same commit passed all checks.

### Verdict

Clean, well-tested fix for a real event-stream leak. LGTM.

gxgeek-n:
Thanks @Copilot — good point, fixed.

`Thread.sleep(300)` replaced with a bounded wait loop that returns as soon as the `AgentEndEvent` is observed (5 s ceiling, 20 ms interval):

```java
long deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(5);
while (emitter.count(AgentEndEvent.class) == 0 && System.nanoTime() < deadline) {
    Thread.sleep(20);
}
```

Both cases still pass, and the cancel case no longer spends a fixed 300 ms waiting. It also still fails without the `doFinally` change in `AgentSpawnTool` (expected 1, got 0), so the bounded wait did not weaken the assertion into a no-op.

jujn (approved):

PR #2491, opened 2026-07-30
fix(harness): bind sandbox to per-call RuntimeContext to prevent concurrent-call corruption

xseruer:
## AgentScope-Java Version

2.0.1-SNAPSHOT (main branch); bug also present in 2.0.0

### Background

`SandboxBackedFilesystem` and `SandboxLifecycleMiddleware` stored per-call sandbox state in **agent-level single slots** (a `volatile Sandbox` field on the filesystem proxy and an `AtomicReference<SandboxAcquireResult>` on the middleware), implicitly assuming at most one in-flight call per agent instance.

However, a single shared agent bean serving many users/conversations is a common setup, and the SDK's `serializeOnKey` only serializes calls of the *same* `(userId, sessionId)` — cross-session calls run in parallel by design. When two calls run concurrently on one agent, both slots get corrupted in three ways:

- a later call overwrote the earlier call's sandbox, silently routing its filesystem/tool operations into the **wrong sandbox** (cross-session file leakage);
- a finishing call nulled the slot while another call was mid-run, failing it with `SandboxConfigurationException: No active sandbox`;
- `releaseForCall` (`getAndSet(null)`) could stop the **other** call's sandbox session and leak its own (never released).

We hit the second failure mode in production at scale: an agent dispatching IM group messages across many conversations had a large fraction of runs fail with `No active sandbox` whenever two conversations were active at once. The only workaround was a custom `SandboxExecutionGuard` serializing *all* calls, degrading cross-session parallelism to FIFO.

### Changes

- **`SandboxLifecycleMiddleware`**: removed the agent-level `currentAcquireResult` slot. `acquireForCall(ctx)` now binds the acquired `Sandbox` and `SandboxAcquireResult` to the per-call `RuntimeContext` (`ctx.put`); `releaseForCall(ctx)` retrieves the acquire result from **its own** context, so persist/release/lease-close are strictly paired per call.
- **`SandboxBackedFilesystem`**: `requireSandbox(ctx)` resolves the sandbox from the caller's `RuntimeContext` first, so concurrent calls each see their own sandbox. The agent-level `SandboxAware` slot is **kept only as a fallback** for context-less internal paths (e.g. `RuntimeContext.empty()` maintenance writes in `SkillUsageStore` / `SkillCurator` / `SkillAuditLog`), backed by an `AtomicReference` with a new CAS-based `clearSandbox(expected)` so a finishing call never wipes a newer call's binding.
- No public API removed: `SandboxAware#setSandbox/getSandbox` behave as before for existing callers. A follow-up could thread a per-call context through the remaining context-less paths and then deprecate the fallback slot.

### How to test

- New regression test `SandboxLifecycleMiddlewarePerCallBindingTest` covers: concurrent calls resolving their own sandbox, a finishing call not disturbing (nor releasing the sandbox of) a still-running call, and idempotent release per context.
- Extended `SandboxBackedFilesystemTest` with context-priority resolution and CAS-clear semantics.
- Full `agentscope-harness` suite: **679 tests, 0 failures**.

## Checklist

- [ ]  Related documentation has been updated (e.g. links, examples, etc.) — N/A, internal concurrency fix with no doc-facing behavior change

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
@@ -29,22 +29,31 @@ import java.util.Base64;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * A {@link BaseSandboxFilesystem} that delegates execution to a live {@link Sandbox}.
  *
- * <p>Stable proxy created at agent build time; a fresh {@link Sandbox} is injected on each call
- * via the volatile {@code sandbox} field by {@link
- * io.agentscope.harness.agent.middleware.SandboxLifecycleMiddleware}.
+ * <p>Stable proxy created at agent build time. The active {@link Sandbox} for an operation is
+ * resolved <b>per call</b> from the {@link RuntimeContext} (bound by {@link
+ * io.agentscope.harness.agent.middleware.SandboxLifecycleMiddleware} via {@code
+ * ctx.put(Sandbox.class, ...)}), so concurrent calls on the same agent instance each see their
+ * own sandbox.
+ *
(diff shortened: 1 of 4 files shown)


zouyx (commented):
I have fixed the CI fail problem and merged into main.
So you can try to merge main into your branch and observe if there are still any problem in CI

PR #2495, opened 2026-07-30
fix(agui): emit AG-UI interrupt for permission-type HITL tool confirmation (#2437)

### LINKED ISSUE #2437: AG-UI adapter drops human-in-the-loop tool confirmation (consumes deprecated v1 Event stream; RequireUserConfirmEvent never bridged)

wzq-xzwj:
## What & Why

Fixes #2437.

The AG-UI adapter only surfaced **suspend-type** HITL (tools that throw `ToolSuspendException` / frontend tools, i.e. `GenerateReason.TOOL_SUSPENDED`). The **permission-type** HITL path was dropped:

- Under `DEFAULT` permission mode, a non-readonly tool makes `ReActAgent` emit `RequireUserConfirmEvent` + `RequestStopEvent(PERMISSION_ASKING)`.
- `RequireUserConfirmEvent` had no converter, so it fell through to the `RAW` fallback and **no `Interrupt` reached `RUN_FINISHED`**.
- The frontend therefore had nothing to confirm/deny, and resume was impossible.

This builds directly on the converter-registry architecture introduced in #2306, extending it to the confirmation path rather than changing its design.

## Changes

- **`PermissionConfirmEventConverter` (new)** — maps `RequireUserConfirmEvent` to one AG-UI `Interrupt` per pending `ToolUseBlock`, with reason `tool_confirmation` and `toolName` / `toolInput` / `toolContent` / `replyId` metadata. Registered in `AgentEventConverterRegistry`. The interrupt id keeps the `replyId:toolCallId` format so resume can recover the tool call id (consistent with `AgentLifecycleEventConverter`).
- **`AguiResumeCoordinator`** — tracks `tool_confirmation` interrupts alongside `tool_call`, exposing them through a new resume-interrupts runtime-context key.
- **`AguiAgentAdapter`** — forwards the resume interrupts to the message converter.
- **`AguiMessageConverter`** — for confirmation interrupts, builds a `ConfirmResult` (approved/denied) from the resume payload and reconstructs the `ToolUseBlock` with **non-null JSON content** (via `JsonUtils.resolveToolCallArgsJson`) to avoid the tool-input validation failure (`argument "content" is null`) on resume. Non-confirmation resumes keep the existing `ToolResultBlock` path.

## Tests

- Added unit tests for resolved / cancelled / explicit-`approved:false` confirmations and the non-confirmation fallback.
- Full `agentscope-extensions-agui` module suite: **435 tests, 0 failures, 0 errors**.

## Notes

- No behavior change for suspend-type HITL; confirmation is a new, separate path selected by interrupt `reason`.


wzq-xzwj:
@jujn 麻烦帮忙审核一下哈，多谢

wzq-xzwj:
@chickenlj AgentScopeJavaBot貌似不工作了😅

jujn:
> @jujn 麻烦帮忙审核一下哈，多谢

core 模块 hitl 还有些问题，先等等。所有的PR后续都会审的

jujn (approved):

PR #2515, opened 2026-08-01
fix(harness): resolve workspace-rooted absolute paths in ProjectAwareOverlay

### LINKED ISSUE #2514: harness: workspace-rooted absolute paths are misrouted when projectWritable(true)
## Summary

With `projectWritable(true)`, `ProjectAwareOverlay` is expected to route non-metadata writes to the project directory. However, when the agent supplies an **absolute** path under the workspace root (e.g. `/your/workspace/src/Foo.java`), the write is always routed to the workspace layer instead of the project.

## Root cause

`ProjectAwareOverlay.isWorkspacePath()` classifies absolute paths purely by prefix:

```java
if (Path.of(normalized).isAbsolute()) {
    return Path.of(normalized).normalize().startsWith(workspaceRoot);
}
```

Any absolute path under the workspace root is unconditionally treated as a workspace path, so the relative part (`src/Foo.java`) is never inspected against `WORKSPACE_PREFIXES`. The write therefore lands in `workspace/src/Foo.java` and `projectWritable(true)` is effectively ignored.

Note that fixing only the classification is not enough: `projectFs` runs in `ROOTED` mode with the workspace listed in its `PathPolicy`, so a workspace-rooted absolute path handed to it directly would still resolve back into the workspace. The path must be relativized before being passed to the project filesystem.

## Expected behaviour

- `write("/your/workspace/src/Foo.java", ...)` → `project/src/Foo.java`
- `write("/your/workspace/MEMORY.md", ...)` → `workspace/MEMORY.md`

## Environment

- Module: `agentscope-harness`
- Class: `io.agentscope.harness.agent.filesystem.ProjectAwareOverlay`

chcodex:
Fixes #2514

## Problem

With `projectWritable(true)`, `ProjectAwareOverlay` misroutes paths when the agent uses an **absolute** path under the workspace root (e.g. `/your/workspace/src/Foo.java`). `isWorkspacePath()` classified any workspace-rooted absolute path as workspace metadata purely by prefix, and read/search operations passed absolute paths straight to the overlay where the lower project filesystem (`SANDBOXED`) re-rooted them under the project root — so reads could not fall back to `project/src/Foo.java`.

## Changes

`ProjectAwareOverlay.java`
- `isWorkspacePath()` now relativizes workspace-rooted absolute paths and classifies the relative part with the same `WORKSPACE_PREFIXES` logic used for relative paths.
- Added `toWorkspaceRelativePath()` (renamed from `toProjectTargetPath()`), which rewrites a workspace-rooted absolute path to its workspace-relative form so the lower `SANDBOXED` project filesystem resolves it against the project root instead of re-rooting it. Without this, `ROOTED` mode would re-resolve the absolute path back into the workspace because the workspace is listed in the project FS `PathPolicy`.
- Write routing: `write`, `edit`, `delete`, `uploadFiles` use the rewritten target for the project branch.
- Read/search routing: `read`, `exists`, `downloadFiles`, `ls`, `grep`, `glob`, `move` now relativize workspace-rooted absolute paths before delegating to overlay semantics (upper workspace first, lower project fallback).

`ProjectAwareOverlayTest.java`
- Fixed `isWorkspacePath_absoluteUnderWorkspace_returnsTrue` (used a non-metadata file, which is no longer a workspace path).
- Added coverage for absolute paths: source write lands in project dir, metadata write lands in workspace, read falls back to project dir, upper read precedence, exists project fallback, ls/glob/grep find project files, and move semantics match relative paths.

## Verification

- `mvn -pl agentscope-harness -am test -Dtest=ProjectAwareOverlayTest` — 26 tests, 0 failures.
- `mvn -pl agentscope-harness test` — 709 tests, 0 failures.


chcodex:
The failing test `HarnessAgentSubagentStreamEventsTest.streamEvents_childAgentStartAndEndEmittedWithSource` is unrelated to this PR (only `ProjectAwareOverlay.java` and its test are touched here).

Verified:
- `ProjectAwareOverlayTest` passes: 19/19.
- `HarnessAgentSubagentStreamEventsTest` passes locally and on 3 repeated runs (4/4 each), so the CI failure is a flaky async timing issue (child AGENT_END event missing from the collected stream), not a regression from this change.

Re-running CI to confirm.

oss-maintainer (approved):
### Summary

---

jobmission:
遇到类似问题

Buktal (approved):
LGTM

PR #2527, opened 2026-08-01
fix(#2509): add per-session interrupt overloads to HarnessAgent

### LINKED ISSUE #2509: [Bug]: The interrupt() method of HarnessAgent does not support passing userId and sessionId
The interrupt() method of HarnessAgent does not support passing userId and sessionId, but ReActAgent does support it. The interrupt() of io. agentscope. core. agent. Agent interface should be updated to receive the InstanceContext or userId and sessionId

Image

waterWang:
## Summary

Adds , , , and  methods to  that delegate to the underlying .

## Problem

 already supports per-session interruption via  and , but  only exposed the deprecated no-arg and -only variants. External callers had no way to interrupt a specific  session through the  API.

## Changes

- Added  — delegates to 
- Added  — delegates to 
- Added  — delegates to 
- Added  — delegates to 

## Testing

This is a pure delegation change — no logic was added. The underlying  methods are already tested for correctness.

Closes #2509

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
@@ -487,6 +487,52 @@ public class HarnessAgent implements Agent, AutoCloseable {
         delegate.interrupt(msg);
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


oss-maintainer (approved):
### Summary

---

PR #2542, opened 2026-08-03
fix(core): route per-build hook/middleware mutations through local copies

xiaowu0203:
Closes #2539

### Problem

`ReActAgent.Builder.build()` mutates the (possibly shared) builder's `hooks` / `middlewares` collections in five places — `configureLongTermMemory`, `configureRAG`, `configureTodoTools`, `configureSkillBox`, and the `DynamicSkillMiddleware` branch. Reusing a shared builder (as the A2A server does with a single builder bean) races on the non-thread-safe `LinkedHashSet` / `ArrayList` (write-while-read during `new ArrayList<>(builder.hooks)` / `mws.addAll(builder.middlewares)`), failing construction with AIOOBE/NPE. Because the installed hooks don't override `equals`/`hashCode`, each build also leaks a fresh instance into the shared set — rebuilding yields agents carrying 2/3/… copies.

### Fix

`build()` now snapshots `hooks` / `middlewares` into per-build local copies, passes them to the `configure*` methods and the `ReActAgent` constructor, and never mutates the builder. `registerToolsFromHooks` runs on the copy taken before `configure*`, so configure-installed hooks never contribute `tools()` — matching first-build behaviour and staying consistent across rebuilds.

### Tests

`ReActAgentSharedBuilderRegressionTest`: a deterministic rebuild test — rebuilding a shared builder (task-list + long-term-memory + RAG configured) always yields exactly one `TaskReminderMiddleware`, one `StaticLongTermMemoryHook` and one `GenericRAGHook`. All existing `ReActAgent` unit tests pass.

Issue #979, opened 2026-03-16
请问是不是一个skill只能在一个ReActAgent里？


czhcc:
看代码是一个skill注册在一个SkillBox里，然后这个SkillBox注册在ReActAgent里

chensk0601:
不是，他可以在多个Agent下使用。

lfz757077613:
是的，这里有并发问题，所以不能并发用，toolkit也要每次新建一个。我也觉得agentscope的这种设计不好，agent/toolkit/skillbox/memory每次发起请求都要新建

<img width="1168" height="568" alt="Image" src="https://github.com/user-attachments/assets/fe3bda16-ba55-4afe-8cd5-6664cc8a432f" />

fang-tech:
> 是的，这里有并发问题，所以不能并发用，toolbox也要每次新建一个。我也觉得agentscope的这种设计不好，agent/toolkit/skillbox/memory每次发起请求都要新建

This design itself is intended to ensure that the agent is stateless. Thus, each user request loads the state from the outside, which is a beneficial design for containerization scenarios.

fang-tech:
> 看代码是一个skill注册在一个SkillBox里，然后这个SkillBox注册在ReActAgent里

The `skill class` is reusable. It is a final class that is thread-safe and can be registered simultaneously in multiple ReactAgents. However, it is recommended that a `skillBox` be placed within only one ReactAgent.

lfz757077613:
> > 看代码是一个skill注册在一个SkillBox里，然后这个SkillBox注册在ReActAgent里
> 
> The `skill class` is reusable. It is a final class that is thread-safe and can be registered simultaneously in multiple ReactAgents. However, it is recommended that a `skillBox` be placed within only one ReactAgent.

这不是明显有并发问题吗，一个skillbox怎么能复用在多个ReActAgent里呢？skillbox是线程安全的不代表可以在多个ReActAgent里复用，这里的toolkit是每个ReActAgent深copy出来的独立的新对象

<img width="1168" height="568" alt="Image" src="https://github.com/user-attachments/assets/b085c9a9-c5b3-4e00-aca6-fd43a5689b73" />
比如一个线程要执行这个apply了，另一个线程新建的ReActAgent正好执行bindToolkit把skillbox的toolkit给换了，这个apply就会因为skill-build-in-tools group没创建而报错

[...]

PR #2543, opened 2026-08-03
fix(skill): give each agent its own SkillBox copy

kratos0718:
### Problem

`SkillBox` holds a mutable reference to the `Toolkit` it serves (`SkillBox.java:146` — `bindToolkit` sets `this.toolkit`), but `Builder.configureSkillBox()` called `bindToolkit()` on the **builder's** box on every `build()`:

```java
private void configureSkillBox(Toolkit agentToolkit) {
    skillBox.bindToolkit(agentToolkit);   // mutates the shared builder field
    ...
    hooks.add(new SkillHook(skillBox));   // every agent gets the SAME instance
}
```

`SkillHook` keeps that instance (`SkillHook.java:46`, `private final SkillBox skillBox`), so building a second agent from one builder silently repoints the **first** agent's hook at the **second** agent's toolkit:

- build #1 → `skillBox.toolkit` = agent 1's toolkit copy
- build #2 → `skillBox.toolkit` = agent 2's toolkit copy

`setSkillActive()` uses that field at runtime (`SkillBox.java:343-354`, `this.toolkit.updateToolGroups(...)`), so activating a skill on agent 1 mutates agent 2's tool groups. Two concurrent `build()` calls also race on the same write — the configuration the A2A server uses, where one builder bean serves every request.

This is the remaining half of the shared-builder problem: #2542 fixes the `hooks` / `middlewares` collections; this fixes the shared `SkillBox` object. The two are independent — this PR does not touch the collections and applies cleanly before or after #2542.

### Fix

`build()` already does `this.toolkit.copy()` per agent. Mirror that for the SkillBox.

`SkillBox.copy(Toolkit)` returns a box bound to the agent's own toolkit, carrying the registered skills and configuration (`workDir`, `fileFilter`, `autoUploadSkill`, instruction, metadata exposure). `uploadDir` is deliberately left unset so it is recomputed lazily against the copy's `workDir`, matching `setWorkDir()`'s existing behaviour. Skills are shared rather than cloned — `AgentSkill` is immutable and, as noted in #979, safe to register in multiple agents.

`configureSkillBox()` now copies instead of rebinding, so the builder's box is never mutated.

This is the approach @fang-tech described in #979 ("I will simplify the setup here by using the same automatic copying method as the toolkit"), and it removes the workaround given there — *"you need to create a new skillBox for each reactagent"*.

### Tests

`SkillBoxCopyTest` — skill carry-over, registry independence in both directions after the copy, config carry-over, null rejection, and that copying leaves the source box bound to its own toolkit (the actual regression).

Verified locally with Maven 3.9.16 / JDK 17 target: `SkillBoxCopyTest` 4/4 green, and `Skill*Test`, `ReActAgent*Test`, `Hook*Test` — 57 test classes — pass with no failures or errors. `spotless:check` clean.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -4714,12 +4714,16 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
          */
         @SuppressWarnings("deprecation")
         private void configureSkillBox(Toolkit agentToolkit) {
-            skillBox.bindToolkit(agentToolkit);
-            skillBox.registerSkillLoadTool();
-            if (skillBox.isAutoUploadSkill()) {
-                skillBox.uploadSkillFiles();
+            // Copy per agent, mirroring the Toolkit.copy() at the top of build(). A SkillBox
+            // holds a mutable reference to the toolkit it serves, so binding the builder's box
+            // to this agent's toolkit repoints any previously built agent's SkillHook at the
+            // wrong toolkit, and two concurrent build() calls race on that field.
+            SkillBox agentSkillBox = skillBox.copy(agentToolkit);
+            agentSkillBox.registerSkillLoadTool();
+            if (agentSkillBox.isAutoUploadSkill()) {
+                agentSkillBox.uploadSkillFiles();
             }
-            hooks.add(new io.agentscope.core.skill.SkillHook(skillBox));
+            hooks.add(new io.agentscope.core.skill.SkillHook(agentSkillBox));
         }
 
         /**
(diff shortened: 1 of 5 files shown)


oss-maintainer (approved):
### Summary

---

kratos0718:
Hi @jujn — this one is green and ready whenever you have a moment: all checks pass, the CLA is signed, and it merges cleanly against `main`.

For context, this is the second half of the shared-builder mutation family that #2542 fixed. #2542 routed the five `hooks`/`middlewares` mutations through per-build local copies, but `configureSkillBox()` still calls `skillBox.bindToolkit(agentToolkit)` on the shared `SkillBox` builder field, and `SkillHook` holds that same instance — so a second `build()` repoints it at the second agent's toolkit, and `setSkillActive()` then mutates the wrong agent's tool groups. This PR gives each agent its own copy.

The regression test reproduces it deterministically with two sequential `build()` calls, so there are no threads involved and it should not be flaky.

Happy to rebase or adjust anything if you'd like changes.

kratos0718:
Still green and still merges cleanly against `main`, so nothing needed from my side. @chickenlj if @jujn is busy, would you be able to take a look?

PR #2545, opened 2026-08-04
fix(core,harness): replace unbounded per-path lock maps with striped locks

### LINKED ISSUE #2486: [Bug]: Per-path lock maps grow without bound — no eviction in SkillBox, WorkspaceManager and LocalFilesystem
## Summary

Three lock maps are keyed by file path and never evict entries. In an agent deployment the paths are chosen by the model rather than by a fixed application config, so the key space is effectively unbounded.

Related to #2261 (`ReActAgent.stateCache` unbounded growth), but these are different maps and one of them is `static`.

## Affected code (current `main`)

| Location | Declaration | Key | Lifetime |
|---|---|---|---|
| `agentscope-core/.../skill/SkillBox.java:57` | `private static final ConcurrentHashMap<String, Object> FILE_LOCKS` | `targetPath.toString()` (line 834) | **JVM-wide, shared by every SkillBox instance** |
| `agentscope-harness/.../workspace/WorkspaceManager.java:116` | `private final Map<String, ReentrantLock> pathLocks` | normalized / relative path (lines 376, 409, 456, 646) | instance |
| `agentscope-harness/.../filesystem/local/LocalFilesystem.java:93` | `private final ConcurrentHashMap<String, ReentrantLock> fileLocks` | lock key (line 359) | instance |

Each is populated exclusively through `computeIfAbsent`, e.g.:

```java
// SkillBox.java:834
Object lock = FILE_LOCKS.computeIfAbsent(targetPath.toString(), k -> new Object());
```

Grepping each file for `remove` / `clear` on these maps returns **0 matches** — entries are only ever added.

## Why it matters here

`FILE_LOCKS` being `static` is the notable one: every distinct path any skill has ever written to retains an entry for the lifetime of the process, across all `SkillBox` instances.

The usual justification for never evicting a per-key lock map is that the key space is bounded by application configuration. That does not hold for an agent runtime — file paths come from model output, so a long-running process accumulates one entry per distinct path the model ever produced.

## Scale, stated honestly

The leaked values are small (`Object`, `ReentrantLock`), so this is a slow leak rather than a fast one — unlike #2261, which retains `AgentState` objects. Reaching memory pressure needs a large number of distinct paths. I am reporting it as a boundedness defect rather than claiming an imminent OOM.

## Possible directions

1. **Striped locking** — a fixed number of locks selected by `key.hashCode()`, so the map never grows (the standard fix; Guava's `Striped` does exactly this). Costs occasional false sharing between unrelated paths.
2. **Remove after release** — `finally { map.remove(key); }`, but that races with another thread that is about to acquire the same lock, so it needs care.
3. **Bounded cache with eviction** — matches whatever approach #2261 settles on, keeping the codebase consistent.

I have not opened a PR because the choice among these is a design call, and option 3 in particular should probably follow whatever #2261 lands on. Happy to implement whichever direction maintainers prefer, with a test asserting the map stays bounded across many distinct paths.

liulangjietou:
## Summary

Fixes #2486.

Three lock maps keyed by file path were populated via `computeIfAbsent` and never evicted. In an agent runtime the paths come from model output, so the key space is effectively unbounded and the maps grow for the lifetime of the process — `SkillBox.FILE_LOCKS` being `static` made it JVM-wide across all instances.

Following the direction suggested in the issue (Option 1, striped locking):

- **New `io.agentscope.core.util.StripedLocks`** — a fixed-size array of `ReentrantLock`s selected by key hash (64 stripes, hash spread as in `ConcurrentHashMap`). Memory footprint is constant regardless of how many distinct paths are ever locked. No new dependency: Guava is only managed in the BOM, not actually used by core/harness, so a ~40-line utility avoids pulling it in.
- **`SkillBox.FILE_LOCKS`** — static unbounded `ConcurrentHashMap<String, Object>` → `StripedLocks`; the `synchronized` block becomes `lock()/finally unlock()` with identical reentrant semantics.
- **`WorkspaceManager.pathLocks`** (4 call sites) and **`LocalFilesystem.fileLocks`** (1 call site) — instance-level unbounded maps → `StripedLocks`, drop-in replacement for the existing `lock()/finally unlock()` pattern.

Correctness notes:
- The same key always maps to the same lock, so mutual exclusion per key is preserved.
- Two distinct keys may share a stripe (false sharing); this only affects throughput, never correctness, and is negligible for file I/O workloads.
- All call sites acquire a single lock at a time (the one nested acquisition in `WorkspaceManager.writeTaskRecord` is same-key reentrancy, which `ReentrantLock` preserves), so striping introduces no deadlock ordering.

## Tests

- New `StripedLocksTest` (5 tests): boundedness across 100k distinct keys (only 64 lock objects exist), same-key-same-lock, power-of-two rounding, argument validation, and an 8-thread mutual-exclusion test that would lose updates if two threads ever held the same key's lock simultaneously.
- Existing suites pass on current `main`: core `SkillBoxTest` + `SkillBoxToolsTest` (32/32), harness `LocalFilesystem*` / `Workspace*` / `Filesystem*` (152/152).
- `mvn spotless:check` passes.


zouyx (commented):
I have fixed the CI fail problem and merged into main.
So you can try to merge main into your branch and observe if there are still any problem in CI

liulangjietou:
Thank you @zouyx for fixing the CI issue on `main` and for the heads-up.

I've merged `main` into this branch and all checks are green now:
- Check License / Check Module Sync — pass
- build (ubuntu-latest) / build (windows-latest) — pass
- codecov/patch — pass

The branch is mergeable with no conflicts. Happy to make any adjustments if you have feedback on the approach (fixed-size lock striping over the three unbounded per-path lock maps).

oss-maintainer (approved):
### Summary

---

agentscope-core/src/main/java/io/agentscope/core/util/StripedLocks.java
```diff
@@ -0,0 +1,81 @@
...
+            this.stripes[i] = new ReentrantLock();
+        }
+    }
+
+    /**
+     * Returns the lock associated with the given key. The same key always maps to the same lock.
+     *
+     * @param key the key to select a stripe for, must not be null
+     * @return the {@link ReentrantLock} guarding the stripe this key belongs to
+     */
```
AgentScopeJavaBot, line 59: [nit] `get(null)` will throw a bare `NullPointerException` from `key.hashCode()` with no context. Consider adding an explicit guard for a clearer diagnostic:
```java
public ReentrantLock get(Object key) {
    Objects.requireNonNull(key, "key must not be null");
    return stripes[indexFor(key)];
}
```
All current call sites pass non-null keys, so this is purely defensive.

agentscope-core/src/main/java/io/agentscope/core/util/StripedLocks.java
```diff
@@ -0,0 +1,81 @@
...
+    private int indexFor(Object key) {
+        int h = key.hashCode();
+        // Spread higher bits downwards (same idea as ConcurrentHashMap) so that keys whose
+        // hash codes differ only in the upper bits still land on different stripes.
+        h ^= (h >>> 16);
+        return h & (stripes.length - 1);
+    }
+
+    private static int ceilingPowerOfTwo(int value) {
+        int highest = Integer.highestOneBit(value);
```
AgentScopeJavaBot, line 78: [nit] `ceilingPowerOfTwo` overflows silently when `value > (1 << 30)`: `highestOneBit` returns `1 << 30`, the left-shift wraps to `Integer.MIN_VALUE`, and the subsequent `new ReentrantLock[negative]` throws `NegativeArraySizeException`. Low risk in practice (callers use 64), but a one-line guard would make the class bulletproof:
```java
if (value > (1 << 30)) {
    throw new IllegalArgumentException("stripeCount too large: " + value);
}
```

agentscope-core/src/test/java/io/agentscope/core/util/StripedLocksTest.java
```diff
@@ -0,0 +1,111 @@
...
+        assertEquals(128, new StripedLocks(65).size());
+    }
+
+    @Test
+    void rejectsNonPositiveStripeCount() {
+        assertThrows(IllegalArgumentException.class, () -> new StripedLocks(0));
+        assertThrows(IllegalArgumentException.class, () -> new StripedLocks(-8));
+    }
+
+    @Test
```
AgentScopeJavaBot, line 69: [nit] The concurrency test only locks a single key (`"shared/file.txt"`). Consider adding a second phase where threads operate on *distinct* keys to verify that striping does not introduce spurious blocking between unrelated keys (i.e., throughput is not serialized). This would give stronger coverage of the striping benefit.

agentscope-core/src/main/java/io/agentscope/core/util/StripedLocks.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * A fixed-size array of striped {@link ReentrantLock}s selected by key hash.
+ *
+ * <p>Use this instead of a {@code Map<String, ReentrantLock>} populated via
+ * {@code computeIfAbsent} when the key space is unbounded (for example, file paths produced by
+ * model output in an agent runtime). A per-key lock map never evicts its entries and therefore
+ * grows without bound over the lifetime of the process, whereas the memory footprint of striped
+ * locks is constant regardless of how many distinct keys are ever locked.
+ *
+ * <p>Two distinct keys may hash to the same stripe and thus contend on the same lock. This false
```
AgentScopeJavaBot, line 29: [praise] Clean, self-contained utility with excellent Javadoc. The decision to avoid pulling in Guava for ~40 lines of code is pragmatic — the `StripedLocks` class is simpler and has zero transitive dependencies. Hash spreading and power-of-two masking follow well-established `ConcurrentHashMap` idioms.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR correctly addresses an unbounded memory growth issue (#2486) by replacing three per-path `ConcurrentHashMap`-based lock registries with a fixed-size striped lock utility. The `StripedLocks` implementation is clean, well-documented, and follows established patterns (hash spreading à la `ConcurrentHashMap`, power-of-two sizing). All call sites are migrated consistently with correct `lock()/finally unlock()` semantics and no nested-locking pitfalls. The accompanying test suite covers the key invariants: bounded memory, deterministic stripe selection, argument validation, and concurrent mutual exclusion. No correctness issues found — this is a solid, minimal fix.

agentscope-core/src/main/java/io/agentscope/core/util/StripedLocks.java
AgentScopeJavaBot, line 59: [nit] `get(null)` will throw a bare `NullPointerException` from `key.hashCode()` with no context. Consider adding an explicit guard for a clearer diagnostic:
```java
public ReentrantLock get(Object key) {
    Objects.requireNonNull(key, "key must not be null");
    return stripes[indexFor(key)];
}
```
All current call sites pass non-null keys, so this is purely defensive.

agentscope-core/src/main/java/io/agentscope/core/util/StripedLocks.java
AgentScopeJavaBot, line 78: [nit] `ceilingPowerOfTwo` overflows silently when `value > (1 << 30)`: `highestOneBit` returns `1 << 30`, the left-shift wraps to `Integer.MIN_VALUE`, and the subsequent `new ReentrantLock[negative]` throws `NegativeArraySizeException`. Low risk in practice (callers use 64), but a one-line guard would make the class bulletproof:
```java
if (value > (1 << 30)) {
    throw new IllegalArgumentException("stripeCount too large: " + value);
}
```

agentscope-core/src/test/java/io/agentscope/core/util/StripedLocksTest.java
AgentScopeJavaBot, line 69: [nit] The concurrency test only locks a single key (`"shared/file.txt"`). Consider adding a second phase where threads operate on *distinct* keys to verify that striping does not introduce spurious blocking between unrelated keys (i.e., throughput is not serialized). This would give stronger coverage of the striping benefit.

agentscope-core/src/main/java/io/agentscope/core/util/StripedLocks.java
AgentScopeJavaBot, line 29: [praise] Clean, self-contained utility with excellent Javadoc. The decision to avoid pulling in Guava for ~40 lines of code is pragmatic — the `StripedLocks` class is simpler and has zero transitive dependencies. Hash spreading and power-of-two masking follow well-established `ConcurrentHashMap` idioms.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR correctly addresses an unbounded memory growth issue (#2486) by replacing three per-path `ConcurrentHashMap`-based lock registries with a fixed-size striped lock utility. The `StripedLocks` implementation is clean, well-documented, and follows established patterns (hash spreading à la `ConcurrentHashMap`, power-of-two sizing). All call sites are migrated consistently with correct `lock()/finally unlock()` semantics and no nested-locking pitfalls. The accompanying test suite covers the key invariants: bounded memory, deterministic stripe selection, argument validation, and concurrent mutual exclusion. No correctness issues found — this is a solid, minimal fix.

PR #2555, opened 2026-08-04
feat(sandbox-e2b): auto-prune old NATIVE_SNAPSHOT snapshots per sandbox

### LINKED ISSUE #2547: [Feature]: e2b: auto-prune old NATIVE_SNAPSHOT snapshots per sandbox to cap cloud storage
### Problem

In `agentscope-extensions-sandbox-e2b`, `E2bPersistenceMode.NATIVE_SNAPSHOT` calls `POST /sandboxes/{id}/snapshots` on every `stop()` (`E2bSandbox.doPersistWorkspace()`). The resulting snapshots are team-level, persistent, TTL-less E2B resources — AgentScope never deletes them, so they accumulate indefinitely and cloud storage costs grow unbounded. Locally, `LocalSnapshotSpec` writes `{sessionId}.tar` and only the latest `snapshotId` is ever referenced (the file is overwritten on each persist), so older cloud snapshots are dead storage that can never be used again.

### Proposal

1. `E2bPlatformHttp`: add `listSnapshots(sandboxId)` (`GET /snapshots?sandboxID=...`) and `deleteSnapshot(snapshotId)` (`DELETE /templates/{snapshotId}`, treating `404` as idempotent — mirrors the E2B Python SDKs `delete_snapshot()`, since the public OpenAPI has no dedicated `/snapshots` DELETE endpoint).
2. Add a configurable `snapshotRetention` (default `1`; `<=0` disables pruning), wired through `E2bSandboxClientOptions` and exposed on `E2bFilesystemSpec`.
3. In `NATIVE_SNAPSHOT` mode, after a successful snapshot create, list the sandboxs snapshots, exclude the just-created one, and delete the rest. Failures are best-effort — warn only, never block persistence.

### Notes / trade-offs

- `GET /snapshots` returns no `createdAt`, so keeping `N > 1` cannot be time-ordered; a default of `1` (only the snapshot referenced by the local `.tar`) is the safe choice.
- The kept snapshot is always the newest — the one the local `.tar` references — so pruning never removes a live recovery point.
- Snapshots survive sandbox deletion, so pruning still works after a sandbox is killed.
- E2Bs `createSnapshot` API has no `memory`/`keepMemory` option (filesystem-only is only available on the `pause` path), so the memory-image overhead cannot be avoided; pruning is the pragmatic mitigation for storage growth.

### Open question: should we also support pause-based persistence?

Beyond pruning, the maintainers may want to consider a second persistence mode (`E2bPersistenceMode.FILESYSTEM_ONLY`) built on the E2B `pause` API (`POST /sandboxes/{id}/pause` with `{"memory": false}`), which captures **filesystem-only** snapshots — no memory image, so faster to create and significantly cheaper to store. Recovery would `connect()` the same paused sandbox (AgentScope already has that path), and paused sandboxes are kept indefinitely with no TTL.

Trade-offs vs. the current snapshot approach:

- **Pros**: true filesystem-only persistence (the memory image is pure waste for the "persist files between calls" use case); no tar transfer cost; resumes the same sandbox.
- **Cons**: not portable to a new sandbox (no reusable snapshot id for cloning); if the sandbox dies while running (e.g. timeout `kill`), there is no backup to restore from — workspace is lost and can only be rebuilt from the manifest; shutdown must not kill the paused sandbox, so paused sandboxes linger on the account and need an explicit cleanup policy.

Would the maintainers be open to adding this as an opt-in mode, or should pruning alone be the scope of this change?

### Related

- `E2bSandbox.java`, `E2bPlatformHttp.java`, `E2bPersistenceMode.java`, `E2bSandboxClientOptions.java`

chcodex:
Closes #2547

Adds a configurable `snapshotRetention` to cap per-session cloud snapshot storage in `NATIVE_SNAPSHOT` mode.

## Changes
- `snapshotRetention` option wired through `E2bSandboxClientOptions` and exposed on `E2bFilesystemSpec`; default `0` (disabled), `<=0` disables pruning
- `E2bSandboxState` gains `snapshotIds`, a per-session record of the snapshots this session has created (most recent last), persisted with the sandbox state; older serialized states lacking the field default to an empty list (backwards compatible)
- Retention keeps the last `retention` ids **by insertion order** (most recent last) and deletes the rest via `DELETE /templates/{snapshotId}` (404 idempotent). Insertion order equals creation time because the list is appended on each persist, so no alias timestamp parsing is needed and any `snapshotID` format is supported.
- Pruning is best-effort: individual failures are logged, the failed id is kept in the record for a later retry, and a failure never blocks persistence
- `deleteSnapshot` builds the URL with `HttpUrl.addPathSegment`, URL-encoding the namespace `/` in a snapshot id (`team/alias:tag` -> `%2F`); an unencoded path was parsed as an extra segment and E2B returned 404

## Design notes (vs. proposal in #2547)
- E2B locks a snapshot template while a sandbox restored from it is running and rejects deleting it with `HTTP 400`. Pruning is therefore deferred to `shutdown()`: `persist` only appends to the per-session record, and after `killSandbox` unlocks the templates `E2bPlatformHttp.cleanupSnapshots` keeps the last `snapshotRetention` ids by insertion order and deletes the rest in one pass, correcting any residue from earlier persists.

- Fixes `SandboxLifecycleMiddleware.releaseForCall()`: it now releases (stop/persist) the sandbox before persisting the state, so mutations made during `stop()` — including the per-session snapshot record updated in `doPersistWorkspace` — are captured by the state persist.

- Issue proposed pruning per sandbox via `GET /snapshots?sandboxID=...`, with a default of `1`. This approach was verified against the live API to be insufficient for sessions resumed from their own snapshots: the source sandbox id changes on every restore, so a `sandboxID` filter never sees the session's older snapshots, and E2B's explicit `kill()` does not delete historic snapshots either.
- This PR therefore keeps a per-session record in `E2bSandboxState` instead of querying `GET /snapshots` at all, making pruning independent of both the sandbox id and any team-wide listing. Default stays `0` to preserve historical behaviour (opt-in).

## Tests
14 e2b module tests pass, including 11 `E2bPlatformHttpTest` cases covering delete semantics (incl. URL encoding), FIFO retention by insertion order (`snap-*` and `team/alias:tag`), and failure tolerance.


oss-maintainer (commented):
### Summary
This PR modifies 8 files with 697 lines of diff.

### Observations
- Files changed: 8
- CLA: ✅ Signed

### Suggestions

---

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bSandbox.java
```diff
@@ -222,6 +228,33 @@ private void restoreSandboxFromSnapshotTemplate(String snapshotTemplateId) throw
...
+        } catch (Exception e) {
+            log.warn("[sandbox-e2b] snapshot cleanup best-effort skipped: {}", e.getMessage());
+        }
+    }
+
+    /**
+     * AgentScope-native snapshot alias: {@code agentscope-<shortId>-<epochMillis>}, where the
+     * middle segment is an 8-hex-char short UUID generated locally. The trailing 13-digit epoch
+     * millis timestamp makes ordering deterministic; anything not matching this format is treated as
+     * a legacy/foreign snapshot and never pruned.
```
oss-maintainer: **[Warning]** Stale javadoc after the retention rework: it still says ids matching `agentscope-<shortId>-<epochMillis>` are ordered by embedded timestamp and that "anything not matching this format is treated as a legacy/foreign snapshot and never pruned". Since a12ed17 / 217ce04, both `pruneSnapshots` and `cleanupSnapshots` drop the timestamp parsing entirely and prune purely by insertion position, so foreign-shaped ids held in `E2bSandboxState.getSnapshotIds()` **will** now be deleted. Please reword this comment to describe position-based retention (the `snapshotName()` helper itself is still correct).

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bPlatformHttp.java
```diff
@@ -67,12 +74,106 @@ JsonNode connectSandbox(String sandboxId, int timeoutSeconds) throws IOException
...
+                                    SandboxErrorCode.SNAPSHOT_PERSIST_ERROR,
+                                    "E2B snapshot delete failed: HTTP " + res.code());
+                        }
+                    }
+                    return null;
+                });
+    }
+
+    /** Retention keeps the last N snapshots by insertion order (most recent last). */
+    List<String> pruneSnapshots(
```
oss-maintainer: **[Info]** `pruneSnapshots` now has no production caller — `E2bSandbox.cleanupSnapshots()` goes through `cleanupSnapshots(snapshotIds, retention)` (line ~158) instead, and only the tests exercise this method. If the persist-path hook was intentionally dropped in favour of the restore-time one-shot cleanup, consider removing `pruneSnapshots` (and folding its deletion-failure-kept tests onto `cleanupSnapshots`) so the two near-duplicate retention implementations can't drift apart.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bPlatformHttp.java
```diff
@@ -67,12 +74,106 @@ JsonNode connectSandbox(String sandboxId, int timeoutSeconds) throws IOException
...
+    }
+
+    private static List<String> append(String keepSnapshotId, List<String> olderSnapshotIds) {
+        List<String> all = new ArrayList<>(olderSnapshotIds);
+        all.add(keepSnapshotId);
+        return all;
+    }
+
+    /** One-shot cleanup keeps the last {@code retention} ids by insertion order. */
+    List<String> cleanupSnapshots(List<String> snapshotIds, int retention) {
```
oss-maintainer, line 119: **[Info]** Minor: neither retention helper de-duplicates its input list. If a persisted id appears twice in `E2bSandboxState.getSnapshotIds()` (e.g. partially-failed state save), positional retention counts it twice and can keep `retention + duplicates - 1` snapshots. The old timestamp code had the same behaviour, so not a regression — a `LinkedHashSet` pass before slicing would make the "keep last N" contract exact.

oss-maintainer (commented):
### Summary
Delta re-review of the 6 commits since `cd4e07bc` (my previous COMMENTED review): snapshot retention switched from embedded-timestamp ordering to insertion-order FIFO, the timestamp regex/parser was removed, and coverage was extended (FIFO `snap-*` ids, deletion-failure retention, fd-leak fix in the persist test). The new semantics are self-consistent and better aligned with how ids are persisted in `E2bSandboxState` — insertion order is authoritative, so opaque/foreign ids no longer escape pruning. Main follow-up is the stale `never pruned` javadoc in `E2bSandbox`; the rest is hygiene. CI on this head is green (build ubuntu/windows, license, module sync).

---

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bSandbox.java
chcodex: Done in 257c6149: rewrote the `snapshotName()` javadoc to describe position-based retention (insertion order, most recent last, any id format prunable; the epoch millis segment is human-readable only and never parsed), and fixed the same stale "embedded timestamp" wording in `E2bSandbox.cleanupSnapshots()`.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bPlatformHttp.java
chcodex: Done in 257c6149: removed the uncalled `pruneSnapshots()` (+ `append` helper); `cleanupSnapshots()` is now the single retention path. Its retention tests were folded onto `cleanupSnapshots` (`cleanupSnapshotsRetentionOneDeletesAllButNewest` incl. namespaced-id encoding assertion, `cleanupSnapshotsRetentionKeepsConfiguredCount`, `cleanupSnapshotsSkipsNullAndBlank`); the delete-failure-kept case was already covered by `cleanupSnapshotsFailurePrependsToKept`. E2bPlatformHttp instr coverage 84% -> 96%.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bPlatformHttp.java
chcodex: Done in 257c6149: `cleanupSnapshots()` now dedupes its input via `LinkedHashSet` (first occurrence wins) and drops blank ids, so "keep last N" counts distinct snapshots. Covered by `cleanupSnapshotsDeduplicatesIds` and `cleanupSnapshotsDuplicatesWithinRetentionNeedNoDelete`.

chcodex:
@Buktal The 3 review comments are addressed in 257c6149 and pushed; the corresponding threads are replied. PTAL when convenient:

- stale `snapshotName()` javadoc rewritten to position-based retention
- uncalled `pruneSnapshots()` removed, tests folded onto `cleanupSnapshots()`
- `cleanupSnapshots()` dedupes input via LinkedHashSet

49 e2b tests pass locally, E2bPlatformHttp instr coverage 84% -> 96%.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bPlatformHttp.java
```diff
@@ -67,12 +75,80 @@ JsonNode connectSandbox(String sandboxId, int timeoutSeconds) throws IOException
...
+    private static List<String> dedupe(List<String> snapshotIds) {
+        if (snapshotIds == null) {
+            return new ArrayList<>();
+        }
+        LinkedHashSet<String> seen = new LinkedHashSet<>();
+        for (String id : snapshotIds) {
+            if (id == null || id.isBlank()) {
+                continue;
+            }
+            seen.add(id);
```
oss-maintainer, line 147: **[Info]** `LinkedHashSet` keeps the *first* occurrence, so a re-added id stays at its older position and is the first candidate for deletion. That is the right call for the partially-failed-save case this dedupe guards (same id = same snapshot), and the comment on line 116-117 already says so explicitly — no change needed. Just flagging that the ordering choice is now load-bearing for the "keep last N" contract, so if snapshot ids ever become re-used for a *new* template, the dedupe would have to flip to last-occurrence-wins.

oss-maintainer (approved):
### Summary
Delta re-review of the single commit since my previous pass: `257c614` addresses both open points — the stale `never pruned` javadoc on `snapshotName()` is rewritten to describe insertion-order retention (and links `E2bPlatformHttp#cleanupSnapshots`), the unused `pruneSnapshots` helper is deleted along with its duplicated test matrix, and `cleanupSnapshots` now de-duplicates its input (my earlier `[Info]`), so a re-persisted id counts once against the retention budget. Blank/null filtering moved into `dedupe`, which also makes the `retention <= 0` / `size <= retention` early returns hand back a clean list instead of the caller's raw list — a small improvement for `E2bSandbox#cleanupSnapshots`, which writes that result straight into `E2bSandboxState`.

### Findings
- **[Info]** `E2bPlatformHttp.java:147` — first-occurrence-wins is documented and correct here; noting only that the ordering is now part of the retention contract.

CI on this head is green (ubuntu + windows build, license, module sync, codecov) and CLA is signed.

oss-maintainer (commented):
### Summary
Re-review of `13d73380` ("roll back snapshot record on persist failure"). The failure mode being fixed is real: `doPersistWorkspace()` appends the new snapshot id before the archive is handed to the snapshot, so a `persist()` throw leaves the shared record pointing at an archive that was never written, and a later `cleanupSnapshots()` could then prune the snapshot the archive actually references. Catching it in `stop()` is the right place, `close()` still logs-and-proceeds so `shutdown()` is unaffected, and the two new MockWebServer tests pin both outcomes precisely (failed stop -> snapshot create + kill only, record unchanged; successful stop -> create + kill + one `DELETE /templates/...`, newest kept). CI is green on this head and CLA is signed.

What I want resolved before this is approved is narrower than the change itself: the rollback is positional rather than identity-based, and its stated justification depends on the corrected record actually reaching the store.

### Findings
- **[Warning]** `E2bSandbox.java:85` — `subList(recordedBefore, size).clear()` drops every id appended after the marker, including any written concurrently or by a shared-workspace counterpart, untracking snapshots that are still referenced.
- **[Warning]** `E2bSandbox.java:77` — `stop()` re-throws, so if `SandboxManager.release` does not still call `persistState` after a stop failure, the rollback stays in memory and the restart/node-switch case the comment cites is not covered.
- **[Info]** `E2bSandboxClient.java:155` — primitive `int` makes `0` mean both "unset" and "disabled", so a call can no longer turn off a client-level retention; a nullable field would keep `E2bFilesystemSpec.snapshotRetention(0)` meaningful.
- **[Info]** `E2bPlatformHttpTest.java:315` — deleting the vacuous assertion also removes the only coverage of the blank-`apiBaseUrl` -> `trimSlash` default.

### Suggestions
Both warnings are small: record the id `doPersistWorkspace()` created and remove that single element on failure, and extend `SandboxLifecycleMiddlewareReleaseOrderTest` with the `stop()`-throws case asserting `persistState` still runs.

---

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bSandbox.java
```diff
@@ -67,6 +69,24 @@ public void start() throws Exception {
...
+            super.stop();
+        } catch (Exception e) {
+            // doPersistWorkspace records the new snapshot id before the archive is persisted;
+            // on failure the archive still references the previous snapshot, so roll the record
+            // back. This keeps the shared state consistent by itself — no local flag could cover
+            // restarts, node switches, or repeated shutdowns.
+            List<String> ids = e2bState.getSnapshotIds();
+            if (ids.size() > recordedBefore) {
+                ids.subList(recordedBefore, ids.size()).clear();
+            }
```
oss-maintainer, line 106: **Positional rollback is not safe if anything else appends to the shared record.** This clears every id added after `recordedBefore`, but the list is shared state (`E2bSandboxState.getSnapshotIds()`, also re-assigned by `cleanupSnapshots()` via `setSnapshotIds`) and nothing here serialises writers, so any id appended concurrently — or by the other side of the `sharedLocalWorkspace` asymmetry #3247 is building toward — is dropped from the record together with this call's. For that other writer the consequence is worse than the failure this commit is fixing: its archive references a snapshot that is no longer tracked, so it can never be pruned.

Rolling back by identity instead of by position keeps the same intent and removes the blast radius: have `doPersistWorkspace()` remember the id it just created (one field on the sandbox, set right before the `add`) and delete exactly that id here, e.g.

```java
String created = lastCreatedSnapshotId;
if (created != null) {
    e2bState.getSnapshotIds().remove(created);
}
```

(then clear `lastCreatedSnapshotId` on the success path).

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bSandbox.java
```diff
@@ -67,6 +69,24 @@ public void start() throws Exception {
         super.start();
     }
 
+    @Override
+    public void stop() throws Exception {
+        int recordedBefore = e2bState.getSnapshotIds().size();
+        try {
+            super.stop();
+        } catch (Exception e) {
```
oss-maintainer, line 99: **Does a `stop()` failure still reach `persistState`?** The comment's rationale is that the corrected record must survive in *persisted* state — that is the whole reason for not using a local flag. But `SandboxLifecycleMiddleware` now documents exactly that ordering (`release` → stop + optional shutdown, then persist sandbox session state), and `stop()` re-throws here. If `SandboxManager.release` propagates that exception instead of continuing to `persistState`, the rollback only exists in memory: the next node / next restart reloads the old record and prunes against it — i.e. the case the comment says "no local flag could cover" is not actually covered.

`SandboxLifecycleMiddlewareReleaseOrderTest` asserts release-before-persist for the success path only. Could it also pin the failure path — `stop()` throwing, `persistState` still called, and the rolled-back `snapshotIds` reaching the store? If the answer is that `release` already swallows stop failures, then a one-line comment here would save the next reader the same trace.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bSandboxClient.java
```diff
@@ -150,6 +150,11 @@ private E2bSandboxClientOptions merge(E2bSandboxClientOptions call) {
         o.setConnectTimeoutSeconds(call.getConnectTimeoutSeconds());
         o.setReadTimeoutSeconds(call.getReadTimeoutSeconds());
         o.setMaxRetries(call.getMaxRetries());
+        // 0 means "left at default" (indistinguishable from an explicit disable on a primitive),
+        // so only a positive per-call value overrides; the client default governs otherwise.
+        if (call.getSnapshotRetention() > 0) {
```
oss-maintainer, line 155: The trade-off is honestly documented, so this is just a note on the shape of it: because `snapshotRetention` is a primitive `int`, `0` means both "unset" and "explicitly disabled", so a call can never turn off a retention configured on the client. The usual fix for a tri-state on a primitive is to make the field nullable (`Integer`, or a `snapshotRetentionSet` flag) rather than special-casing the merge — that also keeps `E2bFilesystemSpec.snapshotRetention(0)` meaningful instead of silently ignored. Happy to see this stay as-is if the asymmetry is considered acceptable.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/test/java/io/agentscope/extensions/sandbox/e2b/E2bPlatformHttpTest.java
```diff
@@ -0,0 +1,315 @@
...
+        com.fasterxml.jackson.databind.JsonNode full =
+                om.readTree(
+                        "{\"sandboxID\":\"sid-2\",\"domain\":\"d.e2b.app\",\"envdAccessToken\":\"tok\",\"envdVersion\":\"0.2.0\"}");
+        platform.applySandboxFields(state, full);
+        assertEquals("sid-2", state.getSandboxId());
+        assertEquals("d.e2b.app", state.getSandboxDomain());
+        assertEquals("tok", state.getEnvdAccessToken());
+        assertEquals("0.2.0", state.getEnvdVersion());
+    }
+}
```
oss-maintainer, line 325: Removing the test is right — `assertTrue(p != null)` asserted nothing that could fail. Worth noting the deletion also removes the last touch point for the blank-`apiBaseUrl` -> `trimSlash` default path, which now has no coverage anywhere; if that path matters, a direct assertion on the resolved base URL (no HTTP call needed) would replace it cheaply.

PR #2560, opened 2026-08-05
feat(kotlin): add Kotlin coroutine extension layer (suspend / Flow APIs)

### LINKED ISSUE #362: [Feature]: Add Kotlin coroutine extension layer (suspend / Flow APIs)
Background

agentscope-java exposes many async APIs based on Reactor (Mono / Flux).
This works well for Java, but is not idiomatic for Kotlin, where suspend functions and Flow<T> are the standard async model.

Proposal

Add a Kotlin coroutine extension module, e.g.:

agentscope-kotlin

This module would not change the core implementation, but provide Kotlin-friendly APIs on top of the existing Reactor-based core.

Goal

At the API level, convert:

Mono<T> → suspend fun

Flux<T> → Flow<T>

So Kotlin users rarely need to interact with Reactor types directly.

Example
suspend fun Agent.send(request: Request): Response
fun Agent.streamEvents(): Flow<Event>

Internally backed by Mono.awaitSingle() and Flux.asFlow().

Summary

Keep Reactor as the core async implementation

Add a Kotlin coroutine extension layer

Improve Kotlin developer experience without impacting Java users

liulangjietou:
Closes #362.

## Summary

Adds `agentscope-extensions-kotlin`, exposing the Reactor-based agent APIs as coroutine-friendly `suspend` functions and `Flow`, so Kotlin users rarely need to touch `Mono`/`Flux` directly. The core is untouched — this only adapts the boundary.

```kotlin
val reply = agent.callSuspend("What is AgentScope?")
agent.streamEventsFlow("Hello").collect { event -> render(event) }
```

The module existed on `1.x` but was not carried over to 2.0 (the BOM still carried a now-unreferenced `kotlin.version`). This is a rebuild against the 2.0 API rather than a port — see the notes below for why a straight port would not have worked.

## What is wrapped

| Reactor API | Kotlin extension |
|---|---|
| `CallableAgent.call(...)` → `Mono<Msg>` | `callSuspend(...)` — 12 overloads (message / list / text × plain, `Class`, `JsonNode`) |
| `ObservableAgent.observe(...)` → `Mono<Void>` | `observeSuspend(...)` |
| `ReActAgent.streamEvents(...)` → `Flux<AgentEvent>` | `streamEventsFlow(...)` → `Flow<AgentEvent>` — 6 overloads |
| `HarnessAgent.streamEvents(...)` | same 6 overloads |

**`StreamableAgent.stream(...)` is deliberately not wrapped**: every overload is `@Deprecated(since = "2.0.0", forRemoval = true)` in favour of `streamEvents`. Shipping a new module built on an API scheduled for removal seemed worse than leaving it out — happy to add it (deprecated as well) if you'd prefer parity with 1.x.

## Design notes

- **Naming.** Extension functions cannot reuse the member names: Kotlin resolves member functions before extensions, so an extension called `call` would be permanently shadowed by `CallableAgent.call` and never invoked. The `Suspend`/`Flow` suffixes match 1.x, so code migrating from 1.x keeps compiling.
- **Harness is `provided`.** The `HarnessAgent` extensions compile into their own class file (`HarnessAgentExtensionsKt`), so a core-only application never triggers its class loading. Verified with `javap`: `AgentExtensionsKt` and `ReActAgentExtensionsKt` contain zero references to harness classes.
- **`observeSuspend` awaits completion, not a value** — `observe` returns `Mono<Void>`, which never emits, so `awaitSingle()` would throw. It uses `awaitSingleOrNull()`.
- **`kotlin.version` 1.9.24 → 2.2.0.** The 1.9 compiler refuses the metadata of our current Kotlin-based dependencies (kotlinx-coroutines 1.10.x ships metadata 2.1, okhttp/okio ship 2.2), so the inherited value simply could not compile this module. The property had been unreferenced since the 1.x module was dropped, so nothing else is affected.
- **Distribution.** Registered in `agentscope-all` and `agentscope-bom`; `.github/scripts/check-shade-and-bom-sync.sh` passes locally.

## Tests

14 tests, all passing. Two layers — a recording fake pins down semantics deterministically, and a real `ReActAgent` + `MockModel` proves it works against the actual implementation:

- delegation of every overload shape to the right member method, including that the structured-output `Class` / `JsonNode` arguments are forwarded unchanged
- error signals surface as thrown exceptions
- **cancellation propagates** — cancelling the coroutine cancels the underlying `Mono` subscription (asserted via `doOnCancel`)
- `Flow` emits the full event stream in order (`AgentStartEvent` … `AgentEndEvent`)
- the returned `Flow` is **cold** — collecting twice starts two agent runs (asserted via `MockModel.callCount`)
- terminating collection early (`first()`) cancels upstream rather than hanging

Locally verified: `mvn test` on the new module (14/14), full rebuild of `agentscope-core` + `agentscope-harness` with the BOM change, `mvn spotless:check`, the module-sync script, and license headers on every new file.


oss-maintainer (commented):
### Summary
This PR modifies 12 files with 986 lines of diff.

### Observations
- Files changed: 12
- CLA: ✅ Signed

### Suggestions

---

PR #2567, opened 2026-08-05
fix(harness): hide host offload path from sandbox agents

### LINKED ISSUE #2566: [Bug]:compaction summary leaks host filesystem path into the agent prompt (invalid inside Docker sandbox)
**Describe the bug**

When compaction is enabled and the agent runs in a Docker sandbox, the injected summary message leaks a **host-side filesystem path** to the original conversation history instead of a sandbox-visible path.

With `CompactionConfig.offloadBeforeCompact` at its default (`true`), `MemoryFlushManager#resolveOffloadPath` returns the host-local session-JSONL path via `workspaceManager.resolveSessionContextFile(...).toString()`, and `ConversationCompactor#buildSummaryMessage` (~L441) embeds it verbatim into the agent-facing summary text:

> *The full conversation history has been saved to `.agentscope\docker-compaction\workspace\alice\agents\compaction-agent\sessions\alice-session-id.jsonl` should you need to refer back to it for details.*

This path does not exist inside the container (where `pwd` is `/workspace`). On a Windows host the path also contains backslashes, which are valid filename characters — not separators — on Linux, so any attempt by the agent to "refer back" to it fails. The statement is factually wrong and can mislead the agent.

**To Reproduce**

1. Build a `HarnessAgent` with a `DockerFilesystemSpec` and a `CompactionConfig` that triggers compaction (default `offloadBeforeCompact(true)`):

```java
        .name("compaction-agent")
        .model(model)
        .workspace(Paths.get(".agentscope/docker-compaction/workspace"))
        .stateStore(new JsonFileAgentStateStore(Paths.get(".agentscope/docker-compaction/state")))
        .compaction(CompactionConfig.builder()
                .triggerMessages(10)
                .keepMessages(3)
                .build())
        .filesystem(new DockerFilesystemSpec()
                .image("ubuntu:24.04")
                .snapshotSpec(new LocalSnapshotSpec(Paths.get(".agentscope/docker-compaction/snapshot"))))
        .build();
RuntimeContext ctx = RuntimeContext.builder().userId("alice").sessionId("alice-session-id").build();
agent.call(new UserMessage("Create an agents.md file declaring your name is coder ..."), ctx).block();
```

2. The single `agent.call` must exercise enough tool calls for the message history to exceed `triggerMessages`. The prompt above triggers 5 tool rounds — `list_files` ×2, `execute("pwd && ls -la")`, `write_file`, `read_file` — i.e. 11 messages, which trips compaction.
3. Inspect the injected summary USER message (`name = __compaction_summary__`) in the session JSONL. It contains the host path shown above; inside the container that path does not exist.

**Expected behavior**

The compaction summary should not present a host-local filesystem path to an agent executing inside a sandbox. Either omit the path from the summary text (the offloaded archive is for operator/external use, not for the agent to `cat` inside the container), or expose a path that is actually resolvable within the sandbox namespace (if a host→sandbox path translation is available).

**Error messages**

No exception or stack trace — this is a silent semantic defect. The observable signal is the incorrect path inside the summary message. If the agent subsequently tries to read it (e.g. `read_file(".agentscope\docker-compaction\...")`), the tool call fails with `Empty or not a directory` / file-not-found because the path does not exist in the container.

**Environment (please complete the following information)**

- Java Version: 17
- OS: Windows (host); sandbox is an `ubuntu:24.04` Docker container

**Additional context**

- Workaround: `CompactionConfig.builder().offloadBeforeCompact(false)` makes `buildSummaryMessage` take the path-less `else` branch. But this also disables message offload, so it is a workaround rather than a fix.
- The bug is mostly invisible under the Local (non-sandbox) filesystem — the path separator is correct and the file lives in the same filesystem. The Docker sandbox amplifies it (path does not exist + `\` is not a separator on Linux), which is likely why it went unnoticed.
- Relevant source (2.0.0): `agentscope-harness/.../memory/compaction/ConversationCompactor.java` (~L441, `buildSummaryMessage`) and `agentscope-harness/.../memory/MemoryFlushManager.java` (~L194, `resolveOffloadPath`).

jaideeppyne:
## Summary

- keep compaction offloading enabled for sandbox-backed filesystems
- omit the host-side session archive path from the agent-facing summary because that path is not resolvable inside the sandbox
- preserve the existing archive-path guidance for local filesystems
- add regression coverage for both sandbox and local behavior

Fixes #2566.

## Validation

- `mvn -pl agentscope-harness -am -Dtest=MemoryFlushManagerOffloadTest -Dsurefire.failIfNoSpecifiedTests=false test` — 5 tests run, 0 failures, 0 errors, 0 skipped; reactor build succeeded
- `mvn -q -pl agentscope-harness spotless:check` — passed
- `git diff --check` — passed

## Impact evidence

`MemoryFlushManager.resolveOffloadPath` currently resolves the archive through the host-side `WorkspaceManager` path, while sandbox filesystem operations use a separate namespace. Returning an empty path activates the compactor’s existing path-less summary branch, so the archive is still persisted but the agent is no longer told it can read a host-only path.

## Risk and rollback

The behavior change is limited to `AbstractSandboxFilesystem` implementations. Local filesystem summaries retain their current archive reference. Reverting this commit restores the previous behavior; no persisted data format or public API changes.

I used a coding agent to help write and verify this change.

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
@@ -24,6 +24,7 @@ import io.agentscope.core.message.ToolResultBlock;
 import io.agentscope.core.message.ToolUseBlock;
 import io.agentscope.core.model.Model;
 import io.agentscope.core.util.JsonUtils;
+import io.agentscope.harness.agent.filesystem.sandbox.AbstractSandboxFilesystem;
 import io.agentscope.harness.agent.memory.compaction.ConversationCompactor;
 import io.agentscope.harness.agent.memory.session.SessionTranscriptWriter;
 import io.agentscope.harness.agent.workspace.WorkspaceConstants;
@@ -192,6 +193,11 @@ public class MemoryFlushManager {
      */
     @Deprecated
     public String resolveOffloadPath(RuntimeContext rc, String agentId, String sessionId) {
+        // Session archives are persisted in the host-side workspace. A sandboxed agent cannot
+        // resolve that path, so do not advertise it as agent-readable in the summary prompt.
+        if (workspaceManager.getFilesystem() instanceof AbstractSandboxFilesystem) {
+            return "";
+        }
         return new SessionTranscriptWriter(workspaceManager)
                 .resolveContextPath(rc, agentId, sessionId);
     }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
@@ -25,6 +25,7 @@ import io.agentscope.core.message.ThinkingBlock;
 import io.agentscope.core.message.ToolResultBlock;
 import io.agentscope.core.message.ToolUseBlock;
 import io.agentscope.core.util.JsonUtils;
+import io.agentscope.harness.agent.filesystem.sandbox.AbstractSandboxFilesystem;
 import io.agentscope.harness.agent.transcript.TranscriptRef;
 import io.agentscope.harness.agent.transcript.TranscriptStore;
 import io.agentscope.harness.agent.workspace.WorkspaceConstants;
@@ -162,8 +163,18 @@ public class SessionTranscriptWriter {
     /**
      * Resolves the workspace-relative path of the session context JSONL (for compaction
      * summary references).
+     *
+     * <p>Returns an empty string for sandbox-backed filesystems: session archives live in the
+     * host-side workspace, which a sandboxed agent cannot resolve, so the path must not be
+     * advertised as agent-readable in the summary prompt. This mirrors the guard in {@link
+     * io.agentscope.harness.agent.memory.MemoryFlushManager#resolveOffloadPath}.
      */
     public String resolveContextPath(RuntimeContext rc, String agentId, String sessionId) {
+        // Session archives are persisted in the host-side workspace. A sandboxed agent cannot
+        // resolve that path, so do not advertise it as agent-readable in the summary prompt.
+        if (workspaceManager.getFilesystem() instanceof AbstractSandboxFilesystem) {
+            return "";
+        }
         try {
             Path file = workspaceManager.resolveSessionContextFile(rc, agentId, sessionId);
             Path ws = workspaceManager.getWorkspace();
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
@@ -16,11 +16,13 @@
 package io.agentscope.harness.agent.memory;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.mock;
 
 import io.agentscope.core.agent.RuntimeContext;
 import io.agentscope.core.message.Msg;
 import io.agentscope.core.message.MsgRole;
 import io.agentscope.core.message.TextBlock;
+import io.agentscope.harness.agent.filesystem.sandbox.AbstractSandboxFilesystem;
 import io.agentscope.harness.agent.memory.compaction.ConversationCompactor;
 import io.agentscope.harness.agent.memory.session.SessionEntry;
 import io.agentscope.harness.agent.memory.session.SessionTree;
@@ -35,6 +37,31 @@ import org.junit.jupiter.api.io.TempDir;
 class MemoryFlushManagerOffloadTest {
 
     @TempDir Path workspace;
+
+    @Test
+    void resolveOffloadPath_omitsHostPathForSandboxFilesystem() {
+        AbstractSandboxFilesystem filesystem = mock(AbstractSandboxFilesystem.class);
+        try (WorkspaceManager workspaceManager = new WorkspaceManager(workspace, filesystem)) {
+            MemoryFlushManager flushManager = new MemoryFlushManager(workspaceManager, null);
+
+            assertEquals(
+                    "",
+                    flushManager.resolveOffloadPath(
+                            RuntimeContext.empty(), "agent-a", "session-1"));
+        }
+    }
+
+    @Test
+    void resolveOffloadPath_keepsReachablePathForLocalFilesystem() {
+        try (WorkspaceManager workspaceManager = new WorkspaceManager(workspace)) {
+            MemoryFlushManager flushManager = new MemoryFlushManager(workspaceManager, null);
+
+            assertEquals(
+                    "agents/agent-a/sessions/session-1.jsonl",
+                    flushManager.resolveOffloadPath(
+                            RuntimeContext.empty(), "agent-a", "session-1"));
+        }
+    }
 
     @Test
     void offloadMessages_skipsAlreadyPersistedPrefix_andKeepsChain() throws Exception {
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriterTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriterTest.java
@@ -19,6 +19,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mock;
 
 import io.agentscope.core.agent.RuntimeContext;
 import io.agentscope.core.message.ImageBlock;
@@ -28,6 +29,7 @@ import io.agentscope.core.message.TextBlock;
 import io.agentscope.core.message.ToolResultBlock;
 import io.agentscope.core.message.ToolUseBlock;
 import io.agentscope.core.message.URLSource;
+import io.agentscope.harness.agent.filesystem.sandbox.AbstractSandboxFilesystem;
 import io.agentscope.harness.agent.transcript.FilesystemTranscriptStore;
 import io.agentscope.harness.agent.transcript.TranscriptRef;
 import io.agentscope.harness.agent.transcript.TranscriptStore;
@@ -176,6 +178,28 @@ class SessionTranscriptWriterTest {
             }
         }
         assertTrue(totalLines >= 2, "segments should contain both entries across flushes");
+    }
+
+    @Test
+    void resolveContextPath_omitsHostPathForSandboxFilesystem() {
+        AbstractSandboxFilesystem filesystem = mock(AbstractSandboxFilesystem.class);
+        try (WorkspaceManager wm = new WorkspaceManager(workspace, filesystem)) {
+            SessionTranscriptWriter writer = new SessionTranscriptWriter(wm);
+
+            assertEquals(
+                    "", writer.resolveContextPath(RuntimeContext.empty(), "agent-a", "session-1"));
+        }
+    }
+
+    @Test
+    void resolveContextPath_keepsReachablePathForLocalFilesystem() {
+        try (WorkspaceManager wm = new WorkspaceManager(workspace)) {
+            SessionTranscriptWriter writer = new SessionTranscriptWriter(wm);
+
+            assertEquals(
+                    "agents/agent-a/sessions/session-1.jsonl",
+                    writer.resolveContextPath(RuntimeContext.empty(), "agent-a", "session-1"));
+        }
     }
 
     private static Msg text(String id, MsgRole role, String text) {


jaideeppyne:
### Rebase complete (conflict resolved)

Rebased onto latest `main` after upstream refactored `resolveOffloadPath` to delegate to `SessionTranscriptWriter#resolveContextPath`.

**Resolution**
- Keep sandbox guard: if filesystem is `AbstractSandboxFilesystem`, return `""` (do not advertise host path).
- Otherwise delegate to `SessionTranscriptWriter` (matches current main).
- Local-path test now expects the **workspace-relative** path returned by `SessionTranscriptWriter` (`agents/agent-a/sessions/session-1.jsonl`).

**Evidence**
- `mvn -pl agentscope-harness -am -Dtest=MemoryFlushManagerOffloadTest -Dsurefire.failIfNoSpecifiedTests=false test` → **5 tests, 0 failures**
- `mvn -q -pl agentscope-harness spotless:check` → **passed**

jaideeppyne:
### Conflict resolution status

Conflict with `main` is real: `resolveOffloadPath` was refactored on `main` to delegate to `SessionTranscriptWriter#resolveContextPath`.

A rebased fix is ready locally and validated:

```text
mvn -pl agentscope-harness -am -Dtest=MemoryFlushManagerOffloadTest \
  -Dsurefire.failIfNoSpecifiedTests=false test
# Tests run: 5, Failures: 0
mvn -q -pl agentscope-harness spotless:check  # passed
```

**Resolution shape (on top of current `main`):**
```java
public String resolveOffloadPath(RuntimeContext rc, String agentId, String sessionId) {
    if (workspaceManager.getFilesystem() instanceof AbstractSandboxFilesystem) {
        return "";
    }
    return new SessionTranscriptWriter(workspaceManager)
            .resolveContextPath(rc, agentId, sessionId);
}
```
Local-path test expects workspace-relative path `agents/agent-a/sessions/session-1.jsonl`.

**Push blocked:** OAuth token lacks `workflow` scope, and rebasing onto `main` updates `.github/workflows/*`, which GitHub rejects without that scope. Branch tip remains the last green commit until a token with `workflow` can force-push the rebase (or conflict is resolved in the GitHub UI).

jaideeppyne:
### Rebased and force-pushed

Conflict with `main` resolved and branch updated.

**Resolution**
- Sandbox filesystems: `resolveOffloadPath` returns `""` (do not advertise host path)
- Local filesystems: delegate to `SessionTranscriptWriter#resolveContextPath` (workspace-relative path)

**Local validation**
- `mvn -pl agentscope-harness -am -Dtest=MemoryFlushManagerOffloadTest -Dsurefire.failIfNoSpecifiedTests=false test` → 5 tests, 0 failures
- `mvn -q -pl agentscope-harness spotless:check` → passed

oss-maintainer (commented):
### Summary
This PR modifies 2 files with 74 lines of diff.

### Observations
- Files changed: 2
- CLA: ✅ Signed

### Suggestions

---

jaideeppyne:
Thanks for the review. To confirm on the two suggestions: the change is backward compatible — `resolveOffloadPath` keeps its signature and only changes what it returns for sandbox filesystems (an empty string, so no host path is advertised) while local filesystems delegate to the existing `SessionTranscriptWriter#resolveContextPath`; no public API changes. Unit coverage for the new logic paths is included (`MemoryFlushManagerOffloadTest`, covering both the sandbox and local cases). Glad to expand it further if you'd like additional cases.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
```diff
@@ -192,6 +193,11 @@ public Mono<Void> flushMemories(RuntimeContext rc, List<Msg> messages) {
      */
     @Deprecated
     public String resolveOffloadPath(RuntimeContext rc, String agentId, String sessionId) {
+        // Session archives are persisted in the host-side workspace. A sandboxed agent cannot
+        // resolve that path, so do not advertise it as agent-readable in the summary prompt.
+        if (workspaceManager.getFilesystem() instanceof AbstractSandboxFilesystem) {
```
AgentScopeJavaBot, line 198: [major] **Symmetry / Migration risk**: The `@deprecated` Javadoc on this method (line 192) says *"Prefer `SessionTranscriptWriter#resolveContextPath`"*. However, `SessionTranscriptWriter.resolveContextPath` (line 166 of `SessionTranscriptWriter.java`) has no sandbox-filesystem guard — it unconditionally resolves and returns the host-side path. The fix is correct for the deprecated method, but the **recommended replacement API is still vulnerable**. When a maintainer migrates callers from `resolveOffloadPath` to `resolveContextPath`, the bug will silently reappear.

**Suggestion**: Move (or duplicate) the `instanceof AbstractSandboxFilesystem` guard into `SessionTranscriptWriter.resolveContextPath` so both APIs behave identically. Alternatively, add an explicit `@deprecated` note in the Javadoc warning callers that `resolveContextPath` must perform its own sandbox check, and file a follow-up issue to track the fix.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
```diff
@@ -36,6 +38,31 @@ class MemoryFlushManagerOffloadTest {
 
     @TempDir Path workspace;
 
+    @Test
+    void resolveOffloadPath_omitsHostPathForSandboxFilesystem() {
```
AgentScopeJavaBot, line 42: [minor] **Test coverage gap**: Both new tests exercise the deprecated `MemoryFlushManager.resolveOffloadPath`. Since the `@deprecated` Javadoc directs callers to `SessionTranscriptWriter.resolveContextPath`, consider adding a parallel test (e.g. in a `SessionTranscriptWriterTest`) that verifies `resolveContextPath` also returns empty for sandbox filesystems — once the guard is propagated there. This ensures the migration path is regression-safe.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR adds a sandbox-aware guard to the deprecated `MemoryFlushManager.resolveOffloadPath` method, returning an empty string when the workspace filesystem is backed by a sandbox. The intent is correct: sandboxed agents cannot resolve host-side workspace paths, so advertising them in the compaction summary prompt is misleading. The fix is minimal, well-commented, and includes two unit tests covering both the sandbox and local-filesystem paths. However, the guard is applied **only** to the deprecated method — the recommended replacement (`SessionTranscriptWriter.resolveContextPath`) has no equivalent protection, creating a migration hazard.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
jaideeppyne: Good catch — confirmed and fixed in 42e8ba7.

`SessionTranscriptWriter.resolveContextPath` did indeed lack the guard: it unconditionally resolved and returned the host-side workspace-relative path, so any caller migrating from the deprecated `resolveOffloadPath` to the recommended replacement would have lost the sandbox protection. I propagated the same `AbstractSandboxFilesystem` guard there so it now returns `""` for sandbox-backed filesystems, mirroring the logic in `resolveOffloadPath`.

Safety check: `resolveContextPath` has no other callers in the codebase (its only current caller path is via `resolveOffloadPath` -> `ConversationCompactor`, which feeds the agent-facing summary), and non-sandbox filesystems are unchanged — the local path is still returned as before, so no legitimate host-side use is broken.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
jaideeppyne: Added in 42e8ba7. `SessionTranscriptWriterTest` now has two parallel tests:

- `resolveContextPath_omitsHostPathForSandboxFilesystem` — asserts `""` is returned for an `AbstractSandboxFilesystem` (mocked), and
- `resolveContextPath_keepsReachablePathForLocalFilesystem` — asserts the reachable `agents/agent-a/sessions/session-1.jsonl` path is still returned for a local filesystem.

I verified the sandbox test genuinely exercises the guard: with the guard reverted it fails with `expected: <> but was: <agents/agent-a/sessions/session-1.jsonl>`, and passes with the guard in place. `mvn -pl agentscope-harness -am -Dtest=SessionTranscriptWriterTest,MemoryFlushManagerOffloadTest test` -> 11 tests, 0 failures; `spotless:check` passes.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
```diff
@@ -162,8 +163,18 @@ public void appendMessages(
...
+     *
+     * <p>Returns an empty string for sandbox-backed filesystems: session archives live in the
+     * host-side workspace, which a sandboxed agent cannot resolve, so the path must not be
+     * advertised as agent-readable in the summary prompt. This mirrors the guard in {@link
+     * io.agentscope.harness.agent.memory.MemoryFlushManager#resolveOffloadPath}.
      */
     public String resolveContextPath(RuntimeContext rc, String agentId, String sessionId) {
+        // Session archives are persisted in the host-side workspace. A sandboxed agent cannot
+        // resolve that path, so do not advertise it as agent-readable in the summary prompt.
+        if (workspaceManager.getFilesystem() instanceof AbstractSandboxFilesystem) {
```
oss-maintainer: This predicate is broader than "sandboxed", so it also hides the archive path for plain **local-shell** workspaces, where the JSONL really is host-readable.

`AbstractSandboxFilesystem` is a shell-capability marker, not an isolation marker: `LocalFilesystemWithShell` implements it (`filesystem/local/LocalFilesystemWithShell.java:50`), `LocalFilesystemSpec.toFilesystem()` wraps that upper layer with `OverlayFilesystem.of(upper, lower)` which returns the `ShellAwareOverlay` subtype (`filesystem/OverlayFilesystem.java:105-118`), and `projectWritable=true` returns `ProjectAwareOverlay extends OverlayFilesystem implements AbstractSandboxFilesystem` (`filesystem/ProjectAwareOverlay.java:54`). HarnessAgent already had to special-case exactly this at `HarnessAgent.java:2699` — *"ShellAwareOverlay is instanceof AbstractSandboxFilesystem, so this branch must come before the sandbox check below to avoid using the sandbox `/workspace` prefix for real host paths"* — and `WorkspaceContextMiddleware.java:358` writes the sandbox branch as `fs instanceof AbstractSandboxFilesystem sandbox && !(fs instanceof OverlayFilesystem)`.

Consequence here: with the default local spec, `resolveContextPath` now returns `""`, `ConversationCompactor.java:194-198` maps a blank path to `null`, and the summary message silently loses its "full history archived at <path>" pointer — the agent can no longer retrieve the transcript it was just told exists. That is a behavior regression for every non-sandbox deployment, which is the common case.

Suggest narrowing the test to the actual sandbox family (`fs instanceof BaseSandboxFilesystem`, i.e. `SandboxBackedFilesystem` / `PinnedSandboxFilesystem`) — or the repo's existing `instanceof AbstractSandboxFilesystem && !(instanceof OverlayFilesystem)` idiom — and adding a small dedicated accessor (e.g. `WorkspaceManager#isHostPathAgentVisible()`) so future wrappers cannot silently re-open this.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
oss-maintainer: This copy of the guard is unreachable by design: `resolveOffloadPath` is `@Deprecated` and its body only delegates to `new SessionTranscriptWriter(workspaceManager).resolveContextPath(...)` (line 201), which applies the identical check. Two copies of the same rule will drift — the comments here and on `resolveContextPath` are already worded differently.

Suggest deleting the check in the deprecated wrapper and letting `SessionTranscriptWriter` be the single decision point.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
```diff
@@ -36,6 +38,31 @@ class MemoryFlushManagerOffloadTest {
...
+
+            assertEquals(
+                    "",
+                    flushManager.resolveOffloadPath(
+                            RuntimeContext.empty(), "agent-a", "session-1"));
+        }
+    }
+
+    @Test
+    void resolveOffloadPath_keepsReachablePathForLocalFilesystem() {
```
oss-maintainer, line 76: `resolveOffloadPath_keepsReachablePathForLocalFilesystem` builds the local case as `new WorkspaceManager(workspace)`, i.e. `filesystem == null`. That does not represent the real local configuration — `HarnessAgentBuilderSupport.java:164-168` always installs a spec-built filesystem, whose local form is `ShellAwareOverlay`/`ProjectAwareOverlay` (both `instanceof AbstractSandboxFilesystem`). So the test suite passes while production local mode takes the new empty-path branch.

Could you add a case that wires the real local spec, e.g. `new WorkspaceManager(workspace, OverlayFilesystem.of(new LocalFilesystemWithShell(...), new LocalFilesystem(...)))` asserting the path is still returned, plus a `SandboxBackedFilesystem` case asserting empty? That would have caught the over-match.

oss-maintainer (commented):
### Summary
Delta re-review after `42e8ba7b` (propagate the guard into `SessionTranscriptWriter`). The direction is right — a host-side archive path is not resolvable from inside a container, and advertising it in the compaction summary sends sandboxed agents chasing a file they cannot open. But the chosen predicate, `getFilesystem() instanceof AbstractSandboxFilesystem`, is a shell-capability marker rather than an isolation marker, so it also matches the plain local-with-shell workspace (`LocalFilesystemSpec` → `ShellAwareOverlay`/`ProjectAwareOverlay`), where the path *is* host-readable. That is flagged inline as the one blocking point; two smaller ones follow.

### Verdict
**COMMENT** — one over-match to settle before merge (`SessionTranscriptWriter.java:175`); the rest is non-blocking. CLA signed, tests included for both branches of the new rule.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
jaideeppyne: Fixed in 851d9ff7. The guard now checks BaseSandboxFilesystem, so real sandbox-backed filesystems return an empty path while local shell-capable overlays retain the workspace-relative path. I added production-shaped local overlay and sandbox-backed tests.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
jaideeppyne: Removed the duplicate guard from deprecated resolveOffloadPath in 851d9ff7. It now delegates to SessionTranscriptWriter as the single decision point.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
jaideeppyne: Updated both test classes in 851d9ff7 to use a shell-aware OverlayFilesystem for the local case and SandboxBackedFilesystem for the sandbox case. Focused Maven tests and Spotless pass.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
```diff
@@ -162,8 +163,17 @@ public void appendMessages(
...
      * summary references).
+     *
+     * <p>Returns an empty string for sandbox-backed filesystems: session archives live in the
+     * host-side workspace, which a sandboxed agent cannot resolve, so the path must not be
+     * advertised as agent-readable in the summary prompt.
      */
     public String resolveContextPath(RuntimeContext rc, String agentId, String sessionId) {
+        // Session archives are persisted in the host-side workspace. A sandboxed agent cannot
+        // resolve that path, so do not advertise it as agent-readable in the summary prompt.
+        if (workspaceManager.getFilesystem() instanceof BaseSandboxFilesystem) {
```
oss-maintainer: **Narrowing to `BaseSandboxFilesystem` also switches off the real sandbox case `RoutedSandboxFilesystem`.**

`RoutedSandboxFilesystem implements AbstractSandboxFilesystem` directly (it is *not* a `BaseSandboxFilesystem`), and `HarnessAgent` installs it as the workspace filesystem for any sandboxed agent that declares file routes — `filesystem = filesystemRoutes.isEmpty() ? capturedSandboxFs : new RoutedSandboxFilesystem(capturedSandboxFs, filesystemRoutes)`, which is then handed to `new WorkspaceManager(resolvedWorkspace, filesystem, ...)`.

So for sandbox + `filesystemRoutes` (the Managed-Agent layout this class exists for) the guard no longer fires and the host-side archive path is advertised to the sandboxed agent again — the leak `resolveOffloadPath` was added to prevent, just in the configuration the old `instanceof AbstractSandboxFilesystem` check did cover.

The repo already has the unwrap idiom twice — `HarnessAgent` shell-policy selection (`filesystem instanceof SandboxBackedFilesystem || (filesystem instanceof RoutedSandboxFilesystem routed && routed.primary() instanceof SandboxBackedFilesystem)`) and `WorkspaceManager.taskRecordsRouteToLiveSandbox` (`if (fs instanceof RoutedSandboxFilesystem routed) fs = routed.backendFor(relPath)`). Suggest one shared predicate rather than a fourth hand-rolled `instanceof` chain, e.g.:

```java
static boolean sandboxBacked(AbstractFilesystem fs) {
    if (fs instanceof RoutedSandboxFilesystem routed) {
        fs = routed.primary();
    }
    if (fs instanceof OverlayFilesystem overlay) {   // ShellAwareOverlay case
        fs = overlay.getUpper();
    }
    return fs instanceof BaseSandboxFilesystem;
}
```

`OverlayFilesystem.of(upper, lower)` returns `ShellAwareOverlay`, which likewise implements `AbstractSandboxFilesystem` without extending `BaseSandboxFilesystem` — so a sandbox upper behind an overlay hits the same hole; if that combination is reachable from a builder today, the helper above covers it as well.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
```diff
@@ -36,6 +41,36 @@ class MemoryFlushManagerOffloadTest {
...
+        try (WorkspaceManager workspaceManager =
+                new WorkspaceManager(
+                        workspace,
+                        OverlayFilesystem.of(
```
oss-maintainer, line 80: **The new local-overlay case is good; the routed-sandbox case is what I would add.**

`OverlayFilesystem.of(new LocalFilesystemWithShell(...), new LocalFilesystem(...))` locks in the behaviour this commit is fixing (`LocalFilesystemWithShell` implements `AbstractSandboxFilesystem`, so the old interface check hid the path for local dev). The mirror case — `new RoutedSandboxFilesystem(sandboxBackedFs, Map.of("agents/", other))` must still return `""` — is not asserted anywhere, which is why the narrowing passes the suite. One `resolveOffloadPath_returnsEmptyForRoutedSandbox` test would pin the whole distinction.

oss-maintainer (commented):
### Summary

Delta re-review after `42e8ba7b` → `851d9ff7` (distinguish sandbox storage from local shell overlays). The motivation is correct and my previous blocking point is fixed: `LocalFilesystemWithShell` implements `AbstractSandboxFilesystem` purely as a *shell-capability* marker, so the old interface check hid a host-readable path from local-dev workspaces that can resolve it fine, and the two new `OverlayFilesystem.of(...)` test cases pin that behaviour properly. Removing the duplicated guard from the deprecated `MemoryFlushManager.resolveOffloadPath` in favour of the single delegate is also the right consolidation.

One issue holds this back: the replacement predicate is now too narrow in the opposite direction. `RoutedSandboxFilesystem implements AbstractSandboxFilesystem` without extending `BaseSandboxFilesystem`, and `HarnessAgent` uses it as the workspace filesystem for any sandboxed agent that declares file routes — so for sandbox + `filesystemRoutes` the host-side archive path is advertised to the agent again, which is the leak this PR was written to plug. The repo already solves this shape twice (`HarnessAgent` shell-policy selection and `WorkspaceManager.taskRecordsRouteToLiveSandbox` both unwrap `RoutedSandboxFilesystem.primary()` / `.backendFor(...)`), so I have suggested a shared `sandboxBacked(...)` predicate instead of a fourth hand-rolled `instanceof` chain.

### Findings

- **[Warning]** `SessionTranscriptWriter.java:174` — `instanceof BaseSandboxFilesystem` misses `RoutedSandboxFilesystem` (and `OverlayFilesystem.ShellAwareOverlay`), re-advertising the host path for sandboxed agents with file routes; suggest one shared unwrapping predicate
- **[Info]** `MemoryFlushManagerOffloadTest.java:62` — the local-overlay branch is now covered, but the routed-sandbox branch (the case that regressed) has no test, which is why the narrowing is green

### Verdict

**COMMENT** — non-blocking on its own if you can confirm `filesystemRoutes` + sandbox is not a supported combination for `HarnessAgent`; if it is, the unwrapping helper is a small change and the missing test would have caught it. CLA signed (`license/cla=success` on `851d9ff7`), CI green on this head (ubuntu + windows + codecov/patch).

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
jaideeppyne: Fixed in 389685c. Added shared FilesystemUtils.isSandboxBacked, which unwraps RoutedSandboxFilesystem and OverlayFilesystem before checking BaseSandboxFilesystem. SessionTranscriptWriter and HarnessAgent shell-policy selection now use the same classifier, covering routed and sandbox-upper overlay combinations while preserving local shell overlays.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerOffloadTest.java
jaideeppyne: Added resolveOffloadPath_omitsHostPathForRoutedSandboxFilesystem in 389685c. The SessionTranscriptWriter sandbox test now also wraps SandboxBackedFilesystem as an overlay upper, while the existing local-overlay test confirms local paths remain visible. Focused Maven tests and Spotless pass.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
```diff
@@ -162,8 +163,17 @@ public void appendMessages(
...
      * summary references).
+     *
+     * <p>Returns an empty string for sandbox-backed filesystems: session archives live in the
+     * host-side workspace, which a sandboxed agent cannot resolve, so the path must not be
+     * advertised as agent-readable in the summary prompt.
      */
     public String resolveContextPath(RuntimeContext rc, String agentId, String sessionId) {
+        // Session archives are persisted in the host-side workspace. A sandboxed agent cannot
+        // resolve that path, so do not advertise it as agent-readable in the summary prompt.
+        if (FilesystemUtils.isSandboxBacked(workspaceManager.getFilesystem())) {
```
oss-maintainer, line 174: **The guard is broader than the leak: the *relative* archive path really is reachable from inside a sandbox, only the absolute fallback is not.**

`resolveContextPath` has two exits. When the archive sits under the host workspace (`file.startsWith(ws)`) it returns a workspace-**relative** path (`agents/<agentId>/sessions/<sessionId>.jsonl`); the absolute host path is only returned by the `file.toString()` fallback. But `SessionTree.scheduleMirror()` mirrors exactly that relative path into the live filesystem — `mirrorToFilesystem(mirrorFs, contextFile, contextRel)` → `fs.uploadFiles(relPath, bytes)` — so for a sandbox-backed workspace the JSONL is present inside the sandbox at the relative path, and the sandboxed agent's file tools can open it.

So `#2566` (a host absolute path advertised to a sandboxed agent) is fixed by hiding the *fallback* branch, while the common branch loses a pointer that was working. Two options:

1. Narrow the new early-return to the unresolvable case, e.g. compute the relative path first and return `""` only when `!file.startsWith(ws)` and the filesystem is sandbox-backed; or
2. Keep the blanket suppression but state the tradeoff in the Javadoc ("sandboxed agents get no transcript pointer, even though `SessionTree` mirrors the archive into the sandbox"), so the next reader does not read this as a leak-only fix.

The same applies to the new `resolveOffloadPath_omitsHostPathForSandboxFilesystem` test: it pins the blanket behavior, so if the intent is (1), the local-workspace case should be asserted as `agents/agent-a/sessions/session-1.jsonl` for a sandbox filesystem too.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/util/FilesystemUtils.java
```diff
@@ -24,6 +28,19 @@ public final class FilesystemUtils {
 
     private FilesystemUtils() {}
 
+    /** Returns whether the filesystem ultimately stores files in a live sandbox. */
+    public static boolean isSandboxBacked(AbstractFilesystem filesystem) {
+        while (true) {
+            if (filesystem instanceof RoutedSandboxFilesystem routed) {
+                filesystem = routed.primary();
```
oss-maintainer, line 35: `primary()` ignores prefix routes, but this repository already has a route-aware idiom for exactly this question.

`RoutedSandboxFilesystem` sends file operations to the longest matching route (see `backendFor(path)`, whose Javadoc is "the backend that would serve `path`"), and `WorkspaceManager.taskRecordsRouteToLiveSandbox(relPath)` uses it for the same reason — a routed sandbox can mount a host-local or persistent store for `agents/...` while the primary stays a sandbox. A `memory-stores/`-style route is already visible in the new test, just not for the session prefix.

For a workspace configured as `RoutedSandboxFilesystem(sandboxBackedFs, {"agents/": hostLocalOrRemoteFs})`, session archives are served by the non-sandbox backend, so the classification here says "sandbox" and the pointer is dropped even though the agent can read it. Suggest an overload `isSandboxBacked(AbstractFilesystem fs, String relPath)` that resolves `routed.backendFor(relPath)` first (mirroring `taskRecordsRouteToLiveSandbox`) and let `SessionTranscriptWriter` pass the session-relative path, or document that primary-only is a deliberate conservative choice.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/util/FilesystemUtils.java
```diff
@@ -24,6 +28,19 @@ public final class FilesystemUtils {
 
     private FilesystemUtils() {}
 
+    /** Returns whether the filesystem ultimately stores files in a live sandbox. */
+    public static boolean isSandboxBacked(AbstractFilesystem filesystem) {
+        while (true) {
```
oss-maintainer, line 33: **Defensive nit: the unwrap loop has no depth bound.**

`while (true)` terminates today because neither `RoutedSandboxFilesystem.primary()` nor `OverlayFilesystem.getUpper()` can return the wrapper itself at construction time, so no cycle exists. The cost of being wrong here is a hang on the compaction path rather than a wrong boolean, and a cap is one line:

```java
for (int depth = 0; depth < 8; depth++) { ... }
return false; // or the last non-wrapped value
```

Also worth a Javadoc sentence on the layering rule being encoded: for an overlay the **upper** layer decides where new files land, so `Overlay(sandboxUpper, localLower)` is classified sandbox-backed while `Overlay(localUpper, sandboxLower)` is not. That is the right rule for "can the agent open a host path", but it is implicit today.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2703,9 +2704,7 @@ public HarnessAgent build() {
                     shellPolicy =
                             io.agentscope.harness.agent.skill.runtime.ShellPathPolicy
                                     .localWithShell(resolvedWorkspace);
-                } else if (filesystem instanceof SandboxBackedFilesystem
-                        || (filesystem instanceof RoutedSandboxFilesystem routed
-                                && routed.primary() instanceof SandboxBackedFilesystem)) {
+                } else if (FilesystemUtils.isSandboxBacked(filesystem)) {
```
oss-maintainer, line 2707: **This branch is a behavior change beyond the offload-path fix, and it is the one path through the PR with no test.**

Before this commit the sandbox branch matched `SandboxBackedFilesystem` or `RoutedSandboxFilesystem(primary = SandboxBackedFilesystem)`. `FilesystemUtils.isSandboxBacked` additionally matches:

- an `OverlayFilesystem` whose upper is a sandbox filesystem — which is what `OverlayFilesystem.of()` produces for a sandbox upper (`ShellAwareOverlay`), and
- an `OverlayFilesystem` produced by `OverlayFilesystem.of()` for a sandbox upper (`ShellAwareOverlay` with a sandbox-capable upper), which previously fell through to the trailing `else`.

Both newly take `ShellPathPolicy.sandbox(wsPrefix)` instead of the trailing `else`, which changes skill/script path rewriting for those workspaces at runtime. It is plausibly the intended "classify consistently" outcome and it matches the `WorkspacePathNormalizer` selection above (which already special-cases the local-overlay case first), but nothing in the PR description or the tests covers it, and `mvn -pl agentscope-harness -Dtest=MemoryFlushManagerOffloadTest` cannot catch a regression here.

Could you add a focused assertion for the shell-policy selection (local overlay → `localWithShell`, wrapped/routed sandbox → `sandbox(wsPrefix)`, no shell → `noShell`) or, if the HarnessAgent edit is only a refactor of the previously matched cases, note that here so reviewers do not have to re-derive the type matrix?

oss-maintainer (commented):
### Summary

Delta re-review after `851d9ff7` → `389685ce` ("classify wrapped sandbox filesystems consistently"). The blocking point from the previous round is resolved: `FilesystemUtils.isSandboxBacked` now unwraps `RoutedSandboxFilesystem.primary()` and `OverlayFilesystem.getUpper()`, so a routed sandbox with file routes is no longer classified as host-readable, and the predicate is shared with the `HarnessAgent` shell-policy branch instead of being re-derived per call site. The routed case is also pinned by `resolveOffloadPath_omitsHostPathForRoutedSandboxFilesystem`. CI is green on ubuntu + windows and CLA is signed.

What is left is a scope question rather than a correctness bug: the guard suppresses the pointer for the *common* case too. `resolveContextPath` returns a workspace-relative path whenever the archive is under the workspace, and `SessionTree.scheduleMirror()` uploads exactly that relative path into the live (sandbox) filesystem — so a sandboxed agent can open it, while only the absolute-path fallback is genuinely unresolvable. Flagged inline with two ways to settle it. Two smaller points follow: route-awareness of `primary()` versus `backendFor(relPath)`, and the `HarnessAgent` branch, which is a behavior change that arrives without a test.

### Verdict

**COMMENT** — no blocking defect found; the four inline points are about narrowing the suppression and covering the `HarnessAgent` classification change. CLA signed, so this stays a comment rather than an approval only because the first point is a real behavior question for the author to answer.

---

PR #2222, opened 2026-07-15
feat(a2a): migrate server runner to streamEvents

guslegend0510:
## Summary

- migrate the A2A server `AgentRunner` contract from deprecated `stream()` / `Event` to `streamEvents()` / `AgentEvent`
- update `BaseReActAgentRunner` to call `ReActAgent.streamEvents()`
- adapt `AgentScopeAgentExecutor` to convert typed agent events into A2A streaming artifacts and final messages
- update A2A server tests for typed events, including lifecycle-event filtering

Fixes #2207

## Test

- `mvn -pl agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server -am test`
- `mvn -pl agentscope-extensions/agentscope-spring-boot-starters/agentscope-a2a-spring-boot-starter -am -DskipTests test`
- `mvn -pl agentscope-extensions/agentscope-spring-boot-starters/agentscope-a2a-spring-boot-starter -am -Dtest=AgentscopeA2aAutoConfigurationTest,A2aJsonRpcControllerTest -DfailIfNoTests=false test`

PR #2577, opened 2026-08-06
feat(a2a): expose fine-grained agent event stream

### LINKED ISSUE #2552: [Feature]: **Title**: `BaseReActAgentRunner.stream()` returns deprecated coarse-grained `Flux<Event>` and does not surface the fine-grained `AgentEvent` stream  **Labels**: `bug` / `enhancement` / `a2a`
## Description

The A2A server's `BaseReActAgentRunner.stream(...)` currently delegates to the **deprecated** `ReActAgent.stream(List<Msg>)` which returns `Flux<io.agentscope.core.agent.Event>` (the coarse-grained v1 event type). As a result, A2A sub-agents cannot emit / consume the fine-grained `AgentEvent` hierarchy (`Flux<io.agentscope.core.event.AgentEvent>`) that covers the full agent lifecycle (reasoning chunks, tool calls/results, HITL `RequireUserConfirmEvent`, `AgentResultEvent`, etc.).

### Location

`agentscope-extensions/.../a2a/.../server/executor/runner/BaseReActAgentRunner.java`

```java
@Override
public Flux<Event> stream(List<Msg> requestMessages, AgentRequestOptions options) {
    ...
    ReActAgent agent = buildReActAgent();
    agentCache.put(options.getTaskId(), agent);
    return agent.stream(requestMessages)                     // <-- deprecated coarse-grained API
            .doFinally(signal -> agentCache.remove(options.getTaskId()));
}
```

`ReActAgent` itself already provides the replacement API:

- `Flux<Event> stream(...)` — `@Deprecated(since = "2.0.0", forRemoval = true)`
- `Flux<AgentEvent> streamEvents(List<Msg>)` / `streamEvents(List<Msg>, RuntimeContext)` — the recommended fine-grained stream

### Impact / Why it matters

1. **Lost observability & control**: `AgentEvent` carries the full lifecycle (reasoning chunks, tool invocations, `RequireUserConfirmEvent` for human-in-the-loop, final `AgentResultEvent`). With only `Flux<Event>`, A2A handlers (e.g. `AgentScopeAgentExecutor`'s streaming/blocking `FluxEventHandler`) cannot distinguish event semantics reliably and must rely on the legacy `EventType` mapping.
2. **HITL breakage**: When a tool requires confirmation, the fine-grained `RequireUserConfirmEvent` is the only typed signal. The coarse `Event` stream does not expose it cleanly, so A2A callers cannot implement human-in-the-loop approval flows properly.
3. **Forward-compat**: `stream(...)` is marked `forRemoval = true`, so this runner will break once it is removed.

### Proposed fix

Make `BaseReActAgentRunner` (or its `AgentRunner` contract) support `AgentEvent`, e.g.:

```java
@Override
public Flux<AgentEvent> streamEvents(List<Msg> requestMessages, AgentRequestOptions options) {
    if (agentCache.containsKey(options.getTaskId())) {
        throw new IllegalStateException("Agent already exists for taskId: " + options.getTaskId());
    }
    ReActAgent agent = buildReActAgent();
    agentCache.put(options.getTaskId(), agent);
    return agent.streamEvents(requestMessages)                 // fine-grained, full lifecycle
            .doFinally(signal -> agentCache.remove(options.getTaskId()));
}
```

and optionally keep a backward-compatible `stream(...)` that adapts `AgentEvent` → `Event` for callers still on the v1 API.

### Environment

- agentscope-java (agentscope-examples/boba-tea-shop A2A setup)
- ReActAgent `streamEvents` available since 2.0.0; `stream(...)` deprecated for removal.

### Reproduction (conceptual)

1. Implement an `AgentRunner` extending `BaseReActAgentRunner`.
2. Subscribe to `runner.stream(msgs, options)`.
3. Observe that emitted elements are `io.agentscope.core.agent.Event`, never `io.agentscope.core.event.AgentEvent` — so lifecycle/HITL events are unavailable to the A2A executor.

mikemikimike:
## Summary
- Add `AgentRunner.streamEvents(...)` to expose fine-grained `AgentEvent` streaming for A2A runners.
- Reuse the existing cached-agent lifecycle for legacy and fine-grained streams, including cleanup on completion and error.
- Keep the existing `stream(...)` API for backward compatibility and add regression coverage.

Closes #2552.

## Tests
- `mvn -pl agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server -am clean verify`
- `mvn -pl agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server -am -Dtest=AgentRunnerTest,AgentScopeAgentExecutorTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `git diff --check`


zcw888:
> PR #2577 added the `AgentRunner.streamEvents()` capability (fine-grained `Flux<AgentEvent>`) and implemented it in `BaseReActAgentRunner`, but the fix looks incomplete: the actual A2A server call chain still uses the legacy coarse-grained stream.
>
> In `AgentScopeAgentExecutor.execute()` (agentscope-extensions-a2a-server), line 101 still calls `agentRunner.stream(...)` returning `Flux<Event>`, and the whole handler chain (`BaseFluxEventHandler`, `BlockingFluxEventHandler`, `StreamingFluxEventHandler`) consumes `Event`. `AgentRunner.streamEvents()` is only exercised by `AgentRunnerTest` and is never consumed by any executor, so A2A clients still cannot receive fine-grained `AgentEvent`s — this is exactly what #2552 asks for.
>
> To fully close this issue, `AgentScopeAgentExecutor` (and its `FluxEventHandler` implementations / tests) should be switched to consume `streamEvents()` / `AgentEvent`, ideally with a backward-compatible adapter so the existing `Event` consumers keep working. Could a follow-up PR cover the executor-side wiring?

oss-maintainer (commented):
### Summary
This PR modifies 3 files with 254 lines of diff.

### Observations
- Files changed: 3
- CLA: ✅ Signed

### Suggestions

---

mikemikimike:
Addressed the executor-side wiring noted in the review comment:

- `AgentScopeAgentExecutor` now prefers `streamEvents()` and routes fine-grained `AgentEvent` values through the blocking and streaming handlers.
- Legacy custom runners remain supported through an adapter/fallback to `stream()`.
- Added regression coverage for fine-grained execution and legacy fallback.

Validation at `489f67d`:
- `mvn -pl agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server -am clean verify` ✅ (core: 2,257 tests; A2A client: 109; A2A server: 138; 8 existing core skips)
- `git diff --check` ✅

The previous required Ubuntu run failed in the unrelated `agentscope-harness` module; the focused A2A reactor passed locally.

mikemikimike:
I rechecked this branch against the current base commit cdf1e5a. Upstream commit d588528 (merged as PR #2222) already migrates AgentRunner, BaseReActAgentRunner, and AgentScopeAgentExecutor to the fine-grained AgentEvent/streamEvents API described by issue #2552. The extra compatibility layer in this PR reintroduces the deprecated stream() method as a required abstract method; because the previous AgentRunner contract only required streamEvents(), that change can break existing custom AgentRunner implementations. The A2A module clean verify passes locally, but the current branch is duplicating the upstream migration and its Codecov patch check is still red. Could a maintainer please review whether this PR is superseded by #2222 and should be closed or reduced to any remaining delta?

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server/src/main/java/io/agentscope/core/a2a/server/executor/runner/AgentRunner.java
```diff
@@ -49,13 +51,31 @@ public interface AgentRunner {
...
+     * <p>Implementations that support the v2 event model should override this method. Existing
+     * custom runners remain source-compatible and can continue serving the legacy
+     * {@link #stream(List, AgentRequestOptions)} contract.
+     *
+     * @param requestMessages the messages from a2a client
+     * @param options the options for agent request, such as {@code taskId}, {@code sessionId} or {@code userId} of this request
+     * @return Flux of fine-grained events emitted during execution, or a Flux that terminates with
+     *     {@link UnsupportedOperationException} when the runner does not support fine-grained events
+     */
+    default Flux<AgentEvent> streamEvents(List<Msg> requestMessages, AgentRequestOptions options) {
```
oss-maintainer, line 74: `UnsupportedOperationException` is doing double duty here: it is the *capability sentinel* for not-yet-migrated runners and simultaneously a plausible genuine failure from a migrated runner's agent pipeline (unsupported modality, an SPI that refuses an operation, a codec path). The executor's `onErrorResume(UnsupportedOperationException.class, ...)` cannot tell the two apart before the first `onNext`, and if it mistakes a real error for "not supported" it re-runs the request through `stream()` — duplicate model calls, duplicate tool side effects, and the original error disappears from the response.

Suggest making the capability explicit instead of inferring it from an exception type: either a dedicated marker (e.g. `UnsupportedAgentEventStreamException extends UnsupportedOperationException` thrown only by this default), or a `default boolean supportsFineGrainedEvents() { return false; }` probe that `AgentScopeAgentExecutor.streamAgentEvents` checks up front. The executor's `emitted` flag only narrows the window, it does not close it.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server/src/main/java/io/agentscope/core/a2a/server/executor/runner/BaseReActAgentRunner.java
```diff
@@ -48,16 +50,24 @@ public String getAgentDescription() {
...
+    private <T> Flux<T> streamWithAgent(
+            String taskId, Function<ReActAgent, Flux<T>> streamFunction) {
+        if (agentCache.containsKey(taskId)) {
+            throw new IllegalStateException("Agent already exists for taskId: " + taskId);
         }
         ReActAgent agent = buildReActAgent();
-        agentCache.put(options.getTaskId(), agent);
-        return agent.streamEvents(requestMessages)
-                .doFinally(signal -> agentCache.remove(options.getTaskId()));
+        agentCache.put(taskId, agent);
```
oss-maintainer, line 69: `streamWithAgent` registers the agent in `agentCache` at *assembly* time but only unregisters in `doFinally`, i.e. at subscribe-terminal time. The fallback path added in this PR can now assemble two streams for the same `taskId` in one request: `streamEvents(...)` is assembled (entry created), it terminates, and `onErrorResume` subscribes `stream(...)` — whose assembly can race the upstream `doFinally` cleanup and trip `IllegalStateException("Agent already exists for taskId: ...")` at line 65. The never-subscribed case is worse: the `fineGrainedStream == null` branch and any assembly-time throw leave the entry in the cache permanently, after which every later request for that taskId fails.

Could the check + `put` move inside `Flux.defer { ... }` (or `doOnSubscribe`) so registration happens exactly when the stream is subscribed and is always paired with the `doFinally` removal? That also makes the idempotency guarantee independent of the new fallback path.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server/src/main/java/io/agentscope/core/a2a/server/executor/AgentScopeAgentExecutor.java
```diff
@@ -318,13 +395,20 @@ void doOnError(Throwable t) {
...
         protected boolean isNoResponseEvent(AgentEvent output) {
-            return !requiredEventTypes.contains(output.getType());
+            if (!requiredEventTypes.contains(output.getType())) {
+                return true;
+            }
+            if (!(output instanceof LegacyAgentEvent legacyEvent)
+                    || !legacyEvent.legacyEvent.isLast()) {
+                return false;
+            }
+            return Objects.equals(lastLegacyEventMsgId, legacyEvent.legacyEvent.getMessageId());
```
oss-maintainer, line 411: Dedup compares the incoming legacy message id against the *single* previously-seen id (`lastLegacyEventMsgId`, written at line 374). If a legacy runner interleaves message ids (chunk `X`, chunk `Y`, terminal `X`), the terminal `X` no longer matches `lastLegacyEventMsgId` (`Y`), so the completed message is streamed again after the deltas — duplicated text at the A2A client. Conversely, if `REASONING` and `SUMMARY` share one reply id, a terminal event can be suppressed even though it carries the authoritative final text.

An id-keyed set of "already streamed non-terminal" ids (rather than the last id only) would make the rule order-independent. A test with two interleaved ids would pin it.

oss-maintainer (commented):
### Summary
Delta re-review after the 8 author commits since `59a23db9`. The compatibility story is now much better: `streamEvents` gained a `default`, `stream` is back for legacy runners, `BaseReActAgentRunner` factors both through one `streamWithAgent` helper, and the executor wraps legacy `Event`s into `AgentEvent`s so the handler path stays single-model. Cache/error coverage is solid (`AgentRunnerTest` +5 cases, `AgentScopeAgentExecutorTest` +3). What is left is about *how* the fallback is triggered and *when* state is registered/suppressed — three inline points below.

### Verdict
**COMMENT** — the sentinel-vs-real-error ambiguity (`AgentRunner.java:74`) is the one I would like resolved or explicitly accepted; the assembly-time cache registration (`BaseReActAgentRunner.java:69`) is the riskier of the two mechanically. CLA signed.

oss-maintainer:
Two follow-ups to go with my review, one of them raised by @mikemikimike in the last comment and worth an explicit :arrow_up: for maintainers.

**1. This PR is largely superseded by #2222 (scope question for a maintainer).** On `main`, `AgentRunner` already declares only `Flux<AgentEvent> streamEvents(...)` as abstract, and `AgentScopeAgentExecutor` / `BaseReActAgentRunner` are already migrated to the fine-grained `AgentEvent` model. The remaining delta here is therefore the *back-compat shim* for the deprecated v1 `stream()` path, not the migration itself. If that shim is judged worth keeping, see point 2; if not, closing in favour of #2222 looks reasonable.

**2. The shim as written is source-breaking in the opposite direction.** Making `stream(...)` abstract again (AgentRunner.java:60) while `streamEvents(...)` becomes a `default` (line 74) means any custom `AgentRunner` implemented against the *current* contract — i.e. one that overrides only `streamEvents`, which is all `main` requires — no longer compiles. The compatibility goal is right; the direction is inverted. Defaulting both sides (with `stream` defaulting to an error signal) preserves both contracts, and a dedicated marker exception instead of a bare `UnsupportedOperationException` keeps "this runner has not migrated" distinguishable from "this runner failed before its first event" — the latter currently re-executes the request through `stream()`, with duplicate model calls and tool side effects.

No action needed from the bot side; flagging for maintainer judgement on scope.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server/src/test/java/io/agentscope/core/a2a/server/executor/runner/AgentRunnerTest.java
```diff
@@ -233,7 +368,7 @@ void testStopAgentAndRemoveFromCache() {
         Flux<AgentEvent> mockFlux = mock(Flux.class);
         when(mockAgent.streamEvents(messages)).thenReturn(mockFlux);
         when(mockFlux.doFinally(any())).thenReturn(mockFlux);
-        runner.streamEvents(messages, requestOptions);
+        runner.streamEvents(messages, requestOptions).subscribe();
```
oss-maintainer, line 371: Test vacuity introduced by the deferral change: the sibling tests were fixed by adding `blockLast` (lines 313/332/354), and `.subscribe()` was added to the first call on this line, but the re-stream check at the end of the test (`Flux<AgentEvent> result = runner.streamEvents(messages, requestOptions);`, line 377) is still only assembled, never subscribed. Now that registration is deferred to subscription time, that call cannot fail even if `stop()` left the entry in `agentCache` - pre-delta the assembly-time `containsKey` is what made the assert meaningful. Restub `streamEvents` to `Flux.empty()` after `runner.stop(taskId)` and `blockLast` the second call so the test actually proves the entry was released.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server/src/main/java/io/agentscope/core/a2a/server/executor/runner/UnsupportedAgentEventStreamException.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.agentscope.core.a2a.server.executor.runner;
+
+/** Signals that an {@link AgentRunner} has not implemented fine-grained event streaming. */
+public final class UnsupportedAgentEventStreamException extends UnsupportedOperationException {
```
oss-maintainer, line 20: Compatibility narrowing worth a changelog line: the documented opt-out signal changed from plain `UnsupportedOperationException` to this marker, and the executor's synchronous catch was narrowed to the marker type (`AgentScopeAgentExecutor.java:190`). Catching code is safe (the marker is a subclass), but a third-party runner that manually threw or returned `Flux.error(new UnsupportedOperationException(...))` from `streamEvents` per the previous javadoc no longer triggers the legacy fallback - the request now fails with the raw error. Acceptable while 2.0 is unreleased; please state it in the migration notes and document on `AgentRunner.streamEvents` that only `UnsupportedAgentEventStreamException` (or not overriding the method) selects the fallback.

oss-maintainer (approved):
### Summary
Delta re-review at `d8e23bf5` (previous pass at `86f870a0`). The single commit since then does exactly what the last round asked for, and all three open findings are resolved:

- **F1 capability sentinel (`AgentRunner.java:81`) - resolved.** The `streamEvents` default now fails with a dedicated `UnsupportedAgentEventStreamException` (a final subclass of `UnsupportedOperationException`), and the executor catches only that marker, both synchronously (`AgentScopeAgentExecutor.java:190-195`) and via `onErrorResume` (177-189), with an emitted-flag guard (174-182) so the fallback cannot start after any `onNext`. A genuine `UnsupportedOperationException` from a migrated runner's pipeline now propagates instead of causing a second execution. `testDoesNotFallBackForAgentUnsupportedOperation`, `testDoesNotFallBackAfterFineGrainedOutput` and `testFallsBackForSynchronousCapabilityMarker` pin all three states.
- **F2 assembly-time cache registration (`BaseReActAgentRunner.java:66`) - resolved.** `streamWithAgent` is wrapped in `Flux.defer`, so the agent is built and registered only at subscription time through `agentCache.compute` (69-79); an existing mapping now raises `IllegalStateException` instead of silently rebuilding (80-84), unregistration uses the two-arg conditional `remove(taskId, agent)` (88) so a stale `doFinally` can no longer evict a newer registration for the same taskId, and a synchronous stream-function failure releases the entry before surfacing the error (85-92). A marker-failing fine-grained stream now registers nothing, so the fallback path can no longer double-register or race the upstream `doFinally`.
- **F3 legacy dedup (`AgentScopeAgentExecutor.java:415`) - resolved.** Dedup records every legacy chunk that actually produced a response as a `LegacyEventKey(EventType, messageId)` (372-379) and suppresses an `isLast` event only when that exact pair was already streamed, so interleaved message ids no longer re-emit a completed message, and the old over-suppression (an `AGENT_RESULT` clobbering the single remembered id) is gone. `testDeduplicatesInterleavedLegacyChunks` is a genuine regression test for the previous logic.

### Verdict
**APPROVE** - the reactive semantics are unambiguous now, the cached-agent lifecycle is atomic, and the delta is backed by targeted tests. Two non-blocking notes are pinned inline: one assertion lost its teeth when registration became lazy (`AgentRunnerTest.java:371`), and the marker-type change deserves a migration note (`UnsupportedAgentEventStreamException.java:20`).

CI is green on this head (Check License, Check Module Sync, ubuntu + windows builds, codecov) and CLA is signed; `mergeState=BLOCKED` reflects the missing-review gate only.

---

mikemikimike:
Addressed the remaining review follow-ups in commit `b5a58287`:

- `AgentRunnerTest` now subscribes to the second stream with `blockLast()`, so lazy re-registration after `stop()` is actually exercised.
- Added migration guidance for `UnsupportedAgentEventStreamException` versus a plain `UnsupportedOperationException` in the `AgentRunner` Javadoc and v2 English/Chinese changelogs.

Validation:
- Focused A2A tests: 44 passed.
- `mvn -pl agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server -am clean verify`: core 2,323 tests (8 existing skips), A2A client 111, A2A server 146.
- GitHub checks: 7/7 passed, including Ubuntu and Windows builds.
- `git diff --check` passed.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-server/src/main/java/io/agentscope/core/a2a/server/executor/runner/AgentRunner.java
```diff
@@ -49,13 +51,39 @@ public interface AgentRunner {
...
+    }
+
+    /**
+     * Start to handle an agent request with fine-grained lifecycle events.
+     *
+     * <p>Implementations that support the v2 event model should override this method. Existing
+     * custom runners remain source-compatible whether they implement this method or only the
+     * legacy {@link #stream(List, AgentRequestOptions)} contract.
+     * The A2A executor selects the legacy stream only when this default implementation is used or
+     * the returned Flux terminates with {@link UnsupportedAgentEventStreamException}; a plain
```
oss-maintainer: **[Info]** Minor accuracy gap in the new fallback contract sentence. `AgentScopeAgentExecutor.streamAgentEvents(...)` also guards on whether anything has already gone downstream (`emitted` flag + `onErrorResume(UnsupportedAgentEventStreamException.class, ...)`): if the fine-grained stream emits at least one element and *then* fails with the marker, the error is propagated instead of re-running through the legacy `stream(...)`. Worth saying "...terminates with `UnsupportedAgentEventStreamException` **before any element is emitted**", otherwise a custom-runner author could reasonably assume late fallback is available. Not blocking — the behavior itself is the right one (falling back after partial output would duplicate text/tool results).

docs/v2/en/docs/change-log.md
```diff
@@ -207,6 +207,8 @@ Detail → [Message & Event](building-blocks/message-and-event.md)
 
 Python 2.0's `agent.reply_stream()` exposes a single streaming signature (`AsyncGenerator[AgentEvent, None]`) that maps directly to Java's fine-grained `io.agentscope.core.event.AgentEvent` hierarchy. To match it, the coarse-grained `Flux<Event> stream(...)` API on the Java side is `@Deprecated` as of 2.0.0:
 
+For A2A `AgentRunner` implementations, leave `streamEvents(...)` unoverridden or return `UnsupportedAgentEventStreamException` when the fine-grained stream is unavailable to select the legacy `stream(...)` fallback. A plain `UnsupportedOperationException` from an overridden `streamEvents(...)` is treated as an execution failure and is not retried through the legacy stream.
```
oss-maintainer: **[Info]** This A2A paragraph is inserted between the lead-in sentence ("... is `@Deprecated` as of 2.0.0:") and the bullet list it introduces, so the colon now reads as if it introduces the A2A note. Same layout in `docs/v2/zh/docs/change-log.md`. Suggest moving the note below the list (or as a `**Note**` block after it) so the deprecation list stays attached to its intro. Docs only, not blocking.

oss-maintainer (approved):
### Summary
Delta re-review at `b5a58287` (previous pass approved `d8e23bf5`). The single new commit is exactly the follow-up the last round asked for and touches no production behavior:

- `AgentRunner.streamEvents(...)` javadoc now states the legacy-fallback contract explicitly (default implementation / marker exception vs. a real `UnsupportedOperationException`).
- `AgentRunnerTest.testStopAgentAndRemoveFromCache` now subscribes to the second stream (`blockLast()`) instead of asserting on an un-subscribed `Flux`, so the "agent was removed from the cache" claim is actually exercised.
- Both `change-log.md` copies (en + zh) document the same rule.

Implementation side is unchanged and still correct: `AgentScopeAgentExecutor.streamAgentEvents(...)` narrows both the eager `catch` and the `onErrorResume` to the dedicated `UnsupportedAgentEventStreamException` subclass, so a genuine `UnsupportedOperationException` from the agent pipeline propagates rather than silently re-executing the agent through the legacy path. CLA signed, 6/6 checks green, `mergeable=true` / `clean` → re-approving on the current head.

### Findings
- **[Info]** `AgentRunner.java:74` — the new sentence omits the "before any element is emitted" guard that the executor really applies (anchored inline).
- **[Info]** `docs/v2/en/docs/change-log.md:210` — the A2A note is inserted between the colon lead-in and the deprecation bullet list; same layout in the zh copy (anchored inline).

Both are documentation wording nits, non-blocking.

---

oss-maintainer (approved):
### Summary

Delta re-review at `4c5e1a3f` (previous pass approved `b5a58287`). The single new commit is a javadoc/documentation clarification of the A2A stream-fallback contract, and it is accurate against the shipped code:

- `AgentRunner.streamEvents(...)` now states the legacy fallback is selected only when the default implementation is used **or** the stream fails with `UnsupportedAgentEventStreamException` **before any element is emitted**. That is exactly what `AgentScopeAgentExecutor.streamAgentEvents(...)` does: the `AtomicBoolean emitted` guard re-emits the error instead of falling back once anything went downstream, so a mid-stream failure can never cause a second agent execution.
- The A2A paragraph in `docs/v2/en/docs/change-log.md` and `docs/v2/zh/docs/change-log.md` moved below the `forRemoval` bullet list; wording is unchanged and en/zh stay in sync.

No production behavior is touched. CI on this head is green (Linux + Windows build, license, module sync, codecov, validate) and `license/cla` is signed. Re-approving — nothing further to ask.

---

Issue #2601, opened 2026-08-07
E2bCodec.JSON mode fails on every successful command due to proto3 default-value semantics


zylgq:
## Environment

| 项 | 值 |
|---|---|
| agentscope-java | 2.0.0-RC5 |
| 模块 | `agentscope-extensions-sandbox-e2b` |
| 受影响类 | `E2bEnvdProcessClient` |
| JDK | 17 |
| 触发条件 | `E2bSandboxClientOptions.setCodec(E2bCodec.JSON)` |

## Summary

When `E2bCodec.JSON` is selected, **any command that exits successfully (exit code 0) throws**:

```
io.agentscope.harness.agent.sandbox.SandboxException$ExecException:
    Command exited with code -2147483648:
```

`-2147483648` is `Integer.MIN_VALUE`, the initial value of the `exit` variable in `drainStartStream()` — meaning no `ProcessEvent.end` frame was ever parsed.

Counter-intuitively, commands that **fail** (non-zero exit code) work correctly.

## Root cause

proto3 semantics: for scalar fields (int32, bool, string...), when the value equals the default (0 for int32), the field is treated as *unset*:

- `message.getAllFields()` does **not** include it
- `message.hasField(field)` returns **false**

`parseJsonStartResponse()` uses `getAllFields().isEmpty()` to decide whether a message has content:

```java
// E2bEnvdProcessClient.parseJsonStartResponse()
if (exitCodeNode.canConvertToInt()) {
    endBuilder.setField(exitCodeField, exitCodeNode.intValue());   // sets exit_code = 0
}
if (!endBuilder.getAllFields().isEmpty()) {        // ← false when exit_code == 0
    event.setField(processEventDesc.findFieldByName("end"), endBuilder.build());
}                                                  // → end event silently dropped

if (!event.getAllFields().isEmpty()) {            // ← also false, cascading
    response.setField(startResponseDesc.findFieldByName("event"), event.build());
}                                                  // → event dropped too
```

Then in `drainStartStream()`:

```java
DynamicMessage sr = parseStartResponseFrame(data);
if (!sr.hasField(srEventF)) {
    continue;                    // ← frame skipped entirely
}
```

And a second instance of the same class of bug:

```java
Descriptors.FieldDescriptor ec = end.getDescriptorForType().findFieldByName("exit_code");
if (end.hasField(ec)) {          // ← also false when exit_code == 0
    exit = ...;
}
```

### Failure chain

```
[...]

PR #2609, opened 2026-08-07
fix(sandbox): preserve zero exit code in E2B JSON stream

### LINKED ISSUE #2603: E2bCodec.JSON mode fails on every successful command due to proto3 default-value semantics
## Environment

| 项 | 值 |
|---|---|
| agentscope-java | 2.0.0-RC5 |
| 模块 | `agentscope-extensions-sandbox-e2b` |
| 受影响类 | `E2bEnvdProcessClient` |
| JDK | 17 |
| 触发条件 | `E2bSandboxClientOptions.setCodec(E2bCodec.JSON)` |

## Summary

When `E2bCodec.JSON` is selected, **any command that exits successfully (exit code 0) throws**:

```
io.agentscope.harness.agent.sandbox.SandboxException$ExecException:
    Command exited with code -2147483648:
```

`-2147483648` is `Integer.MIN_VALUE`, the initial value of the `exit` variable in `drainStartStream()` — meaning no `ProcessEvent.end` frame was ever parsed.

Counter-intuitively, commands that **fail** (non-zero exit code) work correctly.

## Root cause

proto3 semantics: for scalar fields (int32, bool, string...), when the value equals the default (0 for int32), the field is treated as *unset*:

- `message.getAllFields()` does **not** include it
- `message.hasField(field)` returns **false**

`parseJsonStartResponse()` uses `getAllFields().isEmpty()` to decide whether a message has content:

```java
// E2bEnvdProcessClient.parseJsonStartResponse()
if (exitCodeNode.canConvertToInt()) {
    endBuilder.setField(exitCodeField, exitCodeNode.intValue());   // sets exit_code = 0
}
if (!endBuilder.getAllFields().isEmpty()) {        // ← false when exit_code == 0
    event.setField(processEventDesc.findFieldByName("end"), endBuilder.build());
}                                                  // → end event silently dropped

if (!event.getAllFields().isEmpty()) {            // ← also false, cascading
    response.setField(startResponseDesc.findFieldByName("event"), event.build());
}                                                  // → event dropped too
```

Then in `drainStartStream()`:

```java
DynamicMessage sr = parseStartResponseFrame(data);
if (!sr.hasField(srEventF)) {
    continue;                    // ← frame skipped entirely
}
```

And a second instance of the same class of bug:

```java
Descriptors.FieldDescriptor ec = end.getDescriptorForType().findFieldByName("exit_code");
if (end.hasField(ec)) {          // ← also false when exit_code == 0
    exit = ...;
}
```

### Failure chain

```
command succeeds → exitCode = 0
    → setField(exit_code, 0) leaves getAllFields() empty (proto3 default-value semantics)
    → end event not attached to event
    → event not attached to response
    → drainStartStream sees no "event" field → continue → frame skipped
    → loop ends with exit == Integer.MIN_VALUE
    → ExecResult.ok() == false → ExecException thrown
```

## Why this has gone unnoticed

`E2bCodec.PROTO` is the default. The protobuf path uses `DynamicMessage.parseFrom()` and never touches `parseJsonStartResponse()`, so the bug is invisible unless JSON codec is explicitly selected.

JSON codec becomes mandatory when the E2B-compatible backend does not accept `application/connect+proto`. For example, Alibaba Cloud FC Agent Sandbox (envd 0.5.2) returns:

```
HTTP 400: invalid character '/' after top-level value
```

for protobuf-encoded requests, forcing JSON codec — at which point the bug blocks all usage.

## Reproduction

```java
E2bSandboxClientOptions options = new E2bSandboxClientOptions();
options.setApiKey(System.getenv("E2B_API_KEY"));
options.setApiBaseUrl("<any E2B-compatible endpoint>");
options.setDomain("<matching domain>");
options.setTemplateId("code-interpreter-v1");
options.setCodec(E2bCodec.JSON);          // ← the trigger

Sandbox sandbox = new E2bSandboxClient(options, null)
        .create(new WorkspaceSpec(), null, options);
sandbox.start();   // fails here: doSetupWorkspace() runs "mkdir -p ..." which exits 0
```

`start()` throws before any user command runs, because `doSetupWorkspace()` issues `mkdir -p {workspaceRoot}`, which exits 0.

### Actual server response (captured from Alibaba Cloud FC, Content-Type `application/connect+json`)

Connect Protocol frames, 1-byte flags + 4-byte big-endian length + payload:

```
frame 1  flags=00  len=0x1d  {"event":{"start":{"pid":2}}}
frame 2  flags=00  len=0x28  {"event":{"data":{"stdout":"aGVsbG8K"}}}
frame 3  flags=00  len=0x40  {"event":{"end":{"exitCode":0,"exited":true,"status":"exited"}}}
frame 4  flags=02  len=0x02  {}
```

The response is well-formed and spec-compliant. Frame 3 carries `exitCode: 0`, which is exactly what triggers the bug.

## Fix

Three changes, all in `E2bEnvdProcessClient`:

```java
// 1 & 2 — parseJsonStartResponse(): drop the getAllFields().isEmpty() guards.
//         If the JSON contains the node, the field must be set unconditionally.

// before:
//   if (!endBuilder.getAllFields().isEmpty()) {
//       event.setField(processEventDesc.findFieldByName("end"), endBuilder.build());
//   }
// after:
event.setField(processEventDesc.findFieldByName("end"), endBuilder.build());

// before:
//   if (!event.getAllFields().isEmpty()) {
//       response.setField(startResponseDesc.findFieldByName("event"), event.build());
//   }
// after:
response.setField(startResponseDesc.findFieldByName("event"), event.build());


// 3 — drainStartStream(): hasField() → null check.
//     For scalar fields, getField() returns the default value even when "unset".

// before:
//   if (end.hasField(ec)) {
// after:
if (ec != null) {
    Object v = end.getField(ec);
    exit = v instanceof Integer ? (Integer) v : ((Long) v).intValue();
}
```

Verified end-to-end against Alibaba Cloud FC Agent Sandbox after the fix:

```
Sandbox created + started in 2123 ms
echo hello        → exit=0, stdout=[hello]      43 ms
python3 --version → stdout=[Python 3.13.13]     57 ms
Sandbox killed                                  52 ms
```

ump45nose:
## AgentScope-Java Version

2.0.1-SNAPSHOT

## Description

Fixes #2603.

In `E2bCodec.JSON` mode, an explicit `{"end":{"exitCode":0}}` frame was lost twice because proto3 does not include default scalar values in `getAllFields()` and reports no scalar presence through `hasField()`. As a result, successful commands kept the missing-exit sentinel and failed.

This change:

- attaches the JSON `end` event when an integer `exitCode` was explicitly parsed, including zero;
- reads the proto3 default scalar from the present end event instead of checking scalar presence;
- adds a focused regression test for a zero exit code.

Related duplicate reports: #2601, #2602.

Tested with:

```text
mvn -pl agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b -am -Dtest=E2bEnvdProcessClientTest -Dsurefire.failIfNoSpecifiedTests=false test
```

Result: 8 tests passed, including `jsonCodecPreservesZeroExitCode`.

## Checklist

- [x]  All targeted tests are passing
- [x]  Javadoc comments are complete and follow project conventions (no public API changes)
- [x]  Related documentation has been updated (not applicable; behavior fix with regression coverage)

### DIFF
--- a/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
+++ b/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
@@ -205,8 +205,7 @@ final class E2bEnvdProcessClient {
                     Descriptors.FieldDescriptor ec =
                             end.getDescriptorForType().findFieldByName("exit_code");
                     if (ec != null) {
-                        Object v = end.getField(ec);
-                        exit = v instanceof Integer ? (Integer) v : ((Long) v).intValue();
+                        exit = ((Number) end.getField(ec)).intValue();
                     }
                 }
             } catch (IOException e) {
--- a/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
+++ b/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
@@ -204,7 +204,7 @@ final class E2bEnvdProcessClient {
                     DynamicMessage end = (DynamicMessage) pe.getField(peEndF);
                     Descriptors.FieldDescriptor ec =
                             end.getDescriptorForType().findFieldByName("exit_code");
-                    if (end.hasField(ec)) {
+                    if (ec != null) {
                         Object v = end.getField(ec);
                         exit = v instanceof Integer ? (Integer) v : ((Long) v).intValue();
                     }
@@ -344,8 +344,6 @@ final class E2bEnvdProcessClient {
             JsonNode exitCodeNode = endNode.path("exitCode");
             if (exitCodeNode.canConvertToInt()) {
                 endBuilder.setField(exitCodeField, exitCodeNode.intValue());
-            }
-            if (!endBuilder.getAllFields().isEmpty()) {
                 event.setField(processEventDesc.findFieldByName("end"), endBuilder.build());
             }
         }
--- a/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/test/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClientTest.java
+++ b/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/test/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClientTest.java
@@ -147,6 +147,17 @@ class E2bEnvdProcessClientTest {
         assertEquals("", stderr.toString(StandardCharsets.UTF_8));
     }
 
+    @Test
+    void jsonCodecPreservesZeroExitCode() throws Exception {
+        E2bEnvdProcessClient client = new E2bEnvdProcessClient(options(E2bCodec.JSON));
+        ByteArrayOutputStream stdout = new ByteArrayOutputStream();
+        ByteArrayOutputStream stderr = new ByteArrayOutputStream();
+        int exit =
+                drainStartStream(client, connectFrame(responseJson(null, null, 0)), stdout, stderr);
+
+        assertEquals(0, exit);
+    }
+
     private static DynamicMessage dataResponse(
             E2bEnvdProcessClient client, String stdout, String stderr) {
         Descriptors.FileDescriptor fd = client.fileDescriptor();


AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes a proto3 default-value bug where an exitCode of 0 was silently dropped in the E2B JSON codec path. Root cause: DynamicMessage.hasField() returns false for proto3 non-optional scalar fields set to their default value, and getAllFields() omits default-valued fields. The fix is minimal, correct, and directly addresses issue #2603. A focused regression test is included. This also benefits the proto codec path where a binary-encoded EndEvent with exit_code=0 would also be missed.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes a proto3 default-value bug where an exitCode of 0 was silently dropped in the E2B JSON codec path. Root cause: DynamicMessage.hasField() returns false for proto3 non-optional scalar fields set to their default value, and getAllFields() omits default-valued fields. The fix is minimal, correct, and directly addresses issue #2603. A focused regression test is included. This also benefits the proto codec path where a binary-encoded EndEvent with exit_code=0 would also be missed.

Buktal (approved):
LGTM @jujn

jujn (approved):

chcodex:
`3e096a00` correctly fixes the `exitCode 0` loss caused by `hasField`/`getAllFields` (proto3 default-value omission), but has limitations:

1. **`EndEvent.error` is dropped** — `proto: EndEvent { sint32 exit_code=1; optional string error=4 }` is populated on signal-kill (`envd handler.go: ProcessEvent_EndEvent.Error=errMsg`) and consumed by the Python SDK `command_handle.py#L115` (`error=event.event.end.error`) into `stderr`. `2609` `parseJsonStartResponse:340-346` only handles `exitCode` and conditionally attaches `end`, so `{"end":{"error":"signal: 9"}}` yields empty `stderr` under the `JSON` codec, diverging from `2196` `540-549`/`264-269`.

2. **Conditional `attach end` diverges from `PROTO` semantics** — `2609` only does `event.setField(end)` when `canConvertToInt`, so `{"end":{}}` / `{"exitCode":"abc"}` / error-only `end` returns `MIN_VALUE` (sentinel), while `PROTO` returns `0` via `getField→0` for the same payloads. Covered by `2196` tests `exitCodeZeroDefaultWhenOmitted` / `jsonCodecNonIntegerExitCodeIsIgnored`.

3. **Test coverage** — only `jsonCodecPreservesZeroExitCode` (`{"exitCode":0}→0`) is added, leaving the empty/`error` cases uncovered.

PR #2640, opened 2026-08-10
feat(console): transcript overhaul - ordered content blocks, markdown, tool UX

### LINKED ISSUE #2596: [Bug]: Managed session console does not render session.error events - failed turns are silent in the chat UI
## Problem

When a managed session turn fails (e.g. model missing, provider key missing), the backend correctly records the failure but the console chat UI shows nothing.

- `SessionTurnRunner.failTurn()` (`agentscope-service/service-dataplane/.../SessionTurnRunner.java`) appends a `session.error` event (payload `{error: {type, code, message}}`) and sets the session status to `terminated`.
- The event is persisted (`dp.builder_session_event`) and returned by `GET /api/sessions/{id}/events`, but the chat component `agentscope-service/frontend/src/components/ChatPanel.tsx` only handles `user.message`, `agent.message`, `agent.turn_stub`, `agent.tool_use`, `agent.tool_result`, `session.requires_action`, `session.status_idle` — there is no branch for `session.error`, so the error is silently dropped on the client.
- Verified end-to-end: 12 events (incl. 3× `session.error`) returned by the API, but nothing rendered in the chat panel.

## Repro

1. Start agentscope-service locally
2. Create an agent without a model (or with a model whose provider key is unset)
3. Create a session and send a message → turn fails with `Failed to create model for id ...`
4. Chat panel shows no error; session badge shows terminated at most

## Expected

The chat UI displays the failure (code + message) as an error bubble, both on initial load and via SSE during a live turn.

## Environment

- agentscope-service frontend (React/Vite), ChatPanel.tsx + MessageBlock.tsx
- Turn failure path: SessionTurnRunner.runTurnAsync → failTurn

Aias00:
## Summary

Reworks the managed-session chat transcript around **one bubble per user question**, with text and tool calls rendered as ordered content blocks in chronological order.

**ChatPanel.tsx**
- Block-based message model (`ContentBlock[]`): `agent.message` / `agent.tool_use` / `agent.tool_result` append ordered blocks into the current turn bubble; `user.message` / `session.error` close it
- Turn bubbles stay open across `session.status_*` events — the backend emits `status_idle` between model iterations of one user question, which previously split one answer into multiple bubbles
- Auto-expand the latest assistant turn bubble (even when a follow-up user message sits after it)
- Renders `session.error` as a red bubble (fixes #2596)
- Fixed the send path: optimistic user bubble + SSE reconciliation now correctly closes the previous turn (the `seen-id` pre-dedupe was eating the streamed `user.message`, so the next reply appended to the old bubble)
- Removed vestigial `replyMsgIdRef` (it crashed sends after the block refactor)

**managedSessions.ts**
- SSE auto-reconnect with exponential backoff (2s → 30s cap, resets on progress), resuming from the last seen sequence via `getAfter` so no events are lost across plane restarts

**MessageBlock.tsx**
- Renders ordered content blocks; assistant text via `react-markdown` with `.md-text` styles (headings, lists, code blocks, tables, blockquote, images)

**ToolCallBlock.tsx**
- Input / Result sections with labels
- JSON pretty-printing + residual escape unescaping (`\\n` from double-encoded tool output)
- Grep-style `file:line:content` rows highlighted (relative-path prefix stripped)
- Line numbers for plain/file dumps; `prism-react-renderer` syntax highlighting for XML/JSON/Java/Bash/JS
- Tool card opens while the tool is running and collapses once the result lands

Dependencies added: `react-markdown`, `prism-react-renderer` (package.json only; `package-lock.json` was not regenerated locally — CI may want to run `npm install --package-lock`).

## Test plan

- [x] `vite build` succeeds
- [x] Verified via headless Chrome: sequential user messages produce separate assistant bubbles; multi-tool turns render as one bubble in event order; markdown renders; tool Input/Result display formatted content; SSE reconnect resumes from cursor
- [x] Send-path regression: user message → agent reply loop works end-to-end (DeepSeek model)

Supersedes #2598 (session.error rendering was folded into the block model).

### DIFF
--- a/agentscope-service/frontend/package.json
+++ b/agentscope-service/frontend/package.json
@@ -22,8 +22,10 @@
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
     "lucide-react": "^1.27.0",
+    "prism-react-renderer": "^2.4.1",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
+    "react-markdown": "^10.1.0",
     "react-router-dom": "^6.28.0",
     "tailwind-merge": "^3.6.0"
   },
--- a/agentscope-service/frontend/src/api/managedSessions.ts
+++ b/agentscope-service/frontend/src/api/managedSessions.ts
@@ -204,67 +204,93 @@ export interface EventStreamHandle {
   close: () => void;
 }
 
+const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
 /**
- * Subscribes to session events over SSE. Uses fetch (not native EventSource) so JWT Bearer auth works.
+ * Subscribes to session events over SSE. Uses fetch (not native EventSource) so JWT Bearer auth
+ * works. Reconnects automatically when the stream ends or fails, resuming from the last seen
+ * sequence via {@code getAfter} so no events are lost across plane restarts or network drops.
  */
 export function streamEvents(
   sessionId: string,
   onEvent: (event: SessionEvent) => void,
   onError?: (err: Error) => void,
-  options?: { eventDeltas?: string[]; after?: number },
+  options?: {
+    eventDeltas?: string[];
+    after?: number;
+    /** Called before each (re)connect to compute the resume cursor. */
+    getAfter?: () => number | undefined;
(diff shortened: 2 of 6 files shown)


Aias00:
image

jujn (approved):

PR #2643, opened 2026-08-10
fix(harness): target active session in HarnessAgent interrupt API

StevenZhang2002:
### Background

Issue #2610: `HarnessAgent.interrupt()` / `interrupt(Msg)` unconditionally resolved the default
session id (the agent name), so streams started via `streamEvents(List, RuntimeContext)` with a
custom `sessionId` could never be cancelled — the interrupt signal was always sent to the wrong
session slot.

### Purpose

Make the interrupt API target the session that is actually running:

1. The context-free overloads now prefer the active call's `RuntimeContext` (obtained from the
   delegate's current runtime context) when one exists, while keeping the legacy default-session
   fallback when no call is in flight.
2. New `interrupt(RuntimeContext)` / `interrupt(RuntimeContext, Msg)` overloads let callers target
   an explicit `(userId, sessionId)` slot. Because each caller supplies the session it wants to
   cancel, this is safe under concurrency (multiple sessions running in parallel).

### Changes

- `agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java`:
  - `interrupt()` and `interrupt(Msg)` now prefer the active call's runtime context when present,
    and fall back to the default session otherwise.
  - Added `interrupt(RuntimeContext ctx)` and `interrupt(RuntimeContext ctx, Msg msg)` overloads
    that delegate directly to the target session (with Javadoc documenting the issue context).
- `agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentInterruptTest.java`
  (new): a blocking model is used to keep a stream in flight while an interrupt is fired. Covers:
  - `interrupt(ctx, msg)` cancels a stream running under a custom `sessionId` → `GenerateReason.INTERRUPTED`
  - `interrupt(ctx)` cancels a custom-session stream without a message
  - context-free `interrupt(msg)` prefers the active call's custom session
  - context-free `interrupt(msg)` still works for streams under the default session (backward compat)
  - context-free `interrupt()` / `interrupt(msg)` with no active call fall back to the default
    session without throwing (backward compat)

### How to Test
All five test cases in `HarnessAgentInterruptTest` should pass; no existing tests are affected.

## Checklist

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
@@ -575,6 +575,7 @@ public class HarnessAgent implements Agent, AutoCloseable {
 
     @Override
     public void interrupt() {
+        // get the active runtime context
         RuntimeContext active = delegate.getRuntimeContext();
         if (active != null) {
             delegate.interrupt(active);
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
@@ -575,12 +575,47 @@ public class HarnessAgent implements Agent, AutoCloseable {
 
     @Override
     public void interrupt() {
-        delegate.interrupt();
+        RuntimeContext active = delegate.getRuntimeContext();
+        if (active != null) {
+            delegate.interrupt(active);
+        } else {
+            delegate.interrupt();
+        }
     }
 
     @Override
     public void interrupt(Msg msg) {
-        delegate.interrupt(msg);
+        RuntimeContext active = delegate.getRuntimeContext();
+        if (active != null) {
+            delegate.interrupt(active, msg);
+        } else {
+            delegate.interrupt(msg);
+        }
+    }
+
+    /**
+     * Interrupts the in-flight call identified by the given {@link RuntimeContext}.
+     *
+     * <p>Unlike {@link #interrupt()}, this targets the exact {@code (userId, sessionId)} slot
+     * carried by {@code ctx}, so it works for sessions started via
+     * {@link #streamEvents(List, RuntimeContext)} with a custom {@code sessionId} (see issue
+     * #2610). Safe under concurrency: each caller supplies the session it wants to cancel.
+     *
+     * @param ctx the runtime context identifying the session to interrupt
+     */
+    public void interrupt(RuntimeContext ctx) {
+        delegate.interrupt(ctx);
+    }
+
+    /**
+     * Interrupts the in-flight call identified by the given {@link RuntimeContext} with an
+     * associated user message.
+     *
+     * @param ctx the runtime context identifying the session to interrupt
+     * @param msg optional user message to attach to the interrupt signal
+     */
+    public void interrupt(RuntimeContext ctx, Msg msg) {
+        delegate.interrupt(ctx, msg);
     }
 
     // -----------------------------------------------------------------
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentInterruptTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentInterruptTest.java
@@ -1,0 +1,210 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.core.event.AgentResultEvent;
+import io.agentscope.core.message.GenerateReason;
+import io.agentscope.core.message.Msg;
+import io.agentscope.core.message.MsgRole;
+import io.agentscope.core.message.TextBlock;
+import io.agentscope.core.message.UserMessage;
+import io.agentscope.core.model.ChatModelBase;
+import io.agentscope.core.model.ChatResponse;
+import io.agentscope.core.model.GenerateOptions;
+import io.agentscope.core.model.ToolSchema;
+import io.agentscope.core.state.InMemoryAgentStateStore;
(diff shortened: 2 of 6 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -575,12 +575,48 @@ public String getDescription() {
 
     @Override
     public void interrupt() {
-        delegate.interrupt();
+        // get the active runtime context
+        RuntimeContext active = delegate.getRuntimeContext();
```
oss-maintainer, line 579: `delegate.getRuntimeContext()` is `ReActAgent.activeRc` — a single `volatile` field that every call activation overwrites (`this.activeRc = ctx` at call start, `= null` at call end). `AgentBase#getRuntimeContext()` spells out the consequence: *"Because the value is sourced from the agent's most-recently-activated scope, under concurrent calls on one instance this reflects the latest call"*.

So for the case this PR is fixing — several `streamEvents(msgs, ctx)` sessions sharing one `HarnessAgent` — the context-free `interrupt()` is still not deterministic: it cancels whichever session activated last, which may be a *different* user's session (previously it deterministically hit the builder default slot). It also degrades to the default slot when the in-flight call finishes in the window between reading the field and interrupting.

Suggested options:

1. Keep the no-arg / `interrupt(Msg)` overloads bound to the builder default slot (no behavior change) and make the new `interrupt(RuntimeContext[, Msg])` overloads the documented fix for #2610 — that is a purely additive, concurrency-safe contract.
2. If last-activated preference is really intended, say so explicitly in the javadoc and the new docs section ("targets the most recently started call, not necessarily yours"), and log the resolved `(userId, sessionId)` at debug so a mis-routed cancel is diagnosable.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -575,12 +575,48 @@ public String getDescription() {
...
+            delegate.interrupt(active);
+        } else {
+            delegate.interrupt();
+        }
     }
 
     @Override
     public void interrupt(Msg msg) {
-        delegate.interrupt(msg);
+        RuntimeContext active = delegate.getRuntimeContext();
```
oss-maintainer, line 589: Same active-context resolution duplicated in both context-free overloads. Extracting one private helper (e.g. `private RuntimeContext interruptTarget()` returning `activeRc`) keeps the two paths in sync and gives the chosen disambiguation rule a single place to document/change — see the note above on line 579.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -575,12 +575,48 @@ public String getDescription() {
...
+     *
+     * <p>Unlike {@link #interrupt()}, this targets the exact {@code (userId, sessionId)} slot
+     * carried by {@code ctx}, so it works for sessions started via
+     * {@link #streamEvents(List, RuntimeContext)} with a custom {@code sessionId} (see issue
+     * #2610). Safe under concurrency: each caller supplies the session it wants to cancel.
+     *
+     * @param ctx the runtime context identifying the session to interrupt
+     */
+    public void interrupt(RuntimeContext ctx) {
+        delegate.interrupt(ctx);
```
oss-maintainer, line 608: `delegate.interrupt(ctx)` treats a `null` ctx as "no session identity" and `ReActAgent#interrupt(RuntimeContext, Msg)` then falls back to `defaultSessionId`, so `interrupt((RuntimeContext) null)` silently cancels a session the caller never named. For a new public API it is better to fail fast — `Objects.requireNonNull(ctx, "ctx")` — and leave the null-to-default behaviour to the deprecated context-free overloads.

agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentInterruptTest.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    @Test
+    void interruptWithoutActiveCall_fallsBackToDefaultSessionWithoutThrowing() throws Exception {
+        Files.createDirectories(workspace);
+        try (HarnessAgent agent = buildAgent(new BlockingModel(), workspace)) {
+            // No in-flight call: the context-free overloads must keep their legacy fallback
+            // (interrupting the default session slot) instead of failing.
+            agent.interrupt();
```
oss-maintainer, line 206: This test only proves "no exception"; it does not pin the fallback contract it names. Asserting the signal actually landed on the default slot — e.g. `assertTrue(agent.getDelegate().getAgentState(null, agent.getDefaultSessionId()).interruptControl().isInterrupted())` — would make the regression guard meaningful.

oss-maintainer (commented):
### Summary
Routes `HarnessAgent`'s context-free `interrupt()` / `interrupt(Msg)` to the active call's `RuntimeContext`, and adds `interrupt(RuntimeContext[, Msg])` overloads so streams started with a custom `sessionId` can be cancelled (issue #2610). The new overloads are the right shape and the test coverage is real, but the context-free re-routing relies on `ReActAgent.activeRc`, which is a single agent-level field and therefore not a per-session handle under concurrency. That is the one point I would like resolved before this is approvable.

### Findings
- **[Warning]** `HarnessAgent.java:579` — `getRuntimeContext()` reflects the *most recently activated* call (per its own javadoc), so with several concurrent sessions the context-free `interrupt()` can target a different session than the caller intends; this replaces one wrong-session risk with another instead of only adding the explicit overloads.
- **[Warning]** `HarnessAgent.java:608` — a `null` ctx silently degrades to the default session slot inside `ReActAgent.interrupt(ctx, msg)`; a new public overload should fail fast instead.
- **[Info]** `HarnessAgent.java:589` — the same resolution logic is duplicated across both context-free overloads.
- **[Info]** `HarnessAgentInterruptTest.java:206` / `:129` — the fallback test asserts only "does not throw", and no test covers two concurrent sessions where only one must be cancelled.

### Suggestions
1. Land the two `RuntimeContext` overloads (additive and correct), and either keep the context-free overloads bound to the default slot or make "most recently activated" an explicit documented contract with a debug log of the resolved `(userId, sessionId)`.
2. Add `Objects.requireNonNull(ctx, "ctx")` in the new overloads so a null context cannot cancel an unrelated session.
3. Add a two-concurrent-session test: interrupt session A via `interrupt(ctxA)`, assert A ends with `GenerateReason.INTERRUPTED` while B completes normally — that is the regression #2610 actually describes.

---

agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentInterruptTest.java
```diff
@@ -0,0 +1,210 @@
...
+                    executor.submit(
+                            () ->
+                                    agent.streamEvents(List.of(userMsg("hello")), ctx)
+                                            .filter(e -> e instanceof AgentResultEvent)
+                                            .cast(AgentResultEvent.class)
+                                            .next()
+                                            .map(AgentResultEvent::getResult)
+                                            .block());
+            assertTrue(
+                    model.awaitCalled(5, TimeUnit.SECONDS),
```
oss-maintainer, line 129: Every case here runs exactly one in-flight stream, so the multi-session behaviour that motivated #2610 is untested. A test with two concurrent streams on different `sessionId`s, interrupting one via `interrupt(ctx)` and asserting the other still completes normally, is the case worth locking down — it is also the case that decides whether the `activeRc`-based routing above is acceptable.

PR #2661, opened 2026-08-11
feat(sandbox): DockerDaemon: docker-java direct to a remote daemon (avoiding docker-in-docker)

magic-dong:
## AgentScope-Java Version

## Description

## Checklist


magic-dong:
不考虑拓展支持一下嘛？主Agent程序一般公司也可能是docker部署，其他沙箱模式对于一般场景有点小重

oss-maintainer:
**CLA Not Signed**

@magic-dong please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.

---

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonSandbox.java
```diff
@@ -0,0 +1,622 @@
...
+            PortBinding[] bindings = new PortBinding[ports.length];
+            for (int i = 0; i < ports.length; i++) {
+                exposed[i] = ExposedPort.tcp(ports[i]);
+                bindings[i] = PortBinding.parse(ports[i] + ":" + ports[i]);
+            }
+            cmd.withExposedPorts(exposed);
+            hostConfig.withPortBindings(bindings);
+        }
+        String network = dockerState.getNetwork();
+        hostConfig.withNetworkMode(network == null ? "none" : network);
```
oss-maintainer, line 375: **[Warning]** `network` is passed through to `withNetworkMode(...)` verbatim, so a persisted/user-supplied value of `host` (or `container:<id>`) silently escapes sandbox network isolation — the whole point of the default `"none"` is defeated by one unvalidated string. Given this module is explicitly about running agent-generated commands against a *shared/remote* daemon, could you reject or require an opt-in for `host` / `container:*` modes (mirroring the "never trust persisted state blindly" stance already taken in `validatedDockerJavaMount`)? A allowlist of `none` / `bridge` / user-defined network names would be enough.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonSandbox.java
```diff
@@ -0,0 +1,622 @@
...
+    protected void doDestroyWorkspace() throws Exception {
+        String containerId = dockerState.getContainerId();
+        if (containerId == null || containerId.isBlank()) {
+            return;
+        }
+        try {
+            execInContainer(
+                    containerId,
+                    null,
+                    List.of("rm", "-rf", dockerState.getWorkspaceRoot()),
```
oss-maintainer, line 281: **[Warning]** `rm -rf <workspaceRoot>` is executed with a path taken straight from persisted state, while the mount path right next to it (`validatedDockerJavaMount`, line 444) is deliberately re-validated because "persisted state JSON ... never trust it blindly". The same argument applies here: a blank, relative, or `/` `workspaceRoot` in a hand-crafted state file turns workspace cleanup into a destructive `rm -rf /` inside the container. Please validate `workspaceRoot` (absolute, not `/`, sane depth) once when the state is loaded and reuse the validated value for both the cleanup exec and the `sh -c` execs.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonClientFactory.java
```diff
@@ -0,0 +1,98 @@
...
+     * version's env-inheritance behavior for {@code https://} daemon URLs.
+     *
+     * @param state sandbox state carrying the frozen daemon URL
+     * @return docker-java client config
+     */
+    static DefaultDockerClientConfig buildConfig(DockerDaemonSandboxState state) {
+        DefaultDockerClientConfig.Builder builder =
+                DefaultDockerClientConfig.createDefaultConfigBuilder();
+        if (state.getDaemonUrl() != null && !state.getDaemonUrl().isBlank()) {
+            builder.withDockerHost(state.getDaemonUrl());
```
oss-maintainer, line 76: **[Warning]** `resolveDaemonUrl(option, env)` (line 46) is only consulted by `DockerDaemonSandboxClient` when the state is first frozen; `buildConfig` here falls back to `DefaultDockerClientConfig.createDefaultConfigBuilder()`, which reads the ambient `DOCKER_HOST`/`DOCKER_TLS_VERIFY`/`DOCKER_CERT_PATH` of the JVM. On the restore path (`state.getDaemonUrl() == null`) that means a session can be re-attached to a *different daemon* than the one that created the container — and then `stopContainerCmd`/`removeContainerCmd`/`rm -rf` run against whichever host happens to be in the environment. Suggest failing fast when no daemon URL is recorded in the state (or resolving through `resolveDaemonUrl` consistently) so create and restore always target the same endpoint.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonClientFactory.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Builds docker-java {@link DockerClient} instances for a {@link DockerDaemonSandboxState}.
+ *
+ * <p>The client talks to the daemon over its Engine HTTP API via the zerodep transport
+ * ({@code docker-java-transport-zerodep}), so no docker CLI or local socket is required.
+ */
+public final class DockerDaemonClientFactory {
+
+    /** Fixed Engine API version to avoid daemon version drift. */
+    static final String API_VERSION = "1.45";
```
oss-maintainer, line 33: **[Info]** `API_VERSION = "1.45"` pins the request path to `/v1.45/...`. Daemons older than Docker Engine 26 reject a newer client version outright, so the very first call fails with a hard-to-read `client is newer than server (400)` rather than a version-negotiation fallback. Either drop the pin and let docker-java negotiate, or document the minimum supported daemon version (and surface it in the error) — the docs added in this PR only mention "no docker CLI required".

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonSandbox.java
```diff
@@ -0,0 +1,622 @@
...
+
+        TruncatingOutputStream stdout = new TruncatingOutputStream(stdoutMaxBytes);
+        TruncatingOutputStream stderr = new TruncatingOutputStream(OUTPUT_TRUNCATE_BYTES);
+        ExecStartResultCallback callback = new ExecStartResultCallback(stdout, stderr);
+        ExecStartResultCallback returned =
+                dockerClient().execStartCmd(execCreate.getId()).withDetach(false).exec(callback);
+
+        boolean finished;
+        try {
+            finished = returned.awaitCompletion(timeoutSeconds, TimeUnit.SECONDS);
```
oss-maintainer, line 559: **[Info]** On timeout the method throws `ExecTimeoutException` but the `ExecStartResultCallback` (and its zerodep HTTP connection) is never closed, and the remote exec keeps running — the javadoc already notes there is no exec-kill API. Under a stuck-command workload this leaks one held connection + one runaway process per timed-out exec against a shared daemon. A `try/finally { returned.close(); }` around the await, plus a note in the docs that a timed-out exec is not killed (and that callers may want `shutdown()` to force-remove the container), would make the failure mode bounded.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonSandbox.java
```diff
@@ -0,0 +1,622 @@
...
+        if (dockerState.getCpuCount() != null) {
+            hostConfig.withNanoCPUs(dockerState.getCpuCount() * 1_000_000_000L);
+        }
+        if (dockerState.getExposedPorts() != null && dockerState.getExposedPorts().length > 0) {
+            int[] ports = dockerState.getExposedPorts();
+            ExposedPort[] exposed = new ExposedPort[ports.length];
+            PortBinding[] bindings = new PortBinding[ports.length];
+            for (int i = 0; i < ports.length; i++) {
+                exposed[i] = ExposedPort.tcp(ports[i]);
+                bindings[i] = PortBinding.parse(ports[i] + ":" + ports[i]);
```
oss-maintainer, line 369: **[Info]** `PortBinding.parse(hostPort + ":" + containerPort)` binds on all interfaces (`0.0.0.0`) of the daemon host. That is fine for a local daemon but exposes the sandbox's ports publicly when `daemonUrl` points at a remote/shared machine. Consider defaulting to `127.0.0.1:<host>:<container>` and letting callers opt into a different bind address.

oss-maintainer (commented):
### Summary
New `agentscope-extensions-sandbox-docker` `docker-daemon` backend that drives a Docker engine straight through the Engine HTTP API (docker-java + zerodep transport) instead of shelling out to the `docker` CLI, so a host-side agent can use a remote daemon without docker-in-docker. The structure is sound and it is thoroughly unit-tested — re-validated mount specs on every start, orphan cleanup after a failed `start`, capped exec output, and `none` as the default network mode are all the right instincts.

Six inline notes below, all about hardening the trust boundary that a *remote* daemon introduces: ambient `DOCKER_HOST` leaking into the restore path, `network` / `workspaceRoot` taken from persisted state without the validation applied to mounts, and unbounded resources on timed-out execs. None of them are design blockers.

Two housekeeping items:
- **CLA is not signed yet** (`license/cla` is pending), so this review is a `COMMENT` rather than an approval — the PR cannot be approved or merged until the CLA is signed.
- The PR description is still the empty template. Could you fill in **Description** (motivation, what changes, how to test) and tick the **Checklist** items (`mvn spotless:apply`, `mvn test`)? It would also help reviewers to state the minimum Docker Engine API/daemon version supported, since the client pins API `1.45`.

---

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonSandbox.java
```diff
@@ -0,0 +1,622 @@
...
+
+        boolean finished;
+        try {
+            finished = returned.awaitCompletion(timeoutSeconds, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw e;
+        }
+        if (!finished) {
+            throw new SandboxException.ExecTimeoutException(String.join(" ", cmd), timeoutSeconds);
```
oss-maintainer: When `awaitCompletion` times out the method throws, but the `ExecStartResultCallback` created a few lines above is never closed and the exec is not cancelled. Against a **remote** daemon the multiplexed stdout/stderr stream (and the zerodep connection backing it) stays open until the in-container command exits on its own, so a stream of timed-out execs can pile up connections/threads. Suggest `callback.close()` (or `connection.close()`) in a `finally`/`whenComplete`, and note the leak risk in the javadoc if it is intentional.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonSandbox.java
```diff
@@ -0,0 +1,622 @@
...
+        DockerClient client = dockerClient;
+        if (client != null) {
+            try {
+                client.close();
+            } catch (Exception e) {
+                log.warn(
+                        "[sandbox-docker-daemon] Failed to close docker client: {}",
+                        e.getMessage());
+            }
+            dockerClient = null;
```
oss-maintainer: `shutdown()` reads the field and nulls it outside `synchronized (this)`, while `dockerClient()` publishes it under the lock. If a `doExec` is in flight while shutdown runs, that caller can either receive the client right before `close()` (use-after-close) or re-enter `dockerClient()` after the null and build a fresh client that nobody ever closes. Consider making the whole close path synchronized plus a `closed` flag that makes `dockerClient()` fail fast instead of recreating.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonSandbox.java
```diff
@@ -0,0 +1,622 @@
...
+            int timeoutSeconds,
+            int stdoutMaxBytes)
+            throws Exception {
+        try {
+            return doExecInContainer(containerId, workingDir, cmd, timeoutSeconds, stdoutMaxBytes);
+        } catch (SandboxException | InterruptedException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new SandboxException.SandboxRuntimeException(
+                    SandboxErrorCode.WORKSPACE_START_ERROR,
```
oss-maintainer: Every non-`SandboxException` failure inside `execInContainer` is wrapped as `WORKSPACE_START_ERROR` — including plain `doExec` calls and workspace persistence. The error code then mis-describes the failure to callers that switch on it (the new tests assert this code for exec paths). Worth a dedicated `SANDBOX_EXEC_ERROR` (or mapping the code from the caller) before this becomes an API contract.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonClientFactory.java
oss-maintainer: The Engine API version is pinned to `1.45`. Daemons whose maximum supported version is older will reject every request with a 400 that is hard to attribute to the pin. Could this be configurable through `DockerDaemonSandboxClientOptions` (with `1.45` as the default), or fall back to docker-java's version negotiation, and call the failure mode out in `docs/v2/en/docs/harness/sandbox.md`?

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-docker/src/main/java/io/agentscope/extensions/sandbox/dockerdaemon/DockerDaemonClientFactory.java
oss-maintainer: This is a new path that hands the sandbox an *arbitrary* daemon endpoint: TLS/CA material is picked up implicitly from `DOCKER_TLS_VERIFY` / `DOCKER_CERT_PATH`, so an operator who sets `daemonUrl=https://...` without those env vars silently connects without certificate verification, and the docker API is root-equivalent on the daemon host. Please make the insecure route explicit and loud — e.g. refuse `https://` unless verification is configured (or require an opt-in `allowInsecureDaemonTls` flag in `DockerDaemonSandboxClientOptions`) and document the trust boundary in the sandbox docs.

oss-maintainer (commented):
### Summary
New `sandbox-docker-daemon` module that drives a remote Docker Engine through docker-java (zerodep transport) instead of shelling out to the `docker` CLI — no local socket dependency. Structure and test coverage look solid (6 test classes, exec/persist/mount paths all exercised), and the state/Jackson module wiring matches the existing docker sandbox conventions.

Blocking points before this can move forward: the **CLA is still unsigned**, and there are a few resource-safety / trust-boundary issues around talking to a remote daemon. The TLS one is the one I would not merge without a decision.

### Findings
- **[Critical]** `DockerDaemonClientFactory.java:76` — `https://` daemon URLs silently skip certificate verification unless `DOCKER_TLS_VERIFY`/`DOCKER_CERT_PATH` happen to be set in the ambient environment. The Docker API is root-equivalent on the daemon host, so this needs an explicit, opt-in insecure flag.
- **[Warning]** `DockerDaemonSandbox.java:565` — exec callback is never closed when `awaitCompletion` times out; against a remote daemon this leaks the multiplexed stream/connection per timed-out exec.
- **[Warning]** `DockerDaemonSandbox.java:165` — `shutdown()` nulls the cached client outside the lock used by the double-checked `dockerClient()`; a concurrent `doExec` can use a closing client or leak a freshly built one.
- **[Info]** `DockerDaemonSandbox.java:527` — all exec failures are wrapped as `WORKSPACE_START_ERROR`; consider a dedicated exec error code.
- **[Info]** `DockerDaemonClientFactory.java:33` — Engine API version hard-pinned to `1.45`; consider making it configurable and documenting the failure mode on older daemons.

### Notes
`mergeable` is currently `MERGEABLE/CLEAN` (the `BLOCKED` merge state is the usual missing-review / CI gating, not a conflict), so no conflict reminder is needed. Once the CLA check turns green and the TLS default is addressed, I am happy to re-review — the rest is polish.

---

PR #2662, opened 2026-08-11
fix(harness): activate skill-bound tool groups on skill load

### LINKED ISSUE #2653: HarnessAgent  (Skill + ToolGroup: on-demand tool disclosure) not work
Image

not work of HarnessAgent

StevenZhang2002:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

SkillLoadTool now activates SkillToolGroups bound via activateOnSkill whenever a skill resource loads successfully, matching the core SkillToolFactory semantics so HarnessAgent exposes the tools on demand. SkillRuntime injects the agent toolkit into the load tool and binds it to the call-scoped RuntimeContext.

**Background**

Issue #2653: HarnessAgent (Skill + ToolGroup: on-demand tool disclosure) not work. When a skill declares tool groups via `activateOnSkill`, loading the skill through HarnessAgent's `SkillLoadTool` only returned the resource text but never activated the bound `SkillToolGroup`s, so the model could not see or call those tools. The core skill access tool (`SkillToolFactory.activateSkill`, used by ReActAgent) already performed this activation — the harness implementation was missing it.

**Changes made**

- `SkillLoadTool`: after a skill resource loads successfully (SKILL.md / in-memory / lazy), resolves the agent toolkit and activates every inactive `SkillToolGroup` bound via `activateOnSkill` (new `toolkitFor` / `activateSkillTools` helpers). Activation failures are logged as warnings and never block resource delivery.
- `SkillRuntime`: keeps an `AtomicReference<Toolkit>` refreshed in `prepareToolkit`, and binds the live toolkit into the call-scoped `RuntimeContext` in `install`, so the load tool can reach the actual agent toolkit.
- `SkillLoadToolToolGroupActivationTest` (new): verifies that loading a skill activates the associated skill-bound tool group, and that its tools become visible afterwards.

**How to test**
```bash 
mvn test -pl agentscope-harness -Dtest=SkillLoadToolToolGroupActivationTest
```
Fixes #2653

## Checklist

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillLoadTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillLoadTool.java
@@ -20,6 +20,8 @@ import io.agentscope.core.skill.AgentSkill;
 import io.agentscope.core.skill.util.MarkdownSkillParser;
 import io.agentscope.core.tool.AgentTool;
 import io.agentscope.core.tool.ToolCallParam;
+import io.agentscope.core.tool.ToolGroup;
+import io.agentscope.core.tool.Toolkit;
 import io.agentscope.harness.agent.skill.SkillResources;
 import java.util.ArrayList;
 import java.util.LinkedHashSet;
@@ -50,6 +52,11 @@ import reactor.core.publisher.Mono;
  * ToolCallParam#getRuntimeContext() tool call's RuntimeContext}, keeping concurrent calls and
  * sessions isolated without re-registering the tool. A legacy catalog reference is consulted only
  * for deprecated callers that invoke the tool without any RuntimeContext.
+ *
+ * <p>On-demand tool disclosure: once a resource is found, every {@link
+ * io.agentscope.core.tool.SkillToolGroup} bound to the skill via {@code activateOnSkill} is
+ * activated on the call's toolkit, so the skill's tools become visible to the model for the
+ * following turns (see issue #2653).
  */
 @SuppressWarnings("deprecation")
 public final class SkillLoadTool implements AgentTool {
@@ -61,10 +68,11 @@ public final class SkillLoadTool implements AgentTool {
 
     private final AtomicReference<SkillCatalog> legacyCatalogRef;
     private final boolean exposeLegacyCatalogInSchema;
+    private final AtomicReference<Toolkit> toolkitRef;
 
     /** Creates a context-scoped loader with an empty context-free compatibility catalog. */
     public SkillLoadTool() {
-        this(new AtomicReference<>(SkillCatalog.empty()), false);
+        this(new AtomicReference<>(SkillCatalog.empty()), false, new AtomicReference<>());
     }
 
     /**
@@ -80,16 +88,24 @@ public final class SkillLoadTool implements AgentTool {
      */
     @Deprecated(since = "2.2.0")
     public SkillLoadTool(AtomicReference<SkillCatalog> catalogRef) {
-        this(catalogRef, true);
+        this(catalogRef, true, new AtomicReference<>());
     }
 
     SkillLoadTool(
             AtomicReference<SkillCatalog> legacyCatalogRef, boolean exposeLegacyCatalogInSchema) {
+        this(legacyCatalogRef, exposeLegacyCatalogInSchema, new AtomicReference<>());
+    }
+
+    SkillLoadTool(
+            AtomicReference<SkillCatalog> legacyCatalogRef,
+            boolean exposeLegacyCatalogInSchema,
+            AtomicReference<Toolkit> toolkitRef) {
         if (legacyCatalogRef == null) {
             throw new IllegalArgumentException("catalogRef must not be null");
         }
         this.legacyCatalogRef = legacyCatalogRef;
         this.exposeLegacyCatalogInSchema = exposeLegacyCatalogInSchema;
+        this.toolkitRef = toolkitRef != null ? toolkitRef : new AtomicReference<>();
     }
 
     @Override
@@ -177,7 +193,8 @@ public final class SkillLoadTool implements AgentTool {
                                         + "'. Check the available skills list."));
             }
 
-            return Mono.just(loadOne(entry, path));
+            Toolkit toolkit = toolkitFor(param);
+            return Mono.just(loadOne(entry, path, toolkit));
         } catch (Exception e) {
             log.error("load_skill_through_path failed", e);
             return Mono.just(ToolResultBlock.error(e.getMessage()));
@@ -192,17 +209,35 @@ public final class SkillLoadTool implements AgentTool {
         return legacy != null ? legacy : SkillCatalog.empty();
     }
 
-    private ToolResultBlock loadOne(HarnessSkillEntry entry, String path) {
+    /**
+     * Resolves the toolkit used to activate skill-bound tool groups. Prefers the call-scoped
+     * instance (installed by {@link SkillRuntime#install(SkillCatalog,
+     * io.agentscope.core.agent.RuntimeContext, Toolkit)}), falling back to the shared reference
+     * kept current by {@link SkillRuntime#prepareToolkit(Toolkit)} for context-free callers.
+     */
+    private Toolkit toolkitFor(ToolCallParam param) {
+        if (param.getRuntimeContext() != null) {
+            Toolkit toolkit = param.getRuntimeContext().get(Toolkit.class);
+            if (toolkit != null) {
+                return toolkit;
+            }
+        }
+        return toolkitRef.get();
+    }
+
+    private ToolResultBlock loadOne(HarnessSkillEntry entry, String path, Toolkit toolkit) {
         AgentSkill skill = entry.skill();
 
         // 1. Special case: SKILL.md returns the parsed body text.
         if (SKILL_FILE.equals(path)) {
+            activateSkillTools(toolkit, skill);
             return ToolResultBlock.text(formatSkillMarkdown(entry));
         }
 
         // 2. In-memory map (Layer 1/3 host repos + marketplace fully-loaded resources).
         Map<String, String> mem = skill.getResources();
         if (mem != null && mem.containsKey(path)) {
+            activateSkillTools(toolkit, skill);
             return ToolResultBlock.text(formatResource(entry, path, mem.get(path)));
         }
 
@@ -210,12 +245,55 @@ public final class SkillLoadTool implements AgentTool {
         if (entry.lazyResources() != null) {
             Optional<String> lazy = entry.lazyResources().read(path);
             if (lazy.isPresent()) {
+                activateSkillTools(toolkit, skill);
                 return ToolResultBlock.text(formatResource(entry, path, lazy.get()));
             }
         }
 
         // 4. Not found: enumerate both sources so the model sees real options.
         return ToolResultBlock.error(formatNotFound(entry, path));
+    }
+
+    /**
+     * Activates every {@link io.agentscope.core.tool.SkillToolGroup} bound to the loaded skill via
+     * {@code activateOnSkill} so its tools become visible to the model (on-demand tool disclosure).
+     *
+     * <p>Matches the activation semantics of the core skill access tool ({@code
+     * SkillToolFactory.activateSkill}): a group is only activated after a resource was actually
+     * found, and only when it is not already active. The toolkit's shared activation state is
+     * picked up by {@code ReActAgent}'s post-acting {@code syncToolkitToState}, which persists it
+     * into the session's {@code ToolContextState} for subsequent reasoning turns.
+     */
+    private void activateSkillTools(Toolkit toolkit, AgentSkill skill) {
+        if (toolkit == null || skill == null) {
+            return;
+        }
+        try {
+            List<String> boundGroups =
+                    toolkit.findSkillToolGroupsByActivateOnSkill(skill.getName());
+            if (boundGroups.isEmpty()) {
+                return;
+            }
+            List<String> toActivate = new ArrayList<>(boundGroups.size());
+            for (String group : boundGroups) {
+                ToolGroup existing = toolkit.getToolGroup(group);
+                if (existing != null && !existing.isActive()) {
+                    toActivate.add(group);
+                }
+            }
+            if (!toActivate.isEmpty()) {
+                toolkit.updateToolGroups(toActivate, true);
+                log.info(
+                        "Activated skill-bound tool groups {} for skill '{}'",
+                        toActivate,
+                        skill.getName());
+            }
+        } catch (Exception e) {
+            log.warn(
+                    "Failed to activate skill-bound tool groups for skill '{}': {}",
+                    skill.getName(),
+                    e.getMessage());
+        }
     }
 
     // ---------------------------------------------------------------------
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillRuntime.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillRuntime.java
@@ -40,6 +40,7 @@ public final class SkillRuntime {
     private final RuntimeContext legacyContext = RuntimeContext.empty();
     private final AtomicReference<SkillCatalog> legacyCatalogRef =
             new AtomicReference<>(SkillCatalog.empty());
+    private final AtomicReference<Toolkit> toolkitRef = new AtomicReference<>();
     private final SkillLoadTool loadTool;
     private final SkillPromptBuilder promptBuilder;
 
@@ -49,7 +50,7 @@ public final class SkillRuntime {
 
     public SkillRuntime(SkillPromptBuilder promptBuilder) {
         this.promptBuilder = promptBuilder != null ? promptBuilder : new SkillPromptBuilder();
-        this.loadTool = new SkillLoadTool(legacyCatalogRef, false);
+        this.loadTool = new SkillLoadTool(legacyCatalogRef, false, toolkitRef);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


duola8789:
Thanks for this fix — we independently ran into the same bug and can confirm both the diagnosis and the impact.

**Our reproduction (2.0.3 release):** scripted stub `Model` that records the exact tool list each reasoning round receives. `load_skill_through_path` returns SUCCESS and the skill body enters the context, but the bound group's tool is still absent from the tool list in the next reasoning round of the same call. Control experiment: calling `reset_equipped_tools` with the group name makes the tool visible immediately — so group state, session-state propagation and schema filtering all work; the only missing link is load → group activation, exactly as you described.

**Root cause confirmation:** bytecode reference graph on 2.0.3 shows `SkillToolFactory` (which implements `activateSkill` + the group bring-out) is only referenced by `SkillBox` — the harness `skillRepository()` path never touches it. We also verified current `main` still has the plain-reading `SkillLoadTool`, consistent with this PR being unmerged.

**Workaround we're running** (in case it's useful to anyone hitting this before the PR lands): a `MiddlewareBase` overriding `onActing` — on a `ToolResultEndEvent` for `load_skill_through_path` with state `SUCCESS`, resolve the skill name from the repository (`getAllSkills()`, match `getSkillId()`), find groups whose `activateOnSkill` matches, and merge them into the session state's `toolContext.activatedGroups` (mirroring the toolkit flag as well, since per-round tool lists are computed from `toolkit.getToolSchemas(state.getActivatedGroups())`). Verified end-to-end with a real model.

One small documentation note this PR might want to include: the binding match is `String.equals` against `AgentSkill.getName()` (the frontmatter `name`), not the skill-id (`<name>_<source>`) — while the load tool's parameter is the skill-id. That asymmetry is easy to trip over.

Happy to help test this PR against our reproduction if that's useful.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillRuntime.java
```diff
@@ -125,6 +133,7 @@ public void prepareToolkit(Toolkit toolkit) {
     public void install(SkillCatalog catalog, RuntimeContext context, Toolkit toolkit) {
         if (context != null) {
             context.put(SkillCatalog.class, catalog != null ? catalog : SkillCatalog.empty());
+            context.put(Toolkit.class, toolkit);
```
oss-maintainer, line 136: `install` now publishes the agent's **live `Toolkit`** into the call-scoped `RuntimeContext` under `Toolkit.class`. Before this change the context only carried `SkillCatalog`, which the class javadoc explicitly describes as "the authorization boundary used by `SkillLoadTool`". With a raw `Toolkit` in the context, any tool or middleware that receives the same `RuntimeContext` can now mutate registration state (register/remove tools, activate/deactivate groups), not just read the catalog.

Could this be narrowed to the one capability the load tool needs, e.g. put a small functional handle instead of the whole toolkit:

```java
public interface SkillGroupActivator { void activate(String skillName); }
context.put(SkillGroupActivator.class, (SkillGroupActivator) name -> activateBoundGroups(name, toolkit));
```

If keeping `Toolkit.class` is intentional, it is worth stating the trust assumption in the javadoc ("the toolkit in context is writable and must only be installed by harness-owned middleware"), because sub-agents and extensions also share `RuntimeContext`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillLoadTool.java
```diff
@@ -218,6 +254,48 @@ private ToolResultBlock loadOne(HarnessSkillEntry entry, String path) {
...
+            }
+            List<String> toActivate = new ArrayList<>(boundGroups.size());
+            for (String group : boundGroups) {
+                ToolGroup existing = toolkit.getToolGroup(group);
+                if (existing != null && !existing.isActive()) {
+                    toActivate.add(group);
+                }
+            }
+            if (!toActivate.isEmpty()) {
+                toolkit.updateToolGroups(toActivate, true);
```
oss-maintainer, line 285: Activation is written to the **shared toolkit** (`toolkit.updateToolGroups(...)`), and reaches session state only indirectly via `ReActAgent.syncToolkitToState`, i.e. `state.getToolContext().setActivatedGroups(toolkit.getActiveGroups())`. Two consequences worth checking:

1. **Persistence is conditional on the acting path taken.** `syncToolkitToState(state)` is called after the acting loop completes (ReActAgent.java ~2935) — but the branches just above it return early for `isStopRequested()`, `returnDirect` and non-empty `pendingPairs` (suspension / awaiting permission approval). If `load_skill_through_path` is followed by a stop, a direct return, or a suspended tool call, the group is activated in memory but never copied into the persisted `ToolContextState`, so the next call — or a session recovered from an `AgentStateStore`, which is the whole point of `agentscope-distribution` — starts with the toolkit active and the state not, and `getActiveGroups()`/state reconciliation can flip it back off. Would it be more robust to activate through the call's own state (`state.getToolContext().addActivatedGroup(...)`) and let the toolkit mirror follow, the way `reset_equipped_tools` does?
2. **Shared instance, multiple sessions.** A `Toolkit` is per agent instance, while `activatedGroups` is per session. When one `HarnessAgent` instance serves several sessions, a skill loaded in session A activates the group for everyone whose state is synced from that toolkit afterwards. The `SkillCatalog` half of this PR is correctly call-scoped; the toolkit half is not.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillLoadTool.java
```diff
@@ -192,24 +209,43 @@ private SkillCatalog catalogFor(ToolCallParam param) {
...
-    private ToolResultBlock loadOne(HarnessSkillEntry entry, String path) {
+    /**
+     * Resolves the toolkit used to activate skill-bound tool groups. Prefers the call-scoped
+     * instance (installed by {@link SkillRuntime#install(SkillCatalog,
+     * io.agentscope.core.agent.RuntimeContext, Toolkit)}), falling back to the shared reference
+     * kept current by {@link SkillRuntime#prepareToolkit(Toolkit)} for context-free callers.
+     */
+    private Toolkit toolkitFor(ToolCallParam param) {
+        if (param.getRuntimeContext() != null) {
+            Toolkit toolkit = param.getRuntimeContext().get(Toolkit.class);
```
oss-maintainer, line 220: `toolkitFor` falls back to the shared `toolkitRef` whenever the call context has no `Toolkit` — which includes the case where `install` was called with a null toolkit (`agent == null` in `HarnessSkillMiddleware.onSystemPrompt`, line 266 there). So a context-scoped call can silently activate groups on the *last* toolkit `prepareToolkit` saw (constructor-time toolkit, i.e. the pre-`copy()` instance at `ReActAgent.Builder:5535`) rather than the agent's live one. Since the deep copy keeps the tool instances but not the group state, that fallback can activate a group that this agent's schema rendering never reads. Worth an explicit "no reachable toolkit -> log at debug and skip" branch instead of mutating whatever reference is cached.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillLoadTool.java
```diff
@@ -218,6 +254,48 @@ private ToolResultBlock loadOne(HarnessSkillEntry entry, String path) {
...
+     * Activates every {@link io.agentscope.core.tool.SkillToolGroup} bound to the loaded skill via
+     * {@code activateOnSkill} so its tools become visible to the model (on-demand tool disclosure).
+     *
+     * <p>Matches the activation semantics of the core skill access tool ({@code
+     * SkillToolFactory.activateSkill}): a group is only activated after a resource was actually
+     * found, and only when it is not already active. The toolkit's shared activation state is
+     * picked up by {@code ReActAgent}'s post-acting {@code syncToolkitToState}, which persists it
+     * into the session's {@code ToolContextState} for subsequent reasoning turns.
+     */
+    private void activateSkillTools(Toolkit toolkit, AgentSkill skill) {
```
oss-maintainer, line 267: Minor, for discoverability: matching is `findSkillToolGroupsByActivateOnSkill(skill.getName())` — i.e. exact `String.equals` against the frontmatter `name` — while the parameter the model passes to this tool is the **skillId** (`<name>_<source>`). @duola8789 hit exactly this in the thread above. A one-line javadoc note here ("bound by skill *name*, not skill id") plus the same note in the `activateOnSkill` builder javadoc would save the next person a debugging session.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillLoadTool.java
```diff
@@ -218,6 +254,48 @@ private ToolResultBlock loadOne(HarnessSkillEntry entry, String path) {
...
+            }
+            if (!toActivate.isEmpty()) {
+                toolkit.updateToolGroups(toActivate, true);
+                log.info(
+                        "Activated skill-bound tool groups {} for skill '{}'",
+                        toActivate,
+                        skill.getName());
+            }
+        } catch (Exception e) {
+            log.warn(
```
oss-maintainer, line 292: `log.info` for the happy path fires on every successful resource load, not only on the first one for a group (the `isActive()` filter at line 281 covers repeat loads, so in practice it is once per group — fine). The `catch (Exception e)` only logs `e.getMessage()`, which drops the stack for the most likely failure (`updateToolGroups` throwing `IllegalArgumentException` on a vanished group). Consider `log.warn("...", e)` or at least including the exception class.

agentscope-harness/src/test/java/io/agentscope/harness/agent/skill/runtime/SkillLoadToolToolGroupActivationTest.java
```diff
@@ -0,0 +1,433 @@
...
+     * the associated SkillToolGroup, so tools in that group remain hidden from the model.
+     *
+     * <p>Expected behavior: After loading a skill, any SkillToolGroup bound to that skill
+     * via {@code activateOnSkill} should be activated, making its tools available.
+     *
+     * <p>Actual behavior (bug): The SkillToolGroup remains inactive after skill loading.
+     */
+    @Test
+    @DisplayName("Issue #2653: Loading skill should activate associated SkillToolGroup")
+    void loadingSkillShouldActivateAssociatedSkillToolGroup() {
```
oss-maintainer, line 65: The comments in these two tests still read like a bug report against unfixed code — "THIS IS THE BUG: The group should be active after skill loading, but it isn't", "Actual behavior (bug)" in the class javadoc — while the assertions here verify the fixed behaviour. Please reword to describe the guarantee being pinned, otherwise a future reader thinks the test documents a live defect. The coverage itself is good: context toolkit, shared-reference fallback, idempotent reload, no-toolkit-still-succeeds and activation-failure-is-swallowed are all covered. What is missing is an end-to-end assertion on `syncToolkitToState` (tool visible in the *next* reasoning round / persisted `ToolContextState`), which is the half that point 1 above is about.

oss-maintainer (commented):
### Summary
Closes the real gap from #2653: the harness `SkillLoadTool` never performed the `activateOnSkill` group activation that `SkillToolFactory.activateSkill` does on the core path, and the new tests pin the behaviour end-to-end at the toolkit level. Confirmed against `main`: `activateSkillTools` mirrors core semantics (only on a found resource, only when the group is inactive), `findSkillToolGroupsByActivateOnSkill`/`updateToolGroups`/`getActiveGroups` all exist as used, and `HarnessSkillMiddleware` is the only caller of `install`/`prepareToolkit`, so no other call site breaks on the new signature. CLA signed, all six checks green, mergeable.

What I would settle before merge is **where the activation is authoritative**: this PR writes the shared `Toolkit` and lets `ReActAgent.syncToolkitToState` copy it into `ToolContextState` afterwards, and that copy is skipped on the early-return acting paths — plus it needs the agent's writable toolkit inside `RuntimeContext`, which is a step back from the catalog-as-authorization-boundary design the class javadoc describes.

- **[Warning]** `SkillRuntime.java:136` — the call context now carries the agent's mutable `Toolkit`, not just the read-only `SkillCatalog`; suggested narrowing to a single-purpose activator handle or documenting the trust boundary.
- **[Warning]** `SkillLoadTool.java:285` — activation lands on the shared toolkit and only reaches session state via `syncToolkitToState`, which is not called on the `isStopRequested` / `returnDirect` / pending-suspension branches (`ReActAgent.java:~2920-2935`); a skill loaded immediately before a suspension or direct return is active in memory but absent from the persisted state, which is exactly the state-recovery path `agentscope-distribution` relies on. Same shared-toolkit write also crosses sessions when one agent instance serves several.
- **[Info]** `SkillLoadTool.java:220` — the `toolkitRef` fallback can activate a pre-`copy()` toolkit (builder copies at `ReActAgent.java:5535`) when `install` ran with a null agent toolkit.
- **[Info]** `SkillLoadTool.java:267` — the name-vs-skillId matching asymmetry @duola8789 flagged deserves a javadoc line.
- **[Info]** `SkillLoadTool.java:292` — `catch (Exception e)` logs only `e.getMessage()`, dropping the stack for the likeliest failure.
- **[Info]** test file — the "THIS IS THE BUG" comments now describe behaviour the tests assert is fixed; reword, and consider one end-to-end case that reaches the next reasoning round.

Nothing here is a correctness defect in the common path; the two warnings are about persistence and blast radius, which matter for a framework whose harness/distribution modules promise session isolation and recovery.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillLoadTool.java
```diff
@@ -218,6 +254,48 @@ private ToolResultBlock loadOne(HarnessSkillEntry entry, String path) {
         return ToolResultBlock.error(formatNotFound(entry, path));
     }
 
+    /**
+     * Activates every {@link io.agentscope.core.tool.SkillToolGroup} bound to the loaded skill via
+     * {@code activateOnSkill} so its tools become visible to the model (on-demand tool disclosure).
```
oss-maintainer, line 259: `SkillLoadTool` is registered as a **single shared instance** (`SkillRuntime.loadTool` is `final`, built once in the constructor, and returned by `loadTool()`), while the thing this new code hands it — the live toolkit — is per agent build. So the shared-reference fallback is not just for "deprecated callers without a RuntimeContext": after a second `HarnessAgent` is built in the same JVM/process from a skill repository that reuses the middleware (or whenever `install` ran with a null agent), `toolkitRef.get()` points at whichever agent most recently passed through `prepareToolkit`, and activation mutates **that** agent's groups.

Concretely: `prepareToolkit` is called from both `HarnessSkillMiddleware` construction (line 213) and every `onSystemPrompt`, and `context.put(Toolkit.class, toolkit)` is unconditional — so a call whose context has no toolkit (legacy `call(msgs)` path, `AgentBase:214` skips the write when `context == null`) silently activates groups on the last-prepared agent. Suggest either making the absence explicit (`Optional<Toolkit>` semantics: no context toolkit **and** legacy constructor ⇒ log at debug and skip) or binding the toolkit per call rather than per runtime.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillRuntime.java
oss-maintainer, line 136: `prepareToolkit(toolkit)` unconditionally overwrites `toolkitRef` on every system-prompt pass, but the deep-copied toolkit from `ReActAgent.Builder` (line 5535) is what the model schemas are read from, and the middleware holds its own reference (`agentToolkit = agent.getToolkit()`). The two only agree because activation happens after the copy — the javadoc you added here says "always points at the agent's live toolkit", which is true only in the steady state and makes the constructor-time `prepareToolkit` call at `HarnessSkillMiddleware:213` misleading (it pre-binds a toolkit that will be replaced). Worth stating that the reference is best-effort and the context binding is authoritative, so a reader doesn't rely on the constructor bind.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillLoadTool.java
```diff
@@ -218,6 +254,48 @@ private ToolResultBlock loadOne(HarnessSkillEntry entry, String path) {
...
+     * picked up by {@code ReActAgent}'s post-acting {@code syncToolkitToState}, which persists it
+     * into the session's {@code ToolContextState} for subsequent reasoning turns.
+     */
+    private void activateSkillTools(Toolkit toolkit, AgentSkill skill) {
+        if (toolkit == null || skill == null) {
+            return;
+        }
+        try {
+            List<String> boundGroups =
+                    toolkit.findSkillToolGroupsByActivateOnSkill(skill.getName());
```
oss-maintainer, line 273: Activation is per-resource-load and mutates the toolkit, so on a session whose persisted `ToolContextState.activatedGroups` was previously empty the **first** load of a skill permanently flips the group on for the rest of that session's life — and, given point 1, potentially for the shared toolkit. That matches core's `SkillToolFactory.activateSkill` semantics, so this is a question rather than a defect: is there a deactivation story for the harness path? Core at least has `skillRegistry.setSkillActive(skillId, false)` alongside it; here there is no equivalent, so a model that loads a skill and then moves on keeps the tools disclosed. If that's intended ("once disclosed, stays disclosed"), say so in the class javadoc, because the on-demand-disclosure feature is usually sold as the opposite.

oss-maintainer (commented):
Following up on my own review with three points I under-called the first time round, after re-checking the shared-instance lifecycle:

1. **`SkillLoadTool` is one shared instance per `SkillRuntime`** (the field is `final` and built in the constructor), so the `toolkitRef` fallback is not merely a legacy-caller path — it is the path any call takes when its `RuntimeContext` carries no toolkit, and it mutates whichever agent last ran `prepareToolkit`. That is the concrete version of the cross-session concern raised inline.
2. `prepareToolkit` is called from both the middleware constructor and every `onSystemPrompt`, so the constructor-time bind is stale by design; the new javadoc's "always points at the agent's live toolkit" overstates it.
3. There is no deactivation counterpart here (`SkillToolFactory` at least has `setSkillActive(skillId, false)`), so "on-demand disclosure" is one-way per session.

None of these change my read that the fix is correct and worth having — #2653 is a genuine gap versus core. Flagging so the shared-toolkit lifetime is decided explicitly rather than inherited.

Issue #2664, opened 2026-08-11
[Bug] `.skills-cache` is never cleaned up when skill repositories return an empty collection; HarnessAgent still reports installed skills


weixsun:
**Describe the bug**

When using a database-backed skill repository (e.g. `MysqlSkillRepository`), after all skills are removed from the repository so that `getAllSkills()` returns an empty collection, the skill files previously staged under `.agentscope/workspace/.skills-cache` in the `HarnessAgent` workspace are **never cleaned up**.

Since `.skills-cache` is part of the sandbox workspace projection allow-list (`SandboxFilesystemSpec.DEFAULT_WORKSPACE_PROJECTION_ROOTS`), the stale SKILL.md / script files are still hydrated into the sandbox on every sandbox start. The model can see these files via file-browsing/shell tools and, combined with `<available_skills>` context left in the conversation history, keeps answering something like "According to my memory, there is currently 1 available Skill installed: ...", which is inconsistent with the fact that the database has no skills at all.

Root cause: `HarnessSkillMiddleware.onSystemPrompt()` and `prestageMarketplaceSkills()` return early when `merged.isEmpty()` / `visible.isEmpty()` and never invoke `MarketplaceStager.stage()`, while orphan directory reclamation (`garbageCollectOrphans()`) only runs inside `stage()` — as long as the repositories keep returning an empty collection, GC is never triggered.

**To Reproduce**

1. Build a HarnessAgent with `MysqlSkillRepository` as the skill store and sandbox enabled:

```java
MysqlSkillRepository skillRepo = new MysqlSkillRepository(dataSource, true, true);

HarnessAgent agent = HarnessAgent.builder()
        .name("demo")
        .model(model)
        .workspace(workspaceRoot)
        .skillRepository(skillRepo)
        // sandbox mode enabled
        .build();
```

2. Save one skill to the database and call the agent so it gets staged into `.skills-cache`:

```java
AgentSkill skill = new AgentSkill(
        "demo-skill", "demo", "content",
        Map.of("run.sh", "#!/bin/bash\necho hello"), "mysql");
skillRepo.save(List.of(skill), false);

agent.call("list the currently available skills", ctx).block();
// now <workspace>/.skills-cache/<source-ns>/demo-skill/ exists
```

3. Remove all skills from the database so that `skillRepo.getAllSkills()` returns an empty collection, then call the agent again:

```java
skillRepo.delete("demo-skill");  // or TRUNCATE the related tables
[...]

PR #2665, opened 2026-08-11
fix(skills): run orphan GC on empty merged skills to clear stale cache

weixsun:
## Description

Closing #2664 

## Checklist

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
@@ -176,7 +176,8 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
         this.frozenSkills =
                 freezeRepositories
                         ? Collections.unmodifiableMap(
-                                new LinkedHashMap<>(mergeRepositories(RuntimeContext.empty())))
+                                new LinkedHashMap<>(
+                                        mergeRepositories(RuntimeContext.empty()).skills()))
                         : null;
         this.runtime.prepareToolkit(toolkit);
     }
@@ -195,7 +196,8 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
      * Pre-stages marketplace skill resources to {@code .skills-cache/} on the host workspace.
      * Intended to be called <em>before</em> sandbox start so that workspace projection picks up
      * the staged content in the same call. Safe to call multiple times — staging is idempotent
-     * (content-hash guarded).
+     * (content-hash guarded). When no skill survives merging/filtering, orphan GC runs with an
+     * empty retained white-list so stale cache content is not projected into the sandbox.
      *
      * @param ctx the per-call runtime context
      */
@@ -206,15 +208,19 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
         if (ctx == null) {
             ctx = RuntimeContext.empty();
         }
-        Map<String, RepoBound> merged = skillsForCall(ctx);
+        MergeResult merge = skillsForCall(ctx);
+        Map<String, RepoBound> merged = merge.skills();
         if (merged.isEmpty()) {
+            stageEmptyRetainedIfSafe(merge.loadFailed());
             return;
         }
         List<RepoBound> visible = applyVisibility(merged.values(), ctx);
         List<RepoBound> enabled = applySkillFilter(visible, effectiveFilter(ctx));
-        if (!enabled.isEmpty()) {
-            stager.stage(enabled, sourceNamespaces);
+        if (enabled.isEmpty()) {
+            stageEmptyRetainedIfSafe(merge.loadFailed());
+            return;
         }
+        stager.stage(enabled, sourceNamespaces);
     }
 
     @Override
@@ -225,8 +231,10 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
 
         Toolkit agentToolkit = agent != null ? agent.getToolkit() : null;
 
-        Map<String, RepoBound> merged = skillsForCall(ctx);
+        MergeResult merge = skillsForCall(ctx);
+        Map<String, RepoBound> merged = merge.skills();
         if (merged.isEmpty()) {
+            stageEmptyRetainedIfSafe(merge.loadFailed());
             runtime.install(SkillCatalog.empty(), ctx, agentToolkit);
             return Mono.just(currentPrompt);
         }
@@ -234,6 +242,7 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
         List<RepoBound> visible = applyVisibility(merged.values(), ctx);
         List<RepoBound> enabled = applySkillFilter(visible, effectiveFilter(ctx));
         if (enabled.isEmpty()) {
+            stageEmptyRetainedIfSafe(merge.loadFailed());
             runtime.install(SkillCatalog.empty(), ctx, agentToolkit);
             return Mono.just(currentPrompt);
         }
@@ -275,8 +284,19 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
     //  Internals
     // ---------------------------------------------------------------------
 
-    private Map<String, RepoBound> skillsForCall(RuntimeContext ctx) {
-        return frozenSkills != null ? frozenSkills : mergeRepositories(ctx);
+    private MergeResult skillsForCall(RuntimeContext ctx) {
+        return frozenSkills != null ? new MergeResult(frozenSkills, false) : mergeRepositories(ctx);
+    }
+
+    /**
+     * Runs orphan GC with an empty retained white-list so stale {@code .skills-cache} content
+     * is removed when no skill survives merging/filtering. Skipped when a repository failed to
+     * load — an empty result caused by an error must not wipe the cache (outage protection).
+     */
+    private void stageEmptyRetainedIfSafe(boolean loadFailed) {
+        if (stager != null && !loadFailed) {
+            stager.stage(List.of(), sourceNamespaces);
+        }
     }
 
     private SkillFilter effectiveFilter(RuntimeContext ctx) {
@@ -298,13 +318,17 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
         return out;
     }
 
+    /** Outcome of {@link #mergeRepositories}: merged skills plus any-load-failure flag. */
+    private record MergeResult(Map<String, RepoBound> skills, boolean loadFailed) {}
+
     /**
      * Merge skills from every repository, in compose order. Later entries with the same
      * {@code AgentSkill.name} win. Also remembers the source repository per winning skill so
      * subsequent steps (lazy resources, marketplace stage) can act on it.
      */
-    private LinkedHashMap<String, RepoBound> mergeRepositories(RuntimeContext ctx) {
+    private MergeResult mergeRepositories(RuntimeContext ctx) {
         LinkedHashMap<String, RepoBound> merged = new LinkedHashMap<>();
+        boolean loadFailed = false;
         for (AgentSkillRepository repo : repositories) {
             List<AgentSkill> skills;
             try {
@@ -313,6 +337,7 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
                                 ? contextRepository.getAllSkills(ctx)
                                 : repo.getAllSkills();
             } catch (Exception e) {
+                loadFailed = true;
                 log.warn(
                         "Skill repository {} failed to load: {}",
                         repo.getClass().getSimpleName(),
@@ -329,7 +354,7 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
                 merged.put(skill.getName(), new RepoBound(skill, repo));
             }
         }
-        return merged;
+        return new MergeResult(merged, loadFailed);
     }
 
     private List<RepoBound> applyVisibility(
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddlewareEmptyGcTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddlewareEmptyGcTest.java
@@ -1,0 +1,294 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent.middleware;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.core.skill.AgentSkill;
+import io.agentscope.core.skill.repository.AgentSkillRepository;
+import io.agentscope.core.skill.repository.AgentSkillRepositoryInfo;
+import io.agentscope.core.tool.Toolkit;
+import io.agentscope.harness.agent.skill.runtime.MarketplaceStager;
+import io.agentscope.harness.agent.skill.runtime.ShellPathPolicy;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * Verifies that an empty skill set (repository returns no skills, or every skill is filtered
+ * out) still triggers orphan GC on {@code .skills-cache/}, so stale staged content is not
+ * projected into the sandbox — while a repository <em>load failure</em> preserves the cache.
+ */
+class HarnessSkillMiddlewareEmptyGcTest {
+
+    @TempDir Path tempWorkspace;
+
+    @Test
+    void onSystemPromptClearsSkillsCacheWhenRepositoryReturnsEmpty() throws IOException {
+        MutableSkillRepository repo =
+                new MutableSkillRepository(List.of(dbSkill("stale-tool")), "test-db");
+        MarketplaceStager stager = new MarketplaceStager(tempWorkspace);
+        HarnessSkillMiddleware middleware = newMiddleware(List.of(repo), null, stager);
+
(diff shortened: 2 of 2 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
```diff
@@ -275,8 +283,19 @@ public Mono<String> onSystemPrompt(Agent agent, RuntimeContext ctx, String curre
     //  Internals
     // ---------------------------------------------------------------------
 
-    private Map<String, RepoBound> skillsForCall(RuntimeContext ctx) {
-        return frozenSkills != null ? frozenSkills : mergeRepositories(ctx);
+    private MergeResult skillsForCall(RuntimeContext ctx) {
+        return frozenSkills != null ? new MergeResult(frozenSkills, false) : mergeRepositories(ctx);
```
AgentScopeJavaBot, line 288: [nit] When `frozenSkills != null`, `loadFailed` is hardcoded to `false`. This means if all repositories failed during construction (with `freezeRepositories=true`), the `loadFailed` signal is silently discarded and a subsequent empty-merge will trigger cache GC. In practice this is harmless (empty frozen skills + empty cache = no-op GC), but a brief inline comment explaining why frozen mode always reports `loadFailed=false` would help future maintainers understand the intent.

agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddlewareEmptyGcTest.java
```diff
@@ -0,0 +1,294 @@
...
+        assertTrue(Files.exists(cacheDir().resolve("test-db/stale-tool/run.sh")));
+
+        repo.setSkills(List.of());
+
+        middleware.prestageMarketplaceSkills(RuntimeContext.empty());
+        assertCacheEmpty();
+    }
+
+    @Test
+    void onSystemPromptClearsSkillsCacheWhenVisibilityFilterHidesAll() throws IOException {
```
AgentScopeJavaBot, line 83: [nit] There's a test for `onSystemPrompt` + visibility-filter-hides-all (this test), but no symmetric test for `prestageMarketplaceSkills` + visibility-filter-hides-all. The skill-filter variant is covered (`prestageMarketplaceSkillsClearsCacheWhenSkillFilterHidesAll`), but visibility filter takes a different code path (`applyVisibility` returns empty before `applySkillFilter` is even reached). Consider adding a `prestageMarketplaceSkillsClearsCacheWhenVisibilityFilterHidesAll` test for symmetry.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
```diff
@@ -275,8 +283,19 @@ public Mono<String> onSystemPrompt(Agent agent, RuntimeContext ctx, String curre
...
+    }
+
+    /**
+     * Runs orphan GC with an empty retained white-list so stale {@code .skills-cache} content
+     * is removed when no skill survives merging/filtering. Skipped when a repository failed to
+     * load — an empty result caused by an error must not wipe the cache (outage protection).
+     */
+    private void stageEmptyRetainedIfSafe(boolean loadFailed) {
```
AgentScopeJavaBot, line 296: [praise] Clean design: reusing `stager.stage(List.of(), sourceNamespaces)` to trigger orphan GC through the existing public API avoids expanding `MarketplaceStager`'s API surface (no need to widen `garbageCollectOrphans` visibility). The `loadFailed` guard is a well-thought-out outage protection — an empty result caused by a transient error must never wipe the cache.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a real and impactful bug: when all skills are removed from repositories or filtered out by visibility/skill filters, stale content in `.skills-cache/` was never cleaned up, causing outdated skill resources to leak into the sandbox. The fix introduces a `MergeResult` record that carries both the merged skills map and a `loadFailed` flag, enabling a new `stageEmptyRetainedIfSafe()` helper that triggers orphan GC with an empty retained set — but only when the empty result is intentional (not caused by a transient repository failure). The outage-protection design is thoughtful, the code reuse of `stager.stage(List.of(), sourceNamespaces)` is clean (no API surface expansion needed), and the 5-test suite covers the key scenarios well including the critical load-failure preservation case.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
AgentScopeJavaBot, line 288: [nit] When `frozenSkills != null`, `loadFailed` is hardcoded to `false`. This means if all repositories failed during construction (with `freezeRepositories=true`), the `loadFailed` signal is silently discarded and a subsequent empty-merge will trigger cache GC. In practice this is harmless (empty frozen skills + empty cache = no-op GC), but a brief inline comment explaining why frozen mode always reports `loadFailed=false` would help future maintainers understand the intent.

agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddlewareEmptyGcTest.java
AgentScopeJavaBot, line 83: [nit] There's a test for `onSystemPrompt` + visibility-filter-hides-all (this test), but no symmetric test for `prestageMarketplaceSkills` + visibility-filter-hides-all. The skill-filter variant is covered (`prestageMarketplaceSkillsClearsCacheWhenSkillFilterHidesAll`), but visibility filter takes a different code path (`applyVisibility` returns empty before `applySkillFilter` is even reached). Consider adding a `prestageMarketplaceSkillsClearsCacheWhenVisibilityFilterHidesAll` test for symmetry.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
AgentScopeJavaBot, line 296: [praise] Clean design: reusing `stager.stage(List.of(), sourceNamespaces)` to trigger orphan GC through the existing public API avoids expanding `MarketplaceStager`'s API surface (no need to widen `garbageCollectOrphans` visibility). The `loadFailed` guard is a well-thought-out outage protection — an empty result caused by a transient error must never wipe the cache.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a real and impactful bug: when all skills are removed from repositories or filtered out by visibility/skill filters, stale content in `.skills-cache/` was never cleaned up, causing outdated skill resources to leak into the sandbox. The fix introduces a `MergeResult` record that carries both the merged skills map and a `loadFailed` flag, enabling a new `stageEmptyRetainedIfSafe()` helper that triggers orphan GC with an empty retained set — but only when the empty result is intentional (not caused by a transient repository failure). The outage-protection design is thoughtful, the code reuse of `stager.stage(List.of(), sourceNamespaces)` is clean (no API surface expansion needed), and the 5-test suite covers the key scenarios well including the critical load-failure preservation case.

oss-maintainer:

This PR currently conflicts with `main` and cannot be merged. Please rebase or merge `main` into your branch and resolve the conflicts:

```bash
git fetch origin
git checkout main
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
```diff
@@ -275,8 +284,19 @@ public Mono<String> onSystemPrompt(Agent agent, RuntimeContext ctx, String curre
     //  Internals
     // ---------------------------------------------------------------------
 
-    private Map<String, RepoBound> skillsForCall(RuntimeContext ctx) {
-        return frozenSkills != null ? frozenSkills : mergeRepositories(ctx);
+    private MergeResult skillsForCall(RuntimeContext ctx) {
+        return frozenSkills != null ? new MergeResult(frozenSkills, false) : mergeRepositories(ctx);
```
oss-maintainer, line 288: **[Warning]** The frozen path hardcodes `loadFailed = false`, so the outage protection is lost exactly where it matters. If a repository throws during the construction-time `mergeRepositories(RuntimeContext.empty())`, the constructor still stores the (possibly empty) `frozenSkills` map and the failure flag is dropped — every later `skillsForCall()` then reports `loadFailed() == false` and `stageEmptyRetainedIfSafe` wipes `.skills-cache` with an empty white-list. Suggest capturing the flag once at construction (e.g. keep a `frozenLoadFailed` field set from the same `MergeResult`) and returning it here.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
```diff
@@ -275,8 +284,19 @@ public Mono<String> onSystemPrompt(Agent agent, RuntimeContext ctx, String curre
...
+    }
+
+    /**
+     * Runs orphan GC with an empty retained white-list so stale {@code .skills-cache} content
+     * is removed when no skill survives merging/filtering. Skipped when a repository failed to
+     * load — an empty result caused by an error must not wipe the cache (outage protection).
+     */
+    private void stageEmptyRetainedIfSafe(boolean loadFailed) {
```
oss-maintainer, line 296: **[Critical]** `stager.stage(List.of(), sourceNamespaces)` uses the two-arg overload, i.e. `scope == null` -> the shared `.skills-cache/<SHARED_SCOPE>/` subtree. `main` has since moved this class to per-isolation-scope staging (`stager.stage(enabled, sourceNamespaces, scopeKeyFor(ctx))` at lines 257/284), so when you rebase: (a) the empty white-list must be applied to *the same scope subtree the skills were staged into*, otherwise the stale per-user content this PR wants to clear is never reached; and (b) an empty white-list against the shared subtree deletes every non-retained directory there, which is exactly the cross-session hazard the scoped layout was introduced to remove. Please thread `ctx` (or the scope key) into `stageEmptyRetainedIfSafe` and call the three-arg overload.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
```diff
@@ -206,15 +208,19 @@ public void prestageMarketplaceSkills(RuntimeContext ctx) {
         if (ctx == null) {
             ctx = RuntimeContext.empty();
         }
-        Map<String, RepoBound> merged = skillsForCall(ctx);
+        MergeResult merge = skillsForCall(ctx);
+        Map<String, RepoBound> merged = merge.skills();
         if (merged.isEmpty()) {
+            stageEmptyRetainedIfSafe(merge.loadFailed());
```
oss-maintainer, line 214: **[Info]** This turns prompt building into a filesystem-mutating step: with `onSystemPrompt` (lines 237/245) also calling it, every system-prompt assembly for a skill-less agent now walks `.skills-cache` and touches mtimes. It is best-effort inside `MarketplaceStager` (exceptions are caught and logged), but it is still per-call directory I/O on the hot path. Consider running the empty GC once per middleware instance/scope, or only from `prestageMarketplaceSkills`, which is already the sandbox-staging hook.

oss-maintainer (commented):
### Summary
Good catch on the real hole: `prestageMarketplaceSkills` / `onSystemPrompt` returned early whenever the merged or filtered skill set was empty, so orphan GC never ran and stale `.skills-cache` content stayed projected into the sandbox. The `loadFailed` out-guard is the right instinct (an empty result caused by a repository outage must not wipe the cache), and the 294-line test covers empty-repository, filtered-out, and load-failure cases.

Two things need to change before this can land, plus a rebase (the PR currently reports `CONFLICTING` against `main`).

### Findings
- **[Critical]** HarnessSkillMiddleware.java:296 — GC runs through the two-arg `stage(...)` overload, i.e. against the shared scope subtree, while staging is per `scopeKeyFor(ctx)`; stale per-scope content is not cleared and the shared subtree is GC'd with an empty white-list.
- **[Warning]** HarnessSkillMiddleware.java:288 — frozen mode hardcodes `loadFailed = false`, so a construction-time repository failure loses the outage protection and wipes the cache.
- **[Info]** HarnessSkillMiddleware.java:214 — empty GC now runs on every `onSystemPrompt` call; consider once-per-scope or restricting it to the prestage hook.

### Suggestions
```java
private void stageEmptyRetainedIfSafe(RuntimeContext ctx, boolean loadFailed) {
    if (stager != null && !loadFailed) {
        stager.stage(List.of(), sourceNamespaces, scopeKeyFor(ctx));
    }
}
```
and in the constructor:
```java
MergeResult frozen = mergeRepositories(RuntimeContext.empty());
this.frozenSkills = freezeRepositories ? Collections.unmodifiableMap(new LinkedHashMap<>(frozen.skills())) : null;
this.frozenLoadFailed = freezeRepositories && frozen.loadFailed();
```
The test would be stronger with an assertion that the empty GC targets the same scope directory the non-empty staging wrote to — that is the case the current unit tests cannot catch because they use the default (shared) scope.

### Notes
CLA status for the current head: `success`. This PR is `CONFLICTING`/`DIRTY` against `main` (a rebase reminder was already posted earlier), so please rebase onto current `main` before the next review round.

---
*Automated review by @BOT_SIG*@

oss-maintainer (commented):
### Summary
Delta re-review of `67d60604` (single rebased commit; the branch history was rewritten since the `104ee537` review, so this is effectively a fresh review of the whole change). Running orphan GC with an empty retained white-list when nothing survives merging is the right fix for stale cache content being projected into a sandbox, the `loadFailed` outage guard is the right instinct, scope confinement plus the 30-minute grace is respected, and the 340-line test covers repository-empty, filter-empty, null-stager, outage and grace-window cases. Two gaps are worth closing before merge, both in the guard conditions rather than in the GC itself.

### Findings
- **[Warning]** `HarnessSkillMiddleware.java:212` — the freeze path discards `loadFailed`, so a repository that fails at construction yields an empty `frozenSkills` reported as "no load failure" (line 331) and every subsequent call sweeps the cache — the outage case the flag exists to protect.
- **[Warning]** `HarnessSkillMiddleware.java:261` / `:286` — the `enabled.isEmpty()` sweeps key off a *per-call* visibility/filter result while deleting from a *scope-wide* subtree, so a skill-less call can reclaim a directory a concurrent long-running call is still executing against.
- **[Info]** `HarnessSkillMiddleware.java:343` — the sweep now runs on every call for agents without marketplace skills; a latch would make it one walk per instance.

Inline comments carry the details and suggested fixes.

### Suggestions
For the two warnings: keep `loadFailed` across the freeze, and restrict the empty sweep to the case where the information is genuinely scope-wide (the repository returned nothing) rather than where this particular caller was shown nothing. Both keep the fix intact for the bug it addresses. CI and CLA are green, and the branch is mergeable.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
```diff
@@ -208,7 +208,8 @@ private HarnessSkillMiddleware(
         this.frozenSkills =
                 freezeRepositories
                         ? Collections.unmodifiableMap(
-                                new LinkedHashMap<>(mergeRepositories(RuntimeContext.empty())))
+                                new LinkedHashMap<>(
+                                        mergeRepositories(RuntimeContext.empty()).skills()))
```
oss-maintainer, line 212: **[Warning]** The outage protection is lost exactly where it is most needed: the freeze path keeps `.skills()` and throws away `loadFailed`. If a repository throws while `freezeRepositories` is on, `frozenSkills` becomes an empty map, and `skillsForCall` then hard-codes `new MergeResult(frozenSkills, false)` (line 331), so every later call looks like "the source genuinely has no skills" and runs the empty-retained sweep — wiping `.skills-cache/<scope>` that an earlier healthy call staged. That is the failure mode the `loadFailed` flag exists to prevent, just moved to construction time. Suggest storing the flag alongside the snapshot (`this.frozenSkills = result.skills(); this.frozenLoadFailed = result.loadFailed();`) and returning it from `skillsForCall`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
```diff
@@ -247,15 +249,19 @@ public void prestageMarketplaceSkills(RuntimeContext ctx) {
...
         if (merged.isEmpty()) {
+            stageEmptyRetainedIfSafe(ctx, merge.loadFailed());
             return;
         }
         List<RepoBound> visible = applyVisibility(merged.values(), ctx);
         List<RepoBound> enabled = applySkillFilter(visible, effectiveFilter(ctx));
-        if (!enabled.isEmpty()) {
-            stager.stage(enabled, sourceNamespaces, scopeKeyFor(ctx));
+        if (enabled.isEmpty()) {
+            stageEmptyRetainedIfSafe(ctx, merge.loadFailed());
```
oss-maintainer, line 261: **[Warning]** These two `enabled.isEmpty()` sweeps (here and in `onSystemPrompt`, line 286) are triggered by a *per-call* decision, not by the repository going empty. `applyVisibility` and `applySkillFilter` are caller-scoped, so one call whose filter hides every skill (a subagent with no skills enabled, a tenant restriction) sweeps the whole isolation scope, and the default `IsolationScope.USER` puts every session of that user in one subtree. A skill directory staged more than `DEFAULT_ORPHAN_GRACE` (30 min) ago by a *concurrent long-running call* is not in this call's white-list, so it gets deleted while that call is still executing against it — and long-running calls are the deployment this framework is built for. The `merged.isEmpty()` sweeps (lines 255/278) do not have this problem: "no skill exists at the source" is scope-wide information. Recommend sweeping only on repository emptiness, and for the filtered case instead passing the *unfiltered* `merged` values as the retained white-list (`stager.stage(merged.values(), ...)`) so hidden-but-still-published content is preserved without being re-staged... or simply skipping the sweep and letting the next non-empty call GC.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
```diff
@@ -318,8 +327,22 @@ public Mono<String> onSystemPrompt(Agent agent, RuntimeContext ctx, String curre
...
+    /**
+     * Runs orphan GC with an empty retained white-list so stale {@code .skills-cache} content
+     * is removed when no skill survives merging/filtering. The sweep is confined to this
+     * call's isolation scope and still honours
+     * {@link MarketplaceStager#DEFAULT_ORPHAN_GRACE}: a directory another call staged within
+     * the grace window is never mistaken for an orphan. Skipped when a repository failed to
+     * load — an empty result caused by an error must not wipe the cache (outage protection).
+     */
+    private void stageEmptyRetainedIfSafe(RuntimeContext ctx, boolean loadFailed) {
+        if (stager != null && !loadFailed) {
```
oss-maintainer, line 343: **[Info]** `stageEmptyRetainedIfSafe` is on the per-call system-prompt path, so for an agent that has no marketplace skills at all this now does a `Files.list` walk of `.skills-cache/<scope>` (plus the namespace subtrees) on every call, where before the empty case returned immediately. Harmless in absolute terms, but it is per-token-path work for the common no-skills configuration. A latch (`AtomicBoolean emptySweepDone`, reset when a merge yields a non-empty set) would make it one walk per instance instead of one per call; it would also pair naturally with the `frozenLoadFailed` flag above.

PR #2667, opened 2026-08-11
feat(harness): add deliver_artifact tool for sandboxed agents

### LINKED ISSUE #2663: [Bug]: Sandbox prompt references non-existent upload/download tools — agents lack an artifact delivery tool

In sandbox mode, `WorkspaceContextMiddleware` renders a `## Workspace` paragraph that
instructs the model to "use the upload/download tools to move files across the
boundary" (`WorkspaceContextMiddleware.java`, sandbox branch). However, the harness
exposes no upload/download `AgentTool` to the model at all:

- `FilesystemTool` registers only `read_file`, `write_file`, `edit_file`,
  `grep_files`, `glob_files`, `list_files`.
- File transfer into/out of the sandbox (`SandboxFileTransfer`, base64-via-exec) is
  an internal mechanism, invisible to the LLM.

So the instruction is a dead reference: the model is told to use tools that do not
exist, and any artifact it produces inside the sandbox has no supported way to leave
it.

My view: the agent does not actually need a generic upload/download tool. What is
missing is an **artifact delivery tool** — a first-class tool that takes a file
produced in the sandbox workspace and explicitly delivers it to a destination
outside the sandbox (e.g. the host, or a project's artifact store).

Concretely:

- An SPI interface such as `ArtifactDeliveryTarget`
  (`deliver(RuntimeContext, ArtifactDeliveryRequest) -> ArtifactDeliveryResult`)
  so the framework stays business-agnostic and the destination logic (e.g. WebDAV
  upload) is implemented by the application.
- A generic `deliver_artifact` tool in the harness (same style as `FilesystemTool`,
  `@Tool`-annotated), which downloads the file from the sandbox filesystem and
  delegates delivery to the configured target.
- Update the sandbox branch of `buildWorkspaceParagraph` to reference the actual
  tool name instead of the non-existent "upload/download tools".

**Additional context**

Deployment scenario: a sandboxed agent that produces reports/documents and must hand
them off to a project's artifact store (e.g. WebDAV). Currently there is no
supported path for this.

chcodex:
Fixes #2663

## Problem

In sandbox mode, `WorkspaceContextMiddleware` renders a `## Workspace` paragraph telling the model to "use the upload/download tools to move files across the boundary". The harness exposes no such `AgentTool` — `FilesystemTool` only registers `read_file`/`write_file`/`edit_file`/`grep_files`/`glob_files`/`list_files`, and `SandboxFileTransfer` is an internal mechanism invisible to the LLM. The instruction is a dead reference: an agent that produces an artifact inside a sandbox has no supported way to hand it out.

## Change

Introduces a generic, business-agnostic artifact delivery path for sandboxed agents:

- **`ArtifactDeliveryTarget`** SPI — application implements `deliver(RuntimeContext, ArtifactDeliveryRequest) -> ArtifactDeliveryResult` and owns destination logic (WebDAV, object store, project artifact store, …).
- **`deliver_artifact` tool** — registered via `HarnessAgent.Builder.artifactDeliveryTarget(...)`. It downloads the file bytes from the agent filesystem and delegates transport to the target.
- **Sandbox prompt** — when a target is configured, the workspace paragraph tells the model to use `deliver_artifact`; otherwise it states plainly that there is no mechanism for moving files across the boundary. The old dead reference to "upload/download tools" is gone.
- **Safety/limits**:
  - The tool is exposed only to the **main agent**, not to automatically constructed subagents (they return plain text results for the main agent to deliver).
  - It is suppressed when `disableFilesystemTools()` is set, keeping file-access isolation intact.
  - `fileName` must be a plain file name — path separators, `.` and `..` are rejected, and target names are derived from a cleaned basename.
  - `ArtifactDeliveryRequest.fileName()` is documented as already validated, so targets can resolve it directly.

## Tests

- `ArtifactDeliveryToolTest` — download/forward, defaults, normalization, failures, conflict advice, and new fileName validation cases.
- `WorkspaceContextMiddlewareSandboxPromptTest` — both sandbox prompt branches.
- `HarnessAgentTest` — registration, `disableFilesystemTools` interaction, main-agent-only exposure to subagents, and prompt integration.


AgentScopeJavaBot (commented):
🤖 AI Review

This PR introduces a generic, business-agnostic artifact delivery mechanism for sandboxed agents. It adds an `ArtifactDeliveryTarget` SPI, a `deliver_artifact` tool registered only on the main agent, and corresponding sandbox prompt updates. The design is clean: the framework stays transport-agnostic (WebDAV, object store, etc. are implemented by the application), the tool validates `fileName` against path traversal, and the builder correctly suppresses the tool when `disableFilesystemTools` is set. Test coverage is solid with 13+ unit tests covering happy paths, input validation, error branches, and subagent isolation. The PR resolves #2663 appropriately.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ArtifactDeliveryTool.java
```diff
@@ -0,0 +1,186 @@
...
+        if (effectiveFileName.isBlank()) {
+            return "Error: unable to determine a target file name from filePath '" + filePath + "'";
+        }
+        if (!isPlainFileName(effectiveFileName)) {
+            return "Error: fileName must be a plain file name without path separators, '.' or '..'";
+        }
+        boolean effectiveForce = Boolean.TRUE.equals(force);
+
+        List<FileDownloadResponse> responses =
+                filesystem.downloadFiles(runtimeContext, List.of(normalized));
```
AgentScopeJavaBot, line 126: [major] **No file size guard.** `downloadFiles()` loads the entire file into a `byte[]` which is then wrapped in `ArtifactDeliveryRequest` and forwarded to the target. The tool description explicitly mentions "archives" as a use case. A sandboxed agent could produce a very large file and trigger OOM or excessive GC pressure. Consider adding a configurable `maxArtifactSizeBytes` on the builder and returning an error before the download when the filesystem can report size, or after download if `response.content().length` exceeds the limit.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1934,6 +1937,25 @@ public Builder enableAgentTracingLog(boolean enabled) {
...
+         *
+         * <p>When set, the generic {@link ArtifactDeliveryTool} is registered on the main agent and
+         * the sandbox workspace prompt tells the model to use it to hand artifacts it produced to a
+         * destination outside the sandbox. When unset (default), no delivery tool is exposed.
+         *
+         * <p>The tool is exposed only to the main agent — it is not propagated to automatically
+         * constructed subagents, which return plain text results for the main agent to deliver.
+         *
+         * <p>Note: the tool reads files from the agent filesystem, so it is also suppressed when
+         * {@link #disableFilesystemTools()} is used — combine the two only when you intend the
```
AgentScopeJavaBot: [nitpick] **Self-contradictory Javadoc.** The note reads: *"it is also suppressed when `disableFilesystemTools()` is used — combine the two only when you intend the delivery tool to remain available."* The second clause implies the tool *will* remain available when combining both, but the code actually suppresses it. Suggested rewrite to clarify the suppression semantics.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ArtifactDeliveryTool.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private static boolean isPlainFileName(String name) {
+        return name.indexOf('/') < 0
+                && name.indexOf('\\') < 0
+                && !name.equals(".")
+                && !name.equals("..");
+    }
+
+    private static String basename(String path) {
```
AgentScopeJavaBot, line 175: [nitpick] **`isPlainFileName` does not reject NUL bytes.** While LLMs rarely produce `\0`, some target implementations (C-based file APIs, certain object stores) truncate file names at NUL. Adding `&& name.indexOf('\0') < 0` would close this edge case cheaply.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ArtifactDeliveryTool.java
AgentScopeJavaBot, line 126: [major] **No file size guard.** `downloadFiles()` loads the entire file into a `byte[]` which is then wrapped in `ArtifactDeliveryRequest` and forwarded to the target. The tool description explicitly mentions "archives" as a use case. A sandboxed agent could produce a very large file and trigger OOM or excessive GC pressure. Consider adding a configurable `maxArtifactSizeBytes` on the builder and returning an error before the download when the filesystem can report size, or after download if `response.content().length` exceeds the limit.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
AgentScopeJavaBot: [nitpick] **Self-contradictory Javadoc.** The note reads: *"it is also suppressed when `disableFilesystemTools()` is used — combine the two only when you intend the delivery tool to remain available."* The second clause implies the tool *will* remain available when combining both, but the code actually suppresses it. Suggested rewrite to clarify the suppression semantics.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ArtifactDeliveryTool.java
AgentScopeJavaBot, line 175: [nitpick] **`isPlainFileName` does not reject NUL bytes.** While LLMs rarely produce `\0`, some target implementations (C-based file APIs, certain object stores) truncate file names at NUL. Adding `&& name.indexOf('\0') < 0` would close this edge case cheaply.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR introduces a generic, business-agnostic artifact delivery mechanism for sandboxed agents. It adds an `ArtifactDeliveryTarget` SPI, a `deliver_artifact` tool registered only on the main agent, and corresponding sandbox prompt updates. The design is clean: the framework stays transport-agnostic (WebDAV, object store, etc. are implemented by the application), the tool validates `fileName` against path traversal, and the builder correctly suppresses the tool when `disableFilesystemTools` is set. Test coverage is solid with 13+ unit tests covering happy paths, input validation, error branches, and subagent isolation. The PR resolves #2663 appropriately.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ArtifactDeliveryTool.java
chcodex: Thanks for the review. This concern is acknowledged, but it is **not introduced by this PR**: `AbstractFilesystem.downloadFiles(...)` returning `List<FileDownloadResponse>` with a `byte[]` content is the framework's existing contract (also used by `WorkspaceSkillRepository`), and `deliver_artifact` simply reuses that API without adding new in-memory amplification. A real fix needs a streaming download API at the filesystem layer plus streaming support in sandbox backends — a framework-level design change beyond the scope of this PR. We will track it as a follow-up (streaming artifact delivery); this PR intentionally leaves it unchanged.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ArtifactDeliveryTool.java
chcodex: fixed

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
chcodex: fixed

AgentScopeJavaBot: ✅ Verified as addressed in `8b3bd29`: Javadoc rewritten to clearly state Combining both leaves the tool unregistered, resolving the self-contradiction.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ArtifactDeliveryTool.java
AgentScopeJavaBot: ✅ Verified as addressed in `8b3bd29`: isPlainFileName now includes name.indexOf('\0') < 0 check, exactly as suggested.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
AgentScopeJavaBot: ✅ Verified as addressed in `8b3bd29`: Javadoc rewritten to clearly state Combining both leaves the tool unregistered, resolving the self-contradiction.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ArtifactDeliveryTool.java
AgentScopeJavaBot: ✅ Verified as addressed in `8b3bd29`: isPlainFileName now includes name.indexOf('\0') < 0 check, exactly as suggested.

chcodex:
image
@jujn  Actual test results are OK

Buktal (approved):
LGTM

PR #2511, opened 2026-07-31
feat(HITL): Emit UserConfirmResultEvent when resuming permission HITL

jujn:
## Summary

- Emit `UserConfirmResultEvent` when `ReActAgent` accepts `ConfirmResult` payloads while resuming permission HITL.
- Persist the pending confirmation request `replyId` on the live assistant message so the result event can correlate with the earlier `RequireUserConfirmEvent`.
- Validate that resume `ConfirmResult`s fully cover all current `ASKING` tool calls before emitting the event or mutating context.
- Update HITL tests and event docs for the new resume-event behavior.

## Checklist

Please check the following items before code is ready to be reviewed.

- [ ]  Code has been formatted with `mvn spotless:apply`
- [ ]  All tests are passing (`mvn test`)
- [ ]  Javadoc comments are complete and follow project conventions
- [ ]  Related documentation has been updated (e.g. links, examples, etc.)
- [ ]  Code is ready for review

PR #2668, opened 2026-08-11
fix(core): 将用户拒绝工具执行原因写入模型上下文

hanydd:
## 背景

现在`ReactAgent`和`HarnessAgent`已经有工具审批逻辑，发出 `RequireUserConfirmEvent` 和 `UserConfirmResultEvent`事件，用户可以选择通过或者拒绝，但是没有渠道可以把用户主动提供的拒绝调用的原因。已合并的 #2511 建立了工具审批结果的事件和恢复链路，但 `ConfirmResult` 只能表达通过或拒绝，拒绝时写入模型上下文的仍是固定文案，无法携带用户主动提供的原因。这个PR拓展了一下工具调用审批需要的 `ConfirmResult`，加入了可选的`reason`字段，并且将原因加入到了对应的ToolResultBlock里面。

## 改动

- 为 `ConfirmResult` 增加可选 `reason`，保留旧构造函数。
- 拒绝时将原因追加到 `DENIED` 工具结果，使模型下一轮可见。
- 补充 ReActAgent 上下文测试和 Harness 远程事件编解码测试。

## 测试

`ReActAgentHitlTest`、`RemoteEventCodecPassthroughTest` 通过。

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -1963,8 +1963,9 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
          */
         private void applyConfirmResults(List<ConfirmResult> results) {
             // Replace ASKING ToolUseBlocks with possibly-modified ones from the user, and
-            // promote them to ALLOWED. Collect denied ones for separate handling.
-            List<ToolUseBlock> deniedToolCalls = new ArrayList<>();
+            // promote them to ALLOWED. Collect denied results for separate handling so any
+            // user-supplied reason can be included in the denied tool result.
+            List<ConfirmResult> deniedResults = new ArrayList<>();
             Map<String, ToolUseBlock> replacements = new HashMap<>();
             for (ConfirmResult r : results) {
                 ToolUseBlock target = r.getToolCall();
@@ -1981,18 +1982,25 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                         }
                     }
                 } else {
-                    deniedToolCalls.add(target);
+                    deniedResults.add(r);
                 }
             }
             applyToolUseBlockReplacements(replacements);
-            for (ToolUseBlock denied : deniedToolCalls) {
-                ToolResultBlock deniedResult =
-                        ToolResultBlock.text("Permission denied by user")
+            for (ConfirmResult deniedResult : deniedResults) {
+                ToolUseBlock denied = deniedResult.getToolCall();
+                String reason = deniedResult.getReason();
+                String message =
+                        reason == null || reason.isBlank()
+                                ? "Permission denied by user"
+                                : "Permission denied by user. User-provided reason: "
+                                        + reason.strip();
+                ToolResultBlock deniedToolResult =
+                        ToolResultBlock.text(message)
                                 .withIdAndName(denied.getId(), denied.getName())
                                 .withState(ToolResultState.DENIED);
                 Msg deniedMsg =
                         ToolResultMessageBuilder.buildToolResultMsg(
-                                deniedResult, denied, getName());
+                                deniedToolResult, denied, getName());
                 state.contextMutable().add(deniedMsg);
             }
         }
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -217,6 +217,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
     private static final Logger log = LoggerFactory.getLogger(ReActAgent.class);
     private static final GracefulShutdownManager shutdownManager =
             GracefulShutdownManager.getInstance();
+    private static final int MAX_CONFIRM_REASON_CODE_POINTS = 500;
 
     /** Tool name used for the per-call structured-output {@code generate_response} tool. */
     public static final String STRUCTURED_OUTPUT_TOOL_NAME = "generate_response";
@@ -1989,11 +1990,18 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
             for (ConfirmResult deniedResult : deniedResults) {
                 ToolUseBlock denied = deniedResult.getToolCall();
                 String reason = deniedResult.getReason();
+                String normalizedReason = reason == null ? "" : reason.strip();
+                if (normalizedReason.codePointCount(0, normalizedReason.length())
+                        > MAX_CONFIRM_REASON_CODE_POINTS) {
+                    int endIndex =
+                            normalizedReason.offsetByCodePoints(0, MAX_CONFIRM_REASON_CODE_POINTS);
+                    normalizedReason = normalizedReason.substring(0, endIndex) + " <truncated>";
+                }
                 String message =
-                        reason == null || reason.isBlank()
+                        normalizedReason.isEmpty()
                                 ? "Permission denied by user"
                                 : "Permission denied by user. User-provided reason: "
-                                        + reason.strip();
+                                        + normalizedReason;
                 ToolResultBlock deniedToolResult =
                         ToolResultBlock.text(message)
                                 .withIdAndName(denied.getId(), denied.getName())
--- a/agentscope-core/src/main/java/io/agentscope/core/event/ConfirmResult.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/event/ConfirmResult.java
@@ -16,6 +16,7 @@
 package io.agentscope.core.event;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.agentscope.core.message.ToolUseBlock;
 import io.agentscope.core.permission.PermissionRule;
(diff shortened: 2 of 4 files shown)


agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1954,18 +1955,25 @@ private void applyConfirmResults(List<ConfirmResult> results) {
...
             }
             applyToolUseBlockReplacements(replacements);
-            for (ToolUseBlock denied : deniedToolCalls) {
-                ToolResultBlock deniedResult =
-                        ToolResultBlock.text("Permission denied by user")
+            for (ConfirmResult deniedResult : deniedResults) {
+                ToolUseBlock denied = deniedResult.getToolCall();
+                String reason = deniedResult.getReason();
+                String message =
+                        reason == null || reason.isBlank()
```
AgentScopeJavaBot: [nitpick] No length cap on `reason` before embedding it into the tool result text. A buggy or malicious UI could pass a very long user-typed string, wasting LLM context tokens. Consider truncating to a reasonable limit (e.g. 500 chars) before concatenation.

agentscope-core/src/main/java/io/agentscope/core/event/ConfirmResult.java
```diff
@@ -33,20 +34,28 @@ public class ConfirmResult {
...
     private final ToolUseBlock toolCall;
     private final List<PermissionRule> rules;
+    private final String reason;
 
     @JsonCreator
     public ConfirmResult(
             @JsonProperty("confirmed") boolean confirmed,
             @JsonProperty("toolCall") ToolUseBlock toolCall,
-            @JsonProperty("rules") List<PermissionRule> rules) {
+            @JsonProperty("rules") List<PermissionRule> rules,
```
AgentScopeJavaBot, line 43: [nitpick] The `@JsonCreator` constructor accepts `reason` for both `confirmed == true` and `confirmed == false` cases. The Javadoc correctly notes it's primarily meaningful for denials, but there is no validation or logging if a reason is supplied with `confirmed == true`. Not a bug — the field is simply ignored in the allowed path — but a brief note in the constructor Javadoc would help future maintainers understand the intended asymmetry.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR adds an optional `reason` field to `ConfirmResult` so that when a user denies a tool call in the HITL (Human-In-The-Loop) flow, their explanation is propagated into the model's context as part of the DENIED tool result message. The implementation is backward-compatible (new overloaded constructors, `@JsonInclude(NON_NULL)` on the getter, `@JsonCreator` handles missing `reason` as null). Two well-targeted tests cover the core path (reason appears in denied tool result text) and the codec round-trip (reason survives serialization). The change is small, focused, and solves a real usability gap — without it, the model has no idea *why* the user rejected a tool call.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot: [nitpick] No length cap on `reason` before embedding it into the tool result text. A buggy or malicious UI could pass a very long user-typed string, wasting LLM context tokens. Consider truncating to a reasonable limit (e.g. 500 chars) before concatenation.

agentscope-core/src/main/java/io/agentscope/core/event/ConfirmResult.java
AgentScopeJavaBot, line 43: [nitpick] The `@JsonCreator` constructor accepts `reason` for both `confirmed == true` and `confirmed == false` cases. The Javadoc correctly notes it's primarily meaningful for denials, but there is no validation or logging if a reason is supplied with `confirmed == true`. Not a bug — the field is simply ignored in the allowed path — but a brief note in the constructor Javadoc would help future maintainers understand the intended asymmetry.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR adds an optional `reason` field to `ConfirmResult` so that when a user denies a tool call in the HITL (Human-In-The-Loop) flow, their explanation is propagated into the model's context as part of the DENIED tool result message. The implementation is backward-compatible (new overloaded constructors, `@JsonInclude(NON_NULL)` on the getter, `@JsonCreator` handles missing `reason` as null). Two well-targeted tests cover the core path (reason appears in denied tool result text) and the codec round-trip (reason survives serialization). The change is small, focused, and solves a real usability gap — without it, the model has no idea *why* the user rejected a tool call.

hanydd:
@chickenlj 大佬可以看下这个PR吗？改动不大，主要是加了工具审批的时候，可以让调用方把用户拒绝工具执行的原因传进来。目前项目中用到了这个功能，现在只能在外面再加适配层，才能把用户的原因拼进去。

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
hanydd: 加了500字符的长度限制

oss-maintainer:

```bash
git fetch upstream
git checkout feat/tool-approval-refusal-reason
git rebase upstream/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2029,21 +2031,34 @@ private void applyConfirmResults(List<ConfirmResult> results, String replyId) {
...
             MessageUtils.replaceToolUseBlocks(state.contextMutable(), replacements);
-            for (ToolUseBlock denied : deniedToolCalls) {
-                ToolResultBlock deniedResult =
-                        ToolResultBlock.text(PERMISSION_DENIED_BY_USER)
+            for (ConfirmResult deniedResult : deniedResults) {
+                ToolUseBlock denied = deniedResult.getToolCall();
+                String reason = deniedResult.getReason();
+                String normalizedReason = reason == null ? "" : reason.strip();
+                if (normalizedReason.codePointCount(0, normalizedReason.length())
+                        > MAX_CONFIRM_REASON_CODE_POINTS) {
```
oss-maintainer, line 2043: Small drift risk: the no-reason branch re-types the literal `"Permission denied by user"` while the rule-based path right above still goes through `PERMISSION_DENIED_BY_RULES`, and `PERMISSION_DENIED_BY_USER` is still the constant the rest of the code (and the pre-existing assertion in `askingToolResumeWithDeniedConfirmResultProducesDeniedToolResult`) refers to. Keeping one source of truth here means a future wording change can't leave the two denial paths saying different things:

```java
String message =
        normalizedReason.isEmpty()
                ? PERMISSION_DENIED_BY_USER
                : PERMISSION_DENIED_BY_USER + ". User-provided reason: " + normalizedReason;
```

(If the constant's wording is deliberately changing for the reason case, then it is the *other* usages that should move with it, ideally in this PR.)

agentscope-core/src/main/java/io/agentscope/core/event/ConfirmResult.java
```diff
@@ -66,4 +75,18 @@ public ToolUseBlock getToolCall() {
...
+    /**
+     * Optional user-supplied reason for the confirmation decision.
+     *
+     * <p>When {@link #confirmed} is false, the agent includes this reason in the denied tool result
+     * that is added to model context. It is otherwise retained on the confirmation event for
+     * callers that need to inspect it.
+     *
+     * @return decision reason, or null if none was supplied
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
oss-maintainer, line 88: The javadoc says the reason is included in the denied tool result "that is added to model context" — but no transport in this repo populates it yet, so on every path a user actually hits today the new field is always `null`:

- `AguiMessageConverter.java:473` → `new ConfirmResult(approved, toolUseBlock)`
- `AgentProtocolTaskStore.java:395` → `new ConfirmResult(approved, toolCall)`
- `HarnessAgentTaskStarter.java:382` → `new ConfirmResult(decision.allow(), call)`
- plus the `documentation2/hitl` examples

I checked because the `reason` value is the point of the whole change, and the only tests that exercise it build `ConfirmResult` by hand. So the core plumbing is right, but the feature is dark until at least one caller reads a reason off the wire. Could this PR (or a linked follow-up) wire the AG-UI side, whose resume payload is where a human denial comment naturally lives? If the intent was core-only on purpose, please say so in the PR description so a reviewer doesn't have to trace five constructors to find out.

Also worth one line in this javadoc: on the `agentscope-extensions-aistio` path `confirmed` comes from a *remote* decision, so `reason` is not necessarily human-authored text. It now enters both the model context and the `ToolResultTextDeltaEvent` stream, which is a prompt-injection / log-surfacing surface. Not an objection — the 500-code-point cap is the right instinct — but the trust boundary should be documented here, since `review_context` calls out credentials leaking into logs/events for this repo.

oss-maintainer (commented):
### Summary
Adds an optional `reason` to `ConfirmResult` and, on a denied confirmation, feeds `Permission denied by user. User-provided reason: <reason>` into the denied `ToolResultBlock` (model context) and the matching `ToolResultTextDeltaEvent` (stream), so the model can react to *why* a human refused instead of only *that* they refused. Good change, well executed at the core level — and the truncation is done with `codePointCount` / `offsetByCodePoints` rather than `substring`, so an astral character like `😀` cannot be cut in half; the `499 a's + emoji` test proves the author noticed that trap. The new assertions that compare the context copy against the streamed delta are the right way to test a change that writes the same text twice, and the round-trip test on `RemoteEventCodec` closes the distributed path.

Two things to settle before this is useful end to end:

- **[Warning]** No caller in the repo actually passes a `reason` (`AguiMessageConverter`, `AgentProtocolTaskStore`, `HarnessAgentTaskStarter`, the HITL examples all use the 2-arg constructor), so the feature is dead code on every user-facing path until a transport carries it. And `reason` may originate from a *remote* decision, not a human — it lands in model context and the event stream, so the trust boundary deserves a javadoc line.
- **[Info]** The no-reason branch duplicates the `PERMISSION_DENIED_BY_USER` literal instead of using the constant.

If this PR is intentionally core-only, split the transport wiring into a follow-up and link it. CLA signed; CI green on this head (ubuntu + windows build, License, Module Sync, codecov).

---

Issue #2490, opened 2026-07-30
[Bug]: Concurrent calls on one agent corrupt each other's sandbox binding (single-slot `SandboxBackedFilesystem` / `SandboxLifecycleMiddleware`)


xseruer:
**Describe the bug**

`SandboxBackedFilesystem` and `SandboxLifecycleMiddleware` both assume that a given agent instance runs **at most one call at a time**, and store per-call sandbox state in **agent-level single slots**:

- `SandboxBackedFilesystem` is a stable proxy created at agent build time and shared by all tool executions of the agent. The live sandbox is injected into a single field:

  ```java
  // agentscope-harness/.../filesystem/sandbox/SandboxBackedFilesystem.java
  private volatile Sandbox sandbox;   // injected per call, cleared on release
  ```

- `SandboxLifecycleMiddleware` keeps the acquire result of "the current call" in a single reference:

  ```java
  // agentscope-harness/.../middleware/SandboxLifecycleMiddleware.java
  private final AtomicReference<SandboxAcquireResult> currentAcquireResult =
          new AtomicReference<>();
  ```

When the same agent bean is called concurrently for **different sessions** (a common setup: one shared `ReActAgent`/`HarnessAgent` bean serving many users/conversations — the SDK's `serializeOnKey` only serializes calls of the *same* `(userId, sessionId)`, cross-session calls run in parallel by design), the two slots are corrupted in three distinct ways:

1. **Sandbox cross-talk** — call B's `filesystemProxy.setSandbox(sbB)` overwrites call A's `sbA` while A is still running. All subsequent filesystem/tool operations of A silently land in **B's sandbox**: files leak across sessions, workspace projection of one conversation is visible to another.

2. **Spurious `No active sandbox` failures** — when A finishes first, `releaseForCall` executes `filesystemProxy.setSandbox(null)`. B is still mid-run; its next filesystem operation throws:

   ```
   io.agentscope.harness.agent.sandbox.SandboxException$SandboxConfigurationException:
   No active sandbox — sandbox filesystem used outside of a call context
   ```

3. **Wrong-pair release + sandbox session leak** — `releaseForCall` does `currentAcquireResult.getAndSet(null)`. If B's acquire has overwritten the slot, A's release will `persistState`/`release` **B's** `SandboxAcquireResult` (stopping the sandbox B is actively using), while A's own acquire result is lost and its sandbox session is **never released**.

[...]

PR #2675, opened 2026-08-12
 fix(harness): isolate sandbox binding per call to fix concurrent corruption

larry-zy:
Related: #2490

Summary
  
  SandboxLifecycleMiddleware and SandboxBackedFilesystem held the acquired sandbox in a single agent-level slot (an AtomicReference plus the proxy's volatile field). Since
  AgentBase.serializeOnKey only serialises calls with the same (userId, sessionId), calls from distinct sessions run in parallel on one agent bean and raced on that shared slot,
  causing two failures:

  - A call could execute against another session's sandbox.
  - A finishing call's release could tear down a still-live sibling sandbox.

  Changes

  - Bind the acquired SandboxAcquireResult per call on the invocation's RuntimeContext, and resolve it there first, so concurrent distinct-session calls stay isolated.
  - releaseForCall now reads back its own binding and tears down only its own sandbox.
  - Keep the volatile field as a best-effort fallback for context-free internal callers that resolve the filesystem with a shared empty RuntimeContext (e.g.
  WorkspaceMessageBus). To stop that fallback from being clobbered under concurrency, it's now maintained via a synchronized compare-and-clear (clearSandboxIfCurrent), so a
  releasing call never nulls a sibling's binding.

  Testing
  
  Adds SandboxLifecycleConcurrencyReproTest, which drives the legal A.acquire → B.acquire → A.use → A.release interleaving and asserts each call executes against its own sandbox
  and that A's release affects only A.
-Dtest='SandboxLifecycleConcurrencyReproTest,SandboxBackedFilesystemTest,SandboxLifecycleMiddlewareCallbackTest'

  All related tests pass (18 run, 0 failures, 0 errors).


AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes issue #2490 — concurrent calls for distinct (userId, sessionId) sessions on the same agent bean corrupted each other's sandbox binding. The root cause was a single agent-level slot (volatile Sandbox field + currentAcquireResult) shared across all concurrent calls. The fix moves the sandbox binding to the per-call RuntimeContext (typed attribute), with the volatile field retained as a best-effort fallback for context-free callers (e.g., WorkspaceMessageBus). A compareAndClear pattern (synchronized + identity check) prevents one call's release from nulling another's fallback. A deterministic concurrency test reproduces the exact interleaving and validates three isolation invariants. The fix is architecturally sound and correctly addresses the reported issue.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes issue #2490 — concurrent calls for distinct (userId, sessionId) sessions on the same agent bean corrupted each other's sandbox binding. The root cause was a single agent-level slot (volatile Sandbox field + currentAcquireResult) shared across all concurrent calls. The fix moves the sandbox binding to the per-call RuntimeContext (typed attribute), with the volatile field retained as a best-effort fallback for context-free callers (e.g., WorkspaceMessageBus). A compareAndClear pattern (synchronized + identity check) prevents one call's release from nulling another's fallback. A deterministic concurrency test reproduces the exact interleaving and validates three isolation invariants. The fix is architecturally sound and correctly addresses the reported issue.

xiaoyuuuuuupeng:
LGTM. This fixes the sandbox race by binding each sandbox to its own call context.

jujn (approved):

PR #2676, opened 2026-08-12
fix(core): propagate acting middleware result rewrites

### LINKED ISSUE #2672: [Bug]:middleware 工具重写失败
在当前2.0版本中，即便存在middleware，但还是将原始的工具结果直接返回大模型
目前只能通过hook进行结果改写，但hook在2.0版本被废弃，感觉middleware还没有完全能替改hook

Image

而且，io.agentscope.core.ReActAgent.CallExecution#actingStream中直接this::publishEvent 流式返回了，没有middleware处理的结果

dangzitou:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #2672.

`ReActAgent.actingStream` published raw tool-result events before the `onActing` middleware chain. The acting phase then discarded the post-middleware events and continued with the raw `resultHolder`, so a middleware rewrite was visible neither to event consumers nor to the next model call.

This change:

- publishes acting events after the middleware chain;
- detects tool-result delta/end events that middleware replaced or filtered;
- rebuilds only the affected `ToolResultBlock` values from the transformed stream;
- preserves the original result object when middleware passes all result events through unchanged.

The new end-to-end regression test rewrites `search:alpha` to `rewritten`. On current main it failed because the public stream returned `search:alpha`; with this change both the public stream and the second model request contain only `rewritten`.

## Tests

- `mvn -pl agentscope-core -Dtest=ReActAgentNewLoopE2ETest#actingMiddlewareCanRewriteToolResultForEventsAndNextModelCall test` — 1 passed
- `mvn -pl agentscope-core -Dtest=ReActAgentNewLoopE2ETest,ReActAgentMiddlewareIntegrationTest test` — 12 passed
- `mvn -pl agentscope-core test` — 2275 passed, 9 skipped
- Spotless check passed as part of the Maven builds

## Compatibility and risk

No public API signatures change. Existing tool results retain their original `ToolResultBlock` instance unless middleware actually changes the tool-result event sequence. Text and data deltas, metadata, and end state are retained when rebuilding a transformed result.

## Checklist

- [x] All relevant tests are passing
- [x] Related middleware documentation has been updated

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -2732,8 +2732,16 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                     .firePreActing(pendingToolCalls, toolkit)
                     .flatMap(
                             toolCalls -> {
+                                List<AgentEvent> coreToolResultEvents = new ArrayList<>();
+                                List<AgentEvent> transformedToolResultEvents = new ArrayList<>();
                                 Function<ActingInput, Flux<AgentEvent>> actingCore =
-                                        ai -> actingStream(ai.toolCalls(), replyId, resultHolder);
+                                        ai ->
+                                                actingStream(ai.toolCalls(), replyId, resultHolder)
+                                                        .doOnNext(
+                                                                event ->
+                                                                        addToolResultEvent(
+                                                                                coreToolResultEvents,
+                                                                                event));
                                 Flux<AgentEvent> stream =
                                         MiddlewareChain.build(
                                                         middlewares,
@@ -2744,11 +2752,20 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                                 .apply(new ActingInput(toolCalls));
                                 return stream.doOnNext(
                                                 ev -> {
+                                                    addToolResultEvent(
+                                                            transformedToolResultEvents, ev);
+                                                    publishEvent(ev);
                                                     if (ev instanceof RequestStopEvent rs) {
                                                         actingStopRequested.compareAndSet(null, rs);
                                                     }
                                                 })
-                                        .then(Mono.defer(() -> Mono.just(resultHolder.get())));
+                                        .then(
+                                                Mono.fromSupplier(
+                                                        () ->
+                                                                applyToolResultTransformations(
+                                                                        resultHolder.get(),
+                                                                        coreToolResultEvents,
+                                                                        transformedToolResultEvents)));
                             })
                     .flatMap(
                             results -> {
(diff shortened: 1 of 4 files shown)


AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes issue #2672 where ReActAgent.actingStream published raw tool-result events before the onActing middleware chain, causing middleware rewrites to be invisible to downstream subscribers and the next model iteration. The fix publishes events after the middleware chain, detects middleware modifications by comparing pre/post event references per tool call ID, and rebuilds ToolResultBlock values from transformed events when changes are detected. Two E2E tests verify the fix. The approach is correct and well-targeted.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes issue #2672 where ReActAgent.actingStream published raw tool-result events before the onActing middleware chain, causing middleware rewrites to be invisible to downstream subscribers and the next model iteration. The fix publishes events after the middleware chain, detects middleware modifications by comparing pre/post event references per tool call ID, and rebuilds ToolResultBlock values from transformed events when changes are detected. Two E2E tests verify the fix. The approach is correct and well-targeted.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2690,8 +2690,16 @@ private Mono<Msg> acting(int iter) {
                     .firePreActing(pendingToolCalls, toolkit)
                     .flatMap(
                             toolCalls -> {
+                                List<AgentEvent> coreToolResultEvents = new ArrayList<>();
+                                List<AgentEvent> transformedToolResultEvents = new ArrayList<>();
```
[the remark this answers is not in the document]
dangzitou: Addressed in acb2f8d3: replaced the two mutable `ArrayList` collectors with `CopyOnWriteArrayList`, so concurrent acting-stream callbacks cannot corrupt them. Existing CI remains green.

dangzitou:
Rebased onto current `upstream/main` and resolved the test-file conflict while preserving both regression cases (event-emitter restoration and middleware tool-result rewriting).

Local validation: `mvn -pl agentscope-core -am -Dtest=ReActAgentNewLoopE2ETest -Dsurefire.failIfNoSpecifiedTests=false test` — 5 tests passed. Please re-run/confirm repository CI and review when convenient.

dangzitou:
Remote CI completed with both platform jobs failing because the unrelated `agentscope-harness` test `JsonSessionDefaultLocationTest.perUserPartitioning_viaSharedAgentRoutedByRuntimeContext` could not delete a JUnit temp directory (`DirectoryNotEmptyException`); the core module and all 844 tests otherwise passed. This is an environment/test-cleanup failure outside the PR's changed files. I attempted to rerun failed jobs, but GitHub requires repository admin rights. No additional code change is justified from this failure.

dangzitou:
Fixed the CI break introduced by the latest `main` OpenAI Java SDK bump: the new `apiKey(String)` / `apiKey(Optional<String>)` overloads made Mockito `any()` ambiguous, so the matcher is now explicitly typed as `String`.

Validation:
- Targeted OpenAI embedding tests pass with both openai-java 4.28.0 and 4.52.0
- Full GitHub CI is green on Linux and Windows
- License and module-sync checks are green

@agentscope-ai maintainers, this is ready for review.

oss-maintainer:

```bash
git fetch upstream
git checkout agent/fix-middleware-tool-rewrite
git rebase upstream/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2809,6 +2829,107 @@ private Mono<Msg> acting(int iter) {
...
+            if (event instanceof ToolResultDataDeltaEvent dataDelta) {
+                return dataDelta.getToolCallId();
+            }
+            if (event instanceof ToolResultEndEvent endEvent) {
+                return endEvent.getToolCallId();
+            }
+            return null;
+        }
+
+        private boolean sameEventInstances(List<AgentEvent> left, List<AgentEvent> right) {
```
oss-maintainer, line 2914: **[Warning]** `sameEventInstances` decides "nothing was transformed" by **reference identity**. Any middleware that does `.map(...)` over the acting stream produces equal-but-new event instances, so every tool result then goes through the reconstruction path even though no middleware changed semantics. Reconstruction rebuilds `ToolResultBlock` solely from `ToolResultTextDeltaEvent` / `ToolResultDataDeltaEvent` / `ToolResultEndEvent`, so anything the original block carried that is not expressed by those three event types would be dropped. Could this instead compare the event *lists* by equality, or better, have `actingStream` record the per-tool result it put into `resultHolder` and only rebuild when the transformed stream for that `toolCallId` actually differs in content? A regression test that runs a pass-through `.map()` middleware and asserts byte-identical memory content would lock this down.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2809,6 +2829,107 @@ private Mono<Msg> acting(int iter) {
...
+        }
+
+        private List<Map.Entry<ToolUseBlock, ToolResultBlock>> applyToolResultTransformations(
+                List<Map.Entry<ToolUseBlock, ToolResultBlock>> results,
+                List<AgentEvent> coreEvents,
+                List<AgentEvent> transformedEvents) {
+            List<Map.Entry<ToolUseBlock, ToolResultBlock>> updated = new ArrayList<>();
+            for (Map.Entry<ToolUseBlock, ToolResultBlock> entry : results) {
+                String toolCallId = entry.getKey().getId();
+                List<AgentEvent> coreForTool = toolResultEventsFor(toolCallId, coreEvents);
```
oss-maintainer, line 2847: **[Warning]** Cost per acting iteration: for every result entry, `toolResultEventsFor` filters **both** accumulated lists, so this is `O(results x events)`; the lists are `CopyOnWriteArrayList`, so each `addToolResultEvent` copies the whole array — with fine-grained `ToolResultTextDeltaEvent` streams (potentially hundreds per tool call) the accumulation itself is `O(n^2)`. Suggest grouping by `toolCallId` in a single pass into `Map<String, List<AgentEvent>>` (a plain `ArrayList` per key is enough if the sink is serialized, otherwise `Collections.synchronizedList`) and reusing it for both core and transformed lists.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2809,6 +2829,107 @@ private Mono<Msg> acting(int iter) {
...
+                StringBuilder text = new StringBuilder();
+                Map<String, Object> metadata = original.getMetadata();
+                ToolResultState state = original.getState();
+                for (AgentEvent event : transformedForTool) {
+                    if (event instanceof ToolResultTextDeltaEvent textDelta) {
+                        if (textDelta.getDelta() != null) {
+                            text.append(textDelta.getDelta());
+                        }
+                        if (textDelta.getMetadata() != null) {
+                            metadata = textDelta.getMetadata();
```
oss-maintainer, line 2867: **[Info]** `metadata` is last-writer-wins across text delta -> data delta -> end event, and `state` likewise. That is probably intended, but it is a silent semantic: a middleware that stamps metadata on an early delta loses it if a later event carries its own map. Worth one javadoc line on `onActing` stating merge policy (replace vs. putAll).

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2744,11 +2755,20 @@ private Mono<Msg> acting(int iter) {
                                                 .apply(new ActingInput(toolCalls));
                                 return stream.doOnNext(
                                                 ev -> {
+                                                    addToolResultEvent(
+                                                            transformedToolResultEvents, ev);
+                                                    publishEvent(ev);
```
oss-maintainer, line 2760: **[Info]** Moving `publishEvent` off `actingStream` onto the post-middleware `stream` is the right change (emitters now see transformed events), but it means middleware that **swallows** events also suppresses publication entirely, and the ordering of `publishEvent` relative to `actingStopRequested` is now fixed by the outer `doOnNext`. Please confirm the AG-UI / event-bridge paths tolerate an event being withheld by a middleware, and note the contract in `MiddlewareBase.onActing` javadoc alongside the new sentence about result transformations.

oss-maintainer (commented):
### Summary
This closes a real gap: `onActing` middleware could rewrite the event stream but the memory/result path still saw the untransformed `resultHolder` content, and moving `publishEvent` to the post-middleware stream makes the two views consistent. The E2E test asserting the *next model request* carries the rewritten result is exactly the right way to prove it.

Two things to tighten before this is mergeable: the identity-based "was it transformed" check is fragile, and the accumulation/replay is quadratic on high-frequency delta events, which is the hot path for streamed tool output.

### Findings
- **[Warning]** `agentscope-core/.../ReActAgent.java:2914` — identity comparison triggers full reconstruction for any `.map()` middleware; reconstruction only preserves the three tool-result event types
- **[Warning]** `agentscope-core/.../ReActAgent.java:2847` — `O(results x events)` filtering plus `CopyOnWriteArrayList` appends on a hot delta stream
- **[Info]** `agentscope-core/.../ReActAgent.java:2867` — metadata/state merge policy is last-writer-wins, should be documented
- **[Info]** `agentscope-core/.../ReActAgent.java:2760` — middleware can now withhold events from `publishEvent`; confirm downstream consumers

### Suggestions
Replace the two parallel event lists with one grouping pass keyed by `toolCallId`, and gate reconstruction on a content comparison (or an explicit "this middleware rewrote results" marker) instead of instance identity.

### Compatibility
`MiddlewareBase.onActing` is a public SPI and third-party middleware exists in the extensions modules; a middleware that currently passes events through unchanged must keep producing identical memory content after this change. Please grep the extensions for `onActing` implementations and run their tests against the new loop.

Note: `mergeable_state=dirty` — needs a rebase onto `main`. The `OpenAITextEmbeddingEmbedTest` `any()` -> `any(String.class)` edit looks like an unrelated overload-resolution fix; fine to keep, but a one-line note in the PR description helps reviewers.

---

PR #2685, opened 2026-08-13
fix(issue#2684): support explicit no-cache via CACHE_CONTROL=false metadata

xzxiaoshan:
## Summary

补齐 `CACHE_CONTROL` 元数据的 `false`（显式不缓存）语义，使单条消息能显式排除在缓存之外。对应 issue https://github.com/agentscope-ai/agentscope-java/issues/2684

## Changes

- converter（OpenAI + DashScope）识别 `false`：`true → {"type":"ephemeral"}`、`false → {"type":"no_cache"}`。
- 自动缓存策略改用 `shouldAutoCache(cc) = cc == null || ephemeral.equals(cc)`，`no_cache` 与自定义值不被覆盖。
- 更新 `MessageMetadataKeys.CACHE_CONTROL` Javadoc。
- 新增/更新单测：system 消息显式 `false` 不被自动缓存、`false` 标记为 `no_cache`。

## Test

- `OpenAICacheControlTest`、`DashScopeCacheControlTest` 及两个 model 级 cache_control 用例已验证通过。


agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/formatter/OpenAIBaseFormatter.java
```diff
@@ -44,6 +44,9 @@ public abstract class OpenAIBaseFormatter
 
     private static final Map<String, String> EPHEMERAL_CACHE_CONTROL = Map.of("type", "ephemeral");
 
+    /** Explicit <code>no_cache</code> control sentinel marking a message excluded from caching. */
+    private static final Map<String, String> NO_CACHE_CONTROL = Map.of("type", "no_cache");
```
AgentScopeJavaBot: [major] **Protocol compliance risk:** The `no_cache` sentinel `{"type": "no_cache"}` is set directly on the `OpenAIMessage` DTO via `setCacheControl()` and will be serialized as `"cache_control": {"type": "no_cache"}` in the HTTP request body. The OpenAI API only documents `{"type": "ephemeral"}` as a valid `cache_control` type. Sending an unrecognized type may cause a 400 error or undefined behavior.

The same issue exists symmetrically in `DashScopeChatFormatter.java:47` and the two `applyCacheControlFromMetadata` methods (`DashScopeMessageConverter.java:284`, `OpenAIMessageConverter.java:545`).

**Suggested fix:** Strip the sentinel before the request is sent over the wire. For example, add a pre-serialization pass in the model layer (or override `buildRequest`) that sets `cacheControl = null` on any message carrying the `no_cache` sentinel:
```java
for (OpenAIMessage msg : request.getMessages()) {
    if (NO_CACHE_CONTROL.equals(msg.getCacheControl())) {
        msg.setCacheControl(null);
    }
}
```
Alternatively, use an internal boolean field (e.g. `@JsonIgnore boolean excludedFromCaching`) on the DTO to track exclusion without polluting the API payload.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/formatter/DashScopeChatFormatter.java
```diff
@@ -204,4 +207,27 @@ public void applyCacheControl(List<DashScopeMessage> messages) {
...
+     * <p>Returns {@code true} when the cache_control is unset ({@code null}) or already explicitly
+     * marked for caching ({@code {"type": "ephemeral"}}). A message explicitly marked as no-cache
+     * ({@code {"type": "no_cache"}}) or carrying any other custom cache_control is left untouched.
+     *
+     * @param cacheControl the current cache_control of the message, may be {@code null}
+     * @return {@code true} if the message should be auto-cached, {@code false} otherwise
+     */
+    static boolean shouldAutoCache(Map<String, String> cacheControl) {
+        return cacheControl == null || EPHEMERAL_CACHE_CONTROL.equals(cacheControl);
+    }
```
AgentScopeJavaBot, line 237: [nitpick] **Javadoc accuracy:** `shouldAutoCache` returns `true` when `existingCacheControl` equals `EPHEMERAL_CACHE_CONTROL`, meaning the auto-cache strategy will technically 'overwrite' an explicitly-marked `true` message with the same `ephemeral` value. While functionally harmless (idempotent), this contradicts the updated Javadoc in `MessageMetadataKeys.CACHE_CONTROL` which states: *"Explicitly marked messages (either true or false) take priority over the automatic strategy — they will not be overwritten."*

Consider either (a) narrowing `shouldAutoCache` to only return `true` for `null` (so explicit `true` is truly untouched), or (b) softening the Javadoc to say *"will not be altered"* and noting the idempotent behavior.
The same applies to `OpenAIBaseFormatter.java:227`.

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR adds support for `CACHE_CONTROL=false` metadata to explicitly exclude messages from prompt caching. The implementation introduces a `{"type": "no_cache"}` sentinel that is set on message DTOs and a `shouldAutoCache` guard so the automatic cache strategy skips explicitly-excluded messages. The code changes are clean, symmetric across OpenAI and DashScope providers, and well-tested. However, there is one significant concern: the `no_cache` sentinel is a non-standard `cache_control` type that will be serialized into the HTTP request body and sent to the upstream API, which only recognizes `{"type": "ephemeral"}`. This could cause API errors or undefined behavior.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/formatter/OpenAIBaseFormatter.java
AgentScopeJavaBot: [major] **Protocol compliance risk:** The `no_cache` sentinel `{"type": "no_cache"}` is set directly on the `OpenAIMessage` DTO via `setCacheControl()` and will be serialized as `"cache_control": {"type": "no_cache"}` in the HTTP request body. The OpenAI API only documents `{"type": "ephemeral"}` as a valid `cache_control` type. Sending an unrecognized type may cause a 400 error or undefined behavior.

The same issue exists symmetrically in `DashScopeChatFormatter.java:47` and the two `applyCacheControlFromMetadata` methods (`DashScopeMessageConverter.java:284`, `OpenAIMessageConverter.java:545`).

**Suggested fix:** Strip the sentinel before the request is sent over the wire. For example, add a pre-serialization pass in the model layer (or override `buildRequest`) that sets `cacheControl = null` on any message carrying the `no_cache` sentinel:
```java
for (OpenAIMessage msg : request.getMessages()) {
    if (NO_CACHE_CONTROL.equals(msg.getCacheControl())) {
        msg.setCacheControl(null);
    }
}
```
Alternatively, use an internal boolean field (e.g. `@JsonIgnore boolean excludedFromCaching`) on the DTO to track exclusion without polluting the API payload.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/formatter/DashScopeChatFormatter.java
AgentScopeJavaBot, line 237: [nitpick] **Javadoc accuracy:** `shouldAutoCache` returns `true` when `existingCacheControl` equals `EPHEMERAL_CACHE_CONTROL`, meaning the auto-cache strategy will technically 'overwrite' an explicitly-marked `true` message with the same `ephemeral` value. While functionally harmless (idempotent), this contradicts the updated Javadoc in `MessageMetadataKeys.CACHE_CONTROL` which states: *"Explicitly marked messages (either true or false) take priority over the automatic strategy — they will not be overwritten."*

Consider either (a) narrowing `shouldAutoCache` to only return `true` for `null` (so explicit `true` is truly untouched), or (b) softening the Javadoc to say *"will not be altered"* and noting the idempotent behavior.
The same applies to `OpenAIBaseFormatter.java:227`.

AgentScopeJavaBot (changes requested):
🤖 AI Review

This PR adds support for `CACHE_CONTROL=false` metadata to explicitly exclude messages from prompt caching. The implementation introduces a `{"type": "no_cache"}` sentinel that is set on message DTOs and a `shouldAutoCache` guard so the automatic cache strategy skips explicitly-excluded messages. The code changes are clean, symmetric across OpenAI and DashScope providers, and well-tested. However, there is one significant concern: the `no_cache` sentinel is a non-standard `cache_control` type that will be serialized into the HTTP request body and sent to the upstream API, which only recognizes `{"type": "ephemeral"}`. This could cause API errors or undefined behavior.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/formatter/DashScopeChatFormatter.java
AgentScopeJavaBot: ✅ Verified as addressed in `112743d`: shouldAutoCache and EPHEMERAL_CACHE_CONTROL are completely removed; applyCacheControl now uses getCacheControl()==null check so explicit intents are never overwritten, matching the Javadoc.

AgentScopeJavaBot: ✅ Verified as addressed in `112743d`: shouldAutoCache and EPHEMERAL_CACHE_CONTROL are completely removed; applyCacheControl now uses getCacheControl()==null check so explicit intents are never overwritten, matching the Javadoc.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/formatter/OpenAIBaseFormatter.java
AgentScopeJavaBot: ✅ Verified as addressed in `a2eeff9`: Sentinel changed from {"type":"no_cache"} to Collections.emptyMap() with @JsonInclude(NON_EMPTY) on cacheControl, so it is never serialized to the API payload.

AgentScopeJavaBot: ✅ Verified as addressed in `a2eeff9`: Same fix applied symmetrically across OpenAIBaseFormatter, DashScopeChatFormatter, and both converter applyCacheControlFromMetadata methods; empty-map sentinel is stripped by @JsonInclude(NON_EMPTY).

AgentScopeJavaBot (approved):

jujn (approved):

PR #2687, opened 2026-08-13
fix(harness): expose shared knowledge to isolated users

### LINKED ISSUE #2683: [Bug]:<agent-id>/knowledge文件夹下的文件读取不到
按照官网的说法（KNOWLEDGE.md 全文进 <domain_knowledge_context>；
同目录下其它文件（任意深度）只列路径清单进 prompt，让 agent 用 read_file / grep_files / glob_files 按需读取。）<agent-id>/knowledge文件夹下的KNOWLEDGE.md以及其他文件的路径清单都能够正常注入，但是读取文件内容时却找不到指定的文件。

Image

只有当我将文件放在<agent-id>/<user-id>/knowledge用户隔离区时文件才能够正常读取到。

Image

如果是正常的情况是否应该是：优先读取用户隔离区中的/knowledge，当用户隔离区中未找到相应的文件时就会读取agent工作区中的/knowledge？

q827505931:
## What changed

- add a local workspace overlay that resolves knowledge files in this order: isolated user/session workspace, shared agent workspace, then project
- preserve copy-on-write behavior when editing shared knowledge
- keep shell execution and project-writable routing intact
- add regression coverage for shared fallback, isolated overrides, copy-on-write, shell capability, and cross-user isolation

## Root cause

The local filesystem upper layer applies the user/session namespace to every relative path. Once isolation is enabled, `knowledge/foo.md` resolves only under `<workspace>/<namespace>/knowledge/foo.md`; the overlay then falls back directly to the project filesystem and never checks the agent-level `<workspace>/knowledge/foo.md` baseline that was already injected into the prompt.

## Validation

- `mvn -pl agentscope-harness -am test`
- core: 2,274 tests, 0 failures, 0 errors
- harness: 832 tests, 0 failures, 0 errors

Fixes #2683

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
@@ -1,0 +1,225 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent.filesystem;
+
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.harness.agent.filesystem.model.EditResult;
+import io.agentscope.harness.agent.filesystem.model.ExecuteResponse;
+import io.agentscope.harness.agent.filesystem.model.FileDownloadResponse;
+import io.agentscope.harness.agent.filesystem.model.FileInfo;
+import io.agentscope.harness.agent.filesystem.model.GlobResult;
+import io.agentscope.harness.agent.filesystem.model.GrepMatch;
+import io.agentscope.harness.agent.filesystem.model.GrepResult;
+import io.agentscope.harness.agent.filesystem.model.LsResult;
+import io.agentscope.harness.agent.filesystem.model.ReadResult;
+import io.agentscope.harness.agent.filesystem.model.WriteResult;
+import io.agentscope.harness.agent.filesystem.sandbox.AbstractSandboxFilesystem;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Local filesystem overlay that exposes agent-level knowledge as a read-only baseline beneath
+ * the isolated user/session layer.
+ *
+ * <p>Knowledge reads use {@code isolated workspace > agent workspace > project}. Other paths keep
+ * the standard {@link OverlayFilesystem} ordering ({@code isolated workspace > project}), so a
+ * caller cannot traverse another user's namespace through the agent workspace fallback. Writes
+ * always target the isolated upper layer and edits of shared knowledge use copy-on-write.
+ */
+public class LocalWorkspaceOverlay extends OverlayFilesystem implements AbstractSandboxFilesystem {
+
+    private static final String KNOWLEDGE = "knowledge";
+
+    private final AbstractSandboxFilesystem shellBackend;
+    private final AbstractFilesystem sharedWorkspace;
+
+    public LocalWorkspaceOverlay(
+            AbstractSandboxFilesystem upper,
+            AbstractFilesystem sharedWorkspace,
+            AbstractFilesystem project) {
+        super(upper, project);
+        this.shellBackend = upper;
+        this.sharedWorkspace = sharedWorkspace;
+    }
+
+    @Override
+    public String id() {
+        return shellBackend.id();
+    }
+
+    @Override
+    public ExecuteResponse execute(
+            RuntimeContext runtimeContext, String command, Integer timeoutSeconds) {
+        return shellBackend.execute(runtimeContext, command, timeoutSeconds);
+    }
+
+    @Override
+    public ReadResult read(RuntimeContext rc, String filePath, int offset, int limit) {
+        if (!isKnowledgePath(filePath) || upper().exists(rc, filePath)) {
+            return super.read(rc, filePath, offset, limit);
+        }
+        if (sharedWorkspace.exists(rc, filePath)) {
+            return sharedWorkspace.read(rc, filePath, offset, limit);
+        }
+        return lower().read(rc, filePath, offset, limit);
+    }
+
+    @Override
+    public LsResult ls(RuntimeContext rc, String path) {
+        if (!isKnowledgePath(path)) {
+            return super.ls(rc, path);
+        }
+        Map<String, FileInfo> merged = new LinkedHashMap<>();
+        mergeFileInfo(merged, lower().ls(rc, path));
+        mergeFileInfo(merged, sharedWorkspace.ls(rc, path));
+        mergeFileInfo(merged, upper().ls(rc, path));
+        return LsResult.success(new ArrayList<>(merged.values()));
+    }
+
+    @Override
+    public GlobResult glob(RuntimeContext rc, String pattern, String path) {
+        if (!isKnowledgePath(path)) {
+            return super.glob(rc, pattern, path);
+        }
+        Map<String, FileInfo> merged = new LinkedHashMap<>();
+        mergeFileInfo(merged, lower().glob(rc, pattern, path));
+        mergeFileInfo(merged, sharedWorkspace.glob(rc, pattern, path));
+        mergeFileInfo(merged, upper().glob(rc, pattern, path));
+        return GlobResult.success(new ArrayList<>(merged.values()));
+    }
+
+    @Override
+    public GrepResult grep(RuntimeContext rc, String pattern, String path, String glob) {
+        if (!isKnowledgePath(path)) {
+            return super.grep(rc, pattern, path, glob);
+        }
+        Map<String, GrepMatch> merged = new LinkedHashMap<>();
+        mergeGrep(merged, lower().grep(rc, pattern, path, glob));
+        mergeGrep(merged, sharedWorkspace.grep(rc, pattern, path, glob));
+        mergeGrep(merged, upper().grep(rc, pattern, path, glob));
+        return GrepResult.success(new ArrayList<>(merged.values()));
+    }
+
+    @Override
+    public EditResult edit(
+            RuntimeContext rc,
+            String filePath,
+            String oldString,
+            String newString,
+            boolean replaceAll) {
+        if (!isKnowledgePath(filePath) || upper().exists(rc, filePath)) {
+            return super.edit(rc, filePath, oldString, newString, replaceAll);
+        }
+        AbstractFilesystem source =
+                sharedWorkspace.exists(rc, filePath) ? sharedWorkspace : lower();
+        if (!source.exists(rc, filePath)) {
+            return EditResult.fail("File not found: " + filePath);
+        }
+        ReadResult read = source.read(rc, filePath, 0, Integer.MAX_VALUE);
+        if (!read.isSuccess()) {
+            return EditResult.fail("Cannot read shared file for copy-on-write: " + filePath);
+        }
+        WriteResult write = upper().write(rc, filePath, read.fileData().content());
+        if (!write.isSuccess()) {
+            return EditResult.fail("Cannot copy shared file for editing: " + filePath);
+        }
+        return upper().edit(rc, filePath, oldString, newString, replaceAll);
+    }
+
+    @Override
+    public List<FileDownloadResponse> downloadFiles(RuntimeContext rc, List<String> paths) {
+        List<FileDownloadResponse> results = new ArrayList<>();
+        for (String path : paths) {
+            if (!isKnowledgePath(path) || upper().exists(rc, path)) {
+                results.addAll(super.downloadFiles(rc, List.of(path)));
+            } else if (sharedWorkspace.exists(rc, path)) {
+                results.addAll(sharedWorkspace.downloadFiles(rc, List.of(path)));
+            } else {
+                results.addAll(lower().downloadFiles(rc, List.of(path)));
+            }
+        }
+        return results;
+    }
+
+    @Override
+    public WriteResult delete(RuntimeContext rc, String path) {
+        if (!isKnowledgePath(path) || upper().exists(rc, path)) {
+            return super.delete(rc, path);
+        }
+        if (sharedWorkspace.exists(rc, path)) {
+            return WriteResult.fail("Cannot delete shared file: " + path);
+        }
+        return super.delete(rc, path);
+    }
+
+    @Override
+    public WriteResult move(RuntimeContext rc, String fromPath, String toPath) {
+        if (!isKnowledgePath(fromPath) || upper().exists(rc, fromPath)) {
+            return super.move(rc, fromPath, toPath);
+        }
+        if (!sharedWorkspace.exists(rc, fromPath)) {
+            return super.move(rc, fromPath, toPath);
+        }
+        ReadResult read = sharedWorkspace.read(rc, fromPath, 0, Integer.MAX_VALUE);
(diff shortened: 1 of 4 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
```diff
@@ -0,0 +1,225 @@
...
+
+    private static boolean isKnowledgePath(String path) {
+        if (path == null || path.isBlank()) {
+            return false;
+        }
+        String normalized = path.replace('\\', '/').strip();
+        while (normalized.startsWith("/")) {
+            normalized = normalized.substring(1);
+        }
+        return normalized.equals(KNOWLEDGE) || normalized.startsWith(KNOWLEDGE + "/");
```
AgentScopeJavaBot, line 205: [major] **Path traversal in `isKnowledgePath` bypasses knowledge-only restriction on the shared workspace fallback.**

`isKnowledgePath` checks the raw path string against the `knowledge` prefix without normalizing `..` segments. A path like `knowledge/../MEMORY.md` passes the check (starts with `knowledge/`) but resolves to `MEMORY.md` at the workspace root when the underlying `LocalFilesystem` normalizes it.

In a multi-user deployment this means a user can read workspace-level files (e.g. `MEMORY.md`, `AGENTS.md`, other users' data) through the shared workspace fallback, bypassing the user/session namespace isolation that the upper layer provides.

The underlying `LocalFilesystem` prevents escaping the workspace root entirely (so this is not an arbitrary-file-read), but it does expose files within the workspace that are outside the `knowledge/` namespace.

**Fix:** Normalize the path before the prefix check:
```java
private static boolean isKnowledgePath(String path) {
    if (path == null || path.isBlank()) {
        return false;
    }
    String normalized = path.replace('\\', '/').strip();
    while (normalized.startsWith("/")) {
        normalized = normalized.substring(1);
    }
    normalized = Path.of(normalized).normalize().toString().replace('\\', '/');
    return normalized.equals(KNOWLEDGE) || normalized.startsWith(KNOWLEDGE + "/");
}
```
Consider adding a test case: `filesystem.read(alice, "knowledge/../MEMORY.md", ...)` should NOT fall through to the shared workspace.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
```diff
@@ -0,0 +1,225 @@
...
+        if (!read.isSuccess()) {
+            return EditResult.fail("Cannot read shared file for copy-on-write: " + filePath);
+        }
+        WriteResult write = upper().write(rc, filePath, read.fileData().content());
+        if (!write.isSuccess()) {
+            return EditResult.fail("Cannot copy shared file for editing: " + filePath);
+        }
+        return upper().edit(rc, filePath, oldString, newString, replaceAll);
+    }
+
```
AgentScopeJavaBot, line 144: [minor] **`move` on a shared knowledge file copies to upper without removing the source from the shared layer.**

`move()` delegates to `write()` + `delete()`. When the source is a shared knowledge file (exists in the shared layer but not in the isolated layer), `write()` copies it to the upper layer, then `delete()` on the `LocalFilesystem` tries to delete the file at `<namespace>/knowledge/foo.md` — which doesn't exist in the upper layer. The delete silently succeeds (no-op), so the shared file remains accessible for reads.

This means `move` behaves as `copy` for shared knowledge files. Document this limitation or add logic to create a deletion marker in the upper layer to mask the shared file after the move.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR introduces `LocalWorkspaceOverlay`, a new three-layer overlay filesystem that inserts a **shared agent workspace** between the isolated user/session upper layer and the project lower layer — but only for `knowledge/` paths. The read fallback order is `isolated workspace → shared agent workspace → project`. Writes always target the isolated upper layer, and edits of shared knowledge files use copy-on-write. `ProjectAwareOverlay` is updated to extend `LocalWorkspaceOverlay` and its constructor gains a `sharedWorkspace` parameter. The test suite covers the three-layer fallback, copy-on-write edit, delete protection, move semantics, namespace isolation, and non-knowledge path isolation.

The design is sound and the test coverage is thorough. Two findings below.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
AgentScopeJavaBot, line 205: [major] **Path traversal in `isKnowledgePath` bypasses knowledge-only restriction on the shared workspace fallback.**

`isKnowledgePath` checks the raw path string against the `knowledge` prefix without normalizing `..` segments. A path like `knowledge/../MEMORY.md` passes the check (starts with `knowledge/`) but resolves to `MEMORY.md` at the workspace root when the underlying `LocalFilesystem` normalizes it.

In a multi-user deployment this means a user can read workspace-level files (e.g. `MEMORY.md`, `AGENTS.md`, other users' data) through the shared workspace fallback, bypassing the user/session namespace isolation that the upper layer provides.

The underlying `LocalFilesystem` prevents escaping the workspace root entirely (so this is not an arbitrary-file-read), but it does expose files within the workspace that are outside the `knowledge/` namespace.

**Fix:** Normalize the path before the prefix check:
```java
private static boolean isKnowledgePath(String path) {
    if (path == null || path.isBlank()) {
        return false;
    }
    String normalized = path.replace('\\', '/').strip();
    while (normalized.startsWith("/")) {
        normalized = normalized.substring(1);
    }
    normalized = Path.of(normalized).normalize().toString().replace('\\', '/');
    return normalized.equals(KNOWLEDGE) || normalized.startsWith(KNOWLEDGE + "/");
}
```
Consider adding a test case: `filesystem.read(alice, "knowledge/../MEMORY.md", ...)` should NOT fall through to the shared workspace.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
AgentScopeJavaBot, line 144: [minor] **`move` on a shared knowledge file copies to upper without removing the source from the shared layer.**

`move()` delegates to `write()` + `delete()`. When the source is a shared knowledge file (exists in the shared layer but not in the isolated layer), `write()` copies it to the upper layer, then `delete()` on the `LocalFilesystem` tries to delete the file at `<namespace>/knowledge/foo.md` — which doesn't exist in the upper layer. The delete silently succeeds (no-op), so the shared file remains accessible for reads.

This means `move` behaves as `copy` for shared knowledge files. Document this limitation or add logic to create a deletion marker in the upper layer to mask the shared file after the move.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR introduces `LocalWorkspaceOverlay`, a new three-layer overlay filesystem that inserts a **shared agent workspace** between the isolated user/session upper layer and the project lower layer — but only for `knowledge/` paths. The read fallback order is `isolated workspace → shared agent workspace → project`. Writes always target the isolated upper layer, and edits of shared knowledge files use copy-on-write. `ProjectAwareOverlay` is updated to extend `LocalWorkspaceOverlay` and its constructor gains a `sharedWorkspace` parameter. The test suite covers the three-layer fallback, copy-on-write edit, delete protection, move semantics, namespace isolation, and non-knowledge path isolation.

The design is sound and the test coverage is thorough. Two findings below.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
q827505931: Addressed in d3d3b304 (duplicate review thread). The method documentation now explicitly states that moving shared knowledge is copy-on-write and preserves the read-only shared baseline; the existing move test verifies that contract.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
q827505931: Addressed in d3d3b304. isKnowledgePath now normalizes the path before checking the knowledge/ namespace. I also added sharedFallbackDoesNotAllowTraversalOutsideKnowledge, which verifies that knowledge/../MEMORY.md is rejected and cannot reach a workspace-root file. The underlying LocalFilesystem already rejects .. segments; the additional normalization keeps the shared-workspace classification correct as defense in depth.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
q827505931: Documented the intended behavior in d3d3b304. The shared workspace is a read-only baseline, so moving a shared knowledge file uses copy-on-write: it creates the destination in the isolated upper layer and deliberately preserves the shared source. sharedKnowledgeMoveCopiesToIsolatedLayerAndDeleteCannotRemoveBaseline verifies this behavior.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
q827505931: Addressed in d3d3b304 (duplicate review thread). isKnowledgePath now normalizes the path before the namespace check, and the new traversal regression test verifies that knowledge/../MEMORY.md cannot access the shared workspace root.

AgentScopeJavaBot: ✅ Verified as addressed in `d3d3b30`: isKnowledgePath (line 199) now normalizes via Path.of(normalized).normalize() before the knowledge/ prefix check, so knowledge/../MEMORY.md resolves to MEMORY.md and is correctly rejected.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
AgentScopeJavaBot: ✅ Verified as addressed in `d3d3b30`: move (line 175) no longer delegates to super.write+delete for shared knowledge sources; it explicitly reads from sharedWorkspace and writes to upper, eliminating the partial-side-effect bug. Javadoc documents the copy-on-write contract.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
AgentScopeJavaBot: ✅ Verified as addressed in `d3d3b30`: Duplicate of thread 1 — same isKnowledgePath normalization fix at line 199 covers this thread concern.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
AgentScopeJavaBot: ✅ Verified as addressed in `d3d3b30`: Duplicate of thread 2 — same move method fix (explicit read-from-shared + write-to-upper, no delete) at line 175 covers this thread concern.

AgentScopeJavaBot (approved):

oss-maintainer:

```bash
git fetch upstream
git checkout agent/fix-local-knowledge-fallback
git rebase upstream/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
```diff
@@ -0,0 +1,231 @@
...
+        if (!isKnowledgePath(path)) {
+            return super.ls(rc, path);
+        }
+        Map<String, FileInfo> merged = new LinkedHashMap<>();
+        mergeFileInfo(merged, lower().ls(rc, path));
+        mergeFileInfo(merged, sharedWorkspace.ls(rc, path));
+        mergeFileInfo(merged, upper().ls(rc, path));
+        return LsResult.success(new ArrayList<>(merged.values()));
+    }
+
```
oss-maintainer, line 94: **[Warning]** With no namespace (and with AGENT/GLOBAL scope), `upper` and `sharedWorkspace` are the same workspace but report different paths: the sandboxed shared filesystem emits `/knowledge/foo`, while the upper emits `knowledge/foo`. Since the merge keys are the raw paths, `ls` and `glob` duplicate every workspace knowledge entry; `grep` duplicates matches too. Normalize paths/keying or skip the shared merge when it aliases upper, and add a non-namespaced regression test.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
```diff
@@ -0,0 +1,231 @@
...
+        if (!read.isSuccess()) {
+            return EditResult.fail("Cannot read shared file for copy-on-write: " + filePath);
+        }
+        WriteResult write = upper().write(rc, filePath, read.fileData().content());
+        if (!write.isSuccess()) {
+            return EditResult.fail("Cannot copy shared file for editing: " + filePath);
+        }
+        return upper().edit(rc, filePath, oldString, newString, replaceAll);
+    }
+
```
oss-maintainer, line 144: **[Warning]** This copy-on-write sequence races another edit in the same namespace. After both callers see no upper file, one can create it and the other returns `Cannot copy...` solely because `write` sees it already exists. The previous overlay lets `upper.edit` serialize through its per-file lock. If a failed write leaves an upper file, continue with `upper.edit` (or retry), and add a concurrent shared-knowledge edit test.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/LocalWorkspaceOverlay.java
oss-maintainer, line 94: **[Warning]** `ls` now returns `LsResult.success(empty)` whenever all three layers fail. A nonexistent `knowledge` subdirectory, or a knowledge file passed to `ls`, therefore appears as an empty directory instead of preserving OverlayFilesystem's error result; FilesystemTool will report it misleadingly. Return an appropriate failed layer result when no layer succeeds, and cover missing and non-directory knowledge paths.

oss-maintainer (commented):
### Summary
Thanks for the thorough write-up and the regression suite — the three-layer knowledge resolution (`isolated workspace > agent workspace > project`) is the right shape, copy-on-write is kept on the write path, and non-knowledge paths still route through the plain overlay, so shell/project-writable behaviour is untouched. I verified the resolution order, the copy-on-write edit path, and the merge helpers against `LocalFilesystem`/`OverlayFilesystem` on `main`. Three things are worth fixing before this is mergeable, all in `LocalWorkspaceOverlay`; the rest is nits-free.

Headline issue: the merged discovery paths (`ls`/`glob`/`grep`) key on the display path returned by each layer, and the new `sharedWorkspace` layer is constructed as `new LocalFilesystem(workspace, true, 10, null)` — `virtualMode=true` makes it report `/knowledge/...` while the upper layer (default `LocalFsMode.ROOTED`, and always namespace-stripped) reports `knowledge/...`. The same physical entry therefore shows up twice in `ls`/`glob` output for every configuration, including non-namespaced ones that previously got a single entry from `OverlayFilesystem`. A maintainer should confirm, but this looks like a user-visible regression in the default setup rather than an isolated-only edge case.

Note on validation: I could not run `mvn -pl agentscope-harness -am test` in this environment, so the "832 harness tests green" result is taken from your PR description; the findings above come from static reading of the diff plus the `main` sources.

### Findings
- **[Warning]** `LocalWorkspaceOverlay.java:94` — duplicated `ls`/`glob`/`grep` entries caused by the aliased, differently-normalised `sharedWorkspace` layer (see inline comment).
- **[Warning]** `LocalWorkspaceOverlay.java:144` — copy-on-write pre-seeding races a concurrent edit and silently overwrites it with the shared baseline (see inline comment).
- **[Warning]** `LocalWorkspaceOverlay.java:94` — knowledge `ls`/`glob`/`grep` report success with an empty result when every layer failed, which `OverlayFilesystem` deliberately does not do (see inline comment).

---

PR #2697, opened 2026-08-13
issue#2696: 增强结构化输出降级路径可靠性，强制调用 generate_response 并注入模式提示。

xzxiaoshan:
完成 issue https://github.com/agentscope-ai/agentscope-java/issues/2696 内容，相关修改和方案，已在 issue 中陈述完整。

    ┌─────────────────────────────────────────────────────────────┐
    │  L0  原生结构化输出（结构化输出参数 + JSON Schema + 严格模式） │
    │      模型 API 层硬约束，最高可靠性                             │
    ├─────────────────────────────────────────────────────────────┤
    │  L1  合成工具注入（"结构化输出工具"，其参数即目标 Schema）       │
    │      把"输出 JSON"转化为"调用一个工具"                          │
    ├─────────────────────────────────────────────────────────────┤
    │  L2  模式上下文信号（进入/退出 <system-reminder> 持久提示）     │
    │      显式告知模型"当前必须产出结构化 JSON"，而非仅靠工具描述暗示   │
    ├─────────────────────────────────────────────────────────────┤
    │  L3  强制调用（工具强制调用参数，或提示词注入）                   │
    │      当模型未主动调用时，强制其调用结构化输出工具                  │
    ├─────────────────────────────────────────────────────────────┤
    │  L4  重试上限 + 优雅降级（3 次封顶，超限告警并正常结束）          │
    │      防止死循环，保证系统有界、可观测                            │
    └─────────────────────────────────────────────────────────────┘

**L0、L1（原来已实现）
L2、L3、L4（本次PR增加实现）**

---

## 概述
本 PR 旨在解决 agentscope 结构化输出的不可靠问题。
注入到工具列表的 `generate_response` 工具调用非强制，模型可能跳过它直接输出自由文本，导致循环结束可能不含结构化数据。本次实现增强了可靠性。


AgentScopeJavaBot (commented):
🤖 AI Review

This PR enhances the reliability of the structured-output fallback path in `ReActAgent`. When the model does not support native structured output (`response_format`), a synthetic `generate_response` tool is injected; however, the model could previously ignore it and emit free text, leaving the caller without structured data. The fix introduces three mechanisms: (1) persistent enter/exit system-reminder banners that prime the model about the active mode, (2) a force-retry strategy — `ToolChoice.Specific` for OpenAI, prompt-reminder for GLM/MiniMax — capped at 3 retries, and (3) context compression that strips SO-related messages after completion. The implementation is well-structured, thoroughly tested (11 new test cases), and solves the problem described in issue #2696. A few design concerns around persistent reminder accumulation and an optimistic capability default warrant attention.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR enhances the reliability of the structured-output fallback path in `ReActAgent`. When the model does not support native structured output (`response_format`), a synthetic `generate_response` tool is injected; however, the model could previously ignore it and emit free text, leaving the caller without structured data. The fix introduces three mechanisms: (1) persistent enter/exit system-reminder banners that prime the model about the active mode, (2) a force-retry strategy — `ToolChoice.Specific` for OpenAI, prompt-reminder for GLM/MiniMax — capped at 3 retries, and (3) context compression that strips SO-related messages after completion. The implementation is well-structured, thoroughly tested (11 new test cases), and solves the problem described in issue #2696. A few design concerns around persistent reminder accumulation and an optimistic capability default warrant attention.

xzxiaoshan:
> 🤖 AI Review
> 
> This PR enhances the reliability of the structured-output fallback path in `ReActAgent`. When the model does not support native structured output (`response_format`), a synthetic `generate_response` tool is injected; however, the model could previously ignore it and emit free text, leaving the caller without structured data. The fix introduces three mechanisms: (1) persistent enter/exit system-reminder banners that prime the model about the active mode, (2) a force-retry strategy — `ToolChoice.Specific` for OpenAI, prompt-reminder for GLM/MiniMax — capped at 3 retries, and (3) context compression that strips SO-related messages after completion. The implementation is well-structured, thoroughly tested (11 new test cases), and solves the problem described in issue #2696. A few design concerns around persistent reminder accumulation and an optimistic capability default warrant attention.
> 
> _(inline comments could not be attached — line numbers fell outside PR hunks. See archived report.)_

@AgentScopeJavaBot 
Thanks for the review. The summary mentions "A few design concerns around persistent reminder accumulation and an optimistic capability default warrant attention" — but the inline comments were not attached, and I can't access the "archived report" referenced at the end. Could you please paste the specific findings here? I'd especially like the details on those two concerns (persistent reminder accumulation, and the optimistic capability default) so I can address them concretely.

dargoner:
跟进一下这个 PR 的进度。

现状核对：6 commits / 9 files，目前只有机器人 COMMENTED，没有人工 approve；分支落后 main 20 个提交；关联的 issue #2696 仍为 open。

这个改动针对的是结构化输出降级路径的可靠性（模型不支持原生 response_format 时，强制调用 generate_response 并注入模式提示，避免模型自由发挥或陷入反复重试）。从下游使用看这类问题确实存在——我们 fork 里长期带着这份实现跑。

想请维护者/作者确认是否有已知阻塞，例如：

- 是否与 #2410 之后的主线结构存在冲突，需要 rebase 才能复审；
- 方向上是否更倾向原生 response_format 路线、而不接受提示注入这类兜底；
- 还需要补哪些测试或证据才能进入复审。

如果方向已经不被接受，也建议在 issue #2696 里明确说明，免得下游继续依赖这份实现。如果需要我这边补充下游的实际使用数据（什么模型、什么触发条件下会走到降级路径），我可以补到这条 PR 里。

xzxiaoshan:
@jujn 看评论反馈， @dargoner 他们似乎已经做过 review 并已经正式使用没有问题，是否可以 merge 。

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1384,9 +1413,11 @@ private Mono<Msg> doFallbackStructuredCall(List<Msg> msgs, Map<String, Object> j
                                     })
```
oss-maintainer, line 1433: On the structured-output give-up path (`flatMap` with `!soCompleted`), the persistent enter `<system-reminder>` injected at line 1922 is never rolled back and `soToolActive` stays `true`. Error and empty paths call `rollbackInjectedSoEnterReminder()`, but a model that never calls `generate_response` leaves the mode banner in durable state. Roll back the enter reminder in the `!soCompleted` branch before returning the free-text message.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2340,6 +2506,22 @@ private Mono<Msg> reasoning(int iter, boolean ignoreMaxIters) {
...
 
+            // Prompt-based force reminder (providers without ToolChoice.Specific, e.g. GLM /
+            // MiniMax): inject BEFORE firePreReasoning so it is captured in the input-messages
+            // snapshot — PreReasoningEvent defensively copies the context list, so a late
+            // injection here would be invisible to the model on this round.
+            if (soForceToolChoiceCount > 0
+                    && soTool != null
+                    && !model.supportsToolChoiceSpecific()) {
+                state.contextMutable().add(buildSoToolForceReminder());
+                log.debug(
```
oss-maintainer, line 2559: For providers without `ToolChoice.Specific` (GLM/MiniMax), a new force `<system-reminder>` is appended to context on every forced retry. `compressStructuredOutputContext` only removes them after a successful `generate_response` call, so up to three reminders stack before give-up, bloating the prompt and confusing the model. Avoid re-injecting when a force reminder is already present, or clean the previous one before each retry.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2495,6 +2700,30 @@ private Mono<Msg> runPostReasoningPipeline(Msg msg, int iter) {
...
+                                                "Structured output tool '{}' not called after {} "
+                                                        + "forced retries; finishing without "
+                                                        + "structured data (agent={}#{})",
+                                                STRUCTURED_OUTPUT_TOOL_NAME,
+                                                soForceToolChoiceCount,
+                                                getAgentId(),
+                                                getName());
+                                    }
                                     return Mono.justOrEmpty(eventMsg);
                                 }
```
oss-maintainer, line 2770: When the agent gives up after three forced retries, any injected prompt force reminders remain in context because cleanup only runs on the `soCompleted` success path. The stale reminders are persisted and influence later calls. Extend the give-up/error/empty cleanup to also strip messages marked with `MessageMetadataKeys.STRUCTURED_OUTPUT_REMINDER`.

oss-maintainer (commented):
### Summary

Adds a reliable structured-output fallback path: forces `generate_response` via `ToolChoice.Specific` (with a prompt-reminder strategy B for GLM/MiniMax), injects persistent enter/exit mode banners, tracks them with a new persisted `so_tool_active` flag, and rolls the injection back on error/empty turns. The overall design is sound and the rollback on `onErrorResume` / `switchIfEmpty` is a genuine improvement.

The remaining gaps are all in the cleanup paths that do *not* go through `soCompleted == true`: a give-up turn leaves the mode banner and the `soToolActive` flag in durable state, and strategy-B force reminders accumulate across retries. None of these are correctness regressions for the success path, so posting as a comment rather than blocking.

### Findings
7 inline comments — see the diff anchors. Headlines:
- **[Warning]** give-up path (`!soCompleted`) skips `rollbackInjectedSoEnterReminder()` → stale `so_tool_active=true` + banner persisted.
- **[Warning]** strategy-B force reminder is appended on *every* retry, and is only stripped on the success path.
- **[Warning]** `AgentState.so_tool_active` has no serializer test coverage; `MockModel.reset()` does not restore `supportsToolChoiceSpecific`.

### Suggestions
Route every non-completing exit (give-up included) through the same rollback helper, and make the force-reminder injection idempotent (remove/replace the previous `STRUCTURED_OUTPUT_REMINDER`-marked message instead of appending a new one).

CLA `signed` (`license/cla=success`), `mergeable=CLEAN`. CI status was not re-run here; the branch-protection gate is what blocks merging, not a conflict.

---

- **[WARNING]** `agentscope-core/src/test/java/io/agentscope/core/agent/test/MockModel.java:189` — `MockModel.reset()` clears call count, last messages, and error state but leaves `supportsToolChoiceSpecific` at whatever a previous test set. If a test reuses a mock instance after `setSupportsToolChoiceSpecific(false)`, later tests expecting the default `true` will fail. Reset `supportsToolChoiceSpecific` to `true` inside `reset()`. _(line outside diff)_

agentscope-core/src/main/java/io/agentscope/core/state/AgentState.java
```diff
@@ -111,6 +114,7 @@ static AgentState fromJson(
             @JsonProperty("reply_id") String replyId,
             @JsonProperty("cur_iter") Integer curIter,
             @JsonProperty("shutdown_interrupted") Boolean shutdownInterrupted,
+            @JsonProperty("so_tool_active") Boolean soToolActive,
```
oss-maintainer, line 117: The new serialized `so_tool_active` field is added to `AgentState`, but `AgentStateTest` is not updated to cover it. Add assertions for the default value in `defaultsAreApplied`, explicit builder round-trip in `explicitFieldsWin`, JSON property preservation in `jsonRoundTripPreservesScalars`, and omitted-field fallback in `jsonOmittedFieldsFallBackToDefaults`.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1491,6 +1522,86 @@ private ThinkingBlock collectLastThinking(AgentState agentState) {
         return last;
     }
 
+    /**
+     * Build the {@code <system-reminder>} banner injected when the agent enters structured-output
+     * tool mode. This is a persistent message (no metadata) — never cleaned — so consecutive
+     * structured-output calls see a stable context and prompt-cache prefixes stay intact.
+     */
+    private Msg buildSoToolEnterReminder() {
+        return SystemMessage.builder()
```
oss-maintainer, line 1570: Enter/exit `<system-reminder>` banners are intentionally persistent and never cleaned. Over repeated structured/normal call transitions the context accumulates one copy of each per cycle, which gradually consumes context window and may confuse the model. Consider deduplicating or capping retained transition reminders if long-running sessions show degraded behavior.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/OpenAIChatModel.java
```diff
@@ -423,6 +424,23 @@ public Builder nativeStructuredOutputWithTools(boolean nativeStructuredOutputWit
...
+         *
+         * <p>Defaults to {@code true}. Set to {@code false} for OpenAI-compatible providers
+         * (e.g. GLM, MiniMax) that only accept {@code tool_choice="auto"}; the agent will fall
+         * back to a prompt-reminder strategy when it needs to force the {@code generate_response}
+         * tool.
+         *
+         * @param supportsToolChoiceSpecific false if {@code ToolChoice.Specific} is unsupported
+         * @return this builder instance
+         */
+        public Builder supportsToolChoiceSpecific(boolean supportsToolChoiceSpecific) {
```
oss-maintainer, line 439: `OpenAIChatModel.Builder` exposes `supportsToolChoiceSpecific(boolean)`, but `OpenAIModelProviderSupport.applyAdvancedOptions` does not read it from `ModelCreationContext`. Users targeting other OpenAI-compatible providers that only support `tool_choice="auto"` cannot configure the flag without subclassing. Wire it as an advanced boolean option in the provider support helper.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1451,6 +1495,25 @@ private boolean isStructuredOutputRelated(Msg msg) {
...
+     * before injecting a fresh one so forced retries replace — rather than stack on — earlier
+     * reminders, including leftovers persisted by a previous call that gave up.
+     */
+    private void removeSoToolForceReminders(AgentState agentState) {
+        agentState.contextMutable().removeIf(this::isSoToolForceReminder);
+    }
+
+    /** Whether the message is a transient force reminder built by {@link
+     * #buildSoToolForceReminder}. */
+    private boolean isSoToolForceReminder(Msg msg) {
```
oss-maintainer, line 1516: **[Warning]** This cleanup is what makes "leftovers persisted by a previous call that gave up" recoverable, and it depends on the flag surviving a durable round-trip **as a `Boolean`**: `Boolean.TRUE.equals(metadata.get(STRUCTURED_OUTPUT_REMINDER))` is false for `"true"` or `1`, so a codec that widens metadata values to `String`/`Number` on deserialization silently disables the strip, and reminders stack again across restarts — the exact failure this commit is fixing, just moved off the in-memory path.

Note the difference from the pre-existing use: `compressStructuredOutputContext` / `isStructuredOutputRelated` only ever see messages from the *current* call, so they never exercised the round-trip. This is the first reader that consumes the flag after a reload.

The PR already adds `AgentStateTest` coverage for `so_tool_active` JSON round-trip; please add the sibling assertion for the reminder message itself — serialize a context containing one `buildSoToolForceReminder()` message through the same codec and assert `isSoToolForceReminder` still returns true for the decoded message (and that `CACHE_CONTROL=false` survives, since the same `Map.of` carries it).

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/OpenAIModelProviderSupport.java
```diff
@@ -77,5 +78,10 @@ public static void applyAdvancedOptions(
         if (nativeStructuredOutputWithTools != null) {
             builder.nativeStructuredOutputWithTools(nativeStructuredOutputWithTools);
         }
+        Boolean supportsToolChoiceSpecific =
+                booleanOption(context, OPTION_SUPPORTS_TOOL_CHOICE_SPECIFIC);
+        if (supportsToolChoiceSpecific != null) {
+            builder.supportsToolChoiceSpecific(supportsToolChoiceSpecific);
```
oss-maintainer, line 84: **[Warning]** `supportsToolChoiceSpecific` is now a user-settable advanced option, but it is the only one of the four that has no documentation: `docs/v2/en|zh/docs/building-blocks/model.md` document `contextWindowSize`, `nativeStructuredOutput` and `nativeStructuredOutputWithTools` (the `nativeStructuredOutputWithTools(false)` section around line 408 is the closest analogue — same "provider does not honour X, turn it off" story), while `git grep supportsToolChoiceSpecific pr/2697 -- docs/` returns nothing.

This one is also the option where a wrong value has the loudest consequence: setting it to `true` for a gateway that silently ignores `tool_choice="specific"` puts the agent back on strategy A and re-creates issue #2696's non-terminating structured-output path. Suggest one row in both languages next to `nativeStructuredOutputWithTools`, with the same "set to `false` for providers that only accept `tool_choice: auto` (GLM / MiniMax style)" wording.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1380,14 +1409,29 @@ private Mono<Msg> doFallbackStructuredCall(List<Msg> msgs, Map<String, Object> j
...
                                             }
                                             scope.state.contextMutable().add(out);
+                                        } else if (scope.soTool != null) {
+                                            // Gave up: the model never called generate_response.
+                                            // The turn's interaction stays in durable state, so
+                                            // the persistent enter banner and mode flag are kept,
+                                            // but the transient force reminders are stripped —
+                                            // they direct the model to call a tool that is not
+                                            // registered on later normal-mode turns.
+                                            removeSoToolForceReminders(scope.state);
```
oss-maintainer, line 1425: **[Info]** The comment justifies stripping the transient reminders because "they direct the model to call a tool that is not registered on later normal-mode turns" — but the retained enter banner carries the same instruction verbatim:

> `STRUCTURED OUTPUT mode is now active. Your final response MUST be produced by calling the generate_response tool ... Do NOT output free-form text as the final answer.`

The new test confirms that message is still in the follow-up call's input (`assertEquals(1, countTextOccurrences(..., "STRUCTURED OUTPUT mode is now active"))`), so after a give-up the model is told once that the mode is active while `soCompleted` is false and the synthetic tool is gone; the exit banner only closes it, and it is appended *before* the fresh user input in `doCallInner`.

I am not asking for a behaviour change — keeping the banner is a deliberate prompt-cache-prefix tradeoff documented on `buildSoToolEnterReminder()`, and the paired exit banner makes it self-healing for sequential calls. Two cheap hardenings, your call:
- state the residual-directive tradeoff in this comment so the next reader does not "finish the cleanup" and break the cache prefix;
- or extend the give-up assertion with a check that the enter/exit banners are in that order relative to the new user message, which is the property that actually makes the stale directive safe.

oss-maintainer (commented):
### Summary

Delta re-review after `93bcccb8` → `a0c49d70`. Two author commits since my last review, and both are improvements to the same failure mode — the structured-output fallback path no longer leaks state:

- `ad009ec9` makes strategy-B force reminders **replace** instead of stack (`removeSoToolForceReminders` before inject, and a strip on give-up), so at most one reminder is ever in context. The regression test captures the right observable: `reminderCounts == [0, 1, 1]` across three reasoning rounds rather than `[0, 1, 2]`, plus the post-give-up durable context (0 reminders, 1 enter banner, 1 exit banner). This closes the unbounded accumulation across calls that gave up, which is a real cost in a long-lived session, not just a tidiness issue.
- `1a83f8ac7` wires `supportsToolChoiceSpecific` as an advanced `ModelCreationContext` option, mirroring `nativeStructuredOutput`, so the strategy-A/strategy-B choice is configurable per provider instead of requiring a subclass — that is what makes the #2696 workaround usable from config. `ChatModelBase.supportsToolChoiceSpecific()` keeps defaulting to `true` when unset, so no behaviour change for existing configurations.

Two items I would like addressed before this is finished (both non-blocking, described inline):

- **`ReActAgent.java:1510`** — the strip is the first reader of `STRUCTURED_OUTPUT_REMINDER` that consumes the flag *after a durable round-trip*, and it matches on `Boolean` identity. If the metadata codec widens `true` to `String`/`Number` on deserialize, the leftover cleanup silently no-ops and reminders stack across restarts again. The PR already round-trips `so_tool_active` in `AgentStateTest`; one sibling assertion on a decoded reminder message would lock the contract.
- **`OpenAIModelProviderSupport.java:84`** — the option is undocumented. Both `model.md` docs already carry a section for the analogous `nativeStructuredOutputWithTools(false)` provider-workaround knob; this is the only one of the four advanced options with no row, and the one where a wrong value re-creates issue #2696's non-terminating path.

One design note is informational only (inline at `ReActAgent.java:1419`): stripping the transient reminder while deliberately keeping the persistent enter banner leaves a `MUST … call generate_response` directive in context after a give-up, which is self-healing via the paired exit banner but is exactly the rationale the new comment gives for stripping — worth stating the tradeoff explicitly so nobody "finishes" the cleanup and breaks the prompt-cache prefix.

No correctness defect found; `compressStructuredOutputContext` and the new predicate agree, and the give-up path keeps the interaction in durable state rather than discarding it, which is the right call for a stateless horizontal-scaling deployment.

Leaving as `COMMENT`, not `APPROVE`: CI is green at `a0c49d70` (`build` ubuntu/windows, `Check License`, `Check Module Sync`, `codecov/patch`), but `check_cla.py` reports `cla_status=unknown` — no `license/cla` context on this head — so I cannot confirm the CLA is signed and will not approve on that basis. Please get the CLA check to report on the head (or confirm the contributor is covered) and request a re-review; the remaining items above are not blockers for me.

---

oss-maintainer (commented):
### Summary
Delta re-review of the new author commit `a0bf93d2` since my last review on `a0c49d70`. The commit is polish, not behaviour change: it documents why the persistent *enter* banner is deliberately kept while the transient force reminders are stripped on give-up, adds a JSON round-trip test for the reminder metadata types, and documents the `supportsToolChoiceSpecific` opt-out for OpenAI-compatible gateways in both the EN and ZH model docs. This closes the loop on the state-leak concerns raised in the previous round — nothing here blocks the PR.

### Findings
- **[Info]** `agentscope-core/src/test/java/io/agentscope/core/state/AgentStateTest.java:185` — good test: it pins the exact assumption the give-up cleanup depends on (`Boolean.TRUE.equals` on a metadata value that may have come back from a durable store) rather than just asserting the round-trip is lossless. One gap: `STRUCTURED_OUTPUT_REMINDER_TYPE` is asserted to survive as the `String` `"PROMPT"`; if the writer side ever persists an enum name of a different shape, this assertion tracks the implementation instead of the contract. Consider asserting on the enum's name constant rather than a literal, or dropping that line since the comment already says the marker is informational and never read for cleanup.
- **[Info]** `agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java:1418` — the comment now explains the enter/exit banner asymmetry well. Since this is precisely the kind of invariant a later "simplify the cleanup" change would break, it would be worth naming the test that protects it (the give-up/strip test) in the comment, so the two stay in sync.
- **[Info]** `docs/v2/en/docs/building-blocks/model.md:421` (and the mirrored ZH section) — the new "Providers without specific `tool_choice` support" section is the right documentation to add, and the failure mode is stated concretely (retry a constraint the gateway never applies, loop, lose structured data). Minor: the ZH and EN blocks are not literal mirrors — EN says the option "is also available as a provider-config advanced option", ZH phrases it as an additional sentence; worth keeping the wording aligned so neither drifts.

### Verdict
No blocking findings. The remaining item on my side from earlier rounds was the give-up-path state cleanup, and the added round-trip test plus this comment make the intent explicit. Rebase onto `main` when convenient so CI can re-run against the current tree.

---

agentscope-core/src/test/java/io/agentscope/core/state/AgentStateTest.java
```diff
@@ -163,6 +178,38 @@ void jsonOmittedFieldsFallBackToDefaults() throws Exception {
         assertEquals("only-id", decoded.getSessionId());
         assertEquals("", decoded.getSummary());
         assertEquals(0, decoded.getCurIter());
+        assertFalse(decoded.isSoToolActive());
         assertNotNull(decoded.getReplyId());
     }
+
+    @Test
```
oss-maintainer, line 185: **[Info]** Good test — it pins the exact assumption the give-up cleanup depends on (`Boolean.TRUE.equals(...)` over metadata that may have come back from a durable store) instead of only asserting that the round trip is lossless.

One nit: the `STRUCTURED_OUTPUT_REMINDER_TYPE` assertion expects the literal `"PROMPT"`, so it tracks the current writer shape rather than the contract. Since the comment above already says the type marker is informational and never read for cleanup, either assert against the enum constant's name or drop that line — otherwise a future change to how the marker is serialised breaks a test for behaviour nobody depends on.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1380,14 +1409,35 @@ private Mono<Msg> doFallbackStructuredCall(List<Msg> msgs, Map<String, Object> j
                                                                 aggregatedThinking);
                                             }
                                             scope.state.contextMutable().add(out);
+                                        } else if (scope.soTool != null) {
+                                            // Gave up: the model never called generate_response.
+                                            // The turn's interaction stays in durable state, so
+                                            // the persistent enter banner and mode flag are kept,
+                                            // but the transient force reminders are stripped —
+                                            // they direct the model to call a tool that is not
+                                            // registered on later normal-mode turns. The enter
```
oss-maintainer, line 1418: **[Info]** This now explains the enter/exit banner asymmetry clearly, which is the part a future reader would otherwise "simplify" into a bug.

Since it is an invariant rather than a preference, it would help to name the test that protects it here (the give-up / strip test in `ReActAgentStructuredOutputTest`) so the code comment and the test stay linked when either moves.

docs/v2/en/docs/building-blocks/model.md
```diff
@@ -418,6 +418,21 @@ OpenAIChatModel model = OpenAIChatModel.builder()
 
 `DashScopeChatModel` supports this option as well. For native OpenAI models (GPT-4o, etc.) the default behavior handles both correctly — no configuration needed.
 
+#### Providers without specific `tool_choice` support
```
oss-maintainer, line 421: **[Info]** Right documentation to add, and the failure mode is stated concretely (retry a constraint the gateway never applies, loop, give up without structured data).

Minor: the EN and ZH blocks are not literal mirrors — EN closes with "It is also available as a provider-config advanced option", ZH rephrases it as a trailing clause. Worth keeping the two aligned so they do not drift on the next edit.

oss-maintainer (commented):
### Summary
Delta re-review of the new author commit `a0bf93d2` since my last review on `a0c49d70`. The commit is polish, not behaviour change: it documents why the persistent enter banner is deliberately kept while the transient force reminders are stripped on give-up, adds a JSON round-trip test for the reminder metadata types, and documents the `supportsToolChoiceSpecific` opt-out for OpenAI-compatible gateways in both the EN and ZH model docs. Nothing here blocks the PR; the findings above are all non-blocking.

### Verdict
No blocking findings. The item I raised in earlier rounds — state left behind when a structured-output turn gives up — is now covered by both the strip logic and a round-trip test that pins the metadata assumption the cleanup relies on. Rebasing onto `main` when convenient will let CI re-run against the current tree.

---

oss-maintainer:

```bash
git fetch origin
git checkout fix/structured-output-fallback
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1451,6 +1501,25 @@ private boolean isStructuredOutputRelated(Msg msg) {
...
     }
 
+    /**
+     * Remove previously injected structured-output force reminders (messages marked with {@link
+     * MessageMetadataKeys#STRUCTURED_OUTPUT_REMINDER}) from the conversation context. Called
+     * before injecting a fresh one so forced retries replace — rather than stack on — earlier
+     * reminders, including leftovers persisted by a previous call that gave up.
+     */
+    private void removeSoToolForceReminders(AgentState agentState) {
+        agentState.contextMutable().removeIf(this::isSoToolForceReminder);
```
oss-maintainer, line 1511: **[Warning]** `removeSoToolForceReminders` is only invoked on the paths that inject a *fresh* reminder (and on give-up). If a forced retry itself fails or returns an empty completion, the transient MUST-call directive is still persisted with the rolled-back turn, while the L2 mode banner handling in the same path only strips the banner. A subsequent non-structured call can therefore start from a context that still contains a stale force directive. Please strip transient `STRUCTURED_OUTPUT_REMINDER` messages on every failure/empty persistence path and add a regression test for the failed-retry case.

agentscope-core/src/main/java/io/agentscope/core/model/ChatModelBase.java
```diff
@@ -63,6 +65,15 @@ protected void setNativeStructuredOutputWithTools(boolean nativeStructuredOutput
         this.nativeStructuredOutputWithTools = nativeStructuredOutputWithTools;
     }
 
+    @Override
+    public boolean supportsToolChoiceSpecific() {
```
oss-maintainer, line 69: **[Warning]** `supportsToolChoiceSpecific()` defaults to *true* when a provider never sets it, so providers that silently downgrade `ToolChoice.Specific` (the OpenAI-compat Kimi path drops it to `auto` when thinking is enabled, which is mandatory for those models) opt into L3 specific-tool forcing by default. ReActAgent then burns all three forced retries without ever getting structured output instead of falling back to the prompt-injection route. Consider defaulting to `false` for compat providers (or setting it explicitly in `KimiModelProvider`) and covering the capability flag per provider.

oss-maintainer (commented):
### Summary
Thanks for the detailed design write-up — the L2 (persistent enter/exit mode banners) / L3 (per-request specific-tool or prompt forcing) / L4 (three-retry ceiling with graceful give-up) layering is a sensible way to make the synthetic `generate_response` path reliable, and I verified the pieces that matter most: `supportsToolChoiceSpecific()` is added to `Model` as a **default** method so existing SPI implementors stay source/binary compatible, the give-up path is bounded (four model calls, then a normal return), `so_tool_active` is nullable on deserialization so previously persisted `AgentState` payloads still read, and the forcing state lives on `CallExecution` rather than mutating a shared model instance. Two cleanup/capability-default issues are worth fixing before merge; everything else looked fine to me.

Headline issue: the transient `STRUCTURED_OUTPUT_REMINDER` directive survives the failure/empty persistence paths even though the give-up path deliberately strips it, so a later normal call can start from durable context containing a "MUST call generate_response" directive while the tool is not registered — the exact class of confusion L3 was meant to remove.

Note on validation: I could not run `mvn -pl agentscope-core -am test` in this review environment (the declared JDK 17 / Maven paths are unavailable), so the "core 2,274 tests green" figure is taken from your description; the findings above come from reading the diff against the `main` sources.

### Findings
- **[Warning]** `ReActAgent.java:1511` — force reminders are stripped on give-up but not on the failed/empty retry paths (see inline comment).
- **[Warning]** `ChatModelBase.java:69` — `supportsToolChoiceSpecific()` defaults to true for providers that never set it, which misroutes Kimi-style providers that silently downgrade `ToolChoice.Specific` (see inline comment).

---

PR #2699, opened 2026-08-13
fix(harness): respect disabled memory hooks during compaction

### LINKED ISSUE #2694: [Bug]:CompactionMiddleware 自建 MemoryFlushManager，绕过 disableMemoryHooks()——「记忆已关闭」时仍会写记忆文件并额外调用 LLM
> Labels 建议：`bug` / `harness` / `memory`

## 环境

| | |
|---|---|
| agentscope-harness | 2.0.2（2.0.0 起即如此） |
| 场景 | 调用方显式 `disableMemoryHooks()` + `disableMemoryTools()`，即「完全不使用长期记忆」 |

## 问题

调用 `HarnessAgent.Builder.disableMemoryHooks()` 后，`MemoryFlushMiddleware` 与 `MemoryMaintenanceMiddleware` 都不会注册，看起来记忆功能已彻底关闭。

但 `CompactionMiddleware` **自己 new 了一个 `MemoryFlushManager`**，完全不看这个开关：

```java
// CompactionMiddleware.java:102-104
MemoryFlushManager flushManager =
        new MemoryFlushManager(workspaceManager, model);
ConversationCompactor compactor =
        new ConversationCompactor(model, flushManager);
```

而 `CompactionConfig` 的两个相关默认值**都是 true**：

```java
// CompactionConfig.java:273/282/283
private int triggerMessages = 50;
private boolean flushBeforeCompact = true;
private boolean offloadBeforeCompact = true;
```

于是在「记忆已关闭」的前提下，只要会话长到 `triggerMessages`（默认 50 条），压缩前仍会：

1. **执行一次记忆抽取** —— 额外一次 LLM 调用，并把结果写进 `memory/<date>.md`
2. **把整份 messages offload 进 session JSONL**

## 实际后果（我们踩到的）

我们的服务在很长一段时间里都是「记忆关闭」状态（`disableMemoryHooks()` + `disableMemoryTools()`），workspace 挂在 NAS 上。后来准备启用长期记忆、检查 NAS 时，发现上面**已经躺着一批记忆日更文件和会话 JSONL** —— 全部是这条路径在「记忆已关闭」期间写出来的。

也就是说：

- **成本**：每次触发压缩都多一次 LLM 调用，而调用方认为自己关闭了这项功能
- **数据**：产生了从未被审阅、也不预期存在的记忆文件；一旦后续启用记忆，这些旧文件会被 `memory_search` 之类的入口读到
- **认知**：`disableMemoryHooks()` 的语义被违反，且没有任何日志提示

## 复现

1. `HarnessAgent.builder().disableMemoryHooks().disableMemoryTools()`，不设置 `compaction(...)`
2. 跑一个超过 50 条消息的会话，触发压缩
3. 观察 workspace：`memory/<date>.md` 被创建；同时可观察到一次额外的 LLM 调用

## 规避方式（供其他遇到的人参考）

必须**无条件显式**设置 compaction 配置，不能依赖默认值：

```java
builder.compaction(CompactionConfig.builder()
        .flushBeforeCompact(false)
        .offloadBeforeCompact(false)
        .build());
```

## 建议

任一即可：

- `CompactionMiddleware` 中的 flush / offload 受 `disableMemoryHooks` 约束（关闭时跳过，等价于把两个 flag 视为 false）
- 或者把 `MemoryFlushManager` 由外部注入，让它与其他记忆组件共享同一套开关
- 最低限度：在 `disableMemoryHooks()` 与 `CompactionConfig` 的 javadoc 里写明「compaction 内部的 flush/offload **不受** `disableMemoryHooks` 管辖，需单独关闭」

个人认为第一种最符合直觉：一个名为 "disable memory hooks" 的开关，不应该留下一条仍在写记忆文件的路径。

RICARDO51167:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #2694.

`disableMemoryHooks()` already disabled per-call memory flush and background consolidation, but normal compaction and the emergency context-overflow path could still flush long-term memory.

This PR:

- derives an immutable effective `CompactionConfig` that forces `flushBeforeCompact=false` when memory hooks are disabled;
- applies the same rule to emergency compaction and retry;
- keeps compaction summarization and session JSONL offload independent;
- preserves the behavior of directly constructed `CompactionMiddleware` instances;
- adds configuration and Harness integration coverage for normal compaction, emergency compaction, offload, and compatibility cases;
- updates the v2 English and Chinese memory, compaction, and workspace documentation.

Verification:

- `mvn -pl agentscope-harness -am -Dtest=HarnessAgentCompactionMemoryHooksTest,CompactionConfigTest,CompactionMiddlewareTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn spotless:apply spotless:check`
- `mvn test`
- `mvn verify`

## Checklist

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
@@ -180,6 +180,7 @@ public class HarnessAgent implements Agent, AutoCloseable {
     private final SkillCurator skillCurator;
     private final SkillAuditLog skillAuditLog;
     private final MemoryConfig memoryConfig;
+    private final boolean memoryHooksDisabled;
 
     /** The subagent middleware (either SubagentsMiddleware or DynamicSubagentsMiddleware). */
     private final Object subagentMiddleware;
@@ -211,6 +212,7 @@ public class HarnessAgent implements Agent, AutoCloseable {
             SkillCurator skillCurator,
             SkillAuditLog skillAuditLog,
             MemoryConfig memoryConfig,
+            boolean memoryHooksDisabled,
             Object subagentMiddleware,
             DistributedStore distributedStore,
             WorkspacePathNormalizer pathNormalizer) {
@@ -229,6 +231,7 @@ public class HarnessAgent implements Agent, AutoCloseable {
         this.skillCurator = skillCurator;
         this.skillAuditLog = skillAuditLog;
         this.memoryConfig = memoryConfig != null ? memoryConfig : MemoryConfig.defaults();
+        this.memoryHooksDisabled = memoryHooksDisabled;
         this.subagentMiddleware = subagentMiddleware;
         this.distributedStore = distributedStore;
         this.pathNormalizer = pathNormalizer;
@@ -1019,7 +1022,11 @@ public class HarnessAgent implements Agent, AutoCloseable {
                         ? effective.getSessionId()
                         : "default";
 
-        CompactionConfig forceConfig = CompactionConfig.builder().triggerMessages(1).build();
+        CompactionConfig forceConfig =
+                CompactionConfig.builder()
+                        .triggerMessages(1)
+                        .flushBeforeCompact(!memoryHooksDisabled)
+                        .build();
         String effectiveFlushPrompt =
                 memoryConfig.flushPrompt() != null
                         ? memoryConfig.flushPrompt()
@@ -2093,8 +2100,10 @@ public class HarnessAgent implements Agent, AutoCloseable {
         }
 
         /**
-         * Disables memory flush + background consolidation, and removes the "automatically
-         * extracted" Persistence line from the workspace system prompt. Combined with {@link
+         * Disables memory flush (including normal and emergency compaction flush) + background
+         * consolidation, and removes the "automatically extracted" Persistence line from the
+         * workspace system prompt. Compaction summarization and session JSONL offload remain
+         * independently controlled by {@link #compaction(CompactionConfig)}. Combined with {@link
          * #disableMemoryTools()}, also skips {@code MEMORY.md} injection into
          * {@code <memory_context>}.
          */
@@ -2454,11 +2463,18 @@ public class HarnessAgent implements Agent, AutoCloseable {
             }
             CompactionMiddleware compactionHook = null;
             if (!disableCompaction && compactionConfig != null) {
+                CompactionConfig effectiveCompactionConfig =
+                        disableMemoryHooks
+                                ? compactionConfig.withFlushBeforeCompact(false)
+                                : compactionConfig;
                 Model compactionModel =
-                        compactionConfig.getModel() != null ? compactionConfig.getModel() : model;
+                        effectiveCompactionConfig.getModel() != null
+                                ? effectiveCompactionConfig.getModel()
+                                : model;
                 if (compactionModel != null) {
                     compactionHook =
-                            new CompactionMiddleware(wsManager, compactionModel, compactionConfig);
+                            new CompactionMiddleware(
+                                    wsManager, compactionModel, effectiveCompactionConfig);
                     inner.middleware(compactionHook);
                 }
             }
@@ -2814,6 +2830,7 @@ public class HarnessAgent implements Agent, AutoCloseable {
                     pendingSkillCurator,
                     pendingSkillAuditLog,
                     memoryConfig,
+                    disableMemoryHooks,
                     capturedSubagentMw,
                     distributedStore,
                     pathNormalizer);
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/CompactionConfig.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/CompactionConfig.java
@@ -208,12 +208,18 @@ public class CompactionConfig {
         return summaryPrompt;
     }
 
-    /** Whether to flush long-term memories from the prefix before compaction. */
+    /**
+     * Whether to flush long-term memories from the prefix before compaction. When used through
+     * {@code HarnessAgent}, {@code disableMemoryHooks()} overrides this setting to {@code false}.
+     */
     public boolean isFlushBeforeCompact() {
         return flushBeforeCompact;
     }
 
-    /** Whether to offload raw messages to the session JSONL before compaction. */
+    /**
+     * Whether to offload raw messages to the session JSONL before compaction. This setting is
+     * independent of long-term memory hooks.
+     */
     public boolean isOffloadBeforeCompact() {
         return offloadBeforeCompact;
     }
@@ -246,12 +252,31 @@ public class CompactionConfig {
      * context window. Used by {@code CompactionMiddleware} to resolve dynamic defaults.
      */
     public CompactionConfig withEffective(int effectiveTriggerTokens, int effectiveKeepTokens) {
+        Builder b = copyBuilder();
+        b.triggerTokens = effectiveTriggerTokens;
+        b.keepTokens = effectiveKeepTokens;
+        return new CompactionConfig(b);
+    }
+
+    /**
+     * Creates a copy with the pre-compaction long-term memory flush enabled or disabled.
+     *
+     * @param flushBeforeCompact whether to flush long-term memories before compaction
+     * @return a new configuration with the requested flush setting
+     */
+    public CompactionConfig withFlushBeforeCompact(boolean flushBeforeCompact) {
+        Builder b = copyBuilder();
+        b.flushBeforeCompact = flushBeforeCompact;
+        return new CompactionConfig(b);
+    }
+
+    private Builder copyBuilder() {
         Builder b = new Builder();
         b.triggerMessages = this.triggerMessages;
-        b.triggerTokens = effectiveTriggerTokens;
+        b.triggerTokens = this.triggerTokens;
         b.reserved = this.reserved;
         b.keepMessages = this.keepMessages;
-        b.keepTokens = effectiveKeepTokens;
+        b.keepTokens = this.keepTokens;
         b.keepTokensMin = this.keepTokensMin;
         b.keepTokensMax = this.keepTokensMax;
         b.keepTokensRatio = this.keepTokensRatio;
@@ -261,7 +286,7 @@ public class CompactionConfig {
         b.truncateArgsConfig = this.truncateArgsConfig;
         b.pruneConfig = this.pruneConfig;
         b.model = this.model;
-        return new CompactionConfig(b);
+        return b;
     }
 
     public static Builder builder() {
@@ -346,13 +371,20 @@ public class CompactionConfig {
             return this;
         }
 
-        /** Whether to flush long-term memories before compaction (default true). */
+        /**
+         * Whether to flush long-term memories before compaction (default true). When used through
+         * {@code HarnessAgent}, {@code disableMemoryHooks()} overrides this setting to {@code
+         * false}.
+         */
         public Builder flushBeforeCompact(boolean flushBeforeCompact) {
             this.flushBeforeCompact = flushBeforeCompact;
             return this;
         }
 
-        /** Whether to offload raw messages to session JSONL before compaction (default true). */
+        /**
+         * Whether to offload raw messages to session JSONL before compaction (default true).
+         * Independent of long-term memory hooks.
+         */
         public Builder offloadBeforeCompact(boolean offloadBeforeCompact) {
             this.offloadBeforeCompact = offloadBeforeCompact;
             return this;
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentCompactionMemoryHooksTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentCompactionMemoryHooksTest.java
@@ -1,0 +1,241 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.core.message.Msg;
+import io.agentscope.core.message.MsgRole;
+import io.agentscope.core.message.TextBlock;
+import io.agentscope.core.model.ChatModelBase;
+import io.agentscope.core.model.ChatResponse;
+import io.agentscope.core.model.GenerateOptions;
+import io.agentscope.core.model.ToolSchema;
+import io.agentscope.core.state.InMemoryAgentStateStore;
+import io.agentscope.harness.agent.memory.MemoryConfig;
+import io.agentscope.harness.agent.memory.compaction.CompactionConfig;
+import io.agentscope.harness.agent.workspace.WorkspaceConstants;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.LocalDate;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import reactor.core.publisher.Flux;
+
+class HarnessAgentCompactionMemoryHooksTest {
+
+    @TempDir Path workspace;
+
+    @Test
+    void disableMemoryHooksSkipsCompactionFlushAndKeepsOffload() throws Exception {
+        RoutingModel model = new RoutingModel();
+        CompactionConfig config = normalCompactionConfig(true, true);
+
+        try (HarnessAgent agent = buildAgent(model, config, true)) {
+            RuntimeContext context = context("normal-session");
+            assertNotNull(agent.call(userMessage("first request"), context).block());
+            assertNotNull(agent.call(userMessage("second request"), context).block());
+        }
+
+        assertEquals(0, model.flushCalls.get());
+        assertEquals(1, model.summaryCalls.get());
+        assertFalse(hasDailyMemoryFile());
+        assertTrue(hasFileNamed("normal-session" + WorkspaceConstants.SESSION_CONTEXT_EXT));
+    }
+
+    @Test
+    void enabledMemoryHooksPreserveConfiguredCompactionFlush() throws Exception {
+        RoutingModel model = new RoutingModel();
+        CompactionConfig config = normalCompactionConfig(true, false);
+
+        try (HarnessAgent agent = buildAgent(model, config, false)) {
+            RuntimeContext context = context("enabled-session");
+            assertNotNull(agent.call(userMessage("first request"), context).block());
+            assertNotNull(agent.call(userMessage("second request"), context).block());
+        }
+
+        assertEquals(1, model.flushCalls.get());
+        assertEquals(1, model.summaryCalls.get());
+        assertTrue(hasDailyMemoryFile());
+    }
+
+    @Test
+    void explicitDisabledCompactionFlushRemainsDisabled() throws Exception {
+        RoutingModel model = new RoutingModel();
+        CompactionConfig config = normalCompactionConfig(false, false);
+
+        try (HarnessAgent agent = buildAgent(model, config, false)) {
+            RuntimeContext context = context("explicit-disabled-session");
+            assertNotNull(agent.call(userMessage("first request"), context).block());
+            assertNotNull(agent.call(userMessage("second request"), context).block());
+        }
+
+        assertEquals(0, model.flushCalls.get());
+        assertEquals(1, model.summaryCalls.get());
+        assertFalse(hasDailyMemoryFile());
+    }
+
+    @Test
+    void disableMemoryHooksSkipsEmergencyCompactionFlush() throws Exception {
+        RoutingModel model = new RoutingModel();
+        CompactionConfig config =
+                CompactionConfig.builder()
+                        .triggerMessages(0)
+                        .triggerTokens(Integer.MAX_VALUE)
+                        .keepTokens(0)
+                        .flushBeforeCompact(true)
+                        .offloadBeforeCompact(false)
+                        .prune(null)
+                        .build();
+
+        try (HarnessAgent agent = buildAgent(model, config, true)) {
+            RuntimeContext context = context("overflow-session");
+            for (int i = 0; i < 11; i++) {
+                assertNotNull(agent.call(userMessage("request " + i), context).block());
+            }
+
+            model.overflowNextReasoning.set(true);
+            assertNotNull(agent.call(userMessage("trigger overflow"), context).block());
+        }
+
+        assertEquals(0, model.flushCalls.get());
+        assertEquals(1, model.summaryCalls.get());
+        assertFalse(hasDailyMemoryFile());
+    }
+
+    private HarnessAgent buildAgent(
+            RoutingModel model, CompactionConfig config, boolean disableMemoryHooks) {
+        HarnessAgent.Builder builder =
+                HarnessAgent.builder()
+                        .name("compaction-memory-hooks-test")
+                        .model(model)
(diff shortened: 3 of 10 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1019,7 +1022,11 @@ private Mono<Msg> forceCompactAndRetry(List<Msg> msgs, RuntimeContext effective)
                         ? effective.getSessionId()
                         : "default";
 
-        CompactionConfig forceConfig = CompactionConfig.builder().triggerMessages(1).build();
+        CompactionConfig forceConfig =
+                CompactionConfig.builder()
+                        .triggerMessages(1)
+                        .flushBeforeCompact(!memoryHooksDisabled)
```
AgentScopeJavaBot, line 1028: [recommended] **Pre-existing asymmetry worth noting:** The emergency path (`forceCompactAndRetry`) builds a fresh `CompactionConfig` from defaults (`triggerMessages(1)` + `flushBeforeCompact(!memoryHooksDisabled)`), discarding the user's other compaction settings (keepTokens, pruneConfig, summaryPrompt, offloadBeforeCompact, etc.). This is pre-existing behavior — the PR correctly adds only the `flushBeforeCompact` fix — but it means the emergency path doesn't fully respect the user's compaction config the way the normal path now does via `withFlushBeforeCompact(false)`.

The root cause is that `compactionConfig` is only a Builder field, not stored on the `HarnessAgent` instance. A future improvement could store the effective `CompactionConfig` (after the `withFlushBeforeCompact` derivation) as a field and reuse it here.

This would require a `withTriggerMessages(int)` copy method. Non-blocking — flagging for future consideration.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/CompactionConfig.java
```diff
@@ -208,12 +208,18 @@ public String getSummaryPrompt() {
         return summaryPrompt;
     }
 
-    /** Whether to flush long-term memories from the prefix before compaction. */
+    /**
+     * Whether to flush long-term memories from the prefix before compaction. When used through
+     * {@code HarnessAgent}, {@code disableMemoryHooks()} overrides this setting to {@code false}.
```
AgentScopeJavaBot, line 213: [nitpick] The updated Javadoc says "When used through HarnessAgent, disableMemoryHooks() overrides this setting to false" — accurate and helpful. Consider adding a one-line note that this applies to both normal and emergency (context-overflow) compaction paths, so readers don't have to trace the code to confirm the fix covers both.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes issue #2694 where `disableMemoryHooks()` did not fully suppress long-term memory flush during compaction. The fix introduces an immutable `withFlushBeforeCompact(boolean)` copy method on `CompactionConfig` and derives an effective config in both the **normal compaction middleware path** and the **emergency context-overflow path** (`forceCompactAndRetry`), forcing `flushBeforeCompact=false` when memory hooks are disabled.

The approach is clean and minimal: the original `CompactionConfig` is never mutated (immutable copy via `copyBuilder()`), the `withEffective()` method is refactored to share the same helper (DRY), and `memoryHooksDisabled` is threaded through as a `final boolean` on the `HarnessAgent` instance. Javadoc is well-updated on all touched APIs. Test coverage is thorough — four integration scenarios covering both compaction paths plus a dedicated unit test for the `withFlushBeforeCompact` contract and immutability guarantee.

No critical issues found.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes issue #2694 where `disableMemoryHooks()` did not fully suppress long-term memory flush during compaction. The fix introduces an immutable `withFlushBeforeCompact(boolean)` copy method on `CompactionConfig` and derives an effective config in both the **normal compaction middleware path** and the **emergency context-overflow path** (`forceCompactAndRetry`), forcing `flushBeforeCompact=false` when memory hooks are disabled.

The approach is clean and minimal: the original `CompactionConfig` is never mutated (immutable copy via `copyBuilder()`), the `withEffective()` method is refactored to share the same helper (DRY), and `memoryHooksDisabled` is threaded through as a `final boolean` on the `HarnessAgent` instance. Javadoc is well-updated on all touched APIs. Test coverage is thorough — four integration scenarios covering both compaction paths plus a dedicated unit test for the `withFlushBeforeCompact` contract and immutability guarantee.

No critical issues found.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
AgentScopeJavaBot, line 1028: [recommended] **Pre-existing asymmetry worth noting:** The emergency path (`forceCompactAndRetry`) builds a fresh `CompactionConfig` from defaults (`triggerMessages(1)` + `flushBeforeCompact(!memoryHooksDisabled)`), discarding the user's other compaction settings (keepTokens, pruneConfig, summaryPrompt, offloadBeforeCompact, etc.). This is pre-existing behavior — the PR correctly adds only the `flushBeforeCompact` fix — but it means the emergency path doesn't fully respect the user's compaction config the way the normal path now does via `withFlushBeforeCompact(false)`.

The root cause is that `compactionConfig` is only a Builder field, not stored on the `HarnessAgent` instance. A future improvement could store the effective `CompactionConfig` (after the `withFlushBeforeCompact` derivation) as a field and reuse it here.

This would require a `withTriggerMessages(int)` copy method. Non-blocking — flagging for future consideration.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/CompactionConfig.java
AgentScopeJavaBot, line 213: [nitpick] The updated Javadoc says "When used through HarnessAgent, disableMemoryHooks() overrides this setting to false" — accurate and helpful. Consider adding a one-line note that this applies to both normal and emergency (context-overflow) compaction paths, so readers don't have to trace the code to confirm the fix covers both.

oss-maintainer:

```bash
git fetch upstream
git checkout codex/issue-2694-disable-compaction-flush
git rebase upstream/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1070,7 +1073,11 @@ private Mono<Msg> forceCompactAndRetry(List<Msg> msgs, RuntimeContext effective)
...
```
oss-maintainer, line 1079: `forceCompactAndRetry` still builds `forceConfig` from a *fresh* `CompactionConfig.builder()`, so the emergency path honours only `triggerMessages(1)` plus this new flush flag. Everything the caller configured on `.compaction(...)` — `offloadBeforeCompact(false)`, `keepMessages`/`keepTokens`, `summaryPrompt`, dedicated `model`, `truncateArgs`, `prune` — is dropped on this branch. That is the same shape as the bug this PR fixes: the normal path was gated, the overflow path was not. The branch is only reached when compaction *is* configured (the caller checks `!disableCompaction && compactionConfig != null`), so an `offloadBeforeCompact(false)` deployment still gets a raw-message JSONL write here.

Since `memoryHooksDisabled` is now plumbed into the instance anyway, plumbing the effective config too and reusing the helper this PR adds looks small:

```java
CompactionConfig forceConfig = configuredCompactionConfig
        .withFlushBeforeCompact(!memoryHooksDisabled);   // + a withTriggerMessages(1) variant
```

Non-blocking if you consider it out of scope — but worth a follow-up issue either way.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/CompactionConfig.java
```diff
@@ -246,12 +252,31 @@ public Model getModel() {
...
+     * @param flushBeforeCompact whether to flush long-term memories before compaction
+     * @return a new configuration with the requested flush setting
+     */
+    public CompactionConfig withFlushBeforeCompact(boolean flushBeforeCompact) {
+        Builder b = copyBuilder();
+        b.flushBeforeCompact = flushBeforeCompact;
+        return new CompactionConfig(b);
+    }
+
+    private Builder copyBuilder() {
```
oss-maintainer, line 273: `copyBuilder()` is the right extraction and both call sites are well covered. The residual hazard is that the copy is still a hand-maintained field list: a future 16th knob added to `Builder` compiles fine and is silently dropped by every `with*` method — exactly the failure mode behind the `forceCompactAndRetry` gap above.

Consider either a reflection-based test that asserts every declared field of the source survives the copy, or moving to a record so the compiler enforces the full list. Cheap insurance for a config object that now has two `with*` projections.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/CompactionConfig.java
```diff
@@ -246,12 +252,31 @@ public Model getModel() {
...
         Builder b = new Builder();
         b.triggerMessages = this.triggerMessages;
```
oss-maintainer, line 275: Minor: the copy shares the nested mutable configs (`TruncateArgsConfig`, `PruneConfig`) and `model` by reference — asserted deliberately with `assertSame` in `CompactionConfigTest`. Fine while those stay effectively immutable; a one-line javadoc note on `withFlushBeforeCompact` saying the copy is shallow would keep a later caller from mutating the returned config's prune settings and expecting isolation.

oss-maintainer (commented):
### Summary
Solid, well-tested fix. `disableMemoryHooks()` now really means "no memory flush", including inside compaction, and the new `withFlushBeforeCompact` / `copyBuilder()` extraction makes both projections expressible. The four behavioural tests in `HarnessAgentCompactionMemoryHooksTest` cover both hook states, the normal and the overflow flush path, and the explicit-disable case; EN and ZH docs are updated consistently. CI is green, CLA signed, mergeable.

No blocking issue found. The one gap worth tracking is that the emergency path (`forceCompactAndRetry`) still rebuilds its config from scratch, so it honours the new flush flag but drops the rest of the caller's compaction settings — the same class of asymmetry this PR removes for flush. Details inline.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1070,14 +1073,18 @@ private Mono<Msg> forceCompactAndRetry(List<Msg> msgs, RuntimeContext effective)
                         ? effective.getSessionId()
                         : "default";
 
-        CompactionConfig forceConfig = CompactionConfig.builder().triggerMessages(1).build();
+        CompactionConfig forceConfig = effectiveCompactionConfig.withTriggerMessages(1);
```
oss-maintainer: `effectiveCompactionConfig` is still a nullable field — `build()` only assigns it inside `if (!disableCompaction && compactionConfig != null)` (line 2611), and the previous code built a throwaway config here so this path could never NPE. Today the invariant holds only implicitly: `recoverFromOverflow()` gates on `compactionHook != null`, and both are assigned in the same block. Since this is the emergency overflow-recovery path, a future builder branch that wires `compactionHook` without the field (or a second constructor path) would surface as an NPE exactly when the agent is already in trouble. Could you make the coupling explicit, e.g. a local fallback (`CompactionConfig base = effectiveCompactionConfig != null ? effectiveCompactionConfig : CompactionConfig.defaults();`) or `Objects.requireNonNull(effectiveCompactionConfig, "compactionHook requires a compaction config")`, plus a comment stating `compactionHook != null` implies the field is set?

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1070,14 +1073,18 @@ private Mono<Msg> forceCompactAndRetry(List<Msg> msgs, RuntimeContext effective)
...
+        Model compactionModel =
+                effectiveCompactionConfig.getModel() != null
+                        ? effectiveCompactionConfig.getModel()
+                        : getModel();
         String effectiveFlushPrompt =
                 memoryConfig.flushPrompt() != null
                         ? memoryConfig.flushPrompt()
                         : MemoryFlushManager.DEFAULT_FLUSH_PROMPT;
         MemoryFlushManager fm =
```
oss-maintainer, line 1089: Behavior change worth calling out (it goes beyond "preserve compaction config on overflow"): `MemoryFlushManager` now gets `compactionModel` instead of `getModel()`, so when `CompactionConfig.model` is set — typically a cheaper/faster summarisation model — the pre-compaction long-term-memory flush prompts are issued against that model too, whereas before the overflow path always used the agent model. That may well be the intent, but if the flush prompt is tuned for the agent model, memory extraction quality can silently regress. Either split the two (`getModel()` for the flush manager, `compactionModel` for the summariser) or mention the change in the PR description / `docs/v2/*/harness/compaction.md`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/ConversationCompactor.java
```diff
@@ -214,6 +216,68 @@ public Mono<Optional<List<Msg>>> compactIfNeeded(
     // Trigger logic
     // -------------------------------------------------------------------------
 
+    private CompactionConfig resolveEffectiveConfig(CompactionConfig config) {
```
oss-maintainer, line 219: Moving dynamic-default resolution from `CompactionMiddleware` into the compactor removes a real duplication bug (the `forceCompactAndRetry` path used to bypass it), and the `@param`/shallow-copy javadoc updates read well. One follow-up: `ConversationCompactor` is public API, so any caller that passed a config with `triggerTokens == 0` / `keepTokens == -1` used to see the raw values and now sees resolved ones, and `resolveEffectiveConfig` runs on every `compactIfNeeded` call (one extra config copy per non-trivial turn). Worth a line in the changelog since the observable behaviour of the public entry point changed.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/ConversationCompactor.java
```diff
@@ -214,6 +216,68 @@ public Mono<Optional<List<Msg>>> compactIfNeeded(
...
+                int usable = contextWindow - config.getReserved();
+                effectiveKeep =
+                        Math.min(
+                                config.getKeepTokensMax(),
+                                Math.max(
+                                        config.getKeepTokensMin(),
+                                        (int) (usable * config.getKeepTokensRatio())));
+                log.debug("Dynamic keep tokens: {}", effectiveKeep);
+            } else {
+                effectiveKeep = 0;
```
oss-maintainer, line 272: `effectiveKeep = 0` when the model reports no context window means the caller silently falls back to `findMessageBasedCutoff(keepMessages)` in `determineCutoffIndex`. That is carried over verbatim from `CompactionMiddleware`, so not a regression, but on the overflow-recovery path a message-count-only keep is a much blunter instrument than the token-based keep — a truncated `keepTokensMin` floor (or a debug log when the fallback is taken) would make "why did my context collapse" easier to answer.

oss-maintainer (commented):
### Summary
Re-review after the new commit `21793739` (*"preserve compaction config on overflow"*) — the previous review covered `c7eaab2b` and still stands. Moving dynamic-default resolution out of `CompactionMiddleware` into `ConversationCompactor.resolveEffectiveConfig` fixes a real gap: the overflow-recovery path in `HarnessAgent.forceCompactAndRetry` used to build a throwaway `CompactionConfig`, so a user-tuned `triggerTokens` / `keepTokens` / truncate / prune / offload config was silently dropped exactly when the context was already overflowing. Inheriting `effectiveCompactionConfig.withTriggerMessages(1)` is the right shape, the shallow-copy javadoc on the new `withTriggerMessages` is consistent with `withEffective`, and the added coverage in `CompactionConfigTest` / `ConversationCompactorTest` / `HarnessAgentCompactionMemoryHooksTest` matches the change. CI is green (7/7).

### Findings
- **[Warning]** `HarnessAgent.java:1076` — `effectiveCompactionConfig` is nullable; the overflow path now depends on an implicit `compactionHook != null` invariant.
- **[Warning]** `HarnessAgent.java:1085` — `MemoryFlushManager` switches from the agent model to the compaction model; confirm that is intended.
- **[Info]** `ConversationCompactor.java:219` — public entry point now resolves dynamic defaults for every caller; worth a changelog note.
- **[Info]** `ConversationCompactor.java:272` — `effectiveKeep = 0` fallback collapses to message-based keep on the recovery path (pre-existing, not a regression).

### Suggestions
Both warnings are small and non-blocking: a local null guard (or `Objects.requireNonNull` with a message) next to `withTriggerMessages(1)`, and one sentence in the PR description about the flush-model change. Neither changes the correctness of the fix itself, so no re-review is needed unless you disagree with the model-choice point.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
RICARDO51167: Addressed in 21793739: emergency compaction now reuses the effective CompactionConfig via withTriggerMessages(1), including offload, keep, prompt, truncation, and pruning settings. It also honors the configured compaction model, with integration coverage added.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/CompactionConfig.java
RICARDO51167: Addressed in 21793739: added reflection-backed coverage for every declared CompactionConfig field, plus per-field copy assertions. Future fields now require an explicit copy-contract test update.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/CompactionConfig.java
RICARDO51167: Addressed in 21793739: withEffective, withFlushBeforeCompact, and withTriggerMessages now explicitly document their shallow-copy semantics. Nested configs remain immutable, and retaining the model reference is intentional.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
RICARDO51167: Addressed in ea9e78a5: forceCompactAndRetry now uses Objects.requireNonNull with a clear message, and the adjacent comment documents the compactionHook/config invariant.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
RICARDO51167: Agreed and addressed in ea9e78a5: MemoryFlushManager continues to use the agent model, while ConversationCompactor alone uses the configured compaction model. The integration test now verifies this split explicitly.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/ConversationCompactor.java
RICARDO51167: Addressed in ea9e78a5: added matching EN/ZH Unreleased release-note entries documenting dynamic trigger/retention resolution for direct ConversationCompactor callers.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/ConversationCompactor.java
RICARDO51167: Addressed in ea9e78a5: the unknown-context-window branch now logs that dynamic keepTokens is falling back to the configured keepMessages value.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1070,14 +1074,21 @@ private Mono<Msg> forceCompactAndRetry(List<Msg> msgs, RuntimeContext effective)
...
+                        effectiveCompactionConfig,
+                        "compactionHook requires an effective compaction config");
+        CompactionConfig forceConfig = baseConfig.withTriggerMessages(1);
+        Model compactionModel = baseConfig.getModel() != null ? baseConfig.getModel() : getModel();
         String effectiveFlushPrompt =
                 memoryConfig.flushPrompt() != null
                         ? memoryConfig.flushPrompt()
                         : MemoryFlushManager.DEFAULT_FLUSH_PROMPT;
         MemoryFlushManager fm =
                 new MemoryFlushManager(workspaceManager, getModel(), effectiveFlushPrompt);
```
oss-maintainer, line 1090: **[Info]** Resolved exactly as I hoped — the flush stays on `getModel()` while summarisation uses `baseConfig.getModel()`, the invariant is now explicit via `Objects.requireNonNull`, `ConversationCompactor.java:273` logs the no-context-window fallback, and both doc languages state the split. That closes all four points from my review on `21793739`. Approving.

One tiny coverage nit (not blocking): the rewritten `emergencyCompactionUsesConfiguredModelPromptAndOffloadSetting` now proves the half I care about least — `compactionModel.flushCalls == 0` — but nothing asserts the positive side, i.e. that the flush is issued against the **agent** model (a name like `...UsesAgentModelForMemoryFlush` would carry it), and `hasDailyMemoryFile()` is a proxy rather than an assertion about which model produced it. If `MemoryFlushManager` were ever switched back to `compactionModel`, would any test fail? Worth a look — a `reasoningModel.flushPromptContains(...)` style assertion would make the split unmissable.

docs/v2/en/docs/others/release-notes.md
```diff
@@ -7,6 +7,14 @@ This page tracks per-version changes for AgentScope Java 2.0. For the overall mi
 
 ---
 
+## Unreleased
+
+### Changed
+
+- `ConversationCompactor` now resolves dynamic trigger and retention settings for direct callers as well as middleware-managed compaction, keeping emergency recovery consistent with normal compaction ([#2699](https://github.com/agentscope-ai/agentscope-java/pull/2699))
```
oss-maintainer, line 14: **[Info]** The new `Unreleased → Changed` entry documents the `ConversationCompactor` resolution change, which is the one that needs a changelog line for direct callers — :+1:

The other externally visible behaviour in this PR is only in the body prose of `docs/v2/{en,zh}/docs/harness/compaction.md`: emergency compaction now reuses the configured `CompactionConfig` (retention, summary prompt, truncate/prune, offload) instead of a throwaway default, so an overflow recovery that previously used stock settings now follows the user's tuning. Anyone reading release notes rather than the compaction page will miss it. Consider one more bullet:

```markdown
- Emergency (overflow) compaction in `HarnessAgent` now reuses the effective `CompactionConfig`
  instead of default settings; only the message-count trigger is forced ([#2699](...))
```

Nit only, the docs edit itself is good.

oss-maintainer (approved):
### Summary

Approving on `ea9e78a5`. That commit answers every point from my review on `21793739`, and answers them in the way I would have picked:

- `forceCompactAndRetry` no longer reads a nullable field implicitly — `Objects.requireNonNull(effectiveCompactionConfig, ...)` makes the "compactionHook and the effective config are created together" invariant explicit at the boundary, with the comment saying so.
- The flush-model question is settled rather than papered over: memory extraction stays on `getModel()`, only summarisation follows `CompactionConfig.model`, and that split is now stated in both `docs/v2/en` and `docs/v2/zh` plus pinned by `flushCalls` counters in `HarnessAgentCompactionMemoryHooksTest`. Reverting the flush to the agent model also keeps this PR's blast radius to "emergency compaction reuses the user's compaction settings", which is the change the title advertises.
- The `effectiveKeep = 0` fallback is now logged, so a model that reports no context window is at least visible on the recovery path instead of silently degrading to a message-count cutoff.
- Release notes got an `Unreleased` entry for the `ConversationCompactor` resolution change — the part that affects callers who use it directly rather than through middleware.

Underlying fix is unchanged from the previous two passes and still correct: `resolveEffectiveConfig` in the compactor means the overflow-recovery path can no longer diverge from normal compaction, which was the actual bug in #2694.

CI is green on this head (ubuntu + windows build, validate, module-sync, codecov/patch, `license/cla`). Two non-blocking suggestions are inline — one about asserting the positive half of the flush-model split, one about a second release-notes bullet.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
RICARDO51167: Addressed in 82e112b4: the test is now named emergencyCompactionUsesAgentModelForFlushAndConfiguredModelForSummary, records the exact flush prompt, and asserts the agent model received it while the compaction model did not.

docs/v2/en/docs/others/release-notes.md
RICARDO51167: Addressed in 82e112b4: added a second matching EN/ZH Unreleased bullet stating that HarnessAgent emergency overflow compaction reuses the effective CompactionConfig and only forces the message-count trigger.

oss-maintainer (approved):
Re-reviewed the new commit (`82e112b4`) since my previous approval. The delta is test-only plus release-notes lines: `emergencyCompactionUsesAgentModelForFlushAndConfiguredModelForSummary` now pins *which* model sees *which* prompt (`flushPromptSeen` asserting the memory-extraction prompt reached the reasoning model, and `assertNull` on the configured compaction model) instead of only counting calls, and `RoutingModel` records the rendered flush prompt to support it. Renaming the test to match what it now asserts is the right call — the old name promised coverage of the offload setting that the assertions do not check.

The strengthened assertions match the behaviour I verified in the previous pass (flush goes to the agent model, summary goes to the configured model), so the intent of the fix stays pinned. LGTM.

---

PR #2712, opened 2026-08-14
fix(core): use call-scoped state for ReActAgent shutdown retry recovery

### LINKED ISSUE #2708: [Bug]: ReActAgent checks shutdown flag on the wrong AgentState in multi-session calls
## AgentScope-Java Version

2.0.2

## Description

`ReActAgent.doCallInner` checks the shutdown-interrupted flag through the agent instance:

```java
shutdownManager.checkAndClearShutdownInterrupted(ReActAgent.this)
```

However, `doCallInner` runs inside a per-call `CallExecution`, which already contains the `AgentState` resolved from the current `(userId, sessionId)` in `RuntimeContext`.

`GracefulShutdownManager.checkAndClearShutdownInterrupted(Agent)` then calls the no-argument `AgentBase.getAgentState()`. For `ReActAgent`, this resolves to the default slot `(null, defaultSessionId)` rather than the current call's state.

As a result, the shutdown retry recovery logic may read and clear a different `AgentState` from the one that was interrupted and persisted during shutdown.

## To Reproduce

1. Create one `ReActAgent` instance serving at least two sessions, for example `session-a` and `session-b`.
2. Call the agent with an explicit `RuntimeContext`, such as:
   - `userId = user-a`
   - `sessionId = session-a`
3. Let graceful shutdown interrupt the request and persist that session's `AgentState` with `shutdownInterrupted = true`.
4. Restart the application or retry the same request with the same `RuntimeContext`.
5. Observe that `doCallInner` checks the default agent state instead of the `AgentState` for `session-a`.

A deterministic unit test can set `shutdownInterrupted = true` on a non-default session state, invoke the call for that session, and verify that the duplicate input is discarded and only that session's flag is cleared.

## Expected Behavior

The shutdown-interrupted flag should be checked and cleared on the current call's `CallExecution.state`, identified by the current `RuntimeContext` `(userId, sessionId)`.

## Actual Behavior

The no-argument `getAgentState()` reads the default slot. For non-default users or sessions, the retry may:

- fail to discard the duplicate input;
- leave the interrupted session's flag set; or
- if the default slot is flagged, discard input belonging to another session.

## Root Cause

`ReActAgent` maintains per-session state in `stateCache`, keyed by `(userId, sessionId)`.

The call lifecycle correctly resolves the active state into `CallExecution.state`, and `AgentBase` binds that state to the shutdown request. However, `doCallInner` calls the legacy agent-level shutdown API.

`GracefulShutdownManager.checkAndClearShutdownInterrupted(Agent)` effectively does:

```java
AgentState state = agent.getAgentState();
```

And `ReActAgent.getAgentState()` delegates to the default session slot:

```java
getAgentState(null, defaultSessionId)
```

This is inconsistent with the call-scoped state model introduced in AgentScope 2.x.

## Suggested Fix

Add a shutdown check overload accepting the resolved `AgentState`, or otherwise make the check request/call-scoped.

For example:

```java
if (shutdownManager.checkAndClearShutdownInterrupted(state)) {
    msgs = List.of();
}
```

The existing agent-level overload may be retained for compatibility, but the `ReActAgent.doCallInner` path should use the current call's state.

## Environment

- AgentScope-Java: 2.0.2
- JDK: 17+
- Affected module: `agentscope-core`
- Affected classes:
  - `ReActAgent`
  - `GracefulShutdownManager`

jujn:
## Summary

Fixes #2708.

`ReActAgent` now checks and clears the `shutdownInterrupted` flag on the `AgentState` associated with the current `(userId, sessionId)` call.

## Root Cause

`ReActAgent.CallExecution` already resolved the correct per-session `AgentState`, but shutdown retry recovery called the agent-level API:
```java
checkAndClearShutdownInterrupted(ReActAgent.this)
```

For `ReActAgent`, this API reads the default session state instead of the state associated with the current `RuntimeContext`. This could cause duplicate input to be retained for an interrupted session, or cause input from another session to be discarded.

## Changes

- Added a call-scoped shutdown state check in `GracefulShutdownManager`.
- Updated `ReActAgent.CallExecution` to use its resolved `AgentState`.
- Preserved the existing agent-level API for backward compatibility.
- Added regression tests covering both non-default session recovery and cross-session isolation.

## Checklist


agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1701,7 +1701,7 @@ private Mono<Msg> doCallInner(List<Msg> msgs) {
             // by shutdown, the client is likely retrying with the same user prompt that already
             // exists in memory. Discard the duplicate input so the agent resumes purely from its
             // saved memory context.
-            if (shutdownManager.checkAndClearShutdownInterrupted(ReActAgent.this)) {
+            if (shutdownManager.checkAndClearShutdownInterruptedForState(state)) {
```
AgentScopeJavaBot, line 1743: [praise] Precise, minimal fix for the root cause. Using the call-scoped `state` (already resolved by `CallExecution` for the current `RuntimeContext`) instead of `ReActAgent.this` avoids the fallback to the default session's `AgentState`. This correctly handles multi-session agents where each `(userId, sessionId)` pair has its own state.

agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownManager.java
```diff
@@ -132,21 +132,34 @@ public void unbindStateSaver(Agent agent) {
...
-            st.setShutdownInterrupted(false);
+        return checkAndClearShutdownInterruptedForState(ab.getAgentState());
+    }
+
+    /**
+     * Check whether a call-scoped state was previously interrupted by shutdown, and clear the flag.
+     *
+     * <p>The supplied state must be the state resolved for the current call. This avoids falling
+     * back to an agent's default state when one agent instance serves multiple sessions.
+     *
```
AgentScopeJavaBot, line 156: [nit] Javadoc reads "Check whether a call-scoped state was previously interrupted" — consider "a call-scoped `AgentState`" or "the given `AgentState`" for precision, since "a call-scoped state" reads slightly awkwardly. Minor wording nit only.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentPerSessionStateTest.java
```diff
@@ -435,6 +435,48 @@ void userInterruptPersistsRecoveryState() throws Exception {
...
+        assertEquals("ok", response.getTextContent());
+        assertFalse(sessionState.isShutdownInterrupted());
+        assertFalse(defaultState.isShutdownInterrupted());
+        assertTrue(
+                sessionState.getContext().stream()
+                        .noneMatch(msg -> "duplicate prompt".equals(msg.getTextContent())),
+                "the retry input must be discarded for the interrupted session");
+
+        ReActAgent otherAgent =
+                ReActAgent.builder().name("asst").sysPrompt("hi").model(new NoopModel()).build();
```
AgentScopeJavaBot, line 462: [praise] Well-structured regression test covering both the positive case (session-specific `shutdownInterrupted` flag cleared, duplicate input discarded) and the critical cross-session isolation case (default-session flag must NOT affect another session's input). The two-agent design cleanly verifies that state leakage cannot occur across agent instances either.

AgentScopeJavaBot (approved):
🤖 AI Review

This is a well-targeted fix for a real multi-session bug in `ReActAgent` shutdown retry recovery. The root cause analysis is precise: `CallExecution.doCallInner` was calling the agent-level `checkAndClearShutdownInterrupted(ReActAgent.this)` which reads the **default** session's state via `getAgentState()` → `getAgentState(null, defaultSessionId)`, instead of using the already-resolved call-scoped `state` field. For agents serving multiple `(userId, sessionId)` sessions, this caused either duplicate input retention for the interrupted session or incorrect input discard for a non-interrupted session.

The fix is minimal and correct: extract the core check-and-clear logic into `checkAndClearShutdownInterruptedForState(AgentState)`, call it from `CallExecution` with the resolved `state`, and retain the agent-level API for backward compatibility (delegating to the new method). The regression test is well-designed, covering both the positive case (session-specific flag cleared, input discarded) and the cross-session isolation case (default-session flag does NOT affect another session's input). The `GracefulShutdownTest` addition covers the new method's null/false/true/clear contract.

No correctness issues found. Backward compatibility is preserved — the old `checkAndClearShutdownInterrupted(Agent)` behaves identically for default-session callers.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot, line 1743: [praise] Precise, minimal fix for the root cause. Using the call-scoped `state` (already resolved by `CallExecution` for the current `RuntimeContext`) instead of `ReActAgent.this` avoids the fallback to the default session's `AgentState`. This correctly handles multi-session agents where each `(userId, sessionId)` pair has its own state.

agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownManager.java
AgentScopeJavaBot, line 156: [nit] Javadoc reads "Check whether a call-scoped state was previously interrupted" — consider "a call-scoped `AgentState`" or "the given `AgentState`" for precision, since "a call-scoped state" reads slightly awkwardly. Minor wording nit only.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentPerSessionStateTest.java
AgentScopeJavaBot, line 462: [praise] Well-structured regression test covering both the positive case (session-specific `shutdownInterrupted` flag cleared, duplicate input discarded) and the critical cross-session isolation case (default-session flag must NOT affect another session's input). The two-agent design cleanly verifies that state leakage cannot occur across agent instances either.

AgentScopeJavaBot (approved):
🤖 AI Review

This is a well-targeted fix for a real multi-session bug in `ReActAgent` shutdown retry recovery. The root cause analysis is precise: `CallExecution.doCallInner` was calling the agent-level `checkAndClearShutdownInterrupted(ReActAgent.this)` which reads the **default** session's state via `getAgentState()` → `getAgentState(null, defaultSessionId)`, instead of using the already-resolved call-scoped `state` field. For agents serving multiple `(userId, sessionId)` sessions, this caused either duplicate input retention for the interrupted session or incorrect input discard for a non-interrupted session.

The fix is minimal and correct: extract the core check-and-clear logic into `checkAndClearShutdownInterruptedForState(AgentState)`, call it from `CallExecution` with the resolved `state`, and retain the agent-level API for backward compatibility (delegating to the new method). The regression test is well-designed, covering both the positive case (session-specific flag cleared, input discarded) and the cross-session isolation case (default-session flag does NOT affect another session's input). The `GracefulShutdownTest` addition covers the new method's null/false/true/clear contract.

No correctness issues found. Backward compatibility is preserved — the old `checkAndClearShutdownInterrupted(Agent)` behaves identically for default-session callers.

chickenlj (approved):

PR #2716, opened 2026-08-15
fix(harness): route skill_manage operations through the call-scoped RuntimeContext

### LINKED ISSUE #2688: WorkspaceSkillRepository contextSupplier有并发的bug
io.agentscope.harness.agent.skill.WorkspaceSkillRepository#WorkspaceSkillRepository(io.agentscope.harness.agent.filesystem.AbstractFilesystem, java.lang.String, java.util.function.Supplier<io.agentscope.core.agent.RuntimeContext>, java.lang.String, boolean)

获取agentSkill需要拿到一个runtimeContext，然后这个runtimeContext是 reactAgent中的activeRc，如果并发情况下，拿到的可能是别人的runtimeContext

birdie7761:
## AgentScope-Java Version

2.0.3-SNAPSHOT (main @ `00215511`)

## Description

Fixes #2688

`WorkspaceSkillRepository` resolved its context on every write-path operation from the
legacy shared-context supplier — the single `volatile RuntimeContext activeRc` field on
the agent (`ReActAgent.java:282`, wired into the repository at
`HarnessAgent.java:2614-2619`). Under concurrent calls on a shared agent instance, that
field holds whichever call set it last (and is cleared to `null` when the owning call
completes), so `skill_manage` could:

- create / edit / patch a skill **in another user's namespace** (cross-user leak),
- `delete` **archive a stranger's same-named skill** instead of the caller's,
- or, when the supplier returns `null`/empty, **escape user isolation** and write into
  the shared root.

The read path already solved this: `RuntimeContextSkillRepository.getAllSkills(ctx)` +
`HarnessSkillMiddleware` pass the call-scoped context, and `SkillLoadTool` resolves via
`param.getRuntimeContext()` (its javadoc explicitly notes this keeps concurrent calls
correct). The write path behind `SkillManageTool` was left on the racy supplier.

This PR mirrors the established call-scoped pattern on the write path:

- `WorkspaceSkillRepository` gains call-scoped overloads for `getSkill`, `skillExists`,
  `save`, `delete`, `readSkillFile`, `writeSkillFile` and `deleteSkillFile`. Each
  resolves every read and write against the supplied context; a `null` context falls
  back to the legacy supplier so out-of-call callers (curator / promoter / stager) keep
  their current behavior. The existing no-arg methods delegate with `null`, so they are
  unchanged. The class javadoc documents this `null` contract explicitly, including how
  it differs from the overridden `getAllSkills(RuntimeContext)` (which preserves the
  interface contract of normalising `null` to `RuntimeContext.empty()`).
- `SkillManageTool` threads `param.getRuntimeContext()` (populated by `ReActAgent` /
  `ToolExecutor` on every real tool call) through all six actions (`create` / `edit` /
  `patch` / `write_file` / `remove_file` / `delete`), including the security-scan
  rollback paths.

Known limitation (unchanged by this PR): a same-user concurrent `create` of an
identically named skill still has an exists-then-save window (last write wins). That is
pre-existing single-user behavior, orthogonal to the cross-user routing fixed here; the
proper fix is atomic exclusive-create semantics in the filesystem backends, not
repo-level locking.

### Regression tests

New `SkillManageToolCallScopedContextTest` reproduces the issue deterministically: it
wires the repository exactly like `HarnessAgent` does (`IsolationScope.USER` namespace
factory + a supplier reading a shared `AtomicReference` simulating `activeRc`), pins
that supplier to user *bob* (or `null`) while driving the tool with user *alice*'s
context, and asserts:

1. `create` lands under `alice/skills/...` — not `bob/skills/...` and not the shared root;
2. `create` still routes to alice when the shared supplier context is cleared
   (post-`afterAgentExecution` window);
3. `delete` archives alice's copy and leaves bob's same-named skill intact;
4. `patch` reads and writes alice's copy only;
5. the **security-scan rollback** paths for `edit` / `write_file` (triggered by a
   DANGEROUS `rm -rf /` body) never touch another user's namespace — a recording
   `LocalFilesystem` subclass asserts every operation the action performed, rollback
   included, resolved against the tool-call user's context (pre-fix recordings show
   `[bob, bob, bob, bob, bob]`);
6. `create` in **drafts mode** (`autoPromote=false`, the default config) also stays
   inside the caller's namespace.

All seven tests fail on current `main` and pass with this change; the four end-state
tests and both rollback tests were additionally re-run against pre-fix code to confirm
the failing assertions.

Related but out of scope (different subsystems, mentioned in #2679 / #2683): the
workspace-vs-user skill **priority** under isolation (#2679 part 1) and the shared
`knowledge/` fallback for `read_file` (#2683) live in repository composition /
filesystem routing and deserve separate fixes.

### How to test

```bash
# focused regression tests for this change
mvn -pl agentscope-harness -am -Dtest=SkillManageToolCallScopedContextTest \
    -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test

# full module regression
mvn -pl agentscope-core -am install -DskipTests
mvn -pl agentscope-harness test   # 835 tests, 0 failures
```

## Checklist

- [x]  All tests are passing (`mvn test` — agentscope-harness: 835 tests, 0 failures)
- [x]  Related documentation has been updated (no doc changes needed — internal context resolution only)


agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/WorkspaceSkillRepository.java
```diff
@@ -135,10 +146,20 @@ public WorkspaceSkillRepository(
...
+        return getSkill(name, null);
+    }
+
+    /**
+     * Returns the named skill visible to the supplied context. A {@code null} context falls back
+     * to the legacy shared-context supplier, so out-of-call callers keep their current behavior;
+     * in-call callers must pass the call-scoped context to stay correct under concurrency.
+     */
+    @Override
+    public AgentSkill getSkill(String name, RuntimeContext context) {
```
AgentScopeJavaBot, line 158: [nit] The `isEmpty()` → `isBlank()` change here is a subtle behavior improvement — whitespace-only names (e.g. `"  "`) would have passed the old guard but now correctly return `null`. This aligns with the other methods in this class that already use `isBlank()`. Not a bug, just worth noting as a minor behavioral refinement bundled with the concurrency fix.

agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/SkillManageToolCallScopedContextTest.java
```diff
@@ -0,0 +1,424 @@
...
+                                        alice))
+                        .block();
+
+        assertTrue(text(r).startsWith("Error:"), "the DANGEROUS edit should be blocked");
+        String aliceFile = Files.readString(workspace.resolve("alice/skills/demo/SKILL.md"));
+        assertTrue(
+                aliceFile.contains("Alice body.") && !aliceFile.contains("rm -rf"),
+                "the caller's original content must be restored by the rollback");
+        String bobFile = Files.readString(workspace.resolve("bob/skills/demo/SKILL.md"));
+        assertTrue(
```
AgentScopeJavaBot, line 320: [praise] Excellent test design: the `RecordingLocalFilesystem` subclass captures every filesystem operation (including rollback writes triggered by the security scanner) and asserts that **all** operations — not just the final state — resolved against the tool-call user's context. This catches rollback-path cross-user leaks that end-state-only assertions would miss. The pre-fix recordings showing `[bob, bob, bob, bob, bob]` vs the post-fix `[alice, alice, alice, alice, alice]` make the fix's impact immediately visible.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/WorkspaceSkillRepository.java
```diff
@@ -248,6 +277,14 @@ public SkillResources resourcesFor(String skillName, RuntimeContext ctx) {
 
     @Override
     public boolean save(List<AgentSkill> skills, boolean force) {
+        return save(skills, force, null);
+    }
+
```
AgentScopeJavaBot, line 282: [nit] Consider promoting `save(List<AgentSkill>, boolean, RuntimeContext)` and `delete(String, RuntimeContext)` to the `RuntimeContextSkillRepository` interface (as default methods delegating to the no-arg versions) for API symmetry with the read path (`getAllSkills(RuntimeContext)` and `getSkill(String, RuntimeContext)` are already on the interface). This is not blocking — `SkillManageTool` already depends on the concrete type — but would make the context-aware write path a first-class part of the repository contract if other implementations need the same pattern in the future.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a critical concurrency bug (#2688) where `WorkspaceSkillRepository`'s write-path operations resolved their context from a shared mutable supplier (`volatile RuntimeContext activeRc`), allowing cross-user data leaks under concurrent calls. The fix mirrors the established call-scoped pattern from the read path: `WorkspaceSkillRepository` gains `RuntimeContext`-parameterized overloads for all write-path methods, and `SkillManageTool` threads `param.getRuntimeContext()` through all six actions including security-scan rollback paths. The null-convention contract is clearly documented, backward compatibility is preserved (no-arg methods delegate with `null`), and the regression test suite is excellent — 7 tests deterministically reproduce the bug using a recording filesystem subclass that validates rollback-path isolation. This is a well-crafted, production-ready fix.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/WorkspaceSkillRepository.java
AgentScopeJavaBot, line 158: [nit] The `isEmpty()` → `isBlank()` change here is a subtle behavior improvement — whitespace-only names (e.g. `"  "`) would have passed the old guard but now correctly return `null`. This aligns with the other methods in this class that already use `isBlank()`. Not a bug, just worth noting as a minor behavioral refinement bundled with the concurrency fix.

agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/SkillManageToolCallScopedContextTest.java
AgentScopeJavaBot, line 320: [praise] Excellent test design: the `RecordingLocalFilesystem` subclass captures every filesystem operation (including rollback writes triggered by the security scanner) and asserts that **all** operations — not just the final state — resolved against the tool-call user's context. This catches rollback-path cross-user leaks that end-state-only assertions would miss. The pre-fix recordings showing `[bob, bob, bob, bob, bob]` vs the post-fix `[alice, alice, alice, alice, alice]` make the fix's impact immediately visible.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/WorkspaceSkillRepository.java
AgentScopeJavaBot, line 282: [nit] Consider promoting `save(List<AgentSkill>, boolean, RuntimeContext)` and `delete(String, RuntimeContext)` to the `RuntimeContextSkillRepository` interface (as default methods delegating to the no-arg versions) for API symmetry with the read path (`getAllSkills(RuntimeContext)` and `getSkill(String, RuntimeContext)` are already on the interface). This is not blocking — `SkillManageTool` already depends on the concrete type — but would make the context-aware write path a first-class part of the repository contract if other implementations need the same pattern in the future.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a critical concurrency bug (#2688) where `WorkspaceSkillRepository`'s write-path operations resolved their context from a shared mutable supplier (`volatile RuntimeContext activeRc`), allowing cross-user data leaks under concurrent calls. The fix mirrors the established call-scoped pattern from the read path: `WorkspaceSkillRepository` gains `RuntimeContext`-parameterized overloads for all write-path methods, and `SkillManageTool` threads `param.getRuntimeContext()` through all six actions including security-scan rollback paths. The null-convention contract is clearly documented, backward compatibility is preserved (no-arg methods delegate with `null`), and the regression test suite is excellent — 7 tests deterministically reproduce the bug using a recording filesystem subclass that validates rollback-path isolation. This is a well-crafted, production-ready fix.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/WorkspaceSkillRepository.java
birdie7761: Thanks for the review! On promoting the ctx-aware `save`/`delete` overloads to `RuntimeContextSkillRepository`: I kept this concrete-first intentionally, for two reasons.

1. A default method delegating to the no-arg versions would make any other implementation **silently ignore** the supplied context — exactly the failure mode this PR removes (a fallback that looks correct but routes to the wrong namespace). If the contract moves to the interface, implementations that cannot honor it should fail loudly (abstract method or `UnsupportedOperationException`), not silently no-op.
2. Making them abstract is a breaking interface change for existing implementors, so I'd rather leave that call to the maintainers than bundle it into a bugfix PR.

Happy to submit a follow-up PR promoting the contract (with loud-failure semantics) if that's wanted — `SkillManageTool` already depends only on the concrete type, so nothing here blocks on it.

oss-maintainer (approved):
### Summary
First review. `skill_manage` operations now resolve the workspace skill repository from the call-scoped `RuntimeContext` (`ToolCallParam#getRuntimeContext()`) instead of the shared-context supplier, so concurrent tool calls on one agent instance read and write the caller's own skill namespace. The two-context-source contract (explicit `null` = "fall back to the legacy supplier" on the write-path overloads, versus the interface contract of normalising `null` to `RuntimeContext#empty()` in `getAllSkills(RuntimeContext)`) is documented at the class level, which is the right call because the asymmetry is easy to misuse later. `createWritesUnderToolCallUserEvenWhenSupplierPointsAtAnotherUser`, `deleteArchivesCallersSkillAndLeavesOtherUsersCopyIntact`, `editRollbackNeverTouchesAnotherUsersNamespace` and friends assert exactly the cross-user isolation failure mode rather than just checking the call succeeds. No correctness or compatibility issue found; the deprecated no-arg paths keep their previous behaviour. LGTM.

---

PR #2720, opened 2026-08-15
feat(core): expose tool call ID from ToolEmitter

### LINKED ISSUE #695: Feature Request: Add getToolCallId() to ToolEmitter interface
## Background

When implementing streaming tools (e.g., bash/terminal execution), tools need to correlate their output with the frontend UI. The `toolCallId` is the natural identifier for this correlation, but currently tools have no way to access it through the `ToolEmitter` interface.

## Problem

Tools like `BashTool` need to:
1. Write streaming output to a shared buffer
2. Allow frontend to poll/subscribe to this buffer by a unique session ID
3. The natural session ID is `toolCallId`, which is already available in `ToolUseBlock`

Without access to `toolCallId`, tools must generate random UUIDs, causing frontend-backend ID mismatch.

## Proposed Solution

Add a default method to `ToolEmitter` interface:

```java
public interface ToolEmitter {
    void emit(ToolResultBlock chunk);
    
    /**
     * Returns the tool call ID associated with this emitter.
     * This allows tools to correlate their output with frontend components.
     * 
     * @return the tool call ID, or null if not available
     */
    default String getToolCallId() {
        return null;
    }
}
```

And implement it in `DefaultToolEmitter`:

```java
@Override
public String getToolCallId() {
    return toolUseBlock != null ? toolUseBlock.getId() : null;
}
```

## Use Cases

1. **Terminal/Bash Tools**: Write output to a shared `TerminalBuffer` keyed by `toolCallId`, allowing frontend to display real-time output
2. **Long-running Tasks**: Track progress and allow cancellation by `toolCallId`
3. **Streaming File Operations**: Correlate file upload/download progress with UI components

## Backward Compatibility

The default method returns `null`, ensuring existing implementations continue to work without modification.

## Current Workaround

We've implemented this locally, but would prefer official framework support for maintainability.

---

Thank you for considering this feature request!

skt-shinyruo:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Adds a backward-compatible way for streaming tools to access the ID of the current tool call.

- Add the default `ToolEmitter#getToolCallId()` method, returning `null` when call context is unavailable.
- Return the associated `ToolUseBlock` ID from `DefaultToolEmitter`.
- Add focused regression tests for associated and unavailable tool call IDs.
- Document the API and nullable behavior in the English and Chinese v2 tool guides.

Closes #695.

### Testing

- `mvn -pl agentscope-core -Dtest=DefaultToolEmitterTest,NoOpToolEmitterTest test`
- Full reactor `mvn test`, completed with Maven `-rf` resumes after transient Maven Central and npm registry connection failures
- `mvn spotless:check`

All reactor modules completed successfully; the interrupted runs failed only while downloading dependencies, with no test failures.

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/DefaultToolEmitter.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/DefaultToolEmitter.java
@@ -57,4 +57,14 @@ class DefaultToolEmitter implements ToolEmitter {
             chunkCallback.accept(toolUseBlock, chunk);
         }
     }
+
+    /**
+     * Returns the ID of the tool call associated with this emitter.
+     *
+     * @return The tool call ID, or {@code null} if no tool call is available
+     */
+    @Override
+    public String getToolCallId() {
+        return toolUseBlock != null ? toolUseBlock.getId() : null;
+    }
 }
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolEmitter.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolEmitter.java
@@ -68,4 +68,16 @@ public interface ToolEmitter {
      * @param chunk The chunk to emit
      */
     void emit(ToolResultBlock chunk);
+
+    /**
+     * Returns the ID of the tool call associated with this emitter.
+     *
+     * <p>The default implementation returns {@code null} so custom emitters without tool call
+     * context remain compatible.
+     *
+     * @return The tool call ID, or {@code null} if it is not available
+     */
+    default String getToolCallId() {
+        return null;
+    }
 }
--- a/agentscope-core/src/test/java/io/agentscope/core/tool/DefaultToolEmitterTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/tool/DefaultToolEmitterTest.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.tool;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import io.agentscope.core.message.ToolUseBlock;
+import java.util.Map;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+/** Tests for DefaultToolEmitter. */
+@DisplayName("DefaultToolEmitter Tests")
+class DefaultToolEmitterTest {
+
+    @Test
+    @DisplayName("getToolCallId() should return the associated tool call ID")
+    void testGetToolCallId() {
+        ToolUseBlock toolUseBlock =
+                ToolUseBlock.builder().id("call-123").name("test_tool").input(Map.of()).build();
+        ToolEmitter emitter = new DefaultToolEmitter(toolUseBlock, null);
+
+        assertEquals("call-123", emitter.getToolCallId());
+    }
+
+    @Test
+    @DisplayName("getToolCallId() should return null when no tool call is available")
+    void testGetToolCallIdWithoutToolUseBlock() {
+        ToolEmitter emitter = new DefaultToolEmitter(null, null);
+
+        assertNull(emitter.getToolCallId());
+    }
+}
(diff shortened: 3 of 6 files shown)


AgentScopeJavaBot (approved):
🤖 AI Review

This PR exposes the tool call ID from `ToolEmitter` in a backward-compatible manner. The implementation is clean: a default method on the `ToolEmitter` interface returns `null`, while `DefaultToolEmitter` returns the actual ID from the associated `ToolUseBlock`. Tests cover both the happy path and the default (null) case. Documentation is updated in both English and Chinese guides. No issues found — well-structured, minimal, and thoroughly tested change.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR exposes the tool call ID from `ToolEmitter` in a backward-compatible manner. The implementation is clean: a default method on the `ToolEmitter` interface returns `null`, while `DefaultToolEmitter` returns the actual ID from the associated `ToolUseBlock`. Tests cover both the happy path and the default (null) case. Documentation is updated in both English and Chinese guides. No issues found — well-structured, minimal, and thoroughly tested change.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolEmitter.java
```diff
@@ -68,4 +68,16 @@ public interface ToolEmitter {
...
+    /**
+     * Returns the ID of the tool call associated with this emitter.
+     *
+     * <p>The default implementation returns {@code null} so custom emitters without tool call
+     * context remain compatible.
+     *
+     * @return The tool call ID, or {@code null} if it is not available
+     */
+    default String getToolCallId() {
+        return null;
```
oss-maintainer, line 81: `default` returning `null` is the right way to keep the interface source-compatible, but it also means the new accessor silently reports "no id" on paths where the framework injects `NoOpToolEmitter.INSTANCE` instead of `DefaultToolEmitter`. `ToolCallParam#getEmitter()` falls back to the no-op singleton whenever the param was built without `.emitter(...)` (e.g. the structured-output invocation in `ReActAgent` builds a `ToolCallParam` with only `toolUseBlock`/`agent`/`runtimeContext`), and in that case `param.getToolUseBlock().getId()` is available while `getToolCallId()` returns `null`.

Two options to make the capability predictable:

1. Resolve the emitter from the tool call itself, e.g. in `ToolCallParam#getEmitter()` return `new DefaultToolEmitter(toolUseBlock, null)` when `emitter == null && toolUseBlock != null` (emit stays a no-op because the callback is null), so the id is reported consistently.
2. Or keep the current fallback but state it explicitly here: "`null` is also returned for the framework's no-op emitter, i.e. when the tool was invoked without an emitter instance — correlate with the tool result block id in that case."

Option 1 is my preference, since the doc example added here (`if (toolCallId != null) progressByToolCall.put(...)`) would otherwise silently skip registration for part of the call paths.

agentscope-core/src/main/java/io/agentscope/core/tool/DefaultToolEmitter.java
```diff
@@ -57,4 +57,14 @@ public void emit(ToolResultBlock chunk) {
...
     }
+
+    /**
+     * Returns the ID of the tool call associated with this emitter.
+     *
+     * @return The tool call ID, or {@code null} if no tool call is available
+     */
+    @Override
+    public String getToolCallId() {
+        return toolUseBlock != null ? toolUseBlock.getId() : null;
```
oss-maintainer, line 68: The null-guard covers `toolUseBlock == null`, but `ToolUseBlock#getId()` itself can be null for blocks built without an id (locally constructed calls / some replay paths). Worth documenting that second null source in the `@return` line, so callers know a `null` return does not only mean "no tool call context".

agentscope-core/src/test/java/io/agentscope/core/tool/DefaultToolEmitterTest.java
```diff
@@ -0,0 +1,47 @@
...
+class DefaultToolEmitterTest {
+
+    @Test
+    @DisplayName("getToolCallId() should return the associated tool call ID")
+    void testGetToolCallId() {
+        ToolUseBlock toolUseBlock =
+                ToolUseBlock.builder().id("call-123").name("test_tool").input(Map.of()).build();
+        ToolEmitter emitter = new DefaultToolEmitter(toolUseBlock, null);
+
+        assertEquals("call-123", emitter.getToolCallId());
```
oss-maintainer, line 37: Nice unit coverage for both branches. One gap relative to what users actually rely on: nothing here asserts the id survives the injection path. A small case in `ToolEmitterIntegrationTest` (which already wires a real `@Tool` method with a `ToolEmitter` parameter) asserting `emitter.getToolCallId()` equals the `ToolUseBlock` id the agent dispatched would lock down the end-to-end contract and catch the `NoOpToolEmitter` fallback discussed above.

oss-maintainer (commented):
### Summary
Exposes the current tool-call id through `ToolEmitter` so tools can correlate progress with a specific call. The change is backward-compatible (`default` method + `null` fallback), covered by new unit tests, and documented in both the English and Chinese tool guides — a clean, well-scoped addition. No blocking issues; one API-consistency point and two minor notes below.

### Findings
- **[Warning]** `ToolEmitter.java:81` — the `default` accessor returns `null` even on paths where the framework does have a tool-call id, because `ToolCallParam.getEmitter()` falls back to the id-less `NoOpToolEmitter.INSTANCE` whenever the execution param was built without `.emitter(...)`; the id is present on `param.getToolUseBlock()` in exactly those cases.
- **[Info]** `DefaultToolEmitter.java:68` — the null-guard covers a missing `ToolUseBlock`, but `ToolUseBlock.getId()` itself can be null; the javadoc should mention that second case.
- **[Info]** `DefaultToolEmitterTest.java:37` — both accessor branches are covered, but nothing pins the end-to-end injection contract that tool authors rely on.

### Suggestions
Prefer deriving the emitter from the tool call in `ToolCallParam.getEmitter()` (e.g. `new DefaultToolEmitter(toolUseBlock, null)` when no emitter was configured — `emit()` stays a no-op because the callback is null), so `getToolCallId()` never discards information the param already carries. If the current fallback is intentional, state it in the javadoc, because the new doc example keys a map on the returned id and would silently skip registration for those call paths.

---

PR #2731, opened 2026-08-16
fix(core): reload authoritative state before permission admin writes

### LINKED ISSUE #2535: [Bug]:setPermissionMode 读取未回源的进程内 stateCache 并整体写回，多副本下会静默覆盖更新的 AgentState
**版本:** 2.0.0（在发布包上验证；`main` 分支同一份代码）
**模块:** `agentscope-core`，经 `agentscope-harness` 使用
**环境:** 多副本部署（Kubernetes，>1 pod），通过 `.stateStore(...)` 配置 `MysqlAgentStateStore`

## 概述

`ReActAgent#setPermissionMode(...)` 从进程内的 `stateCache` 读取会话的 `AgentState`，**不从已配置的 `AgentStateStore` 回源**，随后又通过 `saveAgentState(...)` 把该实例整体写回存储。

在多副本部署下，这会把**陈旧**状态持久化，静默截断对话历史；也会把一次失败调用遗留的、本不该落盘的半截状态扶正进存储，造成该会话永久不可用。

## 机制

```java
// ReActAgent#setPermissionMode(String, String, PermissionMode)
public void setPermissionMode(String userId, String sessionId, PermissionMode mode) {
    Objects.requireNonNull(mode, "mode must not be null");
    String sid = (sessionId == null || sessionId.isBlank()) ? defaultSessionId : sessionId;
    String slot = slotKey(userId, sid);
    AgentState s = getAgentState(userId, sid);                  // (1) 读缓存，不回源
    s.setPermissionContext(s.getPermissionContext().withMode(mode));
    permissionEngineCache.put(slot, new PermissionEngine(s.getPermissionContext()));
    saveAgentState(userId, sid);                                // (2) 把【整个】 AgentState 写回
}

// (1) ReActAgent#getAgentState(String, String)
public AgentState getAgentState(String userId, String sessionId) {
    return stateCache.computeIfAbsent(slot, k -> loadOrCreateAgentStateForSlot(...));
    //                ^^^^^^^^^^^^^^^ 命中即返回缓存实例，从不回源
}

// (2) ReActAgent#saveAgentState(String, String)
public void saveAgentState(String userId, String sessionId) {
    if (stateStore == null) return;
    AgentState s = stateCache.get(slotKey(userId, sessionId));
    if (s != null) stateStore.save(userId, sessionId, "agent_state", s);
}
```

`stateCache` 是 agent 实例上的普通 `ConcurrentHashMap`。在 `agentscope-core` 与 `agentscope-harness` 中，对它**没有任何** `remove` / `clear` / 失效 / TTL，`activateSlotForContext` 放进去的条目会驻留至 JVM 结束。

### 故障 A —— 跨副本陈旧覆盖（丢失历史）

请求按 pod 分散路由，且会话跨多次调用时：

```
调用1 -> pod A：activateSlotForContext 回源加载，stateCache.put(slot, s_A)
                调用结束，存储为 [u1, a1]；pod A 仍持有 s_A == [u1, a1]
调用2 -> pod B：存储推进到 [u1, a1, u2, a2]
调用3 -> pod A：setPermissionMode -> getAgentState -> 缓存命中 -> s_A == [u1, a1]（陈旧）
                saveAgentState -> 存储被覆盖为 [u1, a1]
                u2 / a2 丢失；无异常，无日志
```

### 故障 B —— 把未落盘的半截状态扶正（毒丸会话）

`saveStateToSession` 只在成功路径执行，因此中途失败的调用**不会**被持久化，这是正确的。但它改动的那个 `AgentState` 实例，正是 `stateCache` 里的那一个——`activateSlotForContext` 执行 `stateCache.put(slot, loaded)` 并把同一引用交给 `CallExecution`。没有任何机制清除它。

若该会话稍后的调用先执行了 `setPermissionMode`，这份半截状态就会被写入存储，例如末尾的 assistant 消息带着 `tool_use` 却没有配对的 `tool_result`。此后每次调用都会回源读到它，`doCallInner` 抛出：

```
java.lang.IllegalStateException: Pending tool calls exist without results.
Enable enablePendingToolRecovery or provide tool results. Pending IDs: [...]
```

该会话从此不可用。我们在生产环境遇到了这个问题：落盘行的最后一条 assistant 消息带有 4 个 `"state":"allowed"` 的 `tool_use` 且无任何 `tool_result`，同时 `shutdown_interrupted: false`（排除了停机状态保存器）。

`enablePendingToolRecovery(true)` 确实能从故障 B 中恢复，但那是治标——不一致的状态仍然进入了存储。

## 为什么这不符合预期

v2.0.0-RC3 的 release note 写道：

> **Distributed state always fresh** — when a state store is configured, agent state and permissions are reloaded from the store at the start of every call, preventing stale cache reads when sessions drift across machines.

`activateSlotForContext` 确实遵守了这一点，而 `setPermissionMode` 绕开了它，并且：

- `setPermissionMode(RuntimeContext, PermissionMode)` 这个重载签名看起来就是每请求 API；
- 它自身的 javadoc 说明了模式变更的语义，却完全没有提到「读取可能陈旧的缓存」以及「重写整个 `AgentState`」；
- 相关警告位于下一层的 `getAgentState` javadoc 中（"returns the locally cached instance ... suitable for admin APIs and tests"）。

一个配置了 `AgentStateStore`、并读过 RC3 那条说明的使用者，有理由认为权限切换同样是与存储一致的。

## 修复建议（任一即可）

1. **配置了 store 时先回源再修改。** 让 `setPermissionMode` 走与 `activateSlotForContext` 相同的权威回源路径并替换缓存条目，而不是 `computeIfAbsent`。
2. **缩小写入范围。** 只持久化 `permission_context` 而非整个 `AgentState`，使该调用无法破坏 `context`。（需要 `AgentStateStore` 提供局部更新能力。）
3. **文档 + 防护。** 至少在 `setPermissionMode` 上注明：它读取进程内状态并重写整个 `AgentState`，在多副本部署下不应按请求调用。可选地，当同一 slot 有调用正在进行时打印告警。

方案 1 最贴近 RC3 的意图，但需要处理一个边界：并发的 in-flight 调用仍持有旧实例的引用。

如果维护者认可其中某个方向，我很乐意提交 PR。

## 变通方案

不要在请求路径上调用 `setPermissionMode`，改为在构建期设置一次：

```java
HarnessAgent.builder()
    .permissionContext(PermissionContextState.builder().mode(PermissionMode.BYPASS).build())
    ...
```

`loadOrCreateAgentStateForSlot` 会把它应用到新建会话并落盘，`activateSlotForContext` 每次调用都会依据存储中的 context 重建 `PermissionEngine`，因此新老会话都能得到预期的模式。

## 关联

`stateCache` 的无界驻留问题另见 #_____（提交 issue ② 后回填编号）。

birdie7761:
## AgentScope-Java Version

2.0.3-SNAPSHOT (main @ `00215511`)

## Description

Fixes #2535

`setPermissionMode` / `replacePermissionContext` read the whole `AgentState` from the never-refreshed in-process `stateCache` (`getAgentState`'s `computeIfAbsent` — a cache hit never goes back to the store) and wrote it back wholesale via `saveAgentState`. With a store configured this had two production failure modes:

- **Cross-replica stale overwrite.** Pod A caches session state v1; pod B advances the store to v2; a permission toggle on pod A then persists v1 again — the newer conversation is silently truncated. The version CAS does not save this: the mismatch falls into the default `ConflictPolicy.OVERWRITE`, which unconditionally writes the stale copy (with only a WARN log).
- **Failed-call poison promotion.** A call that failed mid-flight mutated the cached instance but never persisted it; the store version therefore still matches, so a later `setPermissionMode` CAS-saves that half-state (e.g. a trailing assistant message with `tool_use` and no `tool_result`). Every subsequent call fails with `Pending tool calls exist without results`, permanently breaking the session — the exact production incident in the issue.

This PR applies the issue's suggested direction 1 — the same authoritative reload already used by `activateSlotForContext` on every call and by the `clearContext` fix:

- Both entry points resolve the slot through `loadOrCreateAgentStateForSlot` first (replacing the `stateCache` entry and `slotVersions` slot) when a store is configured, then install and persist the permission context. Without a store the previous in-process behavior is unchanged.
- **In-flight edge** (called out in the issue): an in-flight call still holds the previously cached instance and persists it at call end, which would revert the toggle. The new permission context is mirrored onto that instance **before** the admin save — an end-of-call save racing inside the gap can no longer hit the OVERWRITE policy and swallow the change: saves landing after the mirror carry the new context, saves landing before it are superseded by the admin save. The call keeps its captured call-scoped engine, per the existing javadoc contract ("an in-flight call keeps the engine it started with").
- CAS conflicts inside the reload→save micro-window follow the existing `conflictPolicy`; the overwrite baseline is now the just-reloaded fresh state, shrinking the loss surface from a whole stale session to the micro-window delta.

### Regression tests

New `ReActAgentPermissionAdminWriteTest` uses two agents sharing one store to model a multi-replica deployment (the same pattern as the existing `clearContextReloadsPersistedStateBeforeClearingConversation`):

1. `setPermissionMode` on a stale replica must not truncate the newer persisted conversation — **fails on main** (`restored context was: []`);
2. `replacePermissionContext` likewise — **fails on main**;
3. a never-persisted failed-call tail (dangling `tool_use`) must not be promoted into the store by a permission toggle — **fails on main** (`expected: <1> but was: <2>`);
4. the same poison scenario through `replacePermissionContext` — **fails on main** (guards the shared helper against the two entry points diverging);
5. in-flight cached instances keep receiving both the mode change and a replaced permission context (guards for the mirror);
6. the mirror happens **before** the admin save — an observing store records the previously cached instance's mode at the moment the admin save fires and asserts it already carries the new mode (closes the OVERWRITE gap; **fails on the mirror-after-save ordering** with `expected: <BYPASS> but was: <DEFAULT>`);
7. without a store the toggle still mutates the cached state in place (guard).

The poison tests run against `JsonFileAgentStateStore` so the reload round-trips through serialization, matching the production MySQL/Redis stores (`InMemoryAgentStateStore` aliases instances by reference — noted in the test class javadoc).

**Compatibility**: no storage-format change — `AgentState` fields, serialization and the store API are untouched, and the reload helper is the same production code path already used by `activateSlotForContext` / `clearContext` (including the `LegacyStateLoader` fallback for older formats). Previously persisted sessions can only become *safer*: they are no longer reverted by a stale replica or poisoned by a failed call. The one deliberate behavior change — unpersisted local cache mutations are discarded in favor of the authoritative store state — is documented on the helper; regular call-path persistence goes through `saveStateToSession` and is unaffected.

### How to test

```bash
mvn -pl agentscope-core -Dtest=ReActAgentPermissionAdminWriteTest -DfailIfNoTests=false test
mvn -pl agentscope-core test                                        # 2282 tests, 0 failures
mvn -pl agentscope-core -am install -DskipTests
mvn -pl agentscope-harness test                                     # 828 tests, 0 failures
```

## Checklist

- [x]  All tests are passing (`mvn test` — core: 2282, harness: 828, 0 failures)
- [x]  Related documentation has been updated (javadoc on both entry points documents the reload semantics)


agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentPermissionAdminWriteTest.java
```diff
@@ -0,0 +1,320 @@
...
+        AgentStateStore observing =
+                new JsonFileAgentStateStore(tempDir) {
+                    @Override
+                    public void save(String u, String s, String key, State value) {
+                        AgentState inFlight = inFlightRef.get();
+                        if ("agent_state".equals(key) && inFlight != null && value != inFlight) {
+                            modeAtAdminSave.set(inFlight.getPermissionContext().getMode());
+                        }
+                        super.save(u, s, key, value);
+                    }
```
AgentScopeJavaBot: [nitpick] The `inFlightMirrorPrecedesTheAdminSave` ordering test only covers `setPermissionMode`. A symmetric test for `replacePermissionContext` would guard against future refactoring that might diverge the two entry points' mirror-before-save ordering. The current coverage is acceptable since both share the same `installPermissionContext` helper, but adding the sibling test (similar to how tests 3/4 and 5/6 are symmetric pairs) would make the suite fully airtight.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentPermissionAdminWriteTest.java
```diff
@@ -0,0 +1,320 @@
...
+                .name("user")
+                .role(MsgRole.USER)
+                .content(TextBlock.builder().text(text).build())
+                .build();
+    }
+
+    /** An assistant turn carrying a tool_use block with no matching tool_result. */
+    private static Msg danglingToolUseAssistantMsg() {
+        return Msg.builder()
+                .name("asst")
```
AgentScopeJavaBot, line 355: [nitpick] `withoutStorePermissionToggleMutatesCachedState` only tests `setPermissionMode`. A sibling test for `replacePermissionContext` without a store would complete the guard coverage for the no-store path across both public entry points.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -4226,18 +4236,76 @@ public void replacePermissionContext(
...
+     * first (mirroring the per-call reload in {@code activateSlotForContext}) and replaces the
+     * local cache entry, so the write can neither revert newer persisted state on another replica
+     * nor promote an unpersisted failed-call state into the store. Any unpersisted mutations in
+     * the previously cached local instance are discarded in favor of the authoritative store
+     * state — call-path context changes are persisted by {@code saveStateToSession} and are not
+     * affected. Without a store the locally cached state is used, as before.
+     */
+    private AgentState stateForPermissionAdminWrite(String userId, String sid) {
+        if (stateStore == null) {
+            return getAgentState(userId, sid);
```
AgentScopeJavaBot, line 4348: [nitpick] `stateForPermissionAdminWrite` unconditionally calls `loadOrCreateAgentStateForSlot` when a store is configured, while the nearby `clearContext` method first checks `stateStore.exists()` and falls back to the cached instance when the store has no record. The difference is functionally correct here (creating a state for a permission admin write is intentional, and `loadOrCreateAgentStateForSlot` creates a fresh state when nothing is found), but a brief inline comment noting the deliberate divergence from `clearContext` would help future maintainers avoid 'normalizing' the two paths.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes two real production failure modes in `setPermissionMode` / `replacePermissionContext`: (1) cross-replica stale overwrite when another pod has advanced the persisted state, and (2) promotion of unpersisted failed-call state (e.g., dangling `tool_use` without `tool_result`) into the store. The fix introduces a `stateForPermissionAdminWrite` helper that reloads authoritative state from the store before mutating — the same pattern already used by `activateSlotForContext` and `clearContext`. A `mirrorPermissionContext` guard ensures in-flight call instances also receive the permission update before the admin save, closing the OVERWRITE race window. The implementation is clean, follows established codebase patterns, and the 8-test regression suite is thorough. Javadoc is excellent. One minor asymmetry: the mirror-before-save ordering test and the no-store guard test only cover `setPermissionMode`, not `replacePermissionContext`.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentPermissionAdminWriteTest.java
AgentScopeJavaBot: [nitpick] The `inFlightMirrorPrecedesTheAdminSave` ordering test only covers `setPermissionMode`. A symmetric test for `replacePermissionContext` would guard against future refactoring that might diverge the two entry points' mirror-before-save ordering. The current coverage is acceptable since both share the same `installPermissionContext` helper, but adding the sibling test (similar to how tests 3/4 and 5/6 are symmetric pairs) would make the suite fully airtight.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentPermissionAdminWriteTest.java
AgentScopeJavaBot, line 355: [nitpick] `withoutStorePermissionToggleMutatesCachedState` only tests `setPermissionMode`. A sibling test for `replacePermissionContext` without a store would complete the guard coverage for the no-store path across both public entry points.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot, line 4348: [nitpick] `stateForPermissionAdminWrite` unconditionally calls `loadOrCreateAgentStateForSlot` when a store is configured, while the nearby `clearContext` method first checks `stateStore.exists()` and falls back to the cached instance when the store has no record. The difference is functionally correct here (creating a state for a permission admin write is intentional, and `loadOrCreateAgentStateForSlot` creates a fresh state when nothing is found), but a brief inline comment noting the deliberate divergence from `clearContext` would help future maintainers avoid 'normalizing' the two paths.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes two real production failure modes in `setPermissionMode` / `replacePermissionContext`: (1) cross-replica stale overwrite when another pod has advanced the persisted state, and (2) promotion of unpersisted failed-call state (e.g., dangling `tool_use` without `tool_result`) into the store. The fix introduces a `stateForPermissionAdminWrite` helper that reloads authoritative state from the store before mutating — the same pattern already used by `activateSlotForContext` and `clearContext`. A `mirrorPermissionContext` guard ensures in-flight call instances also receive the permission update before the admin save, closing the OVERWRITE race window. The implementation is clean, follows established codebase patterns, and the 8-test regression suite is thorough. Javadoc is excellent. One minor asymmetry: the mirror-before-save ordering test and the no-store guard test only cover `setPermissionMode`, not `replacePermissionContext`.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentPermissionAdminWriteTest.java
AgentScopeJavaBot: [nitpick] The `inFlightMirrorPrecedesTheAdminSave` ordering test only covers `setPermissionMode`. A symmetric test for `replacePermissionContext` would guard against future refactoring that might diverge the two entry points' mirror-before-save ordering. The current coverage is acceptable since both share the same `installPermissionContext` helper, but adding the sibling test (similar to how tests 3/4 and 5/6 are symmetric pairs) would make the suite fully airtight.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentPermissionAdminWriteTest.java
AgentScopeJavaBot, line 355: [nitpick] `withoutStorePermissionToggleMutatesCachedState` only tests `setPermissionMode`. A sibling test for `replacePermissionContext` without a store would complete the guard coverage for the no-store path across both public entry points.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
AgentScopeJavaBot, line 4348: [nitpick] `stateForPermissionAdminWrite` unconditionally calls `loadOrCreateAgentStateForSlot` when a store is configured, while the nearby `clearContext` method first checks `stateStore.exists()` and falls back to the cached instance when the store has no record. The difference is functionally correct here (creating a state for a permission admin write is intentional, and `loadOrCreateAgentStateForSlot` creates a fresh state when nothing is found), but a brief inline comment noting the deliberate divergence from `clearContext` would help future maintainers avoid 'normalizing' the two paths.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes two real production failure modes in `setPermissionMode` / `replacePermissionContext`: (1) cross-replica stale overwrite when another pod has advanced the persisted state, and (2) promotion of unpersisted failed-call state (e.g., dangling `tool_use` without `tool_result`) into the store. The fix introduces a `stateForPermissionAdminWrite` helper that reloads authoritative state from the store before mutating — the same pattern already used by `activateSlotForContext` and `clearContext`. A `mirrorPermissionContext` guard ensures in-flight call instances also receive the permission update before the admin save, closing the OVERWRITE race window. The implementation is clean, follows established codebase patterns, and the 8-test regression suite is thorough. Javadoc is excellent. One minor asymmetry: the mirror-before-save ordering test and the no-store guard test only cover `setPermissionMode`, not `replacePermissionContext`.

oss-maintainer (approved):
### Summary
First review. `setPermissionMode` / `replacePermissionContext` now resolve the state to mutate through `stateForPermissionAdminWrite`, which reloads the authoritative persisted state (mirroring the per-call reload in `activateSlotForContext`) before the admin write, so a toggle from one replica can no longer revert newer persisted state written by another, and an unpersisted failed-call state is no longer promoted into the store. The ordering choice in `installPermissionContext` (mirror onto the previously cached instance **before** `saveAgentState`) is the subtle part, and it is both documented and pinned by `inFlightMirrorPrecedesTheAdminSave` / `replacePermissionContextMirrorPrecedesTheAdminSave`; `withoutStore*` covers the no-store fallback. 10 new tests cover each branch of the fix rather than the happy path only. No correctness, locking or compatibility issue found. LGTM.

Not a merge decision — CI and maintainer approval still apply.

---

Issue #2583, opened 2026-08-06
[Bug]:MySQL AgentStateStore rejects Harness sandbox session IDs containing


CyberXavier:
```markdown
### Environment

- AgentScope Java: `2.0.1-sn-v2-SNAPSHOT`
- Java: 17
- Store: `MysqlAgentStateStore`
- Harness sandbox enabled

### Problem

When Harness persists sandbox state with session isolation, `SessionSandboxStateStore` generates a session ID like:

```text
sandbox/session/conv-a12d03e7-e213-4905-ad82-ad199ae1ea70
```

`MysqlAgentStateStore` then rejects it in `validateSessionId(...)`:

```text
java.lang.IllegalArgumentException:
AgentStateStore ID cannot contain path separators
```

Relevant call chain:

```text
SandboxManager.persistState(...)
 -> SessionSandboxStateStore.save(...)
 -> slotSessionId(...)
 -> MysqlAgentStateStore.save(...)
 -> validateSessionId(...)
```

### Expected behavior

Harness sandbox state should be persistable through `MysqlAgentStateStore`, or the sandbox store should encode/sanitize its internal scoped session IDs before delegating to `AgentStateStore`.

### Question

Should `SessionSandboxStateStore` use a storage-safe encoding for IDs such as `sandbox/session/<sessionId>`, or should `MysqlAgentStateStore` allow `/` in session IDs?

The database itself can store `/`; the failure comes from the application-level validation.

zouyx:
#2022 have fixed this issue, would you like to confirm?

tsujishiki:
2.0.2依旧有这个问题

haosong384:
The MySQL half of this issue was already fixed on `main` by 769eac17, which removed the path-separator check from `MysqlAgentStateStore.validateSessionId()`.

`PostgresAgentStateStore.validateSessionId()` still enforces it, so the same failure reproduces verbatim when `PostgresAgentStateStore` is used instead:

```
java.lang.IllegalArgumentException: AgentStateStore ID cannot contain path separators
    at PostgresAgentStateStore.validateSessionId(PostgresAgentStateStore.java:721)
    at PostgresAgentStateStore.save(PostgresAgentStateStore.java:265)
```

Grepping the repository confirms Postgres is now the only `AgentStateStore` implementation still enforcing this check, so the implementations are inconsistent.

[...]

PR #2336, opened 2026-07-21
fix(harness): use ':' separator in sandbox slot session id so SQL stores accept it

logicwu0:
## AgentScope-Java Version

2.0.1-SNAPSHOT (main); bug also present in 2.0.0.

## Description

Fixes #2327.

**Background.** `SessionSandboxStateStore.slotSessionId()` packs the sandbox `SandboxIsolationKey` into a single sessionId string using `/` as the segment separator (e.g. `sandbox/session/<id>`, `sandbox/user/<agentId>/<id>`, `sandbox/agent/<agentId>`, `sandbox/global`). That string is then passed as the `sessionId` argument to the backing `AgentStateStore`.

**Problem.** SQL-backed stores validate the sessionId and reject `/` and `\` as path separators — `PostgresAgentStateStore.validateSessionId` and `MysqlAgentStateStore.validateSessionId` both throw `IllegalArgumentException` in that case. As a result, sandbox state persistence fails on **every** SQL backend under **any** `IsolationScope` (SESSION/USER/AGENT/GLOBAL). The original issue reported Postgres, but MySQL is affected too: its `slotId()` uses `:` only to join `userId`/`sessionId`, so it cannot dodge the `/` that already lives inside the sandbox slot's sessionId.

**Fix.** Switch the slot-id segment separator from `/` to `:`. This is exactly the convention `MysqlAgentStateStore#slotId` already relies on (and documents) for the same `validateSessionId` reason, so it stays consistent with the codebase. The change is on the producer side (`SessionSandboxStateStore`), so it fixes all backing stores at once; no store-specific changes needed. No code reverse-parses these slot strings on `/`, so changing the separator is safe.

**How to test.** Added `slotSessionId_hasNoPathSeparators_soSqlStoresAccept_allScopes` in `SessionSandboxStateStoreTest`. It drives the store through a backend that mirrors the `validateSessionId` contract (rejects `/`/`\`) and round-trips save/load/delete for all four isolation scopes. It fails (RED) with the old `/` separators — `IllegalArgumentException: ... cannot contain path separators: sandbox/session/sess-001`, the same error seen on Postgres/MySQL — and passes (GREEN) with the fix. The pre-existing tests (which use `InMemoryAgentStateStore`, no validation) are unchanged.

`mvn -pl agentscope-harness test`: all green; `spotless:check` clean.

## Checklist

Please check the following items before code is ready to be reviewed.

- [x]  Code has been formatted with `mvn spotless:apply`
[...]

PR #2733, opened 2026-08-16
fix(postgresql): remove path-separator check from PostgresAgentStateStore session id validation

haosong384:
## AgentScope-Java Version

2.0.3-SNAPSHOT (rebased onto current `main`, at 48b62328)

### Background

`PostgresAgentStateStore.validateSessionId()` rejects any session id containing `/` or `\`:

```java
if (sessionId.contains("/") || sessionId.contains("\\")) {
    throw new IllegalArgumentException("AgentStateStore ID cannot contain path separators");
}
```

Harness packs the sandbox isolation scope into the session id, so `SessionSandboxStateStore.slotSessionId()` produces values such as `sandbox/session/<sessionId>`. Persisting sandbox state through `PostgresAgentStateStore` therefore fails with:

```
java.lang.IllegalArgumentException: AgentStateStore ID cannot contain path separators
    at PostgresAgentStateStore.validateSessionId(PostgresAgentStateStore.java:721)
    at PostgresAgentStateStore.save(PostgresAgentStateStore.java:265)
```

Call chain: `SandboxManager.persistState(...)` → `SessionSandboxStateStore.save(...)` → `slotSessionId(...)` → `PostgresAgentStateStore.save(...)` → `validateSessionId(...)`.

### Why removing the check is the right fix

This is a filesystem-oriented constraint that has no meaning for a relational store:

- The value is bound as a `PreparedStatement` parameter (`stmt.setString(1, slotId)`) and is never used as a filesystem path, so `/` poses no injection or traversal risk.
- `PostgresAgentStateStore` already joins `userId` and `sessionId` with `:` internally (`slotId(...)`), confirming that `/` carries no special meaning at this layer.
- The identical check was removed from `MysqlAgentStateStore` in 769eac17 for exactly this reason.
- `JsonFileAgentStateStore` — the implementation where the constraint genuinely matters — protects itself instead of rejecting: ids that do not match `^[a-zA-Z0-9_\-.]+$` are Base64-URL encoded.

`PostgresAgentStateStore` was the last `AgentStateStore` implementation still enforcing it (verified by grepping the whole repository), so the implementations were behaving inconsistently.

### Changes

- `PostgresAgentStateStore.validateSessionId()`: drop the path-separator rejection (null/blank and 255-character checks are unchanged).
- `PostgresAgentStateStoreTest`: the existing `rejectsSessionIdWithPathSeparators` asserted the buggy behaviour; it is replaced by `acceptsSessionIdWithPathSeparators`, which uses the real sandbox slot id (`sandbox/session/conv-1`) as input.

### How to test

```bash
mvn -pl agentscope-extensions/agentscope-extensions-postgresql -am \
    -DfailIfNoSpecifiedTests=false -Dtest='Postgres*' test
```

Before the fix, the new test fails with the exception quoted above. After the fix, the whole module passes:

```
Tests run: 144, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

The tests use a Mockito-mocked `DataSource`, so no live PostgreSQL instance is required.

### Relationship to existing issues and PRs

- Completes the Postgres half of #2583. The MySQL half was already fixed by 769eac17.
- #2336 proposes an alternative approach on the Harness side (replacing `/` with `:` in the sandbox slot id). The two are not mutually exclusive: even if #2336 lands, `PostgresAgentStateStore` would still be inconsistent with `MysqlAgentStateStore` and `JsonFileAgentStateStore`, and would still reject session ids that callers are free to choose under the `AgentStateStore` contract. Happy to defer or rebase if maintainers prefer to converge on one approach.

## Checklist

- [x]  Code has been formatted with `mvn spotless:apply` (`spotless:check` passes on all five reactor modules)
- [x]  Javadoc comments are complete and follow project conventions (no signature or contract documentation changed)
- [x]  Related documentation has been updated (e.g. links, examples, etc.) — no documentation references this validation rule; the only mention of path separators is `docs/v1/en/docs/task/session.md`, which concerns the filesystem-backed `JsonSession` and is unaffected

### DIFF
--- a/agentscope-extensions/agentscope-extensions-postgresql/src/main/java/io/agentscope/extensions/postgresql/state/PostgresAgentStateStore.java
+++ b/agentscope-extensions/agentscope-extensions-postgresql/src/main/java/io/agentscope/extensions/postgresql/state/PostgresAgentStateStore.java
@@ -717,9 +717,6 @@ public class PostgresAgentStateStore implements AgentStateStore {
         if (sessionId == null || sessionId.trim().isEmpty()) {
             throw new IllegalArgumentException("AgentStateStore ID cannot be null or empty");
         }
-        if (sessionId.contains("/") || sessionId.contains("\\")) {
-            throw new IllegalArgumentException("AgentStateStore ID cannot contain path separators");
-        }
         if (sessionId.length() > 255) {
             throw new IllegalArgumentException("AgentStateStore ID cannot exceed 255 characters");
         }
--- a/agentscope-extensions/agentscope-extensions-postgresql/src/test/java/io/agentscope/extensions/postgresql/state/PostgresAgentStateStoreTest.java
+++ b/agentscope-extensions/agentscope-extensions-postgresql/src/test/java/io/agentscope/extensions/postgresql/state/PostgresAgentStateStoreTest.java
@@ -15,6 +15,7 @@
  */
 package io.agentscope.extensions.postgresql.state;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
@@ -384,15 +385,15 @@ class PostgresAgentStateStoreTest {
     }
 
     @Test
-    void rejectsSessionIdWithPathSeparators() {
+    void acceptsSessionIdWithPathSeparators() {
         PostgresAgentStateStore store =
                 PostgresAgentStateStore.builder(dataSource).createIfNotExist(false).build();
-        assertThrows(
-                IllegalArgumentException.class,
-                () -> store.save("user", "a/b", "key", new TestState("v")));
-        assertThrows(
-                IllegalArgumentException.class,
-                () -> store.save("user", "a\\b", "key", new TestState("v")));
+        // Harness packs the sandbox isolation scope into the session id, e.g.
+        // "sandbox/session/<sessionId>". Path separators carry no meaning for a relational
+        // store, so they must not be rejected.
+        assertDoesNotThrow(
+                () -> store.save(null, "sandbox/session/conv-1", "key", new TestState("v")));
+        assertDoesNotThrow(() -> store.save("user", "a\\b", "key", new TestState("v")));
     }
 
     @Test


agentscope-extensions/agentscope-extensions-postgresql/src/main/java/io/agentscope/extensions/postgresql/state/PostgresAgentStateStore.java
```diff
@@ -717,9 +717,6 @@ protected void validateSessionId(String sessionId) {
         if (sessionId == null || sessionId.trim().isEmpty()) {
             throw new IllegalArgumentException("AgentStateStore ID cannot be null or empty");
         }
-        if (sessionId.contains("/") || sessionId.contains("\\")) {
-            throw new IllegalArgumentException("AgentStateStore ID cannot contain path separators");
-        }
         if (sessionId.length() > 255) {
```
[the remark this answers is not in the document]
haosong384: Thanks — the observation is accurate: `save(...)` computes `slotId(userId, sessionId)` and passes the composite value to `validateSessionId(...)`, so both the parameter name and the error text read as if they applied to the raw `sessionId`.

I'd prefer to keep it out of this PR, for three reasons:

1. **It is pre-existing and shared, not introduced here.** `MysqlAgentStateStore` has exactly the same shape — 8 call sites of `validateSessionId(slotId)` and the same `protected void validateSessionId(String sessionId)` signature. Its `slotId(...)` Javadoc even documents the coupling explicitly:

   ```java
   /**
    * Combine {@code (userId, sessionId)} into the single {@code session_id} column value.
    * Uses {@code :} as the separator so existing {@link #validateSessionId(String)} (which
    * forbids {@code /} and {@code \\}) still accepts the combined string.
    */
   ```

2. **Renaming only in Postgres would re-introduce the divergence this PR removes.** The whole point of this change is that `PostgresAgentStateStore` had drifted from `MysqlAgentStateStore`; swapping one inconsistency (validation behaviour) for another (helper naming) seems like a poor trade.

3. **`validateSessionId` is `protected`,** i.e. part of the extension surface for subclasses, so renaming it is a breaking change rather than a pure refactor — it warrants its own PR and release note rather than riding along with a 3-line bug fix.

Happy to open a follow-up covering both stores consistently. That change could also fix the Javadoc quoted above, which is now stale: the `/` and `\` restriction it refers to was removed from MySQL in 769eac17, and this PR removes the Postgres counterpart.

AgentScopeJavaBot (approved):
🤖 AI Review

This is a clean, minimal, and well-justified fix. The removal of the path-separator check from `PostgresAgentStateStore.validateSessionId()` correctly eliminates a filesystem-oriented constraint that has no relevance to a relational store. The session ID is always bound via `PreparedStatement.setString()`, so `/` and `\` pose no injection or traversal risk. The change brings `PostgresAgentStateStore` into alignment with `MysqlAgentStateStore` (where the same check was removed in 769eac17) and the `AgentStateStore` interface contract, which only requires sessionId to be "non-null, non-blank". The replacement test uses a realistic sandbox slot ID (`sandbox/session/conv-1`) and properly verifies both forward-slash and backslash acceptance. No issues found.

AgentScopeJavaBot (approved):
🤖 AI Review

This is a clean, minimal, and well-justified fix. The removal of the path-separator check from `PostgresAgentStateStore.validateSessionId()` correctly eliminates a filesystem-oriented constraint that has no relevance to a relational store. The session ID is always bound via `PreparedStatement.setString()`, so `/` and `\` pose no injection or traversal risk. The change brings `PostgresAgentStateStore` into alignment with `MysqlAgentStateStore` (where the same check was removed in 769eac17) and the `AgentStateStore` interface contract, which only requires sessionId to be "non-null, non-blank". The replacement test uses a realistic sandbox slot ID (`sandbox/session/conv-1`) and properly verifies both forward-slash and backslash acceptance. No issues found.

AgentScopeJavaBot (approved):
🤖 AI Review

This is a clean, minimal, and well-justified fix. The removal of the path-separator check from `PostgresAgentStateStore.validateSessionId()` correctly eliminates a filesystem-oriented constraint that has no relevance to a relational store. The session ID is always bound via `PreparedStatement.setString()`, so `/` and `\` pose no injection or traversal risk. The change brings `PostgresAgentStateStore` into alignment with `MysqlAgentStateStore` (where the same check was removed in 769eac17) and the `AgentStateStore` interface contract, which only requires sessionId to be "non-null, non-blank". The replacement test uses a realistic sandbox slot ID (`sandbox/session/conv-1`) and properly verifies both forward-slash and backslash acceptance. No issues found.

xzxiaoshan:
被移除的 sessionId 的校验，本身对 PostgreSQL 没有任何必要性，移除后不会给 PG 造成问题。兄弟模块redis、jdbc也没有这个校验，如果兄弟模块本身运行正常且sesionId包含 /，切换使用 pg 后就会出错。

--- 

LGTM.

agentscope-extensions/agentscope-extensions-postgresql/src/main/java/io/agentscope/extensions/postgresql/state/PostgresAgentStateStore.java
oss-maintainer, line 720: The diagnosis here is correct — `SessionSandboxStateStore.slotSessionId()` really does emit `sandbox/session/<id>`, `sandbox/user/<agentId>/<id>` and `sandbox/agent/<agentId>` (see `agentscope-harness/.../sandbox/SessionSandboxStateStore.java:84-92`), and since `validateSessionId()` is called on the *composed* `slotId` (line 270) rather than a path fragment, separators are meaningless for a relational store. Dropping the check is the right fix.

But the same guard is still live in the sibling JDBC store:

```java
// agentscope-extensions-jdbc/.../state/JdbcAgentStateStore.java:586-588
if (slotId.contains("/") || slotId.contains("\\")) {
    throw new IllegalArgumentException("Session ID cannot contain path separators");
}
```

So after this PR, persisting sandbox state works on MySQL (no separator check) and PostgreSQL, but `sandbox/session/...` keep failing on generic JDBC — exactly the failure mode this PR removes. `MysqlAgentStateStore.validateSessionId()` (line 991) already has no such check, which reinforces that the JDBC one is the outlier rather than a deliberate contract.

Could this PR also relax `JdbcAgentStateStore.validateSlotId()`, or is the JDBC store intentionally kept strict? If it is out of scope, it would be worth an explicit follow-up issue reference so the gap is not lost.

agentscope-extensions/agentscope-extensions-postgresql/src/test/java/io/agentscope/extensions/postgresql/state/PostgresAgentStateStoreTest.java
```diff
@@ -384,15 +385,15 @@ void validateSessionIdRejectsNullOrEmpty() {
...
-                IllegalArgumentException.class,
-                () -> store.save("user", "a/b", "key", new TestState("v")));
-        assertThrows(
-                IllegalArgumentException.class,
-                () -> store.save("user", "a\\b", "key", new TestState("v")));
+        // Harness packs the sandbox isolation scope into the session id, e.g.
+        // "sandbox/session/<sessionId>". Path separators carry no meaning for a relational
+        // store, so they must not be rejected.
+        assertDoesNotThrow(
+                () -> store.save(null, "sandbox/session/conv-1", "key", new TestState("v")));
```
oss-maintainer, line 395: Nice that the rewritten assertion uses `null` as the userId — that mirrors how `SessionSandboxStateStore` actually calls the store (sandbox state is agent-scoped, `userId` is always `null`), so the test now exercises the real production shape rather than an invented one.

Two small coverage gaps worth considering while this test is being touched:

1. Both cases go through `save(...)`, so the read paths (`get`/`getList`/`exists`, which also route through `slotId(...)` + `validateSessionId(...)`) are not pinned. A regression reintroducing a separator check on a read-only helper would not be caught.
2. The prefix shapes that carry the isolation scope — `sandbox/user/<agentId>/<id>` and `sandbox/global` — are not represented. A parameterized test over the four `slotSessionId()` outputs would make the intent ("these ids are the contract, not user input") self-documenting.

Neither blocks the change; just flagging that the test name now asserts a behaviour other stores may rely on.

oss-maintainer (commented):
### Summary
The diagnosis is right and the fix is correct: `SessionSandboxStateStore.slotSessionId()` really does produce `sandbox/session/<id>`-style ids, and since validation runs on the composed `slotId` for a relational table, rejecting path separators there was over-broad. Tests were updated in lockstep rather than just deleted, which is the right call.

Two things keep this from a clean approve:
- **[Warning]** The identical check is still enforced in `JdbcAgentStateStore.validateSlotId()`, so the same sandbox session ids will keep failing on JDBC after this merge — MySQL already has no such check, making JDBC the outlier. Worth relaxing here or referencing a follow-up.
- **[Info]** Coverage is limited to `save(...)` on PostgreSQL; read paths and the other isolation-scope prefixes are unpinned.

Existing discussion (`@xzxiaoshan`) already agrees the check is unnecessary for PG, so this is about scope/completeness rather than correctness.

---

oss-maintainer:

```bash
git fetch origin
git checkout fix/postgres-state-store-path-separator
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #2737, opened 2026-08-17
fix(harness): reclaim .skills-cache orphans when repositories return no skills

haosong384:
## AgentScope-Java Version

2.0.3-SNAPSHOT (`main` at 49cffeb6)

### Background

Orphan reclamation for `.skills-cache` lives entirely inside `MarketplaceStager.stage()`:

```java
garbageCollectOrphans(cacheRoot, retained);   // MarketplaceStager.java:146
```

But `HarnessSkillMiddleware` returned early without ever reaching `stage()` whenever no skill was left to stage:

- `prestageMarketplaceSkills()` — returned on `merged.isEmpty()`, and guarded `stage()` behind `if (!enabled.isEmpty())`
- `onSystemPrompt()` — returned on `merged.isEmpty()` and again on `enabled.isEmpty()`

So once a repository stopped publishing a skill — for example after the rows were deleted from `MysqlSkillRepository` — the directories staged by earlier calls were never reclaimed.

Because `.skills-cache` is part of `SandboxFilesystemSpec.DEFAULT_WORKSPACE_PROJECTION_ROOTS`, those stale `SKILL.md` files and scripts kept being hydrated into every freshly started sandbox. The model could still browse them with its file/shell tools and kept reporting skills that no longer existed anywhere in the repository. Nothing threw, so the divergence was silent.

### Fix

The empty case now still calls `stage()` with an empty visible set, via a small `reclaimStagedSkills()` helper.

This needs no new mechanism, because it is precisely what `stage()` already documents about itself:

> The white-list of staged directories is rebuilt every call; any pre-existing directory under `.skills-cache/<source-ns>/` not in the white-list is removed.

With an empty white-list, every staged directory is an orphan and gets reclaimed. Staging itself is a no-op in that case, and `garbageCollectOrphans()` already returns immediately when the cache directory does not exist, so the added call is cheap and safe when there is nothing to clean.

### Changes

- `HarnessSkillMiddleware`: the three early-return paths (`prestageMarketplaceSkills()` on empty merged view, `onSystemPrompt()` on empty merged and empty enabled views) now reclaim staged orphans before returning. `prestageMarketplaceSkills()` also drops the `!enabled.isEmpty()` guard so an empty enabled view reaches `stage()` as well.
- `HarnessSkillMiddlewareCacheGcTest` (new): regression coverage for both entry points, driven by a stub repository that returns an empty collection.

### How to test

```bash
mvn -pl agentscope-harness -am -DfailIfNoSpecifiedTests=false \
    -Dtest='HarnessSkillMiddlewareCacheGcTest' test
```

Both new tests fail on `main` exactly as reported in the issue — the staged directory survives the call:

```
AssertionFailedError: stale staged skill must be reclaimed:
  .../.skills-cache/mysql/demo-skill ==> expected: <false> but was: <true>
Tests run: 2, Failures: 2
```

After the fix, the full `agentscope-harness` module passes:

```
Tests run: 830, Failures: 0, Errors: 0, Skipped: 3
BUILD SUCCESS
```

(The 3 skips are pre-existing and unrelated.) The tests use `@TempDir` only — no database, sandbox or network required.

## Checklist

- [x]  Code has been formatted with `mvn spotless:apply` (`spotless:check` passes for parent, core and harness)
- [x]  Javadoc comments are complete and follow project conventions (the new private helper documents why the empty case must still reach `stage()`)
- [x]  Related documentation has been updated (e.g. links, examples, etc.) — none needed; no documented behaviour or public API changes, the fix restores the reclamation contract already stated in `MarketplaceStager.stage()`'s Javadoc

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/HarnessSkillMiddleware.java
@@ -208,12 +208,25 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
         }
         Map<String, RepoBound> merged = skillsForCall(ctx);
         if (merged.isEmpty()) {
+            reclaimStagedSkills();
             return;
         }
         List<RepoBound> visible = applyVisibility(merged.values(), ctx);
         List<RepoBound> enabled = applySkillFilter(visible, effectiveFilter(ctx));
-        if (!enabled.isEmpty()) {
-            stager.stage(enabled, sourceNamespaces);
+        stager.stage(enabled, sourceNamespaces);
+    }
+
+    /**
+     * Rebuilds the staged white-list from an empty visible set so that orphaned
+     * {@code .skills-cache} directories are reclaimed even when no skill is staged.
+     *
+     * <p>Orphan GC only runs inside {@link MarketplaceStager#stage}, so the empty case must
+     * still reach it: otherwise skills deleted from their repository keep their staged files on
+     * the host workspace and stay visible through sandbox workspace projection.
+     */
+    private void reclaimStagedSkills() {
+        if (stager != null) {
+            stager.stage(List.of(), sourceNamespaces);
         }
     }
 
@@ -227,6 +240,7 @@ public class HarnessSkillMiddleware implements HarnessRuntimeMiddleware {
 
         Map<String, RepoBound> merged = skillsForCall(ctx);
         if (merged.isEmpty()) {
+            reclaimStagedSkills();
             runtime.install(SkillCatalog.empty(), ctx, agentToolkit);
             return Mono.just(currentPrompt);
         }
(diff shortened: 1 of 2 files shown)


haosong384:
Closing this as a duplicate — #2665 by @weixsun predates it by six days and is the better fix.

I missed the existing PR during triage: my search relied on GitHub's `-linked:pr` filter, which only counts pull requests that declare a closing keyword, so #2664 still looked unclaimed. My mistake.

For the record, #2665 also handles a case this PR got wrong. It threads a `loadFailed` flag out of `mergeRepositories()` and skips reclamation via `stageEmptyRetainedIfSafe(...)` when a repository lookup threw. That distinction matters: an empty `getAllSkills()` can mean "the skills were deleted" *or* "the backing store was briefly unreachable", and this PR treated both the same — so a transient database error would have wiped a user's staged skills. #2665 reclaims only when the empty view is trustworthy, and covers four call sites rather than three.

No changes requested on #2665 from my side; it supersedes this work cleanly.

PR #2746, opened 2026-08-17
fix(harness): resolve sandbox workspace root from state instead of hardcoding /workspace

### LINKED ISSUE #2647: System prompt  Sanbox root directory is hardcoded /workspace
Image

**this should use blow config (or add sandbox worksapce dir ):**

Image

haosong384:
## AgentScope-Java Version

2.0.3-SNAPSHOT (`main` at 49cffeb6)

### Background

`WorkspaceContextMiddleware.buildWorkspaceParagraph()` hardcodes `"/workspace"` in the system prompt for sandbox-backed filesystems:

```java
sb.append("Sandbox root: /workspace (container id: ")
```

But different sandbox backends mount the workspace at different paths:
- Docker: `/workspace` (default)
- E2b: `/home/user`
- AgentRun: `/home/agentscope/workspace`
- Daytona: `/home/daytona/workspace`

The authoritative workspace root lives in each `SandboxState` subclass — written by `SandboxClient.create()` from `SandboxClientOptions.workspaceRoot` at sandbox creation time, and used by all sandbox exec/hydrate/persist operations at runtime. The hardcoded value gives the model incorrect path context when any non-default root is configured.

**Note on `WorkspaceSpec.root`**: this is a *separate* declarative field for workspace projection (initial entries, environment). It may diverge from the actual container path because it comes from a different input source. The fix reads the state field, not the spec, because the state is what the sandbox runtime actually uses.

### Changes

- **`SandboxState.java`**: add `public String getWorkspaceRoot()` returning `null`. All 5 existing backend subclasses (`DockerSandboxState`, `E2bSandboxState`, `AgentRunSandboxState`, `DaytonaSandboxState`, `KubernetesSandboxState`) already have their own `public String getWorkspaceRoot()` with their own field — those automatically become overrides, requiring zero changes to extension modules.
- **`WorkspaceContextMiddleware.java`**: new `resolveSandboxRoot(AbstractFilesystem)` helper reads `SandboxAware.getSandbox().getState().getWorkspaceRoot()`, falling back to `/workspace` when unavailable. The sandbox prompt branch calls it instead of the hardcoded string.
- **`WorkspaceContextMiddlewareSandboxRootTest.java`** (new): 3 tests with a `FakeState` subclass that overrides `getWorkspaceRoot()` — matching exactly how real backends store the value.

### How to test

```bash
mvn -pl agentscope-harness -am -DfailIfNoSpecifiedTests=false \
    -Dtest='WorkspaceContextMiddlewareSandboxRootTest' test
```

Full regression including sandbox extension serde tests:
```bash
mvn -pl agentscope-harness,agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-agentrun,agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b,agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes -am test
# Tests run: 3156, Failures: 0, Errors: 0
# BUILD SUCCESS
```

### Jackson serialization safety

The base-class getter has **no `@JsonIgnore`** — that was tested and found to *break* subclass serialization (Jackson inherits ignore annotations downward). Without the annotation, Jackson correctly resolves the property from each subclass's own field+getter+setter, ignoring the base-class getter entirely. Verified by running `AgentRunSandboxStateSerializationTest` (round-trip serde with explicit `workspaceRoot` assertion).

---
Closes #2647

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxState.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxState.java
@@ -82,4 +82,18 @@ public abstract class SandboxState {
     public void setWorkspaceRootReady(boolean workspaceRootReady) {
         this.workspaceRootReady = workspaceRootReady;
     }
+
+    /**
+     * Returns the absolute path of the workspace root inside the sandbox container.
+     *
+     * <p>Each backend stores this value in its concrete state subclass. This base-class accessor
+     * provides a polymorphic read path for code that only holds a {@code SandboxState} reference
+     * (e.g. middleware prompt builders). Subclasses that define a {@code workspaceRoot} field
+     * override this method with their own getter.
+     *
+     * @return the workspace root path, or {@code null} when the backend has not yet initialised it
+     */
+    public String getWorkspaceRoot() {
+        return null;
+    }
 }
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddlewareSandboxRootTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddlewareSandboxRootTest.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent.middleware;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.harness.agent.filesystem.sandbox.SandboxBackedFilesystem;
+import io.agentscope.harness.agent.sandbox.ExecResult;
+import io.agentscope.harness.agent.sandbox.Sandbox;
+import io.agentscope.harness.agent.sandbox.SandboxState;
+import io.agentscope.harness.agent.sandbox.WorkspaceSpec;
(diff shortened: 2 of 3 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxState.java
```diff
@@ -82,4 +82,18 @@ public boolean isWorkspaceRootReady() {
...
+     * Returns the absolute path of the workspace root inside the sandbox container.
+     *
+     * <p>Each backend stores this value in its concrete state subclass. This base-class accessor
+     * provides a polymorphic read path for code that only holds a {@code SandboxState} reference
+     * (e.g. middleware prompt builders). Subclasses that define a {@code workspaceRoot} field
+     * override this method with their own getter.
+     *
+     * @return the workspace root path, or {@code null} when the backend has not yet initialised it
+     */
+    public String getWorkspaceRoot() {
```
AgentScopeJavaBot, line 96: [nitpick] Now that `getWorkspaceRoot()` is declared on the base class, the five concrete subclasses implicitly override it but lack `@Override`. Adding `@Override` would let the compiler catch future signature drift.

agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddlewareSandboxRootTest.java
```diff
@@ -0,0 +1,168 @@
...
+        wm = new WorkspaceManager(workspace, fs);
+        WorkspaceContextMiddleware mw = new WorkspaceContextMiddleware(wm);
+
+        String prompt = mw.onSystemPrompt(null, RuntimeContext.empty(), "BASE\n").block();
+
+        assertNotNull(prompt);
+        assertTrue(
+                prompt.contains("Sandbox root: /workspace"),
+                "prompt must fall back to /workspace, got:\n" + prompt);
+    }
```
AgentScopeJavaBot, line 95: [nitpick] The `resolveSandboxRoot` method has two additional fallback branches not exercised by tests: (1) `aware.getSandbox()` returning `null`, and (2) `fs` being an `AbstractSandboxFilesystem` that does not implement `SandboxAware`. Adding one test for the null-sandbox case would close the coverage gap cheaply.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a hardcoded `/workspace` string in `WorkspaceContextMiddleware.buildWorkspaceParagraph()` by dynamically resolving the sandbox workspace root from `SandboxState.getWorkspaceRoot()`. A new base-class accessor is added to `SandboxState` (returning `null` by default), and a private `resolveSandboxRoot()` helper reads it via the `SandboxAware` interface with a clean null-safe chain, falling back to `/workspace`. The approach is sound.

PR #2755, opened 2026-08-18
fix(core): retry empty final responses instead of finishing silently

### LINKED ISSUE #2750: [Bug]: ReActAgent treats thinking-only final response as finished - user receives no text message, silent failure
**Summary**

When a reasoning model occasionally emits its entire final answer in the `reasoning_content` channel with empty `content` (`finish_reason=stop`, no tool calls), `ReActAgent` judges the run as finished. The final message contains only a `ThinkingBlock` and no `TextBlock`, so AG-UI emits zero `TEXT_MESSAGE_*` events and the persisted assistant turn has no text. No error, no warning - a silent empty reply.

Observed in production with kimi-k2.6 (DashScope OpenAI-compatible endpoint); intermittent - the same session's previous identical round produced normal text.

**Root Cause** (verified on main @ 49cffeb6, 2.0.3-SNAPSHOT)

| Link | Location | Behavior |
|---|---|---|
| Response parsing | `OpenAIResponseParser.java:431-466` | `reasoning_content` -> `ThinkingBlock`; empty `content` -> no `TextBlock` (faithful parsing, correct) |
| Event emission | `ReActAgent.java:2550-2562` (`emitBlockEvents`) | `TextBlock` -> `TextBlock*Event`; nothing emitted for text when absent |
| Finish check | `ReActAgent.java:3682-3693` (`isFinished`) | `return toolCalls.isEmpty();` - never checks that the final response contains user-visible content |
| AG-UI conversion | `TextBlockEventConverter.java:33-40` | `TextMessage*` events sourced exclusively from `TextBlock*` events |

A final assistant message consisting solely of a `ThinkingBlock` therefore passes `isFinished()` and terminates the ReAct loop with an empty visible reply.

**Impact**

- User gets no answer and no error (run finishes "successfully"); nothing at WARN/ERROR level to alert on
- Assistant turn persisted without text (reasoning only), degrading subsequent multi-turn context
- Affects any reasoning model on the OpenAI-compatible path using `reasoning_content` (kimi-k2 confirmed; DeepSeek-R1-class models share the same risk)
- Intermittent and model-dependent - hard to detect from logs

**Production Evidence**

One run: 6 model calls, 5 tool calls all successful; final generation produced 170 output tokens entirely in reasoning parts, `finish_reason=stop`; trace output parts were `[{"type":"reasoning", ...}]` with no text part. Full AG-UI event stream for the run: `TEXT_MESSAGE*` count = 0 (thinking stream normal), then `RunFinished`. Framework-level WARN/ERROR in the window: 0 - excluding retries/degradation/HTTP errors; purely a response-structure issue.

**Proposed Fix** (either or both)

1. **Built-in empty-final-response retry**: in the postReasoning pipeline, when the final response has no `ToolUseBlock` and no `TextBlock` (thinking-only), trigger the existing `gotoReasoning` mechanism with a nudge message (e.g. "your previous reply had empty content; write the final answer in the content channel"), with a retry cap. The machinery already exists (`ReActAgent.java:2416-2422`) - it is just not used for this case.
2. **Stricter finish check**: `isFinished()` additionally requires a non-empty user-visible block. Riskier as a hard rule - may break legitimate non-text finals (structured output etc.), so direction 1 is safer.

At minimum, a WARN log on "final response has no visible content" would make this detectable.

**Workaround** (application layer)

A PostReasoning middleware performing the check + `gotoReasoning` described in direction 1 - works on 2.0.1 without framework changes.

**Environment**

- AgentScope Java: 2.0.1 (production); verified still present in main 2.0.3-SNAPSHOT @ 49cffeb6
- Model: kimi-k2.6 via DashScope OpenAI-compatible endpoint

Buktal:
Fixes #2750

## Summary

A reasoning model occasionally emits its entire final answer into the `reasoning_content` channel with an empty `content` (`finish_reason=stop`, no tool calls). `ReActAgent.isFinished()` only checked `toolCalls.isEmpty()`, so such a response terminated the ReAct loop **silently**: zero `TEXT_MESSAGE_*` events downstream, empty assistant turn in session history, and nothing logged at WARN/ERROR — an invisible failure for both the user and ops. Verified in production with kimi-k2.6 (DashScope OpenAI-compatible endpoint); any reasoning model over the OpenAI-compatible path using `reasoning_content` is exposed (DeepSeek-R1-class included).

## Root cause chain (main @ 49cffeb6)

| Link | Location | Behavior |
|---|---|---|
| Response parsing | `OpenAIResponseParser.java:431-466` | `reasoning_content` → `ThinkingBlock`; empty `content` → no `TextBlock` (faithful, unchanged) |
| Event emission | `ReActAgent.emitBlockEvents` | no `TextBlock` → no `TextBlock*Event` |
| Finish check | `ReActAgent.isFinished()` | `toolCalls.isEmpty()` only — never checked for visible content |
| AG-UI conversion | `TextBlockEventConverter` | `TextMessage*` events sourced exclusively from `TextBlock*` events |

## The fix

1. **`isFinished()` strengthened** — a tool-free response finishes only when it carries visible content: any non-`ThinkingBlock` block counts, a `TextBlock` only when non-blank. `msg == null` still finishes (unchanged).
2. **Empty final responses loop back through the existing natural path** — `runPostReasoningPipeline` → `acting(iter)` → (empty pending branch) → `executeIteration(iter + 1)` → `reasoning(iter + 1, false)`. The retry therefore **consumes iteration budget and is bounded by `maxIters`** — exactly the semantics the loop already applies to hallucinated tool calls ("Tool not found" loops). No new config, no new per-call state, honest iteration accounting.
3. **Feedback per loop-back** — a WARN is logged unconditionally (observability even when iterations exhaust), and a synthetic reminder (`role=USER`, `name="system"`, `METADATA_SYNTHETIC`, reminder kind `"empty_response"`) is injected, mirroring the "Tool not found" feedback given for hallucinated tools. Construction follows the existing `TaskReminderMiddleware` / `SubagentsMiddleware` synthetic-reminder convention; unlike the transient todo reminder it is deliberately written into the agent context (like `SubagentsMiddleware`'s task-delivery reminder) so the corrective signal stays visible to later turns, bounded by `maxIters` per call.

## Behavior change

Models that legitimately end with empty or thinking-only final responses now trigger additional model calls (bounded by `maxIters`) instead of finishing silently. When iterations exhaust, the existing `summarizing()` safety net applies. The only in-repo test relying on silent thinking-only completion (`ReActAgentThinkingCumulativeTest`, a chunk-mechanics test whose mock model streams thinking-only chunks) was adapted with `.maxIters(1)`.

## Tests

New `ReActAgentEmptyResponseRetryTest`:
- thinking-only first response → retry produces text; 2 model calls; second call's last message is the synthetic reminder (asserts `METADATA_SYNTHETIC`, reminder kind `"empty_response"`, `<system-reminder>` prefix)
- a model that always returns thinking-only with `maxIters=1` → bounded (1 reasoning call + 1 summary call), never an unbounded loop
- blank-text-only response triggers the retry
- tool-call responses do **not** trigger the guard
- normal text responses do **not** trigger a retry (1 call)

Verification: `agentscope-core` (2280 tests) and `agentscope-harness` (828 tests) full suites green; Spotless clean.

## Follow-up (out of scope here)

The synthetic-reminder construction shape now exists in three places (`ReActAgent`, `TaskReminderMiddleware`, `SubagentsMiddleware`); a shared factory (e.g. `Msg.syntheticReminder(kind, text)`) would be a worthwhile standalone cleanup PR.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -2435,10 +2435,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                 // Empty final response: no tool calls and no visible content
                                 // (e.g. a reasoning model that wrote its whole answer into the
                                 // reasoning channel and left the content channel empty). Loop
-                                // back to reasoning with a synthetic reminder — mirroring the
-                                // "Tool not found" feedback given for hallucinated tool calls —
-                                // bounded by maxIters, instead of silently finishing with an
-                                // empty reply (#2750).
+                                // back to reasoning with a synthetic reminder.
                                 if (!hasToolCalls(eventMsg)) {
                                     log.warn(
                                             "Final response has no visible content (empty reply),"
@@ -3698,8 +3695,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
         /**
          * Check if the ReAct loop should terminate.
          *
-         * <p>A response with tool calls (even non-existent ones) continues to the acting phase
-         * where ToolExecutor returns "Tool not found" for the model to see. A tool-free response
+         * <p>A response with tool calls continues to the acting phase. A tool-free response
          * finishes only when it carries visible content: empty or thinking-only responses (the
          * entire answer in the reasoning channel) loop back to reasoning, bounded by {@code
          * maxIters}, instead of silently ending the agent with an empty reply.
@@ -3716,28 +3712,14 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                 return false;
             }
 
-            return hasVisibleText(msg);
-        }
-
-        /**
-         * Check whether the given message requests tool execution. Shared by {@link #isFinished}
-         * and the empty-final-response guard so both derive "tool-free" from one place.
-         */
-        private static boolean hasToolCalls(Msg msg) {
-            return !msg.getContentBlocks(ToolUseBlock.class).isEmpty();
-        }
-
-        /**
-         * Check whether the given message carries user-visible content: a {@link TextBlock}
-         * with non-blank text. Text is the only channel consumers render as the assistant's
-         * visible reply (thinking never counts; tool calls are handled by the caller before
-         * this check; images and other data enter through tool results, not final responses).
-         */
-        private static boolean hasVisibleText(Msg msg) {
             return msg.getContentBlocks(TextBlock.class).stream()
                     .anyMatch(
                             textBlock ->
                                     textBlock.getText() != null && !textBlock.getText().isBlank());
+        }
+
+        private static boolean hasToolCalls(Msg msg) {
+            return !msg.getContentBlocks(ToolUseBlock.class).isEmpty();
         }
 
         /**
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -3716,7 +3716,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                 return false;
             }
 
-            return hasVisibleContent(msg);
+            return hasVisibleText(msg);
         }
 
         /**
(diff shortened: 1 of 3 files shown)


Buktal:
LGTM @jujn

AgentScopeJavaBot (commented):
🤖 AI Review

This PR adds an "empty final response guard" to `ReActAgent` (issue #2750). When a model returns a response with no tool calls and no visible text content (e.g., thinking-only where the entire answer was written to the reasoning channel), instead of silently finishing with an empty reply, the agent loops back to reasoning with a synthetic `<system-reminder>` message. The retry is bounded by `maxIters`. The implementation is clean, well-documented, and includes comprehensive test coverage. The core logic is correct: `isFinished` now returns `false` for empty responses, the guard adds a reminder, and `acting` routes back to `reasoning` when no tool calls are present.

jujn (approved):

PR #2756, opened 2026-08-18
fix(middleware): normalize model call tools

### LINKED ISSUE #2753: [Bug]: Max-iteration summary passes null tools through the onModelCall middleware chain
**Describe the bug**

When a `ReActAgent` reaches `maxIters`, the summary path creates a `ModelCallInput` with `tools = null` and sends it through the regular `MiddlewareBase.onModelCall` chain.

However, the `ModelCallInput` API only describes `tools` as "the tool schemas" and does not document that the list may be null. A middleware that treats it as a normal collection, for example by calling `input.tools().stream()`, fails with a `NullPointerException`. The max-iteration summary never reaches the model.

The relevant code is currently:

`ReActAgent.CallExecution.summaryStream()`:

```java
.apply(new ModelCallInput(messages, null, options, model))
```

This behavior exists in v2.0.0-RC5 and is still present on the current `main` branch.

**To Reproduce**

1. Register an `onModelCall` middleware that processes the tool schema collection:

```java
@Override
public Flux<AgentEvent> onModelCall(
        Agent agent,
        RuntimeContext ctx,
        ModelCallInput input,
    input.tools().stream().forEach(schema -> inspect(schema));
    return next.apply(input);
}
```

2. Configure a `ReActAgent` with a low `maxIters` value and a model response that keeps the reasoning/acting loop running until the limit is reached.
3. Execute the agent.
4. When `summarizing()` invokes `summaryStream()`, the middleware receives `tools = null` and throws before the summary model call.

**Expected behavior**

A model call with no available tools should expose an empty, non-null collection to middleware. For example, the summary path could use `List.of()`, or `ModelCallInput` could normalize a null tools value in its constructor.

If null is intentionally part of the public contract, that nullability should be explicit in the API and middleware documentation. A non-null empty list is preferable because middleware chains commonly transform or inspect tool schemas as collections.

**Error messages**

```text
java.lang.NullPointerException: Cannot invoke "java.util.List.stream()"
because the return value of
"io.agentscope.core.middleware.ModelCallInput.tools()" is null
    at <custom middleware>.onModelCall(...)
    at io.agentscope.core.middleware.MiddlewareChain.lambda$build$0(...)
    at io.agentscope.core.ReActAgent$CallExecution.summaryStream(...)
    at io.agentscope.core.ReActAgent$CallExecution.summarizing(...)
```

**Environment**

- AgentScope-Java Version: 2.0.0-RC5
- Java Version: 21
- OS: Linux

**Additional context**

The normal reasoning path passes the list returned by `Toolkit.getToolSchemas(...)`, while only the max-iteration summary path deliberately passes null. This makes the generic `onModelCall` middleware contract path-dependent.

A focused regression test could register a middleware that asserts `input.tools() != null`, force `maxIters` to be reached, and verify that the summary model call completes.

DennisWLX:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Summary model calls do not expose tools, but they currently pass `null` to middleware through `ModelCallInput`. Middleware that treats `tools()` as a collection can therefore fail only on the summary path.

This change:

- passes an empty tool list from `ReActAgent.summaryStream()`
- normalizes legacy/null inputs at the `ModelCallInput` boundary
- documents the non-null tools contract
- adds regression coverage for summary middleware and direct construction

Fixes #2753

### Testing

- `mvn -pl agentscope-core clean test`: 2277 tests passed, 9 skipped
- focused regression tests: 14 tests passed

## Checklist

- [x] Related documentation has been updated (Javadoc contract; no external docs required)

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -3534,7 +3534,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                             rc,
                             MiddlewareBase::onModelCall,
                             summaryModelCallCore)
-                    .apply(new ModelCallInput(messages, null, options, model))
+                    .apply(new ModelCallInput(messages, List.of(), options, model))
                     .doOnNext(this::publishEvent);
         }
 
--- a/agentscope-core/src/test/java/io/agentscope/core/middleware/ModelCallInputTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/middleware/ModelCallInputTest.java
@@ -15,24 +15,19 @@
  */
 package io.agentscope.core.middleware;
 
-import io.agentscope.core.message.Msg;
-import io.agentscope.core.model.GenerateOptions;
-import io.agentscope.core.model.Model;
-import io.agentscope.core.model.ToolSchema;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 import java.util.List;
+import org.junit.jupiter.api.Test;
 
-/**
- * Input context for {@link MiddlewareBase#onModelCall}.
- *
- * @param messages the messages to send to the model
- * @param tools    the non-null tool schemas, empty when no tools are available
- * @param options  generation options
- * @param model    the model instance to call
- */
-public record ModelCallInput(
-        List<Msg> messages, List<ToolSchema> tools, GenerateOptions options, Model model) {
+class ModelCallInputTest {
 
-    public ModelCallInput {
-        tools = tools == null ? List.of() : tools;
+    @Test
+    void nullToolsAreNormalizedToEmptyList() {
+        ModelCallInput input = new ModelCallInput(List.of(), null, null, null);
+
+        assertNotNull(input.tools());
+        assertTrue(input.tools().isEmpty());
     }
 }
--- a/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentNewLoopReplyTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentNewLoopReplyTest.java
@@ -46,6 +46,8 @@ import io.agentscope.core.message.ThinkingBlock;
 import io.agentscope.core.message.ToolResultBlock;
 import io.agentscope.core.message.ToolResultState;
 import io.agentscope.core.message.ToolUseBlock;
+import io.agentscope.core.middleware.MiddlewareBase;
+import io.agentscope.core.middleware.ModelCallInput;
 import io.agentscope.core.model.ChatModelBase;
 import io.agentscope.core.model.ChatResponse;
 import io.agentscope.core.model.GenerateOptions;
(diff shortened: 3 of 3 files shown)


agentscope-core/src/main/java/io/agentscope/core/middleware/ModelCallInput.java
```diff
@@ -25,9 +25,14 @@
...
- * @param tools    the tool schemas
+ * @param tools    the non-null tool schemas, empty when no tools are available
  * @param options  generation options
  * @param model    the model instance to call
  */
 public record ModelCallInput(
-        List<Msg> messages, List<ToolSchema> tools, GenerateOptions options, Model model) {}
+        List<Msg> messages, List<ToolSchema> tools, GenerateOptions options, Model model) {
+
+    public ModelCallInput {
```
AgentScopeJavaBot, line 35: [nitpick] The compact constructor normalizes `tools` but not `messages`. While `messages` is always non-null at current call sites, consider applying the same defensive normalization (`messages = messages == null ? List.of() : messages;`) for API consistency — records expose their components directly, so any null leaks through to middleware consumers. Not blocking.

agentscope-core/src/test/java/io/agentscope/core/middleware/ModelCallInputTest.java
```diff
@@ -0,0 +1,33 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class ModelCallInputTest {
+
+    @Test
+    void nullToolsAreNormalizedToEmptyList() {
+        ModelCallInput input = new ModelCallInput(List.of(), null, null, null);
```
AgentScopeJavaBot, line 28: [nitpick] Only the null-tools path is tested. Consider adding a complementary test that passes a non-null, non-empty tool list and asserts it is preserved unchanged (identity passthrough). This guards against a future regression where the compact constructor accidentally wraps or copies non-null inputs.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -3534,7 +3534,7 @@ Flux<AgentEvent> summaryStream(
                             rc,
                             MiddlewareBase::onModelCall,
                             summaryModelCallCore)
-                    .apply(new ModelCallInput(messages, null, options, model))
+                    .apply(new ModelCallInput(messages, List.of(), options, model))
```
AgentScopeJavaBot, line 3537: [nitpick] Symmetry note: `OtelTracingMiddleware.java:187` still includes `input.tools() != null ? (long) input.tools().size() : 0L` — this null branch is now dead code since the compact constructor guarantees non-null tools. Consider a follow-up cleanup to simplify it to `(long) input.tools().size()` for consistency. Not blocking for this PR.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR hardens `ModelCallInput` against null `tools` by adding a compact constructor that normalizes `null` to `List.of()`, and updates the one call site in `ReActAgent.summaryStream()` that previously passed `null` explicitly. The change is clean, focused, and well-tested at both unit and integration levels. The belt-and-suspenders approach (fixing the call site **and** adding constructor-level normalization) is good defensive practice. One minor cleanup opportunity exists in downstream code that still checks for null tools.

zouyx (approved):

PR #2757, opened 2026-08-18
fix(core): preserve failed summary termination state

### LINKED ISSUE #2754: [Feature]: Preserve max-iteration terminal semantics when summary generation fails

Yes. When a `ReActAgent` reaches `maxIters`, `summarizing()` emits an `ExceedMaxItersEvent` and attempts one final summary model call.

If that summary succeeds, the returned message is tagged with `GenerateReason.MAX_ITERATIONS`. If the summary fails, however, `handleSummaryError()` catches the exception and returns a plain `AssistantMessage` containing text such as:

```text
Maximum iterations (100) reached. Error generating summary: ...
```

That fallback message is not tagged with `GenerateReason.MAX_ITERATIONS` and carries no structured summary-failure information. The agent then completes normally, so adapters commonly emit an ordinary agent result followed by agent end.

Consumers that determine run status from the final result can therefore classify an exhausted and unsuccessfully summarized run as successful. The only reliable signal is the earlier event, which requires every adapter and orchestrator to retain cross-event state.

Please preserve max-iteration terminal semantics even when summary generation fails.

At minimum:

1. Tag the fallback message returned by `handleSummaryError()` with `GenerateReason.MAX_ITERATIONS`.
2. Preserve a structured indication that summary generation failed, for example through message metadata or a dedicated event that includes the failure category without exposing sensitive provider details.
3. Document that reaching `maxIters` is a terminal partial/failure outcome even though the agent returns a graceful final message.

A possible shape is:

```java
AssistantMessage.builder()
        .name(getName())
        .content(...)
        .generateReason(GenerateReason.MAX_ITERATIONS)
        .metadata(Map.of("summaryStatus", "failed"))
        .build();
```

The exact metadata/event API is open for discussion; the key requirement is that callers should not need to parse English text or correlate a previous event to identify the terminal outcome.

It would also be useful to add tests for both branches:

- max iterations + successful summary -> final message has `MAX_ITERATIONS`;
- max iterations + failed summary -> fallback message still has `MAX_ITERATIONS` and exposes structured summary failure.

- **Only consume `ExceedMaxItersEvent`:** works, but forces adapters to maintain state across events and is easy to miss when only the final agent result is persisted.
- **Parse the fallback text:** brittle, locale-dependent, and couples consumers to an implementation string.
- **Propagate the summary exception instead of returning a message:** preserves failure semantics but removes the useful graceful fallback response. A structured fallback result provides both.
- **Treat any returned `AssistantMessage` as success:** this is the current behavior in many integrations and is the source of the ambiguity.

**Additional context**

In `ReActAgent.CallExecution.summarizing()`, the successful post-summary branch explicitly calls `withGenerateReason(GenerateReason.MAX_ITERATIONS)`. The `handleSummaryError()` branch builds a new `AssistantMessage` without the equivalent reason.

This request is independent of the specific cause of a summary failure. The model provider, a middleware, a hook, or stream processing can all fail during the summary call; terminal semantics should remain consistent for all of them.

Observed with AgentScope-Java 2.0.0-RC5; the same behavior is present on the current `main` branch.

DennisWLX:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

When the ReAct loop reaches `maxIters` and final summary generation also fails, the fallback message currently uses the default `MODEL_STOP` reason. Downstream callers cannot distinguish this failure from a normal model completion.

This change:

- preserves `GenerateReason.MAX_ITERATIONS` on the summary fallback message
- adds internal `_summary_failed=true` metadata for explicit failure detection
- adds a regression test for the returned reason, metadata, and error text

Fixes #2754

### Testing

- `mvn -pl agentscope-core clean test`: 2276 tests passed, 9 skipped

## Checklist

- [x] Related documentation has been updated (metadata Javadoc; no external docs required)


agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -3648,6 +3648,8 @@ private Mono<Msg> handleSummaryError(Throwable error) {
                                                                     + " generating summary: %s",
                                                             maxIters, error.getMessage()))
                                             .build())
+                            .metadata(Map.of(MessageMetadataKeys.SUMMARY_FAILED, true))
```
AgentScopeJavaBot, line 3703: [nitpick] The `.metadata(Map.of(...))` call creates an unmodifiable map. The subsequent `.generateReason()` call correctly wraps it into a `HashMap` (via the `instanceof HashMap` check in `Msg.Builder.generateReason()`), so the current code works correctly. However, this creates an implicit ordering dependency: if a future refactoring moves `.metadata()` after `.generateReason()`, the `metadata()` call would silently overwrite the generate-reason entry. Consider adding a brief comment like `// .metadata() must precede .generateReason()` or using `new HashMap<>(Map.of(...))` directly to make the intent explicit and prevent accidental breakage.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR adds metadata markers to the summary-error fallback path in `ReActAgent.handleSummaryError()`. Specifically, it sets `SUMMARY_FAILED=true` in the message metadata and explicitly assigns `GenerateReason.MAX_ITERATIONS` to the error fallback message. This brings the error path into symmetry with the successful summary path (which already sets `MAX_ITERATIONS` via `withGenerateReason`), while the new `SUMMARY_FAILED` flag allows callers to distinguish a failed summary from a successful one. The change is minimal, well-targeted, and includes a focused unit test that validates all three assertions (generate reason, metadata flag, and error message text). No bugs or correctness issues found.

zouyx (changes requested):
有个冲突要解决下

zouyx (approved):

PR #2760, opened 2026-08-18
fix(core): propagate agent state load failures

### LINKED ISSUE #2741: [Bug]: ReActAgent silently resets conversation memory when AgentStateStore throws (and drops the stack trace)
**Describe the bug**

In `ReActAgent.loadOrCreateAgentStateForSlot` (agentscope-core, `ReActAgent.java:441-448`), the `catch (Exception e)` around state loading falls back to a brand-new empty `AgentState` for **any** exception thrown by the configured `AgentStateStore` — not just "state not found". A transient backend failure (Redis/MySQL blip, connection timeout, serialization error after an upgrade, etc.) is therefore indistinguishable from a genuinely fresh session: the agent continues this turn with empty conversation history, and on the next successful save it can overwrite the previously persisted state.

The same catch also logs `e.getMessage()` instead of `e`, so the stack trace is dropped. For exceptions whose `getMessage()` is null or terse (NPE, some Jackson `JsonMappingException` paths, timeout wrappers), the log line gives no place to start from.

Related but distinct: #2702 (onError path does not persist), #2475 (sessionId with `/` corrupts the slot), #2535 (stale stateCache write-back overwrites newer state). This one is the read-side complement.

**To Reproduce**

1. Configure a `ReActAgent` / `HarnessAgent` with any `AgentStateStore` whose `getVersioned(...)` throws — e.g. a stub that simulates a transient `SQLException`/`RedisConnectionFailureException`:

   ```java
   AgentStateStore brokenStore = new AgentStateStore() {
       @Override
       public <T> Optional<VersionedState<T>> getVersioned(
               String userId, String sessionId, String key, Class<T> type) {
           throw new RuntimeException("simulated transient store failure");
       }
       // ... other methods delegate to a working in-memory store
   };
   ```

2. Put a few turns into a session so history exists in the working backend.
3. Make the backend throw once on the next `agent.call(...)` (flip a flag, kill the Redis proxy, etc.).
4. Send another message.

**Observed behavior**

- The turn proceeds as if the session were brand new; no prior messages are visible to the model.
- The only log line is:
  ```
  WARN  ReActAgent - Failed to load AgentState for slot (userId=alice, sessionId=demo): simulated transient store failure
  ```
  No stack trace.
- On the next successful save, the empty in-memory state can overwrite the previously persisted history.

**Expected behavior**

- A transient backend failure should propagate (or the turn should fail fast) rather than silently treating the session as new — the cost of a failed turn is much lower than the cost of silently wiping memory.
- The "fall back to a fresh state" path should be reserved for the genuine absence case, which the code already distinguishes (`versioned.isPresent()` / `legacy.found()` both return false without throwing).
- The log should include the exception itself (`log.warn("...", e)`), not just `e.getMessage()`.

**Error messages**

```
WARN  i.a.core.ReActAgent - Failed to load AgentState for slot (userId=alice, sessionId=demo): simulated transient store failure
```

- AgentScope-Java Version: `2.0.3-SNAPSHOT` (main, commit `49cffeb6`)
- Java Version: 17
- OS: macOS 15.3 (darwin)

**Additional context**

The offending block is `ReActAgent.java:441-448`:

```java
} catch (Exception e) {
    log.warn(
            "Failed to load AgentState for slot (userId={}, sessionId={}): {}",
            userId,
            sessionId,
            e.getMessage());
    long version = stateStore.supportsVersioning() ? 0L : AgentStateStore.UNVERSIONED;
    return new VersionedState<>(fresh, version);
}
```

Happy to send a PR once we agree on the desired semantics — my one-line hypothesis is that the catch should (a) log `e` for the stack trace and (b) only return `fresh` when the exception is a known "not found" signal from the store, letting storage/serialization errors bubble.

ordinary-s:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Summary

Prevent `ReActAgent` from silently replacing conversation state with a fresh session when `AgentStateStore` loading fails.

## Root Cause

`loadOrCreateAgentStateForSlot` caught every `Exception` raised by versioned or legacy state loading and returned a fresh `AgentState`. This made backend, connection, I/O, and deserialization failures indistinguishable from a successful not-found result.

The fresh state could then enter the per-slot cache and be saved with a zero or unversioned baseline, potentially overwriting valid persisted conversation history.

## Changes

- Remove the broad state-load exception handler.
- Let the original load or decoding exception propagate to the caller.
- Continue creating fresh state only after both versioned and legacy loads successfully report that no state exists.
- Add regression coverage for versioned store failures, non-versioned legacy failures, genuinely missing state, existing state restoration, decoding failures, and cache/persisted-state integrity.
- Update a harness test fixture so its mocked `AgentStateStore` delegates to a contract-compliant in-memory implementation.

## Tests

- `mvn -pl agentscope-core -Dtest=ReActAgentStateLoadFailureTest -DfailIfNoTests=false test`
  - 5 tests passed
- `mvn -pl agentscope-harness test`
  - 830 tests run, 0 failures, 0 errors, 3 skipped
- `mvn spotless:check`
  - all 88 reactor modules passed
  - 2,280 tests ran with 0 failures and 2 errors; both errors are pre-existing Windows symlink tests that cannot create symbolic links under the current user's permissions. The unrelated `GracefulShutdownTest` also passed 49/49 when rerun independently after one shared-static-state flaky failure in an exclusion run.

## Risk

The intentional behavior change is that a turn now fails when persisted state cannot be restored. Valid missing-state, existing-state, versioned, non-versioned, and legacy paths remain unchanged. No public API or storage format changes are introduced.

Fixes #2741

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -422,30 +422,20 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
         if (stateStore == null) {
             return new VersionedState<>(fresh, AgentStateStore.UNVERSIONED);
         }
-        try {
-            VersionedState<AgentState> versioned =
-                    stateStore.getVersioned(userId, sessionId, "agent_state", AgentState.class);
-            if (versioned.isPresent()) {
-                return versioned;
-            }
-            LegacyStateLoader.LegacyLoadResult legacy =
-                    LegacyStateLoader.loadFromLegacySessionWithPresence(
-                            stateStore, userId, sessionId);
-            if (legacy.found()) {
-                // Legacy keys have no version; treat as create-if-absent baseline.
-                long version = stateStore.supportsVersioning() ? 0L : AgentStateStore.UNVERSIONED;
-                return new VersionedState<>(legacy.state(), version);
-            }
-            long version = stateStore.supportsVersioning() ? 0L : AgentStateStore.UNVERSIONED;
-            return new VersionedState<>(fresh, version);
-        } catch (Exception e) {
-            log.warn(
-                    "Failed to load AgentState for slot (userId={}, sessionId={})",
-                    userId,
-                    sessionId,
-                    e);
-            throw e;
+        VersionedState<AgentState> versioned =
+                stateStore.getVersioned(userId, sessionId, "agent_state", AgentState.class);
+        if (versioned.isPresent()) {
+            return versioned;
         }
+        LegacyStateLoader.LegacyLoadResult legacy =
+                LegacyStateLoader.loadFromLegacySessionWithPresence(stateStore, userId, sessionId);
+        if (legacy.found()) {
+            // Legacy keys have no version; treat as create-if-absent baseline.
+            long version = stateStore.supportsVersioning() ? 0L : AgentStateStore.UNVERSIONED;
+            return new VersionedState<>(legacy.state(), version);
+        }
+        long version = stateStore.supportsVersioning() ? 0L : AgentStateStore.UNVERSIONED;
+        return new VersionedState<>(fresh, version);
     }
 
     private static AgentState freshState(
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -422,31 +422,20 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
         if (stateStore == null) {
             return new VersionedState<>(fresh, AgentStateStore.UNVERSIONED);
         }
-        try {
-            VersionedState<AgentState> versioned =
-                    stateStore.getVersioned(userId, sessionId, "agent_state", AgentState.class);
-            if (versioned.isPresent()) {
-                return versioned;
-            }
-            LegacyStateLoader.LegacyLoadResult legacy =
-                    LegacyStateLoader.loadFromLegacySessionWithPresence(
-                            stateStore, userId, sessionId);
-            if (legacy.found()) {
-                // Legacy keys have no version; treat as create-if-absent baseline.
-                long version = stateStore.supportsVersioning() ? 0L : AgentStateStore.UNVERSIONED;
-                return new VersionedState<>(legacy.state(), version);
-            }
-            long version = stateStore.supportsVersioning() ? 0L : AgentStateStore.UNVERSIONED;
-            return new VersionedState<>(fresh, version);
-        } catch (Exception e) {
-            log.warn(
-                    "Failed to load AgentState for slot (userId={}, sessionId={}): {}",
-                    userId,
-                    sessionId,
-                    e.getMessage());
-            long version = stateStore.supportsVersioning() ? 0L : AgentStateStore.UNVERSIONED;
-            return new VersionedState<>(fresh, version);
+        VersionedState<AgentState> versioned =
+                stateStore.getVersioned(userId, sessionId, "agent_state", AgentState.class);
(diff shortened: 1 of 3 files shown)


agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -422,31 +422,20 @@ private static VersionedState<AgentState> loadOrCreateAgentStateForSlot(
...
-            }
-            long version = stateStore.supportsVersioning() ? 0L : AgentStateStore.UNVERSIONED;
-            return new VersionedState<>(fresh, version);
-        } catch (Exception e) {
-            log.warn(
-                    "Failed to load AgentState for slot (userId={}, sessionId={}): {}",
-                    userId,
-                    sessionId,
-                    e.getMessage());
+        VersionedState<AgentState> versioned =
```
AgentScopeJavaBot, line 432: [nitpick] The removed `catch` block included `log.warn("Failed to load AgentState for slot (userId={}, sessionId={}): {}", userId, sessionId, e.getMessage())`, which provided structured session context for diagnostics. With the catch removed, exceptions propagate (correct), but the userId/sessionId context is no longer attached to the log entry. Consider wrapping the body in a try-catch that logs context and rethrows: `try { ... } catch (Exception e) { log.warn("Failed to load AgentState for slot (userId={}, sessionId={})", userId, sessionId, e); throw e; }`. This preserves observability without swallowing the failure. The `AgentBase.createErrorHandler` → `notifyError` hook does propagate the error, but the `ErrorEvent` does not carry userId/sessionId, making production debugging of intermittent store failures harder when many sessions are active.

AgentScopeJavaBot (approved):
🤖 AI Review

This PR fixes a data-loss bug in `ReActAgent.loadOrCreateAgentStateForSlot` where a broad `try-catch (Exception)` silently swallowed state store failures (network errors, deserialization failures, etc.) and returned a fresh empty state. This could cause the agent to "forget" persisted conversation history and even overwrite it on the next save. The fix removes the catch block so exceptions propagate to callers, causing the agent call to fail instead of silently losing state. The change is correct, well-scoped, and backed by comprehensive regression tests covering versioned store failures, legacy store failures, decoding failures, and happy-path scenarios. The harness test fix (replacing bare mocks with delegate-backed mocks) is a necessary adaptation. One minor suggestion: preserve the structured userId/sessionId logging that was removed with the catch block.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
ordinary-s: Addressed in c793f725. `loadOrCreateAgentStateForSlot` now logs `userId` and `sessionId` with the original exception as the final SLF4J argument, then rethrows that exact exception. This preserves the stack trace and fail-fast behavior without changing missing-state, legacy restoration, or versioning semantics. The focused regression tests and all GitHub CI checks pass.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -440,12 +440,11 @@ private static VersionedState<AgentState> loadOrCreateAgentStateForSlot(
             return new VersionedState<>(fresh, version);
         } catch (Exception e) {
             log.warn(
-                    "Failed to load AgentState for slot (userId={}, sessionId={}): {}",
+                    "Failed to load AgentState for slot (userId={}, sessionId={})",
```
hexly666: 建议： 重新抛出异常的情况下，在这里不打印日志，调用方可能会 catch 或者做异常处理的时候重新打印 stack trace，导致二次打印 stack trace

jujn (approved):

jujn:
@hexly666 不知道为啥Assign不了你（可能github bug），评论记录一下

PR #2762, opened 2026-08-18
fix(sandbox): make concurrent sandbox uploads safe (unique hydrate temp names, native transfer for relative paths)

### LINKED ISSUE #2761: [Bug]: Concurrent sandbox file uploads race on the deterministic hydrate temp tar — sessions.json / transcript uploads fail intermittently with "tar extract failed (exit=2)"
**Describe the bug**

`SandboxBackedFilesystem.uploadFiles` handles uploads that `supportsFileTransfer` rejects by building a single-entry tar and pushing it through `Sandbox.hydrateWorkspace`. On the Kubernetes backend that hydrate stages a temp archive whose name is derived from the sandbox session id alone (`KubernetesSandbox.java:177`):

```java
String tmpRel = TMP_DIR_REL + "/ws-hydrate-" + sessionHash() + ".tar";
```

`sessionHash()` depends only on the sandbox session id, and the upload → extract → cleanup path is not synchronized. Two concurrent uploads on the same sandbox therefore share one tar path, and this interleaving fails one of them:

```
writer A:  files().write(tar)   OK
writer B:  files().write(tar)   OK      (overwrites the same name)
writer A:  tar -xf              OK
writer A:  rm -f tar            OK      (finally cleanup)
writer B:  tar -xf              FAILS   -> "Cannot open: No such file or directory"
```

The losing writer's exception is caught in `uploadFiles` (`SandboxBackedFilesystem.java:142-145`) and downgraded to `FileUploadResponse.fail` + WARN, so the file silently never lands in the sandbox.

The window is hit in practice by transcript persistence: at every agent-call teardown, the transcript segment `.jsonl` (mirrored asynchronously by the static single-thread `MIRROR_EXECUTOR`, `SessionTree.java:90`) and the `sessions.json` session index (written immediately after `flush()` returns, `WorkspaceManager.updateSessionIndex`, `WorkspaceManager.java:404`) both call `uploadFiles` on the same sandbox session.

These writes take the hydrate fallback at all because `uploadFiles` probes `supportsFileTransfer` with the raw path (`SandboxBackedFilesystem.java:110-111`), but `WorkspaceManager.writeUtf8WorkspaceRelative` passes workspace-relative paths, which never match `KubernetesSandbox.toFileApiRelative` — it only accepts absolute paths under `fileApiBaseDir` (`KubernetesSandbox.java:291`). So each single-file workspace write rides the racy temp tar instead of the native `uploadFile` channel.

**To Reproduce**

1. Configure an agent with a sandbox filesystem backed by `KubernetesSandbox` (file API enabled) and transcript persistence on the workspace filesystem.
2. Run a multi-turn conversation — each call teardown runs the segment mirror and the session-index write in the same window.
3. Observe intermittent WARNs (frequency depends on thread scheduling):

```text
WARN  ... SandboxBackedFilesystem - [sandbox-fs] uploadFiles failed for path: agents/<agentId>/sessions/sessions.json
io.agentscope.harness.agent.sandbox.SandboxException$SandboxRuntimeException: tar extract failed (exit=2): tar: /workspace/.agentscope-tmp/ws-hydrate-<hash>.tar: Cannot open: No such file or directory
    at io.agentscope.extensions.sandbox.kubernetes.KubernetesSandbox.extractArchive(KubernetesSandbox.java:214)
    ...
```

4. The file is absent under `/workspace` in the sandbox (and missing from the persisted snapshot) even though the run itself succeeded.

**Expected behavior**

Concurrent uploads to the same sandbox must not interfere with each other — every successful `uploadFiles` response should correspond to a file that actually landed.

**Error messages**

See the WARN block above.

- AgentScope-Java Version: main @ 2.0.3-SNAPSHOT (since #2557, merged 2026-08-06; no tagged release carries it yet)
- Java Version: 21
- OS: Windows 11 client against Kubernetes sandbox pods (same deterministic name is also used by the exec fallback when the file API is disabled: `/tmp/ws-hydrate-<hash>.tar`, `KubernetesSandbox.java:189`)

**Additional context**

Fix proposal in #2762, two parts: resolve workspace-relative paths against the workspace root so ordinary writes go through `SandboxFileTransfer.uploadFile` instead of the hydrate fallback, and give each staged archive a unique name so concurrent hydrate calls cannot collide.

Buktal:
## AgentScope-Java Version

main @ 2.0.3-SNAPSHOT

## Description

Fixes #2761.

Two commits:

1. `fix(sandbox-kubernetes)`: `hydrateViaFileApi` / `hydrateViaExec` / `persistViaFileApi` staged their temp tar at a name derived from the sandbox session id alone. Concurrent transfers on one sandbox shared that path, so one side's `rm -f` cleanup deleted the other side's archive between its upload and `tar -xf` (`Cannot open: No such file or directory`). The name now appends a UUID per transfer.

2. `fix(harness)`: `uploadFiles` probed `supportsFileTransfer` with the raw path. Workspace-relative paths (session index, transcript segments written via `WorkspaceManager`) never matched the absolute-path contract and always took the archive-hydrate fallback where that race lives. Relative paths are now resolved against the sandbox workspace root before probing and uploaded through `SandboxFileTransfer.uploadFile`. Absolute-path behavior and the hydrate fallback for sandboxes/paths without transfer support are unchanged.

**Tests**

- `KubernetesSandboxTest`: two hydrate calls on the same sandbox write distinct temp archive paths.
- `SandboxBackedFilesystemTest`: a workspace-relative path resolves against the workspace root and rides the native transfer (no hydrate); an unavailable workspace root falls back to hydration; paths outside the file-API root still hydrate.

## Checklist

### DIFF
--- a/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/test/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxTest.java
+++ b/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/test/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxTest.java
@@ -107,6 +107,24 @@ class KubernetesSandboxTest {
     }
 
     @Test
+    void hydrateUsesUniqueTempArchivePerCall() throws Exception {
+        when(commands.run(anyString())).thenReturn(new ExecutionResult("", "", 0));
+
+        sandbox.doHydrateWorkspace(new ByteArrayInputStream("tar-one".getBytes()));
+        sandbox.doHydrateWorkspace(new ByteArrayInputStream("tar-two".getBytes()));
+
+        // Concurrent transfers must never share a temp archive: a deterministic name lets
+        // one call's rm -f cleanup delete another call's archive before tar -xf runs.
+        ArgumentCaptor<String> path = ArgumentCaptor.forClass(String.class);
+        verify(files, org.mockito.Mockito.times(2)).write(path.capture(), any(byte[].class));
+        String first = path.getAllValues().get(0);
+        String second = path.getAllValues().get(1);
+        assertTrue(first.startsWith(".agentscope-tmp/ws-hydrate-"));
+        assertTrue(second.startsWith(".agentscope-tmp/ws-hydrate-"));
+        assertTrue(!first.equals(second));
+    }
+
+    @Test
     void persistUsesFileApiWhenConfigured() throws Exception {
         when(commands.run(anyString())).thenReturn(new ExecutionResult("", "", 0));
         byte[] tar = "fake-tar".getBytes();
--- a/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/test/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxTest.java
+++ b/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/test/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxTest.java
@@ -125,6 +125,25 @@ class KubernetesSandboxTest {
     }
 
     @Test
+    void persistUsesUniqueTempArchivePerCall() throws Exception {
+        when(commands.run(anyString())).thenReturn(new ExecutionResult("", "", 0));
+        when(files.read(startsWith(".agentscope-tmp/ws-persist-")))
+                .thenReturn("fake-tar".getBytes());
+
+        sandbox.doPersistWorkspace();
+        sandbox.doPersistWorkspace();
+
+        // Same uniqueness contract as hydrate: concurrent persists must not share a
+        // temp archive, even if a future refactor splits the naming method.
+        ArgumentCaptor<String> path = ArgumentCaptor.forClass(String.class);
+        verify(files, org.mockito.Mockito.times(2)).read(path.capture());
+        String first = path.getAllValues().get(0);
+        String second = path.getAllValues().get(1);
+        assertTrue(first.startsWith(".agentscope-tmp/ws-persist-"));
+        assertTrue(!first.equals(second));
+    }
+
+    @Test
     void persistUsesFileApiWhenConfigured() throws Exception {
         when(commands.run(anyString())).thenReturn(new ExecutionResult("", "", 0));
         byte[] tar = "fake-tar".getBytes();
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
@@ -107,16 +107,30 @@ public class SandboxBackedFilesystem extends BaseSandboxFilesystem implements Sa
             String path = file.getKey();
             byte[] content = file.getValue();
 
-            if (active instanceof SandboxFileTransfer transfer
-                    && transfer.supportsFileTransfer(path)) {
+            if (active instanceof SandboxFileTransfer transfer) {
+                String transferPath = null;
                 try {
-                    transfer.uploadFile(path, content);
-                    results.add(FileUploadResponse.success(path));
-                } catch (Exception e) {
-                    log.warn("[sandbox-fs] native upload failed for path: {}", path, e);
+                    transferPath = resolveTransferPath(active, path);
+                } catch (IllegalArgumentException e) {
+                    // Same contract as the archive fallback: an invalid path fails this file.
+                    log.warn("[sandbox-fs] uploadFiles failed for path: {}", path, e);
                     results.add(FileUploadResponse.fail(path, e.getMessage()));
+                    continue;
+                } catch (IOException e) {
+                    log.debug(
+                            "[sandbox-fs] Workspace root unavailable, keeping archive fallback: {}",
+                            path);
                 }
-                continue;
+                if (transferPath != null && transfer.supportsFileTransfer(transferPath)) {
+                    try {
+                        transfer.uploadFile(transferPath, content);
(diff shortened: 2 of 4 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
```diff
@@ -107,16 +107,30 @@ public List<FileUploadResponse> uploadFiles(
...
+                    continue;
+                } catch (IOException e) {
+                    log.debug(
+                            "[sandbox-fs] Workspace root unavailable, keeping archive fallback: {}",
+                            path);
+                }
+                if (transferPath != null && transfer.supportsFileTransfer(transferPath)) {
+                    try {
+                        transfer.uploadFile(transferPath, content);
+                        results.add(FileUploadResponse.success(path));
```
AgentScopeJavaBot, line 131: [major] `content` is not validated for null before `transfer.uploadFile(transferPath, content)`. The archive fallback has an explicit guard in `buildSingleFileArchive` (`"File content must not be null"`), but the native transfer path passes null directly to the sandbox implementation, which will produce an unclear NPE or vendor-specific error. With relative paths now eligible for native transfer, this gap is more exposed. Add a null guard before the native transfer attempt, e.g.:
```java
if (content == null) {
    results.add(FileUploadResponse.fail(path, "File content must not be null"));
    continue;
}
```
Place it right after `byte[] content = file.getValue();` (line 109) so both paths benefit.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/test/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxTest.java
```diff
@@ -106,6 +106,24 @@ void hydrateUsesExecWhenFileApiDisabled() throws Exception {
         verify(commands).run(contains("echo '" + b64 + "' | base64 -d > /tmp/ws-hydrate-"));
     }
 
+    @Test
```
AgentScopeJavaBot, line 109: [minor] `hydrateUsesUniqueTempArchivePerCall` covers the hydrate path, but `persistViaFileApi` also calls `tempArchiveName("ws-persist")` and has no corresponding uniqueness test. Since both code paths share the same method, the risk is low, but a symmetric test (e.g. `persistUsesUniqueTempArchivePerCall` invoking `doPersistWorkspace()` twice and asserting the two `files.read()` capture paths differ) would close the coverage gap and guard against future refactors that might split the method.

AgentScopeJavaBot (commented):
🤖 AI Review

This PR fixes a genuine concurrency race condition in `KubernetesSandbox` where concurrent hydrate/persist calls shared a deterministic temp archive name (derived from `sessionHash()`), allowing one call's `rm -f` cleanup to delete another call's archive before its `tar -xf` extraction completes. The fix introduces `tempArchiveName()` which appends `UUID.randomUUID()` to guarantee per-call uniqueness — correct and minimal. The second change enables native single-file transfer for relative paths in `SandboxBackedFilesystem` by resolving them against the workspace root, avoiding the slower tar-hydration round-trip. The path resolution logic is clean, security is preserved (`validatePath` rejects `..`, `toFileApiRelative` double-checks segments), and test coverage is solid with three new test cases covering the happy path, fallback when root is unavailable, and unsupported-path hydration. One defensive-programming gap is worth addressing: null `content` bypasses validation on the native transfer path.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
Buktal: Addressed in 06577573: `uploadFiles` now fails the file up front with `File content must not be null`, right after reading the entry, so both the native transfer and archive paths share the same contract. Added `uploadFiles_rejectsNullContentOnNativeTransferPath`.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/test/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxTest.java
Buktal: Addressed in 73e400a6: added `persistUsesUniqueTempArchivePerCall`, invoking `doPersistWorkspace()` twice and asserting the two captured `files.read()` paths differ.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
AgentScopeJavaBot: ✅ Verified as addressed in `9277d19`: Null guard added at line 111-113 right after `byte[] content = file.getValue();`, exactly as the bot suggested: fails the file with "File content must not be null" and continues, protecting both the native transfer and archive fallback paths.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/test/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxTest.java
AgentScopeJavaBot: ✅ Verified as addressed in `9277d19`: `persistUsesUniqueTempArchivePerCall` test added (lines 126-142). It calls `doPersistWorkspace()` twice, captures both paths via ArgumentCaptor, asserts the first starts with `.agentscope-tmp/ws-persist-`, and asserts the two captured paths differ — symmetric to the existing hydrate uniqueness test as the bot requested.

AgentScopeJavaBot (approved):

jujn (approved):

PR #2764, opened 2026-08-18
feat(harness): add disableLocalWorkspace() to opt out of local workspace materialisation

### LINKED ISSUE #2763: Harness always materialises .agentscope/workspace under the working directory — no opt-out for SaaS / sandboxed deployments
# Issue draft — post on https://github.com/agentscope-ai/agentscope-java/issues

---
**Title:** Harness always materialises `.agentscope/workspace` under the working directory — no opt-out for SaaS / sandboxed deployments

**Body:**

### Problem

`HarnessAgent` always resolves a local workspace. When `workspace(...)` is not set it falls back
to `${user.dir}/.agentscope/workspace` (see `resolveDefaultWorkspace()`), and several
workspace-backed components are wired unconditionally — the default local filesystem overlay
(`LocalFilesystemSpec`), the workspace-backed `WorkspaceMessageBus` / `WorkspaceAsyncToolRegistry`
fallbacks, the default `WorkspaceTaskRepository` (`agents/<agentId>/tasks`), skills staging
(`.skills-cache`), and (on `main`) transcripts under `workspace/.agentscope/transcripts`.

In a SaaS / containerised deployment where **all file IO happens inside a remote sandbox or an
in-memory / distributed store**, that local workspace is pure dead weight. Even after
`disableFilesystemTools().disableShellTool()` the harness still creates a `.agentscope` tree in
the application's working directory (and the default `JsonFileAgentStateStore` falls back to
`~/.agentscope/state/<agentId>` when no state store is given). Note the project's own
`.gitignore` already ignores `.agentscope/` — the directory is known to be an unwanted
by-product, but there is no way to prevent it.

The existing escape hatches (`workspace(Path)`, `agentscope.workspace` system property,
`AGENTSCOPE_WORKSPACE` env var) only *relocate* the directory; none of the 15 `disable*`
builder switches stops the local materialisation.

### Proposed solution

Add `HarnessAgent.Builder.disableLocalWorkspace()` (feature PR #…):

- workspace resolves to an ephemeral `${java.io.tmpdir}/agentscope-workspace/<agentId>` when
  `workspace(...)` is unset, never `${user.dir}`;
- no default local filesystem is created (`resolveFilesystem` returns `null`);
- missing explicit `.stateStore(...)` / `.taskRepository(...)` fail fast with descriptive
  errors instead of persisting to `~/.agentscope/state` / `agents/<agentId>/tasks`;
- transcripts and skills staging are skipped when there is no local filesystem.

Behaviour is unchanged for everyone who does not call the new switch.

### Environment

- agentscope-java `main` (2.0.3-SNAPSHOT); also present in 2.0.1/2.0.2
- JDK 17+, Windows / Linux

Letter2025:
﻿# feat(harness): add `disableLocalWorkspace()` to opt out of local workspace materialisation

Closes #2763

## Motivation

In SaaS / containerised deployments where **all file IO happens inside a remote sandbox or an
in-memory / distributed store**, the harness still resolves and uses a local workspace at
`${user.dir}/.agentscope/workspace` by default. Even when the File/Shell tools are disabled via
`disableFilesystemTools().disableShellTool()`, components that are intrinsic to the harness
(workspace-backed default `TaskRepository`, default `JsonFileAgentStateStore`, skills staging
`.skills-cache`, transcripts, message bus) keep writing into the application's working
directory — a `.agentscope` tree appears in the project root for every deployment that never
wanted one.

The existing escape hatch (`workspace(...)` / `agentscope.workspace` / `AGENTSCOPE_WORKSPACE`)
only **relocates** the directory; it cannot turn it off, and requiring every SaaS service to
know about and point at a local directory is exactly the kind of implicit state this mode wants
to eliminate.

## What this PR adds

`HarnessAgent.Builder.disableLocalWorkspace()` — a per-build opt-out:

- When no explicit `.workspace(...)` is set, the workspace resolves to an **ephemeral
  `${java.io.tmpdir}/agentscope-workspace/<agentId>`** instead of `${user.dir}/.agentscope/workspace`.
- **No default local filesystem is created** (`resolveFilesystem` returns `null` instead of
  materialising a `LocalFilesystemSpec` overlay). This also disables the workspace-backed
  `WorkspaceMessageBus` / `WorkspaceAsyncToolRegistry` fallbacks.
- **Fail-fast validation** replaces silent persistence:
  - missing explicit `.stateStore(...)` → `IllegalStateException` (the default
    `JsonFileAgentStateStore` would otherwise persist to `~/.agentscope/state/<agentId>`);
  - subagents enabled without an explicit `.taskRepository(...)` → `IllegalStateException`
    (the default `WorkspaceTaskRepository` would otherwise materialise
    `agents/<agentId>/tasks`).
- Transcripts are **skipped with a warning** instead of being persisted to
  `workspace/.agentscope/transcripts` when no `.transcriptStore(...)` is supplied and no
  filesystem exists.
- Skills staging (`.skills-cache`) and the `FilesystemTool` registration are skipped when there
  is no local filesystem.

Backward compatible: default behaviour is unchanged; the flag is additive and only takes effect
when explicitly enabled. No existing API is broken.

## Behaviour table

| Aspect | Default | `disableLocalWorkspace()` |
|--------|---------|---------------------------|
| Workspace location (no `workspace(...)`) | `${user.dir}/.agentscope/workspace` | `${java.io.tmpdir}/agentscope-workspace/<agentId>` |
| Default filesystem | `user.dir` + workspace overlay | none (`null`) |
| Default `AgentStateStore` | `~/.agentscope/state/<agentId>` | build fails fast, requires `.stateStore(...)` |
| Default `TaskRepository` | `agents/<agentId>/tasks` | build fails fast, requires `.taskRepository(...)` |
| MessageBus / async-tool registry | workspace-backed | skipped |
| Transcripts | persisted in workspace | skipped with warning (unless `transcriptStore` given) |
| Skills staging (`.skills-cache`) | materialised | not staged |

## Tests

`HarnessAgentLocalWorkspaceDisabledTest` (5 tests):

- no `.agentscope` created under the working directory in disabled mode;
- missing explicit state store fails fast;
- subagents enabled without explicit task repository fails fast;
- ephemeral workspace resolves under `${java.io.tmpdir}`, never the working directory;
- blank agent id sanitisation.

## Documentation

- `docs/v2/en/docs/harness/workspace.md` + `docs/v2/zh/docs/harness/workspace.md`:
  new "Disabling the local workspace" section.
- `docs/v2/en/docs/others/release-notes.md`: "Next" entry.

## Checklist

- [x] Conventional Commit message
- [x] Unit tests included and passing
- [x] Docs updated (EN + ZH)
- [x] Backward compatible, no API breakage
- [ ] Spotless formatting applied (run `mvn spotless:apply`)

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
@@ -1136,9 +1136,8 @@ public class HarnessAgent implements Agent, AutoCloseable {
      * @return the ephemeral workspace directory; never {@code null}
      */
     static Path resolveEphemeralWorkspace(String agentId) {
-        String safeAgentId =
-                agentId.replaceAll("[/\\\\\\s\\p{Cntrl}]", "_").replaceAll("^[_\\-\\.]+", "");
-        if (safeAgentId.isBlank()) {
+        String safeAgentId = HarnessAgentBuilderSupport.sanitizeIdentifier(agentId);
+        if (safeAgentId == null || safeAgentId.isBlank()) {
             safeAgentId = "ReActAgent";
         }
         return Paths.get(System.getProperty("java.io.tmpdir"))
@@ -2506,9 +2505,7 @@ public class HarnessAgent implements Agent, AutoCloseable {
                                                 .resolve(".agentscope/transcripts"));
                     }
                 }
-                if (effectiveTranscriptStore == null) {
-                    // Transcripts simply not recorded in this mode (see warning above).
-                } else {
+                if (effectiveTranscriptStore != null) {
                     inner.middleware(
                             new TranscriptMiddleware(
                                     wsManager, effectiveTranscriptStore, transcriptTenant));
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
@@ -1127,6 +1127,26 @@ public class HarnessAgent implements Agent, AutoCloseable {
     }
 
     /**
+     * Resolves the ephemeral workspace used when {@link Builder#disableLocalWorkspace()} is
+     * enabled and no explicit {@link Builder#workspace(Path)} is set. The path lives under the
+     * JVM temp directory so that nothing is materialised in the application's working directory;
+     * it is defensive fallback and is not created eagerly.
+     *
+     * @param agentId the resolved agent id (already non-blank)
+     * @return the ephemeral workspace directory; never {@code null}
+     */
+    static Path resolveEphemeralWorkspace(String agentId) {
+        String safeAgentId =
+                agentId.replaceAll("[/\\\\\\s\\p{Cntrl}]", "_").replaceAll("^[_\\-\\.]+", "");
+        if (safeAgentId.isBlank()) {
+            safeAgentId = "ReActAgent";
+        }
+        return Paths.get(System.getProperty("java.io.tmpdir"))
+                .resolve("agentscope-workspace")
+                .resolve(safeAgentId);
+    }
+
+    /**
      * Returns true when the given session is a local in-process implementation that cannot share
      * state across nodes. Used by sandbox / remote-filesystem fail-fast checks to reject
      * configurations that would silently leak per-node state in distributed deployments.
@@ -1199,6 +1219,16 @@ public class HarnessAgent implements Agent, AutoCloseable {
         String transcriptTenant;
         boolean disableSessionPersistence = false;
         boolean disableWorkspaceContext = false;
+
+        /**
+         * When enabled, the harness never materialises the default local workspace
+         * ({@code ${user.dir}/.agentscope/workspace}) for this build. Intended for
+         * SaaS / containerised deployments where all file IO is handled by a remote
+         * sandbox or an in-memory / distributed store, and a local workspace directory
+         * would only pollute the application's working directory.
+         */
+        boolean disableLocalWorkspace = false;
+
         boolean disableAtPathExpansion = false;
         boolean disableSubagents = false;
         boolean disableDynamicSkills = false;
@@ -2145,6 +2175,42 @@ public class HarnessAgent implements Agent, AutoCloseable {
             return this;
         }
 
+        /**
+         * Prevents the harness from materialising the default local workspace
+         * ({@code .agentscope/workspace} under the current working directory) for this agent.
+         *
+         * <p>When enabled and no explicit {@link #workspace(Path)} is set, the workspace path
+         * resolves to an ephemeral location under the JVM temp directory
+         * ({@code ${java.io.tmpdir}/agentscope-workspace/<agentId>}) instead of
+         * {@code ${user.dir}/.agentscope/workspace}, and no default local filesystem is created.
+         * This targets SaaS / container deployments that run agents inside a remote sandbox or
+         * with an in-memory / distributed store, and never want a {@code .agentscope} directory
+         * to appear in the application's working directory.
+         *
+         * <p><strong>Required companion configuration:</strong> because the workspace-backed
+         * defaults (JsonFileAgentStateStore, WorkspaceTaskRepository, local filesystem) are not
+         * available in this mode, the build fails fast with a descriptive error unless the caller
+         * supplies the corresponding custom implementations:
+         * <ul>
+         *   <li>an explicit {@link #stateStore(io.agentscope.core.state.AgentStateStore)} (or a
+         *       {@code DistributedStore});</li>
+         *   <li>an explicit {@link #taskRepository(io.agentscope.harness.agent.subagent.task.TaskRepository)}
+         *       when subagents are enabled;</li>
+         *   <li>a set of opt-outs for workspace-local subsystems that are not desired in this
+         *       mode, e.g. {@link #disableFilesystemTools()}, {@link #disableShellTool()},
+         *       {@link #disableMemoryTools()}, {@link #disableMemoryHooks()},
+         *       {@link #disableDynamicSkills()}, {@link #disableDefaultWorkspaceSkills()} and
+         *       {@link #disableTranscript()}. Any such subsystem that is left enabled reads and
+         *       writes the ephemeral temp workspace instead of {@code ${user.dir}}.</li>
+         * </ul>
+         *
+         * @return this builder
+         */
+        public Builder disableLocalWorkspace() {
+            this.disableLocalWorkspace = true;
+            return this;
+        }
+
         public Builder disableAtPathExpansion() {
             this.disableAtPathExpansion = true;
             return this;
@@ -2241,11 +2307,16 @@ public class HarnessAgent implements Agent, AutoCloseable {
                         "abstractFilesystem() is an escape hatch and is mutually exclusive with"
                                 + " filesystem(...) specs");
             }
-            Path resolvedWorkspace = workspace != null ? workspace : resolveDefaultWorkspace();
             String resolvedAgentId =
                     agentId != null && !agentId.isBlank()
                             ? agentId
                             : (name != null && !name.isBlank() ? name : "ReActAgent");
+            Path resolvedWorkspace =
+                    workspace != null
+                            ? workspace
+                            : (disableLocalWorkspace
+                                    ? resolveEphemeralWorkspace(resolvedAgentId)
+                                    : resolveDefaultWorkspace());
             // ---- DistributedStore auto-wiring ----
             // distributedStore provides storage components; filesystem mode is user's choice.
             // Priority: explicit builder methods > distributedStore > workspace defaults
@@ -2280,6 +2351,14 @@ public class HarnessAgent implements Agent, AutoCloseable {
                             : new LocalPeriodicGate();
 
             AgentStateStore effectiveSession = stateStoreOverride;
+            if (disableLocalWorkspace && effectiveSession == null) {
+                throw new IllegalStateException(
+                        "disableLocalWorkspace() leaves no default AgentStateStore: the default"
+                            + " JsonFileAgentStateStore would persist to ~/.agentscope/state, which"
+                            + " is exactly what this mode must avoid. Pass an explicit"
+                            + " .stateStore(...) or configure a DistributedStore that supplies"
+                            + " one.");
+            }
             IsolationScope fsIsolationScope = IsolationScope.USER;
             if (remoteFilesystemSpec != null && remoteFilesystemSpec.getIsolationScope() != null) {
                 fsIsolationScope = remoteFilesystemSpec.getIsolationScope();
@@ -2412,6 +2491,13 @@ public class HarnessAgent implements Agent, AutoCloseable {
                     if (wsManager.getFilesystem() != null) {
                         effectiveTranscriptStore =
                                 new ObjectStoreTranscriptStore(wsManager.getFilesystem());
+                    } else if (disableLocalWorkspace) {
+                        // No local workspace and no filesystem: transcripts would have to be
+                        // persisted to ${user.dir}, which this mode explicitly forbids.
+                        log.warn(
+                                "[harness] disableLocalWorkspace() leaves no transcript backend;"
+                                    + " transcripts are disabled. Supply .transcriptStore(...) or"
+                                    + " omit disableLocalWorkspace() to persist session logs.");
                     } else {
                         effectiveTranscriptStore =
                                 new FilesystemTranscriptStore(
@@ -2420,9 +2506,13 @@ public class HarnessAgent implements Agent, AutoCloseable {
                                                 .resolve(".agentscope/transcripts"));
                     }
                 }
-                inner.middleware(
-                        new TranscriptMiddleware(
-                                wsManager, effectiveTranscriptStore, transcriptTenant));
+                if (effectiveTranscriptStore == null) {
+                    // Transcripts simply not recorded in this mode (see warning above).
+                } else {
+                    inner.middleware(
+                            new TranscriptMiddleware(
+                                    wsManager, effectiveTranscriptStore, transcriptTenant));
+                }
             }
             Model memoryModel = memoryConfig.model() != null ? memoryConfig.model() : model;
             if (memoryModel != null && !disableMemoryHooks) {
@@ -2472,7 +2562,9 @@ public class HarnessAgent implements Agent, AutoCloseable {
                     inner.middleware(compactionHook);
                 }
             }
-            if (!disableToolResultEviction && toolResultEvictionConfig != null) {
+            if (!disableToolResultEviction
+                    && toolResultEvictionConfig != null
+                    && filesystem != null) {
                 inner.middleware(
                         new ToolResultEvictionMiddleware(filesystem, toolResultEvictionConfig));
             }
@@ -2575,7 +2667,7 @@ public class HarnessAgent implements Agent, AutoCloseable {
                 pathNormalizer =
                         WorkspacePathNormalizer.of(resolvedWorkspace.toAbsolutePath().toString());
             }
-            if (!disableFilesystemTools) {
+            if (!disableFilesystemTools && filesystem != null) {
                 agentToolkit.registerTool(new FilesystemTool(filesystem, pathNormalizer));
             }
             if (!disableShellTool && filesystem instanceof AbstractSandboxFilesystem sandbox) {
@@ -2724,7 +2816,7 @@ public class HarnessAgent implements Agent, AutoCloseable {
 
             if (!orderedSkillRepos.isEmpty()) {
                 io.agentscope.harness.agent.skill.runtime.MarketplaceStager stager =
-                        resolvedWorkspace != null
+                        resolvedWorkspace != null && !disableLocalWorkspace
                                 ? new io.agentscope.harness.agent.skill.runtime.MarketplaceStager(
                                         resolvedWorkspace)
                                 : null;
@@ -2801,7 +2893,7 @@ public class HarnessAgent implements Agent, AutoCloseable {
                     "HarnessAgent '{}' built [workspace={}, filesystem={}, subagents={}]",
                     name,
                     resolvedWorkspace,
-                    filesystem.getClass().getSimpleName(),
+                    filesystem != null ? filesystem.getClass().getSimpleName() : "none",
                     !leafSubagent && !disableSubagents && model != null);
 
             // ---- Build inner ReActAgent ----
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
@@ -162,6 +162,14 @@ final class HarnessAgentBuilderSupport {
         if (b.localFilesystemSpec != null) {
             return b.localFilesystemSpec.toFilesystem(workspace, nsFactory);
         }
+        if (b.disableLocalWorkspace) {
+            // No default local filesystem: nothing to materialise under the working directory.
+            // Workspace-backed components (WorkspaceMessageBus, WorkspaceAsyncToolRegistry,
+            // workspace task repository, transcript-on-disk, skills staging…) are consequently
+            // skipped or require an explicit custom implementation — exactly the intent for
+            // SaaS/sandboxed builds that never want a .agentscope directory locally.
+            return null;
+        }
         // Default: route through LocalFilesystemSpec so the default project (= ${user.dir})
         // is overlaid below the agent workspace, matching the Claude-Code-style two-layer model.
         return new LocalFilesystemSpec().toFilesystem(workspace, nsFactory);
@@ -759,6 +767,13 @@ final class HarnessAgentBuilderSupport {
             if (distributed != null) {
                 return distributed;
             }
+        }
+        if (b.disableLocalWorkspace) {
+            throw new IllegalStateException(
+                    "disableLocalWorkspace() prevents the workspace-backed default TaskRepository"
+                            + " (which would materialise agents/<agentId>/tasks under the working"
+                            + " directory). Pass an explicit .taskRepository(...), or configure a"
+                            + " DistributedStore that supplies one.");
         }
         Objects.requireNonNull(
                 wsManager,
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentLocalWorkspaceDisabledTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentLocalWorkspaceDisabledTest.java
@@ -1,0 +1,240 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.core.message.Msg;
+import io.agentscope.core.message.TextBlock;
+import io.agentscope.core.model.ChatResponse;
+import io.agentscope.core.model.GenerateOptions;
+import io.agentscope.core.model.Model;
+import io.agentscope.core.model.ToolSchema;
+import io.agentscope.core.state.InMemoryAgentStateStore;
+import io.agentscope.harness.agent.subagent.task.BackgroundTask;
+import io.agentscope.harness.agent.subagent.task.TaskRepository;
+import io.agentscope.harness.agent.subagent.task.TaskRunSpec;
+import io.agentscope.harness.agent.subagent.task.TaskStatus;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import reactor.core.publisher.Flux;
+
+/**
+ * Tests for {@link HarnessAgent.Builder#disableLocalWorkspace()}: a "no local workspace" build
+ * must never materialise {@code .agentscope} under the working directory, and it must fail fast
+ * with a descriptive error when the workspace-backed defaults (state store, task repository)
+ * cannot be avoided.
+ */
+class HarnessAgentLocalWorkspaceDisabledTest {
+
+    @TempDir Path workingDirectory;
+
+    private String previousUserDir;
+    private String previousWorkspaceProperty;
+
+    @BeforeEach
+    void pointWorkingDirectoryAtTempDir() {
+        previousUserDir = System.getProperty("user.dir");
+        previousWorkspaceProperty = System.getProperty(HarnessAgent.WORKSPACE_PROPERTY);
+        // The default workspace would resolve against ${user.dir}; redirect it to a scratch dir
+        // so the test can assert that NOTHING is created there in disabled mode.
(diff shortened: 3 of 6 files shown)


AgentScopeJavaBot (commented):
🤖 AI Review

This PR adds `disableLocalWorkspace()` to `HarnessAgent.Builder`, allowing SaaS / containerised deployments to opt out of materialising the default `${user.dir}/.agentscope/workspace` directory. When enabled, the workspace path falls back to an ephemeral temp directory, no default local filesystem is created, and missing state-store / task-repository fail fast with descriptive errors. The implementation is well-structured: the fail-fast checks are correctly placed, the transcript and skill-staging subsystems degrade gracefully, test coverage is solid (including edge cases for `resolveEphemeralWorkspace`), and both English and Chinese docs are updated. Two areas for improvement are noted below.

Letter2025:
Hi @AgentScopeJavaBot — thanks for the review! Could you re-post the full report including the two areas for improvement? The inline comments didn't attach (line numbers fell outside the PR hunks), so the archived report isn't visible on our side. We'd like to address every point before the maintainers take a look.

Letter2025:
Hi @oss-maintainer — I've rebased this PR onto the latest \main\ (6c8ccc5), so it's up to date and mergeable (0 conflicts, verified via merge-tree).

Summary of the change: \HarnessAgent.Builder.disableLocalWorkspace()\ opts a build out of materialising the default \\/.agentscope/workspace\ directory — intended for SaaS / containerised deployments where all file IO happens inside a remote sandbox or an in-memory/distributed store. When enabled, the workspace resolves to an ephemeral \\\ location, no default local filesystem is created, and missing \.stateStore(...)\ / \.taskRepository(...)\ fail fast with descriptive errors. Covered by 5 new tests (all passing locally) plus docs (EN/ZH) and release notes; prior CI runs (ubuntu + windows) were all green.

Could you please take a look and approve if everything looks good? Happy to adjust anything per your feedback.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2712,7 +2801,7 @@ public HarnessAgent build() {
                 pathNormalizer =
                         WorkspacePathNormalizer.of(resolvedWorkspace.toAbsolutePath().toString());
             }
-            if (!disableFilesystemTools) {
+            if (!disableFilesystemTools && filesystem != null) {
```
oss-maintainer, line 2828: The null-filesystem guards are added per call site, and `deliver_artifact` looks like a gap: `artifactDeliveryEnabled` is only `artifactDeliveryTarget != null && !disableFilesystemTools` (HarnessAgent.java:2805-2810), so a disabled-mode build that sets an artifact target registers `ArtifactDeliveryTool` with a null filesystem, and `filesystem.downloadFiles(...)` (ArtifactDeliveryTool.java:126) NPEs at call time instead of either failing the build or omitting the tool. Suggest `&& filesystem != null` here (or fold it into `artifactDeliveryEnabled`), plus the same audit for the other ungated consumers reached via `wsManager`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1177,6 +1177,25 @@ static Path resolveDefaultWorkspace() {
...
+     * @param agentId the resolved agent id (already non-blank)
+     * @return the ephemeral workspace directory; never {@code null}
+     */
+    static Path resolveEphemeralWorkspace(String agentId) {
+        String safeAgentId = HarnessAgentBuilderSupport.sanitizeIdentifier(agentId);
+        if (safeAgentId == null || safeAgentId.isBlank()) {
+            safeAgentId = "ReActAgent";
+        }
+        return Paths.get(System.getProperty("java.io.tmpdir"))
+                .resolve("agentscope-workspace")
```
oss-maintainer, line 1200: `${java.io.tmpdir}/agentscope-workspace/<agentId>` is host-global and never reclaimed: two replicas (or two tenants) of the same agentId on one host collide in the same tree — `.skills-cache`, plan files, bus dirs — which is exactly the multi-replica SaaS layout this mode targets, and the previous `${user.dir}` location at least died with the deployment. Consider mixing a per-JVM nonce (or the `IsolationScope`/userId component already used elsewhere) into the path, and documenting that the tree is not deleted on exit.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2414,6 +2484,14 @@ public HarnessAgent build() {
                             : new LocalPeriodicGate();
 
             AgentStateStore effectiveSession = stateStoreOverride;
+            if (disableLocalWorkspace && effectiveSession == null) {
```
oss-maintainer: The message names `~/.agentscope/state`, but `defaultStateDir()` also honors `-Dagentscope.state.home=...` (HarnessAgent.java:1136-1143) — and this throw runs before that fallback, so an operator who only wants the state tree outside `$HOME` cannot use it to satisfy this mode. Either allow an explicit `agentscope.state.home` to count as "caller took responsibility" or drop the `~/.agentscope/state` wording and just say "no default AgentStateStore in this mode".

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
```diff
@@ -163,6 +163,14 @@ static AbstractFilesystem resolveFilesystem(
         if (b.localFilesystemSpec != null) {
             return b.localFilesystemSpec.toFilesystem(workspace, nsFactory);
         }
+        if (b.disableLocalWorkspace) {
```
oss-maintainer, line 166: Returning null here only covers the parent build. Child builders copy `disableFilesystemTools/ShellTool/MemoryTools/MemoryHooks/SessionPersistence/WorkspaceContext` but not `disableLocalWorkspace`, and they are handed `.workspace(workspace)` (HarnessAgentBuilderSupport.java:382) — so with `b.workspace != null` they fall through to the default `new LocalFilesystemSpec().toFilesystem(...)` branch below and get a real local filesystem over the ephemeral root. "No local filesystem" therefore holds for the parent only. Please propagate the flag to sub-builders (or state the asymmetry in the javadoc/docs table).

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2549,6 +2627,13 @@ public HarnessAgent build() {
                     if (wsManager.getFilesystem() != null) {
                         effectiveTranscriptStore =
                                 new ObjectStoreTranscriptStore(wsManager.getFilesystem());
+                    } else if (disableLocalWorkspace) {
+                        // No local workspace and no filesystem: transcripts would have to be
+                        // persisted to ${user.dir}, which this mode explicitly forbids.
```
oss-maintainer, line 2656: Transcripts warn-and-skip, but the memory/session tools stay on by default (`!disableMemoryTools`, HarnessAgent.java:2779-2784) and `WorkspaceManager` writes them to local disk under the ephemeral root (WorkspaceManager.java:858/891 `Files.createDirectories`). So in this mode MEMORY/sessions silently persist to a location an OS tmp cleaner can sweep. The javadoc does say "left enabled … reads and writes the ephemeral temp workspace" — a line in the docs table about durability (and tmp-reaper risk) would make that explicit for operators.

oss-maintainer (commented):
### Summary
Adds `HarnessAgent.Builder.disableLocalWorkspace()`: the default workspace resolves under `${java.io.tmpdir}` instead of `${user.dir}/.agentscope/workspace`, `resolveFilesystem()` returns no default local filesystem in that mode, and the workspace-backed defaults that would still touch disk (`JsonFileAgentStateStore`, `WorkspaceTaskRepository`) fail fast with actionable messages instead of silently writing. Solid shape — the fail-fast-instead-of-silent-degradation choice is the right one, the two throw paths are unit-pinned, and the docs table in both languages matches the code.

The gaps below are all in the same family: `filesystem` becomes nullable, and only some of its consumers learned about it. None of them is a blocker for the intended (fully-opted-out) configuration, but each is reachable from a builder combination a user can write.

### Findings
- **[Warning]** `HarnessAgent.java:2804` — `deliver_artifact` is registered without a null-filesystem guard (`artifactDeliveryEnabled` at :2805 only checks the target + `!disableFilesystemTools`) → NPE at call time in this mode.
- **[Warning]** `HarnessAgent.java:1195` — the ephemeral path is host-global per `agentId`: collides across replicas/tenants on one host and is never reclaimed.
- **[Warning]** `HarnessAgentBuilderSupport.java:166` — sub-agent builders do not inherit `disableLocalWorkspace()`, so children still get a default local filesystem over the ephemeral root.
- **[Info]** `HarnessAgent.java:2487` — the throw ignores `agentscope.state.home`, which is the other supported way to move the state tree off `$HOME`.
- **[Info]** `HarnessAgent.java:2632` — transcripts warn-and-skip, but memory/session tools keep writing to the temp root; durability (tmp reaper) is worth a docs line.

### Suggestions
```java
boolean artifactDeliveryEnabled =
        artifactDeliveryTarget != null && !disableFilesystemTools && filesystem != null;
```
and for the sub-agent path, capture the flag next to the other `capturedDisable*` values (`if (capturedDisableLocalWorkspace) sub.disableLocalWorkspace();`) so the promise is per-agent-tree rather than per-builder.

Positives: `resolveEphemeralWorkspace()` sanitizes the agent id and has a blank-id fallback; the log line no longer NPEs on `filesystem.getClass()`; `MarketplaceStager` is explicitly skipped rather than pointed at a half-wired root.

CLA signed (`license/cla=success` on `d6c22d39`), CI green (`statusCheckRollup=SUCCESS`), `mergeable=MERGEABLE/clean`. Re-review after the rebase is not needed for the above — the delta is confined to the two files touched here. Not approving: the `deliver_artifact` path is a reachable runtime NPE, and the sub-agent asymmetry changes what the flag means for anyone using subagents.

---

Letter2025:
Thanks @oss-maintainer for the thorough review — all five points are addressed in the new commit `a4db909f`:

**1. `deliver_artifact` NPE (warning)** — `artifactDeliveryEnabled` now also requires `filesystem != null` (`HarnessAgent.java`), so a disabled-mode build with an artifact target simply skips registering `ArtifactDeliveryTool` instead of registering a tool that would NPE at call time.

**2. Ephemeral path collisions (warning)** — `resolveEphemeralWorkspace()` now inserts a per-JVM nonce: `${java.io.tmpdir}/agentscope-workspace/<nonce>/<agentId>`. Replicas / tenants on one host no longer collide in the same tree. Tests updated to pin the new layout (nonce + trailing agent id) and the blank-id fallback.

**3. Sub-agent flag propagation (warning)** — `disableLocalWorkspace()` is now captured and re-applied in both `buildGeneralPurposeFactory` and `buildDeclaredFactory` (`HarnessAgentBuilderSupport.java`), so "no local filesystem" holds for the whole agent tree, not just the parent. (Note: sub-builders still get the parent's explicit `workspace(...)` / filesystem spec when one is set — only the default-local branch is suppressed.)

**4. `agentscope.state.home` (info)** — the fail-fast check now stands down when `-Dagentscope.state.home` is set (the default state tree then lives outside `$HOME`), and the error message mentions the property as an option. New test `disabledBuild_withStateHomeOverride_doesNotFailFast` covers it.

**5. Durability of the temp workspace (info)** — the docs table (EN + ZH) now has an explicit row: memory / session tools left enabled write to the ephemeral temp tree which is **not durable** (OS tmp reaper), steering operators to disable them or back them with a remote store. The builder javadoc mirrors this.

All changes were formatted with `spotless:apply`; `HarnessAgentLocalWorkspaceDisabledTest` (now 6 tests) passes locally, and CI is re-running on the pushed commit. Could you re-review when the checks go green?

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1177,6 +1182,28 @@ static Path resolveDefaultWorkspace() {
...
+    static Path resolveEphemeralWorkspace(String agentId) {
+        String safeAgentId = HarnessAgentBuilderSupport.sanitizeIdentifier(agentId);
+        if (safeAgentId == null || safeAgentId.isBlank()) {
+            safeAgentId = "ReActAgent";
+        }
+        return Paths.get(System.getProperty("java.io.tmpdir"))
+                .resolve("agentscope-workspace")
+                // Per-JVM nonce keeps multi-replica / multi-tenant deployments on one host from
+                // colliding in the same tree; the previous pattern was host-global per agentId.
+                .resolve(EPHEMERAL_WORKSPACE_NONCE)
```
oss-maintainer, line 1203: **[Info]** The per-JVM nonce closes the cross-replica collision from the previous round, but it makes the old orphan problem structural rather than incidental: every JVM start now writes into a brand-new `${java.io.tmpdir}/agentscope-workspace/<nonce>/<agentId>` tree, and `close()` does not remove it, so a long-lived host that runs many short-lived agents accumulates one unreachable tree per JVM forever. Worth either deleting the nonce root when the agent owned it (mirroring the `ownedWorkspaceIndex` pattern already in `close()`), or sweeping stale nonce directories on startup. Not blocking — the mode is explicitly documented as non-durable now.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2414,6 +2497,17 @@ public HarnessAgent build() {
                             : new LocalPeriodicGate();
 
             AgentStateStore effectiveSession = stateStoreOverride;
+            if (disableLocalWorkspace
+                    && effectiveSession == null
+                    && System.getProperty("agentscope.state.home") == null) {
```
oss-maintainer, line 2510: **[Info]** Accepting `-Dagentscope.state.home` as a third way to satisfy the fail-fast is the right call — it is already the supported override for the default state tree, and refusing it would have forced a custom store on anyone who had deliberately relocated state. One asymmetry to keep in mind: the property is read here at build time, while `resolveEphemeralWorkspace()` derives its path from `java.io.tmpdir` at first use, so the workspace location and the state location are resolved through two different mechanisms. Fine as is; a follow-up that exposes both through the builder would make `disableLocalWorkspace()` fully property-free for tests. The new `disabledBuild_withStateHomeOverride_doesNotFailFast` test pins the behaviour, which is the important part.

agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentLocalWorkspaceDisabledTest.java
```diff
@@ -0,0 +1,278 @@
...
+                        .disableDefaultWorkspaceSkills()
+                        .disableWorkspaceContext()
+                        .disableTranscript()
+                        .disableCompaction()
+                        .disableToolResultEviction()
+                        .taskRepository(new NoopTaskRepository())
+                        .build()) {
+            // build succeeded without an explicit state store — the default JsonFile store is
+            // permitted because state.home moves it outside the workspace concern.
+            assertTrue(true);
```
oss-maintainer: **[Warning]** `assertTrue(true)` asserts nothing — the test is really "`build()` must not throw", which JUnit already reports as an error if it does, so the literal only adds noise at the failure site. Either drop it, or make the test say what it means, e.g. assert that the state tree actually landed under the override (`assertTrue(Files.startsWith(stateHome, <resolved state dir>))`) so the test would also catch the override being accepted but ignored.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
```diff
@@ -565,6 +576,7 @@ static SubagentFactory buildDeclaredFactory(
             if (capturedDisableShellTool) sub.disableShellTool();
             if (capturedDisableMemoryTools) sub.disableMemoryTools();
             if (capturedDisableMemoryHooks) sub.disableMemoryHooks();
+            if (capturedDisableLocalWorkspace) sub.disableLocalWorkspace();
```
oss-maintainer, line 579: **[Info]** Good catch propagating this to both factories. Worth noting the sub-agent then relies on `capturedStateStore` being non-null (lines 419/569) to pass the same fail-fast, so a parent that satisfies the check via `-Dagentscope.state.home` while a *declared* sub-agent is built before the property is visible would be the only way to hit the new throw from a child. `buildDeclaredFactory` also does not capture `capturedDisableWorkspaceContext` / `disableDefaultWorkspaceSkills` on the same footing as the general-purpose factory — pre-existing and out of scope here, just flagging that the parity list is now 10 flags in one factory and 7 in the other.

oss-maintainer (approved):
### Summary
Re-review at `a4db909f` after @Letter2025 addressed all five points from the previous round. Every one of them is fixed, and fixed the way it was suggested:

- `deliver_artifact` NPE → `artifactDeliveryEnabled` now also requires `filesystem != null` (`HarnessAgent.java:2622`), which is exactly the registration gate at `build()`
- host-global ephemeral path → per-JVM `EPHEMERAL_WORKSPACE_NONCE` segment (`:173`, `:1203`), with the Javadoc and the layout assertions in `ephemeralWorkspace_resolvesOutsideWorkingDirectory` updated in step
- sub-agents not inheriting the flag → `capturedDisableLocalWorkspace` captured and applied in both `buildGeneralPurposeFactory` and `buildDeclaredFactory`
- fail-fast ignoring `agentscope.state.home` → the property now satisfies the check (`:2502`) with a new test pinning it
- non-durability of the temp root → documented explicitly in the `disableLocalWorkspace()` Javadoc, in both the EN and ZH workspace docs, and in the release notes

Approving. The four inline notes are follow-up grade, not blockers.

### Findings
- **[Info]** HarnessAgent.java:1203 — the nonce makes orphaned ephemeral trees structural: `close()` never removes the per-JVM root, so a host running many short-lived agents accumulates one unreachable tree per JVM
- **[Info]** HarnessAgent.java:2502 — workspace path and state path are now resolved through two mechanisms (builder/first-use vs. a build-time property read); fine here, worth unifying later
- **[Warning]** HarnessAgentLocalWorkspaceDisabledTest.java:144 — `assertTrue(true)` asserts nothing; drop it or assert the override is actually honoured
- **[Info]** HarnessAgentBuilderSupport.java:579 — propagation is right in both factories; note the two factories' flag-parity lists are no longer the same length (pre-existing)

### Verification
CI green on `a4db909f` (`validate`, `build (ubuntu-latest)`, `build (windows-latest)`, `Check License`, `Check Module Sync`, `codecov/patch`, `license/cla`), CLA signed, `MERGEABLE/BLOCKED` only because a review was required — this review supplies it.

---

Letter2025:
Thanks for the approval @oss-maintainer! One quick follow-up on your non-blocking notes: the `assertTrue(true)` in `disabledBuild_withStateHomeOverride_doesNotFailFast` is now a real assertion — it verifies the default `JsonFileAgentStateStore` directory actually lands under `${agentscope.state.home}` (pushed as `bb56d37f`, 6/6 tests pass locally). The other follow-ups (nonce-tree cleanup on `close()`, path-resolution unification) are noted as future work. Thanks again for the thorough review!

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -1177,6 +1182,28 @@ static Path resolveDefaultWorkspace() {
...
+     * enabled and no explicit {@link Builder#workspace(Path)} is set. The path lives under the
+     * JVM temp directory so that nothing is materialised in the application's working directory;
+     * it is defensive fallback and is not created eagerly.
+     *
+     * @param agentId the resolved agent id (already non-blank)
+     * @return the ephemeral workspace directory; never {@code null}
+     */
+    static Path resolveEphemeralWorkspace(String agentId) {
+        String safeAgentId = HarnessAgentBuilderSupport.sanitizeIdentifier(agentId);
+        if (safeAgentId == null || safeAgentId.isBlank()) {
```
oss-maintainer, line 1196: The per-JVM nonce makes the ephemeral workspace path non-deterministic and un-collectable:

1. **Nothing removes these trees.** `disableLocalWorkspace()` is aimed exactly at containerised/SaaS deployments, where `${java.io.tmpdir}` is frequently a size-bounded tmpfs — every JVM start of every agent id leaks another `<nonce>/<agentId>` tree forever.
2. **No sharing.** Two replicas of the same deployment can never resolve the same ephemeral workspace, and a restart loses whatever the workspace-backed fallbacks did write (transcript/`skills-cache` are skipped, but anything materialised later is orphaned).

For the collision problem this was added to solve, could you either add a cleanup path (e.g. register the root for deletion on close / prune sibling nonces older than the JVM's own start under `agentscope-workspace`), or document `disableLocalWorkspace()` as "a throwaway path per JVM lifetime, caller-managed cleanup"? `Files.createTempDirectory("agentscope-ws-")` + `deleteOnExit`, or a nonce-less path with an explicit `workspace(...)` escape hatch, would both be simpler to reason about.

oss-maintainer (commented):
### Summary
Adds an opt-in `HarnessAgent.Builder.disableLocalWorkspace()` so SaaS/containerised deployments stop materialising `${user.dir}/.agentscope/workspace`: the workspace resolves to an ephemeral path under `${java.io.tmpdir}`, no default local filesystem is created, and persistence gaps become fail-fast `IllegalStateException`s (missing explicit state store / task repository) or warned degradations (transcripts, `.skills-cache`, `FilesystemTool`). The design is right — additive flag, default behaviour untouched, explicit fail-fast instead of silent local writes, and 280 lines of dedicated tests plus EN/ZH doc updates. CLA signed, all checks green, `mergeStateStatus=CLEAN`.

Re-reviewing the new commit (`bb56d37f`): it only replaces the `assertTrue(true)` placeholder in `disabledBuild_withStateHomeOverride_doesNotFailFast` with a real assertion, and the expected path is consistent with the `workingDirectory.resolve("state-home")` value set in the test — good change, production code is unchanged since my previous approval.

### Findings
- **[Warning]** `HarnessAgent.java:1196` — the per-JVM nonce makes the ephemeral workspace path un-collectable (nothing prunes old nonce trees, and `/tmp` is usually a bounded tmpfs in exactly the deployments this flag targets) and prevents two replicas of one deployment from sharing it. Please add cleanup or document caller-managed cleanup.
- **[Warning]** `HarnessAgent.java:2503` — per-build validation depends on the global `agentscope.state.home` system property, making the fail-fast contract order-dependent; the same coupling forces the new test to mutate shared JVM state. A builder-level `stateHome(...)` (property kept as a documented fallback) would make both deterministic.
- **[Info]** `HarnessAgent.java:2648` — `disableLocalWorkspace()` now silently degrades transcripts, skills staging and filesystem tools; consider one aggregated "capabilities disabled" log line at build time, or a strict variant that turns them into errors.

None of these are blockers for the flag as proposed; they are about the lifecycle of the ephemeral tree and the determinism of the validation. Holding off on re-approving on the new head until the nonce/cleanup question is answered, since it decides the public contract of `disableLocalWorkspace()`.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2414,6 +2497,17 @@ public HarnessAgent build() {
                             : new LocalPeriodicGate();
 
             AgentStateStore effectiveSession = stateStoreOverride;
+            if (disableLocalWorkspace
+                    && effectiveSession == null
+                    && System.getProperty("agentscope.state.home") == null) {
+                throw new IllegalStateException(
```
oss-maintainer, line 2511: Two things bundled into this check:

- It reads the **global** `agentscope.state.home` system property to make a **per-build** decision, so the fail-fast contract is order-dependent: agent A built before the property is set throws, agent B built after does not. A deployment that sets the property programmatically (not via `-D`) gets different validation results over its lifetime.
- The companion test drives this by `System.setProperty`/`clearProperty` on shared JVM state, which is why it has to be sequential. That is fine for one test, but it will make the whole `HarnessAgent...Test` class fragile if more builds start touching `state.home` in parallel (surefire parallel mode / future `@ParameterizedTest`).

Would it be possible to expose the value on the builder instead (e.g. `Builder#stateHome(Path)` / `stateStore(...)` only), keeping the system property as a fallback read once at a documented point? That would make the fail-fast deterministic and let the test assert on the resolved store path rather than on global state.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2549,6 +2643,13 @@ public HarnessAgent build() {
                     if (wsManager.getFilesystem() != null) {
                         effectiveTranscriptStore =
                                 new ObjectStoreTranscriptStore(wsManager.getFilesystem());
+                    } else if (disableLocalWorkspace) {
+                        // No local workspace and no filesystem: transcripts would have to be
+                        // persisted to ${user.dir}, which this mode explicitly forbids.
```
oss-maintainer, line 2656: With `disableLocalWorkspace()` there are now three silent degradations: transcripts (warn here), skills staging (`.skills-cache` not staged) and `FilesystemTool` (not registered). In a production SaaS deployment the most likely failure is a configured skill or a prompt that assumes file tools quietly doing nothing, with only a single `log.warn` line to explain it.

Two cheap options, either is fine:
- a one-shot aggregated `log.info` at the end of `build()` listing every capability that was disabled by this flag ("workspace: ephemeral / transcripts: disabled / skills staging: disabled / filesystem tool: disabled");
- or an opt-in strict variant (`.disableLocalWorkspace(StrictMode.FAIL)` / `failOnDisabledCapabilities()`) so teams can turn the silent degradations into build-time errors in production.

oss-maintainer:
Thanks for the follow-up @Letter2025 — the real assertion in `disabledBuild_withStateHomeOverride_doesNotFailFast` (checking that the default `JsonFileAgentStore` directory actually lands under the overridden `state.home`) is exactly what I was asking for, and it is a better test than the placeholder was.

The item still gating a re-approval is the ephemeral-workspace lifecycle, not the test: each `disableLocalWorkspace()` build mints a fresh nonce directory under `${java.io.tmpdir}` and nothing prunes the older ones. In the SaaS/container deployments this flag targets `/tmp` is often a bounded tmpfs, so a process that rebuilds agents (config reload, tenant churn, tests) accumulates a tree per build. A one-line answer is enough: either (a) document that the caller owns cleanup, or (b) reuse a deterministic path per agent instead of a per-build nonce.

Agreed on deferring the nonce-tree cleanup and the path-resolution unification as follow-up work — flagging here so the deferral is visible rather than silent, and I will not hold the PR on the other two notes.

---

Letter2025:
Good catch on the lifecycle, and thanks for keeping the bar — quick clarification plus the doc fix, pushed as `02d7e03a`:

**On "a fresh nonce directory per build":** the nonce is actually `static final` per JVM (`EPHEMERAL_WORKSPACE_NONCE`), so all builds inside one process share the same `${java.io.tmpdir}/agentscope-workspace/<nonce>/` root — the same agentId **reuses** its directory across rebuilds; no per-build accumulation occurs within a process. What does accumulate is one root per JVM start / per replica on a host, since nothing prunes old roots.

Addressing your option (a) — caller-owned cleanup — is now documented in three places:

- `disableLocalWorkspace()` javadoc: the ephemeral tree is never deleted by `close()` / the harness; per-JVM nonce reuse within a process; deployments that rebuild agents frequently should purge `${java.io.tmpdir}/agentscope-workspace` (e.g. at process start) or rely on the OS tmp reaper
- workspace docs table (EN + ZH): new "Cleanup of the ephemeral tree" row stating the same contract

That closes out the gating item. The nonce-tree cleanup on `close()` and path-resolution unification stay deferred as agreed. Thanks again!

oss-maintainer (approved):
### Summary
Re-reviewed `bb56d37f..2700dc6e` (two commits). The delta answers the one question I was holding re-approval on, so I am approving this head.

`02d7e03a` resolves the ephemeral-tree lifecycle point by stating the ownership instead of leaving it implied, and the wording matches the code: `EPHEMERAL_WORKSPACE_NONCE` is a `static final` per-JVM value (`HarnessAgent.java:173`) applied above the `agentId` segment in `resolveEphemeralWorkspace` (`HarnessAgent.java:1199-1204`), so repeated builds of one agent inside a process really do land in the same tree and nothing accumulates per build — my earlier concern about a fresh directory on every rebuild was wrong, and your correction is the accurate one. The remaining exposure is exactly what the new paragraph says: a JVM restart or a second replica on the same host mints a new root, so `close()` deliberately does not delete and the caller purges `${java.io.tmpdir}/agentscope-workspace` if rebuilds are frequent. Stating that in the javadoc plus the EN/ZH capability tables is the right resolution for a flag whose whole purpose is not to own the deployment's disk.

`2700dc6e` fixes the unmatched `</li>` left in that paragraph, which is a genuine doclint break rather than cosmetics.

### Still open from the previous round (non-blocking, unchanged by this delta)
- `HarnessAgent.java:2508` — the fail-fast check keys off the global `agentscope.state.home` system property, so whether `build()` throws depends on JVM-wide state rather than the builder. A builder-level `stateHome(...)` would make the contract order-independent; worth a follow-up issue rather than expansion of this PR.
- `HarnessAgent.java:2654` — the transcript-backend warning is still per-degradation; one aggregated "capabilities disabled under disableLocalWorkspace()" line at build time would be easier to spot in a startup log.

### Checks
`license/cla` signed, `Check License` / `Check Module Sync` / builds green (`mergeable=true`). No local build run on my side, so this rests on GitHub's own checks plus reading the two patches. Merging stays with the maintainers.

---

PR #2767, opened 2026-08-18
feat(models): add support for gemini server tools (#2241)

### LINKED ISSUE #2241: [Feature]: Gemini 内置工具支持
当前仅支持用户自定义 function calling，不支持 Gemini 内置 grounding/server tools。无法用 Google Search grounding、Maps grounding、URL context 等。

pony-huang:
## AgentScope-Java Version

## Description

Fixes #2241 Gemini 支持官方内置工具，如ContextUrl, GoogleMaps, GoogleSearch等。

## Simple
```java
public class Main {
    public static void main(String[] args) throws JsonProcessingException {

        GeminiChatModel gemini =
                GeminiChatModel.builder()
                        .modelName("gemini-3.5-flash")
                        .streamEnabled(true)
                        .serverTools(List.of(GeminiServerTool.googleSearch().build()))
                        .build();

        WeatherTools weatherTools = new WeatherTools();
        toolkit.registerTool(weatherTools);
        ReActAgent agent =
                ReActAgent.builder()
                        .name("weather-agent")
                        .model(gemini)
                        .toolkit(toolkit)
                        .maxIters(6)
                        .build();

        Msg response =
                agent.call(
                                new UserMessage(
                                        """
                                        What is the southernmost city in the China? What's the weather like there today?\
                                        """))
                        .block();

        assert response != null;
        System.out.println(response.getTextContent());
        System.out.println(new ObjectMapper().writeValueAsString(response.getMetadata()));
    }

    static final class WeatherTools {

        @Tool(
                name = "getWeather",
                description = "Get the weather in a given location",
                readOnly = true)
        public String getWeather(
                @ToolParam(
                                name = "location",
                                description = "The city and state, e.g. San Francisco, CA",
                                required = true)
                        String location) {

            System.out.println("getWeather called with location: " + location);
            return """
            {
              "response": "Very hot. it's  96.8°F",
            }
            """;
        }
    }
}
```

## Checklist

- [√]  Code has been formatted with `mvn spotless:apply`
- [√]  All tests are passing (`mvn test`)
- [√]  Javadoc comments are complete and follow project conventions
- [√]  Related documentation has been updated (e.g. links, examples, etc.)
- [√]  Code is ready for review

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -1798,6 +1798,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                             .collect(Collectors.toSet());
 
             return lastAssistant.getContentBlocks(ToolUseBlock.class).stream()
+                    .filter(toolUseBlock -> !toolUseBlock.isServer())
                     .map(ToolUseBlock::getId)
                     .filter(id -> !existingResultIds.contains(id))
                     .collect(Collectors.toSet());
@@ -3289,7 +3290,24 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
             // No tool calls - finished
             // If there are tool calls (even non-existent ones), continue to acting phase
             // where ToolExecutor will return "Tool not found" error for the model to see
-            return toolCalls.isEmpty();
+            if (toolCalls.isEmpty()) {
+                return true;
+            }
+
+            // Server tool calls are executed by the provider and their results arrive in the
+            // same assistant message: if every tool call is a server tool with its result
+            // present, there is nothing left to act on. A server tool call without a result
+            // (e.g. pause_turn) keeps the loop running so the conversation goes back to the
+            // provider to continue.
+            Set<String> inlineResultIds =
+                    msg.getContentBlocks(ToolResultBlock.class).stream()
+                            .map(ToolResultBlock::getId)
+                            .collect(Collectors.toSet());
+            return toolCalls.stream()
+                    .allMatch(
+                            toolCall ->
+                                    toolCall.isServer()
+                                            && inlineResultIds.contains(toolCall.getId()));
         }
 
         /**
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
@@ -173,6 +173,7 @@ class ToolExecutor {
      *
      * <p>This method handles:
      * <ul>
+     *   <li>Server-side tool calls short-circuit (never executed locally)</li>
      *   <li>Tool lookup and validation</li>
      *   <li>Group activation check</li>
      *   <li>Parameter merging (preset + input)</li>
@@ -183,6 +184,21 @@ class ToolExecutor {
      */
     private Mono<ToolResultBlock> executeCore(ToolCallParam param) {
         ToolUseBlock toolCall = param.getToolUseBlock();
+
+        // Server-side (built-in) tools are executed by the model provider itself and their
+        // results (if any) already arrive in the model response. Never dispatch them to the
+        // local toolkit, even if a local tool happens to share the same name.
+        if (toolCall.isServer()) {
+            return Mono.just(
+                    ToolResultBlock.suspended(
+                            toolCall,
+                            new ToolSuspendException(
+                                    "Server-side tool '"
+                                            + toolCall.getName()
+                                            + "' is executed by the model provider and cannot"
+                                            + " be dispatched locally")));
+        }
+
         AgentTool tool = toolRegistry.getTool(toolCall.getName());
 
         if (tool == null) {
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/GeminiChatModel.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/GeminiChatModel.java
@@ -34,6 +34,8 @@ import io.agentscope.core.model.ModelException;
 import io.agentscope.core.model.ToolSchema;
 import io.agentscope.core.model.transport.ProxyConfig;
 import io.agentscope.extensions.model.gemini.formatter.GeminiChatFormatter;
+import io.agentscope.extensions.model.gemini.formatter.GeminiToolsHelper;
+import io.agentscope.extensions.model.gemini.tool.GeminiServerTool;
 import java.time.Instant;
 import java.util.List;
 import java.util.Objects;
@@ -55,6 +57,8 @@ import reactor.core.scheduler.Schedulers;
  * <ul>
  * <li>Text generation with streaming and non-streaming modes</li>
  * <li>Tool/function calling support</li>
+ * <li>Gemini server-side tools, including Google Search, Google Maps, URL Context, and code
+ *     execution</li>
  * <li>Multi-agent conversation with history merging</li>
  * <li>Vision capabilities (images, audio, video)</li>
  * <li>Thinking mode (extended reasoning)</li>
@@ -77,6 +81,7 @@ public class GeminiChatModel extends ChatModelBase {
     private final GenerateOptions defaultOptions;
     private final Formatter<Content, GenerateContentResponse, GenerateContentConfig.Builder>
             formatter;
+    private final List<GeminiServerTool> serverTools;
 
     /**
      * Creates a new Gemini chat model instance.
@@ -96,6 +101,8 @@ public class GeminiChatModel extends ChatModelBase {
      * @param defaultOptions default generation options
      * @param formatter      the message formatter to use (null for default Gemini
      *                       formatter)
+     * @param serverTools    Gemini built-in tools executed by the model provider; copied
+     *                       defensively, or empty when null
      */
     public GeminiChatModel(
             String apiKey,
@@ -109,7 +116,8 @@ public class GeminiChatModel extends ChatModelBase {
             GoogleCredentials credentials,
             ClientOptions clientOptions,
             GenerateOptions defaultOptions,
-            Formatter<Content, GenerateContentResponse, GenerateContentConfig.Builder> formatter) {
+            Formatter<Content, GenerateContentResponse, GenerateContentConfig.Builder> formatter,
+            List<GeminiServerTool> serverTools) {
         this.apiKey = apiKey;
         this.modelName = Objects.requireNonNull(modelName, "Model name is required");
         this.streamEnabled = streamEnabled;
@@ -122,6 +130,7 @@ public class GeminiChatModel extends ChatModelBase {
         this.defaultOptions =
                 defaultOptions != null ? defaultOptions : GenerateOptions.builder().build();
         this.formatter = formatter != null ? formatter : new GeminiChatFormatter();
+        this.serverTools = serverTools != null ? List.copyOf(serverTools) : List.of();
 
         // Initialize Gemini client
         Client.Builder clientBuilder = Client.builder();
@@ -202,7 +211,8 @@ public class GeminiChatModel extends ChatModelBase {
                 credentials,
                 clientOptions,
                 defaultOptions,
-                formatter);
+                formatter,
+                null);
     }
 
     private static HttpOptions resolveHttpOptions(String baseUrl, HttpOptions httpOptions) {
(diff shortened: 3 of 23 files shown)


oss-maintainer:
**CLA Not Signed**

@pony-huang please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.

---

agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
```diff
@@ -183,6 +184,21 @@ Mono<ToolResultBlock> execute(ToolCallParam param) {
      */
     private Mono<ToolResultBlock> executeCore(ToolCallParam param) {
         ToolUseBlock toolCall = param.getToolUseBlock();
+
+        // Server-side (built-in) tools are executed by the model provider itself and their
+        // results (if any) already arrive in the model response. Never dispatch them to the
+        // local toolkit, even if a local tool happens to share the same name.
+        if (toolCall.isServer()) {
```
oss-maintainer, line 191: Returning a **suspended** result here has a semantic that doesn't match the situation. `ToolResultBlock.suspended(...)` sets `METADATA_SUSPENDED`, which in `ReActAgent` routes the call into `pendingPairs` -> `buildSuspendedMsg(...)` -> `GenerateReason.TOOL_SUSPENDED`, and `determineToolResultState()` reports it as `RUNNING`. That path means "a human/external executor must still produce this result", but a server-side tool call is already done from the provider's side and nobody will ever resume it, so the loop can park instead of making progress.

Also note this guard looks unreachable via the normal ReAct path: `getPendingToolUseIds()` in this same PR already filters out `isServer()` blocks, so `extractPendingToolCalls()` never hands a server call to the executor.

Suggest either (a) dropping the guard and keeping the `isServer()` filter as the single source of truth, or (b) returning an error-style result (like the existing "Tool not found" branch below) so the model gets a recoverable signal instead of a suspension. If (b) is chosen, a regression test asserting the ReAct loop still terminates would be valuable.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiResponseParser.java
```diff
@@ -178,21 +200,114 @@ protected void parseToolCall(
...
+     *
+     * @param toolCall         Gemini ToolCall object
+     * @param thoughtSignature Thought signature from the Part (may be null)
+     * @param blocks           List to add parsed ToolUseBlock to
+     */
+    protected void parseToolCall(
+            ToolCall toolCall, byte[] thoughtSignature, List<ContentBlock> blocks) {
+        try {
+            String id = toolCall.id().orElse("tool_call_" + System.currentTimeMillis());
+            String name = toolCall.toolType().map(ToolType::toString).orElse("");
```
oss-maintainer, line 227: `ToolType::toString` yields the SDK enum's UPPER_SNAKE name — your own tests assert `"GOOGLE_SEARCH_WEB"` / `"GOOGLE_MAPS"` — while `GeminiServerTool` uses lower-case wire constants (`google_search`, `google_map`, `url_context`, `code_execution`). So the value parsed out of the response is not the same namespace as the value configured on the request, and `GeminiMessageConverter` later echoes `tub.getName()` straight back into `.toolType(...)`.

Could you confirm the round-trip is lossless, i.e. that a response parsed as `GOOGLE_SEARCH_WEB` re-serialises to a `toolType` the API accepts? A test that asserts *emitted request part == parsed response part* for each supported tool type would lock this down and prevent a silent "server tool call disappears on turn 2" bug. Ideally there would be a single canonical naming source shared by `GeminiServerTool` and the parser.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiMessageConverter.java
```diff
@@ -137,30 +151,59 @@ public List<Content> convertMessages(List<Msg> msgs) {
...
+                    // for an independent user Content after the current message.
                     String textOutput = convertToolResultToString(trb.getOutput());
+                    if (trb.isServer()) {
+                        // Create ToolResponse for server-side (built-in) tools, the output is
+                        // placed under the "response" key
+                        Map<String, Object> responseMap;
+                        try {
+                            // The parser serializes the tool response map to a JSON string, so
+                            // parse it back rather than converting the raw string value.
+                            responseMap = JsonUtils.getJsonCodec().fromJson(textOutput, Map.class);
```
oss-maintainer, line 164: This round-trips structured data through a JSON string: `parseToolResponse` serialises `toolResponse.response()` into a single `TextBlock`, and here you parse it back to satisfy Gemini's "response must be an object" contract.

Two concerns:
1. The fallback `Map.of("output", textOutput)` silently changes the payload shape the provider expects, so a parse failure degrades into a wrong-but-valid request rather than a visible error. Worth at least distinguishing "text was not a JSON object" from "text was corrupted", or logging at error level.
2. Any intermediate transformation of that text block (context compaction, memory flush, sanitisation, truncation of long grounding results) will break the re-parse. Since `ToolResultBlock` is a general message model shared with harness/distribution, storing the raw map in `metadata` (or as a dedicated block type) and only rendering text for display would be more robust.

Minor: `fromJson(textOutput, Map.class)` is a raw type and needs `@SuppressWarnings("unchecked")` or a `TypeReference<Map<String, Object>>` for consistency with the rest of the codebase.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiMessageConverter.java
```diff
@@ -113,16 +116,27 @@ public List<Content> convertMessages(List<Msg> msgs) {
...
-                    // Build Part with FunctionCall and optional thought signature
-                    Part.Builder partBuilder = Part.builder().functionCall(functionCall);
+                    // Build Part with FunctionCall or ToolCall and optional thought signature
+                    Part.Builder partBuilder = null;
+                    if (tub.isServer()) {
+                        // Create ToolCall for server-side (built-in) tools
+                        ToolCall toolCall =
+                                ToolCall.builder()
+                                        .id(tub.getId())
+                                        .toolType(tub.getName())
```
oss-maintainer, line 126: Nit on symmetry: for local calls you keep `id`/`name`, here `toolType` is fed from `tub.getName()` (see the parser comment about `GOOGLE_SEARCH_WEB` vs `google_search`). If a server tool call ever reaches this path without going through `parseToolCall` — e.g. a session restored from an older persisted state where `server` defaulted to `false`, or a locally registered tool that happens to share a built-in tool's name — the request shape silently changes between `functionCall` and `toolCall`.

Consider documenting which producer is allowed to set `server=true`, since it is persisted in `AgentState` and therefore crosses version boundaries.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -3289,7 +3290,24 @@ private boolean isFinished(Msg msg) {
             // No tool calls - finished
             // If there are tool calls (even non-existent ones), continue to acting phase
             // where ToolExecutor will return "Tool not found" error for the model to see
-            return toolCalls.isEmpty();
+            if (toolCalls.isEmpty()) {
```
oss-maintainer, line 3293: Nice that the mixed local/server case is handled. One edge to confirm: when a server tool call arrives **without** an inline result (the `pause_turn` case in your comment), `isFinished` returns `false`, `getPendingToolUseIds()` excludes it, so `acting()` sees an empty `pendingToolCalls` and falls through to `executeIteration(iter + 1)` — an extra model round-trip whose only termination guard is `maxIters`.

Is that intentional, and is it bounded in practice? If the provider can answer again with the same unresolved server call, this becomes `maxIters` wasted requests. A test covering "server call with no result" (asserting iteration count / final reason) would make the intended behaviour explicit, and the Javadoc here could state who is responsible for terminating the pause_turn loop.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/GeminiChatModel.java
```diff
@@ -109,7 +116,8 @@ public GeminiChatModel(
             GoogleCredentials credentials,
             ClientOptions clientOptions,
             GenerateOptions defaultOptions,
-            Formatter<Content, GenerateContentResponse, GenerateContentConfig.Builder> formatter) {
+            Formatter<Content, GenerateContentResponse, GenerateContentConfig.Builder> formatter,
+            List<GeminiServerTool> serverTools) {
```
oss-maintainer, line 120: Plumbing check: `serverTools` is only applied inside `buildGenerateContentConfig`, which is reached solely from `doStream`. Please confirm the non-streaming path (`streamEnabled == false`) and any other request-building site (token counting, formatter defaults) go through the same method — otherwise built-in tools are configured on the model but silently ignored for non-streaming calls, which is a confusing failure mode given `Builder.serverTools(...)` advertises unconditional support.

Also: `@JsonProperty("server") boolean server` on the `ToolUseBlock`/`ToolResultBlock` `@JsonCreator` is a primitive. Missing field in old persisted state deserialises to `false` (good), but an explicit JSON `null` throws for primitives. Since these blocks are stored by `AgentStateStore` implementations across harness/distribution, a `@Nullable Boolean` normalised in the constructor would be safer — plus a deserialisation test for a pre-change state JSON document.

oss-maintainer (commented):
### Summary

Adds first-class support for Gemini provider-executed (server-side) built-in tools — Google Search / Maps / URL Context / code execution — by introducing a `server` flag on `ToolUseBlock`/`ToolResultBlock`, parsing `tool_call`/`tool_response` parts, keeping server results inline in the `model` turn, short-circuiting them away from the local toolkit, and teaching `ReActAgent` to recognise an all-server-turn as finished. 23 files, ~2.0k additions, of which ~800 are tests.

This is a well-structured change: the core/extension split is clean, `enrichToolUseBlockWithId` and `ToolCallsAccumulator` were both updated so the new flags survive stream accumulation (an easy thing to miss), the parallel-`functionResponse` merging in `appendFunctionResponses` is a genuine correctness improvement, and test coverage is thorough. The ReAct-loop reasoning in the `isFinished` comment is exactly the right framing.

Overall verdict: **direction is good, but the PR currently has a merge conflict with `main` and the CLA is not signed, so it cannot be approved or merged as-is.** Beyond those blockers, the findings above cluster on two themes: (1) `server` is persisted state crossing version boundaries in a multi-module codebase, so its deserialisation and naming contracts need to be pinned down explicitly, and (2) two of the new code paths (the suspended-result short-circuit in `ToolExecutor`, and the unresolved-server-call loop in `ReActAgent`) have termination/resume semantics that look unintentional and deserve either a comment or a test.

### Findings
- **[Warning]** `ToolExecutor.java:191` — suspended result implies "awaiting external resume" and is reported as `RUNNING`; also appears unreachable given the new `isServer()` filter in `getPendingToolUseIds()`.
- **[Warning]** `GeminiResponseParser.java:227` — `ToolType::toString` produces `GOOGLE_SEARCH_WEB`, but `GeminiServerTool` constants are `google_search`; request/response naming namespaces diverge.
- **[Warning]** `GeminiMessageConverter.java:164` — structured data round-tripped through a JSON string; the raw-text fallback changes the payload shape silently, and any intermediate text transform breaks re-parse.
- **[Warning]** `GeminiChatModel.java:120` — confirm the non-streaming path also applies `serverTools`; primitive `boolean server` in `@JsonCreator` throws on explicit JSON `null` for persisted state.
- **[Info]** `GeminiMessageConverter.java:126` — document which producer may set `server=true`, since it is persisted in `AgentState`.
- **[Info]** `ReActAgent.java:3293` — server call without an inline result relies on `maxIters` as its only termination guard; worth a test asserting intended behaviour.

### Next steps
1. Rebase onto `main` and resolve the conflict (this PR touches `agentscope-core` message model, which has moved recently — rebase, don't merge).
2. Sign the Contributor License Agreement (see the CLA reminder comment).
3. Address the naming-namespace and non-streaming-path questions above; the others can be follow-ups.

Feel free to @mention me for a re-review once the conflicts are resolved — happy to look again.

---

oss-maintainer:

```bash
git fetch origin
git checkout issue-2241-support-gemini-server-tools
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #2769, opened 2026-08-18
fix(core): preserve permission context when loading legacy v1 session state

### LINKED ISSUE #2768: ReActAgent legacy-state load silently drops permissionContext (downgrades to DEFAULT)
# Issue draft — A3: LegacyStateLoader drops permission context (silent downgrade)

---
**Title:** `ReActAgent` legacy-state load silently drops `permissionContext` (downgrades to DEFAULT)

**Body:**

### Problem

`ReActAgent.loadOrCreateAgentStateForSlot(...)` loads a session's state in this order:

1. a new-format `agent_state` entry from the `AgentStateStore`,
2. else v1 **legacy** session keys (`memory_messages` / `toolkit_activeGroups`) via
   `LegacyStateLoader`,
3. else a fresh state built from the caller's parameters.

Step 2 is leaky: `LegacyStateLoader.loadFromLegacySessionWithPresence(...)` reconstructs the
`AgentState` with only `context(msgs)` (+ tool context), and **never sets `permissionContext`**.
Legacy keys carry no permission context (the concept did not exist in v1), so the reconstructed
state falls back to the default `DEFAULT` mode — even when the caller explicitly supplied a
non-default permission context (for example `BYPASS`) that `freshState(permCtx, ...)` (step 3)
would have preserved, and even though `permCtx` is already in scope at the call site.

Concrete trace (on `main`):

```java
// ReActAgent.loadOrCreateAgentStateForSlot
AgentState fresh = freshState(permCtx, agentId, userId, sessionId, initialActiveToolGroups); // has permCtx
...
        LegacyStateLoader.loadFromLegacySessionWithPresence(stateStore, userId, sessionId); // permCtx NOT passed
if (legacy.found()) {
    return new VersionedState<>(legacy.state(), version); // permission mode silently DEFAULT
}
```

```java
// LegacyStateLoader.loadFromLegacySessionWithPresence
AgentState.Builder builder = AgentState.builder().context(msgs); // no .permissionContext(...)
```

Impact: any session that has v1-era keys but no new-format `agent_state` yet (typical for the
first post-upgrade turn) silently loses its caller-configured permission mode. This can weaken
permission semantics (e.g. a pre-seeded `BYPASS` session degrading to `DEFAULT`).

### Proposed fix

- `LegacyStateLoader.loadFromLegacySession(...)` / `...WithPresence(...)`: add overloads accepting
  a `PermissionContextState`; attach it to the reconstructed `AgentState` when non-null (old
  overloads kept for backward compatibility).
- `ReActAgent.loadOrCreateAgentStateForSlot(...)`: pass the in-scope `permCtx` into the legacy
  load, exactly like `freshState` already does.

Behavior for callers without a specific permission context is unchanged.

### Environment

- agentscope-java `main` (2.0.3-SNAPSHOT); reproduced against 2.0.1 sources too (same path)

Letter2025:
# fix(core): preserve permission context when loading legacy v1 session state

Closes #2768

## Problem

`ReActAgent.loadOrCreateAgentStateForSlot(...)` loads a session's state in this order:

1. a new-format `agent_state` entry from the `AgentStateStore`,
2. else v1 **legacy** session keys (`memory_messages` / `toolkit_activeGroups`) via
   `LegacyStateLoader`,
3. else a fresh state built from the caller's parameters (`freshState(permCtx, ...)`).

Step 2 is leaky: `LegacyStateLoader` reconstructs the `AgentState` with only the legacy
conversation context (+ tool activation groups) and **never sets `permissionContext`**. Legacy
keys cannot carry one — the concept did not exist in v1 — so the reconstructed state silently
falls back to the default `DEFAULT` permission mode. This happens even though the caller
explicitly supplied a non-default permission context (for example `BYPASS`) that step 3 would
have preserved, and even though `permCtx` is already in scope at the call site.

Affected sessions: any session that still has v1-era keys but no new-format `agent_state` yet —
typically the first turn after a 1.x → 2.0 migration. The caller-configured permission mode is
silently downgraded, weakening permission semantics.

## Change

- `LegacyStateLoader.loadFromLegacySession(...)` / `loadFromLegacySessionWithPresence(...)`:
  new overloads accepting a `PermissionContextState`; when non-null it is attached to the
  reconstructed `AgentState`. The existing overloads are kept and delegate with `null`, so
  behaviour for callers without a specific permission context is unchanged.
- `ReActAgent.loadOrCreateAgentStateForSlot(...)`: pass the in-scope `permCtx` into the legacy
  load, exactly like `freshState` already does.

## Tests

`LegacyStateLoaderTest` (5 tests):

- supplied permission context is preserved (presence + non-presence variants);
- no permission context keeps the default mode (old overload behaviour);
- legacy messages and tool activation groups survive reconstruction alongside the permission
  context;
- no legacy keys reports `found=false`.

## Checklist

- [x] Conventional Commit message
- [x] Unit tests included
- [x] Backward compatible, no API breakage (old overloads kept)
- [ ] Full build verified locally (blocked by sandbox; CI runs `mvn clean verify` on Linux + Windows)

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -430,7 +430,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
             }
             LegacyStateLoader.LegacyLoadResult legacy =
                     LegacyStateLoader.loadFromLegacySessionWithPresence(
-                            stateStore, userId, sessionId);
+                            stateStore, userId, sessionId, permCtx);
             if (legacy.found()) {
                 // Legacy keys have no version; treat as create-if-absent baseline.
                 long version = stateStore.supportsVersioning() ? 0L : AgentStateStore.UNVERSIONED;
--- a/agentscope-core/src/main/java/io/agentscope/core/state/LegacyStateLoader.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/state/LegacyStateLoader.java
@@ -16,6 +16,7 @@
 package io.agentscope.core.state;
 
 import io.agentscope.core.message.Msg;
+import io.agentscope.core.permission.PermissionContextState;
 import io.agentscope.core.state.legacy.ToolkitState;
 import java.util.List;
 import java.util.Optional;
@@ -27,6 +28,11 @@ import java.util.Optional;
  * {@code toolkit_activeGroups}) and constructs an equivalent {@link AgentState} object. The
  * original session data is not modified or migrated in place; subsequent saves will use the new
  * format automatically.
+ *
+ * <p>Legacy keys carry no 2.0-era {@link PermissionContextState}, so callers that need a specific
+ * permission context (for example {@code BYPASS} for a pre-seeded privileged session) must pass it
+ * in explicitly — otherwise the reconstructed state silently falls back to the default
+ * permission mode.
  */
 public final class LegacyStateLoader {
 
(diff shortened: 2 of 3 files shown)


dailingtao:
LGTM. The legacy v1 session loading path now correctly preserves the caller-supplied permission context while keeping the existing overload behavior backward compatible. The tests cover permission propagation, default behavior, and preservation of legacy session data.

dailingtao (approved):

jujn (approved):

PR #2775, opened 2026-08-19
fix(harness): keep persisted snapshot id when resume falls back to fresh create

### LINKED ISSUE #2771: [Bug]: Kubernetes sandbox workspace never survives across calls — per-call pod termination makes resume fall back to fresh create and discard the persisted snapshot id
**Describe the bug**

With the stock `KubernetesSandboxClient` (Mode 2 `filesystem(SandboxFilesystemSpec)`, a `SandboxSnapshotSpec` and a state store configured), workspace contents are silently lost at the end of every call, and each call orphans the previous snapshot archive.

Trigger chain:

1. `KubernetesSandboxClient.create()` sets `claimOwned = true`. `HarnessAgent` auto-registers `SandboxLifecycleMiddleware`, which releases the sandbox after **every** call: `SandboxManager.release()` (self-managed path) → `stop()` (uploads the workspace archive under `snapshot.getId()`) → `shutdown()`. `KubernetesSandbox.shutdown()` sees `claimOwned == true` and calls `terminate()`, deleting the claim and the pod — every call.
2. On the next call, `SandboxManager.acquire()` Priority 3 loads the persisted state and calls `KubernetesSandboxClient.resume()` → `getSandbox(claimName, namespace)`, which throws because the claim was terminated in step 1. `acquire()` catches this, logs a warn, and falls through to Priority 4 fresh create.
3. Priority 4 creates a sandbox with a **new** sessionId, so `snapshotSpec.build(newSessionId)` produces a **new** snapshot id. The previous archive is never referenced again (orphaned), and `start()` runs the fresh-init branch (workspace rebuilt from spec only) — all files written in previous calls are gone. The only trace is the Priority 3 fallback warn log; the agent call itself succeeds.

This contradicts the documented 4-branch workspace-start logic in `AbstractBaseSandbox.start()`, whose Branch B/C exist precisely to restore the workspace from the snapshot once the container is gone. On Docker this works: `DockerSandboxClient.resume()` does not require a live container, so the next call reaches Branch B and restores the workspace. On Kubernetes, the per-call claim termination in step 1 makes those restore branches unreachable, and the Priority 4 fallback drops the persisted snapshot id.

**To Reproduce**

1. Build a `HarnessAgent` with `filesystem(new KubernetesFilesystemSpec(...).snapshotSpec(new RemoteSnapshotSpec(client)) ...)` and a distributed state store.
2. First call: write a file into the sandbox workspace. Observe the archive uploaded to remote storage and the claim/pod terminated at release.
3. Second call with the same scope: observe in the logs `Failed to load persisted state for scope ..., falling through to fresh create`, a new pod with a new sessionId, an empty workspace, and the previous archive orphaned in remote storage.

**Expected behavior**

The workspace persists across calls via the snapshot mechanism, matching the Docker behavior. Either:

- `resume()` should not hard-fail when the claim is gone, letting `start()` degrade to the Branch B restore path, or
- the fresh-create fallback should seed the new sandbox's snapshot with the persisted snapshot id so the workspace is restored from the previous archive.

**Error messages**

```
[sandbox] Failed to load persisted state for scope ..., falling through to fresh create: ...
```

Then a silent workspace reset — no further error.

- AgentScope-Java Version: 2.0.1 (also verified on current main `bf7b7dae`)
- Java Version: 21
- OS: any
- Module: agentscope-extensions-sandbox-kubernetes

**Additional context**

Workaround in use locally: subclass `KubernetesSandboxClient` and flip `claimOwned` to `false` right after `create()`, so the per-call `shutdown()` only closes the connection and the pod survives across calls; eviction deletes the pod explicitly. This confirms the trigger chain above but is fragile — it depends on overriding lifecycle ownership semantics that the harness assumes it controls.

Buktal:
## AgentScope-Java Version

2.0.1 (fix developed against current `main`, `bf7b7dae`)

## Description

Fixes #2771.

**Background.** With a harness-managed sandbox (`filesystem(SandboxFilesystemSpec)`) backed by the stock `KubernetesSandboxClient`, the workspace silently resets on every call and each snapshot archive is orphaned:

1. `create()` sets `claimOwned=true`, so the per-call release (`stop()` → `shutdown()` → `terminate()`) deletes the claim/pod at the end of every call.
2. On the next call, Priority 3 resume throws (`getSandbox(claimName)` fails — the claim is gone) and `SandboxManager.acquire()` falls through to the Priority 4 fresh create.
3. The fresh create uses a new sessionId, so `snapshotSpec.build(newSessionId)` produces a **new** snapshot id: the previous archive is never referenced again, and `start()` runs the fresh-init branch — all files written in previous calls are gone, with only the P3 fallback warn as a trace.

The 4-branch workspace-start logic (Branch B/C restore) exists for exactly this situation but is unreachable on Kubernetes, because the resume never succeeds. Docker works because its `resume()` does not require a live container.

**Change.** In `SandboxManager.acquire()`, when a persisted state was loaded but the resume failed, the fresh-create fallback now carries the **persisted snapshot id** over to the new sandbox (rebuilding the snapshot from the current spec, so it carries a bound client regardless of deserialization). The next `start()` then restores the workspace from the previous archive (Branch C), and the next `stop()` overwrites that same archive instead of orphaning it.

The fix is client-agnostic: it applies to any `SandboxClient` whose resume can fail after its backing runtime is gone.

**How to test.** Two new unit tests in `SandboxManagerIsolationTest`:
- `priority3_resumeFails_freshCreateCarriesOverPersistedSnapshotId` — resume throws → fresh create re-seeds `state.snapshot` via `snapshotSpec.build(persistedId)`
- `priority3_resumeFails_noPersistedSnapshot_keepsFreshSnapshot` — no persisted snapshot → fresh snapshot untouched

Full harness module suite: `mvn -pl agentscope-harness test` — all passing.

## Checklist

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxManager.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxManager.java
@@ -16,6 +16,7 @@
 package io.agentscope.harness.agent.sandbox;
 
 import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.harness.agent.sandbox.snapshot.SandboxSnapshotSpec;
 import java.util.Objects;
 import java.util.Optional;
 import org.slf4j.Logger;
@@ -26,7 +27,9 @@ import org.slf4j.LoggerFactory;
  *
  * <p>Acquire priority: {@link SandboxContext#getExternalSandbox()} &gt; {@link
  * SandboxContext#getExternalSandboxState()} &gt; persisted {@link SandboxState} &gt; {@link
- * SandboxClient#create}.
+ * SandboxClient#create}. When a persisted state exists but the resume fails, the fresh-create
+ * fallback keeps the persisted snapshot id so the workspace restores from the previous archive
+ * instead of silently resetting.
  *
  * <p>When a {@link SandboxExecutionGuard} is configured, the manager acquires an execution
  * {@link SandboxLease} before sandbox resume/create for isolation keys that are present. The
@@ -95,6 +98,7 @@ public class SandboxManager {
         }
 
         try {
+            String persistedSnapshotId = null;
             if (scopeKey.isPresent()) {
                 try {
                     Optional<String> stateJson = stateStore.load(scopeKey.get());
@@ -108,6 +112,9 @@ public class SandboxManager {
                         // Overwrite stale WorkspaceSpec with current application config
                         if (sandboxContext.getWorkspaceSpec() != null) {
                             state.setWorkspaceSpec(sandboxContext.getWorkspaceSpec().copy());
+                        }
+                        if (state.getSnapshot() != null) {
+                            persistedSnapshotId = state.getSnapshot().getId();
                         }
                         Sandbox sandbox = client.resume(state);
                         return SandboxAcquireResult.selfManaged(sandbox, lease);
@@ -136,6 +143,8 @@ public class SandboxManager {
                             spec,
                             sandboxContext.getSnapshotSpec(),
                             sandboxContext.getClientOptions());
+            carryOverPersistedSnapshotId(
+                    sandbox, persistedSnapshotId, sandboxContext.getSnapshotSpec());
             return SandboxAcquireResult.selfManaged(sandbox, lease);
 
         } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxManager.java
```diff
@@ -109,6 +113,9 @@ public SandboxAcquireResult acquire(
                         if (sandboxContext.getWorkspaceSpec() != null) {
                             state.setWorkspaceSpec(sandboxContext.getWorkspaceSpec().copy());
                         }
+                        if (state.getSnapshot() != null) {
+                            persistedSnapshotId = state.getSnapshot().getId();
+                        }
                         Sandbox sandbox = client.resume(state);
```
zouyx, line 119: 如果这里执行失败，被 try/catch 之后，persistedSnapshotId 这个 id 被初始化了，我看后续会用到。这样是否会有影响？

zouyx (changes requested):

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxManager.java
Buktal: 1. carryOverPersistedSnapshotId 方法的入参有 sandbox 跟 sandboxContext，这两部分不会产生 NPE
2. persistedSnapshotId 主要用来记录之前的快照ID，用来恢复沙箱快照文件；即使某种原因导致 sandbox resume 失败，重新创建了新的 sandbox，也要旧 snapshot ID 设置到新 Sandbox，这样工作区的文件才能尽可能地恢复之前的数据

辛苦再确认下哈

jujn (approved):

zouyx (approved):

Issue #2682, opened 2026-08-13
MemoryFlushMiddleware delays agent event stream completion, causing perceived latency after reply


chcodex:
### Description
After the agent finishes streaming its reply, the event stream stays open for several seconds. SSE/HTTP consumers keep the "generating" state until the stream completes. Root cause: `MemoryFlushMiddleware.onAgent` appends a **synchronous LLM extraction call** to the agent event stream via `.concatWith()`:

```java
return next.apply(input)
    .concatWith(
        Mono.defer(() -> doFlush(agent, rc))
            .subscribeOn(Schedulers.boundedElastic())
            ...
```

`concatWith` semantics require the appended `flushMemories` (an extra `model.stream` call in `MemoryFlushManager`) to finish before the main stream completes. `subscribeOn(boundedElastic)` only moves the work to another thread — it does not shorten the completion wait.

Secondary impact: `HarnessGateway.withGatedStream` serialises turns per `gateKey`; while a flush is pending, the next user message in the same session is blocked until it completes.

### Repro
- Default `MemoryConfig.FlushTrigger.always()` (flush after every call).
- Observe the gap between the last text event and stream `onComplete` / SSE connection close ≈ duration of one extra LLM round-trip.

### Expected
The event stream should complete immediately after the last agent event; memory extraction should run in the background (e.g. fire-and-forget on a daemon executor, like `SkillCuratorMiddleware`), with pending flushes drained on `HarnessAgent.close()` so no memory is lost on shutdown.

### Related
- `agentscope-harness/.../middleware/MemoryFlushMiddleware.java`
- `agentscope-harness/.../memory/MemoryFlushManager.java` (`flushMemories`)
- `MemoryMaintenanceMiddleware` has the same pattern (and a production `.block()`) — may warrant a follow-up.

dangzitou:
I would like to work on this issue. I plan to move the memory extraction off the user-visible stream, serialize background flushes to protect shared memory files, and add a regression test proving the agent stream completes before a slow flush. I will also add a graceful shutdown drain so pending flushes are not abandoned.

chcodex:
> I would like to work on this issue. I plan to move the memory extraction off the user-visible stream, serialize background flushes to protect shared memory files, and add a regression test proving the agent stream completes before a slow flush. I will also add a graceful shutdown drain so pending flushes are not abandoned.

https://github.com/agentscope-ai/agentscope-java/pull/2777 
Look here, it seems my issue submission was duplicated.

PR #2777, opened 2026-08-19
fix(harness): make memory flush fire-and-forget to unblock conversation completion

### LINKED ISSUE #2774: MemoryFlushMiddleware flush blocks the conversation completion signal
## Description

`MemoryFlushManager.flushMemories()` performs a full LLM extraction call plus disk writes, and it currently blocks the end of the agent's reply stream.

## Location

`agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java` (lines 143-152):

```java
return next.apply(input)
        .concatWith(
                        .onErrorResume(e -> { ... })
                        .then(Mono.<AgentEvent>empty()));
```

## Problem

`concatWith` chains the flush strictly after the main response stream. Although `subscribeOn(Schedulers.boundedElastic())` moves the flush work off the event-loop thread, it does **not** change the timing: the downstream `onComplete` signal is only emitted after `flushMemories` finishes, i.e. after:

1. reading MEMORY.md + today's daily ledger,
2. one additional LLM extraction call via `model.stream(...)`,
3. writing the daily `*.md` file.

So every call's completion is delayed by an extra synchronous LLM round-trip, adding visible tail latency to every user turn (default `FlushMode.ALWAYS`).

## Expected behaviour

The memory flush should be fully asynchronous/fire-and-forget w.r.t. the conversation stream — e.g. dispatched on a separate scheduler without being composed into the reply Flux (possibly bounded/rate-limited to avoid unbounded pending flushes), so the response completes independently of the flush.

## Environment

- Module: agentscope-harness
- Default config: `FlushTrigger.always()`

chcodex:
## Summary

- `MemoryFlushMiddleware` used `concatWith` to chain the memory flush after the agent stream, so the downstream `onComplete` was delayed by a full LLM extraction round-trip plus disk writes on every call (default `FlushMode.ALWAYS`). Switched to `doOnComplete` + `subscribeOn(boundedElastic()).subscribe()` so the flush runs in the background and the conversation stream completes immediately.
- Snapshot the conversation list (`new ArrayList<>(state.getContext())`) before the async flush to avoid concurrent modification if the next agent call clears the state while the flush is still running.
- Applied the same fire-and-forget pattern to `MemoryMaintenanceMiddleware`, which had the identical `concatWith` structure.
- Added `MemoryFlushMiddlewareAsyncFlushTest` to verify the agent stream completes before a never-finishing flush does, and that the flush is still triggered.

## Test plan

- [x] New `MemoryFlushMiddlewareAsyncFlushTest` passes (verifies stream completes before flush + flush still triggered).
- [x] Existing `MemoryFlushMiddlewareTriggerTest` (17 tests) passes.
- [x] Existing `MemoryMaintenanceMiddlewareStaleEntryTest` (3 tests) passes.
- [x] Full `agentscope-harness` module test suite passes (831 tests, 0 failures).

Closes #2774

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
@@ -457,6 +457,10 @@ public class HarnessAgent implements Agent, AutoCloseable {
             // race with resource cleanup (e.g., temp workspace deletion in tests).
             io.agentscope.harness.agent.memory.session.SessionTree.awaitMirrorQuiescence(
                     5, java.util.concurrent.TimeUnit.SECONDS);
+            // Drain fire-and-forget memory flush/maintenance so async memory/*.md writes do not
+            // race with resource cleanup (e.g., temp workspace deletion in tests).
+            io.agentscope.harness.agent.memory.MemoryBackgroundTasks.awaitQuiescence(
+                    5, java.util.concurrent.TimeUnit.SECONDS);
             shutdownTaskRepository();
         } finally {
             try {
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryBackgroundTasks.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryBackgroundTasks.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent.memory;
+
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Tracks in-flight fire-and-forget memory background tasks (flush, maintenance) so
+ * {@code HarnessAgent.close()} can wait for them to quiesce before releasing resources.
+ *
+ * <p>The middleware instances that dispatch these tasks are created per agent call, so they
+ * cannot be reached from the agent's {@code close()} method. Like
+ * {@link io.agentscope.harness.agent.memory.session.SessionTree#awaitMirrorQuiescence}, the
+ * in-flight count is process-wide: {@code close()} blocks until every background task started
+ * before the call has finished, so async workspace writes do not race with resource cleanup
+ * (e.g. temp workspace deletion in tests).
+ */
+public final class MemoryBackgroundTasks {
+
+    private static final Logger log = LoggerFactory.getLogger(MemoryBackgroundTasks.class);
+
+    private static final Object MONITOR = new Object();
+    private static int inFlight = 0;
+
+    private MemoryBackgroundTasks() {}
+
+    /** Marks the start of a fire-and-forget background task. */
+    public static void begin() {
+        synchronized (MONITOR) {
+            inFlight++;
+        }
+    }
+
+    /** Marks the end of a fire-and-forget background task (must pair with {@link #begin()}). */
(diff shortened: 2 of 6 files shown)


chcodex:
## Update: graceful shutdown for the async flush (commit `5d99f005`)

The first version made the flush fire-and-forget but did not handle shutdown: `HarnessAgent.close()` could release the workspace while the background flush was still writing `memory/*.md`, racing with `@TempDir` deletion in tests (intermittent `Failed to delete temp directory` under a full test run).

### What changed

- **New `MemoryBackgroundTasks`** (`io.agentscope.harness.agent.memory`): process-wide in-flight counter with `begin()/end()` and `awaitQuiescence(timeout, unit)`. The middleware instances that dispatch the flush are created per agent call, so they are not reachable from `close()` — the counter is process-wide for the same reason `SessionTree.awaitMirrorQuiescence` already is.
- **`MemoryFlushMiddleware` / `MemoryMaintenanceMiddleware`**: the async task chain now wraps `doOnSubscribe(begin())` / `doFinally(end())` so the counter is always paired (success, error, and cancel all release it).
- **`HarnessAgent.close()`**: waits for memory-task quiescence (5s) before releasing resources, mirroring the existing `SessionTree.awaitMirrorQuiescence` call directly above it.
- **`MemoryFlushMiddlewareAsyncFlushTest`**: previously used `Flux.never()`, which would hold the in-flight counter open forever and force every subsequent `close()` to wait the full 5s timeout. The test now *gates* the flush stream behind a latch, asserts the agent stream completes while the flush is in-flight, then releases the gate and asserts the tracker quiesces back to zero.

### Why this is safe

- The non-blocking behavior from the original fix is preserved: the conversation stream still completes immediately; `close()` is the only place that waits, and only for tasks that were already running.
- The 5s cap means a genuinely stuck flush cannot hang shutdown indefinitely (same timeout as the session-tree drain).

### Verification

- `LocalFilesystemPersonalAssistantExampleTest` — 10 consecutive full runs, all green (was ~50% flaky before).
- Full `agentscope-harness` suite — 831 tests, 0 failures.

dangzitou:
Thanks for taking this on. I had started the same fix from issue #2682, but your PR predates my branch and already covers the stream completion, background-task draining, and maintenance path with passing CI. I’m closing my duplicate PR # to avoid fragmenting review.

Buktal (changes requested):
1. 考虑按 isolation key 加一下并发控制（比如 putIfAbsent + doFinally）：主要防止同会话在连续请求或并发时，多个 flush 并行跑导致往 MEMORY.md 里写入重复记忆
2. 建议把 doFlush 整体用 Mono.defer 包一层再 subscribeOn：避免在 onComplete 的回调线程（比如 Netty 事件循环线程）上同步做上下文序列化和文件读取，完全沉到 boundedElastic 去跑

chcodex:
@Buktal Both suggestions are addressed in commit `66f8f9bb`:

**1. Per-isolation-key concurrency control** — flushes now go through a process-wide `ConcurrentHashMap<String, FlushQueue>` keyed by the same `compositeTimerKey` used for throttling (`isolationScope:userId/sessionId`). `compute` atomically claims the slot: if a flush for the key is already running, the new request is queued (FIFO); when the running flush terminates, `doFinally` drains the queue. This prevents the back-to-back double-extract that could previously write overlapping memories (the second flush used to read the dedup context before the first had appended to it). Queued tasks re-snapshot the conversation at execution time — a superset of the blocked predecessor's window — so nothing is lost, and idle keys are evicted from the map. Covered by a new test `flushesForSameIsolationKeyRunSerially` (asserts the second flush never starts while the first is in flight, and observed max concurrency is 1).

**2. `Mono.defer` around the whole `doFlush`** — you're right, the fire-and-forget rework had dropped the `defer`, so the state snapshot, message serialization and ledger reads ran synchronously on the completing thread. `runFlush` now does `Mono.defer(() -> doFlush(...)).subscribeOn(boundedElastic())`, so every blocking step runs on the background scheduler. `MemoryBackgroundTasks.begin()` fires synchronously at scheduling time (before the hand-off, so `close()` cannot race past a scheduled-but-not-yet-started flush; queued tasks stay counted until they actually finish) and `end()` in `doFinally`.

Note the previous "track only when there is actual work" tweak was intentionally reverted: with a queue, a scheduled task's no-op-ness is unknown until it runs, and close() must drain queued tasks too — so counting happens per scheduling instead, at the cost of one trivial synchronized pair per call.

Rebased onto latest main; full `agentscope-harness` suite green (850 tests, 0 failures).

Buktal (commented):
(Withdrawn — will push the fixes directly on the branch instead.)

Buktal (approved):
LGTM

PR #2796, opened 2026-08-21
fix(core): make JsonSchemaUtils schema generation thread-safe

### LINKED ISSUE #2795: [Bug]: JsonSchemaUtils shared SchemaGenerator is not thread-safe (ConcurrentModificationException under concurrent structured calls)
**Describe the bug**

`JsonSchemaUtils` initializes a single victools `SchemaGenerator` in a static
block (`agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java`,
lines 62–84) and shares it across the whole JVM **without any synchronization**.
When multiple threads perform structured calls targeting the same output class
(e.g. multiple agents running in parallel), `generateSchemaFromClass()` /
`generateSchemaFromType()` can fail with a `ConcurrentModificationException`
wrapped in a `RuntimeException`.

The failure is timing-sensitive and intermittent: in our production setup (two
expert agents reviewing in parallel), it triggered once across many runs — the
losing agent fails immediately at its first structured call and returns an
ERROR, while the other agent completes normally.

**To Reproduce**

1. Run two (or more) `ReActAgent` instances in parallel threads, both performing
   a structured call with the same target class:
   ```java
   // thread A
   agentA.call(messages, MyReport.class, context);
   // thread B, overlapping in time
   agentB.call(messages, MyReport.class, context);
   ```
2. Any overlap of two `doStructuredCall` invocations that both reach
   `JsonSchemaUtils.generateSchemaFromClass(...)` for a not-yet-introspected
   class races on the shared `SchemaGenerator`.
3. Occasionally one thread throws (see error below).

Note: the race window is narrow — two threads must be introspecting the same
not-yet-cached class at the same time. We could not reproduce it
deterministically locally (800 concurrent calls across two stress rounds stayed
green), which matches the intermittent production behavior. The race is still
real: there is no happens-before edge between concurrent calls.

**Expected behavior**

Schema generation from shared static state should be thread-safe (or the class
should be explicitly documented as not thread-safe). Concurrent structured
calls are a normal usage pattern for multi-agent applications.

**Error messages**

```
java.lang.RuntimeException: Failed to generate JSON schema for com.example.MyStructuredOutput
    at io.agentscope.core.util.JsonSchemaUtils.generateSchemaFromClass(JsonSchemaUtils.java:102)
    at io.agentscope.core.ReActAgent.doStructuredCall(ReActAgent.java:1054)
    ...
Caused by: java.util.ConcurrentModificationException
    at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1013)
    at com.fasterxml.jackson.databind.introspect.POJOPropertiesCollector._sortProperties(POJOPropertiesCollector.java:1868)
    at com.fasterxml.jackson.databind.introspect.POJOPropertiesCollector.collectAll(POJOPropertiesCollector.java:502)
    at com.fasterxml.jackson.databind.introspect.BasicBeanDescription.findProperties(BasicBeanDescription.java:236)
    at com.github.victools.jsonschema.module.jackson.JacksonModule.shouldIgnoreField(JacksonModule.java:306)
    at com.github.victools.jsonschema.generator.SchemaGeneratorConfigPart.lambda$shouldIgnore$0(SchemaGeneratorConfigPart.java:135)
```

(stack trace trimmed to the relevant frames)

- AgentScope-Java Version: 2.0.1 (also verified the code is unchanged on latest `main`, commit `0de9ed43`)
- Java Version: 17
- OS: macOS

**Additional context**

Root cause analysis:

1. **Shared singleton without synchronization** — `JsonSchemaUtils` holds a
   `private static final SchemaGenerator` (one instance per JVM); neither
   `generateSchemaFromClass` nor `generateSchemaFromType` guards the
   `schemaGenerator.generateSchema(...)` call.
2. **victools `SchemaGenerator` is not thread-safe** — a known limitation of
   victools/jsonschema-generator. In `jsonschema-module-jackson` 4.38.0,
   `JacksonModule` keeps an instance-level `Map<Class, BeanDescription>`
   introspection cache (a plain `HashMap`, no concurrency protection).
   Concurrent `getBeanDescriptionForClass` calls can share one
   `BeanDescription`, and two threads racing on its lazily-populated
   `POJOPropertiesCollector` (one iterating in `_sortProperties` while the
   other inserts) throw the CME above.
3. **Wide trigger surface** — `ReActAgent.doStructuredCall` regenerates the
   schema on *every* structured call (no caching), and the
   `call(msgs, Class, ctx)` entry always goes through the shared generator.
   So any two overlapping structured calls — not just agent startup, any ReAct
   iteration — form a data race. CME is only the most visible symptom; without
   a happens-before edge, the shared HashMap cache can also silently lose
   entries or get structurally corrupted.

The same class of bug was reported and fixed in Spring AI:
[spring-ai#6207](https://github.com/spring-projects/spring-ai/issues/6207).
The Spring AI team resolved it by guarding every `SchemaGenerator.generateSchema(...)`
call with `synchronized` (commit `9b8eb07`, "Guard JSON schema generation"),
plus concurrent regression tests.

Suggested fix: apply the same minimal guard in `JsonSchemaUtils` — wrap the two
`schemaGenerator.generateSchema(...)` call sites in a `synchronized` block
(a dedicated lock object, or `synchronized (schemaGenerator)` as Spring AI did).
This keeps the public API unchanged. A schema cache could be a separate
enhancement later.

Happy to submit a PR for this.

jnduan:
## AgentScope-Java Version

2.0.3-SNAPSHOT (latest `main`, commit `0de9ed43`)

## Description

Fixes #2795

`JsonSchemaUtils` shares a single static victools `SchemaGenerator` across the
JVM without synchronization. victools' `SchemaGenerator` is not thread-safe
(its `JacksonModule` keeps an unsynchronized `Map<Class, BeanDescription>`
introspection cache), so concurrent structured calls — e.g. multiple agents
running in parallel with the same target class — race on the shared generator
and can fail with a `ConcurrentModificationException`. The linked issue has the
full root-cause analysis, including the identical bug fixed in Spring AI
(spring-ai#6207, resolved with the same synchronized-guard approach).

Changes:

- Added a private `SCHEMA_LOCK` and guarded the two
  `schemaGenerator.generateSchema(...)` call sites
  (`generateSchemaFromClass` / `generateSchemaFromType`). Public API is
  unchanged.
- Class Javadoc now documents the thread-safety guarantee.
- Added concurrent regression tests for both guarded methods: 12 threads ×
  240 tasks released simultaneously via `CountDownLatch`, alternating between
  multiple POJOs so the victools introspection cache starts cold. Any exception
  propagates through `Future#get` and fails the test; all results are
  validated.

Note: the pre-fix race window is narrow and cannot be reproduced
deterministically, so the tests are regression protection rather than a
reproducer.

`mvn -pl agentscope-core test`: 2277 tests, 0 failures.

## Checklist

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
@@ -53,6 +53,10 @@ import java.util.Map;
  *   <li>{@code @JsonClassDescription(...)} - add class description</li>
  * </ul>
  *
+ * <p>All public methods are thread-safe. Schema generation through the shared victools
+ * {@code SchemaGenerator} is serialized by an internal lock, because the generator itself
+ * is not designed for concurrent use.</p>
+ *
  * @hidden
  */
 public class JsonSchemaUtils {
@@ -60,6 +64,13 @@ public class JsonSchemaUtils {
     private static final boolean PROPERTY_REQUIRED_BY_DEFAULT = false;
 
     private static final SchemaGenerator schemaGenerator;
+
+    /**
+     * Guards the shared victools {@link SchemaGenerator}, which is not thread-safe: its
+     * JacksonModule keeps an unsynchronized introspection cache, so concurrent schema
+     * generation must be serialized.
+     */
+    private static final Object SCHEMA_LOCK = new Object();
 
     static {
         // JacksonModule to support @JsonProperty, @JsonPropertyDescription annotations
@@ -95,7 +106,10 @@ public class JsonSchemaUtils {
      */
     public static Map<String, Object> generateSchemaFromClass(Class<?> clazz) {
         try {
-            JsonNode schemaNode = schemaGenerator.generateSchema(clazz);
+            JsonNode schemaNode;
+            synchronized (SCHEMA_LOCK) {
+                schemaNode = schemaGenerator.generateSchema(clazz);
+            }
             return JsonUtils.getJsonCodec()
                     .convertValue(schemaNode, new TypeReference<Map<String, Object>>() {});
         } catch (Exception e) {
@@ -130,7 +144,10 @@ public class JsonSchemaUtils {
      */
     public static Map<String, Object> generateSchemaFromType(Type type) {
         try {
-            JsonNode schemaNode = schemaGenerator.generateSchema(type);
+            JsonNode schemaNode;
+            synchronized (SCHEMA_LOCK) {
+                schemaNode = schemaGenerator.generateSchema(type);
+            }
             return JsonUtils.getJsonCodec()
                     .convertValue(schemaNode, new TypeReference<Map<String, Object>>() {});
         } catch (Exception e) {
--- a/agentscope-core/src/test/java/io/agentscope/core/util/JsonSchemaUtilsTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/util/JsonSchemaUtilsTest.java
@@ -23,11 +23,22 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.function.IntFunction;
 import org.junit.jupiter.api.Test;
 
 class JsonSchemaUtilsTest {
+
+    private static final int CONCURRENT_THREAD_COUNT = 12;
+
+    private static final int CONCURRENT_CALL_COUNT = 240;
 
     static class SimpleModel {
         public String name;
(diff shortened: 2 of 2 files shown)


jnduan:
The failing check is unrelated to this PR. I dug into the job log:

- The error comes from `agentscope-harness`, in `JsonSessionDefaultLocationTest.perUserPartitioning_viaSharedAgentRoutedByRuntimeContext`
- The test's own assertions actually passed; it failed in JUnit's `@TempDir` teardown: `Failed to delete temp directory ... DirectoryNotEmptyException: .../bob` — a cleanup race where an async session-store write lands after teardown starts deleting
- Neither this PR (`agentscope-core`/`JsonSchemaUtils`) nor the harness module reference each other; locally `agentscope-core` (2,282 tests) and `agentscope-harness` (833 tests) both pass on the rebased head `8882daa8`

Looks like a flaky test. Could someone with permissions re-run the failed jobs? Thanks!

skrcode:
I ran JAIPilot Cloud against this exact PR head. It found one follow-up: cache generated JSON schemas by Class/Type so repeated structured-output and tool schemas do not serialize on SCHEMA_LOCK and rerun reflective generation.

In the managed sandbox, the identical 16-thread, 80,000-call workload over two repeated classes moved from a 20,323 ms median to 899 ms across five observations per side. Four behavior locks preserve schema equality, independent returned-map mutation, and null handling; downstream structured-output tests and mvn -pl agentscope-core -am verify passed.

PR directly onto this source branch: https://github.com/jnduan/agentscope-java/pull/1  
Cloud-generated draft and full evidence: https://github.com/skrcode/agentscope-java/pull/2

The cache retains one entry per encountered Class/Type, so this is optional and the lifecycle assumption should be reviewed for dynamic class-loading environments. No absolute production-latency claim is intended.

jnduan:
Thanks for the analysis! Schema caching is indeed the planned follow-up — this PR was deliberately kept to the minimal synchronization fix (single concern), and caching will be submitted as a separate enhancement once this lands. The per-class lifecycle note is a good point to factor into that work.

jujn (approved):

PR #2797, opened 2026-08-21
fix(harness): recover lost async task completions in legacy wait

### LINKED ISSUE #2791: [Bug]: WaitAsyncResultsTool waits on the inbox notification chain instead of task state, stalling the main agent when any link breaks
**Describe the bug**

`WaitAsyncResultsTool` builds its waiting semantics on an indirect, multi-hop chain rather than
observing task state directly:

```
subtask completes -> TaskRepository completionCallback -> MessageBus inbox push
                  -> InboxMiddleware injects before next LLM call -> LLM sees the result
```

The tool itself only polls the inbox every 3s. If any link in that chain fails to deliver, the
tool never observes the fact that the task has actually finished — and it never cross-checks the
real task status in `TaskRepository`.

Combined with the `MAX_CONSECUTIVE_EMPTY_WAITS` rejection, the consequence escalates from "waits
too long" to "hard stall": after two consecutive empty waits the tool refuses to wait any longer,
while the main agent has neither received results nor been given any other way to make progress.

**To Reproduce**

1. You code

```java
HarnessAgent supervisor = HarnessAgent.builder()
        .model(model)
        .messageBus(messageBus)
        .subagentFactory("worker", name -> workerAgent)
        .build();

// Spawn one or more async subagents (background task mode), then let the supervisor
// call wait_async_results to collect their results.
supervisor.streamEvents(new UserMessage(promptThatSpawnsThenWaits), ctx).subscribe();
```

2. How to execute

Run the flow, and make any one link of the injection chain fail. Any of these is enough:

- another `WorkspaceTaskRepository.setCompletionCallback` call replaces the callback, so the
  inbox push no longer happens (later registration overwrites the earlier one);
- the task reaches a terminal state between `putTask` and callback registration (race), so the
  completion event is lost;
- the inbox message is consumed but injection lands after the round that needed it (injector race).

Then compare the subagent's real state with what the tool reports.

3. See error

`TaskRepository` shows the task as `COMPLETED`, the inbox is empty, and the tool keeps waiting
until timeout, then refuses to wait again. The LLM never receives the results.

**Expected behavior**

Once a task has actually reached a terminal state, `wait_async_results` should observe it and hand
the result to the LLM. Even if the inbox notification side-channel fails, the tool must not treat
"notification missing" as "task not finished".

**Error messages**

No exception is raised. The failure is silent, so the observable symptoms are:

- the tool blocks for the full 60s timeout although the task is already terminal;
- after two consecutive empty waits, waiting is refused via `MAX_CONSECUTIVE_EMPTY_WAITS` and the
  conversation cannot proceed;
- when the inbox is empty but all tasks are terminal, the tool returns a promise such as
  `results will be injected automatically`, which is never fulfilled in this scenario and actively
  misleads the LLM into waiting further.

- AgentScope-Java Version: 2.0.0 (agentscope-harness 2.0.0)
- Java Version: 17
- OS: macos

**Additional context**

*Root cause.* The tool's wait condition is not the same thing as the semantics it means to express:

| | Intended semantics | What is actually observed |
|---|---|---|
| Goal | whether tasks reached a terminal state | task status in `TaskRepository` |
| Actual | whether the inbox has a message | a product of the notification chain |

Task status is the **authoritative fact** and is guaranteed to happen (unless the task itself
hangs). The inbox message is only a **derived notification** with several failure points along the
way. Observing the derived signal without checking the authoritative one makes "lost notification"
indistinguishable from "task not finished".

`MAX_CONSECUTIVE_EMPTY_WAITS` amplifies the defect. It assumes repeated empty waits mean the LLM
is abusing the wait, but when the notification chain is broken, repeated empty waits are a symptom
of the tool's own faulty observation — refusing to wait at that moment only pushes the LLM into a
dead end.

*Suggested fix.* Change the wait condition to poll `TaskRepository` for terminal task state
directly. `TaskRepository.listTasks(ctx, sessionId, null)` already returns all tasks with their
`TaskStatus`, and `TaskStatus.isTerminal()` already exists, so no new API is required. Once tasks
are terminal, return the result text as the tool's return value so result collection closes within
a single `call()` and no longer depends on cross-round injection. Inbox injection can stay as a
fast path, but should not be the only correctness dependency.

Secondary points:

1. Remove or reposition `MAX_CONSECUTIVE_EMPTY_WAITS`. If the wait observes real task state, a
   terminal state is guaranteed to arrive and the rejection is meaningless; if the intent is to
   guard against hung tasks, the timeout itself is the better mechanism.
2. When the inbox is empty and all tasks are terminal, do not return
   `will be injected automatically`. Return the actual results, or state the current status
   explicitly, instead of making a promise that will not be kept.

Affected classes:

- `io.agentscope.harness.agent.tool.WaitAsyncResultsTool`
- `io.agentscope.harness.agent.subagent.task.WorkspaceTaskRepository` (completion callback registration)

Daihonghui:
## Summary

- snapshot non-terminal async tasks when legacy `wait_async_results` starts
- keep inbox notifications as the first fast path
- fall back to authoritative `TaskRepository` state when a tracked task becomes terminal
- return the completed task's actual result/error and mark it delivered

## Compatibility

- preserves legacy ANY-result semantics: one tracked completion returns immediately
- excludes tasks already terminal before wait start
- leaves `task_ids` and `wait_all=true` barrier semantics unchanged
- preserves inbox fast-path and timeout behavior

## Tests

- reproduced the bug before the production change: the core regression timed out and failed
- `WaitAsyncResultsToolTest`: 23 tests passed
- `WaitAsyncResultsToolTest`, `WorkspaceTaskRepositoryTest`, and `WorkspaceTaskRepositoryDeliveryTest`: 54 tests passed
- Spotless check passed

Fixes #2791

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/WaitAsyncResultsTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/WaitAsyncResultsTool.java
@@ -32,14 +32,14 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Tool that blocks until async results arrive, either for a specific task barrier or (legacy)
- * until any message appears in the session inbox.
+ * until any inbox message arrives or any task running at wait start reaches terminal state.
  *
  * <p>Prefer barrier mode ({@code task_ids} or {@code wait_all=true}): when the wait set reaches
  * terminal state, this tool embeds each task's status and result in the tool return so the model
  * can continue reasoning immediately without waiting for push-back delivery.
  *
- * <p>Without {@code task_ids} / {@code wait_all}, this keeps the legacy inbox-any behavior: the
- * tool returns as soon as <em>any</em> inbox message is present — it is not a wait-all barrier.
+ * <p>Without {@code task_ids} / {@code wait_all}, this keeps the legacy ANY-result behavior: the
+ * tool returns for <em>any</em> inbox message or tracked task completion, not a wait-all barrier.
  *
  * <p>Guardrails prevent unbounded blocking:
  * <ul>
@@ -103,9 +103,10 @@ public class WaitAsyncResultsTool {
                             + "wait_all=true waits for the snapshot of currently running tasks "
                             + "(tasks created while waiting are not added) and also returns their "
                             + "results. "
-                            + "Without task_ids and without wait_all, this is legacy inbox-any "
-                            + "mode: returns when ANY inbox message arrives — not wait-all. For "
-                            + "must-collect-all groups use task_ids or wait_all=true. "
+                            + "Without task_ids and without wait_all, this is legacy ANY-result "
+                            + "mode: returns when ANY inbox message arrives or any task running at "
+                            + "wait start becomes terminal — not wait-all. For must-collect-all "
+                            + "groups use task_ids or wait_all=true. "
                             + "Also covers AgentTeams teammate work via the external-work probe. "
                             + "Max timeout is 120 seconds. "
                             + "If you have already waited without results, use task_list or "
@@ -179,9 +180,10 @@ public class WaitAsyncResultsTool {
                     + "blocking.";
         }
 
+        List<String> legacyWaitSet = List.of();
         if (taskRepository != null) {
-            boolean hasNonTerminal =
-                    hasNonTerminalTasks(runtimeContext, sessionId) || hasExternalWork();
+            legacyWaitSet = snapshotNonTerminalTaskIds(runtimeContext, sessionId);
+            boolean hasNonTerminal = !legacyWaitSet.isEmpty() || hasExternalWork();
             if (!hasNonTerminal) {
                 Boolean hasMessages = messageBus.inboxHasMessages(sessionId).block();
                 if (!Boolean.TRUE.equals(hasMessages)) {
@@ -199,18 +201,14 @@ public class WaitAsyncResultsTool {
         int timeout = normalizeTimeout(timeoutSeconds, sessionId);
 
         log.info(
-                "wait_async_results: waiting up to {}s for any inbox message (legacy inbox-any),"
-                        + " session={}",
+                "wait_async_results: waiting up to {}s for any inbox message or tracked task"
+                        + " completion (legacy ANY-result), session={}",
                 timeout,
                 sessionId);
 
         long deadlineMs = System.currentTimeMillis() + (timeout * 1000L);
 
         while (true) {
-            long remainingMs = deadlineMs - System.currentTimeMillis();
-            if (remainingMs <= 0) {
-                break;
-            }
             Boolean hasMessages = messageBus.inboxHasMessages(sessionId).block();
             if (Boolean.TRUE.equals(hasMessages)) {
                 log.info("wait_async_results: inbox has messages, session={}", sessionId);
@@ -220,8 +218,22 @@ public class WaitAsyncResultsTool {
                         + "automatically. Prefer wait_async_results(task_ids=...) or "
                         + "wait_all=true when you need every task in a group.";
             }
+            BackgroundTask terminalTask =
+                    findFirstTerminalTrackedTask(runtimeContext, sessionId, legacyWaitSet);
+            if (terminalTask != null) {
+                log.info(
+                        "wait_async_results: tracked task {} is terminal, session={}",
+                        terminalTask.getTaskId(),
+                        sessionId);
+                emptyWaits.set(0);
+                return formatLegacyResult(runtimeContext, sessionId, terminalTask);
+            }
+            long remainingMs = deadlineMs - System.currentTimeMillis();
+            if (remainingMs <= 0) {
+                break;
+            }
             // Cap sleep to the remaining budget so the tool never overshoots the caller's timeout.
-            Thread.sleep(Math.min(POLL_INTERVAL_MS, remainingMs));
+            sleepForPollInterval(remainingMs);
         }
 
         int emptyCount = emptyWaits.incrementAndGet();
@@ -364,15 +376,52 @@ public class WaitAsyncResultsTool {
         return formatBarrierResults(runtimeContext, sessionId, terminalTasks);
     }
 
+    private BackgroundTask findFirstTerminalTrackedTask(
+            RuntimeContext runtimeContext, String sessionId, List<String> waitSet) {
+        if (waitSet.isEmpty()) {
+            return null;
+        }
+        Collection<BackgroundTask> currentTasks =
+                taskRepository.listTasks(runtimeContext, sessionId, null);
+        for (String taskId : waitSet) {
+            for (BackgroundTask task : currentTasks) {
+                if (taskId.equals(task.getTaskId()) && task.getTaskStatus().isTerminal()) {
+                    return task;
+                }
+            }
+        }
+        return null;
+    }
+
+    private String formatLegacyResult(
+            RuntimeContext runtimeContext, String sessionId, BackgroundTask task) {
+        return formatTerminalResults(
+                runtimeContext,
+                sessionId,
+                List.of(task),
+                "A tracked background task is terminal. Its result is included below.");
+    }
+
     /**
      * Embeds each terminal task's status/result in the tool return so the parent can continue
      * without waiting for inbox push-back. Marks tasks delivered to avoid duplicate reminders.
      */
     private String formatBarrierResults(
             RuntimeContext runtimeContext, String sessionId, List<BackgroundTask> tasks) {
+        return formatTerminalResults(
+                runtimeContext,
+                sessionId,
+                tasks,
+                "All requested background tasks are terminal. Results are included below.");
+    }
+
+    private String formatTerminalResults(
+            RuntimeContext runtimeContext,
+            String sessionId,
+            List<BackgroundTask> tasks,
+            String heading) {
         StringBuilder sb = new StringBuilder();
-        sb.append("All requested background tasks are terminal. Results are included below.")
-                .append('\n');
+        sb.append(heading).append('\n');
         for (BackgroundTask task : tasks) {
             task.updateLastCheckedAt();
             if (task.isCompleted() && task.getTaskStatus().isTerminal()) {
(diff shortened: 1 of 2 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/WaitAsyncResultsTool.java
```diff
@@ -179,9 +180,10 @@ public String waitForResults(
                     + "blocking.";
         }
 
+        List<String> legacyWaitSet = List.of();
         if (taskRepository != null) {
-            boolean hasNonTerminal =
-                    hasNonTerminalTasks(runtimeContext, sessionId) || hasExternalWork();
+            legacyWaitSet = snapshotNonTerminalTaskIds(runtimeContext, sessionId);
```
xiaoyuuuuuupeng, line 185: [P1] Recover terminal-but-undelivered tasks before taking this snapshot

This snapshot only includes tasks that are non-terminal when the legacy wait
starts. If a task completes before this call and its completion notification
is lost, it is excluded from legacyWaitSet and its result is not returned.
That leaves the callback-registration race described in #2791 unresolved.

Please check for terminal-but-undelivered tasks before building this snapshot,
using findPendingDeliveries or equivalent delivery metadata. Already-delivered
historical tasks should remain excluded. A regression test should start with
a terminal, undelivered task and an empty inbox, then verify that legacy wait
returns the result and marks it delivered.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/WaitAsyncResultsTool.java
```diff
@@ -220,8 +218,22 @@ public String waitForResults(
                         + "automatically. Prefer wait_async_results(task_ids=...) or "
                         + "wait_all=true when you need every task in a group.";
             }
+            BackgroundTask terminalTask =
+                    findFirstTerminalTrackedTask(runtimeContext, sessionId, legacyWaitSet);
```
xiaoyuuuuuupeng, line 222: [P2] Make repository recovery reachable after the wait budget is exhausted

This fallback is never reached once MAX_CONSECUTIVE_EMPTY_WAITS has been hit,
because the legacy path rejects earlier after checking only the inbox.

For example, two waits may time out while the task is still running; the task
then completes, its inbox notification is lost, and the next legacy wait is
rejected without consulting TaskRepository.

Please perform one terminal-but-undelivered repository check before applying
the exhausted-budget rejection. A regression test should cover completion
after the second timeout with an empty inbox.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/WaitAsyncResultsTool.java
```diff
@@ -179,9 +180,10 @@ public String waitForResults(
                     + "blocking.";
         }
 
+        List<String> legacyWaitSet = List.of();
```
xiaoyuuuuuupeng, line 183: [Suggestion] Check the inbox before querying TaskRepository

The initial non-terminal snapshot queries TaskRepository before the first
inbox check. With a remote repository, an already-available inbox result may
therefore be delayed; if listTasks throws, the call fails without observing
the inbox result.

Please check inboxHasMessages before taking the repository snapshot. Repository
fallback failures should also degrade to the original inbox-only behavior
rather than preventing it.

A small test with inbox=true and a repository whose listTasks throws would
verify the ordering and degradation behavior.

xiaoyuuuuuupeng (changes requested):
Thanks for addressing the notification-loss failure in the legacy
wait_async_results path. Falling back to authoritative TaskRepository state
is the right direction.

I found one case that still leaves a reproduction from #2791 unresolved:
a task that becomes terminal before legacy wait starts is excluded from the
non-terminal snapshot, even when its result has not been delivered. The tool
then reports that all tasks are complete without returning the lost result.

Please recover terminal-but-undelivered tasks before taking the non-terminal
snapshot. The existing findPendingDeliveries / delivery metadata appears to
provide the distinction needed to avoid returning already-delivered historical
tasks.

I also left two smaller comments:

- repository recovery is bypassed after the consecutive-wait budget is
  exhausted;
- the initial repository snapshot runs before the first inbox check, so the
  claimed inbox fast path can still be delayed or broken by repository
  failures.

The main repository fallback is a good improvement, but I think the
terminal-before-wait case should be addressed before this PR closes #2791.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/WaitAsyncResultsTool.java
```diff
@@ -364,15 +376,52 @@ private String completeTaskBarrierIfReady(
         return formatBarrierResults(runtimeContext, sessionId, terminalTasks);
     }
 
+    private BackgroundTask findFirstTerminalTrackedTask(
+            RuntimeContext runtimeContext, String sessionId, List<String> waitSet) {
+        if (waitSet.isEmpty()) {
+            return null;
+        }
+        Collection<BackgroundTask> currentTasks =
```
oss-maintainer, line 384: `listTasks(rc, sessionId, null)` re-materialises the whole session task list on every 3 s tick, and the nested `waitSet × currentTasks` loop is O(n·m) per tick. The barrier path right next to it uses point lookups (`taskRepository.getTask(rc, sessionId, taskId)` in `completeTaskBarrierIfReady`), which is both cheaper and stable when a session accumulates many historical tasks. Since `legacyWaitSet` is already the exact id set you care about, could the fallback iterate it with `getTask` (or index the listing into a `HashMap` once per tick)? Purely a cost/consistency point — behaviour is equivalent.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/WaitAsyncResultsTool.java
```diff
@@ -220,8 +218,22 @@ public String waitForResults(
...
+                    findFirstTerminalTrackedTask(runtimeContext, sessionId, legacyWaitSet);
+            if (terminalTask != null) {
+                log.info(
+                        "wait_async_results: tracked task {} is terminal, session={}",
+                        terminalTask.getTaskId(),
+                        sessionId);
+                emptyWaits.set(0);
+                return formatLegacyResult(runtimeContext, sessionId, terminalTask);
+            }
+            long remainingMs = deadlineMs - System.currentTimeMillis();
```
oss-maintainer, line 231: Moving the deadline check below the two probes means the final iteration now runs `inboxHasMessages(...).block()` plus the repository scan *after* the budget has expired, so an explicit `timeoutSeconds=1` call can return slightly later than before (one extra probe round). That is arguably an improvement — a completion landing exactly at the deadline is still caught — but it does relax the guardrail the comment right below advertises ("never overshoot the caller's timeout"). If intentional, please say so in the loop comment; otherwise checking `remainingMs` at the top and doing one final probe after `break` keeps the old bound.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/WaitAsyncResultsTool.java
```diff
@@ -179,9 +180,10 @@ public String waitForResults(
                     + "blocking.";
         }
 
+        List<String> legacyWaitSet = List.of();
         if (taskRepository != null) {
-            boolean hasNonTerminal =
-                    hasNonTerminalTasks(runtimeContext, sessionId) || hasExternalWork();
+            legacyWaitSet = snapshotNonTerminalTaskIds(runtimeContext, sessionId);
+            boolean hasNonTerminal = !legacyWaitSet.isEmpty() || hasExternalWork();
             if (!hasNonTerminal) {
```
oss-maintainer, line 187: The wait set is snapshotted once, before `normalizeTimeout`, so a sub-agent spawned *while* this call is blocked is still invisible to the fallback: with an empty inbox it will keep ticking to the timeout even though a tracked task completed mid-wait. That mirrors the documented `wait_all` semantics ("tasks created while waiting are not added"), so it may well be deliberate for legacy mode as well — but the failure this PR fixes is exactly "a completion nobody noticed", so it is worth stating the choice in the javadoc above, or re-snapshotting the non-terminal set on each tick.

oss-maintainer (commented):
### Summary

Makes the legacy (`task_ids`/`wait_all` absent) mode of `wait_async_results` recover a sub-agent completion that never arrived as an inbox notification: it snapshots the non-terminal task ids at wait start, keeps the inbox as the fast path, and on each tick falls back to the authoritative `TaskRepository` state, returning the first tracked task that reached terminal state through the extracted `formatTerminalResults` (so results, errors and cancellations all surface, and delivery is marked exactly once).

I read the change against the legacy ANY contract and it holds: tasks already terminal before the wait are excluded by `snapshotNonTerminalTaskIds`, the `task_ids` / `wait_all=true` barrier is untouched, and the new tests pin the fast path, the historical-completion exclusion, the failed and cancelled returns. The points below are cost, timeout-precision and one semantics question — none of them block the fix.

### Findings
- **[Warning]** `WaitAsyncResultsTool.java:384` — `listTasks(rc, sessionId, null)` rescans the whole session task list on every 3 s tick, with a nested `waitSet × tasks` loop; the neighbouring barrier path already uses `getTask` point lookups.
- **[Info]** `WaitAsyncResultsTool.java:231` — the deadline check moved below the two probes, so the last iteration can overshoot the caller's timeout by one probe round; fine if intentional, but it softens the guardrail the comment right below still advertises.
- **[Info]** `WaitAsyncResultsTool.java:183` — the wait set is snapshotted once, so a sub-agent spawned while this call is blocked stays invisible and the tool times out even when that task completed mid-wait.

### Suggestions
For the rescan, `getTask` is a drop-in for the tracked set:

```java
for (String taskId : waitSet) {
    BackgroundTask task = taskRepository.getTask(runtimeContext, sessionId, taskId);
    if (task != null && task.getTaskStatus().isTerminal()) {
        return task;
    }
}
```

If the single snapshot is deliberate (it matches the documented `wait_all` semantics), one line in the class javadoc saying so will save the next reader the same question.

Verified: CLA signed, CI green at head `84893b96` (`build (ubuntu-latest)`, `build (windows-latest)`, `Check License`, `Check Module Sync`, `codecov/patch` all SUCCESS), `mergeable=true`, `mergeable_state=BLOCKED` (branch protection, not a conflict). No blocking issue found — LGTM for the maintainer's call once the inline items are triaged.

---

PR #2803, opened 2026-08-22
fix(harness): route task records around the per-call sandbox proxy

### LINKED ISSUE #2743: [Bug]: WorkspaceTaskRepository orphan sweeper and heartbeat fail under sandbox filesystem
**Describe the bug**

When a `HarnessAgent` is configured with a sandbox-backed filesystem (e.g. `SandboxBackedFilesystem` via the sandbox lifecycle middleware), the background maintenance threads of `WorkspaceTaskRepository` (`ws-task-maint-*`) cannot persist or read task records. The sandbox instance is injected into the filesystem only during an agent call (`SandboxLifecycleMiddleware.acquireForCall` → `setSandbox`, cleared in `releaseForCall`), but the heartbeat and orphan sweeper run on a standalone daemon thread with `sandbox == null`. As a result:

- The orphan sweeper throws `SandboxException.SandboxConfigurationException: No active sandbox — sandbox filesystem used outside of a call context` every sweep cycle and only logs a warning, so **orphaned tasks are never marked FAILED**.
- The heartbeat silently fails the same way (caught at `debug` level), so **`lastUpdatedAt` is never refreshed and RUNNING tasks never progress** — the whole task-liveness mechanism is broken in sandbox mode.

**To Reproduce**

1. Configure `HarnessAgent`/`ReActAgent` with a sandbox-backed workspace (sandbox mode + `SandboxBackedFilesystem`).
2. Submit an async background subagent task (`timeout_seconds=0`) via `AgentSpawnTool`.
3. Observe logs every ~5 min:

```
[ws-task-maint-176] WARN i.a.h.a.s.t.WorkspaceTaskRepository: - Orphan sweeper encountered an error: No active sandbox — sandbox filesystem used outside of a call context
```

**Expected behavior**

Task records are cross-call orchestration metadata and must not depend on a per-call sandbox instance. The sweeper should detect and mark orphaned tasks as `FAILED`, and the heartbeat should refresh `lastUpdatedAt` even when no agent call is in progress.

**Root cause (code pointers)**

- `io.agentscope.harness.agent.filesystem.sandbox.SandboxBackedFilesystem#requireSandbox` (line ~194) throws when `sandbox == null`.
- `WorkspaceTaskRepository#sweepOrphanedTasks` calls `WorkspaceManager#listAllTaskRecords` → `filesystem.glob(...)` on the maintenance thread.
- `WorkspaceManager` task-record paths (`readTaskMap`/`persistTaskMap`/`listAllTaskRecords`) route through the filesystem layer (`readWithOverride`/`writeUtf8WorkspaceRelative`), which requires a live sandbox.
- `SandboxLifecycleMiddleware` sets/clears the sandbox only around `ReActAgent.beforeAgentExecution`/`afterAgentExecution`.

- AgentScope-Java Version: latest (2.0.3-SNAPSHOT)
- Java Version: TBD
- OS: Linux

**Additional context**

The `ControlPlaneTaskRepository` (aistio) is not affected because it persists via the control-plane HTTP API and has no client-side sweeper.

birdie7761:
Fixes #2743

## Description

Under a sandbox-backed workspace, `WorkspaceTaskRepository`'s maintenance threads (`ws-task-maint-*`) cannot persist or read task records: the sandbox instance is injected into `SandboxBackedFilesystem` only during an agent call (`SandboxLifecycleMiddleware.acquireForCall` → `releaseForCall`), while the heartbeat and orphan sweeper run on a standalone scheduler with `sandbox == null`. Three failure modes, all reproduced by the new tests before the fix:

- the orphan sweeper throws `SandboxException.SandboxConfigurationException` on every cycle (unguarded `filesystem.glob` in `listAllTaskRecords`), so **orphaned tasks are never marked FAILED**;
- the heartbeat fails silently at `debug` level (the throwing filesystem layer bypasses `readWithOverride`'s local-disk fallback), so **`lastUpdatedAt` is never refreshed**;
- an async local task whose originating call already ended **cannot even start** — the supplier's initial task-record read throws the same way.

Task records are cross-call orchestration metadata and must not depend on a per-call sandbox instance. `WorkspaceManager` now routes task-record IO (`readTaskMap`, `persistTaskMap`, and the glob in `listAllTaskRecords`) to the host workspace whenever the filesystem layer serving those paths is the per-call sandbox proxy — the same location used when no filesystem layer is configured:

```java
private boolean taskRecordsRouteToLiveSandbox(String relPath) {
    AbstractFilesystem fs = this.filesystem;
    if (fs instanceof RoutedSandboxFilesystem routed) {
        fs = routed.backendFor(relPath);
    }
    return fs instanceof SandboxBackedFilesystem;
}
```

Two deliberate type boundaries:

- the check targets `SandboxBackedFilesystem` specifically, not `AbstractSandboxFilesystem` — the latter is also implemented by persistent filesystems (`LocalFilesystemWithShell`, `OverlayFilesystem`, …) that only expose `execute()`;
- explicit prefix routes are respected: `RoutedSandboxFilesystem.backendFor` (new, delegating to the new `CompositeFilesystem.filesystemFor`) resolves the backend actually serving the path, so a persistent store mounted for `agents/` keeps serving task records unchanged — locked by a dedicated test.

`RemoteFilesystemSpec` deployments are unaffected (task records still route to shared storage; guarded by the existing `WorkspaceTaskRepositoryTest` coverage).

## Test evidence (TDD red → green)

New `WorkspaceTaskRepositorySandboxMaintenanceTest` (3 cases). The first two assemble the real classes — a `SandboxBackedFilesystem` that never had a sandbox injected models the outside-a-call state:

| Case | Before (main) | After |
|---|---|---|
| orphan sweeper marks orphaned tasks FAILED | ❌ `expected: <FAILED> but was: <RUNNING>` | ✅ |
| heartbeat refreshes `lastUpdatedAt` | ❌ `task supplier never started` | ✅ |
| explicit `agents/` persistent route still serves task records (contract lock: no silent host-disk fallback, read-back and sweeper listing go through the routed backend) | — (added during review) | ✅ |

Regression: targeted task-repository family + `CompositeFilesystemTest` green, full harness suite **836 tests / 0 failures** (3 skipped pre-existing), `spotless:check` clean.

## Checklist


xiaoyuuuuuupeng (approved):
LGTM — clear fix for #2743, and the tests lock the failure modes that used to break outside a call context.

## Summary

Under a sandbox-backed workspace, `WorkspaceTaskRepository`'s maintenance threads (heartbeat / orphan sweeper) and late-starting async local tasks need task-record IO *after* `releaseForCall`, when `SandboxBackedFilesystem` holds no live sandbox. Routing those paths to the host workspace (the same place used when no filesystem layer is configured) is the right boundary: task records are cross-call orchestration metadata, not per-call sandbox content.

I walked the three IO entry points (`readTaskMap` / `persistTaskMap` / `listAllTaskRecords` glob), the `RoutedSandboxFilesystem.backendFor` longest-prefix lookup, and the new regression suite. No blocking issues.

## Verified

- Production assembly in `HarnessAgent` is either bare `SandboxBackedFilesystem` or `RoutedSandboxFilesystem` — both hit `taskRecordsRouteToLiveSandbox` correctly.
- Explicit `agents/` persistent routes stay on the routed backend (no silent host fallback).
- Path validation / per-file locks are preserved on the host branch.
- Local: `WorkspaceTaskRepositorySandboxMaintenanceTest` green; CI green.

Nit (non-blocking): `persistTaskMap`'s host branch calls `writeLocalFile(rel)` without an explicit `requireSafeRelativePath` (read side has one). `rel` comes from `taskRecordPath` and `writeLocalFile` still refuses writes outside `workspace`, so this is fine as-is.

jujn (approved):

jujn:

PR #2804, opened 2026-08-22
fix(sandbox-e2b): preserve zero exit codes in envd streams

### LINKED ISSUE #2793: fix(sandbox-e2b): E2bEnvdProcessClient drops the end event when exit_code is 0 - every successful command is reported as a failure
## Summary

`E2bEnvdProcessClient` (agentscope-extensions-sandbox-e2b) uses proto3 field-presence
APIs (`getAllFields()` / `hasField()`) to decide whether an `end` event exists and
whether `exit_code` was set. `exit_code` is a proto3 singular scalar **without
explicit presence**, so the value `0` is indistinguishable from "absent" through
those APIs. The result inverts the meaning of success:

- a command that exits with **0 (success)** → the `end` event is dropped / the exit
  code is never read → the caller sees the initial sentinel `Integer.MIN_VALUE` →
  `ExecException: Command exited with code -2147483648`;
- a command that exits with **non-zero (failure)** → everything works normally.

**Every successful command is reported as a failure.** Since sandbox/workspace
bootstrap itself runs commands that exit 0, the E2B backend cannot even hand out a
sandbox handle when the envd endpoint speaks JSON.

## Affected versions

- 2.0.1, 2.0.2, 2.0.2-subagent-bugfix — `E2bEnvdProcessClient.class` is byte-identical
  across these three releases on Maven Central.
- Still present on `main` as of 2026-08-20 (checked at 643905e6).

## Environment where it bites

Any e2b-compatible envd that serves **JSON** frames. Concretely reproduced against
Alibaba Cloud FC's e2b-compatible service (`*.e2b.fc.aliyuncs.com`): FC's envd
rejects the PROTO codec with HTTP 400 and only speaks JSON, so the JSON parsing
path is mandatory there.

Server-side evidence that the endpoint is fine: the same request sent with `curl`
returns three complete frames —

```
{"event":{"start":{...}}}
{"event":{"data":{"stdout":"aGkK"}}}
{"event":{"end":{"exitCode":0}}}
```

The `end` frame is present with `exitCode: 0`; it is the client that discards it.

## Standalone reproduction (no network, unmodified Central artifact)

A minimal project with a single dependency on
`io.agentscope:agentscope-extensions-sandbox-e2b:2.0.1` feeds the exact envd JSON
frames above into `parseJsonStartResponse` via reflection. Output:

```
== Part 1: the proto3 fact the bug rests on (pure protobuf, no agentscope logic) ==
EndEvent builder after setField(exit_code, 0) -> getAllFields().isEmpty() = true
EndEvent builder after setField(exit_code, 1) -> getAllFields().isEmpty() = false

== Part 2: agentscope 2.0.1 parseJsonStartResponse on real envd JSON frames ==
envd frame: {"event":{"end":{"exitCode":0}}}
  end event survives parsing : false   <== END EVENT DROPPED BY THE CLIENT
  exit code the caller sees  : -2147483648   <== ExecException: Command exited with code -2147483648
envd frame: {"event":{"end":{"exitCode":1}}}
  end event survives parsing : true
  exit code the caller sees  : 1
```

The `exitCode:1` control case shows the parser is otherwise fine — the failure is
specific to the success value `0`.

## Root cause — two spots, same mistake

**1. JSON → DynamicMessage assembly, `parseJsonStartResponse`:**

```java
JsonNode exitCodeNode = endNode.path("exitCode");
if (exitCodeNode.canConvertToInt()) {
    endBuilder.setField(exitCodeField, exitCodeNode.intValue());
}
if (!endBuilder.getAllFields().isEmpty()) {           // <-- bug
    event.setField(processEventDesc.findFieldByName("end"), endBuilder.build());
}
```

`exit_code` is a proto3 singular int32 without explicit presence. After
`setField(exitCodeField, 0)`, `endBuilder.getAllFields()` is **empty** — that is
documented proto3 behavior, not a protobuf bug. So the presence of the `end` event
is judged from the wrong place, and the event is attached only when the exit code
is non-zero.

**2. Stream draining, `drainStartStream`:**

```java
if (pe.hasField(peEndF)) {
    DynamicMessage end = (DynamicMessage) pe.getField(peEndF);
    Descriptors.FieldDescriptor ec =
            end.getDescriptorForType().findFieldByName("exit_code");
    if (end.hasField(ec)) {                            // <-- bug
        Object v = end.getField(ec);
        exit = v instanceof Integer ? (Integer) v : ((Long) v).intValue();
    }
}
```

Same presence trap on the read side: for `exit_code == 0`, `hasField()` is false and
`exit` keeps its `Integer.MIN_VALUE` initializer. Note this read-side guard is wrong
even for the binary PROTO path — proto3 wire format omits default-valued singular
scalars, so a server that legitimately encodes `exit_code = 0` also trips it.

## Suggested fix

**`parseJsonStartResponse`** — judge presence from the JSON, not from
`getAllFields()`:

```java
JsonNode exitCodeNode = endNode.path("exitCode");
boolean hasExitCode = exitCodeNode.canConvertToInt();
if (hasExitCode) {
    endBuilder.setField(exitCodeField, exitCodeNode.intValue());
}
if (hasExitCode || !endBuilder.getAllFields().isEmpty()) {
    event.setField(processEventDesc.findFieldByName("end"), endBuilder.build());
}
```

**`drainStartStream`** — read unconditionally; the proto3 default for an absent
field is `0`, which is exactly the right value here:

```java
Object v = end.getField(ec);
exit = v instanceof Integer ? (Integer) v : ((Long) v).intValue();
```

We have been running both changes as a local patch in production use against FC's
e2b service; happy to turn this into a PR if the approach looks right to you.

## Regression test suggestion

A unit test in the same package pins the behavior without any network:

```java
@Test
void anExitCodeOfZeroSurvivesJsonParsing() throws Exception {
    E2bSandboxClientOptions options = new E2bSandboxClientOptions();
    options.setApiKey("test-only-not-used");
    E2bEnvdProcessClient client = new E2bEnvdProcessClient(options);

    byte[] frame = "{\"event\":{\"end\":{\"exitCode\":0}}}"
            .getBytes(StandardCharsets.UTF_8);
    DynamicMessage response = /* invoke parseJsonStartResponse(frame) */;

    Descriptors.FieldDescriptor eventField =
            response.getDescriptorForType().findFieldByName("event");
    assertThat(response.hasField(eventField)).isTrue();

    DynamicMessage event = (DynamicMessage) response.getField(eventField);
    Descriptors.FieldDescriptor endField =
            event.getDescriptorForType().findFieldByName("end");
    assertThat(event.hasField(endField)).isTrue();   // fails before the fix

    DynamicMessage end = (DynamicMessage) event.getField(endField);
    Object exitCode = end.getField(
            end.getDescriptorForType().findFieldByName("exit_code"));
    assertThat(((Number) exitCode).intValue()).isZero();
}
```

The negative control that documents the proto3 fact this bug rests on: build an
`EndEvent` via `DynamicMessage.newBuilder(endDesc)`, `setField(exit_code, 0)`, and
assert `builder.getAllFields()` is empty — while `setField(exit_code, 1)` makes it
non-empty. That is why the `getAllFields().isEmpty()` heuristic can never work for
this field.

mikemikimike:
## Summary

Fixes #2793.

`E2bEnvdProcessClient` incorrectly relies on protobuf field presence for the proto3 scalar `exit_code`. That causes JSON `exitCode: 0` end events to be dropped, and causes both JSON and binary streams to retain `Integer.MIN_VALUE` instead of the successful exit code.

This change:
- preserves an end event when JSON explicitly contains `exitCode`, including `0`;
- reads the proto3 scalar default unconditionally from parsed end events;
- adds a regression test through the JSON stream-draining path.

## Validation

- `mvn -pl agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b -am -Dspotless.check.skip=true -Dtest=E2bEnvdProcessClientTest -Dsurefire.failIfNoSpecifiedTests=false test` — 8 passed
- `mvn -pl agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b spotless:check` — passed
- `git diff --check` — passed

The Maven commands were run with JDK 21 because the repository's Spotless plugin requires Java 17+; the shell's default `JAVA_HOME` points to JDK 8.


agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
```diff
@@ -341,8 +341,11 @@ private DynamicMessage parseJsonStartResponse(byte[] data) throws IOException {
             DynamicMessage.Builder endBuilder = DynamicMessage.newBuilder(endDesc);
             Descriptors.FieldDescriptor exitCodeField = endDesc.findFieldByName("exit_code");
             JsonNode exitCodeNode = endNode.path("exitCode");
-            if (exitCodeNode.canConvertToInt()) {
+            boolean hasExitCode = exitCodeNode.canConvertToInt();
+            if (hasExitCode) {
                 endBuilder.setField(exitCodeField, exitCodeNode.intValue());
+            }
+            if (hasExitCode || !endBuilder.getAllFields().isEmpty()) {
```
oss-maintainer, line 351: **[Warning]** `parseJsonStartResponse` only ever sets `exit_code` on `endBuilder` (lines 345-347), so the second disjunct `!endBuilder.getAllFields().isEmpty()` can never be true while `hasExitCode` is false — the condition is currently equivalent to the old `if (exitCodeNode.canConvertToInt())`. That means behavior for an explicit `"exitCode": 0` is unchanged from `main` (the end event was already attached there, since `canConvertToInt()` is true for `0`).

The key question for #2793: what does envd actually send when the process exits with 0? If it uses proto3 JSON default omission (frame looks like `{"event":{"end":{}}}` with no `exitCode` field), this change still drops the end event and `drainStartStream` still returns `Integer.MIN_VALUE`, i.e. the reported failure mode would remain. Could you share a real envd capture for a successful command? If envd really omits the field, consider treating a present-but-empty `end` object as a valid end event carrying the proto3 default exit code `0` — consistent with the binary path, where an empty `EndEvent` on the wire yields `end.getField(exit_code)` = `0`.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/test/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClientTest.java
```diff
@@ -158,6 +158,17 @@ void jsonCodecPreservesZeroExitCode() throws Exception {
...
 
+    @Test
+    void jsonCodecIgnoresEndWithoutExitCode() throws Exception {
+        E2bEnvdProcessClient client = new E2bEnvdProcessClient(options(E2bCodec.JSON));
+        ByteArrayOutputStream stdout = new ByteArrayOutputStream();
+        ByteArrayOutputStream stderr = new ByteArrayOutputStream();
+        int exit =
+                drainStartStream(client, connectFrame("{\"event\":{\"end\":{}}}"), stdout, stderr);
+
+        assertEquals(Integer.MIN_VALUE, exit);
```
oss-maintainer: **[Info]** This assertion pins `Integer.MIN_VALUE` for `{"event":{"end":{}}}`. If envd omits a zero `exitCode` in its JSON (proto3 default omission), that frame is exactly the success-case wire shape described in #2793 — in that scenario this test would codify the buggy behavior instead of preventing it. Worth verifying against a real envd stream.

Also note that open PR #2828 removes the `MIN_VALUE` sentinel entirely and throws `IOException` when the stream ends without an exit code, while editing the same method and tests. The two approaches conflict; they should be reconciled (pick one end state and rebase the other) to avoid semantic conflicts at merge time.

oss-maintainer (commented):
### Summary
This PR reworks the JSON `end`-event handling in `E2bEnvdProcessClient.parseJsonStartResponse` so a present `exitCode` (including `0`) preserves the end event, and adds a regression test. The direction matches issue #2793, and the CLA is signed — thanks for the clear write-up and tests.

### Findings
- **[Warning]** `E2bEnvdProcessClient.java:348` — the new condition `hasExitCode || !endBuilder.getAllFields().isEmpty()` is currently equivalent to `hasExitCode`, because this code path only ever sets `exit_code` on `endBuilder`. That makes the change a no-op when envd emits an explicit `"exitCode": 0` (the old code already attached the end event there), and still drops the end event when envd omits a zero exit code (proto3 JSON default omission). Please confirm what a real envd stream actually sends on success (a capture of the final frame would settle it).
- **[Info]** `E2bEnvdProcessClientTest.java:169` — the new test pins `Integer.MIN_VALUE` for `{"end":{}}`; if envd omits a zero `exitCode`, that frame is exactly the success shape from #2793, so the test could end up codifying the buggy behavior. Also heads-up that open PR #2828 removes the `MIN_VALUE` sentinel entirely (throws on missing exit code) and edits the same method/tests — worth coordinating so the two don't land in conflict.

### Suggestions
- Capture one real envd `end` frame for a successful (exit 0) command and paste it here. If it is `{"end":{}}` / has no `exitCode`, the fix needs to treat a present-but-empty `end` as exit code `0` (aligning with the binary path, where `end.getField(exit_code)` returns the proto3 default `0`). If it contains `"exitCode": 0`, we should revisit what actually reproduced #2793, since `canConvertToInt()` is already true for `0` on `main`.

Overall a solid first contribution — just want to nail down the root cause before approving. Happy to re-review once the envd frame shape is confirmed.
---

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
mikemikimike: Thanks for raising this. I checked the capture in #2793: the successful Alibaba FC envd frame is `{"event":{"end":{"exitCode":0}}}`, rather than an empty `end` object. The original JSON parser did call `setField(exit_code, 0)`, but the subsequent `getAllFields().isEmpty()` presence check drops proto3 scalar defaults, so the end event was still omitted. The new `hasExitCode` condition deliberately preserves the JSON-level field presence and fixes that explicit-zero frame. I have kept the regression focused on this captured wire shape.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/test/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClientTest.java
mikemikimike: Agreed that this assertion would create an unnecessary semantic commitment for a frame that is not the captured #2793 success shape. I removed `jsonCodecIgnoresEndWithoutExitCode` in 6df1778f, leaving the regression focused on the explicit `exitCode: 0` capture. #2828 can define and test the separate incomplete-stream behavior (missing end/exit code) without this PR pinning it to `Integer.MIN_VALUE`.

mikemikimike:
Addressed the outstanding stream-completion concern in commit 5d837e8. The issue's real envd success frame carries an explicit exitCode: 0, which remains preserved; an empty end object now remains absent so drainStartStream treats the stream as incomplete, consistent with #2828. Local validation: targeted reactor test/build passed (E2bEnvdProcessClientTest 10/10), Spotless check passed, and git diff --check passed.

PR #2830, opened 2026-08-24
fix(tool): honor execution config retries for tool exceptions

### LINKED ISSUE #2829: [Bug]: ExecutionConfig retry settings are ignored for tool failures
**Describe the bug**

When a tool call fails with an exception, the retry policy configured in `ExecutionConfig` (`maxAttempts`, `retryOn`, backoff) is never applied. The tool is invoked exactly once and the failure is immediately returned as a `ToolResultBlock` error result, even for clearly retryable errors such as `IOException` (network/transport failures).

Root cause: `ToolExecutor.executeWithInfrastructure(...)` runs the compatibility entry `execute(...)`, which converts tool failures into `ToolResultBlock` error results inside `executeCore(...)` **before** the scheduling/timeout/retry layers run. By the time `applyRetry(...)` executes, failures are already completed results, so `retryWhen` never sees an error signal. Only the error emitted by the `timeout(...)` operator can still reach the retry layer.

**To Reproduce**

1. Add the following test class to `agentscope-core/src/test/java/io/agentscope/core/tool/RetryReproTest.java`:

```java
package io.agentscope.core.tool;

import io.agentscope.core.message.ToolResultBlock;
import io.agentscope.core.message.ToolUseBlock;
import io.agentscope.core.model.ExecutionConfig;
import java.io.IOException;
import java.time.Duration;
import java.util.List;
import java.util.Map;
import java.util.concurrent.atomic.AtomicInteger;
import reactor.core.publisher.Mono;

/** Minimal reproduction for tool retry behavior. */
class RetryReproTest {

    static final AtomicInteger CALLS = new AtomicInteger(0);

    static class FlakyTools {
        @Tool(name = "flaky", description = "Fails once, then recovers")
        public String flaky() throws IOException {
            if (CALLS.incrementAndGet() == 1) {
                throw new IOException("Network down");
            }
            return "recovered";
        }
    }

    @Test
    void reproduce() {
        toolkit.registerTool(new FlakyTools());

        ExecutionConfig config =
                ExecutionConfig.builder()
                        .maxAttempts(3) // 1 initial attempt + up to 2 retries
                        .initialBackoff(Duration.ofMillis(100))
                        .maxBackoff(Duration.ofMillis(200))
                        .retryOn(ExecutionConfig.RETRYABLE_ERRORS) // IOException is retryable
                        .build();

        ToolUseBlock call =
                ToolUseBlock.builder()
                        .id("call-1")
                        .name("flaky")
                        .input(Map.of())
                        .content("{}") // required: schema validation reads ToolUseBlock.content()
                        .build();

        List<ToolResultBlock> results =
                toolkit.callTools(List.of(call), config, null, null).block();

        System.out.println("REPRO attempts = " + CALLS.get());
        System.out.println("REPRO state    = " + results.get(0).getState());
        System.out.println("REPRO output   = " + results.get(0).getOutput());
    }

    @Test
    void timeoutComparison() {
        AtomicInteger timeoutCalls = new AtomicInteger(0);
        toolkit.registerTool(
                new AgentTool() {
                    @Override
                    public String getName() {
                        return "never";
                    }

                    @Override
                    public String getDescription() {
                        return "never completes";
                    }

                    @Override
                    public Map<String, Object> getParameters() {
                        return Map.of();
                    }

                    @Override
                    public Mono<ToolResultBlock> callAsync(ToolCallParam param) {
                        timeoutCalls.incrementAndGet();
                        return Mono.never();
                    }
                });

        ExecutionConfig config =
                ExecutionConfig.builder()
                        .timeout(Duration.ofMillis(100))
                        .maxAttempts(3)
                        .initialBackoff(Duration.ofMillis(50))
                        .maxBackoff(Duration.ofMillis(100))
                        .build();

        ToolUseBlock call =
                ToolUseBlock.builder()
                        .id("call-2")
                        .name("never")
                        .input(Map.of())
                        .content("{}")
                        .build();

        List<ToolResultBlock> results =
                toolkit.callTools(List.of(call), config, null, null).block();

        System.out.println("REPRO timeout attempts = " + timeoutCalls.get());
        System.out.println("REPRO timeout state    = " + results.get(0).getState());
        System.out.println("REPRO timeout output   = " + results.get(0).getOutput());
    }
}
```

2. Run it from the repository root:

```
mvn -pl agentscope-core test -Dtest=RetryReproTest -Dsurefire.failIfNoSpecifiedTests=false
```

3. See the printed output and the surefire report (see **Error messages**).

**Expected behavior**

The first failure should surface as an error signal to the retry layer, so the tool is re-invoked and recovers:

```
REPRO attempts = 2
REPRO state    = RUNNING
REPRO output   = ["recovered"]
```

(`RUNNING` is the framework's default state for a plain successful tool result.) In the timeout comparison, the tool should be re-invoked on every attempt and the final error result should keep the root cause:

```
REPRO timeout attempts = 3
REPRO timeout state    = ERROR
REPRO timeout output   = [Error: Tool execution failed: Tool execution timeout after PT0.1S]
```

**Error messages**

No exception is thrown; the symptom is the missing retry. Actual output on the current `main`:

```
REPRO attempts = 1
REPRO state    = ERROR
REPRO output   = [Error: Tool execution failed: Network down]
REPRO timeout attempts = 1
REPRO timeout state    = ERROR
REPRO timeout output   = [Error: Tool execution failed: Retries exhausted: 2/2]
```

Observations:

- For the flaky tool, no `Retrying tool call 'flaky' ...` warning ever appears: the retry layer never runs for tool exceptions.
- The timeout comparison isolates the two halves of the bug: the timeout error signal **is** retried (the `Retries exhausted: 2/2` message proves 3 attempts), but the tool is still only invoked **once** (`REPRO timeout attempts = 1`) — retries merely re-subscribe the already-assembled chain instead of re-invoking the tool. Additionally, the exhausted-retry message replaces the real cause (`Tool execution timeout after PT0.1S`) with the generic `Retries exhausted: 2/2`.

- AgentScope-Java Version: 2.0.3-SNAPSHOT (reproduced on `main` at commit `ddad42e3`; the same code path exists in earlier 2.0.x)
- Java Version: 17
- OS: macOS (the issue is OS-independent)

**Additional context**

- Relevant code: `ToolExecutor.executeCore(...)` converts failures with `ToolResultBlock.error(...)`; `ToolExecutor.executeWithInfrastructure(...)` then pipes that chain through `applyScheduling` / `applyTimeout` / `applyRetry` / `applyShutdownGuard`, so `applyRetry` only ever sees completed results (except timeouts).
- A fix needs the tool invocation to keep failures as reactive error signals until after the timeout/retry layers (for example, a separate execution channel that is re-invoked per attempt), while preserving the existing `Toolkit.callTool(...)` contract that direct calls still receive error results.
- The fix must also keep `ToolSuspendException` (external-execution suspension) from ever being retried, even when it is wrapped in reflection/future/arbitrary exceptions, since replaying a suspension duplicates side effects.
- I am preparing a fix and will open a PR referencing this issue.

kevinyang03:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #2829.

**Background:** Tool calls never honored the retry settings in `ExecutionConfig`. `ToolExecutor.executeCore(...)` converted tool failures into completed `ToolResultBlock` error results before the timeout/retry layers ran, so `applyRetry` only ever saw completed results. Verified on `main` with the reproduction from #2829: a tool throwing `IOException` once was invoked only once (`maxAttempts=3` with `RETRYABLE_ERRORS` configured), and in the timeout comparison the retry layer ran 3 attempts (`Retries exhausted: 2/2`) without ever re-invoking the tool.

**Changes:**

- Add an execution channel: `AgentTool.callAsyncForExecution(...)` (new `default` method, backward compatible) plus `ToolMethodInvoker.invokeRawAsync(...)`, which keep failures as reactive error signals. `ToolExecutor.executeWithInfrastructure(...)` now runs this channel through scheduling → timeout → retry → shutdown and converts the final error into an error result only after retries are exhausted. The existing `callAsync(...)` contract — failures surface as error results on direct calls — is unchanged and covered by tests for custom, annotation-based, and MCP tools.
- Make tool timeouts retryable: the timeout fallback now emits `TimeoutException`, which `ExecutionConfig.RETRYABLE_ERRORS` recognizes.
- Preserve the root cause when retries are exhausted via `onRetryExhaustedThrow((spec, signal) -> signal.failure())`, instead of the generic `Retries exhausted: N/N` message.
- Detect `ToolSuspendException` anywhere in the cause chain (`ExceptionUtils.findToolSuspendException(...)`, identity-guarded against circular causes) so suspensions — even wrapped in reflection/future/arbitrary exceptions — become suspended results and are never retried, including under the retry-everything fallback when no `retryOn` predicate is configured.
- Keep MCP protocol-level business errors (`isError=true`) as completed error results so they are never retried.
- Document the execution-channel contract on `AgentTool` and the tool retry semantics on `ExecutionConfig.TOOL_DEFAULTS`.

**How to test:**

- `mvn -pl agentscope-core test -Dtest=ToolExecutorTest` — 18 new end-to-end tests (class total: 30), covering retry of annotation tools (sync/Mono/CompletableFuture) and MCP transport failures, no-retry for protocol-level business errors and `ToolSuspendException` (raw and wrapped), retry-exhaustion error results with id/name, timeout retries, retry isolation in parallel batches, and the direct-call compatibility contract.
- The reproduction from #2829 now returns the recovered result after 2 attempts (`state=RUNNING`, output `["recovered"]`), and the timeout comparison re-invokes the tool on every attempt with the root cause preserved (`Tool execution timeout after PT0.1S`).

## Checklist

- [x] All tests are passing (`mvn test`) — `mvn -pl agentscope-core clean verify` passes locally: 2312 tests, 0 failures, 0 errors (9 pre-existing skips); the full-reactor `clean verify` runs in CI
- [x] Javadoc comments are complete and follow project conventions — `mvn -pl agentscope-core javadoc:javadoc` passes with `doclint=all,-missing`
- [x] Related documentation has been updated (e.g. links, examples, etc.) — no doc changes needed; the new API is a backward-compatible `default` method, and the `ExecutionConfig` javadoc is updated in code

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeoutException;
 import java.util.function.BiConsumer;
 import java.util.function.Predicate;
 import org.slf4j.Logger;
@@ -159,13 +160,47 @@ class ToolExecutor {
     // ==================== Single Tool Execution ====================
 
     /**
-     * Execute a single tool call with full infrastructure support.
+     * Execute a single tool call without the infrastructure layers (scheduling, timeout, retry,
+     * shutdown guard).
+     *
+     * <p>This is the compatibility entry point used by {@link Toolkit#callTool(ToolCallParam)}: it
+     * keeps the {@link AgentTool#callAsync(ToolCallParam)} contract that failures surface as
+     * {@link ToolResultBlock} error results instead of reactive error signals.
      *
      * @param param Tool call parameters
      * @return Mono containing execution result
      */
     Mono<ToolResultBlock> execute(ToolCallParam param) {
-        return TracerRegistry.get().callTool(this.toolkit, param, () -> executeCore(param));
+        return executeWithTracing(param, false)
+                .onErrorResume(e -> Mono.just(buildToolErrorResult(e)));
+    }
+
+    /**
+     * Execute a single tool call through the raw execution channel, keeping failures as reactive
+     * error signals for the infrastructure layers (scheduling, timeout, retry, shutdown guard) to
+     * act on. Used only by {@link #executeWithInfrastructure}.
+     *
+     * @param param Tool call parameters
+     * @return Mono containing execution result, or signalling an error on failure
+     */
+    private Mono<ToolResultBlock> executeRaw(ToolCallParam param) {
+        return executeWithTracing(param, true);
+    }
+
+    /**
+     * Run the traced core execution for the given tool call.
+     *
+     * @param param Tool call parameters
+     * @param useExecutionPath whether to use the raw execution channel
+     *     ({@link AgentTool#callAsyncForExecution(ToolCallParam)}) that keeps failures as error
+     *     signals, or the compatibility channel ({@link AgentTool#callAsync(ToolCallParam)}) that
+     *     converts failures into error results
+     * @return Mono containing execution result
+     */
+    private Mono<ToolResultBlock> executeWithTracing(
+            ToolCallParam param, boolean useExecutionPath) {
+        return TracerRegistry.get()
+                .callTool(this.toolkit, param, () -> executeCore(param, useExecutionPath));
     }
 
     /**
@@ -181,7 +216,7 @@ class ToolExecutor {
      *   <li>Actual tool invocation</li>
      * </ul>
      */
-    private Mono<ToolResultBlock> executeCore(ToolCallParam param) {
+    private Mono<ToolResultBlock> executeCore(ToolCallParam param, boolean useExecutionPath) {
         ToolUseBlock toolCall = param.getToolUseBlock();
         AgentTool tool = toolRegistry.getTool(toolCall.getName());
 
@@ -263,8 +298,34 @@ class ToolExecutor {
                         .emitter(toolEmitter)
                         .build();
 
-        return tool.callAsync(executionParam)
-                .onErrorResume(
+        // Invoke the tool through the selected channel.
+        //
+        // Compatibility channel (useExecutionPath=false): failures are converted into
+        // ToolResultBlock error results right here, preserving the AgentTool contract for callers
+        // that execute tools directly without infrastructure.
+        //
+        // Execution channel (useExecutionPath=true): failures stay as reactive error signals and
+        // flow through the timeout/retry/shutdown layers applied by executeWithInfrastructure,
+        // which converts them into ToolResultBlock error results only after retries are
+        // exhausted. Mono.defer re-invokes callAsyncForExecution on every subscription so each
+        // retry attempt rebuilds and re-runs the tool call instead of resubscribing to the same
+        // pre-assembled publisher.
+        Mono<ToolResultBlock> invocation;
+        if (useExecutionPath) {
+            invocation =
+                    Mono.defer(() -> tool.callAsyncForExecution(executionParam))
+                            // Unwrap reflection/future wrappers so the retry predicate sees the
+                            // original exception (IOException, transport errors, ...) instead of
+                            // InvocationTargetException or ExecutionException.
+                            .onErrorMap(ExceptionUtils::unwrapExecutionWrapper);
+        } else {
+            invocation = tool.callAsync(executionParam);
+        }
+
+        // ToolSuspendException is business-level suspension, not a failure: both channels convert
+        // it to a suspended result without retrying.
+        Mono<ToolResultBlock> chain =
+                invocation.onErrorResume(
                         ToolSuspendException.class,
                         e -> {
                             // Convert ToolSuspendException to suspended result
@@ -273,21 +334,23 @@ class ToolExecutor {
                                     toolCall.getName(),
                                     e.getReason() != null ? e.getReason() : "no reason");
                             return Mono.just(ToolResultBlock.suspended(toolCall, e));
-                        })
-                .onErrorResume(
-                        e -> {
-                            String errorMsg =
-                                    e.getMessage() != null
-                                            ? e.getMessage()
-                                            : e.getClass().getSimpleName();
-                            return Mono.just(
-                                    ToolResultBlock.error("Tool execution failed: " + errorMsg));
-                        })
-                .switchIfEmpty(
-                        Mono.just(
-                                ToolResultBlock.error(
-                                        "Tool execution failed: Tool completed without returning a"
-                                                + " result")));
+                        });
+        if (!useExecutionPath) {
+            chain = chain.onErrorResume(e -> Mono.just(buildToolErrorResult(e)));
+        }
+        return chain.switchIfEmpty(
+                Mono.just(
+                        ToolResultBlock.error(
+                                "Tool execution failed: Tool completed without returning a"
+                                        + " result")));
+    }
+
+    /**
+     * Build the standard {@link ToolResultBlock} error result for a failed tool call.
+     */
+    private ToolResultBlock buildToolErrorResult(Throwable e) {
+        String errorMsg = e.getMessage() != null ? e.getMessage() : e.getClass().getSimpleName();
+        return ToolResultBlock.error("Tool execution failed: " + errorMsg);
     }
 
     // ==================== Batch Tool Execution ====================
@@ -384,8 +447,10 @@ class ToolExecutor {
                         .runtimeContext(agentRuntimeContext)
                         .build();
 
-        // Get core execution
-        Mono<ToolResultBlock> execution = execute(param);
+        // Use the raw execution channel so tool failures remain reactive error signals while
+        // the timeout/retry/shutdown layers below run; this final onErrorResume converts the
+        // error that remains after retries are exhausted into a ToolResultBlock error result.
+        Mono<ToolResultBlock> execution = executeRaw(param);
 
         // Apply infrastructure layers
         execution = applyScheduling(execution);
@@ -424,9 +489,11 @@ class ToolExecutor {
         Duration timeout = config.getTimeout();
         logger.debug("Applied timeout: {} for tool: {}", timeout, toolCall.getName());
 
+        // TimeoutException (rather than a bare RuntimeException) so that
+        // ExecutionConfig.RETRYABLE_ERRORS recognizes tool timeouts as retryable.
         return execution.timeout(
                 timeout,
-                Mono.error(new RuntimeException("Tool execution timeout after " + timeout)));
+                Mono.error(new TimeoutException("Tool execution timeout after " + timeout)));
     }
 
     private Mono<ToolResultBlock> applyRetry(
@@ -450,6 +517,10 @@ class ToolExecutor {
                         .maxBackoff(maxBackoff)
                         .jitter(0.5)
                         .filter(retryOn)
+                        // On exhaustion, propagate the last failure itself instead of Reactor's
+                        // RetryExhaustedException wrapper so the error result surfaces the real
+                        // failure message (e.g. the timeout or the tool exception).
+                        .onRetryExhaustedThrow((spec, signal) -> signal.failure())
                         .doBeforeRetry(
                                 signal ->
                                         logger.warn(
--- a/agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
@@ -160,12 +160,10 @@ class ToolExecutor {
     // ==================== Single Tool Execution ====================
 
     /**
-     * Execute a single tool call without the infrastructure layers (scheduling, timeout, retry,
-     * shutdown guard).
+     * Execute a single tool call without the infrastructure layers.
      *
-     * <p>This is the compatibility entry point used by {@link Toolkit#callTool(ToolCallParam)}: it
-     * keeps the {@link AgentTool#callAsync(ToolCallParam)} contract that failures surface as
-     * {@link ToolResultBlock} error results instead of reactive error signals.
+     * <p>Compatibility entry point used by {@link Toolkit#callTool(ToolCallParam)}: failures
+     * surface as {@link ToolResultBlock} error results, not reactive error signals.
      *
      * @param param Tool call parameters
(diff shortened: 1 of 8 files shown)


guslegend0510 (approved):
LGTM

Leslie-ZH1023:
期望尽快合并

agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
```diff
@@ -450,6 +504,8 @@ private Mono<ToolResultBlock> applyRetry(
                         .maxBackoff(maxBackoff)
                         .jitter(0.5)
                         .filter(retryOn)
```
oss-maintainer: **[Critical] Silent behavior change: `maxAttempts > 1` now retries non-idempotent tools.**

Before this PR, a tool exception was converted into a completed `ToolResultBlock` error result *before* `applyRetry` ran, so `retryOn` never matched and `maxAttempts` was effectively inert for tool failures. Wiring the raw channel through makes the setting live — but `retryOn` defaults to `error -> true` when the caller configures only `maxAttempts` (as `retryConfig(int)` in the new tests does):

```java
Predicate<Throwable> retryOn =
        config.getRetryOn() != null ? config.getRetryOn() : error -> true;
```

So any existing user who set `maxAttempts=3` (e.g. copied from a model-call config) will now silently re-execute side-effecting tools — file writes, `ShellExecuteTool`, MCP POST-style calls — after a transient failure. That is a correctness regression risk for upgrade, not just a fix.

Suggested, either:
1. Default the tool-path predicate to a whitelist (`ExecutionConfig.RETRYABLE_ERRORS` already exists and is the safer default), or
2. Gate retries on `AgentTool.isReadOnly()`, which is already on the interface, so only read-only tools are retried implicitly, or
3. Keep `error -> true` but state the new behavior prominently in the changelog / `TOOL_DEFAULTS` javadoc, since the added note documents the mechanics but not the "previously inert, now active" migration hazard.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
```diff
@@ -263,31 +294,52 @@ private Mono<ToolResultBlock> executeCore(ToolCallParam param) {
...
-                            return Mono.just(ToolResultBlock.suspended(toolCall, e));
-                        })
-                .onErrorResume(
+        // Compatibility channel converts failures to error results here; the execution channel
+        // keeps them as error signals for the timeout/retry layers. Mono.defer re-invokes the
+        // tool per subscription so each retry runs a fresh attempt.
+        Mono<ToolResultBlock> invocation;
+        if (useExecutionPath) {
+            invocation =
+                    Mono.defer(() -> tool.callAsyncForExecution(executionParam))
```
oss-maintainer: **[Warning] The `ToolEmitter` is created once per `executeCore` call but reused by every retry attempt.**

`toolEmitter` (line ~273) and `executionParam` are built outside the `Mono.defer`, so re-subscription re-invokes `callAsyncForExecution` with the *same* emitter, whose `chunkCallback` is the caller's sink. With `maxAttempts > 1`, the partial chunks emitted by the discarded attempt have already been delivered before the retry re-emits from scratch, so the UI / streaming consumer can see duplicated or interleaved tool output.

The `doBeforeRetry` log already admits the previous "attempt is cancelled and may still be consuming resources" — cancellation stops further emissions but cannot retract what was already pushed.

Consider moving the emitter + `executionParam` construction inside the `Mono.defer` so each attempt gets a fresh emitter, or suppress chunk callbacks for attempts after the first (e.g. an attempt counter captured by the emitter) and note the chosen semantic in the `callAsyncForExecution` javadoc.

agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
```diff
@@ -424,9 +477,10 @@ private Mono<ToolResultBlock> applyTimeout(
         Duration timeout = config.getTimeout();
         logger.debug("Applied timeout: {} for tool: {}", timeout, toolCall.getName());
 
+        // TimeoutException so ExecutionConfig.RETRYABLE_ERRORS recognizes timeouts as retryable
         return execution.timeout(
                 timeout,
-                Mono.error(new RuntimeException("Tool execution timeout after " + timeout)));
+                Mono.error(new TimeoutException("Tool execution timeout after " + timeout)));
```
oss-maintainer: **[Warning] Timeout is applied per attempt, so worst-case latency becomes `maxAttempts × timeout` plus backoff.**

`applyTimeout` runs before `applyRetry`, which is the right ordering for making timeouts retryable, but it means the configured `TOOL_DEFAULTS` 5-minute timeout is now a *per-attempt* budget. With `maxAttempts=3` a single tool call can occupy the ReAct loop for 15+ minutes, which callers of `callTools(...)` and any upstream agent/turn deadline probably do not expect.

Would be good to either bound the total elapsed time (apply the timeout outside the retry layer as an overall deadline in addition to the per-attempt one), or document explicitly in the `ExecutionConfig` javadoc that `timeout` is per attempt, not cumulative.

agentscope-core/src/main/java/io/agentscope/core/tool/mcp/McpTool.java
```diff
@@ -178,6 +178,32 @@ public Mono<PermissionDecision> checkPermissions(
...
+     * <p>Transport-level failures stay as reactive error signals so the executor can apply its
+     * retry policy. Protocol-level business errors ({@code isError=true}) remain completed error
+     * results and are never retried.
+     *
+     * @param param The tool call parameters containing toolUseBlock, input, and agent
+     * @return a Mono that emits the tool result, or signals an error on failure
+     */
+    @Override
+    public Mono<ToolResultBlock> callAsyncForExecution(ToolCallParam param) {
+        Objects.requireNonNull(param, "param must not be null");
```
oss-maintainer: **[Info] `Objects.requireNonNull` at assembly time escapes the compat `onErrorResume`.**

`callAsync` is now `callAsyncForExecution(param).onErrorResume(...)`, but `callAsyncForExecution` validates `param` eagerly, so a null argument throws synchronously at method-invocation time instead of surfacing through the `Mono`. The `.onErrorResume` in `callAsync` cannot catch it, which breaks the "direct path always returns an error result" contract this PR documents (and asserts in `shouldKeepCallAsyncContractOnDirectCalls`).

Wrapping the body in `Mono.defer(() -> { ... })` keeps the validation on the subscribe path so both channels behave uniformly. Same pattern as the `Mono.defer` you already added in `ToolExecutor.executeCore`.

agentscope-core/src/main/java/io/agentscope/core/util/ExceptionUtils.java
```diff
@@ -79,4 +83,46 @@ public static boolean containsInterruptedException(Throwable error) {
...
+    /**
+     * Finds a {@link ToolSuspendException} anywhere in the given throwable's cause chain.
+     *
+     * <p>The cause chain is walked with an identity set guard so circular causes cannot cause an
+     * infinite loop.
+     *
+     * @param error the throwable to inspect (may be {@code null})
+     * @return the first {@link ToolSuspendException} found, or {@code null} if none is present
+     */
+    public static ToolSuspendException findToolSuspendException(Throwable error) {
```
oss-maintainer: **[Info] Whole-cause-chain matching can turn a genuine failure into a permanent suspension.**

`findToolSuspendException` walks the entire chain, so any error that merely *wraps* a `ToolSuspendException` (rather than signalling a suspension) is reported as `ToolResultBlock.suspended(...)`, and the agent will then wait for an external result that can never arrive. `shouldNotRetryWrappedToolSuspension` shows this is deliberate, but the blast radius is larger than it looks once tools start composing sub-calls.

Two small hardening options: bound the walk to the first N causes (or only direct cause + one level), or log at `warn` when a *wrapped* suspension is detected so a mis-classified suspension is diagnosable instead of silent. Also worth noting `io.agentscope.core.util` now depends on `io.agentscope.core.tool`, a small layering inversion for a util package — the tool-facing helper could live next to `ToolSuspendException` instead.

oss-maintainer (commented):
﻿### Summary

Wiring the raw execution channel so `ExecutionConfig` retries actually apply to tool exceptions is the right fix, and the new test matrix (sync / Mono / CompletableFuture / MCP transport vs protocol / suspend / retryOn / parallel isolation / direct-path contract) is thorough. The `useExecutionPath` split correctly keeps `Toolkit.callTool` semantics intact, which was the risky part.

My main concern is upstream of the code: this turns a previously **inert** setting into a live one. Anyone who already set `maxAttempts > 1` on a tool path (with the default permissive `retryOn`) will start re-executing side-effecting tools after a transient failure once they upgrade, and the worst-case latency of a single tool call becomes `maxAttempts x timeout`. That deserves an explicit decision on the default retry predicate plus a changelog note before merge.

### Findings
- **[Critical]** `ToolExecutor.java:506` - `retryOn` defaults to `error -> true`, so `maxAttempts>1` now retries non-idempotent tools; previously this was a no-op for tool exceptions.
- **[Warning]** `ToolExecutor.java:303` - one `DefaultToolEmitter` is reused across retry attempts, so chunks emitted by the discarded attempt are already delivered before the retry re-emits.
- **[Warning]** `ToolExecutor.java:483` - timeout is applied per attempt, not cumulatively; document it or bound total elapsed time.
- **[Info]** `McpTool.java:206` - eager `Objects.requireNonNull` escapes the `onErrorResume` compat contract; wrap in `Mono.defer`.
- **[Info]** `ExceptionUtils.java:117` - whole-cause-chain suspension lookup can permanently suspend a genuine failure; consider bounding the walk. Also note `core.util` now depends on `core.tool`.

### Suggestions
- Default the tool retry predicate to `ExecutionConfig.RETRYABLE_ERRORS` (already available), or gate implicit retries on `AgentTool.isReadOnly()`, keeping an opt-in for callers that need the current behavior.
- Build `executionParam` / `toolEmitter` inside the `Mono.defer` so each attempt is isolated.
- Add a test asserting chunk callbacks are not duplicated across a retry - that gap is exactly what makes the emitter finding invisible to CI.

Overall: solid fix and excellent test coverage; I would only settle the default-retry-policy question so the upgrade path is safe for existing configurations.

---

oss-maintainer:

This PR currently conflicts with the target branch and cannot be merged. Please rebase or merge the latest target branch into your branch and resolve the conflicts:

```bash
git fetch origin
git checkout fix/tool-retry-on-exceptions
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #2839, opened 2026-08-25
fix(harness): prevent pipe deadlock in LocalFilesystemWithShell.execute

### LINKED ISSUE #2838: [Bug]: LocalFilesystemWithShell.execute deadlocks when child process output exceeds pipe buffer (waitFor before draining stdout/stderr)
`execute()` starts the child process and immediately calls `Process.waitFor(timeout)`, and only afterwards reads `getInputStream().readAllBytes()` / `getErrorStream().readAllBytes()`:

```java
Process process = pb.start();
boolean finished = process.waitFor(timeout, TimeUnit.SECONDS);   // waits first
String out = new String(process.getInputStream().readAllBytes(), charset);  // drains after
String err = new String(process.getErrorStream().readAllBytes(), charset);
```

If the child writes more than the OS pipe buffer (≈4 KB on Windows, 64 KB default on Linux) it blocks in `write()` and never exits, while the parent blocks in `waitFor()` and never drains the pipe — a classic pipe deadlock. Every such command is misreported as a timeout (exit code 124) and forcibly killed, even though it would have finished in well under a second.

**To Reproduce**

Any shell command producing output larger than the pipe buffer, e.g. a Python script printing a ~64 KB JSON document:

```java
LocalFilesystemWithShell fs = new LocalFilesystemWithShell(Path.of("/tmp/ws"));
ExecuteResponse resp = fs.execute(null, "python3 gen_64kb_output.py", 60);
```

A pure-shell reproducer (used in our regression test, no Python needed):

```bash
# Linux/macOS: prints ~68 KB
i=0; while [ "$i" -lt 4000 ]; do echo 0123456789abcdef; i=$((i+1)); done
rem Windows (cmd /c): prints ~72 KB
for /l %i in (1,1,4000) do @echo 0123456789abcdef
```

**Expected behavior**

The command completes with exit code 0 in well under a second, with full output captured.

**Error messages**

```
Error: Command timed out after 60 seconds (custom timeout). The command may be stuck or require more time.
```

(exit code 124, process forcibly killed)

**Environment**

- AgentScope-Java Version: 2.0.1 (also reproducible on `main` / 2.0.3-SNAPSHOT)
- Java Version: 17+
- OS: Windows (≈4 KB pipe buffer, triggered by almost any large output); also Linux for outputs > 64 KB

**Additional context**

Observed with a real workload: a ~64 KB (64,404 bytes) JSON response printed to stdout hangs on Windows; the same script completes in 0.6 s when stdout is consumed concurrently. `redirectErrorStream` does not help since neither stream is drained while waiting.

Note: the exact same deadlock was already fixed in `agentscope-core`'s `ShellCommandTool` ("Start asynchronous stream readers immediately to prevent pipe buffer deadlock", draining both streams via a thread pool while `waitFor` blocks). `LocalFilesystemWithShell` in `agentscope-harness` appears to have been missed by that fix.

Suggested fix: drain stdout/stderr concurrently with `waitFor` — mirroring the `ShellCommandTool` approach, e.g. two daemon threads copying each stream into a buffer, then `waitFor(timeout)`, then join the drainers (destroying the process first on timeout so the drainers terminate). We have a fix + regression test ready and are happy to open a PR.

zzz-ghost:
## What this PR does

Fixes a pipe deadlock in `LocalFilesystemWithShell.execute()`: the child process's stdout/stderr are now drained on daemon threads **concurrently** with `Process.waitFor()`, instead of only after it returns.

## Root cause

Previously `execute()` called `Process.waitFor(timeout)` first and read `getInputStream()` / `getErrorStream()` only afterwards. If the child wrote more than the OS pipe buffer (≈4 KB on Windows, 64 KB default on Linux), it blocked in `write()` forever while the parent blocked in `waitFor()` — a classic pipe deadlock. Every such command was misreported as a timeout (exit 124) and forcibly killed, even though it would have finished in well under a second.

## Fix

- Start two daemon drainer threads (8 KB chunks into `ByteArrayOutputStream`) immediately after `ProcessBuilder.start()`, then `waitFor(timeout)`, then join the drainers — destroying the process first on timeout so the drainers terminate.
- This mirrors the fix already applied to `ShellCommandTool` in `agentscope-core` ("Start asynchronous stream readers immediately to prevent pipe buffer deadlock"), which `LocalFilesystemWithShell` was missed by.
- Behavior is otherwise unchanged: same timeout messages, exit code 124 semantics, charset selection, and `maxOutputBytes` truncation.

## Testing

New regression test `execute_outputLargerThanOsPipeBufferCompletesWithoutDeadlock` prints ~70 KB via shell builtins (works on both `sh` and `cmd.exe`, no Python needed):

| | Before fix | After fix |
|---|---|---|
| Result | deadlock → killed after 60 s, exit **124** | completes in **0.4 s**, exit **0** |
| Output | — | all 4000 lines captured, not truncated |

Verified on Windows 10 (≈4 KB pipe buffer); the threshold (68–72 KB) also exceeds the 64 KB Linux default pipe buffer, so the test covers both platforms.

Fixes #2838


zzz-ghost:
Note: the two earlier `build (ubuntu-latest)` failures on this PR were a pre-existing flaky
test, unrelated to this change — `HarnessAgentDynamicHookBuilderTest` failed in *different*
methods each run with `JUnitException: Failed to close extension context`
(`DirectoryNotEmptyException` on the JUnit `@TempDir`). It looks like a race between JUnit
temp-dir cleanup and the framework's fire-and-forget session/transcript mirror writes;
`HarnessAgent.close()` documents this race, but these tests never close the agent.
CI is green after re-running. Happy to file a separate issue for the flaky test if you'd like.

dailingtao (approved):
LGTM cc @jujn

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -432,6 +456,38 @@ private Path resolveExecuteCwd(RuntimeContext rc) {
...
+     */
+    private static Thread drainAsync(InputStream in, ByteArrayOutputStream buf) {
+        Thread t =
+                new Thread(
+                        () -> {
+                            byte[] chunk = new byte[DRAIN_CHUNK_BYTES];
+                            int n;
+                            try {
+                                while ((n = in.read(chunk)) != -1) {
+                                    buf.write(chunk, 0, n);
```
xiaoyuuuuuupeng, line 472: nit / follow-up: please bound memory while draining, not only after assembling the final string.

`maxOutputBytes` currently truncates post-hoc, so a huge command can OOM before truncation. Cap `buf.write` once `buf.size()` hits the limit, but keep `read`ing so the OS pipe stays drained (same idea as gemini-cli / qwen-code). Mark `truncated` when bytes were discarded.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -432,6 +456,38 @@ private Path resolveExecuteCwd(RuntimeContext rc) {
...
+                            }
+                        });
+        t.setDaemon(true);
+        t.start();
+        return t;
+    }
+
+    private static void joinQuietly(Thread t) {
+        try {
+            t.join(DRAIN_JOIN_TIMEOUT_MILLIS);
```
xiaoyuuuuuupeng, line 485: nit / follow-up: please align reader completion with `ShellCommandTool`.
﻿
`join(5s)` may return while the drainer is still alive; `ByteArrayOutputStream` is not thread-safe, so calling `toString()` concurrently with `write` is racy. Prefer `Future.get(timeout)` + `cancel(true)` and discard untrusted output on timeout, or after `join` check `isAlive()` → interrupt and do not trust the buffer.

xiaoyuuuuuupeng (approved):
Deadlock fix LGTM.

Please open follow-up PR(s)/issues for:
1) Cap writes inside `drainAsync` using `maxOutputBytes` (keep reading so the pipe stays drained; mark truncated). Post-hoc truncate can OOM before truncation runs. gemini-cli / qwen-code both bound capture during async drain.
2) Align reader teardown with `ShellCommandTool` (`Future.get` + `cancel(true)`, or after `join` if `isAlive()` → interrupt and discard the untrusted BAOS).

Optional: add a stderr / dual-stream regression test.

zzz-ghost:
Thanks! Filed #2850 to track these — will send a follow-up PR once this one lands.

jujn:
Assign to @dailingtao @xiaoyuuuuuupeng (Review)

jujn (approved):

PR #2844, opened 2026-08-26
fix(accumulator): warn on malformed streamed tool arguments

### LINKED ISSUE #2841: [Bug]: ToolCallsAccumulator silently swallows JSON parse failures for streamed tool call arguments and the raw string is lost
## Describe the bug

`ToolCallsAccumulator$ToolCallBuilder.build()` parses the fully accumulated tool-call
arguments with `JsonUtils.getJsonCodec().fromJson(rawContentStr, Map.class)` inside a
`try/catch (Exception)` that does nothing on failure — no log, no event, no counter.
When the accumulated arguments are not valid JSON, the parsed input map stays empty and
`content` falls back to `"{}"`, so the string the model actually emitted is gone from the
entire system.

We hit this with Chinese LLMs (Qwen via OpenAI-compatible streaming), which fairly often
emit unescaped ASCII double quotes inside string values — a *complete* argument JSON that
is broken. App-side symptom: the tool is invoked with an empty parameter map and fails with
"missing required parameter". That looks like a model/tool-schema mismatch, and nothing in
the framework logs, events or traces hints at the real cause. We could only find the root
cause by decompiling the jar — the framework gives zero signal.

Verified at three points in time:
- agentscope-core 2.0.1: the bytecode catch handler is empty (it only stores the exception)
- agentscope-core 2.0.2 (latest release): bytecode identical to 2.0.1
- main branch (checked 2026-08-25): still `catch (Exception ignored) { // Parsing failed, keep previously merged args }` (ToolCallsAccumulator.java:116-118)

## To Reproduce

A streamed tool call whose accumulated arguments string is:

```json
{"projectName": "改造"文化礼堂"及配套设施"}
```

Minimal reproduction via the public API (same shape as `ToolCallsAccumulatorTest`):

```java
ToolCallsAccumulator accumulator = new ToolCallsAccumulator();
accumulator.add(ToolUseBlock.builder()
        .id("call_1")
        .name("write_text_file")
        .content("{\"projectName\": \"改造\"文化礼堂\"及配套设施\"}")
        .build());
ToolUseBlock toolCall = accumulator.buildAllToolCalls().get(0);
// toolCall.getInput()   -> {}        (empty)
// toolCall.getContent() -> "{}"
```

What `build()` does (ToolCallsAccumulator.java:100-131 on main):

1. `fromJson(rawContent, Map.class)` throws (unescaped quotes) — caught and completely ignored.
2. `rawContent` is non-empty, but `JsonUtils.isValidJsonObject(rawContent)` is false, so `content` becomes `"{}"`.
3. Result: `ToolUseBlock` with `input = {}` and `content = "{}"`. The raw string is gone and this code path produces no log line at all.

The tool then fails with its own missing-parameter error, which is a misleading dead end.

## Expected behavior

The `{}` fallback itself is reasonable — the comment at lines 121-123 explains it exists
to prevent persisting malformed JSON when streaming is interrupted. The problem is that
the failure is completely unobservable:

1. Minimum: log a `warn` in the catch handler, including the raw string and the underlying
   exception. A `catch (Exception ignored)` that silently drops a parse error makes this
   whole class of failures effectively undiagnosable.
2. Discussion: expose the raw arguments string somewhere reachable for diagnostics without
   polluting the persisted `content` — e.g. a dedicated field on `ToolUseBlock`, or via its
   existing metadata, so applications can retry, repair, or at least show the user what the
   model actually emitted.

## Environment

- agentscope-core 2.0.1 (also verified in 2.0.2 and on main as of 2026-08-25)
- JDK 21, Qwen via OpenAI-compatible endpoint, streaming tool calls

## Additional context

Malformed tool-call arguments are a recurring pain point here, but nobody has raised the
diagnosability angle:

- #847 — feature request for built-in multi-stage repair of malformed arguments (its
  examples even include curly quotes). We deliberately do *not* ask for that here: we
  repair app-side via the existing `JsonUtils.setJsonCodec()` extension point. The
  blocker is that the default path gives no observable signal at all.
- #2648 — describes the same `{}` fallback for arguments truncated by `finish_reason=length`;
  it likewise treats the fallback as existing behavior and asks for truncation detection.
- #1309 — stream=true malformed arguments (`unexpected end-of-input in property name`);
  different root cause (incomplete chunks), same silent-failure symptom.
- #2451 (merged) — made `build()` always parse the accumulated raw JSON and repair null
  input values, but only when the raw string is valid JSON; the invalid-JSON path is
  still silent.

Happy to submit a PR if you agree on the direction (warn log in the catch handler plus an
observable channel for the raw string).

mikemikimike:
## Summary

- warn when accumulated streamed tool-call arguments cannot be parsed as JSON
- preserve the existing `{}` content fallback and merged-argument behavior
- add regression coverage for malformed arguments emitted by a model

## Testing

- `mvn -pl agentscope-core -Dtest=ToolCallsAccumulatorTest test`
- `mvn -pl agentscope-core spotless:check`
- `git diff --check`

Closes #2841

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -3336,7 +3336,8 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                                                         toolCall,
                                                         ToolResultBlock.error(
                                                                 toolCall.getId(),
-                                                                "Tool execution rejected: malformed tool arguments")))
+                                                                "Tool execution rejected: malformed"
+                                                                        + " tool arguments")))
                                 .toList());
             }
             return dispatchToolCalls(toolCalls)
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -3338,6 +3338,18 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
          */
         private Mono<List<Map.Entry<ToolUseBlock, ToolResultBlock>>> executeToolCalls(
                 List<ToolUseBlock> toolCalls) {
+            if (toolCalls.stream().anyMatch(t -> t.getState() == ToolCallState.PARSE_FAILED)) {
+                return Mono.just(
+                        toolCalls.stream()
+                                .map(
+                                        toolCall ->
+                                                Map.entry(
+                                                        toolCall,
+                                                        ToolResultBlock.error(
+                                                                toolCall.getId(),
+                                                                "Tool execution rejected: malformed tool arguments")))
+                                .toList());
+            }
             return dispatchToolCalls(toolCalls)
                     .map(
                             results ->
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
@@ -23,6 +23,8 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Tool calls accumulator for accumulating streaming tool call chunks.
@@ -38,6 +40,8 @@ import java.util.stream.Collectors;
  * @hidden
  */
 public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
+
+    private static final Logger log = LoggerFactory.getLogger(ToolCallsAccumulator.class);
 
     // Map to support multiple parallel tool calls
     // Key: tool identifier (ID, name, or index)
@@ -113,8 +117,13 @@ public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
                             }
                         }
                     }
-                } catch (Exception ignored) {
-                    // Parsing failed, keep previously merged args
+                } catch (Exception e) {
+                    // Keep previously merged args, but retain enough context to diagnose the
+                    // malformed arguments emitted by the model.
+                    log.warn(
+                            "Failed to parse accumulated tool call arguments: {}",
+                            rawContentStr,
+                            e);
                 }
             }
 
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
@@ -124,8 +124,10 @@ public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
                     }
                 } catch (Exception e) {
                     state = ToolCallState.PARSE_FAILED;
-                    // Keep previously merged args, but retain enough context to diagnose the
-                    // malformed arguments emitted by the model.
+                    // Do not leave stale or partial arguments executable after final parsing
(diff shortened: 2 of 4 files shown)


gilgold:
The added observability is useful, but I think this exact form creates two safety problems:

1. `log.warn(... rawContentStr, e)` puts the full tool payload into default telemetry. Tool arguments routinely contain credentials, customer text, file contents, and prompt-injection material. I’d log only safe metadata by default: tool-call id/name, provider/model if available, byte length, SHA-256, parser exception class/offset, and stream finish reason. Any raw sample should require an explicit debug policy and be redacted + size-capped.
2. The new test codifies a more serious ambiguity: after the final JSON parse fails, previously merged `input` remains executable while `content` becomes `{}`. That can combine stale/partial fields with a malformed final payload and makes a parse failure indistinguishable from a genuinely valid empty object downstream.

A minimal fail-closed contract would be:

- mark the block with a typed `PARSE_FAILED` outcome (or metadata flag) distinct from valid `{}` and `INTERRUPTED`;
- preserve raw arguments only in an ephemeral caller-controlled diagnostic channel, never checkpoints/default logs;
- prevent tool execution whenever final parse failed;
- if repair is enabled: strict parse → conservative deterministic repair → tool-schema validation → execute. Unescaped quotes are ambiguous, so if schema validation cannot prove one reconstruction, return the typed error or make one bounded model retry rather than guessing;
- change the regression to assert malformed input cannot reach the executor as retained partial args, while valid `{}` still can.

For AgentScope, this belongs locally in the accumulator/`JsonCodec` seam. Disclosure: I maintain [JSONSuture](https://gilgold.github.io/jsonsuture-api/?source=github), a hosted option for teams that need one repair + schema-validation boundary across Java/Python/Node services; it is not a replacement for the framework-level fix here.

mikemikimike:
Addressed in 6106aa9e: malformed tool-call diagnostics now log only tool id/name, UTF-8 byte length, SHA-256 digest, exception type, and message. The raw argument payload is no longer emitted to default logs. Existing {} fallback and merged-argument behavior remain unchanged. Focused tests (16/16), module tests, Spotless, and git diff --check pass. The earlier full CI failure was unrelated, in agentscope-harness teardown.

gilgold:
Thanks — `6106aa9e` resolves the default-log payload exposure cleanly.

The execution-integrity issue is still present, though, and the new regression currently locks it in: `input.path=output.txt` remains executable even though the final accumulated payload is malformed and `content` is emitted as `{}`. Unless a later layer has an independent parse-failure guard, `write_text_file` can still run with stale/partial arguments while downstream code cannot distinguish that state from a valid empty object.

A minimal follow-up can stay local and avoid choosing a repair policy here:

- set a typed `PARSE_FAILED` outcome/metadata flag in the catch path;
- clear or quarantine executable `input` for that block;
- have the executor reject `PARSE_FAILED` before tool dispatch;
- change this regression to assert malformed final JSON is non-executable, while a genuinely valid `{}` remains executable.

If changing `ToolUseBlock` is too broad for this PR, a dedicated follow-up issue is reasonable, but I would not merge the retained-input test as the intended safety contract.

mikemikimike:
Addressed the remaining execution-integrity concern in `f71d92ca`:

- malformed final JSON now clears any previously merged partial input before building the block;
- the typed `PARSE_FAILED` state remains set, so `ReActAgent` rejects the call before tool dispatch;
- valid `{}` remains `PENDING` and keeps the existing executable empty-input behavior;
- fixed the existing `ReActAgent` Spotless violation from the prior commit.

Local validation with JDK 21:
- `mvn -pl agentscope-core -Dtest=ToolCallsAccumulatorTest test -Dcheckstyle.skip=true` — 17 passed
- `mvn -pl agentscope-core spotless:check -Dcheckstyle.skip=true` — passed
- `git diff --check` — passed

PR #2846, opened 2026-08-26
fix(sandbox): serialize same-key concurrent calls to prevent persisted state races

### LINKED ISSUE #2845: [Bug]:Same-isolation-key concurrent calls race on sandbox state slot
Description
  
  On a single HarnessAgent, two concurrent calls that resolve to the same SandboxIsolationKey operate on the
  same persisted sandbox state record. The sandbox acquire → resume → persist → release window runs outside
  the delegate's serializeOnKey gate, and the sandbox-layer guard defaults to noop() — so there is no
  serialisation.

  Both calls therefore start independent containers from the same state, mutate them in parallel, and
  overwrite each other on completion. Workspace changes from one call are silently lost (last write wins),
  with a transient double-container in between.

  Expected behaviour
  
  Concurrent calls resolving to the same isolation key should serialise and hand off state sequentially. No
  call's workspace changes should be clobbered by a concurrent peer.

  Actual behaviour
  
  Calls run fully in parallel on the shared state slot; whichever call persists last wins, and the other
  call's changes are dropped.

  Steps to reproduce
  
  1. Build a HarnessAgent with the default scope (or explicit GLOBAL / AGENT).
  2. Fire two concurrent calls that resolve to the same isolation key, each mutating the sandbox workspace
  (e.g. writing a file).
  3. Inspect the final persisted state — one call's changes have been overwritten and lost.

  Affected scopes

  Any two concurrent calls that resolve to the same persisted state slot:

  - SESSION — same sessionId
  - USER (default) — same userId
  - AGENT — same agentId (fixed at build time)
  - GLOBAL — constant key "__global__", shared by all calls (easiest to hit)

  Notes
  
  - This is not covered by #2490. That fix isolates the in-memory live binding for different-key concurrent
  calls; it does not prevent same-key calls from racing on the persisted state.
  - serializeOnKey does not help here: its gate key (userId, sessionId) protects ReActAgent's session state
  and runs inside the delegate, whereas the sandbox lifecycle is wrapped outside it.

  Possible direction (non-binding)

  Serialise same-key concurrent calls at the sandbox layer — e.g. a JVM-local default guard keyed by
  SandboxIsolationKey. Multi-instance deployments would need a distributed guard (Redis SET NX, etc.).
  Implementation details TBD in the PR.

larry-zy:
Closes #2845

## Background

On a single `HarnessAgent`, two concurrent calls that resolve to the **same `SandboxIsolationKey`** read and write the **same persisted sandbox state record**.

The sandbox `acquire → resume → persist → release` window runs **outside** the delegate's `serializeOnKey` gate, and the sandbox-layer guard defaulted to `noop()` — so there was no serialisation. Both calls started independent containers from the same state, mutated them in parallel, and overwrote each other on completion: **workspace changes are silently lost (last write wins)**, with a transient double-container in between.

This is **not** covered by #2490. That fix isolates the in-memory *live binding* for **different-key** concurrent calls; it does not stop **same-key** calls from racing on the *persisted state*.

## Changes

- **Add `InProcessSandboxExecutionGuard` (JVM-local default guard):** serialises per `SandboxIsolationKey`, one fair `Semaphore(1)` per key. `Semaphore` rather than `ReentrantLock` because acquire happens in the reactive resource supplier and release in the corresponding cleanup, which may run on different threads — `Semaphore` is not thread-ownership bound.
- **Flip the default guard from `noop()` to `inProcess()`** (`HarnessAgent`): single-instance deployments serialise out of the box; multi-instance deployments can still supply a distributed guard via `SandboxFilesystemSpec#executionGuard` or a `DistributedStore`.
- **Move sandbox acquire/release onto a `boundedElastic` thread:** the guard now genuinely blocks (potentially for the full duration of a same-slot peer call), so it must never tie up a WebFlux event-loop thread and starve unrelated sessions.
- **Release the lease in a `finally`** (`SandboxLifecycleMiddleware`): `lease.close()` is guaranteed regardless of what persist/release throw, preventing a permit leak that would block every future same-slot call forever; also restores the interrupt flag cleared by `InterruptedException`.
- **Add `inProcess(Duration)`** as a backstop against a wedged holder, throwing `SandboxExecutionTimeoutException` on expiry. This is a wedged-holder backstop, not a contention timeout — set it well above the maximum realistic call duration.

## Tests

- `InProcessSandboxExecutionGuardTest` — queued serialisation, timeout, no slot leak, cross-thread release.
- `SandboxSameSessionSerializationTest` — same-key concurrent calls no longer overwrite each other end-to-end.
- `HarnessAgentSandboxOffThreadTest` — acquire/release never run on the subscriber thread.

## Deployment note

This guard coordinates within a **single JVM** only. When the same slot can be contended across multiple instances, supply a distributed implementation (e.g. Redis `SET NX`).


agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -890,22 +891,67 @@ public Toolkit getToolkit() {
...
+                    }
+                });
+    }
+
+    /**
+     * Runs the (potentially blocking) {@code release} on a boundedElastic thread. Package-private so
+     * a test can assert the release never runs on the subscriber's thread either.
+     */
+    static Mono<Void> releaseOffThread(Runnable release) {
+        return Mono.<Void>fromRunnable(release).subscribeOn(Schedulers.boundedElastic());
```
guslegend0510, line 1063: [P1] boundedElastic starvation deadlock

Blocking same-key acquire waiters and the only lease-release cleanup share the same boundedElastic pool. Once waiters consume its thread cap, the holder's cleanup cannot run, so the permit is never released. Use a non-blocking keyed gate or ensure release cannot queue behind blocking waiters.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -890,22 +891,67 @@ public Toolkit getToolkit() {
...
-                        },
+                Mono.usingWhen(
+                        acquireSandboxOffThread(effective),
                         eff -> inner.get(),
-                        eff -> {
-                            if (sandboxLifecycleMw != null) {
-                                sandboxLifecycleMw.releaseForCall(eff);
-                            }
-                        });
+                        this::releaseSandboxOffThread);
```
guslegend0510, line 1072: [P1] Cancellation can leak acquired resources

Cancellation before the resource supplier emits does not invoke usingWhen cleanup. If acquireForCall has obtained the lease and create/start ignores interruption, the RuntimeContext is discarded while the sandbox and permit remain held permanently. Add cancellation-safe acquisition compensation and a cancel-during-start regression test.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2335,10 +2363,14 @@ public HarnessAgent build() {
...
                         new SessionSandboxStateStore(effectiveSession, resolvedAgentId);
+                // Default to a JVM-local guard so same-slot concurrent calls (e.g. two requests on
+                // one sessionId) serialise their acquire/persist window instead of racing on the
+                // shared state slot (issue #2800). Multi-instance deployments override this with a
+                // distributed guard via the spec or a DistributedStore.
                 SandboxExecutionGuard executionGuard =
                         sandboxFilesystemSpec.getExecutionGuard() != null
                                 ? sandboxFilesystemSpec.getExecutionGuard()
-                                : SandboxExecutionGuard.noop();
+                                : SandboxExecutionGuard.inProcess();
```
guslegend0510, line 2539: [P1] DistributedStore noop bypasses the new default

DistributedStore auto-wiring can populate the spec with its default noop guard before this fallback is evaluated. COS, OSS, and an unconfigured composite store therefore still race on same-key persisted state. Preserve an unset state or otherwise distinguish an absent distributed guard from an explicit noop.

guslegend0510 (changes requested):

muranchenhui (approved):
LGTM

muranchenhui (approved):
LGTM

guslegend0510 (approved):
LGTM

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -939,22 +943,137 @@ public Toolkit getToolkit() {
...
                             }
-                            return effective;
-                        },
-                        eff -> inner.get(),
-                        eff -> {
-                            if (sandboxLifecycleMw != null) {
-                                sandboxLifecycleMw.releaseForCall(eff);
+                            return ctx;
+                        })
+                .subscribeOn(SANDBOX_ACQUIRE_SCHEDULER)
```
oss-maintainer: CI is red on the current head (`b50e0dbf`): `Java CI with Maven` fails with `HarnessAgentTest.executionHandleCancelsTheHarnessAndReleasesItsSandboxBinding:667 expected: <RUNNING> but was: <QUEUED>` (main is green at `3c1c29c0`), and `HarnessAgentTest.java` is not among this PR's 11 files, so the break is caused by this change rather than by a stale test in the branch.

The cause is this line. Before the change `Mono.using` acquired synchronously on the subscriber's thread, so once `subscribe(...)` returned, `run.status()` was already `RUNNING` and `SandboxAcquireResult` was already bound on the context. `subscribeOn(SANDBOX_ACQUIRE_SCHEDULER)` turns that hand-off asynchronous, so the same call now observes `QUEUED`.

That is an observable contract change, not only a test artifact: any caller that treats 'subscribe returned' as 'the run is live and its sandbox is bound' now races the acquire. Please either restore a synchronous status flip (flip to RUNNING before scheduling the acquire, and surface a distinct ACQUIRING/QUEUED state only while the guard is contended if that distinction is wanted), or update the test deliberately and state in the description that the transition is now asynchronous. A red CI plus an un-examined guarantee is the worst of both outcomes. A same-session concurrency test asserting the guard is already held when `subscribe` returns would settle which behaviour is intended.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2393,7 +2494,12 @@ public HarnessAgent build() {
                     if (sandboxFilesystemSpec.getSnapshotSpecOverride() == null) {
                         sandboxFilesystemSpec.snapshotSpec(distributedStore.sandboxSnapshotSpec());
                     }
-                    if (sandboxFilesystemSpec.getExecutionGuard() == null) {
+                    // Only adopt the store's guard when it actually supplies one. A store that
+                    // returns null is opting out, NOT requesting a no-op guard — leaving the spec
+                    // null lets the inProcess() default below still serialise same-key calls
+                    // (issue #2800). Injecting a noop here would silently suppress that default.
+                    if (sandboxFilesystemSpec.getExecutionGuard() == null
+                            && distributedStore.sandboxExecutionGuard() != null) {
```
oss-maintainer: `distributedStore.sandboxExecutionGuard()` is invoked **twice** here — once for the null check, once for the value that is installed. The `DistributedStore` SPI promises no 'same instance per call' identity, and at least one in-repo implementation builds a fresh guard every invocation (`RedisDistributedStore.sandboxExecutionGuard()` → `RedisSandboxExecutionGuard.builder(jedis).keyPrefix(...).build()`). So the instance the guard clause validates is not necessarily the instance the agent ends up using, and stores that allocate in the accessor pay for it twice. Please hoist it:

```java
SandboxExecutionGuard storeGuard = distributedStore.sandboxExecutionGuard();
if (sandboxFilesystemSpec.getExecutionGuard() == null && storeGuard != null) {
    sandboxFilesystemSpec.executionGuard(storeGuard);
}
```

That also puts the new 'null means opt out, not no-op' contract in one readable place.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -939,22 +943,137 @@ public Toolkit getToolkit() {
...
+     * Dedicated scheduler for the blocking guard acquire, kept separate from {@link
+     * Schedulers#boundedElastic()} on purpose. The guard's {@code semaphore.acquire()} can park a
+     * thread for the entire duration of a same-slot peer call (issue #2800); if those long-parked
+     * waiters shared the pool that runs the holder's release, they could consume its whole thread
+     * cap and the release that frees the permit would queue behind them forever — a starvation
+     * deadlock. Releasing stays on {@code boundedElastic}, so it can always make progress and wake a
+     * waiter here. Daemon threads, shared JVM-wide like {@code boundedElastic}, so no disposal.
+     */
+    private static final Scheduler SANDBOX_ACQUIRE_SCHEDULER =
+            Schedulers.newBoundedElastic(
```
oss-maintainer: Splitting acquire from `boundedElastic` protects the *release* path, but waiters on one wedged key still occupy this pool's fixed thread cap (`DEFAULT_BOUNDED_ELASTIC_SIZE`), and tasks for **unrelated** keys queue behind them in the same bounded queue. Combined with the default guard waiting indefinitely (`waitTimeout == null` → `semaphore.acquire()`), a holder that never releases (killed container, lost lease) parks threads until the pool saturates and then stalls sandbox acquires JVM-wide — the same starvation class this scheduler exists to avoid, moved one layer down.

Consider a bounded default wait (or a documented 'a wedged holder degrades unrelated keys after N threads park' note), and/or per-key admission that does not consume a worker thread while waiting. The trade-off is real either way — an indefinite wait is what makes same-key fairness useful — so this is about making the failure mode explicit rather than demanding a timeout.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
```diff
@@ -124,20 +124,31 @@ public void acquireForCall(RuntimeContext ctx) {
...
             }
         } catch (Exception e) {
             log.error("[sandbox-mw] Failed to acquire/start sandbox", e);
+            // Throwing InterruptedException clears the thread's interrupt flag; restore it so
+            // callers up the stack can still observe the interrupt (the guard's tryEnter blocks
+            // interruptibly while waiting for a busy slot — issue #2800).
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
             throw new RuntimeException(e);
```
oss-maintainer: Wrapping in a bare `RuntimeException(e)` erases the new typed failure. `SandboxExecutionTimeoutException` already extends `RuntimeException`, so a caller who configured `inProcess(Duration)` specifically to tell 'wedged slot' apart from 'broken sandbox' now receives `RuntimeException(cause=SandboxExecutionTimeoutException)` and has to walk the cause chain to branch on it. Please rethrow `SandboxExecutionTimeoutException` (and `RuntimeException` generally) as-is and wrap only checked exceptions. Same point for the interrupt path just above: restoring the flag is correct, but the type the caller sees is still opaque.

agentscope-harness/src/main/java/io/agentscope/harness/agent/IsolationScope.java
```diff
@@ -47,8 +47,12 @@
  * </ul>
  *
  * <p><b>Concurrency note:</b> for sandbox mode this is sequential-reuse sharing, not
- * live-instance sharing. Concurrent calls at the same scope each get their own running container;
- * they converge on the last persisted snapshot at the end of the call.
+ * live-instance sharing. By default a JVM-local
```
oss-maintainer: Good doc update, and it makes clear this is a default **behaviour change**, not only a bug fix: previously 'concurrent calls at the same scope each get their own running container', now they serialise per slot key. For SESSION/USER/AGENT-scoped deployments that fan out concurrent calls onto one key, per-call wall clock can grow by the number of queued peers, and `docs/v2/en/docs/harness/sandbox.md` currently documents `executionGuard` as opt-in. Worth a migration note in the changelog plus that doc page so the flip is discoverable by anyone who does not read the javadoc diff.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -939,22 +943,137 @@ public Toolkit getToolkit() {
...
-                                sandboxLifecycleMw.releaseForCall(eff);
+                            return ctx;
+                        })
+                .subscribeOn(SANDBOX_ACQUIRE_SCHEDULER)
+                .doOnCancel(
+                        () -> {
+                            // Cancel: publish bit 1. If acquire already set bit 0, the resource is
+                            // now orphaned (usingWhen registered no cleanup for it) — release it.
+                            if ((state.getAndUpdate(v -> v | 0b10) & 0b01) != 0) {
+                                compensate.run();
```
oss-maintainer: The 2-bit reconcile reads as correct to me: a `Mono.fromCallable` source that has already emitted cannot take a later cancel signal, so exactly one of the two branches runs `compensate`, and `releaseForCall` is idempotent via the per-call `SandboxAcquireResult` clear. One asymmetry though — on this branch `compensate.run()` performs the release **inline on the cancelling thread**, whereas the normal path deliberately routes release through `releaseOffThread` because persist + container stop + `lease.close()` block. If a caller cancels from an event-loop thread, this orphan-release reintroduces blocking there. Consider `releaseOffThread(() -> ...)` on this path too.

oss-maintainer (commented):
### Summary
Fixes a real and well-diagnosed race: with the default `noop()` guard, two concurrent calls resolving to the same `SandboxIsolationKey` both resumed from the same persisted state and last-writer-wins on `persistState`. The `InProcessSandboxExecutionGuard` design is careful — refcounted slot recycling under `compute`, a fair `Semaphore(1)` rather than a `ReentrantLock` because release lands on a different thread, and an idempotent lease — and the 700 lines of added tests across three files target exactly the window being closed. The lease-close-in-`finally` pair in `SandboxLifecycleMiddleware` is the right shape: a permit that leaks on a rollback path wedges the slot permanently.

Two things block approval. First, CI is red on this head and the failure is caused by the change, not by a stale branch: moving the acquire to `subscribeOn(...)` makes the status flip asynchronous and `HarnessAgentTest` asserts it synchronously (detailed at HarnessAgent.java:1034). That is also a user-visible contract question, so it needs a deliberate answer rather than a rebase to see. Second, the guard accessor is called twice in `build()`, which matters for implementations that construct a new guard per call.

The rest is about failure modes and reachability: an indefinite default wait on a bounded pool turns one wedged holder into a JVM-wide acquire stall, and wrapping everything in `RuntimeException(e)` hides the `SandboxExecutionTimeoutException` that the same PR introduces for precisely that diagnosis. Flipping `noop()` → `inProcess()` is also a behaviour change for anyone fanning out concurrent calls on one key, so it deserves a migration note next to the javadoc.

### Findings
- **[Critical]** `HarnessAgent.java:1034` — off-thread acquire breaks the synchronous RUNNING-after-subscribe guarantee; `HarnessAgentTest:667` fails on this head while main is green.
- **[Warning]** `HarnessAgent.java:2502` — `sandboxExecutionGuard()` invoked twice; null-checked instance need not be the installed instance (Redis builds one per call).
- **[Warning]** `HarnessAgent.java:956` — indefinite default wait parks the fixed-size acquire pool; one wedged holder stalls unrelated keys.
- **[Warning]** `SandboxLifecycleMiddleware.java:152` — `RuntimeException(e)` wrapping erases `SandboxExecutionTimeoutException` / interrupt typing.
- **[Info]** `IsolationScope.java:50` — default flip is a behaviour change; add a changelog + doc migration note.
- **[Info]** `HarnessAgent.java:1040` — cancel-path release runs inline on the cancelling thread while the normal path releases off-thread.

---

Issue #2850, opened 2026-08-26
[Enhancement]: LocalFilesystemWithShell.execute — bound capture during drain, Future-based teardown, dual-stream test


zzz-ghost:
Follow-ups from the review of #2839 (which fixed the pipe deadlock by draining stdout/stderr concurrently with `waitFor`):

1. **Bound memory while draining, not after.** `maxOutputBytes` is applied post-hoc to the assembled string, so a huge output can OOM before truncation runs. Cap `buf.write` once the buffer hits `maxOutputBytes`, keep reading-and-discarding so the OS pipe stays drained, and mark the result `truncated` when bytes were dropped.

2. **Align reader teardown with `ShellCommandTool`.** `join(timeout)` may return while the drainer thread is still alive, and `ByteArrayOutputStream` is not thread-safe, so `toString()` can race with `write`. Prefer `Future.get(timeout)` + `cancel(true)` (discarding the untrusted buffer), or interrupt the drainer when it is still alive after join.

3. **Add a stderr / dual-stream regression test** alongside the stdout one added in #2839.

Happy to send the PR(s) once #2839 is merged.

oss-maintainer:
**Issue Evaluation**
Category: `enhancement` | Area: `area/harness` | Status: **Evaluated**

**Feasibility:** Feasible, well-scoped. All three improvement points are confirmed on current `main`:

1. **Unbounded drain** — `drainAsync()` (LocalFilesystemWithShell.java:464-481) writes into `ByteArrayOutputStream` at line 472 with no size cap; `maxOutputBytes` is only checked post-hoc on the merged string at line 404. Capping `buf.write` during drain and marking `truncated` is straightforward.
2. **Thread teardown** — `joinQuietly` (line 483-489) calls `join(5000ms)` but never interrupts the drainer on timeout. `ShellCommandTool` (line 691) uses `Future.cancel(true)` as the reference pattern. Switching to `Future.get(timeout)` + `cancel(true)` aligns the two code paths.
3. **Missing test coverage** — `LocalFilesystemWithShellTest` has only one `execute` test (line 50) which produces stdout only. No stderr or dual-stream regression test exists.

**Scope:** Confined to `LocalFilesystemWithShell.java` and its test file. No API changes.

**Compatibility:** Internal implementation only; `ExecuteResponse` contract unchanged.

*Automated evaluation by github-manager-bot*

PR #2853, opened 2026-08-26
fix(harness): bound LocalFilesystemWithShell output capture

guslegend0510:
## Summary

Follow up on #2839 and address the output-capture lifecycle issues tracked in #2850.

- Bound stdout and stderr capture while continuing to drain excess bytes.
- Replace raw drainer-thread joins with `Future`-based completion.
- Use one shared teardown deadline for both output streams.
- Close process streams and cancel unfinished reader tasks on timeout or interruption.
- Preserve exit code `124` for command timeouts and restore the caller's interrupt flag.
- Add large stdout, stderr, dual-stream, bounded-capture, timeout, and interruption regression tests.
- Add best-effort cleanup for snapshotted command descendants.

## Dependency

This PR depends on #2839.

Because #2839 has not merged yet, this branch currently contains its prerequisite pipe-deadlock commit. After #2839 is merged, the branch will be rebased and the duplicate prerequisite commit will be dropped.

## Validation

- `LocalFilesystemWithShellTest`: 13 tests passed
- `agentscope-core`: 2294 passed, 9 skipped
- `agentscope-harness`: 851 passed, 3 skipped
- Spotless and `git diff --check` passed

## Draft follow-ups

Before marking this PR ready for review:

- Propagate unexpected stdout/stderr `IOException`s instead of treating partial capture as successful.
- Clarify or reduce the best-effort process-tree cleanup scope, since `ProcessHandle.descendants()` cannot guarantee cleanup after the parent has exited or descendants have been reparented.

Refs #2850  
Depends on #2839


CryoThrust:
I found one lifecycle edge case in the current cleanup path:

`execute_shellExitWithBackgroundPipeHolderReturnsPromptly` demonstrates that the shell can exit while a background descendant keeps stdout/stderr open. In that situation `awaitCapture` reaches its deadline, closes/cancels the readers, and the method enters `finally`, but `destroyProcessTree` returns immediately when `process.toHandle()` is no longer alive. The descendant snapshot is therefore never taken and the background process can remain alive after `execute` returns. The test explicitly destroys the child in its own `finally`, so it does not catch this production leak.

Could the cleanup contract cover this parent-exited case, for example by retaining a descendant snapshot taken immediately after launch (before `waitFor`) and using it during final cleanup, or by documenting that only live-root descendants are in scope and adding a test that verifies the orphan case? The snapshot needs to happen before the parent exits; attempting to discover descendants only after root termination is too late because they may already be reparented.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -344,28 +371,26 @@ public ExecuteResponse execute(
                 pb.environment().putAll(env);
```
oss-maintainer, line 376: A descendant spawned after the initial snapshot can be reparented before cleanup when the shell exits quickly, so it is never rediscovered or terminated. This leaves background commands running after execute returns; use a process group/job object for lifecycle ownership rather than relying on a pre-wait snapshot.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -263,6 +278,10 @@ public LocalFilesystemWithShell(
         if (timeout <= 0) {
```
oss-maintainer, line 283: Rejecting maxOutputBytes == 0 changes the behavior of a public constructor: previously callers could use zero to suppress retained command output. Preserve zero as a valid no-capture limit, or explicitly version/document this breaking API change.

oss-maintainer (commented):
### Summary
Re-review of PR #2853 at its current head (automated review of the changes in this diff). 2 item(s) raised below; nothing here blocks the approach itself — details inline.

### Findings
- **[Warning]** `agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java:371` — A descendant spawned after the initial snapshot can be reparented before cleanup when the shell exits quickly, so it is never rediscovered or terminated. This leaves background commands running after execute returns; use a process group/job object for lifecycle ownership rather than relying on a pre-wait snapshot.
- **[Warning]** `agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java:278` — Rejecting maxOutputBytes == 0 changes the behavior of a public constructor: previously callers could use zero to suppress retained command output. Preserve zero as a valid no-capture limit, or explicitly version/document this breaking API change.

### Notes

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -400,15 +443,14 @@ public ExecuteResponse execute(
 
             String outputStr = output.isEmpty() ? "<no output>" : output.toString();
 
-            boolean truncated = false;
+            boolean truncated = stdoutBuf.truncated() || stderrBuf.truncated();
             if (outputStr.length() > maxOutputBytes) {
```
oss-maintainer: With `maxOutputBytes == 0` the synthetic `"<no output>"` sentinel is itself subject to the cap, so the zero case is not really "capture disabled": `outputStr.length()` (11) > 0 truncates the sentinel to `""` and sets `truncated = true` unconditionally, and `ShellExecuteTool` then renders `"(output was truncated)"` on top of `"... Output truncated at 0 bytes."` for *every* command — including commands that printed nothing. Suggest short-circuiting the zero case (return an empty/`"<no output>"` response with `truncated = false`) before the cap check, or excluding the sentinel from `outputStr.length() > maxOutputBytes`, so the flag keeps meaning "bytes were actually dropped".

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -241,7 +256,7 @@ public LocalFilesystemWithShell(
      * @param mode path-resolution policy ({@code null} treated as {@link LocalFsMode#UNRESTRICTED})
      * @param pathPolicy allow-list for {@link LocalFsMode#ROOTED}; ignored otherwise
      * @param timeout default shell timeout (seconds, must be positive)
-     * @param maxOutputBytes byte cap for captured shell output
+     * @param maxOutputBytes nonnegative byte cap for captured shell output; zero disables capture
```
oss-maintainer: The javadoc now promises a *byte* cap, but this final check compares `outputStr.length()` — UTF-16 chars of the merged `stdout` + `"[stderr] "`-prefixed string — against `maxOutputBytes`. The per-stream `BoundedCapture` does cap bytes correctly; this tail pass does not, so for non-ASCII output the returned string can still exceed the documented byte budget (and for `maxOutputBytes < 11` it is the sentinel length that trips it, which is what makes the zero case above look wrong). Worth using an encoded byte length (or applying the cap before merging the streams) now that the contract is spelled out here.

oss-maintainer (commented):
### Summary
Re-review after the new commit `aa9154db` ("fix(harness): preserve zero output capture limit"). The previous **[Warning]** about the public constructor rejecting `maxOutputBytes == 0` is **resolved the right way**: the guard is now `< 0`, the javadoc states the contract, and the two constructor tests were split into `acceptsZeroMaxOutputBytes` / `rejectsNegativeMaxOutputBytes` so the boundary is pinned. CI is green on this head (build ubuntu/windows, validate, Check License, Check Module Sync, codecov/patch, `license/cla` all pass).

The other item from my earlier review — `LocalFilesystemWithShell.java:371`, descendants observed only from the pre-`wait` snapshot can escape cleanup — is unchanged by this commit, so it stays open; it is now documented in `docs/v2/{en,zh}/harness/filesystem.md` as best-effort, which is the honest thing to do, but the discovery window itself is still the gap.

What is left is that the newly-blessed `0` value does not behave like "capture disabled" end-to-end — one inline finding below, plus a byte-vs-char note on the same tail path.

### Verdict
`COMMENT` — the compatibility regression is fixed; the zero-cap semantics above are worth tightening before merge but do not change my read that the overall approach (bounded per-stream capture + explicit cleanup) is sound.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -398,16 +446,20 @@ public ExecuteResponse execute(
...
-                        outputStr.substring(0, maxOutputBytes)
-                                + "\n\n... Output truncated at "
-                                + maxOutputBytes
-                                + " bytes.";
-                truncated = true;
+            boolean hasOutput = !output.isEmpty();
+            String outputStr = hasOutput ? output.toString() : "<no output>";
+            boolean truncated = stdoutBuf.truncated() || stderrBuf.truncated();
+            if (maxOutputBytes == 0) {
+                outputStr = "<no output>";
```
oss-maintainer: `maxOutputBytes == 0` forces `truncated = false`, so a command that wrote 10 MB comes back as the bare sentinel `"<no output>"` with no truncation marker (`ShellExecuteTool` only appends `"(output was truncated)"` when `result.truncated()` is true). The caller then cannot distinguish "the command printed nothing" from "capture was turned off and everything was dropped" — which is exactly the signal an agent needs before concluding a build/compile succeeded silently. The new test locks this in. Suggest keeping `truncated = true` for the zero-capture case (or a distinct sentinel such as `"<output capture disabled>"`) and documenting that the zero branch is lossy.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -398,16 +446,20 @@ public ExecuteResponse execute(
...
+                truncated = false;
+            } else {
+                if (hasOutput && outputStr.getBytes(outputCharset).length > maxOutputBytes) {
```
oss-maintainer: This measures the merged stdout+stderr string by encoding it in full (`getBytes` allocates a byte[] the size of the whole capture) purely to compare a length, and then `truncateToByteLimit` encodes it a second time. For a per-call hot path with a large cap that is one extra full-size allocation per `execute()`. Cheaper and single-pass: call the encoder with a `ByteBuffer.allocate(maxOutputBytes + 1)` and treat "no OVERFLOW" as "nothing to truncate" — the buffer size already proves the cap, and `input.position()` gives the split point when it does overflow.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -398,16 +446,20 @@ public ExecuteResponse execute(
...
+                    outputStr = truncateToByteLimit(outputStr, outputCharset, maxOutputBytes);
+                    truncated = true;
+                }
+                if (truncated) {
+                    outputStr += "\n\n... Output truncated at " + maxOutputBytes + " bytes.";
```
oss-maintainer, line 464: The suffix is appended *after* the byte cap is enforced, so the response is `maxOutputBytes + len("\n\n... Output truncated at N bytes.")` bytes; the same applies to the `"\n\nExit code: N"` suffix below. If the intent of this PR is a hard budget for what reaches the model context, the budget is still exceeded by a predictable ~40 bytes. Reserve the marker up front (truncate to `maxOutputBytes - suffixBytes`) or state in the javadoc that the cap covers captured payload only, excluding the trailer.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -429,9 +481,37 @@ public ExecuteResponse execute(
...
+        }
+        CharBuffer input = CharBuffer.wrap(value);
+        ByteBuffer output = ByteBuffer.allocate(maxBytes);
+        try {
+            CoderResult result =
+                    charset.newEncoder()
+                            .onMalformedInput(CodingErrorAction.REPLACE)
+                            .onUnmappableCharacter(CodingErrorAction.REPLACE)
+                            .encode(input, output, true);
+            if (result.isError()) {
```
oss-maintainer: With `onMalformedInput(REPLACE)` / `onUnmappableCharacter(REPLACE)`, `CoderResult.isError()` can never be true, so `throwException()` and the `CharacterCodingException` catch are dead paths. The result that actually happens on a full buffer is `OVERFLOW`, which is deliberately ignored and relied upon via `input.position()`. That is correct, but it reads as an oversight: add a short comment that OVERFLOW is the expected terminal result, and drop or replace the unreachable error handling. Also worth a test that asserts the *byte* cap holds for a long multi-byte run (the new test only checks 2-char boundaries) and one for the `maxOutputBytes == 0` + nonzero-exit-code combination.

oss-maintainer (commented):
Delta re-review after `aa9154db` → `cdba9143`. The byte-vs-char bug from my previous review is fixed properly: the merged output is now measured in bytes of `outputCharset` instead of `String.length()`, and `truncateToByteLimit` cuts on a character boundary. CI is green (build ubuntu/windows, Check License, Check Module Sync, validate, codecov/patch) and CLA is signed on `cdba9143`.

Four remaining points, all about the contract rather than the fix itself:

- `maxOutputBytes == 0` now means "capture nothing", and because `BoundedCapture(0)` already retains zero bytes, the *only* effect of that branch is to reset `truncated` to `false`. The caller therefore sees a plain `<no output>` for a command that wrote megabytes — indistinguishable from a genuinely silent command, and `ShellExecuteTool` suppresses its `(output was truncated)` note. That is the signal an agent uses before concluding "the build printed nothing", so I would keep the flag set, or use a distinct sentinel such as `<output capture disabled>`.
- The truncation trailer (and the `Exit code:` trailer) is appended after the cap is applied, so the response is still a few dozen bytes over `maxOutputBytes`. Reserve room for the trailer, or scope the javadoc to captured payload.
- The cap is checked by encoding the whole merged string with `getBytes`, then encoded again inside `truncateToByteLimit`; a single pass with `maxOutputBytes + 1` as the buffer bound avoids both the extra full-size allocation and the second encode.
- With `CodingErrorAction.REPLACE` on malformed input and unmappable characters, `CoderResult.isError()` can never be true, so `throwException()`/the `CharacterCodingException` catch are unreachable while the result that actually occurs (`OVERFLOW`) is silently relied upon. Worth an explicit comment, plus a byte-cap test over a long multi-byte run (the new test covers only a 2-character boundary).

Not blocking — flagging for the author's call, and I will re-review on the next head.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -241,7 +260,8 @@ public LocalFilesystemWithShell(
      * @param mode path-resolution policy ({@code null} treated as {@link LocalFsMode#UNRESTRICTED})
      * @param pathPolicy allow-list for {@link LocalFsMode#ROOTED}; ignored otherwise
      * @param timeout default shell timeout (seconds, must be positive)
-     * @param maxOutputBytes byte cap for captured shell output
+     * @param maxOutputBytes nonnegative byte cap for captured stdout/stderr payload; diagnostic
+     *     trailers are excluded, and zero disables capture
```
oss-maintainer, line 264: **[Warning]** The javadoc now advertises `zero disables capture`, but `0` is not reachable through the configuration surface this PR also documents: `LocalFilesystemSpec.maxOutputBytes(int)` still throws for `bytes <= 0` (`maxOutputBytes must be positive, got 0`), and `docs/v2/{en,zh}/docs/harness/filesystem.md` — touched by this PR — still presents `maxOutputBytes(int)` as *the* knob with default 100,000 and no zero case.

So today the branch at line 453 is only reachable via the public constructor, i.e. the documented contract and the documented API disagree. Suggest one of:
- accept `0` in `LocalFilesystemSpec.maxOutputBytes(int)` and add a `0 = disable capture` row/note to both language docs, or
- keep the spec strict and scope the javadoc to "the constructor accepts `0` to disable capture; the spec requires a positive cap".

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -398,16 +446,23 @@ public ExecuteResponse execute(
...
-                        outputStr.substring(0, maxOutputBytes)
-                                + "\n\n... Output truncated at "
-                                + maxOutputBytes
-                                + " bytes.";
-                truncated = true;
+            boolean hasOutput = !output.isEmpty();
+            String outputStr = hasOutput ? output.toString() : "<no output>";
+            boolean truncated = stdoutBuf.truncated() || stderrBuf.truncated();
+            if (maxOutputBytes == 0) {
+                outputStr = "<output capture disabled>";
```
oss-maintainer, line 453: **[Info]** `<output capture disabled>` is the right fix for the ambiguity I raised earlier — a caller can now distinguish "capture off" from "command was silent", and the new `Exit code: 7` assertion pins the disabled + non-zero-exit combination.

Two follow-ups so the sentinel does not become an implicit contract:
- `ShellExecuteTool` keys only off `truncated()`, so it is unaffected, but nothing in this PR records the sentinel string as part of the observable `ExecuteResponse` contract. A one-line note next to the `maxOutputBytes(int)` docs row would keep downstream prompt/snapshot assertions honest.
- Consider asserting the same sentinel in a `LocalFilesystemSpec`-driven test once (if) zero becomes spec-reachable, so the two paths cannot drift.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -429,9 +484,38 @@ public ExecuteResponse execute(
...
+                (long) Math.ceil((double) encoder.maxBytesPerChar() * value.length());
+        int capacity = (int) Math.min(maxBytes, Math.min(Integer.MAX_VALUE, encodedUpperBound));
+        ByteBuffer output = ByteBuffer.allocate(capacity);
+        CoderResult result = encoder.encode(input, output, true);
+        if (result.isUnderflow()) {
+            return value;
+        }
+        // OVERFLOW is expected when the payload reaches the byte cap; input.position() remains
+        // on a valid character boundary.
+        return value.substring(0, input.position());
```
oss-maintainer, line 518: **[Info]** Single-pass encoding with an over-allocated bound is a clear improvement over `getBytes()` + re-encode, and dropping the unreachable `CharacterCodingException` path is correct now that `OVERFLOW` is the branch actually relied on.

One precision nit: `encode(input, output, true)` reports `UNDERFLOW` when the *input* is exhausted, but for stateful charsets the trailing `flush()` bytes are never emitted here, so `return value` on underflow can exceed `maxBytes` by the flush tail (relevant only for e.g. ISO-2022 charsets; UTF-8 and the Windows code pages `outputCharset` selects are stateless). Either call `flush()` into the same buffer before the underflow return, or note in the comment that the cap is exact for stateless charsets only.

oss-maintainer (commented):
### Summary

Delta re-review after `cdba9143` → `3f542ae7` (`fix(harness): clarify disabled output capture`).

All four contract points from my previous review are addressed properly, and the fix is real, not cosmetic:

- The disabled-capture ambiguity is gone — `maxOutputBytes == 0` now returns the distinct `<output capture disabled>` sentinel instead of `<no output>`, so a caller can tell "capture off" from "command was silent". The new assertion also pins the `<sentinel>\n\nExit code: 7` combination.
- The cap is no longer a documented fiction: the javadoc now says the byte cap covers the captured payload and excludes diagnostic trailers, which is what the code actually produces.
- `truncateToByteLimit` is single-pass: `capacity` is clamped by `min(maxBytes, ceil(maxBytesPerChar * length()))`, so there is no whole-payload `getBytes()` and no second encode. The `long` upper bound also removes the int-overflow risk in that sizing.
- `CoderResult.OVERFLOW` is now an explicit, commented expectation instead of an unreachable `CharacterCodingException` path, and the byte cap is tested over a long multi-byte run (100 × `中` at a 31-byte cap) rather than only a 2-character boundary.

Remaining items are non-blocking: the `0 = disable capture` contract is not reachable through `LocalFilesystemSpec.maxOutputBytes(int)`, which still rejects non-positive values, and is not reflected in the `maxOutputBytes(int)` docs rows this PR touches (inline at `LocalFilesystemWithShell.java:264`); plus two precision notes on the sentinel's contract surface and on `UNDERFLOW` for stateful charsets.

Not blocking — leaving as `COMMENT` rather than `APPROVE` because CI is red at this head: `build (windows-latest)` fails in `agentscope-extensions-aistio` (`HttpSelfRegistrationTest…:106 expected: <true> but was: <false>`) and `build (ubuntu-latest)` was cancelled. Neither module is touched by this PR (4 files, harness + docs), and `60934595a` on `main` is green, so this reads as an unrelated aistio flake — a re-run should confirm it. CLA is signed on `3f542ae7`.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/spec/LocalFilesystemSpec.java
```diff
@@ -109,12 +109,15 @@ public LocalFilesystemSpec executeTimeoutSeconds(int seconds) {
...
+     * Executions then return {@code <output capture disabled>} instead of captured output.
+     *
+     * @param bytes byte cap (must be nonnegative)
      * @return this spec
      */
     public LocalFilesystemSpec maxOutputBytes(int bytes) {
-        if (bytes <= 0) {
-            throw new IllegalArgumentException("maxOutputBytes must be positive, got " + bytes);
+        if (bytes < 0) {
+            throw new IllegalArgumentException("maxOutputBytes must be nonnegative, got " + bytes);
```
oss-maintainer, line 120: This relaxation changes what an existing call site means: `maxOutputBytes(0)` used to be a loud `IllegalArgumentException`, and it is now "silently discard every byte of stdout/stderr". The risk is not the explicit `0` in the new test — it is a spec assembled from a plain `int` config field, where `0` is also the Java zero value for an unset or typo'd property. That caller flips from fail-fast to "the tool returned no output and I cannot tell whether the command was broken", which is exactly the ambiguity the `<output capture disabled>` sentinel was introduced to remove one layer down.

If the `0`-means-off contract is intended here (it is documented now, and the sentinel keeps it distinguishable), then the remaining gap is disclosure: this is a silent semantic change to a public builder setter, so it is worth one line in the v2 migration/changelog notes. The alternative is an explicit opt-in (e.g. `disableOutputCapture()`, or rejecting `0` in the spec while the lower-level constructor keeps accepting it), which keeps the zero value an error.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -429,9 +484,40 @@ public ExecuteResponse execute(
...
+                        .onMalformedInput(CodingErrorAction.REPLACE)
+                        .onUnmappableCharacter(CodingErrorAction.REPLACE);
+        long encodedUpperBound =
+                (long) Math.ceil((double) encoder.maxBytesPerChar() * value.length());
+        int capacity = (int) Math.min(maxBytes, Math.min(Integer.MAX_VALUE, encodedUpperBound));
+        ByteBuffer output = ByteBuffer.allocate(capacity);
+        CoderResult result = encoder.encode(input, output, true);
+        if (result.isUnderflow()) {
+            // outputCharset() selects UTF-8 or a Windows native/default code-page charset. Those
+            // execution-path encoders are stateless, so they do not emit additional flush bytes.
```
oss-maintainer, line 513: The conclusion is right, but the reason given is narrower than the code path. `return value` is safe here because underflow means the whole input fit into `capacity`, and `capacity` is already clamped to `maxBytes` — that holds for any charset.

The statelessness claim does not: `encoder.flush()` can still append bytes for a stateful encoder, e.g. when `outputCharset()` falls back to `Charset.defaultCharset()` under `-Dfile.encoding=ISO-2022-JP` (or any other ISO-2022 family charset), where flush emits the escape back to ASCII state. Those bytes are outside the buffer, so the returned string can encode to slightly more than the cap. Not a correctness bug for the shell path today, but rewording the comment to the capacity invariant keeps the fast path justified without asserting a charset property that can be configured away.

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShellTest.java
```diff
@@ -46,6 +67,97 @@ void outputCharset_fallsBackToDefaultWhenWindowsNativeEncodingIsUnavailable() {
...
+                        new LocalFilesystemSpec()
+                                .project(tempDir)
+                                .maxOutputBytes(0)
+                                .toFilesystem(tempDir, null));
+
+        ExecuteResponse response = fs.execute(null, "echo output", 60);
+
+        assertEquals(0, response.exitCode());
+        assertEquals("<output capture disabled>", response.output());
+        assertFalse(response.truncated());
```
oss-maintainer, line 116: Good addition — this is the only test that pins the spec -> filesystem wiring of `maxOutputBytes`. One cheap hardening: mirror the constructor-level test's non-zero-exit case here (`execute(null, isWindows() ? "exit /b 7" : "exit 7", 60)` asserting `"<output capture disabled>\n\nExit code: 7"`). The exit-code trailer is appended after the disabled-capture branch, so a future refactor that moves or short-circuits that branch would still pass this test while dropping the diagnostic callers depend on.

oss-maintainer (commented):
### Summary

Delta re-review after `3f542ae7` → `843569cd` (`fix(harness): expose disabled output capture in spec`, +30/-6 across 5 files).

The commit moves the disabled-capture contract up to the public builder: `LocalFilesystemSpec.maxOutputBytes(int)` now accepts `0` (matching the `maxOutputBytes < 0` guard already in `LocalFilesystemWithShell`'s constructor), documents it in the javadoc, adds the first spec-level test for the wiring, and updates the EN and ZH option tables identically. The `truncateToByteLimit` change is comment-only, and CI is green on this head (`build ubuntu-latest` / `build windows-latest` / `validate` / `Check License` all `success`), so nothing here is a blocker for the behavior the PR is trying to land.

One item is worth a decision before merge, plus two optional nits — all three are inline:

- **[Warning]** `LocalFilesystemSpec.java:120` — the relaxation is a silent semantic change on a public setter: a `0` that today fails loudly (unset/typo'd `int` config) becomes "capture off" after this change. Either make the opt-in explicit, or record the change in the v2 migration notes.
- **[Info]** `LocalFilesystemWithShell.java:513` — the early `return value` is guaranteed by the `capacity <= maxBytes` clamp, not by encoder statelessness; the comment's premise does not survive every charset `outputCharset()` can return.
- **[Info]** `LocalFilesystemWithShellTest.java:116` — the new spec-level test only covers exit `0`; the exit-code trailer is appended after the disabled branch, so mirroring the `Exit code: 7` case would pin the diagnostics too.

Verified in this pass: the sentinel branch at `LocalFilesystemWithShell.java:452-455` still forces `truncated = false`, so `<output capture disabled>` is reported as complete output as intended; `maxOutputBytes` has no other production consumer in the repo (only the spec and the shell filesystem), so no additional call sites needed updating.

---

PR #2854, opened 2026-08-26
fix(sandbox): retain active fallback across concurrent calls

### LINKED ISSUE #2849: `SandboxConfigurationException: No active sandbox` with concurrent sessions on a shared agent — the per-call binding fix (#2490) is partial; the legacy `volatile sandbox` fallback still races
---                                                                                                                                                                                                                                                                         
  ### Title
  `SandboxConfigurationException: No active sandbox` with concurrent sessions on a shared agent — the per-call binding fix (#2490) is partial; the legacy `volatile sandbox` fallback still races                                                                             
                                                            
  ### Summary
  After the #2490 fix (sandbox bound per-call on `RuntimeContext`), two **different sessions** running **concurrently on the same `HarnessAgent`** still intermittently throw:

  io.agentscope.harness.agent.sandbox.SandboxException$SandboxConfigurationException:
  No active sandbox — sandbox filesystem used outside of a call context
      at io.agentscope.harness.agent.filesystem.sandbox.SandboxBackedFilesystem.requireSandbox(SandboxBackedFilesystem.java:245)

  Direct tool calls are isolated correctly (the per-call binding works). The failure comes from the **context-free internal readers** — `WorkspaceMessageBus`, `WorkspaceAsyncToolRegistry`, `WorkspaceTaskRepository`, and the baked `getWorkspaceManager()` — which still
  resolve the sandbox through the single shared `volatile Sandbox sandbox` field.

  ### Expected behavior
  Two distinct sessions on one agent must never observe "No active sandbox", including for internal filesystem readers that do not carry a per-call `RuntimeContext`.

  ### Actual behavior / race timeline
  Session A and B stream concurrently on the same agent:

  | step | event | shared field |
  |---|---|---|
  | 1 | A `acquireForCall` → `setSandbox(sbA)` | `sbA` |
  | 2 | B `acquireForCall` → `setSandbox(sbB)` | `sbB` (last-writer-wins overwrite) |
  | 3 | B `releaseForCall` → `clearSandboxIfCurrent(sbB)` — field `== sbB` → cleared | `null` |
  | 4 | A still streaming; next context-free read (bus / async registry / task repo / plan capture via `getWorkspaceManager()`) → `requireSandbox(emptyCtx)` → no per-call binding **and** field is `null` → throw at line 245 | `null` |

  `clearSandboxIfCurrent`'s compare-and-clear only guards the *reverse* race (a releasing call never nulls a field that already points at a sibling's sandbox). It cannot help when the releasing call **overwrote** the field via `setSandbox` (last-writer-wins): at release
   time the field points at *its own* sandbox, so it legitimately nulls it while the sibling session is still mid-call. The single slot has no memory that another call is in flight.

  ### Root cause
  The #2490 fix made only the tool-call path context-bound. The internal machinery is still per-agent singleton and passes a shared static empty context to the filesystem:

  - `WorkspaceMessageBus` — `private static final RuntimeContext RC = RuntimeContext.empty()`; every `fs.*(RC, ...)` call uses it
  - `WorkspaceAsyncToolRegistry` — same static empty RC
  - `WorkspaceTaskRepository` — `rc != null ? rc : RuntimeContext.empty()` fallbacks throughout (including the orphan sweeper)
  - `BakedContextFilesystem` / `getWorkspaceManager()` — bakes only `userId`+`sessionId`, **no** `SandboxAcquireResult`

  These callers fall through to the single `volatile Sandbox sandbox` field, which the javadoc itself describes as "last-writer-wins under concurrency and must not be relied on for isolation".

  ### Suggested fix
  Make sandbox resolution **session-scoped** instead of a single global slot:

  **Option 1 — session-keyed fallback map:** replace `volatile Sandbox sandbox` in `SandboxBackedFilesystem` with `ConcurrentHashMap<sessionKey, Sandbox>`; `acquireForCall` binds by the call's session, `releaseForCall` removes only its own key; context-free readers
  resolve by the session key they already know (bus/async/task records are keyed by `sessionId`).

  **Option 2 — finish #2490: thread the per-call `RuntimeContext` into the internal readers.** Add a `RuntimeContext` parameter to the message bus / async registry / task repository methods and pass the current call's context at the call sites, so every filesystem
  access resolves through the per-call binding and the shared field is no longer needed.

  ### App-level workaround (unblocks us)
  Cache the `HarnessAgent` per `(agentId, sessionId)` so each session owns a distinct `SandboxBackedFilesystem` instance. This removes cross-session contention but does not fix the underlying framework gap.

  ### Environment
  - agentscope-java `main` (includes #2490)
  - `IsolationScope.SESSION`, `SandboxExecutionGuard` configured
  - Two concurrent sessions sharing one `HarnessAgent`

  ---

Oxygen56:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #2849.

When concurrent calls shared one agent, the context-free filesystem fallback remembered only the
latest sandbox. If that call released while an earlier call was still active, it cleared the
fallback and caused the remaining call's internal filesystem work to fail.

This change retains every active fallback binding in acquisition order. Releasing a call removes
one matching binding and restores the most recently acquired binding that is still active. It also
removes non-current bindings so they cannot be restored after release, and preserves repeated
bindings to the same sandbox until every matching call has released. Per-call context binding
remains authoritative and unchanged. Startup cleanup also records whether the current call
actually registered a fallback, so a failed startup cannot remove an identical sandbox binding
that belongs to another active call.

This follow-up is intentionally limited to the reported empty-fallback window. Context-free
callers still use the most recently acquired active fallback and do not gain session routing;
per-call context remains the isolation mechanism.

The failed-start regression was run against both versions: it reaches the reported missing-sandbox
failure before this follow-up and passes afterward. The existing release-order regressions and
focused lifecycle checks also pass locally.

The null-binding regression confirms that a rejected null update leaves both the current and previous active bindings intact.

Local `clean verify` passes for the Harness module and its dependencies, including formatting and the full relevant test suites. GitHub CI runs the complete cross-platform build and remaining repository checks.

## Checklist

- [x]  Related documentation has been updated (not applicable; behavior is internal)

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
@@ -53,15 +53,18 @@ import org.slf4j.LoggerFactory;
  * {@code volatile sandbox} field is retained only as a best-effort fallback for context-free
  * internal callers that resolve the filesystem with a shared empty {@link RuntimeContext} (e.g.
  * {@link io.agentscope.harness.agent.bus.WorkspaceMessageBus}, which carries no per-call binding).
- * The middleware still maintains that field via {@link #setSandbox} on acquire and {@link
- * #clearSandboxIfCurrent} on release, so it remains last-writer-wins under concurrency and must not
- * be relied on for isolation.
+ * The middleware maintains that fallback via {@link #setSandbox} on acquire and {@link
+ * #clearSandboxIfCurrent} on release. Active bindings are retained in acquisition order so a
+ * release restores another still-active binding instead of transiently leaving context-free
+ * callers without a sandbox (issue #2849). The fallback still must not be relied on for session
+ * isolation; per-call bindings remain authoritative.
  */
 public class SandboxBackedFilesystem extends BaseSandboxFilesystem implements SandboxAware {
 
     private static final Logger log = LoggerFactory.getLogger(SandboxBackedFilesystem.class);
 
     private final String fsId;
+    private final List<Sandbox> fallbackBindings = new ArrayList<>();
     private volatile Sandbox sandbox;
 
     public SandboxBackedFilesystem() {
@@ -70,6 +73,12 @@ public class SandboxBackedFilesystem extends BaseSandboxFilesystem implements Sa
 
     @Override
     public synchronized void setSandbox(Sandbox sandbox) {
+        if (sandbox == null) {
+            fallbackBindings.clear();
+            this.sandbox = null;
+            return;
+        }
+        fallbackBindings.add(sandbox);
         this.sandbox = sandbox;
     }
 
@@ -79,16 +88,25 @@ public class SandboxBackedFilesystem extends BaseSandboxFilesystem implements Sa
     }
 
     /**
-     * Clears the fallback {@code sandbox} field only if it still points at {@code expected}. Used by
-     * {@link io.agentscope.harness.agent.middleware.SandboxLifecycleMiddleware} on release so a
-     * finishing call never nulls a concurrent sibling call's fallback binding (issue #2490).
+     * Removes one fallback binding for {@code expected} and restores the most recently acquired
+     * binding that remains active. Used by {@link
+     * io.agentscope.harness.agent.middleware.SandboxLifecycleMiddleware} on release so a finishing
+     * call neither clears nor later restores a concurrent sibling call's binding (issues #2490 and
+     * #2849).
      *
      * @param expected the sandbox this call bound at acquire time
      */
     public synchronized void clearSandboxIfCurrent(Sandbox expected) {
-        if (this.sandbox == expected) {
-            this.sandbox = null;
+        for (int i = fallbackBindings.size() - 1; i >= 0; i--) {
+            if (fallbackBindings.get(i) == expected) {
+                fallbackBindings.remove(i);
+                break;
+            }
         }
+        this.sandbox =
+                fallbackBindings.isEmpty()
+                        ? null
+                        : fallbackBindings.get(fallbackBindings.size() - 1);
     }
 
     @Override
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
@@ -36,6 +36,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
 import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
@@ -71,13 +72,15 @@ public class SandboxBackedFilesystem extends BaseSandboxFilesystem implements Sa
         this.fsId = "sandbox-" + UUID.randomUUID().toString().substring(0, 8);
     }
 
+    /**
+     * Registers an active fallback binding.
+     *
+     * @param sandbox the active sandbox; must not be {@code null}
+     * @throws NullPointerException if {@code sandbox} is {@code null}
+     */
     @Override
     public synchronized void setSandbox(Sandbox sandbox) {
-        if (sandbox == null) {
-            fallbackBindings.clear();
-            this.sandbox = null;
-            return;
-        }
+        Objects.requireNonNull(sandbox, "sandbox");
         fallbackBindings.add(sandbox);
         this.sandbox = sandbox;
     }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
@@ -159,8 +159,8 @@ public class SandboxLifecycleMiddleware implements HarnessRuntimeMiddleware {
             return;
         }
         ctx.put(SandboxAcquireResult.class, null);
-        // Compare-and-clear the fallback field so a releasing call never nulls a concurrent
-        // sibling's binding (issue #2490); it only clears the field when it still points here.
+        // Remove this call's fallback binding. The proxy restores the most recently acquired
+        // binding that is still active, so a sibling call is neither cleared nor resurrected.
         filesystemProxy.clearSandboxIfCurrent(result.getSandbox());
         SandboxContext sandboxContext = ctx.get(SandboxContext.class);
         try {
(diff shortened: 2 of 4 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
```diff
@@ -70,6 +73,12 @@ public SandboxBackedFilesystem() {
 
     @Override
     public synchronized void setSandbox(Sandbox sandbox) {
+        if (sandbox == null) {
```
xiaoyuuuuuupeng: Suggestion: setSandbox(null) clears the whole fallbackBindings list, not just one entry — so any other active call loses its fallback too. Nothing in prod calls it today; I'd just drop the null branch or requireNonNull so nobody uses it by accident.

xiaoyuuuuuupeng (approved):

xiaoyuuuuuupeng:
LGTM Fixes the #2849 empty-fallback window: fallbackBindings stack + fallbackBound look correct. Concurrent lifecycle tests pass; per-call binding unchanged. One non-blocking nit on setSandbox(null) clearing the whole stack — doesn't block merge.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
Oxygen56: Good catch. I now reject null fallback bindings before any state mutation and added a regression test that preserves both the current and previous active bindings.

xiaoyuuuuuupeng (approved):

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
```diff
@@ -53,23 +54,34 @@
...
+ * release restores another still-active binding instead of transiently leaving context-free
+ * callers without a sandbox (issue #2849). The fallback still must not be relied on for session
+ * isolation; per-call bindings remain authoritative.
  */
 public class SandboxBackedFilesystem extends BaseSandboxFilesystem implements SandboxAware {
 
     private static final Logger log = LoggerFactory.getLogger(SandboxBackedFilesystem.class);
 
     private final String fsId;
+    private final List<Sandbox> fallbackBindings = new ArrayList<>();
```
oss-maintainer, line 68: **Binding lifetime is only shortened by `clearSandboxIfCurrent`, so any acquire that never reaches `releaseForCall` leaks a permanent entry.** `releaseForCall` returns early when `ctx.get(SandboxAcquireResult.class)` is `null`, and the fallback field is written nowhere else — so for such a call the list keeps growing and, worse, that stale (possibly stopped) `Sandbox` can later be restored as the fallback for context-free callers, which would run their filesystem work against a released sandbox. Previously the failure mode was bounded: a stale value was simply overwritten by the next `setSandbox`.

Could you either bound the list (cap + warn log) or clear all bindings on agent shutdown / sandbox-manager close, so a leaked acquire cannot be resurrected? Happy to see a test for the "acquire without release" path too.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
```diff
@@ -53,23 +54,34 @@
...
 
+    /**
+     * Registers an active fallback binding.
+     *
+     * @param sandbox the active sandbox; must not be {@code null}
+     * @throws NullPointerException if {@code sandbox} is {@code null}
+     */
     @Override
     public synchronized void setSandbox(Sandbox sandbox) {
+        Objects.requireNonNull(sandbox, "sandbox");
```
oss-maintainer, line 83: `SandboxAware#setSandbox` is a public interface method of `agentscope-harness`, and `null` used to be an accepted way to clear the fallback. Rejecting it is fine, but the contract now differs between the interface (no `@throws`, no `@NonNull`) and this implementation. Worth documenting the non-null requirement on `SandboxAware.setSandbox` itself so external implementors do not get an NPE from a call site that used to work.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
```diff
@@ -79,16 +91,25 @@ public Sandbox getSandbox() {
...
+        for (int i = fallbackBindings.size() - 1; i >= 0; i--) {
+            if (fallbackBindings.get(i) == expected) {
+                fallbackBindings.remove(i);
+                break;
+            }
         }
+        this.sandbox =
+                fallbackBindings.isEmpty()
+                        ? null
+                        : fallbackBindings.get(fallbackBindings.size() - 1);
```
oss-maintainer, line 112: Here the field is reassigned even when the loop removed nothing (an unrelated or duplicated `expected`). Today that is still a no-op because `sandbox == fallbackBindings.tail` holds after every mutation, but the invariant is implicit. Guarding the restore with the `removed` flag from the loop above would make it explicit and keep the old "only touch the field when this call owns it" guarantee from #2490 visible at the call site.

oss-maintainer (commented):
### Summary

Turns the context-free sandbox fallback in `SandboxBackedFilesystem` from a single last-writer-wins `volatile` field into an ordered binding stack, so that when one of several concurrent calls sharing an agent releases, the fallback restores a still-active sibling instead of transiently dropping to `null` (issue #2849). Also stops a failed `sandbox.start()` from clearing a binding it never registered, via the `fallbackBound` flag in `SandboxLifecycleMiddleware`. The stack model is a genuine improvement over compare-and-clear for the reported failure, the identity-based removal keeps #2490's per-call guarantee, and the five new tests (remaining-sandbox restore, non-current release not resurrected, shared-sandbox refcounting, null rejection, failed-start) cover exactly the interesting orderings.

### Findings
- **[Warning]** `SandboxBackedFilesystem.java:68` — the stack only shrinks through `clearSandboxIfCurrent`, so an acquire whose release never runs leaves a permanent entry that can later be restored as the active fallback (a released sandbox handed to context-free callers). Previously that stale value was simply overwritten by the next acquire.
- **[Info]** `SandboxBackedFilesystem.java:83` — `Objects.requireNonNull` tightens `SandboxAware#setSandbox`, a public interface method that accepted `null` before; the contract is documented on the implementation only.
- **[Info]** `SandboxBackedFilesystem.java:112` — the field is reassigned even when nothing was removed, which is safe only while the implicit `sandbox == fallbackBindings.tail` invariant holds; an early return when no match was found would keep #2490's "only touch it if this call owns it" visible.

### Suggestions
For the first point, something as simple as a size guard plus a `WARN` makes a leak diagnosable instead of silent:

```java
if (fallbackBindings.size() > MAX_TRACKED_FALLBACK_BINDINGS) {
    log.warn("[sandbox-fs] {} fallback bindings still registered for {}, dropping oldest",
            fallbackBindings.size(), fsId);
    fallbackBindings.remove(0);
}
```

…or pruning on `stop`/agent close, if a bounded stack is not acceptable for long-lived mirrors. None of these look blocking to me — flagging so the resurrection path gets an explicit decision.

Verified: CLA signed (`license/cla` success), `build (ubuntu-latest)` / `build (windows-latest)` / `Check License` / `Check Module Sync` / `codecov/patch` all passing at head `a97c7d13`, `mergeable=true`, `mergeable_state=blocked` (branch protection, not a conflict).

---

PR #2856, opened 2026-08-26
fix(agui): isolate HITL sessions by user

### LINKED ISSUE #2855: [Bug]: AG-UI HITL sessions mix tenants by threadId
**Describe the bug**

AG-UI HITL with server-side memory is not tenant-safe and can fail to resume interrupts.

1. `ThreadSessionManager` / `AgentResolver` key sessions only by `threadId`. Different users that share a thread id reuse the same agent and `hasMemory` result, so conversation state leaks across tenants.
2. Inspecting session memory (and related unwrap of a live `ReActAgent` / harness delegate) can close the in-session agent. The next turn then loses the live HITL agent instead of reusing it.
3. When server-side memory is enabled, `extractLatestUserMessage` keeps only the last user message. Trailing HITL tool results after the last assistant turn are dropped, so interrupt resume cannot continue.
4. Frontend tools should not emit AG-UI interrupts; backend HITL tools still should.

**To Reproduce**

1. Enable AG-UI server-side memory (`ThreadSessionManager`).
2. Two users start runs with the same `threadId` but different `userId` — memory and agent reuse mix.
3. Trigger a backend HITL tool suspend (`TOOL_SUSPENDED`) — interrupt should emit. A frontend tool suspend should not.
4. Resume a HITL interrupt while server-side memory is on — tool-result messages after the last assistant turn are discarded.

**Expected behavior**

- Sessions are keyed by `(userId, threadId)` so tenants never share an agent slot.
- Reading `hasMemory` / unwrapping a harness `ReActAgent` does not close the live session agent.
- Follow-up input after the last assistant turn (including HITL tool results) is kept so interrupt resume can continue.
- Do not emit interrupts for frontend tools.

**Environment**

- AgentScope-Java Version: main
- Java Version: 17
- OS: Windows

**Additional context**

- PR: https://github.com/agentscope-ai/agentscope-java/pull/2856

dragondyt:
## Related issues
- Fixes #2855
- 
- ## Summary
- Key `ThreadSessionManager` / `AgentResolver` by `(userId, threadId)` so `hasMemory` and agent reuse no longer mix tenants.
- Inspect session memory without closing the live `ReActAgent`.
- Keep messages after the last assistant turn, including tool results, so HITL can resume.
- Do not emit interrupts for frontend tools.
- Unwrap harness/stop interrupts via `AguiUtil.asReActAgent` so demo `stopThread` and processor interrupt target the live session.
- Add unit coverage for user-scoped sessions, default `AgentResolver`, lifecycle interrupt filtering, and `ProcessResult.interrupt()`.

## Breaking Changes

This PR is source- and behavior-incompatible. Upgrading requires recompilation + minor migration.

### API (source-incompatible)
- `AgentResolver#hasMemory(String threadId)` -> `hasMemory(RuntimeContext)`. Custom `AgentResolver`
  implementations must update the signature; the single-arg form is removed.
- `ThreadSessionManager#hasMemory(String threadId)` -> `hasMemory(RuntimeContext)`. Callers must pass
  a `RuntimeContext` carrying `userId` + `sessionId`.

### Behavior (runtime-incompatible)
- Session keying changed `threadId` -> `(userId, threadId)`. Different users on the same thread id no
  longer share an agent instance. Anonymous callers (`userId == null`) stay on the shared `__anon__`
  slot, so single-tenant behavior is unchanged.
- `ThreadSessionManager#getSessions()` now returns a threadId-keyed snapshot that is lossy when
  multiple users share a thread id; use `getSession(userId, threadId)` for tenant-safe lookup.
- AG-UI event stream: frontend/external tools (`SchemaOnlyTool` / `@Tool(externalTool=true)`) no longer
  emit a `TOOL_SUSPENDED` interrupt. Results are delivered via message-based resume (a trailing
  `role:"tool"` message in the next run's transcript) instead of `RunAgentInput.resume`.

### DIFF
--- a/agentscope-examples/agentscope-copilotkit/src/main/java/io/agentscope/examples/copilotkit/config/CopilotKitRouteConfiguration.java
+++ b/agentscope-examples/agentscope-copilotkit/src/main/java/io/agentscope/examples/copilotkit/config/CopilotKitRouteConfiguration.java
@@ -114,20 +114,25 @@ public class CopilotKitRouteConfiguration {
 
     private Mono<ServerResponse> stopThread(
             ServerRequest request, ThreadSessionManager threadSessionManager) {
-        String agentId = request.pathVariable("agentId");
         String threadId = request.pathVariable("threadId");
+        String userId = resolveDemoUserId(request);
         threadSessionManager
-                .getSession(threadId)
+                .getSession(userId, threadId)
                 .ifPresent(
                         threadSession -> {
                             if (threadSession.getAgent() instanceof ReActAgent reActAgent) {
-                                reActAgent.interrupt(null, threadId);
+                                reActAgent.interrupt(userId, threadId);
                             } else {
                                 // todo Harness?
                                 threadSession.getAgent().interrupt();
                             }
                         });
         return Mono.empty();
+    }
+
+    private static String resolveDemoUserId(ServerRequest request) {
+        String token = request.headers().firstHeader("X-Token");
+        return token == null || token.isBlank() ? DemoThreadStore.DEMO_USER_ID : token;
     }
 
     private static Mono<ServerResponse> listThreads(
--- a/agentscope-examples/agentscope-copilotkit/src/main/java/io/agentscope/examples/copilotkit/config/CopilotKitRouteConfiguration.java
+++ b/agentscope-examples/agentscope-copilotkit/src/main/java/io/agentscope/examples/copilotkit/config/CopilotKitRouteConfiguration.java
@@ -16,6 +16,7 @@
 package io.agentscope.examples.copilotkit.config;
 
 import io.agentscope.core.ReActAgent;
+import io.agentscope.core.agui.AguiUtil;
 import io.agentscope.core.agui.model.RunAgentInput;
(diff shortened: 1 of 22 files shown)


agentscope-examples/agentscope-copilotkit/src/main/java/io/agentscope/examples/copilotkit/config/CopilotKitRouteConfiguration.java
```diff
@@ -114,14 +114,14 @@ public RouterFunction<ServerResponse> copilotKitRoutes(
...
-                .getSession(threadId)
+                .getSession(userId, threadId)
                 .ifPresent(
                         threadSession -> {
                             if (threadSession.getAgent() instanceof ReActAgent reActAgent) {
-                                reActAgent.interrupt(null, threadId);
+                                reActAgent.interrupt(userId, threadId);
                             } else {
                                 // todo Harness?
                                 threadSession.getAgent().interrupt();
```
dragondyt, line 129: 需要处理else分支

dragondyt (commented):
仅在demo的stopThread有遗漏，对核心无影响

jujn (approved):

PR #2862, opened 2026-08-27
fix(core): default retryOn to RETRYABLE_ERRORS when unset in ModelUtils

### LINKED ISSUE #2861: [Bug]: ModelUtils retries auth/request-side errors when retryOn is unset, contradicting RETRYABLE_ERRORS docs
## Summary

`ModelUtils.applyTimeoutAndRetry(...)` applies `error -> true` when `ExecutionConfig.getRetryOn()` is `null`:

```java
if (retryOn == null) {
    retryOn = error -> true; // retry all errors by default
}
```

(`agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java`, `applyTimeoutAndRetry`)

This contradicts the documented default of `ExecutionConfig.RETRYABLE_ERRORS` (which backs `ExecutionConfig.MODEL_DEFAULTS`): its javadoc explicitly lists HTTP 401/403 (authentication) and other 4xx client errors as **non-retryable**, while 429/5xx/timeout/network errors are retryable.

Because the field is nullable and `ExecutionConfig.mergeConfigs` keeps `null` when neither side supplies a predicate, any caller that builds an `ExecutionConfig` with only `maxAttempts` / backoff (or relies on the agent-level `maxRetries` budget) silently gets retry-all semantics: an invalid API key (401/403) or a permanently bad request (400/422) is retried `maxAttempts - 1` extra times, each guaranteed to fail again — burning latency, quota, and log noise.

## Impact

- Every official model extension (OpenAI, Anthropic, DashScope, Gemini, Ollama) routes through `ModelUtils.applyTimeoutAndRetry`, so the bug affects all providers when `retryOn` is not explicitly configured.
- Auth and request-side failures are retried pointlessly; the final error surfaces only after the full retry budget is exhausted.

## Suggested Fix

When `retryOn` is `null`, default to `ExecutionConfig.RETRYABLE_ERRORS` (429/5xx/timeout/network) so runtime behavior matches the documented default. This is a behavior change for callers that relied on retry-all — it restores the semantics the docs already promise.

## Related

- #2829 observes the tool-execution path also has a retry-default problem (`ToolExecutor.applyRetry` shares the same `error -> true` fallback); note that tool retry is currently dead code there because `execute()` converts failures into normal `ToolResultBlock.error` signals before `retryWhen` can see them. This issue fixes only the live model path; the tool path deserves separate tracking.

Letter2025:
## Summary

`ModelUtils.applyTimeoutAndRetry(...)` falls back to `error -> true` (retry **all** errors) when `ExecutionConfig.getRetryOn()` is `null`, contradicting the documented `RETRYABLE_ERRORS` default and pointlessly retrying auth (401/403) and request-side (400/422) errors — each attempt is guaranteed to fail again, burning latency, quota, and log noise.

## Change

```diff
if (retryOn == null) {
-   retryOn = error -> true; // retry all errors by default
+   // Match the documented default: only retry transient/retryable errors
+   // (429, 5xx, timeouts, network errors). Retrying auth (401/403) or
+   // request-side (400/422) errors is guaranteed to fail again.
+   retryOn = ExecutionConfig.RETRYABLE_ERRORS;
}
```

Runtime behavior now matches the javadoc of `ExecutionConfig.RETRYABLE_ERRORS` (which also backs `MODEL_DEFAULTS`), affecting every official model extension (OpenAI, Anthropic, DashScope, Gemini, Ollama) that routes through `applyTimeoutAndRetry`.

## Behavior change notice (intentional)

This is a deliberate bug fix that changes default behavior for callers relying on retry-all semantics: 401/403/400/422 will no longer be retried unless an explicit `retryOn` predicate is configured. Explicitly configured predicates are untouched.

## Tests

New `ModelUtilsRetryDefaultTest` (5 cases):
- 401 is **not** retried by default
- 429 **is** retried up to maxAttempts
- 400 is **not** retried by default
- timeout **is** retried up to maxAttempts
- an explicit `retryOn` predicate is still respected

All pass locally (`mvn test -pl agentscope-core -Dtest=ModelUtilsRetryDefaultTest`), plus related regression suites (ExecutionConfigTest, GenerateOptionsTest, ModelConfigTest, ModelTimeoutRetryTest) remain green.

## Related

Closes #2861
- #2829 tracks the companion issue on the tool-execution path (`ToolExecutor.applyRetry` shares the same `error -> true` fallback, but tool retry is currently dead code because `execute()` converts failures into normal `ToolResultBlock.error` signals before `retryWhen` can observe them). This PR intentionally fixes only the live model path; the tool path deserves separate tracking and a follow-up.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
@@ -16,6 +16,7 @@
 package io.agentscope.core.model;
 
 import java.time.Duration;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Predicate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -82,12 +83,16 @@ public final class ModelUtils {
             // Apply timeout if configured
             Duration timeout = execConfig.getTimeout();
             if (timeout != null) {
+                // Keep the TimeoutException type on the cause chain so the default retry
+                // predicate (RETRYABLE_ERRORS) classifies this produced timeout as retryable.
                 responseFlux =
                         responseFlux.timeout(
                                 timeout,
                                 Flux.error(
                                         new ModelException(
                                                 "Model request timeout after " + timeout,
+                                                new TimeoutException(
+                                                        "Model request timeout after " + timeout),
                                                 modelName,
                                                 provider)));
                 LOG.debug("Applied timeout: {} for model: {}", timeout, modelName);
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
@@ -108,7 +108,10 @@ public final class ModelUtils {
                     maxBackoff = Duration.ofSeconds(10);
                 }
                 if (retryOn == null) {
-                    retryOn = error -> true; // retry all errors by default
+                    // Match the documented default: only retry transient/retryable errors
+                    // (429, 5xx, timeouts, network errors). Retrying auth (401/403) or
+                    // request-side (400/422) errors is guaranteed to fail again.
+                    retryOn = ExecutionConfig.RETRYABLE_ERRORS;
                 }
 
                 Retry retrySpec =
--- a/agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsRetryDefaultTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsRetryDefaultTest.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.model;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import io.agentscope.core.model.transport.HttpTransportException;
+import java.time.Duration;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Flux;
+import reactor.test.StepVerifier;
+
+/**
+ * Verifies the default {@code retryOn} semantics of
+ * {@link ModelUtils#applyTimeoutAndRetry} when no explicit predicate is configured.
+ *
+ * <p>Regression guard: when {@link ExecutionConfig#getRetryOn()} is {@code null} the retry
+ * predicate must default to {@link ExecutionConfig#RETRYABLE_ERRORS} (retry 429/5xx/timeouts/
+ * network errors only) instead of retrying every error — retrying auth (401/403) or request-side
+ * (400/422) failures is guaranteed to fail again.
+ */
+@Tag("unit")
+@DisplayName("ModelUtils Default RetryOn Semantics Tests")
+class ModelUtilsRetryDefaultTest {
+
+    private static final String MODEL = "test-model";
+    private static final String PROVIDER = "test-provider";
+
+    private GenerateOptions optionsWithMaxAttemptsButNoRetryOn() {
+        return GenerateOptions.builder()
+                .executionConfig(
+                        ExecutionConfig.builder()
+                                .maxAttempts(3)
+                                .initialBackoff(Duration.ofMillis(5))
+                                .maxBackoff(Duration.ofMillis(20))
+                                .build())
+                .build();
+    }
+
+    @Test
+    @DisplayName("Default retryOn must not retry auth errors (401)")
+    void defaultRetryOnDoesNotRetryAuthErrors() {
+        AtomicInteger attempts = new AtomicInteger();
+
+        Flux<ChatResponse> source =
+                Flux.defer(
+                        () -> {
+                            attempts.incrementAndGet();
+                            return Flux.error(
+                                    new HttpTransportException(
+                                            "unauthorized",
+                                            401,
+                                            "{\"error\":\"invalid_api_key\"}"));
+                        });
+
+        StepVerifier.create(
+                        ModelUtils.applyTimeoutAndRetry(
+                                source,
+                                optionsWithMaxAttemptsButNoRetryOn(),
+                                null,
+                                MODEL,
+                                PROVIDER))
+                .expectError()
+                .verify(Duration.ofSeconds(5));
+
+        assertEquals(1, attempts.get(), "401 must not be retried by the default predicate");
+    }
+
+    @Test
+    @DisplayName("Default retryOn must retry rate-limit errors (429)")
+    void defaultRetryOnRetriesRateLimitErrors() {
+        AtomicInteger attempts = new AtomicInteger();
+
+        Flux<ChatResponse> source =
+                Flux.defer(
+                        () -> {
+                            attempts.incrementAndGet();
+                            return Flux.error(
+                                    new HttpTransportException(
+                                            "rate limited",
+                                            429,
+                                            "{\"error\":\"rate_limit_exceeded\"}"));
+                        });
+
+        StepVerifier.create(
+                        ModelUtils.applyTimeoutAndRetry(
+                                source,
+                                optionsWithMaxAttemptsButNoRetryOn(),
+                                null,
+                                MODEL,
+                                PROVIDER))
+                .expectError()
+                .verify(Duration.ofSeconds(10));
+
+        assertEquals(3, attempts.get(), "429 must be retried up to maxAttempts");
+    }
+
+    @Test
+    @DisplayName("Default retryOn must not retry request-side errors (400)")
+    void defaultRetryOnDoesNotRetryBadRequestErrors() {
+        AtomicInteger attempts = new AtomicInteger();
+
+        Flux<ChatResponse> source =
+                Flux.defer(
+                        () -> {
+                            attempts.incrementAndGet();
+                            return Flux.error(
+                                    new HttpTransportException(
+                                            "bad request", 400, "{\"error\":\"invalid_request\"}"));
+                        });
+
+        StepVerifier.create(
+                        ModelUtils.applyTimeoutAndRetry(
+                                source,
+                                optionsWithMaxAttemptsButNoRetryOn(),
+                                null,
+                                MODEL,
+                                PROVIDER))
+                .expectError()
+                .verify(Duration.ofSeconds(5));
+
+        assertEquals(1, attempts.get(), "400 must not be retried by the default predicate");
+    }
+
+    @Test
+    @DisplayName("Default retryOn must retry timeout errors")
+    void defaultRetryOnRetriesTimeouts() {
+        AtomicInteger attempts = new AtomicInteger();
+
+        Flux<ChatResponse> source =
+                Flux.defer(
+                        () -> {
+                            attempts.incrementAndGet();
+                            return Flux.error(new TimeoutException("read timed out"));
+                        });
+
+        StepVerifier.create(
+                        ModelUtils.applyTimeoutAndRetry(
+                                source,
+                                optionsWithMaxAttemptsButNoRetryOn(),
+                                null,
+                                MODEL,
+                                PROVIDER))
+                .expectError()
+                .verify(Duration.ofSeconds(10));
+
+        assertEquals(3, attempts.get(), "timeouts must be retried up to maxAttempts");
+    }
+
+    @Test
+    @DisplayName("Explicit retryOn must still be respected when configured")
+    void explicitRetryOnIsRespected() {
+        AtomicInteger attempts = new AtomicInteger();
+
+        Flux<ChatResponse> source =
+                Flux.defer(
+                        () -> {
+                            attempts.incrementAndGet();
+                            return Flux.error(
+                                    new HttpTransportException(
+                                            "unauthorized", 401, "{\"error\":\"invalid_key\"}"));
+                        });
+
+        GenerateOptions options =
+                GenerateOptions.builder()
+                        .executionConfig(
+                                ExecutionConfig.builder()
+                                        .maxAttempts(3)
+                                        .initialBackoff(Duration.ofMillis(5))
+                                        .maxBackoff(Duration.ofMillis(20))
+                                        .retryOn(error -> true) // explicit: retry everything
+                                        .build())
+                        .build();
+
+        StepVerifier.create(ModelUtils.applyTimeoutAndRetry(source, options, null, MODEL, PROVIDER))
+                .expectError()
+                .verify(Duration.ofSeconds(10));
+
+        assertEquals(3, attempts.get(), "an explicit retryOn=true must retry 401");
+    }
+}
--- a/agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsRetryDefaultTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsRetryDefaultTest.java
@@ -140,8 +140,41 @@ class ModelUtilsRetryDefaultTest {
     }
 
     @Test
-    @DisplayName("Default retryOn must retry timeout errors")
-    void defaultRetryOnRetriesTimeouts() {
+    @DisplayName("Default retryOn must retry an actually timing-out source")
+    void defaultRetryOnRetriesTimingOutSource() {
+        AtomicInteger attempts = new AtomicInteger();
+
+        // A source that never emits: the timeout operator inside applyTimeoutAndRetry must
+        // fire, and the produced error (a ModelException carrying a TimeoutException cause)
+        // must be classified as retryable by the default predicate.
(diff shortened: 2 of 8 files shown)


agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsRetryDefaultTest.java
```diff
@@ -0,0 +1,198 @@
...
+                                optionsWithMaxAttemptsButNoRetryOn(),
+                                null,
+                                MODEL,
+                                PROVIDER))
+                .expectError()
+                .verify(Duration.ofSeconds(10));
+
+        assertEquals(3, attempts.get(), "429 must be retried up to maxAttempts");
+    }
+
```
guslegend0510, line 114: [P1] Preserve timeout classification

The timeout operator above emits a cause-less ModelException, so RETRYABLE_ERRORS rejects the timeout produced by this method itself. With maxAttempts=3 the source is subscribed only once. Preserve a TimeoutException cause/type and test an actually timing-out source.

[P1] Real Anthropic 429 is not recognized

Anthropic RateLimitException reports status 429 but does not implement ModelHttpException, so the new default predicate returns false. ReActAgent's default maxAttempts-only configuration therefore stops after one 429. Adapt provider exceptions to the provider-neutral HTTP contract and add a real SDK-exception test.

guslegend0510 (changes requested):

agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsRetryDefaultTest.java
Letter2025: ﻿Thanks for the careful review — both issues are real, and both are fixed in `5f973767`.

**P1 (timeout classification)** — Confirmed. The `timeout(...)` operator produced a cause-less `ModelException`, so `RETRYABLE_ERRORS` rejected the timeout generated by this method itself (`maxAttempts` ended up subscribing once). Fixed in `ModelUtils.applyTimeoutAndRetry`: the produced error now carries a `TimeoutException` cause, so classification matches the documented default. `ModelUtilsRetryDefaultTest` now uses an **actually timing-out source** (`Flux.never()` + short timeout) and asserts 3 subscriptions at `maxAttempts=3`; a second test asserts the `TimeoutException` is preserved on the cause chain.

**P1 (real Anthropic 429)** — Confirmed and fixed provider-side. Anthropic's `RateLimitException` carries status 429 but does not implement `ModelHttpException`, so the new default predicate returned false and a genuine 429 stopped after one attempt. `AnthropicChatModel` now runs `adaptSdkException(...)` over the whole response flux (covers synchronous call, non-streaming `CompletableFuture`, and stream parsing): any `AnthropicServiceException` found on the cause chain is remapped to `HttpTransportException(status, body, cause)` before retry/fallback classification sees it. Tests build a **real SDK `RateLimitException`** (429) plus a wrapped-through-cause-chain case, and assert `RETRYABLE_ERRORS` classifies the adapted error as retryable.

While auditing I found the same gap in `DashScopeHttpException` and `OllamaHttpException` (they already carried `getStatusCode()` but did not implement `ModelHttpException`); both now implement the contract so 429/5xx classify consistently, with contract tests on each client test class.

Local verification before pushing: core `ModelUtilsRetryDefaultTest` 7/7, `AnthropicChatModelTest` 29/29, `DashScopeHttpClientTest` 62/62, `OllamaHttpClientTest` 14/14, all green; spotless applied.

agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
```diff
@@ -108,7 +113,10 @@ public static Flux<ChatResponse> applyTimeoutAndRetry(
                     maxBackoff = Duration.ofSeconds(10);
                 }
                 if (retryOn == null) {
-                    retryOn = error -> true; // retry all errors by default
+                    // Match the documented default: only retry transient/retryable errors
+                    // (429, 5xx, timeouts, network errors). Retrying auth (401/403) or
+                    // request-side (400/422) errors is guaranteed to fail again.
+                    retryOn = ExecutionConfig.RETRYABLE_ERRORS;
```
oss-maintainer, line 119: Aligning the unset default with the documented `RETRYABLE_ERRORS` is right, and the commit chain shows the hard part was done properly — the timeout signal now carries a real `TimeoutException` cause precisely so it still classifies as retryable, which is easy to miss.

The part that looks unfinished is the other side of the flip. `OpenAIException` already implements the contract, and this PR adapts Anthropic (`adaptSdkException`) plus DashScope and Ollama (their `*HttpException` now implements `ModelHttpException`) — but **Gemini goes through the same helper and is not adapted**:

```java
// GeminiChatModel.doStream0, line 318-323
} catch (Exception e) {
    return Flux.error(new ModelException("Gemini API call failed: " + e.getMessage(), e));
}
```

`com.google.genai.errors.ApiException` (and its `ClientException` / `ServerException` subclasses) extends `RuntimeException` and exposes the status only through `code()`, so `isRetryableError` matches no branch, walks a cause chain that is also `RuntimeException`-shaped, and returns `false`. Under the previous `error -> true` those 429/5xx responses were retried; after this change a Gemini caller with a hand-built `ExecutionConfig` (no explicit `retryOn`) silently loses retry. Worth deciding explicitly: either add the equivalent adapter for Gemini — the `RETRYABLE_ERRORS` javadoc already claims timeouts and 5xx are retried, so today that guarantee is provider-dependent — or note the gap in the PR description so it lands as a follow-up rather than riding in with a "fix". The `agentscope-extensions-rag-simple` embedding clients also call `applyTimeoutAndRetry` and are worth a look in the same pass.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
```diff
@@ -256,7 +258,49 @@ protected Flux<ChatResponse> doStream(
 
         // Apply timeout and retry if configured
         return ModelUtils.applyTimeoutAndRetry(
-                responseFlux, options, defaultOptions, modelName, "anthropic");
+                // Adapt Anthropic SDK HTTP exceptions (RateLimitException and friends carry a
+                // status code but do not implement the provider-neutral ModelHttpException
+                // contract) to HttpTransportException so the default RETRYABLE_ERRORS predicate
+                // classifies real 429/5xx as retryable. Applies uniformly to errors raised by the
+                // synchronous call, the non-streaming CompletableFuture, and stream parsing.
+                responseFlux.onErrorResume(e -> Flux.error(adaptSdkException(e))),
```
oss-maintainer, line 266: `onErrorResume` rewrites the error for **every** subscriber of the stream, not only the retry operator, so the exception type that leaves `doStream` changes for all Anthropic users regardless of whether retry is configured: a 401/400 that previously surfaced as `ModelException` (with `modelName`/`provider` populated) now surfaces as `HttpTransportException`, and `Retry.backoff(...)` re-throws the last failure once attempts are exhausted. That is a wider contract change than the title implies, and `ModelException` carries the model/provider fields that callers and the fallback chain use for diagnostics.

Two ways to keep the classification without changing what callers see:

1. Adapt only for the decision — build the predicate here and pass it through `GenerateOptions`/`ExecutionConfig.retryOn` instead of rewriting the stream:

```java
Predicate<Throwable> retryOn = e ->
        ExecutionConfig.RETRYABLE_ERRORS.test(e) || isRetryableSdkStatus(e);
```

2. Or keep the rewrite but re-wrap the non-retryable cases so the public type is stable, i.e. only convert when `ase.statusCode() == 429 || ase.statusCode() >= 500` and return `error` unchanged otherwise. That also stops auth/400 errors from changing type on their way out.

Option 2 is the smaller diff and keeps the cause-chain walk in `adaptSdkException` useful; it just narrows the substitution to the errors the adapter exists to classify.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
```diff
@@ -256,7 +258,49 @@ protected Flux<ChatResponse> doStream(
...
+    static Throwable adaptSdkException(Throwable error) {
+        Throwable current = error;
+        while (current != null) {
+            if (current instanceof AnthropicServiceException ase) {
+                return new HttpTransportException(
+                        "Anthropic API request failed with status "
+                                + ase.statusCode()
+                                + (error.getMessage() != null ? ": " + error.getMessage() : ""),
+                        ase.statusCode(),
+                        ase.body() != null ? ase.body().toString() : null,
```
oss-maintainer: `ase.body().toString()` goes into `responseBody`, and `HttpTransportException.getMessage()` appends it (truncated to 1000 chars) — which `ModelUtils` then logs on every retry attempt via `signal.failure().getMessage()`. Provider error bodies can echo fragments of the request payload, so for a framework whose agents routinely put user content into prompts it is worth confirming that this log path is acceptable, or leaving the body out and keeping only the status code. `HttpTransportException.toString()` is more explicit than `getMessage()` here if the intent is operator-facing output only.

Otherwise this reads well: the `while (current != null)` cause-chain walk handles the `ModelException` wrapper from the synchronous `catch`, `error` is preserved as the cause so the original stack survives, and `ModelUtilsRetryDefaultTest` + the three adapter tests pin down exactly the cases that matter (401 not retried, 429 retried, 400 not retried, explicit `retryOn` still respected).

oss-maintainer (commented):
﻿### Summary

Fixes a real defaulting bug: when `ExecutionConfig.retryOn` was unset, `ModelUtils` used `error -> true`, so a 401 or a 400 was retried with backoff even though it can never succeed — while `ExecutionConfig.MODEL_DEFAULTS` and the docs both describe `RETRYABLE_ERRORS` as the default. Making the unset case match the documented default is the right direction, the timeout path is handled thoughtfully (a real `TimeoutException` cause so produced timeouts stay retryable, rather than letting the stricter predicate silently swallow them), and the provider-side contract work (`DashScopeHttpException` / `OllamaHttpException` implementing `ModelHttpException`, the Anthropic SDK adapter) is the correct shape for it. `ModelUtilsRetryDefaultTest` covers the semantics that actually changed: 401 no, 429 yes, 400 no, real timeout yes, explicit predicate respected.

Verdict: **comment** — two things I would like resolved before this merges, both about the scope of the behaviour change rather than the direction of the fix.

### Findings

- **[Warning]** `ModelUtils.java:119` — the new default is applied per provider only where an exception carries a status the predicate can read. OpenAI already did, Anthropic/DashScope/Ollama are adapted here, but Gemini is not, so a Gemini 429/5xx moves from "retried" to "not retried" as a side effect.
- **[Warning]** `AnthropicChatModel.java:266` — `onErrorResume` changes the exception type that leaves `doStream` for *all* subscribers, not just for the retry decision, so non-retryable failures stop arriving as `ModelException` (with its `modelName`/`provider` fields) even when retry is off. Narrowing the substitution to the statuses being classified keeps that surface stable.
- **[Info]** `AnthropicChatModel.java:298` — the raw provider error body ends up in the message that the retry logger prints; fine if intended, worth a conscious call.

### Suggestions

For the Anthropic surface, the one-line narrowing keeps both behaviours intact:

```java
if (current instanceof AnthropicServiceException ase
        && (ase.statusCode() == 429 || (ase.statusCode() >= 500 && ase.statusCode() < 600))) {
    return new HttpTransportException(...);
}
```

Everything else keeps flowing through untouched as `ModelException`, and `RETRYABLE_ERRORS` still sees a status-coded error exactly where the adaptation was needed.

### Cross-repo Note

`ModelUtils.applyTimeoutAndRetry` is shared by every model extension and by the `agentscope-extensions-rag-simple` embedding clients, so the retry-default change is wider than the five providers named in the commits. A short line in the changelog/release notes about the new default would save downstream surprises.

---

Letter2025:
﻿Thanks for the careful review — all four points addressed in `c06bcb2c`:

1. **Gemini not adapted (W1)** — `GeminiChatModel` now adapts genai SDK exceptions the same way as Anthropic: a real `ApiException` (and its `ClientException`/`ServerException` subclasses) with 429/5xx is rewritten to `HttpTransportException(status, no body, cause)` so the default `RETRYABLE_ERRORS` predicate classifies it correctly; `GenAiIOException` (transport failure without an HTTP status) adapts to a status-less `HttpTransportException` which stays retryable. Non-retryable statuses keep their original exception type.

2. **Anthropic `onErrorResume` rewrites for every subscriber (W2)** — narrowed: `adaptSdkException` now only rewrites retryable statuses (429 / 5xx); anything else flows through unchanged, so `ModelException` with `modelName`/`provider` stays the public surface for auth/400 failures.

3. **Provider error body in retry logs (Info)** — both adapters no longer carry the response body into `HttpTransportException`; only the status is preserved for classification, so per-retry log lines never echo payload fragments.

4. **Changelog (Note)** — added an `Unreleased → Compatibility` entry in release-notes (en/zh) documenting the new `retryOn` default.

Tests: `GeminiChatModelTest` 43/43 (incl. real-SDK 429/5xx/4xx/IO adaptation + a `doStream` retry-count regression), `AnthropicChatModelTest` 31/31 (incl. 400-keeps-type and no-body assertions), `ModelUtilsRetryDefaultTest` 7/7; spotless applied; CI re-running.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/GeminiChatModel.java
```diff
@@ -325,6 +337,54 @@ protected Flux<ChatResponse> doStream0(
...
+    static Throwable adaptSdkException(Throwable error) {
+        Throwable current = error;
+        while (current != null) {
+            if (current instanceof ApiException apie) {
+                int status = apie.code();
+                if (status == 429 || (status >= 500 && status < 600)) {
+                    return new HttpTransportException(
+                            "Gemini API request failed with status " + status, status, null, error);
+                }
+                return error;
```
oss-maintainer, line 373: **[Warning]** The `return error` inside the `ApiException` branch aborts the whole cause-chain walk, so a retryable fault nested *deeper* than a non-retryable `ApiException` (or a `GenAiIOException` wrapped underneath it) is never classified. If that ordering is intentional, a one-line comment saying "first hit wins by design" would help; otherwise `current = current.getCause(); continue;` keeps searching. Same shape in `AnthropicChatModel.adaptSdkException`.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/GeminiChatModel.java
```diff
@@ -325,6 +337,54 @@ protected Flux<ChatResponse> doStream0(
...
+            }
+            if (current instanceof GenAiIOException) {
```
oss-maintainer, line 375: **[Info]** `GenAiIOException` is matched after `ApiException`. If the genai SDK ever throws an `ApiException` subtype whose `code()` is a non-HTTP sentinel (0/-1) for a transport failure, this branch is unreachable and the failure lands as non-retryable. Worth a test that pins the hierarchy assumption (e.g. assert a `GenAiIOException` cause nested under a status-less `ApiException` still retries), or check `GenAiIOException` first.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
```diff
@@ -256,7 +258,57 @@ protected Flux<ChatResponse> doStream(
...
+     * provider error body is deliberately not carried, so the per-retry log lines never echo
+     * payload fragments that may contain user content. The original SDK exception is preserved
+     * as the cause for diagnostics.
+     *
+     * @param error the raw SDK exception (may be wrapped by the synchronous catch
+     *     {@link ModelException})
+     * @return an {@link HttpTransportException} when the cause chain carries a retryable
+     *     {@link AnthropicServiceException}, otherwise the original error unchanged
+     */
+    static Throwable adaptSdkException(Throwable error) {
```
oss-maintainer, line 295: **[Info]** Good narrowing on the privacy side (no provider body into the retry/log path). Two follow-ups: (1) the 429 message no longer carries any provider detail, so a support ticket built from logs alone loses the `retry-after`/error type — consider keeping the SDK `toString()` (no body) or the `retry-after` header; (2) `adaptSdkException` now exists per-provider with subtly different shapes (Anthropic: status-only; Gemini: status + IO branch). OpenAI/DashScope have no equivalent adapter, so if their SDK exceptions also lack `ModelHttpException`, the same 429-misclassification the Anthropic branch fixed is still live there. Extracting one core helper (`HttpTransportException forSdk(Throwable, int)` + a shared predicate) and applying it to all providers would keep retry semantics aligned as providers are added.

oss-maintainer (commented):
### Summary
Follow-up commit addresses my earlier `retryOn` review feedback well: Anthropic's adaptation is narrowed to genuinely retryable statuses (429/5xx) so non-retryable 4xx keep the stable `ModelException` type, the provider error body is dropped from the retry/log path, and Gemini gains the equivalent `adaptSdkException` so a real Gemini 429/5xx is no longer misclassified as terminal. New tests pin all three behaviours (400 passthrough, body-free 429, retryable classification) — nice. CLA signed, `MERGEABLE/BLOCKED` (blocked only on review state). Verdict: **COMMENT** — three non-blocking points below, no changes requested.

### Findings
- **[Warning]** `GeminiChatModel.java:373` — early `return error` aborts the cause-chain walk (mirrors Anthropic).
- **[Info]** `GeminiChatModel.java:375` — branch ordering pins an untested assumption about the genai SDK exception hierarchy.
- **[Info]** `AnthropicChatModel.java:295` — 429 log detail, and the per-provider duplication of `adaptSdkException` (OpenAI/DashScope still unadapted).

---

PR #2865, opened 2026-08-27
feat(harness): built-in ask_user tool — model-initiated questions to the user (HITL ask direction)

### LINKED ISSUE #2860: [Feature]: Built-in ask_user tool - model-initiated questions to the user (HITL ask direction)
### Problem

AgentScope's HITL is currently **one-directional**: the *user* decides whether a *tool* may run (permission ALLOW/DENY/ASK, `RequireUserConfirmEvent`, `GenerateReason.PERMISSION_ASKING`). There is no way for the **model** to proactively request *information* from the user during a run.

This is a standard capability in current coding-agent products:

| Product | Mechanism |
|---|---|
| OpenAI Codex CLI | `request_user_input` (id/header/question/isOther/isSecret/options, isBlocking + autoResolutionMs) |
| Claude Code / Agent SDK | `AskUserQuestion` tool (questions + choices, "type your own answer if none fit") |
| Cline | AskUserQuestion TUI (options + trailing free-text input row) |
| DeepSeek Harness | `ask_user_question` (questions/options/multi_select + custom free text) |

Open issues already asking for the "give new information to the agent" direction:
- agentscope-ai/agentscope#1120 — "[human in loop] how to interrupt the agent flow and give a confirm/**new information** to the agent to continue the flow"
- agentscope-ai/agentscope#926 — "[Feature]: human in loop wanted"
- agentscope-ai/agentscope-java#2168 — "[Bug]: Fail to continue Human-in-the-Loop"

### Proposed design

A framework-first-class, opt-in capability:

1. **New permission decision signal** — `PermissionBehavior.ASK_USER` + `PermissionDecision.askUser(...)`. It is produced only by a tool's own `checkPermissions()` (i.e. the bypass-immune tool-check phase), so it interrupts in **every** `PermissionMode` (incl. `BYPASS`), and is not registerable as a rule.
2. **Built-in `AskUserTool`** (registered by `HarnessAgent.Builder.enableAskUser()`) exposing a JSON-schema'd `ask_user` tool: `questions[] {id, header, question, type: single|multiple|free|secret, options[], required}`, batch questions in one tool call, `required` skip support.
3. **Distinct interrupt semantics** — `GenerateReason.ASK_USER_ASKING` (parallel to `PERMISSION_ASKING`), new `RequireUserAskEvent` (parallel to `RequireUserConfirmEvent`, carries the pending `ToolUseBlock`s with the questions). The tool is **never executed**; the framework pauses, the caller renders the questions, then resumes with answers.
4. **First-class resume** — new `AskUserResult` + `Msg.METADATA_ASK_USER_RESULTS` (parallel to `ConfirmResult` / `METADATA_CONFIRM_RESULTS`); the framework formats the answers into the `ask_user` tool result itself, so the next reasoning iteration reads the user's answers without re-executing the tool. Streaming gets `UserAskResultEvent` for correlation.

### Scope

- `agentscope-core`: decision signal, gate routing, new event(s), `GenerateReason`, resume payload handling (parallel to the existing `PERMISSION_ASKING` machinery).
- `agentscope-harness`: `AskUserTool` + `enableAskUser()` builder switch (opt-in, default off).
- Documentation + runnable example + tests (incl. BYPASS-mode interrupt, multi-question batch, skip semantics, resume-does-not-execute).

### Non-goals (for this PR)

- No AG-UI / protocol bridging (that's app-layer work).
- No secret masking policy, prompt-segment engineering, or question frequency limits — these are host-application concerns.

### API sketch

```java
        .model(model)
        .enableAskUser()   // registers ask_user into the toolkit
        .build();
```

```java
// call returns Msg with GenerateReason.ASK_USER_ASKING;
// subsequent agent.call(List.of(resumeMsg)) with
//   Msg.METADATA_ASK_USER_RESULTS = List.of(new AskUserResult(toolCallId, answers))
// continues the run; the ask_user call is not executed.
```

Letter2025:
## Summary

Adds the HITL "ask" direction to AgentScope: **the model can proactively ask the user structured questions during a run**, complementing the existing user-confirms-tool flow.

A tool whose `checkPermissions()` returns `PermissionDecision.askUser(...)` pauses the run with the new `GenerateReason.ASK_USER_ASKING` — in **every** `PermissionMode` (including `BYPASS`, since it rides the bypass-immune tool self-check) — and is **never executed**. Callers render the questions and resume with `AskUserResult`(s) under `Msg.METADATA_ASK_USER_RESULTS`; the framework formats the answers into the `ask_user` tool result so the model reads them on the next reasoning iteration.

The harness ships a built-in `ask_user` tool (JSON-schema'd `questions[]`: id/header/question/type single|multiple|free|secret/options/required), enabled via `HarnessAgent.Builder.enableAskUser()` (opt-in, off by default — no behavior change for existing builders).

## Motivation

HITL today only answers "**may this tool run?**" (`PERMISSION_ASKING` + `RequireUserConfirmEvent`). There is no way for the model to request *information* from the user. This is a standard capability in current coding agents (OpenAI Codex `request_user_input`, Claude Code / Agent SDK `AskUserQuestion`, Cline, DeepSeek Harness `ask_user_question`) and has been requested in:
- agentscope-ai/agentscope#1120 — "give a confirm/**new information** to the agent to continue the flow"
- agentscope-ai/agentscope#926 — "[Feature]: human in loop wanted"
- agentscope-ai/agentscope-java#2168 — "Fail to continue Human-in-the-Loop"

## Core changes (`agentscope-core`)

| Area | Change |
|---|---|
| `PermissionBehavior` | new `ASK_USER` constant (not registerable as a rule; produced only by a tool's `checkPermissions()`) |
| `PermissionDecision` | new `askUser(String)` factory |
| `PermissionEngine` | `ASK_USER` short-circuits immediately in the bypass-immune tool-sef-check phase (not swallowed by rule/BYPASS handling) |
| `ReActAgent` | gate routes `ASK_USER` to a dedicated pending list; `actingStream` emits `RequireUserAskEvent` + `RequestStopEvent(ASK_USER_ASKING)`; `acting()` returns the pausing assistant `Msg`; resumes with answers via `validateAndAcceptAskUserResults` which formats them into the tool result without executing it |
| Events | new `RequireUserAskEvent` / `UserAskResultEvent` (replyId-correlated, parallel to the confirm flow; registered in `AgentEvent` subtypes) |
| `GenerateReason` | new `ASK_USER_ASKING` |

## Harness changes (`agentscope-harness`)

- Built-in `AskUserTool` in `io.agentscope.harness.agent.tool` (name `ask_user`, `checkPermissions()` → `askUser(...)`)
- `HarnessAgent.Builder.enableAskUser()` (no-arg + boolean overload), registers the tool at `build()`

## Docs & examples

- `docs/v1/{en,zh}/docs/task/hitl.md` — new "Ask the user for input" section
- `docs/v2/{en,zh}/docs/building-blocks/permission-system.md` — new "Ask the user (model-initiated questions)" section (incl. `DONT_ASK` note)
- `agentscope-examples/.../hitl/AskUserHITLExample.java` — runnable interactive example

## Tests

- `ReActAgentAskUserTest` (core): pause with `ASK_USER_ASKING`, tool not executed, pause under `BYPASS`, resume → answers visible in next reasoning, resume-without-answers guidance error, streaming `RequireUserAskEvent`/`UserAskResultEvent` correlation, answer formatting (scalar/list/selected+text/skipped)
- `HarnessAgentAskUserTest` (harness): `enableAskUser()` registers the tool and pauses; BYPASS still pauses; full resume round-trip (answer reaches the model, `ask_user` not executed); tool self-description
- Existing HITL/permission suites stay green

## Compatibility

Opt-in feature: existing `HarnessAgent`/`ReActAgent` builders are unaffected (tool not registered, no new gate entries, `GenerateReason` gains one value). `ASK_USER` is not a rule behavior, so no permission-config surface changes.

Closes #2860


Letter2025:
﻿Non-blocking note on CI: the remaining build failures are a pre-existing harness test flake, not caused by this PR.

The failing jobs report `Failed to close extension context` / `Failed to delete temp directory` at JUnit `@TempDir` teardown time, in a *different* test class on every run:

- a7250f90 → `HarnessAgentSubagentStreamTest` + `HarnessAgentIntegrationExampleTest`
- 64c83f3a → `HarnessAgentDynamicHookBuilderTest`
- a0d565a4 → `HarnessAgentSubagentStreamTest` (again)
- 75c73811 → `JsonSessionDefaultLocationTest`

The same failure pattern hit main's own CI on an unrelated commit (guava bump, [run 32981277153](https://github.com/agentscope-ai/agentscope-java/actions/runs/32981277153) — `HarnessAgentSubagentStreamTest.call_localSubagent_returnsReplyWithoutStreaming` -> `Failed to close extension context`), and all of these classes pass locally (and pass on CI in runs where they are not the flaky one). Root-cause hypothesis: harness default memory flush/maintenance middleware keeps writing workspace files on a background scheduler after the test body completes, racing `@TempDir` cleanup.

The deterministic failures caused by this PR were fixed and verified:
- added `ASK_USER` to `PermissionBehaviorTest` expectations
- added `REQUIRE_USER_ASK` / `USER_ASK_RESULT` samples to `RemoteEventCodecPassthroughTest`
- added `@JsonCreator`/`@JsonProperty` to `AskUserResult` so `UserAskResultEvent` round-trips through `RemoteEventCodec`

Happy to rebase/adapt if maintainers prefer a different approach.

Letter2025:
﻿Hi @mikemikimike @helloworldtang @wzq-xzwj — could someone rerun the failed CI jobs on this PR? The failure is a pre-existing harness `@TempDir` teardown flake, unrelated to the change itself:

- The only failing test in the last runs is `HarnessAgentSubagentStreamTest.call_localSubagent_returnsReplyWithoutStreaming` (and in earlier runs `HarnessAgentDynamicHookBuilderTest` / `JsonSessionDefaultLocationTest`) — each failing at `Failed to close extension context` / `Failed to delete temp directory` during teardown, with the test bodies passing.
- The same failure appeared on main's own CI for the unrelated guava-bump commit ([run 32981277153](https://github.com/agentscope-ai/agentscope-java/actions/runs/32981277153): same test class, same `Failed to close extension context`), and all these classes pass locally and on CI in runs where they are not the flaky one.
- Suspected root cause: the harness default memory flush/maintenance middleware keeps writing workspace files on a background scheduler after the test body completes, racing `@TempDir` cleanup. Happy to file/adjust a follow-up if a global fix is preferred.

This branch already verified locally: full core + harness suites pass (only known local-only errors: `DangerousPathBypassTest` symlink cases on Windows without symlink privileges).

oss-maintainer:

```bash
git fetch origin
git checkout feat/ask-user-tool
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AskUserTool.java
```diff
@@ -0,0 +1,188 @@
...
+                Map.of(
+                        "type",
+                        "string",
+                        "enum",
+                        List.of("single", "multiple", "free", "secret"),
+                        "default",
+                        "single",
+                        "description",
+                        "single=choose one option; multiple=choose any number of options; "
+                                + "free=free-text answer; secret=masked input (API keys, "
```
oss-maintainer, line 139: `secret` is advertised in the schema as "masked input (API keys, passwords) — the host must not persist or log the raw value", but the framework persists the raw value no matter what the host does, so the guarantee cannot be honoured from the host side:

1. `ReActAgent.validateAndAcceptAskUserResults` renders every answer verbatim through `AskUserResult.formatAnswers(...)` into the `ask_user` tool result and appends it to `state.contextMutable()` (ReActAgent line 2016) — that is the persisted agent state, which is exactly what an `AgentStateStore` writes and what session recovery replays.
2. `publishEvent(new UserAskResultEvent(replyId, results))` (line 2026) carries the same raw answers on the event stream, and `AskUserResult.answers` is a `@JsonProperty`, so the values are serialised for tracing exporters and for the sub-agent / distribution wire path.
3. The next reasoning turn sends them to the model provider in the request body.

Nothing downstream can redact them either: `answers` is keyed by question id with no type information attached, so neither the state store nor an event sink can tell which entry was a `secret`. Since this repo's own review bar is that credentials must not leak into logs or events, I would not merge the `secret` variant in this shape. Two options, in rough order of preference:

- **Drop `secret` from the enum** (`single` / `multiple` / `free`) and reintroduce it later with a redaction design. Collecting a password through a model-visible tool result is questionable regardless of masking, because the secret still transits the model context and the provider.
- Or make the answer self-describing so the framework can redact it — e.g. accept `{"q_1": {"value": "...", "secret": true}}` — and in `validateAndAcceptAskUserResults` substitute a fixed placeholder for both the persisted tool result and `UserAskResultEvent`, so the raw value is never written anywhere. That needs a test asserting the secret string appears neither in `state.contextMutable()` nor in the serialised event.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2855,23 +2985,35 @@ Flux<AgentEvent> actingStream(
...
                                     return runToolBatch(
                                             toolCalls, autoDenied, replyId, resultHolder);
                                 }
 
+                                // Ask-user HITL: the model asked the user questions. Surface the
+                                // pending ask_user calls, persist the reply id for resume
+                                // correlation, then signal stop via RequestStopEvent. The agent's
+                                // acting() will set GenerateReason to ASK_USER_ASKING and return;
+                                // the tool is never executed.
+                                if (!pendingAskUser.isEmpty()) {
```
oss-maintainer, line 3004: This early return skips the rest of the block, including `writeAutoDeniedResults(toolCalls, autoDenied)` at line 3022. `updateToolCallStates` deliberately leaves auto-denied calls out of the state map on the assumption that a DENIED `ToolResultBlock` is written here ("DENIED tools don't need a state change — they'll get a DENIED ToolResultBlock", line 2982), so in a batch where the model emits `ask_user` alongside a rule-denied call, that call ends up with a `tool_use` in context and no matching `tool_result` — which the Anthropic-style APIs this code targets reject outright on the next turn. The run also looks like a clean question interrupt from the host's side, so the breakage is far from its cause.

Lifting the denied write above the interrupt keeps the pause shape intact:

```java
if (!pendingAskUser.isEmpty()) {
    resultHolder.set(List.of());
    if (!autoDenied.isEmpty()) {
        writeAutoDeniedResults(toolCalls, autoDenied);
    }
    persistPendingRequestReplyId(Msg.METADATA_ASK_REQUEST_REPLY_ID, replyId);
    return Flux.<AgentEvent>just(
            new RequireUserAskEvent(replyId, pendingAskUser),
            new RequestStopEvent("ask user", GenerateReason.ASK_USER_ASKING));
}
```

The same branch also drops `pending` (permission confirmations): only `RequireUserAskEvent` is emitted, so the host is never told those calls are waiting, while `updateToolCallStates` has already parked them in `ASKING` — resuming the answers clears the question and leaves the confirmation gate pending with no event ever having surfaced it. Worth deciding explicitly: either surface both events in one pause, or reject the mixed batch at gate construction with a clear message. `ReActAgentAskUserTest` covers single-kind pauses only, so neither outcome is currently pinned.

agentscope-core/src/main/java/io/agentscope/core/event/AskUserResult.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @JsonCreator
+    public AskUserResult(
+            @JsonProperty("toolCallId") String toolCallId,
+            @JsonProperty("answers") java.util.Map<String, Object> answers) {
+        if (toolCallId == null || toolCallId.isEmpty()) {
+            throw new IllegalArgumentException("AskUserResult.toolCallId must not be empty");
+        }
+        this.toolCallId = toolCallId;
+        this.answers = answers != null ? java.util.Map.copyOf(answers) : java.util.Map.of();
```
oss-maintainer, line 50: `Map.copyOf` does not preserve insertion order, but `formatAnswers` walks `answers.entrySet()` to build the text the model reads — so the host's question order is not what comes back out. The javadoc promises "a stable, human-readable rendering", and `formatAnswersHandlesScalarsListsMapsAndSkips` asserts `"q_a: 1\nq_b: 2"`, but it builds a `LinkedHashMap` and calls `formatAnswers(multi)` directly, bypassing this constructor — the ordering that actually reaches a real resume goes through `Map.copyOf`, which is unspecified. Two keys survive by accident; three or more is where the sequence visibly diverges, and it costs prompt-prefix stability for anything caching the conversation.

```java
this.answers = answers != null
        ? Collections.unmodifiableMap(new LinkedHashMap<>(answers))
        : Map.of();
```

and it is worth re-running the ordering assertion through `new AskUserResult(id, multi).getAnswers()` so the test covers the path callers take rather than only the static helper.

Minor, same file: `java.util.Map` / `java.util.List` are written fully-qualified throughout, which is out of step with the rest of `io.agentscope.core.event` — plain imports would match the surrounding style.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1910,6 +1920,121 @@ private void validateAndAcceptConfirmResults(List<Msg> msgs, List<ToolUseBlock>
...
+                                + "\", e.g.:\n"
+                                + "    UserMessage.builder()\n"
+                                + "        .metadata(Map.of(Msg.METADATA_ASK_USER_RESULTS,\n"
+                                + "            List.of(new AskUserResult(toolCallId, answers))))\n"
+                                + "        .build();\n"
+                                + "Tip: capture the ToolUseBlocks from the RequireUserAskEvent"
+                                + " emitted when the agent paused.");
+            }
+
+            Set<String> expectedIds =
```
oss-maintainer, line 1991: `expectedIds` comes from `asking = askingToolCalls()`, and `updateToolCallStates` (line 2988) marks *both* gated kinds with the same `ToolCallState.ASKING` — there is no marker distinguishing "paused for a question" from "paused for a permission decision". So an `AskUserResult` whose `toolCallId` names a permission-pending call passes validation here, and its formatted answers are written as that call's tool result, clearing the permission gate without any `ConfirmResult` ever having been supplied. The host is trusted either way, so this is not privilege escalation, but the two HITL channels become interchangeable at resume and a host that mixes up its payloads gets a plausible-looking wrong transcript instead of an error — and `ToolResultBlock.text(formatAnswers(...))` has no `DENIED`/`ALLOWED` state, so the transcript no longer records that a permission decision was ever pending.

Since `RequireUserAskEvent` already carries only the question calls, the cheapest discriminator is the tool name (`AskUserTool.TOOL_NAME`, or any tool whose `checkPermissions()` returned `ASK_USER`) recorded on the block alongside `ASKING`. That would also let `doCallInner` route on the pending kind instead of `isAskUserPaused() || hasAskUserResults(msgs)` taking precedence over the confirm branch at line 1769.

oss-maintainer (commented):
﻿### Summary

Adds the "model asks the user" direction of HITL: a new non-bypassable `PermissionBehavior.ASK_USER` decision, a `RequireUserAskEvent` / `UserAskResultEvent` pair correlated by `replyId` (same shape as the existing confirmation flow), an `ASK_USER_ASKING` generate reason, and an opt-in `ask_user` tool in the harness whose `checkPermissions()` always interrupts so the tool itself never runs. Answers are formatted into the paused call's tool result and written to context, which is the right way to make them visible to the next reasoning turn without inventing a second message channel. Scoping is careful where it counts: `HarnessAgent.Builder.enableAskUser()` defaults to off, the decision short-circuits before rule/BYPASS handling rather than registering as a rule, `PermissionVerdict` gained the extra arm, and both the English and Chinese HITL/permission docs plus a runnable example were updated. Test volume is good for a first contribution — 405 lines of `ReActAgent` pause/resume tests, 213 harness lines, event round-trip and `RemoteEventCodecPassthrough` coverage, all green on Linux and Windows, CLA signed.

Verdict: **comment**. One thing I would hold the merge on and three smaller ones; the design itself I would not change.

### Findings

- **[Critical]** `AskUserTool.java:139` — the schema offers `type: "secret"` for "API keys, passwords" and instructs the host not to persist the raw value, but the framework persists and broadcasts it: `validateAndAcceptAskUserResults` writes `formatAnswers(...)` into `state.contextMutable()` (persisted agent state, replayed on session recovery), publishes `UserAskResultEvent` with the raw answers (serialised for tracing and the sub-agent/distribution path), and the next turn ships them to the provider. Nothing downstream can redact them, because `answers` carries no type information. Either drop `secret` from the v1 enum, or make the answer self-describing and redact at both write sites.
- **[Warning]** `ReActAgent.java:3004` — the ask-user early return skips `writeAutoDeniedResults` (line 3022), so a batch mixing `ask_user` with a rule-denied call leaves a `tool_use` with no matching `tool_result`; and the same return drops `pending`, so permission confirmations in that batch are parked as `ASKING` without ever emitting `RequireUserConfirmEvent`.
- **[Warning]** `AskUserResult.java:50` — `Map.copyOf` loses insertion order, while `formatAnswers` renders in iteration order and the javadoc promises a "stable" rendering; the ordering test bypasses the constructor, so the path callers take is not the path asserted.
- **[Warning]** `ReActAgent.java:1991` — `ASKING` does not distinguish a question pause from a permission pause, so an `AskUserResult` naming a permission-pending call clears that gate with formatted "answers" and no `ConfirmResult` state recorded. Not privilege escalation (the host supplies both), but the two channels are interchangeable at resume, which makes a host bug silent rather than loud.

### Suggestions

The ordering fix is a one-liner (`Collections.unmodifiableMap(new LinkedHashMap<>(answers))`), and the auto-denied fix is a two-line move of the existing guard above the interrupt. The `secret` item is the one that needs a decision rather than a patch — I would take the enum out for now, since even masked, a credential that transits the model context and the event stream is not a design you can later call safe.

`PermissionGate` growing a third list also makes `isInAny` scan two lists per tool call; with `pending` and `pendingAskUser` both small this is irrelevant, just noting it is the kind of thing that wants a `Set<String>` if a third pause kind ever arrives.

### Cross-repo Note

`PermissionBehavior` / `AgentEventType` / `GenerateReason` are `agentscope-core` surfaces that cascade into harness, distribution and extensions. A rule configuration in `agentscope-extensions-nacos-*` or an external studio consumer that parses `ask_user` will get an unrecognised value, and the two new event subtypes change what older consumers must tolerate — worth a note in the release description even though `AgentEvent` deserialisation by `@JsonSubTypes` name keeps existing payloads working.

---

PR #2868, opened 2026-08-27
fix(core): default model retry to RETRYABLE_ERRORS when retryOn is unset

### LINKED ISSUE #2861: [Bug]: ModelUtils retries auth/request-side errors when retryOn is unset, contradicting RETRYABLE_ERRORS docs

chaucerj:
## What

`ModelUtils.applyTimeoutAndRetry` fell back to `retryOn = error -> true` (retry **every** error) when no `retryOn` predicate was configured. This contradicts the documented default of `ExecutionConfig`, where only retryable errors — HTTP 429/5xx, timeout, network/IO — are retried, and non-retryable errors (400, 401/403, other 4xx) fail fast.

## Why it matters

Without this fix, a model call that retries with the default config (e.g. `maxAttempts=3` without an explicit `retryOn`) will:

- retry authentication failures (401/403) that can never succeed on retry, wasting backoff time (up to `maxBackoff` per attempt) before surfacing the real error;
- retry unknown programming errors, masking request-side bugs behind repeated failing calls.

`ExecutionConfig.MODEL_DEFAULTS` already sets `retryOn(RETRYABLE_ERRORS)`, so this only affects configs built without an explicit predicate (custom `ExecutionConfig` or merged configs where `retryOn` is null).

## Change

- `ModelUtils.applyTimeoutAndRetry`: when `retryOn` is null, default to `ExecutionConfig.RETRYABLE_ERRORS` instead of `error -> true`. One-line behavior change plus javadoc.
- New test `ModelUtilsRetryDefaultsTest` covering the default policy: 401 → 1 attempt (no retry), unknown error → 1 attempt, transient IOException → retried and succeeds.

## Behavior comparison (retryOn unset)

| error | before | after |
|---|---|---|
| HTTP 401/403, 4xx | retried until exhausted | fails fast (1 attempt) |
| unknown `RuntimeException` | retried | fails fast |
| HTTP 429/5xx, timeout, IOException | retried | retried (unchanged) |

Fixes #2861

---

*This patch was developed with AI assistance (Claude Code); the change, tests and PR description were reviewed by the author.*

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
@@ -55,7 +55,9 @@ public final class ModelUtils {
      * <ul>
      *   <li>If RetryConfig is provided, failed requests will be retried with exponential backoff
      *   <li>Retries respect the maxAttempts, initialBackoff, and maxBackoff settings
-     *   <li>Only errors matching the retryOn predicate will be retried
+     *   <li>Only errors matching the retryOn predicate will be retried; when no predicate is
+     *       configured, only retryable errors (HTTP 429/5xx, timeout, network/IO) are retried,
+     *       per {@link ExecutionConfig#RETRYABLE_ERRORS}
      *   <li>Each retry is logged with attempt number and failure reason
      * </ul>
      *
@@ -108,7 +110,10 @@ public final class ModelUtils {
                     maxBackoff = Duration.ofSeconds(10);
                 }
                 if (retryOn == null) {
-                    retryOn = error -> true; // retry all errors by default
+                    // Default to the documented retryable-error policy (429/5xx, timeout,
+                    // network/IO) instead of retrying every error, so that auth (401/403)
+                    // and other request-side failures fail fast
+                    retryOn = ExecutionConfig.RETRYABLE_ERRORS;
                 }
 
                 Retry retrySpec =
--- a/agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsRetryDefaultsTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/model/ModelUtilsRetryDefaultsTest.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.model;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import io.agentscope.core.message.TextBlock;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Flux;
+import reactor.test.StepVerifier;
+
+/**
+ * Tests for the default retry policy of {@link ModelUtils#applyTimeoutAndRetry} when no {@code
+ * retryOn} predicate is configured: the default must follow {@link ExecutionConfig#RETRYABLE_ERRORS}
+ * instead of retrying every error.
+ */
+@DisplayName("ModelUtils default retry policy when retryOn is unset")
+class ModelUtilsRetryDefaultsTest {
+
+    @Test
+    @DisplayName("Should not retry non-retryable (401) errors when retryOn is unset")
+    void shouldNotRetryAuthErrorsByDefault() {
+        AtomicInteger attempts = new AtomicInteger();
+        Flux<ChatResponse> flux =
+                Flux.defer(
+                        () -> {
+                            attempts.incrementAndGet();
+                            return Flux.error(new TestModelHttpException(401));
+                        });
+
+        StepVerifier.create(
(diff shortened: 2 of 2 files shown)


agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
```diff
@@ -108,7 +110,10 @@ public static Flux<ChatResponse> applyTimeoutAndRetry(
                     maxBackoff = Duration.ofSeconds(10);
                 }
                 if (retryOn == null) {
-                    retryOn = error -> true; // retry all errors by default
+                    // Default to the documented retryable-error policy (429/5xx, timeout,
+                    // network/IO) instead of retrying every error, so that auth (401/403)
+                    // and other request-side failures fail fast
+                    retryOn = ExecutionConfig.RETRYABLE_ERRORS;
```
Aias00, line 116: This default is correct in isolation, but it is not safe without the two companion changes that #2862 carries, and landing it alone **regresses timeouts**.

`ExecutionConfig.isRetryableError` only recognises `TimeoutException` by *type or cause chain*. The timeout this method produces 20 lines above is:

```java
new ModelException("Model request timeout after " + timeout, modelName, provider)
```

`ModelException(String, String, String)` sets no cause, so the chain is empty -> not `TimeoutException`, not `IOException`, not `ModelHttpException` -> `isRetryableError` returns `false`. With `retryOn = error -> true` today the timeout is retried; with this change it is not. Timeouts are the single most valuable thing to retry, and the javadoc you are updating explicitly lists them as retryable.

Second regression: the Anthropic / DashScope / Ollama SDK exceptions do not implement `ModelHttpException` (or `HttpTransportException`), so a real 429/5xx from those providers is classified as an unknown error and stops after one attempt. #2862 adapts each of them.

Since #2862 already contains this exact line **plus** the `TimeoutException` cause and the three provider adapters, I would close this PR as a duplicate and land #2862 alone. If you would rather keep this one, please at minimum (a) add `new TimeoutException(...)` as the cause of the timeout `ModelException` and (b) cover the timeout path in `ModelUtilsRetryDefaultsTest` — the current three tests never exercise it.

Aias00 (changes requested):
The intent is right and the analysis in the description is right: `retryOn = error -> true` contradicts the documented `ExecutionConfig` default, and the behaviour table is accurate.

I am requesting changes because this PR is a **partial duplicate of #2862** (same author, same file, same line) and, taken on its own, it introduces two regressions rather than fixing one:

1. **Timeouts stop being retried.** `ExecutionConfig.isRetryableError` detects timeouts by type or cause chain, but the `ModelException` raised by the timeout operator in this very method is built with the `ModelException(String, String, String)` constructor, which sets no cause. It is therefore classified as an unknown error and not retried — the opposite of what the javadoc you are updating promises. #2862 fixes this by passing a `TimeoutException` as the cause.
2. **Provider 429/5xx stop being retried.** The Anthropic, DashScope and Ollama SDK exceptions do not implement `ModelHttpException` / `HttpTransportException`, so they too fall through to "unknown -> do not retry". #2862 adds adapters for all three.

`ModelUtilsRetryDefaultsTest` only covers 401, an unknown `RuntimeException` and a raw `IOException`, so neither regression is caught.

Recommendation: close this PR as a duplicate and land #2862, which contains this line plus both companion fixes. If you prefer to keep this one, details are inline. Either way, please add a test that drives an actual `execConfig.timeout(...)` through `applyTimeoutAndRetry` and asserts the retry happens.

agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
chaucerj: You are right, and thank you for the detailed analysis — I had not traced the timeout path through `isRetryableError`. Confirmed: the timeout `ModelException` from the 3-arg constructor carries no cause, so under `RETRYABLE_ERRORS` it would land in the unknown-error bucket and stop after one attempt. That is a real regression this PR does not fix, and #2862 carries the complete change (the default plus the `TimeoutException` cause and the provider adapters).

Closing this as a duplicate — #2862 should land alone. One note for the tracking issue: could you link #2861 to #2862 so it auto-closes when that merges?

chaucerj:
Closing as a duplicate of #2862, which carries this default change plus the timeout-cause fix and the provider adapters needed to make it safe. Thanks for the review!

PR #2869, opened 2026-08-27
fix(core): log tool call argument JSON parse failures instead of swallowing them

### LINKED ISSUE #2841: [Bug]: ToolCallsAccumulator silently swallows JSON parse failures for streamed tool call arguments and the raw string is lost

chaucerj:
## What

Implements the **minimum** fix requested in #2841: `ToolCallsAccumulator$ToolCallBuilder.build()` caught JSON parse failures of the accumulated tool-call arguments with an empty handler — `catch (Exception ignored)` — so a malformed accumulated string was silently replaced by `input = {}` / `content = "{}"` with no log, event or trace anywhere in the framework.

The catch handler now logs a `warn` including the tool name, the **raw accumulated string** and the underlying exception. Runtime behavior is otherwise unchanged: previously merged args are still kept and `content` still falls back to `"{}"` (that fallback is intentional per the existing comment about interrupted streams).

## Why

With models that emit broken-but-complete argument JSON (the reporter's example: unescaped double quotes inside string values, seen with Qwen via OpenAI-compatible streaming), the tool ends up invoked with an empty parameter map and fails with its own missing-parameter error, which points debugging in the wrong direction. Today the only way to find the real cause is decompiling the jar. One log line makes the whole failure class diagnosable.

## Scope

Deliberately limited to observability. The second (discussion) item in #2841 — exposing the raw arguments on `ToolUseBlock` for app-side repair — is an API design question and is left for maintainers to decide; the log already carries the raw string for diagnosis.

## Testing

Existing `ToolCallsAccumulatorTest` covers the `"{}"` fallback path (15 tests, all green). The repo has no SLF4J backend on the test classpath and no log-assertion precedent, so the log line itself has no new test — verified manually that the warn fires for the issue's reproduction case.

Fixes #2841

---

*This patch was developed with AI assistance (Claude Code); the change and PR description were reviewed by the author.*

### DIFF
--- a/agentscope-core/pom.xml
+++ b/agentscope-core/pom.xml
@@ -203,6 +203,13 @@
             <artifactId>kubernetes-server-mock</artifactId>
             <scope>test</scope>
         </dependency>
+        <!-- SLF4J backend for tests that assert log output (e.g. accumulator parse warnings) -->
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>1.5.18</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/agentscope-core/pom.xml
+++ b/agentscope-core/pom.xml
@@ -215,7 +215,9 @@
     <build>
         <plugins>
             <!-- Publish a test-jar so downstream modules (agentscope-harness) can reuse the
-                 shared test fixtures (MockModel, MockToolkit, TestConstants, TestUtils). -->
+                 shared test fixtures (MockModel, MockToolkit, TestConstants, TestUtils).
+                 logback-test.xml is excluded so it cannot override downstream modules'
+                 own test logging configuration. -->
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
@@ -224,6 +226,11 @@
                         <goals>
                             <goal>test-jar</goal>
                         </goals>
+                        <configuration>
+                            <excludes>
+                                <exclude>logback-test.xml</exclude>
+                            </excludes>
+                        </configuration>
                     </execution>
                 </executions>
             </plugin>
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
@@ -283,9 +283,18 @@ public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
      */
     public List<ToolUseBlock> buildAllToolCalls() {
         // Finalization boundary: the stream has ended, so malformed accumulated JSON is a
-        // real defect worth reporting — once per tool call, sanitized
-        builders.values().forEach(ToolCallBuilder::warnIfMalformedOnce);
-        return builders.values().stream().map(ToolCallBuilder::build).collect(Collectors.toList());
+        // real defect worth reporting — once per tool call, sanitized. build() runs first so
+        // parseFailure reflects the FINAL raw content: a prefix that failed mid-stream but
+        // completed validly must not warn, and a call that was never built mid-stream must
+        // still warn on this first finalization.
+        return builders.values().stream()
+                .map(
+                        builder -> {
+                            ToolUseBlock block = builder.build();
+                            builder.warnIfMalformedOnce();
+                            return block;
+                        })
+                .collect(Collectors.toList());
     }
 
     /**
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
@@ -59,6 +59,10 @@ public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
         Map<String, Object> args = new HashMap<>();
         StringBuilder rawContent = new StringBuilder();
         Map<String, Object> metadata = new HashMap<>();
+        // Name of the exception from the last failed JSON parse of rawContent, cleared on a
+        // successful parse; only used to report the failure once at finalization
+        String parseFailure;
+        boolean parseWarned;
 
         void merge(ToolUseBlock block) {
             // Update ID if present
@@ -106,6 +110,7 @@ public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
                     @SuppressWarnings("unchecked")
                     Map<String, Object> parsed =
                             JsonUtils.getJsonCodec().fromJson(rawContentStr, Map.class);
+                    parseFailure = null;
                     if (parsed != null) {
                         for (Map.Entry<String, Object> entry : parsed.entrySet()) {
                             if (entry.getValue() == null) {
@@ -118,15 +123,12 @@ public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
                         }
                     }
                 } catch (Exception e) {
-                    // Parsing failed, keep previously merged args; log the raw string so a
-                    // malformed tool-call argument is diagnosable instead of silently lost
-                    LOG.warn(
-                            "Failed to parse accumulated tool call arguments as JSON for tool"
-                                    + " '{}'; falling back to previously merged args. Raw"
-                                    + " arguments: {}",
-                            name,
-                            rawContentStr,
-                            e);
+                    // Incomplete JSON prefixes are expected while streamed fragments are still
+                    // arriving, so build() stays silent here; buildAllToolCalls() reports the
+                    // failure once at finalization if the raw content never became valid JSON.
+                    // Only the exception name is kept: its message may quote raw arguments,
+                    // which can contain credentials or other sensitive data.
+                    parseFailure = e.getClass().getSimpleName();
                 }
             }
 
@@ -149,6 +151,28 @@ public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
                     .content(contentStr)
                     .metadata(metadata.isEmpty() ? null : metadata)
                     .build();
+        }
+
+        /**
+         * Emits at most one sanitized warning when the accumulated raw content never parsed as
+         * valid JSON.
+         *
+         * <p>Called only from the finalization boundary ({@link #buildAllToolCalls()}), not from
+         * per-fragment builds, because incomplete JSON prefixes are normal mid-stream. The
+         * warning omits the raw arguments (they may contain credentials, PII, or user
+         * documents) and carries no stack trace.
+         */
+        private void warnIfMalformedOnce() {
+            if (parseFailure != null && !parseWarned) {
+                parseWarned = true;
+                LOG.warn(
+                        "Tool call '{}' arguments are not valid JSON after the stream ended"
+                                + " (raw length: {}); using partially accumulated arguments"
+                                + " instead. Cause: {}",
+                        name,
+                        rawContent.length(),
+                        parseFailure);
+            }
         }
 
         private boolean isPlaceholder(String name) {
@@ -258,6 +282,9 @@ public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
      * @return List of tool calls
      */
     public List<ToolUseBlock> buildAllToolCalls() {
+        // Finalization boundary: the stream has ended, so malformed accumulated JSON is a
+        // real defect worth reporting — once per tool call, sanitized
+        builders.values().forEach(ToolCallBuilder::warnIfMalformedOnce);
         return builders.values().stream().map(ToolCallBuilder::build).collect(Collectors.toList());
     }
 
--- a/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
@@ -23,6 +23,8 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Tool calls accumulator for accumulating streaming tool call chunks.
@@ -38,6 +40,8 @@ import java.util.stream.Collectors;
  * @hidden
  */
 public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ToolCallsAccumulator.class);
 
     // Map to support multiple parallel tool calls
     // Key: tool identifier (ID, name, or index)
@@ -113,8 +117,16 @@ public class ToolCallsAccumulator implements ContentAccumulator<ToolUseBlock> {
                             }
                         }
                     }
-                } catch (Exception ignored) {
-                    // Parsing failed, keep previously merged args
+                } catch (Exception e) {
+                    // Parsing failed, keep previously merged args; log the raw string so a
+                    // malformed tool-call argument is diagnosable instead of silently lost
+                    LOG.warn(
+                            "Failed to parse accumulated tool call arguments as JSON for tool"
+                                    + " '{}'; falling back to previously merged args. Raw"
+                                    + " arguments: {}",
+                            name,
+                            rawContentStr,
+                            e);
                 }
             }
 
--- a/agentscope-core/src/test/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulatorParseWarningTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulatorParseWarningTest.java
@@ -71,12 +71,14 @@ class ToolCallsAccumulatorParseWarningTest {
     void validMultiFragmentEmitsNoWarning() {
         accumulator.add(
                 ToolUseBlock.builder().id("call-1").name("get_weather").content("{\"ci").build());
-        accumulator.add(
-                ToolUseBlock.builder().name("__fragment__").content("ty\": \"Shanghai\"}").build());
 
         // Per-fragment build path (hook dispatch during streaming) sees an incomplete JSON
-        // prefix and must not warn
+        // prefix: parse fails here, but the prefix later completes, so finalization must not
+        // emit the stale failure
         assertNotNull(accumulator.getAccumulatedToolCall("call-1"));
+
+        accumulator.add(
+                ToolUseBlock.builder().name("__fragment__").content("ty\": \"Shanghai\"}").build());
 
         ToolUseBlock result = accumulator.buildAllToolCalls().get(0);
         assertEquals("{\"city\": \"Shanghai\"}", result.getContent());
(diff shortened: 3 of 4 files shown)


agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
```diff
@@ -113,8 +117,16 @@ ToolUseBlock build() {
...
+                    // Parsing failed, keep previously merged args; log the raw string so a
+                    // malformed tool-call argument is diagnosable instead of silently lost
+                    LOG.warn(
+                            "Failed to parse accumulated tool call arguments as JSON for tool"
+                                    + " '{}'; falling back to previously merged args. Raw"
+                                    + " arguments: {}",
+                            name,
+                            rawContentStr,
+                            e);
                 }
```
guslegend0510, line 132: [P1] Please avoid logging parse failures from this incremental build() path. build() is invoked for every streamed tool-call fragment through fireReasoningChunk(), so normal incomplete JSON prefixes are expected here. This produces repeated WARNs and stack traces—potentially twice for ID-less fragments—and repeatedly logs the growing raw argument string. Besides causing near O(n²) log volume, raw tool arguments may contain credentials, PII, user documents, or source code. Please emit at most one sanitized warning at the finalization/interruption boundary and avoid logging raw arguments by default. Add regression tests asserting that a valid multi-fragment call emits no warning and a final malformed call emits one redacted warning.

guslegend0510 (changes requested):

agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
chaucerj: Good catch on all three points — fixed in 25866a8:

- **Per-fragment spam**: `build()` is now fully silent. It only records the parse-failure cause and clears it on a successful parse (so a prefix that later completes correctly never warns). `getAccumulatedToolCall`/hook paths stay quiet during streaming.
- **Single sanitized warning at finalization**: `buildAllToolCalls()` — the end-of-stream boundary used by `buildFinalMessage()` — emits at most one WARN per tool call: tool name, raw length, and cause only. No raw arguments (not even argument keys), no stack trace, and no exception message either, since Jackson/json-p messages can quote the offending token.
- **Regression tests** (`ToolCallsAccumulatorParseWarningTest`, via test-scoped logback-classic + a WARN-level `logback-test.xml`): a valid multi-fragment call (with an intermediate `getAccumulatedToolCall` build) asserts zero warnings; a truncated final call asserts exactly one warning that contains the tool name but neither the argument keys nor the secret value, with no throwable. Full agentscope-core suite: 2304 tests, 0 failures.

On the raw-string-on-ToolUseBlock discussion from the issue: agreed that's a larger API decision, so I've left it out of this PR as you suggested.

guslegend0510: The warning order is still incorrect. buildAllToolCalls() calls warnIfMalformedOnce() before build() updates parseFailure from the final raw content. As a result, the Issue #2841 reproduction emits no warning on the first finalization, while a previously incomplete call that later becomes valid can emit a stale warning. Please build/parse each tool call before checking the warning state, and add regression tests for both transitions. Also, please avoid publishing the module-wide logback-test.xml through the core test-jar.

chaucerj: Both points fixed in 2a83f82:

- **Warning order**: `buildAllToolCalls()` now runs `build()` first, so `parseFailure` reflects the *final* raw content when `warnIfMalformedOnce()` checks it. The issue #2841 reproduction (no mid-stream builds) warns on the first finalization, and a prefix that failed mid-stream but completed validly stays silent.
- **Regression tests for both transitions**: `validMultiFragmentEmitsNoWarning` now builds at the genuinely incomplete point (first fragment only) before completing the call; `truncatedWithoutIntermediateBuildsWarnsOnFirstFinalization` covers no-mid-stream-build truncation — exactly one warning on the first finalization, silent on repeated finalization. I verified both tests fail against the previous ordering (stale warning / no warning respectively) before landing the fix.
- **Test-jar**: the core `maven-jar-plugin` test-jar execution now excludes `logback-test.xml` (verified: not present in `agentscope-core-*-tests.jar`, 560 test classes still packaged).

Full agentscope-core suite: 2305 tests, 0 failures.

guslegend0510 (approved):
LGTM

agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
```diff
@@ -113,8 +122,13 @@ ToolUseBlock build() {
...
                     }
-                } catch (Exception ignored) {
-                    // Parsing failed, keep previously merged args
+                } catch (Exception e) {
+                    // Incomplete JSON prefixes are expected while streamed fragments are still
+                    // arriving, so build() stays silent here; buildAllToolCalls() reports the
+                    // failure once at finalization if the raw content never became valid JSON.
+                    // Only the exception name is kept: its message may quote raw arguments,
+                    // which can contain credentials or other sensitive data.
+                    parseFailure = e.getClass().getSimpleName();
```
Aias00, line 131: Dropping the exception message is the right call (Jackson messages quote the offending input, which here is raw tool arguments that may carry credentials or user documents). But keeping only the simple class name throws away the one content-free, high-signal detail: **where** parsing stopped.

Consider also capturing the location when the exception is a `JsonProcessingException`:

```java
parseFailure = e.getClass().getSimpleName();
if (e instanceof com.fasterxml.jackson.core.JsonProcessingException jpe
        && jpe.getLocation() != null) {
    parseFailure += " at line " + jpe.getLocation().getLineNr()
            + ", column " + jpe.getLocation().getColumnNr();
}
```

That tells an operator "the JSON broke at line 3, column 17" without logging a single argument value, which buys back most of the diagnosability that #2841 is asking for. As it stands, the only positional hint is `rawContent.length()`.

agentscope-core/src/test/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulatorParseWarningTest.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    @AfterEach
+    void detachAppender() {
+        logger.detachAppender(appender);
+    }
+
+    private List<ILoggingEvent> warnings() {
+        return appender.list.stream()
+                .filter(event -> event.getLevel() == Level.WARN)
```
Aias00, line 65: Nit: `detachAppender()` leaves the appender running — call `appender.stop()` after detaching so the instance is fully released (`ListAppender` is not thread-safe, and a started appender that is still attached anywhere keeps buffering events).

Aias00 (approved):
Good, tight observability fix. The important design choices are right:

- **Correct finalization boundary.** I traced the callers: `buildAllToolCalls()` is reached from `ReasoningContext.buildFinalMessage()` and the studio aggregator, while mid-stream callers go through `getAccumulatedToolCall(id)`. `ToolCallsAccumulator.buildAggregated()` has no production caller, so "build() is silent, finalization reports once" genuinely holds and streamed JSON prefixes stay quiet.
- **No stale warnings.** Clearing `parseFailure` on a successful parse means a prefix that failed mid-stream but completed validly stays silent — covered explicitly by the first test.
- **Sanitized by construction.** Logging only the exception class name, the raw length and the tool name, with no throwable, keeps credentials and user documents out of the logs. That is the right trade-off for a framework-level log line.
- The `logback-test.xml` exclusion from the published test-jar is a thoughtful touch, and `logback-classic` is test-scoped so it does not leak onto the harness test classpath (Maven drops test-scoped transitive deps of the test-jar).

**On CI:** the `build (ubuntu-latest)` failure is not from this change. It is `HarnessAgentSubagentStreamTest.call_localSubagent_returnsReplyWithoutStreaming` failing in *extension teardown* — `Failed to delete temp directory ... DirectoryNotEmptyException` — in a module this PR does not touch, and with 0 test failures and 1 teardown error across 844 tests. Windows was then cancelled by fail-fast. Please re-run the workflow.

Two non-blocking suggestions inline: capture the Jackson parse location (line/column) to restore most of the diagnosability without logging any argument value, and stop the `ListAppender` in `@AfterEach`. Approving.

agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
chaucerj: Good point — the location is content-free signal and Jackson exposes it directly. I will add `JsonProcessingException` line/column capture as a follow-up after this merges, rather than pushing a new commit now (it would dismiss your approval and re-run the whole matrix for a logging-only tweak). The `rawContent.length()` hint stays until then.

agentscope-core/src/test/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulatorParseWarningTest.java
chaucerj: Right, `stop()` after detach is the correct cleanup — will include it in the same follow-up as the location capture, so this PR can merge on the current approval.

chaucerj:
Thanks for the review and for merging main!

The ubuntu failure on the merge commit looks like a JUnit `@TempDir` teardown flake rather than a code regression:

- `HarnessAgentSubagentStreamTest.call_localSubagent_returnsReplyWithoutStreaming` — the test body itself **passed** (0.141s); the error is `Failed to close extension context` ← `IOException: Failed to delete temp directory /tmp/junit-…` ← `DirectoryNotEmptyException` during temp-dir cleanup.
- The windows job was **cancelled** (fail-fast), with no test failures of its own.
- I ran the full `agentscope-harness` suite locally on the merge commit: **842 tests, 0 failures**.
- The merge introduced no untested code: my half was green in the previous run (2a83f82), and upstream main's CI is green with the merged commits.

Could you re-run the failed jobs? Happy to dig deeper if it reproduces.

chaucerj:
Thanks @jujn for taking a look! I noticed the `TBD` label — happy to join the discussion.

For context, this PR implements the minimum behavior requested in #2841 (log instead of silently swallowing). If the concern is the broader design — e.g. whether malformed accumulated JSON should emit a framework event, log at `error` level, or fail the tool call instead of falling back to `input = {}` — I'm glad to extend or rework the approach. Just let me know which direction you'd like to take.

oss-maintainer (commented):
### Summary

Good, well-scoped fix for #2841 — replacing `catch (Exception ignored)` with a single sanitized warning at the finalization boundary is a real observability win, and keeping the raw arguments and exception message out of the log (they routinely carry credentials/PII) is the right call. The `parseFailure` clear-on-success plus `parseWarned` once-guard also avoids the mid-stream false-positive spam that a naive `build()`-level warn would produce, and the regression tests cover the tricky cases (prefix completes later, repeated per-fragment builds, credential-looking payload, never-built call). CI is green and the diff is reviewable end to end.

Four non-blocking points before this is mergeable from my side: the hardcoded `logback-classic` version bypasses the BOM, the new module-wide `logback-test.xml` changes logging config for every `agentscope-core` test, the warn is only reachable from one drain path (`buildAllToolCalls()`) so abort/cancel and `getAccumulatedToolCall()` paths can still swallow the defect, and the message is uncorrelatable when `name` is null/placeholder. None of them need to block a merge of the core behaviour change — the first two are build hygiene, the last two are follow-up friendly.

---

agentscope-core/pom.xml
```diff
@@ -203,12 +203,21 @@
             <artifactId>kubernetes-server-mock</artifactId>
             <scope>test</scope>
         </dependency>
+        <!-- SLF4J backend for tests that assert log output (e.g. accumulator parse warnings) -->
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>1.5.18</version>
```
oss-maintainer, line 210: This pins `logback-classic` with an inline version, while the rest of the build centralises versions in `agentscope-dependencies-bom` (e.g. `slf4j.version=2.0.17` lives there). A hardcoded version in one module drifts silently and, if another module later adds a different logback, produces a duplicate-binding conflict on the test classpath. Could you either add `logback.version` + a `dependencyManagement` entry in the BOM, or drop `<version>` if the parent already manages it?

While at it, is `logback-classic` already on `agentscope-core`'s test classpath transitively? If it is, this explicit pin (rather than a plain test dependency) becomes the only guard, and `mvn -pl agentscope-core dependency:tree -Dincludes=org.slf4j,ch.qos.logback` would be worth a run to confirm no second SLF4J binding sneaks in (several extension modules exclude `slf4j-simple`/`slf4j-log4j12`).

agentscope-core/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Keep test logging quiet: warnings and errors only, to stderr -->
+<configuration>
+    <appender name="STDERR" class="ch.qos.logback.core.ConsoleAppender">
+        <target>System.err</target>
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} %-5level %logger{36} - %msg%n</pattern>
+        </encoder>
+    </appender>
+    <root level="WARN">
```
oss-maintainer, line 10: `<root level="WARN">` plus a stderr appender is now the logging configuration for *every* test in `agentscope-core`, not just the new parse-warning tests — previously the module had no `logback-test.xml` at all. That is usually harmless, but it can change output/behaviour for tests that assert on stderr or that relied on the default configuration, and it hides INFO-level diagnostics when someone debugs a failing core test.

Two options: keep the file but narrow the change to what the test needs (e.g. leave `root` at the default and only set the level for `io.agentscope.core.agent.accumulator`), or call the module-wide effect out in the PR description so reviewers know it is deliberate.

agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
```diff
@@ -246,7 +282,19 @@ public ContentBlock buildAggregated() {
...
+        // Finalization boundary: the stream has ended, so malformed accumulated JSON is a
+        // real defect worth reporting — once per tool call, sanitized. build() runs first so
+        // parseFailure reflects the FINAL raw content: a prefix that failed mid-stream but
+        // completed validly must not warn, and a call that was never built mid-stream must
+        // still warn on this first finalization.
+        return builders.values().stream()
+                .map(
+                        builder -> {
+                            ToolUseBlock block = builder.build();
+                            builder.warnIfMalformedOnce();
```
oss-maintainer, line 294: Attaching the warning to `buildAllToolCalls()` covers the normal completion path, but `buildAllToolCalls()` is not the only way a builder is drained: `getAccumulatedToolCall(...)` (line 309, used by `ReActAgent.resolveToolCallId` and `LegacyHookDispatcher`) and any abort/cancel path call `build()` directly and never reach `warnIfMalformedOnce()`. In the scenario from #2841 the stream can end via error/cancellation, so the malformed payload would still be swallowed with no log — exactly the failure mode this PR sets out to fix.

Would it be more robust to warn from `build()` itself, gated on a `terminal` flag (or on `JsonUtils.isValidJsonObject(rawContentStr)` failing at the point where `content` is forced to `"{}"`), so the report is tied to the moment the value is actually discarded rather than to one particular drain path? Happy to see this handled in a follow-up commit — the current version is already a strict improvement over `catch (Exception ignored)`.

agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
```diff
@@ -139,6 +153,28 @@ ToolUseBlock build() {
...
+         * documents) and carries no stack trace.
+         */
+        private void warnIfMalformedOnce() {
+            if (parseFailure != null && !parseWarned) {
+                parseWarned = true;
+                LOG.warn(
+                        "Tool call '{}' arguments are not valid JSON after the stream ended"
+                                + " (raw length: {}); using partially accumulated arguments"
+                                + " instead. Cause: {}",
+                        name,
```
oss-maintainer, line 172: Good call keeping the raw payload and the exception message out of the log — tool arguments routinely carry API keys and user documents. One gap for correlation though: `name` is null or a placeholder (`__fragment__` / `__pending__` / any `__*`) whenever the provider streamed fragments without a resolved tool name, and then this line reads `Tool call 'null' ...`, which is not traceable to a call.

Suggestion: include `toolId` (already resolved in `build()`) alongside the name, e.g. `"Tool call '{}' (id '{}') arguments are not valid JSON ..."`, and fall back to the builder key when `isPlaceholder(name)` is true. That gives operators something to match against the surrounding trace/span ids.

PR #2874, opened 2026-08-28
fix(agui): emit frontend tool args from fragment deltas

### LINKED ISSUE #2873: [Bug]: AG-UI frontend tools miss TOOL_CALL_ARGS when streamed name is __fragment__
**Describe the bug**
When `emitToolCallArgs` is disabled, frontend tools still need `TOOL_CALL_ARGS` so CopilotKit / AG-UI clients can invoke them. `ToolCallEventConverter` currently matches frontend tools with `delta.getToolCallName()`. Streaming argument chunks use the placeholder name `__fragment__`, so the frontend-tool check fails and no args are emitted.

This is different from #498, which was about accumulating streamed args in general and is already closed.

**To Reproduce**
1. Register a frontend tool via `RunAgentInput.tools` (for example `lookup`).
2. Set `AguiAdapterConfig.emitToolCallArgs(false)` so backend tool args are hidden.
3. Stream a tool call as:
   - `ToolCallStartEvent(id=tool-1, name=lookup)`
   - `ToolCallDeltaEvent(id=tool-1, name=__fragment__, delta={q)`
   - `ToolCallDeltaEvent(id=tool-1, name=__fragment__, delta=:agent})`
   - `ToolCallEndEvent(id=tool-1, name=lookup)`
4. Observe that `TOOL_CALL_ARGS` events are missing, so the frontend cannot call the tool.

**Expected behavior**
Frontend tools should emit `TOOL_CALL_ARGS` even when the streamed delta name is `__fragment__`. The real tool name is already known from `ToolCallStart` and should be resolved by `toolCallId`. Backend tools should still hide args when `emitToolCallArgs=false`.

**Error messages**
No exception. The AG-UI stream only contains `TOOL_CALL_START` / `TOOL_CALL_END`, so the client cannot reconstruct tool parameters.

**Environment**
- AgentScope-Java Version: main
- Java Version: 17+
- OS: Windows

**Additional context**
Fix: record the real tool name on `ToolCallStart` in `AguiStreamContext`, then decide whether to emit args with `isFrontendToolCall(toolCallId)` instead of the placeholder delta name.

dragondyt:
## Related issues
- Fixes #2873

## Summary
- Record the real tool name on `ToolCallStart` and match frontend tools by `toolCallId`.
- Emit `TOOL_CALL_ARGS` for frontend tools even when streamed deltas use `__fragment__`.
- Keep backend tool args hidden when `emitToolCallArgs` is false.
- Add unit coverage for frontend and backend fragment deltas.

## Breaking Changes
None.

### DIFF
--- a/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/strategy/AguiStreamContext.java
+++ b/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/strategy/AguiStreamContext.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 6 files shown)


dragondyt:
CI失败与本次提交无关

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/strategy/AguiStreamContext.java
```diff
@@ -187,6 +192,9 @@ public void startToolCall(String toolCallId, String toolCallName) {
             return;
         }
         if (startedToolCalls.add(toolCallId)) {
+            if (toolCallName != null && !toolCallName.isBlank()) {
+                startedToolCallNames.put(toolCallId, toolCallName);
```
Aias00, line 218: Nit: `startedToolCallNames` is never pruned when the tool call ends, so it grows for the lifetime of the context (an agent loop can issue hundreds of tool calls in one run). `startedToolCalls` has the same behaviour today, so this is consistent — but since this map is only consulted between `ToolCallStart` and `ToolCallEnd`, consider removing the entry in `endToolCall()` to keep the context bounded.

Aias00 (approved):
Matches #2873 exactly: the real tool name is captured on `ToolCallStart` and the frontend-tool decision is made from `toolCallId` instead of the `__fragment__` placeholder, so frontend tools keep receiving `TOOL_CALL_ARGS` while backend tools stay hidden under `emitToolCallArgs=false`.

Deliberate details I checked and like:
- The raw (pre-normalisation) name is stored, so matching is against what the client actually declared in `RunAgentInput.tools`.
- `Collectors.toSet()` rather than `toUnmodifiableSet()` — the latter would NPE on a null `AguiTool.name`.
- `isFrontendToolCall` degrades to `false` when the start event was never seen, so the previous opt-in behaviour is preserved.

Tests cover both directions (frontend emits, backend does not) with the exact placeholder-name stream from the issue.

Note on CI: the `build (windows-latest)` failure is unrelated — it is `SessionEventLogCrossProcessTest.subscribeAfterCursorSkipsEarlierEvents` in `service-common` failing with a Reactor `OverflowException` from interval backpressure, a known Windows timing flake.

One non-blocking nit inline about pruning the new map. Approving.

PR #2877, opened 2026-08-28
feat(harness): report MCP server registration results

guslegend0510:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

MCP server registration is intentionally best-effort, but callers currently only receive log output when a configured server cannot initialize or list its tools. This leaves host services without a machine-readable way to identify and retire unhealthy MCP configurations.

This PR:

- adds an optional `McpServerRegistrationListener` and immutable result payload with `SUCCESS`, `FAILED`, and `SKIPPED` terminal states;
- exposes the listener through `HarnessAgent.Builder` for both file-based and programmatic tools configuration;
- preserves the existing two-argument registration API and best-effort behavior;
- isolates listener failures so they cannot interrupt later MCP registrations;
- closes a created MCP client wrapper if Toolkit registration fails, preserving close failures as suppressed exceptions;
- adds coverage for invalid transport configuration, initialization/list-tools failures, listener isolation, cleanup, and HarnessAgent integration.

Related to #2875.

## Validation

- `git diff --check`
- `mvn -pl agentscope-harness -am spotless:check -DskipTests`
- `mvn -pl agentscope-harness -am -Dtest=McpClientManagerTest,McpServerRegistrarTest,HarnessAgentToolsConfigTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -pl agentscope-harness -am test` — 3,160 tests, 0 failures, 0 errors

## Checklist

- [x] All affected module tests are passing
- [x] No documentation update is required for this API-only change


agentscope-harness/src/main/java/io/agentscope/harness/agent/tools/McpServerRegistrar.java
```diff
@@ -65,25 +87,65 @@ public static void register(Toolkit toolkit, Map<String, McpServerConfig> server
...
-            reg.enableTools(enableTools);
+        List<String> enableTools;
+        try {
+            Toolkit.ToolRegistration reg = toolkit.registration().mcpClient(wrapper);
+            enableTools = cfg.getEnableTools();
+            if (enableTools != null && !enableTools.isEmpty()) {
+                reg.enableTools(enableTools);
+            }
+            reg.apply();
+        } catch (RuntimeException | Error failure) {
```
Aias00, line 124: Nit: catching `Error` here is broader than most code in this repo. The intent (release the wrapper, then rethrow unchanged) is fine, but please narrow this to `RuntimeException` or add a short comment explaining why `Error` (e.g. `StackOverflowError` from a deep transport init) must also release the client. Otherwise a future reader may assume Errors are swallowed.

agentscope-core/src/test/java/io/agentscope/core/tool/McpClientManagerTest.java
```diff
@@ -413,4 +416,39 @@ void testRegisterMcpClient_PreservesOutputSchemaInRegisteredTool() {
         assertEquals("test-client", registeredClientName[0]);
         assertEquals(toolPresetParams, registeredPresetParams[0]);
     }
+
+    @Test
+    void testRegisterMcpClient_PropagatesInitializationFailure() {
```
Aias00, line 421: Scope nit: these two tests only lock in behaviour that `McpClientManager.registerMcpClient` already has (`initialize().then(defer(listTools))` propagates errors) — no production code in `agentscope-core` changes in this PR. They are harmless, but they belong in a separate commit/PR so the harness-side change stays reviewable on its own.

Aias00 (approved):
The listener plumbing is clean and backwards compatible: the existing two-arg `register(toolkit, servers)` overload is preserved as a no-listener path, results keep configuration order, listener exceptions are isolated per entry, and the new `closeAfterFailedRegistration` closes a real resource leak when `apply()` fails after the client was built.

Coverage is solid — failure ordering, per-transport validation errors, success path, suppressed close failures, and listener-throws-are-contained are all exercised.

Two non-blocking nits inline (scope of `Error` in the catch, and the unrelated `agentscope-core` test additions). Approving.

jujn (approved):

PR #2878, opened 2026-08-28
fix(model): move cache control to content blocks

### LINKED ISSUE #1363: [Bug] DashScope native API: cache_control placed at message level instead of content block level

charlotteForever:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #1363

Supersedes #1377

DashScope native and OpenAI-compatible requests currently serialize explicit
`cache_control` markers at message level. The provider wire format expects the
marker on a content block instead.

This change:

- places automatic and metadata-driven markers on the final suitable content
  block of each selected message;
- converts marked string content to a text content-part array;
- preserves existing multimodal content and changes only its final suitable
  block;
- migrates legacy message-level markers and clears the invalid placement;
- preserves the empty-map `CACHE_CONTROL=false` sentinel so no marker is
  serialized and repeated automatic passes cannot add one; and
- carries explicit cache metadata through the multi-agent merge paths. When
  multiple source messages become one output block, the last explicit Boolean
  directive applies to that merged boundary.

The existing automatic selection remains unchanged: all system messages plus
the final message. Dynamic Hook caching policy and the provider's four-marker
limit are intentionally out of scope.

Thanks to @flystar32 for the original implementation and investigation in
#1377. This PR refreshes the fix for the current provider-module layout and
also handles legacy-marker migration and the false sentinel.

### Testing

- TDD regression cycle for `DashScopeCacheControlTest` and
  `OpenAICacheControlTest` (expected failures recorded before implementation)
- `mvn -B -T1 -pl agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope,agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai -am clean verify`

The verification covers local formatting, compilation, Javadocs, unit tests,
and final JSON serialization. No live provider E2E call was performed.

## Checklist

- [x] Relevant module and dependency tests pass with scoped `clean verify`
- [x] Documentation impact was assessed; no user-facing documentation change is required
- [ ] Code is ready for review (kept unchecked while this PR is a draft)

### DIFF
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/formatter/DashScopeChatFormatter.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/formatter/DashScopeChatFormatter.java
@@ -21,6 +21,7 @@ import io.agentscope.core.model.ChatResponse;
 import io.agentscope.core.model.GenerateOptions;
 import io.agentscope.core.model.ToolChoice;
 import io.agentscope.core.model.ToolSchema;
+import io.agentscope.extensions.model.dashscope.dto.DashScopeContentPart;
 import io.agentscope.extensions.model.dashscope.dto.DashScopeInput;
 import io.agentscope.extensions.model.dashscope.dto.DashScopeMessage;
 import io.agentscope.extensions.model.dashscope.dto.DashScopeParameters;
@@ -46,9 +47,9 @@ public class DashScopeChatFormatter
     private static final Map<String, String> EPHEMERAL_CACHE_CONTROL = Map.of("type", "ephemeral");
 
     /**
-     * Sentinel for an explicit "no cache" intent. An empty map is serialized away (via
+     * Sentinel for an explicit "no marker" intent. An empty map is serialized away (via
      * {@code @JsonInclude(NON_EMPTY)} on {@link DashScopeMessage#cacheControl}), so the upstream
-     * API receives no {@code cache_control} field and therefore performs no caching.
+     * API receives no explicit {@code cache_control} marker.
      */
     private static final Map<String, String> NO_CACHE_CONTROL = Collections.emptyMap();
 
@@ -183,25 +184,69 @@ public class DashScopeChatFormatter
     /**
      * Apply cache control to DashScope messages.
      *
-     * <p>Adds <code>cache_control: {"type": "ephemeral"}</code> to all system messages and the last
-     * message in the list. Messages that already carry a cache_control value (including the empty
-     * "no cache" sentinel) are left untouched.
+     * <p>Adds <code>cache_control: {"type": "ephemeral"}</code> to the last content part of all
+     * system messages and the last message in the list. Legacy message-level values are migrated to
+     * the last content part. Messages explicitly excluded from caching are left untouched.
      *
      * @param messages the list of formatted DashScope messages
      */
     public void applyCacheControl(List<DashScopeMessage> messages) {
+        applyCacheControlToMessages(messages);
+    }
+
+    static void applyCacheControlToMessages(List<DashScopeMessage> messages) {
         if (messages == null || messages.isEmpty()) {
             return;
         }
         for (DashScopeMessage msg : messages) {
+            migrateLegacyCacheControl(msg);
+        }
+        for (DashScopeMessage msg : messages) {
             if ("system".equals(msg.getRole()) && shouldAutoCache(msg)) {
-                msg.setCacheControl(EPHEMERAL_CACHE_CONTROL);
+                setCacheControlOnContent(msg, EPHEMERAL_CACHE_CONTROL);
             }
         }
         DashScopeMessage lastMsg = messages.get(messages.size() - 1);
         if (shouldAutoCache(lastMsg)) {
-            lastMsg.setCacheControl(EPHEMERAL_CACHE_CONTROL);
+            setCacheControlOnContent(lastMsg, EPHEMERAL_CACHE_CONTROL);
         }
+    }
+
+    static void setCacheControlOnContent(
+            DashScopeMessage message, Map<String, String> cacheControl) {
+        DashScopeContentPart lastPart = getOrCreateLastContentPart(message);
+        if (lastPart != null && lastPart.getCacheControl() == null) {
+            lastPart.setCacheControl(cacheControl);
+        }
+        message.setCacheControl(null);
+    }
+
+    private static void migrateLegacyCacheControl(DashScopeMessage message) {
+        Map<String, String> legacyCacheControl = message.getCacheControl();
+        if (legacyCacheControl != null && !legacyCacheControl.isEmpty()) {
+            setCacheControlOnContent(message, legacyCacheControl);
+        }
+    }
+
+    private static DashScopeContentPart getOrCreateLastContentPart(DashScopeMessage message) {
+        Object content = message.getContent();
+        if (content instanceof String text) {
+            DashScopeContentPart textPart = DashScopeContentPart.text(text);
+            message.setContent(List.of(textPart));
+            return textPart;
+        }
+        return findLastContentPart(content);
+    }
+
+    private static DashScopeContentPart findLastContentPart(Object content) {
+        if (content instanceof List<?> parts) {
+            for (int i = parts.size() - 1; i >= 0; i--) {
+                if (parts.get(i) instanceof DashScopeContentPart part) {
+                    return part;
+                }
+            }
+        }
+        return null;
     }
 
     /**
@@ -214,9 +259,9 @@ public class DashScopeChatFormatter
     }
 
     /**
-     * Get the "no cache" sentinel constant.
+     * Get the "no marker" sentinel constant.
      *
-     * @return unmodifiable empty map representing an explicit "no cache" intent
+     * @return unmodifiable empty map representing an explicit "no marker" intent
      */
     static Map<String, String> getNoCacheControl() {
         return NO_CACHE_CONTROL;
@@ -225,14 +270,18 @@ public class DashScopeChatFormatter
     /**
      * Whether the automatic cache-control strategy should mark a message as ephemeral.
      *
-     * <p>Returns {@code true} only when the message has no cache_control value at all. Any non-null
-     * value — an explicit {@code {"type": "ephemeral"}}, a custom value, or the empty "no cache"
-     * sentinel — is left unchanged.
+     * <p>Returns {@code true} only when neither the legacy message-level state nor the last content
+     * part carries a cache-control value. The empty message-level "no marker" sentinel is therefore
+     * preserved across repeated automatic-strategy passes.
      *
      * @param message the message to inspect
      * @return {@code true} if the message should be auto-cached, {@code false} otherwise
      */
     static boolean shouldAutoCache(DashScopeMessage message) {
-        return message.getCacheControl() == null;
+        if (message.getCacheControl() != null) {
+            return false;
+        }
+        DashScopeContentPart lastPart = findLastContentPart(message.getContent());
+        return lastPart == null || lastPart.getCacheControl() == null;
     }
 }
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/formatter/OpenAIBaseFormatter.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/formatter/OpenAIBaseFormatter.java
@@ -20,6 +20,7 @@ import io.agentscope.core.model.ChatResponse;
 import io.agentscope.core.model.GenerateOptions;
 import io.agentscope.core.model.ToolChoice;
 import io.agentscope.core.model.ToolSchema;
+import io.agentscope.extensions.model.openai.dto.OpenAIContentPart;
 import io.agentscope.extensions.model.openai.dto.OpenAIMessage;
 import io.agentscope.extensions.model.openai.dto.OpenAIRequest;
 import io.agentscope.extensions.model.openai.dto.OpenAIResponse;
@@ -46,9 +47,9 @@ public abstract class OpenAIBaseFormatter
     private static final Map<String, String> EPHEMERAL_CACHE_CONTROL = Map.of("type", "ephemeral");
 
     /**
-     * Sentinel for an explicit "no cache" intent. An empty map is serialized away (via
+     * Sentinel for an explicit "no marker" intent. An empty map is serialized away (via
      * {@code @JsonInclude(NON_EMPTY)} on {@link OpenAIMessage#cacheControl}), so the upstream API
-     * receives no {@code cache_control} field and therefore performs no caching.
+     * receives no explicit {@code cache_control} marker.
      */
     private static final Map<String, String> NO_CACHE_CONTROL = Collections.emptyMap();
 
@@ -180,9 +181,9 @@ public abstract class OpenAIBaseFormatter
     /**
      * Apply cache control to OpenAI messages.
      *
-     * <p>Adds <code>cache_control: {"type": "ephemeral"}</code> to all system messages and the last
-     * message in the list. Messages that already carry a cache_control value (including the empty
-     * "no cache" sentinel) are left untouched.
+     * <p>Adds <code>cache_control: {"type": "ephemeral"}</code> to the last content part of all
+     * system messages and the last message in the list. Legacy message-level values are migrated to
+     * the last content part. Messages explicitly excluded from caching are left untouched.
      *
      * @param messages the list of formatted OpenAI messages
      */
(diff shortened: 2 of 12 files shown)


oss-maintainer (commented):
﻿### Summary

Moves `cache_control` from the message level down onto the last content part for both the DashScope and OpenAI extensions, which matches what the upstream APIs actually consume, and keeps a migration path for callers that were setting the legacy message-level field. The design is careful in the places that matter: the empty-map "no marker" sentinel stays meaningful across repeated passes, `lastPart.getCacheControl() == null` protects a caller's own placement, the migration pass runs before the auto strategy, and the multi-agent formatters now propagate `CACHE_CONTROL` metadata for the system message, assistant tool calls, tool results and merged conversation groups — previously those paths dropped it. Tests were extended on both providers rather than only one. CI is green on Linux and Windows, CLA signed.

Verdict: **comment** — no design objection, but the migration has two paths whose behaviour I could not confirm from the diff, and one invariant change worth a look.

### Findings

- **[Warning]** `OpenAIBaseFormatter.java:213` (and `DashScopeChatFormatter.java:221`) — the legacy message-level value is cleared even when no content part was found to receive it, so a marker the caller explicitly set can disappear without a trace.
- **[Warning]** `OpenAIBaseFormatter.java:227` (and `DashScopeChatFormatter.java:235`) — the `String` → single-part conversion now applies to system messages and the *last message in the list*, which in a ReAct turn is usually a `tool` result or an `assistant` with `tool_calls` and empty text; that changes the wire shape for roles where `content` is typed as string-or-null, and can emit `{"type":"text","text":""}`. Worth confirming against the endpoints (native DashScope text-generation vs multimodal vs OpenAI-compatible) rather than only the DTO tests.
- **[Warning]** `DashScopeChatFormatter.java:284` — `shouldAutoCache` inspects only the last part, so a message carrying an explicit marker on an earlier part is auto-marked again; the old per-message field made one-marker-per-message implicit, and providers bound the number of breakpoints per request. Same at `OpenAIBaseFormatter.java:276`.
- **[Info]** The five helpers are a near-verbatim third copy of this policy in the repo (the other two being the OpenAI formatter and `AnthropicBaseFormatter`), including an identical `applyMergedCacheControlMetadata` in both multi-agent formatters. Fine to defer; noted because "last marker wins" and sentinel semantics are what drift.

### Suggestions

The `lastPart == null` case is the one I would change before merge, since it is a silent-loss path with no test and no log:

```java
OpenAIContentPart lastPart = getOrCreateLastContentPart(message);
if (lastPart == null) {
    log.debug("No content part available for cache_control on role '{}'", message.getRole());
    return; // leave the legacy value in place instead of dropping it
}
if (lastPart.getCacheControl() == null) {
    lastPart.setCacheControl(cacheControl);
}
message.setCacheControl(null);
```

### Cross-repo Note

`MessageMetadataKeys.CACHE_CONTROL` in `agentscope-core` is a per-message `Boolean` contract, and this PR changes what a provider does with it (placement moves to a content part, `shouldAutoCache` becomes per-last-part). The core javadoc itself is still accurate, so no change is required there — but if more providers adopt part-level placement, the "one explicit marker per message blocks the automatic strategy" expectation is worth restating somewhere provider-neutral, because that is the invariant that shifted here.

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/formatter/OpenAIBaseFormatter.java
```diff
@@ -180,27 +181,66 @@ public OpenAIRequest buildRequest(
...
+            setCacheControlOnContent(lastMsg, EPHEMERAL_CACHE_CONTROL);
+        }
+    }
+
+    static void setCacheControlOnContent(OpenAIMessage message, Map<String, String> cacheControl) {
+        OpenAIContentPart lastPart = getOrCreateLastContentPart(message);
+        if (lastPart != null && lastPart.getCacheControl() == null) {
+            lastPart.setCacheControl(cacheControl);
+        }
+        message.setCacheControl(null);
```
oss-maintainer, line 213: `message.setCacheControl(null)` runs even when `getOrCreateLastContentPart` returned `null`, so the migration can *destroy* a caller's explicit directive instead of moving it. Concretely: a message whose `content` is `null`, an empty `List`, or a `List` of elements that are not `OpenAIContentPart` reaches `migrateLegacyCacheControl` with a non-empty legacy map, sets nothing on the part side, and then clears the message-level value — the request goes out with no `cache_control` at all, silently. The same applies when `applyCacheControlFromMetadata` marks a message that has no usable part.

Suggest only clearing after a successful move, and logging the case that cannot be moved:

```java
OpenAIContentPart lastPart = getOrCreateLastContentPart(message);
if (lastPart == null) {
    log.debug("Skipping cache_control migration for role '{}': no text content part available", message.getRole());
    return; // keep the legacy value rather than dropping it
}
if (lastPart.getCacheControl() == null) {
    lastPart.setCacheControl(cacheControl);
}
message.setCacheControl(null);
```

`DashScopeChatFormatter.setCacheControlOnContent` (line 221) has the same shape and needs the same fix. If the intent really is "an unmigratable marker is dropped", a test asserting it would be better than silence — right now nothing covers the `lastPart == null` path.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/formatter/OpenAIBaseFormatter.java
```diff
@@ -180,27 +181,66 @@ public OpenAIRequest buildRequest(
...
+        if (legacyCacheControl != null && !legacyCacheControl.isEmpty()) {
+            setCacheControlOnContent(message, legacyCacheControl);
         }
     }
 
+    private static OpenAIContentPart getOrCreateLastContentPart(OpenAIMessage message) {
+        Object content = message.getContent();
+        if (content instanceof String text) {
+            OpenAIContentPart textPart = OpenAIContentPart.text(text);
+            message.setContent(List.of(textPart));
```
oss-maintainer, line 227: Widening the payload shape: this converts `"content": "..."` into `"content": [{"type":"text",...}]` for *every* message the auto strategy marks, which is system messages plus the last message in the list. In an agent loop the last message is very often `role: "tool"`, or `role: "assistant"` with `tool_calls` and empty text (`convertToolMessage` and the `builder.content("")` paths in `OpenAIMessageConverter`), so those requests now carry an array content part — and when the text was empty, a part with `"text": ""`. The OpenAI chat-completions schema types `tool` message content as string-or-null, so an array there is at best unspecified — do you know whether the endpoints you validated accept it, or is this reachable only when `cache_control` metadata is set on a tool result?

The DashScope side (same conversion at `DashScopeChatFormatter.java:235`) has an extra dimension: `DashScopeHttpClient` targets `/api/v1/services/aigc/text-generation/generation` for non-multimodal models and `/api/v1/services/aigc/multimodal-generation/generation` for multimodal ones, and it is the native protocol here rather than the OpenAI-compatible one. Could you confirm which endpoint(s) were exercised against the live API for `cache_control` inside an array `content`? If only the compatible-mode path was verified, marking text-generation requests in this form is the risky case.

Options worth considering: only migrate when the content is already in parts form; skip the array conversion for roles whose `content` is specified as string-or-null; or skip when `text.isEmpty()`. At minimum it would be good to state on `applyCacheControl` that marking the last message changes its `content` representation, since that is the observable contract change here, and to add a `role: "tool"` case to `OpenAICacheControlTest` — the current tests cover system/user text and multimodal parts, not the tool shape that dominates a ReAct turn.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/formatter/DashScopeChatFormatter.java
```diff
@@ -225,14 +270,18 @@ static Map<String, String> getNoCacheControl() {
...
      *
      * @param message the message to inspect
      * @return {@code true} if the message should be auto-cached, {@code false} otherwise
      */
     static boolean shouldAutoCache(DashScopeMessage message) {
-        return message.getCacheControl() == null;
+        if (message.getCacheControl() != null) {
+            return false;
+        }
+        DashScopeContentPart lastPart = findLastContentPart(message.getContent());
```
oss-maintainer, line 284: `shouldAutoCache` inspects only the *last* content part, so a multimodal message that already carries `cache_control` on an earlier part still counts as "unmarked" and receives a second marker here. The legacy per-message field made "one marker per message" the effective invariant; with markers now living on parts, this reads as "one marker per message *at most, on the last part*", which is not the same guarantee and can push a request past the provider's breakpoint budget for a single message. Since the surrounding code goes out of its way to respect explicit caller intent (the empty-map sentinel, the `lastPart.getCacheControl() == null` guard), scanning all parts looks like the consistent behaviour:

```java
private static boolean hasAnyCacheControl(List<DashScopeContentPart> parts) { ... }
```

Same in `OpenAIBaseFormatter.shouldAutoCache` (line 276).

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/formatter/DashScopeChatFormatter.java
```diff
@@ -183,27 +184,71 @@ public DashScopeRequest buildRequest(
...
             }
         }
         DashScopeMessage lastMsg = messages.get(messages.size() - 1);
         if (shouldAutoCache(lastMsg)) {
-            lastMsg.setCacheControl(EPHEMERAL_CACHE_CONTROL);
+            setCacheControlOnContent(lastMsg, EPHEMERAL_CACHE_CONTROL);
+        }
+    }
+
+    static void setCacheControlOnContent(
```
oss-maintainer, line 215: The five helpers here (`setCacheControlOnContent`, `migrateLegacyCacheControl`, `getOrCreateLastContentPart`, `findLastContentPart`, `shouldAutoCache`) plus the two constants are reproduced near-verbatim in `OpenAIBaseFormatter` and are the third copy of this policy in the repo (`AnthropicBaseFormatter.applyCacheControl` is the first). The subtle parts — sentinel handling, "last marker wins", migrate-then-clear — are exactly what drifts between copies, and `applyMergedCacheControlMetadata` is already identical in both multi-agent formatters.

Understood that `DashScopeContentPart` and `OpenAIContentPart` share no supertype, so this is not a mechanical extraction, and a core interface change cascades widely. A follow-up that gives both DTOs a small common contract (e.g. `CacheControlAware { Map<String,String> getCacheControl(); void setCacheControl(Map<String,String>); }` implemented in the extensions module, not core) would let the policy live once. Flagging it as a maintenance note, not something to solve in this PR.

oss-maintainer (commented):
### Summary
Re-review of `6c9255593`: moving `cache_control` from the message level to the last content part in both the DashScope and OpenAI extensions is the right direction, the legacy migration + String-content wrapping are covered by the extended tests, and the sentinel semantics rename ("no cache" -> "no marker") is honestly documented.

Two things to fix or call out:

`[Warning]` `DashScopeChatFormatter.java` (`setCacheControlOnContent`) — `message.setCacheControl(null)` runs unconditionally even when `getOrCreateLastContentPart` returns `null` (empty content list, or a list whose tail elements are not `DashScopeContentPart`). In that case the cache marker (or a legacy non-empty message-level value being migrated) is **silently dropped**. Same path applies via `migrateLegacyCacheControl`. Recommend only clearing the message-level field when the part-level set actually succeeded, and adding a test for the null-last-part edge. (Applies to the mirrored OpenAIBaseFormatter code too.)

`[Info]` wire compatibility — serialized requests no longer carry a message-level `cache_control` field at all. Any consumer/proxy that reads the message-level field (or an API tier that only honors it) will see the marker disappear; please confirm the target gateway/API version accepts part-level `cache_control` exclusively, and mention it in release notes if not already there.

Fix the drop-on-null path and this looks good to re-review.

---

jujn (approved):

PR #2885, opened 2026-08-28
fix(sandbox): handle SIGTERM in docker keep-alive to avoid 30s stop delay

### LINKED ISSUE #2884: [Bug]: Docker sandbox docker stop waits full 30s due to SIGTERM-ignoring entrypoint
**Describe the bug**

When using Docker sandbox mode, every agent turn incurs a ~30 second delay between the agent completing its work (last event emitted) and the `agent.call()` Flux emitting its terminal signal (onComplete). This blocks the SSE stream completion, making the user wait ~30 seconds after the agent's final response before the conversation turn is marked as finished.

**Root cause**: The container entrypoint command `sh -c "while :; do sleep 3600; done"` does not handle SIGTERM. When `DockerSandbox.shutdown()` runs `docker stop --time=30`, Docker sends SIGTERM to PID 1 (`sh`), but `sh` is blocked waiting for its child process `sleep 3600` and does not exit. Docker waits the full 30 seconds before sending SIGKILL to force-kill the container.

The call chain:
```
Flux.using() cleanup (HarnessAgent.java:924)
  -> SandboxLifecycleMiddleware.releaseForCall()
    -> SandboxManager.release()
      -> sandbox.shutdown()
        -> docker stop --time=30 <containerId>   <- blocks ~30s
        -> docker rm --force <containerId>
```

This happens on **every turn**, regardless of conversation length, because `SandboxManager.release()` always calls `stop()` + `shutdown()` after each agent call.

**To Reproduce**

1. Configure an agent with Docker sandbox mode (`sandbox.type=docker`, any isolation scope).
2. Disable snapshot (`sandbox.snapshot.type=noop` or default).
3. Send any message to the agent (even a simple "hello").
4. Measure the wall-clock time between the last SSE content chunk and the SSE stream completion - it is consistently ~30 seconds regardless of conversation length or complexity.

Verified by disabling compaction (`compactionEnabled=false`) - the delay persists unchanged, confirming it is not caused by context compression.

**Expected behavior**

`docker stop` should complete in <1 second. The container entrypoint should properly handle SIGTERM so the container exits gracefully upon receiving the signal.

Suggested fix - change the entrypoint in `DockerSandbox.java` (line ~525) from:
```java
cmd.add("sh");
cmd.add("-c");
cmd.add("while :; do sleep 3600; done");
```
to one of:
```java
// Option A: trap SIGTERM
cmd.add("sh");
cmd.add("-c");
cmd.add("trap 'exit 0' TERM; while :; do sleep 3600 & wait $!; done");

// Option B: tail (naturally responds to SIGTERM)
cmd.add("tail");
cmd.add("-f");
cmd.add("/dev/null");

// Option C: exec sleep (sleep becomes PID 1, receives signals directly)
cmd.add("sh");
cmd.add("-c");
cmd.add("exec sleep infinity");
```

Alternatively, add `--init` to the `docker run` command to use tini as PID 1, which properly forwards signals to child processes.

**Error messages**

No error - this is a silent performance issue. The turn eventually completes, but with an unnecessary ~30 second delay on every agent call.

- Java Version: 21
- OS: Linux (Docker), also reproduced on Windows with Docker Desktop
- Docker: `docker stop --time=30` behavior is standard across Docker versions

**Additional context**

The affected code locations:
- `DockerSandbox.java:71` - `CONTAINER_STOP_TIMEOUT_SECONDS = 30`
- `DockerSandbox.java:97-130` - `shutdown()` method calling `docker stop --time=30`
- `DockerSandbox.java:525-528` - container entrypoint `while :; do sleep 3600; done`
- `SandboxManager.java:140-160` - `release()` calling `stop()` + `shutdown()` synchronously
- `HarnessAgent.java:924` - `Flux.using()` with synchronous cleanup

This issue affects all Docker sandbox users regardless of isolation scope (SESSION/USER/AGENT/GLOBAL), because `SandboxManager.release()` always calls `shutdown()` after each turn.

The ~30 second delay is easily observable by any SSE client: measure the time between the last content chunk and the stream completion event.

larry-zy:
## Problem

In Docker sandbox mode, every agent turn incurs a ~30s delay between the agent's last emitted event and the `agent.call()` Flux terminal signal (onComplete), blocking SSE stream completion.

**Root cause:** the container keep-alive entrypoint `sh -c "while :; do sleep 3600; done"` leaves PID 1 (`sh`) blocked in a foreground `sleep` that never observes SIGTERM. In a PID namespace the kernel drops signals to PID 1 that have no installed handler, so when `SandboxManager.release()` → `sandbox.shutdown()` runs `docker stop --time=30` (synchronously, on the `Flux.using()` teardown path), Docker waits the full 30s grace period before escalating to SIGKILL. This happens on **every turn** regardless of conversation length. Fixes #2884.

## Fix

Change the entrypoint to install an explicit SIGTERM trap and wait on a backgrounded sleep:

```
sh -c "trap 'exit 0' TERM; while :; do sleep 3600 & wait $!; done"
```

- `trap 'exit 0' TERM` registers a handler on PID 1, so the kernel delivers SIGTERM instead of dropping it.
- `sleep 3600 &` + `wait $!` makes `sh` block on an **interruptible** `wait` (EINTR). SIGTERM interrupts the wait, the trap runs `exit 0`, and `docker stop` returns in <1s. A foreground `sleep` would defer the trap until the command returns.

### Why not `exec sleep infinity`

`sleep infinity` is a GNU coreutils extension that BusyBox/Alpine `sleep` rejects, so it would break container startup on minimal images. Bare `sleep` as PID 1 also has no SIGTERM handler and would still rely on `--init` to be reliable. The trap approach uses only `sh` + integer `sleep` — the exact same runtime primitives as the old entrypoint — so it stays portable and needs no extra `docker run` flag.

## Verification

Measured real `docker stop --time=30` with `sh -c` as PID 1 across images:

| image | shell | new (trap+wait) | old (foreground) |
|---|---|---|---|
| alpine:latest | BusyBox ash | 0.33s, exit 0 | 30.21s, exit 137 |
| ubuntu:22.04 | dash | 0.14s, exit 0 | 30.16s, exit 137 |
| busybox:latest | BusyBox ash | 0.12s, exit 0 | — |
| debian:bookworm-slim | dash | 0.18s, exit 0 | — |

Old entrypoint consistently hits the 30s grace period then SIGKILL (exit 137); new entrypoint exits cleanly in <0.35s.

## Tests

Adds `DockerSandboxRunCommandTest` asserting the generated `docker run` entrypoint installs the SIGTERM trap, waits on a backgrounded sleep, does not revert to the foreground sleep loop, and avoids the non-portable `sleep infinity`.

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
@@ -522,10 +522,14 @@ public class DockerSandbox extends AbstractBaseSandbox {
         }
 
         cmd.add(dockerState.getImage());
-        // Keep the container alive with an idle shell loop
+        // Keep the container alive with an idle shell loop. Install a SIGTERM trap and
+        // background the sleep with `wait`, so PID 1 (sh) exits promptly on `docker stop`
+        // instead of blocking in a foreground `sleep` until the stop timeout forces a
+        // SIGKILL. Relies only on `sh` + integer `sleep`, so it stays portable across
+        // minimal images (e.g. BusyBox/Alpine) where `sleep infinity` is unsupported.
         cmd.add("sh");
         cmd.add("-c");
-        cmd.add("while :; do sleep 3600; done");
+        cmd.add("trap 'exit 0' TERM; while :; do sleep 3600 & wait $!; done");
 
         return cmd;
     }
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandboxRunCommandTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandboxRunCommandTest.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.harness.agent.sandbox.impl.docker;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.agentscope.harness.agent.sandbox.WorkspaceSpec;
+import java.lang.reflect.Method;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit tests for the {@code docker run} command assembled by {@link DockerSandbox}, focused on
+ * the container keep-alive entrypoint.
+ *
+ * <p>Regression guard for issue #2884: the old entrypoint {@code while :; do sleep 3600; done}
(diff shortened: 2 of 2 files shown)


Buktal (approved):
LGTM

PR #2886, opened 2026-08-28
fix(core): pass builder permission context when loading legacy v1 session state

### LINKED ISSUE #2768: ReActAgent legacy-state load silently drops permissionContext (downgrades to DEFAULT)

chang6666:
Fixes #2768 / re-lands #2769

Commit 605978292 (#2760) rewrote loadOrCreateAgentStateForSlot and dropped the permCtx argument, calling the 3-arg LegacyStateLoader.loadFromLegacySessionWithPresence overload. Legacy v1 keys (memory_messages / toolkit_activeGroups) never carry a permission context, so the first turn after a 1.x -> 2.0 migration silently downgraded the caller-configured permission mode to DEFAULT, re-introducing the bug fixed by #2769.

## Changes
- loadOrCreateAgentStateForSlot passes the builder-time permission context to the 4-arg LegacyStateLoader overload (original #2769 re-land).
- LegacyStateLoader now also preserves the slot identity on the reconstructed state: sessionId and userId were previously left unset, so AgentState auto-generated a random sessionId that got persisted into the migrated agent_state.
- The 3-arg overloads on LegacyStateLoader (no permission context) are deprecated so future callers cannot silently fall back to DEFAULT again.
- End-to-end regression test (ReActAgentLegacyPermissionContextTest) covering: builder BYPASS surviving legacy load, migrated session id stability, and an existing v2 state staying authoritative over the builder template.

## Verification
- mvn -pl agentscope-core test -Dtest=ReActAgentLegacyPermissionContextTest,LegacyStateLoaderTest (JDK 17): 8/8 pass
- spotless:check and git diff --check clean

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/state/LegacyStateLoader.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/state/LegacyStateLoader.java
@@ -51,7 +51,13 @@ public final class LegacyStateLoader {
      * @param userId nullable user identifier
      * @param sessionId the session identifier (required)
      * @return a new AgentState populated with the legacy data
+     * @deprecated Use
+     *     {@link #loadFromLegacySession(AgentStateStore, String, String, PermissionContextState)}
+     *     instead. Legacy keys never carry a permission context, so this variant reconstructs the
+     *     state with the default permission mode and silently discards any caller-configured
+     *     context.
      */
+    @Deprecated
     public static AgentState loadFromLegacySession(
             AgentStateStore stateStore, String userId, String sessionId) {
         return loadFromLegacySessionWithPresence(stateStore, userId, sessionId, null).state();
@@ -64,6 +70,10 @@ public final class LegacyStateLoader {
      * Legacy keys carry no permission context; passing {@code permCtx} is the only way to
      * preserve a non-default permission mode (for example {@code BYPASS}) when reconstructing
      * state from a v1 session.
+     *
+     * <p>Unlike {@code freshState}, v1 keys must be upgraded in place: the reconstructed state
+     * keeps the {@code userId} / {@code sessionId} identity of the migrated session so it stays
+     * the same slot.
      *
      * @param stateStore the state store to read from
      * @param userId nullable user identifier
@@ -90,7 +100,14 @@ public final class LegacyStateLoader {
      * context to the reconstructed state; legacy keys themselves never carry one. Callers that
      * do not need a specific permission mode may use the {@code (stateStore, userId, sessionId)}
      * variant.
+     *
+     * @deprecated Use
+     *     {@link #loadFromLegacySessionWithPresence(AgentStateStore, String, String, PermissionContextState)}
+     *     instead. Legacy keys never carry a permission context, so this variant reconstructs the
+     *     state with the default permission mode and silently discards any caller-configured
+     *     context.
      */
+    @Deprecated
     public static LegacyLoadResult loadFromLegacySessionWithPresence(
             AgentStateStore stateStore, String userId, String sessionId) {
         return loadFromLegacySessionWithPresence(stateStore, userId, sessionId, null);
@@ -119,7 +136,10 @@ public final class LegacyStateLoader {
         Optional<ToolkitState> toolkitState =
                 stateStore.get(userId, sessionId, "toolkit_activeGroups", ToolkitState.class);
 
-        AgentState.Builder builder = AgentState.builder().context(msgs);
+        AgentState.Builder builder = AgentState.builder().sessionId(sessionId).context(msgs);
+        if (userId != null) {
+            builder.userId(userId);
+        }
         if (permCtx != null) {
             builder.permissionContext(permCtx);
         }
--- a/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentLegacyPermissionContextTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentLegacyPermissionContextTest.java
@@ -17,6 +17,7 @@ package io.agentscope.core.agent;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import io.agentscope.core.ReActAgent;
@@ -77,6 +78,11 @@ class ReActAgentLegacyPermissionContextTest {
                 PermissionMode.BYPASS,
                 state.getPermissionContext().getMode(),
                 "builder-supplied permission context must survive legacy v1 session loading");
+        assertEquals(
+                SESSION,
+                state.getSessionId(),
+                "migrated state must keep the slot's session id instead of a random value");
+        assertNull(state.getUserId(), "anonymous slot keeps a null user id");
         assertTrue(
                 state.getContext().stream()
                         .anyMatch(m -> "hello from v1".equals(m.getTextContent())),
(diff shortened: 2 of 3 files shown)


guslegend0510 (approved):
LGTM

chang6666 (commented):
Done

chang6666 (commented):
Done

agentscope-core/src/main/java/io/agentscope/core/state/LegacyStateLoader.java
```diff
@@ -119,7 +136,10 @@ public static LegacyLoadResult loadFromLegacySessionWithPresence(
         Optional<ToolkitState> toolkitState =
                 stateStore.get(userId, sessionId, "toolkit_activeGroups", ToolkitState.class);
 
-        AgentState.Builder builder = AgentState.builder().context(msgs);
+        AgentState.Builder builder = AgentState.builder().sessionId(sessionId).context(msgs);
```
oss-maintainer, line 139: **[Warning]** Forwarding `permCtx` closes one half of the builder-vs-legacy asymmetry, but the same path still drops `initialActiveToolGroups`. `ReActAgent.freshState()` seeds `ToolContextState` from the builder's initial groups (and from `agentId` as the session fallback); here `toolContext` is only set when the legacy `toolkit_activeGroups` key exists. So for a v1 session that has `memory_messages` but no `toolkit_activeGroups`, `found()` is true, the migrated state wins, and the builder's initial groups are silently lost; when the key exists they are discarded rather than merged.

Not introduced by this PR — but since this change is exactly about "the builder's intent must survive legacy loading", it is worth doing the tool-group dimension in the same sweep, or opening a follow-up issue so the two dimensions do not drift apart again. Either way the precedence rule (what wins on migration: persisted v1 data or builder template) should be written down once and applied uniformly to permission context, tool groups and session identity.

agentscope-core/src/main/java/io/agentscope/core/state/LegacyStateLoader.java
```diff
@@ -119,7 +136,10 @@ public static LegacyLoadResult loadFromLegacySessionWithPresence(
...
+        if (userId != null) {
```
oss-maintainer, line 140: **[Info]** Minor asymmetry in the null handling: `sessionId` is applied unconditionally (a `null` still lands on `AgentState`'s `newHex()` fallback, `AgentState.java:82`) while `userId` is guarded. The javadoc does say the session id is required, and `ReActAgent` always resolves one, so this is not a live bug — but a `sessionId != null ? sessionId : <something explicit>` guard (or `Objects.requireNonNull`) here would stop a future caller from silently reproducing the very bug this PR fixes, since the failure only surfaces on the shutdown-saver path that reads the id back off the state object (`ReActAgent.java:365-388`).

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentLegacyPermissionContextTest.java
```diff
@@ -0,0 +1,140 @@
...
+        assertNotNull(reply);
+
+        AgentState state = agent.getAgentState();
+        assertEquals(
+                PermissionMode.BYPASS,
+                state.getPermissionContext().getMode(),
+                "builder-supplied permission context must survive legacy v1 session loading");
+        assertEquals(
+                SESSION,
+                state.getSessionId(),
```
oss-maintainer, line 83: **[Info]** Good assertion — this is the one that would catch a regression, because per-call persistence keys off `CallExecution.slotKey`, whereas the *shutdown* saver re-derives `(userId, sessionId)` from the `AgentState` itself and would flush a migrated session into a random new slot. Worth strengthening with a read-back from the store after `agent.call(...)`:

```java
AgentState persisted = stateStore.get(null, SESSION, "agent_state", AgentState.class).orElse(null);
assertNotNull(persisted, "migrated session must be written back to its own slot");
assertEquals(SESSION, persisted.getSessionId());
```

That pins the slot identity where it actually matters rather than only in the in-memory object.

oss-maintainer (approved):
### Summary

Correct, well-scoped fix for the v1 → v2 legacy migration path, and the blast radius is bigger than the diff suggests. Verified against `main`:

- `ReActAgent.loadOrCreateAgentStateForSlot` called the permission-context-less `loadFromLegacySessionWithPresence(stateStore, userId, sessionId)` overload, so a builder-configured `permissionContext` (e.g. `BYPASS`) was silently replaced by `AgentState`'s default on any session that still lived under v1 keys — a permission-engine behaviour difference between "fresh session", "v2 session" and "migrated v1 session". Forwarding `permCtx` closes that.
- Setting `sessionId` / `userId` on the reconstructed state matters for persistence, not just logging: `AgentState` falls back to `newHex()` when the builder has no session id (`AgentState.java:82`), and the shutdown state saver derives `uid`/`sid` **from the state object** (`ReActAgent.java:365-388`), so a migrated session could be flushed to a brand-new slot. The added `assertEquals(SESSION, state.getSessionId())` is the right regression guard.
- Deprecating the two context-less overloads instead of deleting them keeps this source-compatible for `agentscope-harness` / `-distribution` / extensions, and both E2E tests (builder context survives legacy load; pre-existing v2 state wins over the builder template) pin the precedence rule in both directions.

Two non-blocking notes below: the same path still drops `initialActiveToolGroups` (pre-existing, not introduced here), and the new test could also assert the *persisted* slot.

CI green (License, Module Sync, ubuntu + windows builds), CLA signed, mergeable.

---

PR #2497, opened 2026-07-30
docs: fix README event count and Model javadoc to match code

helloworldtang:
## What & Why

Found two documentation/code inconsistencies while reviewing the 2.0 GA codebase. This PR fixes both — **docs only, no behavior change**.

### 1. Typed event count is 31, not 28
`AgentEventType` defines **31** enum values and `AgentEvent` registers **31** `@JsonSubTypes.Type`, but `README.md` (×2) and `README_zh.md` still advertise "28 typed events". None of the 31 events are deprecated, so 28 is simply stale.

Updated all three occurrences (README.md ×2, README_zh.md ×1) to 31.

### 2. `Model#getContextWindowSize()` javadoc points at a non-core component
The javadoc on `Model#getContextWindowSize()` says *"Used by the compaction middleware to ..."*, but `CompactionMiddleware` lives in **agentscope-harness**, not `agentscope-core`. The claim is accurate (the harness `CompactionMiddleware` does call `model.getContextWindowSize()` at `CompactionMiddleware.java:157`), but *"the compaction middleware"* reads as a core concept and misleads readers exploring core in isolation. Qualified it to *"the harness-layer compaction middleware"*.

## Verification
- `mvn -pl agentscope-core spotless:check` ✅ passes
- Docs-only change; no tests or behavior affected

## Checklist
- [x] Conventional commit messages (`docs(scope): subject`)
- [x] Focused on a single concern (documentation/code consistency)
- [x] No breaking changes

PR #2887, opened 2026-08-28
docs(tool): fix ToolBase builder javadoc example (no build() method)

chang6666:
`ToolBase.Builder` has no `build()` terminal method — the builder is passed directly to the `ToolBase(Builder)` constructor (as `ToolBaseTest` already does), so the class javadoc example did not compile.

Removes the stray `.build()` from the example. Same class of issue as #2497 (javadoc not matching code).


xzxiaoshan:
LGTM

jujn (approved):

Issue #2251, opened 2026-07-16
[Bug]: LocalFilesystemPersonalAssistantExampleTest flaky JUnit @TempDir cleanup race with session-tree-mirror thread


Asaduddin18:
**Describe the bug**

`LocalFilesystemPersonalAssistantExampleTest` (`agentscope-harness`, package `io.agentscope.harness.agent.example`) intermittently fails in CI with:

```
org.junit.platform.commons.JUnitException: Failed to close extension context
Caused by: java.io.IOException: Failed to delete temp directory /tmp/junit-... The following paths could not be deleted (see suppressed exceptions for details): <root>, agents, agents/my-local-assistant, agents/my-local-assistant/sessions
Suppressed: java.nio.file.DirectoryNotEmptyException: ...
```

This happens during JUnit's `@TempDir` teardown: it walks the temp directory tree to delete it, but a background thread named **`session-tree-mirror`** is still writing files into that same directory at teardown time, so the delete hits `DirectoryNotEmptyException` mid-walk. It's a teardown race between the test's async session-mirroring thread and JUnit's directory cleanup, not a logic bug in the test's assertions.

**To Reproduce**

Not reliably reproducible on demand — it's timing/scheduler dependent. Observed across multiple unrelated PRs in CI:

- PR #2113 (unrelated `OpenAITextEmbedding` change): failed on `ubuntu-latest` — [run](https://github.com/agentscope-ai/agentscope-java/actions/runs/29156943370/job/86555498510), method `localFilesystem_directDiskAccessFromHostProcess`.
- PR #2248 (unrelated `MemoryMaintenanceMiddleware` change): failed on `windows-latest` — [run](https://github.com/agentscope-ai/agentscope-java/actions/runs/29481707253/job/87566739340), method `localFilesystem_workspaceIsNotPartitionedByUserOrSession`. A CI run immediately prior on the exact same code passed cleanly.

Same test class, different failing methods, different OS, inconsistent pass/fail on identical code — consistent with a timing race rather than a deterministic bug.

**Expected behavior**

The test suite should not intermittently fail due to teardown races. Either:
- The `session-tree-mirror` background thread should be joined/stopped before the test method returns (so nothing writes to the temp dir during `@TempDir` cleanup), or
- The test should use a teardown mechanism that tolerates a still-active writer (e.g. retry-on-`DirectoryNotEmptyException` deletion, or an explicit shutdown hook for the mirror thread in an `@AfterEach`).

[...]

PR #2890, opened 2026-08-29
fix(agui): share resume coordination state across replicas

### LINKED ISSUE #2773: [Bug]: AguiResumeCoordinator loses HITL interrupt state across instances
**Describe the bug**

The AG-UI interrupt/resume contract state (`pendingInterruptsByThread`, `activeRunsByThread`) lives only in process-local `ConcurrentHashMap`s inside `AguiResumeCoordinator`. When the service is deployed with multiple replicas (e.g. 2 pods on k8s) behind a round-robin load balancer, a resume request can be routed to a replica that never saw the original interrupt. `validate()` then fails with `AGUI_INTERRUPT_CONTRACT_ERROR` (`"RunAgentInput.resume does not match any open interrupt"`), even though the client is correctly following the AG-UI resume contract.

The same state is lost on a single replica whenever the process restarts (rolling update, OOM eviction, failover).

Relevant code:

- `AguiResumeCoordinator.java:43-45` — state held in process-local `ConcurrentHashMap`s
- `AguiResumeCoordinator.java:56-61` — the failing check (`pending == null && hasResume() == true`)
- `AguiRequestProcessor.java:76` — coordinator instantiated inside the processor; no injection or extension point
- `AguiRequestProcessor.java:130-174` — `beginRun` / `trackPendingInterrupts` / `finishRun` wiring

**To Reproduce**

1. Deploy the AG-UI service with 2 replicas behind a round-robin LB (k8s Service / Nginx Ingress):

```yaml
apiVersion: apps/v1
kind: Deployment
spec:
  replicas: 2
```

1. Client sends a run request **without** `resume` → the agent issues a `tool_call` → server replies with `RUN_FINISHED` carrying an `interrupt` (`reason: "tool_call"`). The interrupt state is recorded **only on the replica that handled this request** (pod A).
2. Client sends the follow-up run request **with** `resume` (same `threadId`, same `interruptId`) → the LB routes it to the other replica (pod B).
3. Pod B has no record of the interrupt → emits `RUN_ERROR` with `code=AGUI_INTERRUPT_CONTRACT_ERROR`.

With 2 replicas and round-robin routing, roughly 50% of resume attempts fail.

**Expected behavior**

A resume request should be accepted regardless of which replica serves it. Either:

- the interrupt/resume state should be shared across replicas (e.g. a pluggable state store with a Redis backend, consistent with the existing `RedisDistributedStore` approach), or
- the limitation should be explicitly documented (e.g. "sticky sessions required for AG-UI tool-call interrupts").

**Error messages**

```json
{"type":"RUN_ERROR","threadId":"1787102227357","runId":"run-1787102359437","message":"RunAgentInput.resume does not match any open interrupt","code":"AGUI_INTERRUPT_CONTRACT_ERROR","timestamp":1787102359738}
```

- AgentScope-Java Version: 2.0.1
- Java Version: 17
- OS: Linux (k8s pods); reproduced from a Windows dev client

**Additional context**

- The official multi-replica guidance (`docs/v2/en/docs/others/going-to-production.md`) recommends `RedisDistributedStore` + `OssFilesystemSpec` and states "no sticky sessions required", but that applies to session/workspace state. The AG-UI resume coordinator has **no distributed backend and no SPI** to plug one in — `AguiResumeCoordinator` is a package-private `final` class with hardcoded `ConcurrentHashMap`s.
- `validate()` also requires the resume `interruptId` set to match the pending set **exactly** (`AguiResumeCoordinator.java:84-95`, missing/unknown ids are reported), so any mismatch fails hard — there is no lenient path.
- There is no TTL/cleanup for pending interrupt entries (`trackPendingInterrupts`, `AguiResumeCoordinator.java:180-198`); entries stay in memory until the next run of the same thread.

Suggested direction:

1. Extract the state access behind a small SPI (e.g. `AguiResumeStateStore`) with the in-memory implementation as the default, and provide a Redis-backed implementation in `agentscope-extensions-redis`.
2. Preserve cross-replica atomicity: `putIfAbsent`-style claim for `beginRun` (SET NX) and conditional delete for `finishRun` (compare-and-delete).
3. Either way, document the multi-replica requirement for AG-UI interr

ordinary-s:
### Summary

Fixes #2773. AG-UI resume requests can reach a different replica from the one that recorded an interrupt. This PR moves pending interrupts and active-run ownership behind a shared store so replicas validate the same pending state and compete for one atomic claim.

### Implementation

- Add `AguiResumeStateStore` with immutable pending snapshots, atomic claim, owner-checked release, and atomic owner-checked pending replacement. Keep `InMemoryAguiResumeStateStore` as the default.
- Add `RedisAguiResumeStateStore`, backed by one Redis hash per thread and Lua scripts. Claim before validating resume input; stale runs cannot overwrite pending state or release another run's owner.
- Support explicit processor store injection and optional Spring store beans in both MVC and WebFlux.

### Request lifecycle and recovery

- Run synchronous resume-store calls on Reactor's `boundedElastic` scheduler, including pending writes and cleanup after cancellation. Ordinary token events do not require a separate blocking task.
- Use a per-subscription resource to coordinate claim, state writes, and cleanup. Cancellation waits for an in-flight storage operation before releasing a confirmed claim; rejected or indeterminate claims are never blindly released.
- Convert claim, validation, resume-injection, and pending-write failures into AG-UI `RUN_ERROR` events. Emit `RUN_STARTED` only if it has not already been sent, respect `emitRunFinishedAfterError`, and withhold the successful `RUN_FINISHED` when its pending write fails.
- Attempt owner-checked release at most three times, with 50 ms and 100 ms retry delays, on completion, error, or cancellation. Log failed attempts, recovery, and exhaustion with thread/run IDs, phase, attempt count, and the exception. Retry only release, never agent execution.
- Preserve the original processing exception and attach exhausted cleanup failures as suppressed exceptions. Cleanup failures are logged rather than emitted as dropped Reactor errors or a second terminal lifecycle.
- Document asynchronous disposal, finite client timeouts, logging alerts, and administrative recovery that preserves pending interrupts in the English and Chinese v2 guides.

### Validation

Validated on JDK 21 with Java 17 compilation, including a disposable Redis 7.4 instance and actual Reactor Netty HTTP connections:

- 517 AG-UI tests passed.
- 12 Redis tests passed, including 7 real Redis recovery and failure-lifecycle cases (`-Dagui.redis.integration=true`; local port 16389, configurable via `-Dagui.redis.port`).
- 56 Spring configuration, resolver, session, and HTTP cases passed, including store selection, SSE errors, disconnect behavior, and contextual cleanup logging on completion/error/cancellation.
- 5 MVC controller cases passed in a separate targeted run.
- New lifecycle coverage includes cancellation during claim, bounded cleanup retries, lost release replies, same-run-ID rejected claims, preservation of pending state after errors, and storage calls off nonblocking threads.
- Repository-wide `mvn spotless:check` and `git diff --check` passed.

### Recovery limits

Release retries are bounded best effort; this PR does not introduce leases, fencing tokens, or TTL-based crash recovery. Exhausted cleanup, a process crash, or a lost claim reply can leave ownership behind. A claim with an unknown outcome is reported as an error without automatic claim retry or release.

Use fresh run IDs because they also identify ownership. Before administrative `releaseRun(threadId, oldRunId)`, confirm that the old agent, tools, and cleanup attempts have stopped. This owner-checked operation preserves pending interrupts; deleting the whole Redis key would lose resume state. Tools that ignore cancellation may outlive the protocol subscription, and a write whose reply was lost may already have committed.

### DIFF
--- a/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiRequestProcessor.java
+++ b/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiRequestProcessor.java
@@ -205,7 +205,13 @@ public class AguiRequestProcessor {
                                                         resumeCoordinator.finishRun(
                                                                 threadId, runId));
                             } catch (Throwable error) {
-                                resumeCoordinator.finishRun(threadId, runId);
+                                try {
+                                    resumeCoordinator.finishRun(threadId, runId);
+                                } catch (Throwable cleanupError) {
+                                    if (cleanupError != error) {
+                                        error.addSuppressed(cleanupError);
+                                    }
+                                }
                                 return processorErrorEvents(input, error);
                             }
                         });
--- a/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiRequestProcessor.java
+++ b/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiRequestProcessor.java
@@ -81,7 +81,11 @@ public class AguiRequestProcessor {
                 builder.adapterFactory != null
                         ? builder.adapterFactory
                         : AguiAgentAdapterFactory.defaultFactory();
-        this.resumeCoordinator = new AguiResumeCoordinator();
+        this.resumeCoordinator =
+                new AguiResumeCoordinator(
+                        builder.resumeStateStore != null
+                                ? builder.resumeStateStore
+                                : new InMemoryAguiResumeStateStore());
         this.runtimeContextResolver = builder.runtimeContextResolver;
     }
 
@@ -363,6 +367,7 @@ public class AguiRequestProcessor {
         private AguiAdapterConfig config;
         private AguiAgentAdapterFactory adapterFactory;
         private AguiRuntimeContextResolver runtimeContextResolver;
+        private AguiResumeStateStore resumeStateStore;
 
         /**
          * Set the agent resolver.
@@ -406,6 +411,22 @@ public class AguiRequestProcessor {
          */
         public Builder runtimeContextResolver(AguiRuntimeContextResolver runtimeContextResolver) {
             this.runtimeContextResolver = runtimeContextResolver;
+            return this;
+        }
+
+        /**
+         * Set the store used to coordinate active runs and pending interrupts.
+         *
+         * <p>When omitted, each processor uses its own {@link InMemoryAguiResumeStateStore},
+         * preserving the process-local default. Multi-replica deployments should provide a
+         * shared implementation with the atomic semantics required by {@link
+         * AguiResumeStateStore}.
+         *
+         * @param resumeStateStore The resume coordination state store
+         * @return This builder
+         */
+        public Builder resumeStateStore(AguiResumeStateStore resumeStateStore) {
+            this.resumeStateStore = resumeStateStore;
             return this;
         }
 
--- a/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiRequestProcessor.java
+++ b/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiRequestProcessor.java
@@ -207,10 +207,8 @@ public class AguiRequestProcessor {
                             } catch (Throwable error) {
                                 try {
                                     resumeCoordinator.finishRun(threadId, runId);
-                                } catch (Throwable cleanupError) {
-                                    if (cleanupError != error) {
-                                        error.addSuppressed(cleanupError);
-                                    }
+                                } catch (RuntimeException ignored) {
+                                    // Best-effort cleanup; preserve the original processing error.
                                 }
                                 return processorErrorEvents(input, error);
                             }
--- a/agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/agui/RedisAguiResumeStateStore.java
+++ b/agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/agui/RedisAguiResumeStateStore.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.extensions.redis.agui;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import io.agentscope.core.agui.event.AguiEvent;
+import io.agentscope.core.agui.processor.AguiResumeStateStore;
+import io.agentscope.core.util.JsonUtils;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import redis.clients.jedis.UnifiedJedis;
+
+/**
+ * Redis-backed {@link AguiResumeStateStore}.
+ *
+ * <p>Each thread is stored in one Redis hash. Run claims, owner-checked releases, and
+ * owner-checked pending interrupt replacements each execute as a single Lua script so replicas
+ * cannot interleave the read and mutation.
+ *
+ * <p>Example:
+ *
+ * <pre>{@code
+ * AguiResumeStateStore resumeStore = new RedisAguiResumeStateStore(jedis);
+ * AguiRequestProcessor processor = AguiRequestProcessor.builder()
+ *     .agentResolver(resolver)
+ *     .resumeStateStore(resumeStore)
+ *     .build();
+ * }</pre>
+ */
+public final class RedisAguiResumeStateStore implements AguiResumeStateStore {
+
+    /** Default key prefix for AG-UI resume coordination state. */
+    public static final String DEFAULT_KEY_PREFIX = "agentscope:agui:resume:";
+
+    private static final String ACTIVE_RUN_FIELD = "activeRunId";
+    private static final String PENDING_INTERRUPTS_FIELD = "pendingInterrupts";
+
+    private static final String CLAIM_RUN_SCRIPT =
+            "local owner = redis.call('HGET', KEYS[1], ARGV[1]) "
+                    + "if owner then return owner end "
+                    + "redis.call('HSET', KEYS[1], ARGV[1], ARGV[2]) "
+                    + "return false";
+
+    private static final String RELEASE_RUN_SCRIPT =
+            "local owner = redis.call('HGET', KEYS[1], ARGV[1]) "
+                    + "if owner ~= ARGV[2] then return 0 end "
+                    + "redis.call('HDEL', KEYS[1], ARGV[1]) "
+                    + "if redis.call('HLEN', KEYS[1]) == 0 then redis.call('DEL', KEYS[1]) end "
+                    + "return 1";
+
+    private static final String REPLACE_PENDING_SCRIPT =
+            "local owner = redis.call('HGET', KEYS[1], ARGV[1]) "
+                    + "if owner ~= ARGV[2] then return 0 end "
+                    + "if ARGV[4] == '1' then "
+                    + "redis.call('HDEL', KEYS[1], ARGV[3]) "
+                    + "else redis.call('HSET', KEYS[1], ARGV[3], ARGV[5]) end "
+                    + "return 1";
+
+    private static final TypeReference<Map<String, AguiEvent.Interrupt>> INTERRUPTS_TYPE =
+            new TypeReference<>() {};
+
+    private final UnifiedJedis jedis;
+    private final String keyPrefix;
+
+    /** Create a store with the {@linkplain #DEFAULT_KEY_PREFIX default key prefix}. */
+    public RedisAguiResumeStateStore(UnifiedJedis jedis) {
+        this(jedis, DEFAULT_KEY_PREFIX);
+    }
+
+    /**
+     * Create a store with a custom key prefix.
+     *
+     * @param jedis initialized Jedis client
+     * @param keyPrefix prefix for AG-UI resume state keys
+     */
+    public RedisAguiResumeStateStore(UnifiedJedis jedis, String keyPrefix) {
+        this.jedis = Objects.requireNonNull(jedis, "jedis");
+        this.keyPrefix = normalizePrefix(keyPrefix);
+    }
+
+    @Override
+    public Map<String, AguiEvent.Interrupt> getPendingInterrupts(String threadId) {
+        String json = jedis.hget(key(threadId), PENDING_INTERRUPTS_FIELD);
+        if (json == null || json.isEmpty()) {
+            return Map.of();
+        }
+        Map<String, AguiEvent.Interrupt> interrupts =
+                JsonUtils.getJsonCodec().fromJson(json, INTERRUPTS_TYPE);
+        return interrupts == null || interrupts.isEmpty() ? Map.of() : Map.copyOf(interrupts);
+    }
+
+    @Override
+    public RunClaim claimRun(String threadId, String runId) {
+        requireIdentifier(runId, "runId");
+        Object result =
+                jedis.eval(
+                        CLAIM_RUN_SCRIPT, List.of(key(threadId)), List.of(ACTIVE_RUN_FIELD, runId));
+        String activeRunId = asString(result);
+        return activeRunId == null ? RunClaim.acquired() : RunClaim.rejected(activeRunId);
+    }
+
+    @Override
+    public void releaseRun(String threadId, String runId) {
+        requireIdentifier(runId, "runId");
+        jedis.eval(RELEASE_RUN_SCRIPT, List.of(key(threadId)), List.of(ACTIVE_RUN_FIELD, runId));
+    }
+
+    @Override
+    public boolean replacePendingInterrupts(
+            String threadId, String runId, Map<String, AguiEvent.Interrupt> pendingInterrupts) {
+        requireIdentifier(runId, "runId");
+        Map<String, AguiEvent.Interrupt> snapshot =
+                Map.copyOf(Objects.requireNonNull(pendingInterrupts, "pendingInterrupts"));
+        boolean clear = snapshot.isEmpty();
+        String json = clear ? "" : JsonUtils.getJsonCodec().toJson(snapshot);
+        Object result =
+                jedis.eval(
+                        REPLACE_PENDING_SCRIPT,
+                        List.of(key(threadId)),
+                        List.of(
+                                ACTIVE_RUN_FIELD,
+                                runId,
+                                PENDING_INTERRUPTS_FIELD,
+                                clear ? "1" : "0",
+                                json));
+        return asLong(result) == 1L;
+    }
+
+    private String key(String threadId) {
+        requireIdentifier(threadId, "threadId");
+        return keyPrefix + threadId;
+    }
+
+    private static void requireIdentifier(String value, String name) {
+        Objects.requireNonNull(value, name);
+        if (value.isBlank()) {
+            throw new IllegalArgumentException(name + " must not be blank");
+        }
+    }
+
+    private static String normalizePrefix(String prefix) {
+        if (prefix == null || prefix.isBlank()) {
+            return DEFAULT_KEY_PREFIX;
+        }
+        return prefix.endsWith(":") ? prefix : prefix + ":";
+    }
+
+    private static String asString(Object value) {
+        if (value == null) {
+            return null;
+        }
+        if (value instanceof byte[] bytes) {
+            return new String(bytes, StandardCharsets.UTF_8);
+        }
+        return value.toString();
+    }
+
+    private static long asLong(Object value) {
+        if (value instanceof Number number) {
+            return number.longValue();
+        }
+        String string = asString(value);
+        return string != null ? Long.parseLong(string) : 0L;
+    }
+}
--- a/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiResumeCoordinator.java
+++ b/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiResumeCoordinator.java
@@ -27,9 +27,8 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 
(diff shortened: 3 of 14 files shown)


ordinary-s:
The CI failure appears to be caused by the known flaky @TempDir cleanup race in agentscope-harness (DirectoryNotEmptyException, tracked in #2251). This PR does not modify the agentscope-harness module, and the failure occurred before the AG-UI/Redis modules were reached in the reactor build. Could someone please re-run the failed job when convenient?

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiRequestProcessor.java
```diff
@@ -152,67 +157,98 @@ public ProcessResult process(AguiRuntimeContextRequest<?> request) {
...
+                                            });
                         });
         return new ProcessResult(agent, events, runtimeContext);
     }
 
-    private Flux<AguiEvent> processorErrorEvents(RunAgentInput input, Throwable error) {
+    private Flux<AguiEvent> prepareRun(
+            RunAgentInput input, RuntimeContext runtimeContext, Agent agent, AguiRunLifecycle run) {
+        AguiResumeCoordinator.ResumeContractResult beginResult = run.begin();
+        if (beginResult.isError()) {
```
oss-maintainer, line 201: On this path the run has already been claimed (`beginRun` invokes `onClaimed` before `validate(input)`), but the release only runs from the `Flux.usingWhen` completion handler, i.e. after the `contractErrorEvents(...)` flux terminates. A subscriber that applies no demand to that error stream therefore keeps the per-thread owner slot held, and every later run/resume on the same threadId is rejected with "Thread already has an active run" until the subscription is cancelled. `beginRun(RunAgentInput)` releases synchronously on a validation error and `beginRunReleasesOwnershipWhenValidationFails` pins that — this second caller does not get the same guarantee. Since the contract-error stream carries no owner state, could the claim be released right here (or the error flux made demand-independent), so a rejected run cannot pin the thread slot behind a slow client?

oss-maintainer (commented):
### Summary
First review of this PR (never reviewed before). Replacing the coordinator's in-memory `activeRunsByThread` / `pendingInterruptsByThread` maps with the `AguiResumeStateStore` SPI (in-memory + Redis implementations, atomic Lua claim/release/replace) is the right shape for stateless horizontal scaling: ownership is now an explicit `RunClaim`, `releaseRun` is a conditional delete, `replacePendingInterrupts` is owner-checked in **both** implementations (the in-memory one compares `state.activeRunId()`, the Redis script compares `ACTIVE_RUN_FIELD`), and a store failure is propagated instead of being treated as missing state — which is the failure mode that would otherwise silently allow two concurrent runs for one thread. `AguiRunLifecycle` + `Flux.usingWhen` make claim/setup/cancel/release one unit, and the test set is unusually thorough for this area (cross-replica resume, duplicate-rejected-claim-must-not-release-the-owner, lost-release-reply retry, cancel-during-claim, non-blocking-thread offload).

Coverage caveat, so this is not read as a full sign-off: the diff is +2608/-105 over 22 files. This pass read the AG-UI core (`AguiRequestProcessor`, `AguiResumeCoordinator`, `AguiRunLifecycle`, `AguiResumeStateStore`, the in-memory store) and the Redis store's owner-check/claim paths in full, and the Spring starters/tests only as a summary. No build or test run was possible here — this host has no JDK 17 or Maven (`runtime-env` reports all toolchains `[NOT FOUND]`), so the concurrency claims are unverified by execution.

One concern is inline: the contract-error path in `AguiRequestProcessor` leaves an already-claimed run to be released by the `usingWhen` completion handler, so a subscriber that never applies demand can pin the thread's owner slot.

---

oss-maintainer:

```bash
git fetch origin
git checkout fix/agui-shared-resume-state
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiRequestProcessor.java
ordinary-s: Thanks for pointing this out. Addressed in 8944797.

The contract-error path now awaits run.finish("validation", null) before subscribing to the error-event stream, so cleanup no longer depends on downstream demand for those events.

This reuses the lifecycle's shared, cached cleanup process and a single bounded release-retry budget across early cleanup and subsequent completion/cancellation. Rejected and indeterminate claims remain excluded from release.

Added regression coverage in AguiContractErrorLifecycleTest for zero/partial demand, a subsequent valid run acquiring the same thread, late completion/cancellation preserving the new owner, and release retry/failure scenarios.

ordinary-s:
The latest CI is now green, and the PR is mergeable without conflicts. @oss-maintainer could you please take another look when convenient?

agentscope-extensions/agentscope-extensions-redis/src/test/java/io/agentscope/extensions/redis/agui/RedisAguiResumeLifecycleTest.java
```diff
@@ -0,0 +1,393 @@
...
+import org.reactivestreams.Subscription;
+import reactor.core.Disposable;
+import reactor.core.publisher.BaseSubscriber;
+import reactor.core.publisher.Flux;
+import reactor.core.scheduler.Schedulers;
+import redis.clients.jedis.JedisPooled;
+import redis.clients.jedis.exceptions.JedisConnectionException;
+
+/** Run with -Dagui.redis.integration=true against a local, disposable Redis on port 16389. */
+@EnabledIfSystemProperty(named = "agui.redis.integration", matches = "true")
```
oss-maintainer, line 67: **[Warning]** The only tests that exercise the cross-replica semantics of this PR against a real Redis — `RedisAguiResumeLifecycleTest` (lost release reply, indeterminate claim, exhausted release, zero-demand contract error across two stores) — are gated behind `@EnabledIfSystemProperty(named = "agui.redis.integration", matches = "true")` plus a local Redis on port 16389, and `.github/workflows/maven-ci.yml` runs plain `mvn -B -T1 clean verify`, so they never execute in CI. The always-on `RedisAguiResumeStateStoreTest` mocks `jedis.eval(...)` entirely, which means the actual Lua bodies (`CLAIM_RUN_SCRIPT`, `RELEASE_RUN_SCRIPT`, `REPLACE_PENDING_SCRIPT` — KEYS/ARGV indices, owner checks, the Lua `false`→nil reply mapping in `claimRun`, the conditional `DEL` in `RELEASE_RUN_SCRIPT`) are never executed by the default suite; a typo or semantic regression in any script would pass CI while silently breaking claim/release/ownership across replicas. I verified the scripts line-by-line and they look correct, so this is a coverage gap, not a known bug. Suggestion: run this test class in CI against an ephemeral Redis (testcontainers or a `redis` service container) by setting `-Dagui.redis.integration=true`, or at minimum add always-on unit tests asserting the exact script strings and the `KEYS`/`ARGV` composition for all three operations (today only `releaseUsesOneOwnerCheckedRedisOperation` and `pendingReplacementReturnsOwnerCheckResultAndClearsWithSameOperation` check argument lists, and neither validates the script text or its Lua semantics).

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/agui/RedisAguiResumeStateStore.java
```diff
@@ -0,0 +1,180 @@
...
+
+    private String key(String threadId) {
+        requireIdentifier(threadId, "threadId");
+        return keyPrefix + threadId;
+    }
+
+    private static void requireIdentifier(String value, String name) {
+        Objects.requireNonNull(value, name);
+        if (value.isBlank()) {
+            throw new IllegalArgumentException(name + " must not be blank");
```
oss-maintainer, line 152: **[Info]** The two `AguiResumeStateStore` implementations disagree on blank identifiers: `RedisAguiResumeStateStore.key()` → `requireIdentifier` throws `IllegalArgumentException` for a blank/whitespace `threadId`, while `InMemoryAguiResumeStateStore` only does `Objects.requireNonNull` (e.g. `getPendingInterrupts` at InMemoryAguiResumeStateStore.java:33 accepts `""`). A request with `threadId=""` is therefore processed normally with the default in-memory store but fails with `RUN_ERROR`/`INVALID_INPUT_ERROR` once a shared Redis store is configured — exactly the kind of deployment-dependent behaviour change the SPI is meant to prevent, and the `AguiResumeStateStore` javadoc does not state any identifier requirements. Suggestion: document the `threadId`/`runId` identifier contract (non-null, non-blank) on the `AguiResumeStateStore` interface and align `InMemoryAguiResumeStateStore` to reject blank values the same way, so both stores behave identically for malformed inputs.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/test/java/io/agentscope/core/agui/processor/AguiResumeLifecycleTest.java
```diff
@@ -0,0 +1,519 @@
...
+            await(validated);
+            assertEquals(1, acquired.get(), "the real claim must have succeeded");
+            boolean cleanupObserved = released.await(1, TimeUnit.SECONDS);
+            assertTrue(events.isEmpty());
+            assertFalse(
+                    subscriber.isDisposed(), "no cancellation or terminal delivery before check");
+            assertEquals(0, executed.get());
+            assertEquals(PENDING, store.getPendingInterrupts("T"));
+            AguiResumeStateStore.RunClaim next = store.claimRun("T", "B");
+            System.out.println(
```
oss-maintainer: **[Info]** `contractErrorReleasesOwnershipWithoutDemand` contains a leftover `System.out.println("C01: acquired=...")` debug block (lines 148–158) that dumps internal claim/cleanup state on every run of the test. It adds noise to CI output and looks like an unfinished debugging aid from the commit that introduced the test. Suggestion: delete the println block — the surrounding `assertTrue(...)`/latch assertions already capture the same state — or replace it with a targeted assertion message if some of that detail was meant to be kept.

oss-maintainer (commented):
### Summary

Since the last review round (commit 168c76f9), 5 new commits landed, the largest being 89447971 "coordinate early cleanup and disconnect cancellation". Re-reviewed the full 25-file diff at head, focused on the ownership semantics:

- **SPI (`AguiResumeStateStore`)**: `claimRun` (create-if-absent), owner-checked `releaseRun`, and owner-checked atomic `replacePendingInterrupts` are correctly specified; `InMemoryAguiResumeStateStore` implements them atomically via `ConcurrentHashMap.compute`/`computeIfPresent`, and `RedisAguiResumeStateStore` implements them as three single-key Lua scripts. I verified both implementations agree on all edge cases I could construct (double claim, same-run-id reclaim, release of a non-owner, replace after release, pending-interrupt preservation across release).
- **Redis Lua scripts**: KEYS/ARGV usage is correct, each script touches a single key (hash-slot safe in Redis Cluster), `return false` → nil maps correctly to `RunClaim.acquired()`, the release script is idempotent and preserves `pendingInterrupts`, and plain `EVAL` avoids NOSCRIPT issues. There is deliberately **no TTL/lease** on the coordination keys — this is explicitly documented as a bounded best-effort design with manual recovery steps, so it is a stated limit rather than a defect.
- **Ownership lifecycle (`AguiRunLifecycle`)**: the cached `close().then(release())` cleanup (verified that Reactor's `Mono.cache()` does not cancel the source on subscriber cancel) gives one shared cleanup per subscription with a single 3-attempt/50–100 ms release retry budget; cancellation during an in-flight claim is serialized through the lifecycle monitor; a lost claim reply is treated as indeterminate and never blindly released; a lost release reply is safe because the release script is owner-checked and idempotent. Suppressed-exception attachment, `RUN_STARTED` at most once, and withholding `RUN_FINISHED` on a failed pending write all match the code.
- **Reactor**: all store calls (claim, validate reads, pending writes, release, cancellation interrupt) run on `boundedElastic`; terminal events are delivered after cleanup completes; tests assert no dropped errors (`Hooks.onErrorDropped`) and no event-loop blocking.
- **Spring auto-config**: MVC/WebFlux expose `ObjectProvider<AguiResumeStateStore>` and default to per-processor in-memory state when absent; `interruptOnDisconnect` now maps to the processor's `interruptOnCancel` with disposal-based cancellation. Backward compatible — only additive builder methods; no existing public signatures changed; the redis module pom addition follows the module's existing `provided`+`optional` pattern and both modules are already registered in `agentscope-all`/`agentscope-bom`.
- **Docs (en + zh)**: every behavioural claim I checked (cleanup budget and log levels, indeterminate claims, RUN_STARTED/RUN_FINISHED rules, no-TTL limits, manual `releaseRun` recovery) matches the implementation in both languages.
- **Tests**: cancellation-during-claim, cross-coordinator/cross-replica races, lost replies, exhausted release, zero-demand cleanup, and run-id reuse are all covered — though the real-Redis cross-replica tests are opt-in (see findings).

Overall: no correctness defects found in the distributed-ownership core; the remaining findings are a CI test-coverage gap and two minor items.

### Previous findings status

- **AguiRequestProcessor.java ~211 (UNRESOLVED from last round)** — "**On this path the run has already been claimed (beginRun invokes onClaimed before validate(input)), but the release only runs from the Flux.using resource cleanup...**" → **FIXED at head 89447971.** `prepareRun` now handles the validation-failure path with `run.finish("validation", null).thenMany(contractErrorEvents(...))` (AguiRequestProcessor.java:210-219), which runs the owner-checked release as part of the subscribed stream before the error events are delivered, even with zero downstream demand or a subscriber that never requests events. The claim/validate ordering is unchanged (`AguiResumeCoordinator.beginRun(input, onClaimed)` invokes `onClaimed` before `validate`), but every exit path — validation error, validate-read failure, synchronous setup failure, cancellation during claim, claim failure (indeterminate, no release) — now flows through `Flux.usingWhen`'s complete/error/cancel hooks into the shared cached cleanup. Directly covered by `AguiResumeLifecycleTest.contractErrorReleasesOwnershipWithoutDemand`, `validationReadFailureReleasesWithoutDemand`, `cancelDuringSuccessfulClaimWaitsForClaimAndReleasesWithoutRunningAgent`, and the Redis-gated `zeroDemandContractErrorAllowsReplicaResumeAndPreservesNewOwner`.


### Verdict

**COMMENT** — not a block: the ownership/claim-release core is correct and the previously raised leak path is fixed and regression-tested at `89447971`. Only the CI coverage gap (real-Redis Lua scripts never execute in `mvn verify`) is worth addressing before merge; the two Info items are quick cleanups. CI is green on the current head and the PR is mergeable, so this is ready for a maintainer to look at once the unresolved thread can be resolved.

### Findings
- **[WARNING]** `RedisAguiResumeLifecycleTest.java:67` — The only tests that exercise the cross-replica semantics of this PR against a real Redis — `RedisAguiResumeLifecycleTest` (lost release reply, indeterminate claim, exhausted release, zero-demand contract error across two stores) — are gated behind `@EnabledIfSystemProperty(named = "agui.redis.integration", matches = "true")` plus a local Redis on port 16389, and `.github/workflows/maven-ci.yml` runs plain `mvn -B -T1 clean verify`, so they never execute in CI. The always-on `RedisAguiResumeStateStoreTest` mocks `jedis.eval(...)` entirely, which means the actual Lua bodies (`CLAIM_RUN_SCRIPT`, `RELEASE_RUN_SCRIPT`, `REPLACE_PENDING_SCRIPT` — KEYS/ARGV indices, owner checks, the Lua `false`→nil reply mapping in `claimRun`, the conditional `DEL` in `RELEASE_RUN_SCRIPT`) are never executed by the default suite; a typo or semantic regression in any script would pass CI while silently breaking claim/release/ownership across replicas. I verified the scripts line-by-line and they look correct, so this is a coverage gap, not a known bug. Suggestion: run this test class in CI against an ephemeral Redis (testcontainers or a `redis` service container) by setting `-Dagui.redis.integration=true`, or at minimum add always-on unit tests asserting the exact script strings and the `KEYS`/`ARGV` composition for all three operations (today only `releaseUsesOneOwnerCheckedRedisOperation` and `pendingReplacementReturnsOwnerCheckResultAndClearsWithSameOperation` check argument lists, and neither validates the script text or its Lua semantics).
- **[INFO]** `RedisAguiResumeStateStore.java:152` — The two `AguiResumeStateStore` implementations disagree on blank identifiers: `RedisAguiResumeStateStore.key()` → `requireIdentifier` throws `IllegalArgumentException` for a blank/whitespace `threadId`, while `InMemoryAguiResumeStateStore` only does `Objects.requireNonNull` (e.g. `getPendingInterrupts` at InMemoryAguiResumeStateStore.java:33 accepts `""`). A request with `threadId=""` is therefore processed normally with the default in-memory store but fails with `RUN_ERROR`/`INVALID_INPUT_ERROR` once a shared Redis store is configured — exactly the kind of deployment-dependent behaviour change the SPI is meant to prevent, and the `AguiResumeStateStore` javadoc does not state any identifier requirements. Suggestion: document the `threadId`/`runId` identifier contract (non-null, non-blank) on the `AguiResumeStateStore` interface and align `InMemoryAguiResumeStateStore` to reject blank values the same way, so both stores behave identically for malformed inputs.
- **[INFO]** `AguiResumeLifecycleTest.java:148` — `contractErrorReleasesOwnershipWithoutDemand` contains a leftover `System.out.println("C01: acquired=...")` debug block (lines 148–158) that dumps internal claim/cleanup state on every run of the test. It adds noise to CI output and looks like an unfinished debugging aid from the commit that introduced the test. Suggestion: delete the println block — the surrounding `assertTrue(...)`/latch assertions already capture the same state — or replace it with a targeted assertion message if some of that detail was meant to be kept.

---
*Automated review by github-manager-bot — re-review after the rebase/fix push; replying to the request for re-review.*

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiRequestProcessor.java
oss-maintainer: @ordinary-s re-review posted on the current head `89447971`. This thread is fixed: `prepareRun` now releases the claim on the contract-error path via `run.finish("validation", null).thenMany(contractErrorEvents(...))` (AguiRequestProcessor.java:210-219), so ownership is released even with zero downstream demand, and `AguiResumeLifecycleTest.contractErrorReleasesOwnershipWithoutDemand` guards it. Resolving this is up to you — no further action needed from my side.

The only substantive item left is the CI coverage gap in `RedisAguiResumeLifecycleTest` (real Redis + `-Dagui.redis.integration=true`), so the Lua claim/release/replace scripts never run in `mvn verify`. Thanks for the thorough cancellation-lifecycle work in this round.

---

ordinary-s:
The latest CI is now green, and the PR is mergeable without conflicts. @oss-maintainer could you please take another look when convenient?

oss-maintainer (approved):
### Summary
The new commit is a clean response to the identifier-contract feedback: `AguiResumeStateStore` now states the contract (non-null, non-blank, used raw — no trimming), `InMemoryAguiResumeStateStore` and `RedisAguiResumeStateStore` validate identically via one helper, the Redis tests prove rejection happens *before* any Jedis interaction (`verifyNoInteractions`), and the `validThreadIdIsUsedWithoutNormalization` cases lock in the raw-key semantics on both backends. The new `redis-integration` CI job — real `redis:7.4.11` service plus a surefire-XML assertion that fails on zero tests or any skip — is exactly the guard that keeps the shared-state claim from silently rotting. Approving from a review standpoint; the two notes below are non-blocking.

### Notes
- **[Info]** Key naming is open-ended by design: a `threadId` of `" x "` or with a newline passes validation and becomes part of the Redis key (`<prefix>" x "`), and Redis keys are unlimited so nothing will error — it just becomes hard to inspect from a client. Fine to keep raw (the contract explicitly forbids normalisation), but if upstream AG-UI ever supplies untrusted ids, a length/charset bound in the store layer would be cheaper to debug than a stuck interrupt key.
- **[Info]** `redis-integration` only runs `RedisAguiResumeLifecycleTest` with `-am`; the report check catches "did not run", so a renamed class fails loudly rather than passing vacuously. Consider a `paths-ignore` for docs-only pushes later if the extra 30-minute job starts costing queue time.

---

PR #2892, opened 2026-08-29
fix(harness): skip composite routes for PathPolicy host mounts

### LINKED ISSUE #2690: [Bug]:CompositeFilesystem 的路由前缀会劫持同名的宿主挂载点（/knowledge、/skills…），静默读不到文件
> Labels 建议：`bug` / `harness` / `filesystem`

## 环境

| | |
|---|---|
| agentscope-harness | 2.0.2 |
| store | `agentscope-extensions-mysql` 2.0.2（`JdbcStore` → MySQL 8.0.21） |
| 场景 | **非 sandbox**，`CompositeFilesystem` + 远程 store，K8s 2 副本 |
| 宿主挂载 | `/repos`（源码，只读）、`/knowledge`（知识库，只读）、`/skills`（技能，只读）、`/workspace`（emptyDir，可写） |

## 现象

系统提示词里明确告诉模型「知识库根目录 = `/knowledge`，需用全路径访问」。模型照做发出 `read_file("/knowledge/services.md")`，**读不到**。

没有异常、没有 WARN，只是返回一个失败结果 —— 从日志上看不出任何异常。

## 根因

`CompositeFilesystem.routeForPath()` 在匹配前会**剥掉前导斜杠**，因此**宿主绝对路径也会参与路由前缀匹配**：

```java
// CompositeFilesystem.java:100-128
private RouteResult routeForPath(String path) {
    String matchPath = stripLeadingSlash(path);            // :106  "/knowledge/x.md" → "knowledge/x.md"
    for (RouteEntry entry : sortedRoutes) {
        ...
        if (matchPath.startsWith(normalizedPrefix)) {      // :122  命中 "knowledge/" 路由
            String suffix = matchPath.substring(normalizedPrefix.length());
            String backendPath = suffix.isEmpty() ? "/" : "/" + suffix;   // :124
            return new RouteResult(entry.backend(), backendPath, entry.prefix());
        }
    }
    return new RouteResult(defaultBackend, path, null);
}
```

于是 `/knowledge/services.md` 被转发到 `knowledge/` 这条路由：

- 上层 = `RemoteFilesystem`（`knowledge` 段），我们从不往这里写 → 空
- 下层 = `LocalFilesystem(workspace/knowledge, SANDBOXED)`，workspace 是 emptyDir → 空

**读取失败。而宿主机上 `/knowledge/services.md` 明明存在**，只是永远不会被访问到 —— 那需要落到 defaultBackend、由它的 `PathPolicy` 白名单放行。

`RemoteFilesystemSpec.toFilesystem()` 默认注册的 10 条路由里，`knowledge/`、`skills/`、`subagents/`、`plans/`、`memory/` 都可能与宿主挂载点撞名。我们的 `/repos` 之所以没事，纯粹因为没有叫 `repos/` 的路由。

## 复现

1. 用 `RemoteFilesystemSpec` + 任意 `BaseStore` 构建 agent
2. 在宿主机上准备 `/knowledge/services.md`，并把 `/knowledge` 加进可读根（我们是自建 `PathPolicy`；官方 spec 下 defaultBackend 是 `UNRESTRICTED`，理论上任何绝对路径都该放行）
3. 调用 `filesystem.read(rc, "/knowledge/services.md", 0, 10)`
4. 预期读到内容；实际失败 —— 请求被转发到了远程 `knowledge` 段

## 影响

- **静默失败**：无异常、无日志，只有"文件读不到"这一个现象，排查成本高
- 任何把只读卷挂在与路由前缀同名路径上的部署都会中招，而这几个名字（knowledge、skills、plans）恰恰是最自然的命名
- 我们最终只能**弃用 `knowledge/` 这条路由**（我们从不往那儿写，弃用无损），让 `/knowledge/**` 正常落 defaultBackend

## 建议

匹配前区分「宿主绝对路径」与「workspace 相对路径」，例如任一：

- 只对**不以 `/` 开头**的路径做路由匹配；以 `/` 开头的先尝试 defaultBackend，失败再回退路由（会改变现有语义，需评估）
- 提供 `RemoteFilesystemSpec#excludeHostRoots(Path...)`，匹配时显式排除这些前缀
- 至少在路由未命中任何文件、而该路径在宿主机上存在时打一条 WARN，把「疑似被路由劫持」暴露出来

## 相关

可能与 [#2683](https://github.com/agentscope-ai/agentscope-java/issues/2683) 相关，但根因不同：那条是 workspace **内部** `<agent-id>/knowledge` 的路径解析；本条是**外部宿主挂载点**与路由前缀撞名。

jialiuyang:
## AgentScope-Java Version

2.0.x (main)

## Description

`CompositeFilesystem.routeForPath` strips a leading slash before prefix matching so `/skills/foo` and `skills/foo` hit the same store. That also makes a host path such as `/knowledge/services.md` match the `knowledge/` route. The request is forwarded to the remote/overlay segment (usually empty) and the file on the host mount is never read — no exception, just a failed result.

This change treats extra `PathPolicy` roots on a `LocalFilesystem` default backend (anything other than the workspace cwd) as host mounts and keeps those absolute paths on the default backend. An optional constructor argument accepts the same roots when the default store is not `LocalFilesystem`. Workspace-relative `knowledge/...` paths still use the route.

Fixes #2690

## Checklist

- [x] Targeted tests passing (`CompositeFilesystemTest`)
- [x] Related documentation has been updated — N/A

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/CompositeFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/CompositeFilesystem.java
@@ -16,6 +16,7 @@
 package io.agentscope.harness.agent.filesystem;
 
 import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.harness.agent.filesystem.local.LocalFilesystem;
 import io.agentscope.harness.agent.filesystem.local.LocalFilesystemWithShell;
 import io.agentscope.harness.agent.filesystem.model.EditResult;
 import io.agentscope.harness.agent.filesystem.model.FileDownloadResponse;
@@ -29,9 +30,11 @@ import io.agentscope.harness.agent.filesystem.model.ReadResult;
 import io.agentscope.harness.agent.filesystem.model.WriteResult;
 import io.agentscope.harness.agent.filesystem.sandbox.AbstractSandboxFilesystem;
 import java.nio.file.FileSystems;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.PathMatcher;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
@@ -60,11 +63,18 @@ import java.util.Map;
  * fs.read(RuntimeContext.empty(), "/memories/notes.md", 0, 100);  // → storeFs.read(...)
  * fs.read(RuntimeContext.empty(), "/src/Main.java", 0, 100);      // → localFs.read(...)
  * }</pre>
+ *
+ * <p>Absolute paths that fall under a host mount (extra {@link
+ * io.agentscope.harness.agent.workspace.PathPolicy} roots on a {@link LocalFilesystem} default
+ * backend, or roots passed to {@link #CompositeFilesystem(AbstractFilesystem, Map, Collection)})
+ * skip prefix routing. Otherwise {@code /knowledge/file.md} would match the {@code knowledge/}
+ * store after the leading slash is stripped, and a host file at that mount would never be read.
  */
 public class CompositeFilesystem implements AbstractFilesystem {
 
     private final AbstractFilesystem defaultBackend;
     private final List<RouteEntry> sortedRoutes;
+    private final List<Path> hostRoots;
 
     /**
      * Creates a composite filesystem with a default backend and prefix-based routes.
@@ -75,6 +85,21 @@ public class CompositeFilesystem implements AbstractFilesystem {
      */
     public CompositeFilesystem(
             AbstractFilesystem defaultBackend, Map<String, AbstractFilesystem> routes) {
+        this(defaultBackend, routes, null);
+    }
+
+    /**
+     * Same as {@link #CompositeFilesystem(AbstractFilesystem, Map)}, plus extra host-mount roots
+     * that must not be claimed by a same-named route prefix.
+     *
+     * @param hostRoots absolute directories that should stay on {@code defaultBackend}; {@code
+     *     null} entries are ignored. Roots already advertised on a {@link LocalFilesystem}
+     *     {@code PathPolicy} (other than the workspace cwd) are included automatically.
+     */
+    public CompositeFilesystem(
+            AbstractFilesystem defaultBackend,
+            Map<String, AbstractFilesystem> routes,
+            Collection<Path> hostRoots) {
         if (defaultBackend == null) {
             throw new IllegalArgumentException("defaultBackend must not be null");
         }
@@ -88,6 +113,7 @@ public class CompositeFilesystem implements AbstractFilesystem {
         }
         entries.sort(Comparator.comparingInt((RouteEntry e) -> e.prefix().length()).reversed());
         this.sortedRoutes = List.copyOf(entries);
+        this.hostRoots = resolveHostRoots(defaultBackend, hostRoots);
     }
 
     // ==================== Routing ====================
@@ -98,6 +124,9 @@ public class CompositeFilesystem implements AbstractFilesystem {
             AbstractFilesystem backend, String backendPath, String routePrefix) {}
 
     private RouteResult routeForPath(String path) {
+        if (isHostMountPath(path)) {
+            return new RouteResult(defaultBackend, path, null);
+        }
         // Canonicalize both sides by stripping any leading slash before matching so callers can
         // pass either "/skills/foo" (the {@link AbstractFilesystem} contract) or "skills/foo"
         // (the convention used by {@code WorkspaceManager.writeUtf8WorkspaceRelative}) and route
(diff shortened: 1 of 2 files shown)


oss-maintainer (commented):
﻿### Summary

The bug is real and the diagnosis in the javadoc is exactly right: `routeForPath` canonicalises both sides with `stripLeadingSlash`, so an absolute host path like `/knowledge/services.md` is indistinguishable from the workspace-relative `knowledge/services.md` and gets stolen by the `knowledge/` route. Short-circuiting on "this is an absolute path under a known host mount" before prefix matching is the correct place to fix it, the leading `/` is preserved when delegating to `defaultBackend`, and routing an invalid path never throws because `Path.of` is guarded. Existing relative-path behaviour is untouched, so this is not a breaking change for callers.

Verdict: **comment** — two things I could not resolve from the diff alone, both about whether the bypass is actually reachable where it is needed.

### Findings

- **[Warning]** `CompositeFilesystem.java:206` — the `instanceof LocalFilesystem` auto-detection does not match what production wiring passes in as `defaultBackend`; no caller in the repo uses the new 3-arg constructor either, so on the default paths `hostRoots` stays empty and behaviour is unchanged. Please see the inline comment for the specific call sites.
- **[Warning]** `CompositeFilesystem.java:185` — the `isAbsolute()` gate is platform-dependent, and `build (windows-latest)` on this head fails on precisely that: `CompositeFilesystemTest.absoluteHostMount_isNotHijackedBySameNamedRoute` reports `FAILURE!` at `CompositeFilesystemTest.java:254` (`Tests run: 5, Failures: 1`), i.e. the new test passes on Linux and fails on Windows. This is the strongest evidence that the check needs to be separator-/root-form aware rather than relying on `Path.isAbsolute()`.
- **[Info]** `CompositeFilesystem.java:232` — minor `List.contains` de-dup, ignore freely.

### Suggestions

For the Windows leg, the least invasive option is to treat a leading `/` as root-anchored before consulting `Path`, so the same absolute-path intent is recognised on both platforms:

```java
boolean rootAnchored = candidate.isAbsolute() || path.startsWith("/");
if (!rootAnchored) {
    return false;
}
```

If you would rather keep `isAbsolute()` as the only gate, then the test should be annotated to say so (`@DisabledOnOs(OS.WINDOWS)` plus a note in the class javadoc) — currently `build (windows-latest)` fails and it reads like an unrelated flake rather than a deliberate platform boundary.

### Cross-repo Note

Nothing here touches `agentscope-core` interfaces, so no cascade into `distribution`/`extensions` is expected.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/CompositeFilesystem.java
```diff
@@ -139,6 +168,72 @@ private static String stripLeadingSlash(String s) {
...
+    /**
+     * Collects host-mount directories that must bypass route prefixes: extra {@code PathPolicy}
+     * roots on a local default backend (excluding the workspace cwd) plus any caller-supplied
+     * roots.
+     */
+    private static List<Path> resolveHostRoots(
+            AbstractFilesystem defaultBackend, Collection<Path> extra) {
+        List<Path> roots = new ArrayList<>();
+        Path cwd = null;
+        if (defaultBackend instanceof LocalFilesystem local) {
```
oss-maintainer, line 206: `resolveHostRoots` only auto-derives roots when the default backend is a bare `LocalFilesystem`, but that is not what production wiring hands in, so the fix looks unreachable on the paths where the collision actually shows up:

- `HarnessAgent` (line 2485) wraps whatever `HarnessAgentBuilderSupport.resolveFilesystem` returned. In every default branch that is an `OverlayFilesystem` / `ProjectAwareOverlay` (`LocalFilesystemSpec.toFilesystem`) or an already-built `CompositeFilesystem` (`RemoteFilesystemSpec.toFilesystem`) — never a `LocalFilesystem`, so `hostRoots` stays empty.
- `RemoteFilesystemSpec` line 212 builds `new LocalFilesystem(workspace, false, 10, localNamespaceFactory)`, i.e. the legacy constructor that maps to `LocalFsMode.UNRESTRICTED` with `PathPolicy.empty()` — `getPathPolicy().roots()` is empty there too.
- The new 3-arg constructor has no non-test caller in the repo.

So the only configuration where this changes behaviour is a user-supplied `Builder.abstractFilesystem` that is itself a `LocalFilesystem` with a non-empty `PathPolicy`. Could the wiring be closed as part of this PR — unwrap `OverlayFilesystem.getUpper()` (and the `CompositeFilesystem` case) to find the underlying local backend, or pass the spec's `additionalRoots` explicitly at the two call sites? A test that goes through `HarnessAgent.Builder` with a host mount would also pin the regression at the level the bug was reported.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/CompositeFilesystem.java
```diff
@@ -139,6 +168,72 @@ private static String stripLeadingSlash(String s) {
...
+        if (hostRoots.isEmpty() || path == null || path.isBlank()) {
+            return false;
+        }
+        Path candidate;
+        try {
+            candidate = Path.of(path);
+        } catch (InvalidPathException e) {
+            return false;
+        }
+        if (!candidate.isAbsolute()) {
```
oss-maintainer, line 185: `isHostMountPath` gates the whole bypass on `candidate.isAbsolute()`, which is OS-dependent: on Windows `Path.of("/knowledge/services.md")` has no root component, so `isAbsolute()` returns `false` and the bypass never fires for the leading-slash form that `AbstractFilesystem` documents as its contract. The class javadoc and the fix itself both talk about `/knowledge/...`, so on `windows-latest` (which this repo's CI runs) the hijack would still be there — and the new test would not catch it.

Consider matching the POSIX-style form explicitly, e.g. treat a path starting with `/` as root-anchored regardless of `Path.isAbsolute()`, or compare the separator-normalised string against the host roots. If Windows is genuinely out of scope here, it is worth stating that in the javadoc so the next reader does not assume a cross-platform guarantee.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/CompositeFilesystem.java
```diff
@@ -139,6 +168,72 @@ private static String stripLeadingSlash(String s) {
...
+        }
+        Path normalized = root.toAbsolutePath().normalize();
+        if (cwd != null && (normalized.equals(cwd) || cwd.startsWith(normalized))) {
+            // Workspace itself, or a parent of the workspace, is not a colliding host mount.
+            return;
+        }
+        if (cwd != null && normalized.startsWith(cwd)) {
+            return;
+        }
+        if (!roots.contains(normalized)) {
```
oss-maintainer, line 232: Minor: `roots.contains(normalized)` makes `addHostRoot` quadratic. The list is short today, so this is not a hot-path concern — a `LinkedHashSet` accumulated in `resolveHostRoots` and returned as `List.copyOf(...)` expresses "unique, insertion-ordered" directly. Feel free to ignore.

PR #2910, opened 2026-08-30
fix(model): preserve thought signatures across JSON persistence round-trips

### LINKED ISSUE #2909: Thought signatures lost after JSON persistence round-trip cause 400 INVALID_ARGUMENT on Gemini 3 tool calls
## Description

Gemini 3 models require a `thought_signature` to be replayed together with each `functionCall` part. The Gemini parser in this repo already stores the signature correctly as `byte[]` in `ToolUseBlock` metadata (`ToolUseBlock.METADATA_THOUGHT_SIGNATURE`), and direct in-memory conversion works.

However, when agent state goes through a JSON persistence round-trip (any state store built on `JsonUtils` / Jackson: save session → load session), typed metadata values do not survive:

- `byte[]` → Jackson persists it as a **Base64 `String`**
- typed objects such as `OpenAIReasoningDetail` → restored as **`LinkedHashMap`**

The message converters only `instanceof`-check the original types, so the signature is silently dropped. On the next turn the request goes out without the signature and the Gemini API rejects the whole request:

```
400 INVALID_ARGUMENT
"Function call is missing a thought_signature in functionCall parts.
This is required for tools to work correctly for Gemini 3 models."
```

This breaks any multi-turn tool-calling loop that persists agent state between turns.

## Reproduction

```java
// 1. What the Gemini parser produces after a tool-call turn (works):
byte[] signature = ...; // from the model response
Map<String, Object> metadata = new HashMap<>();
metadata.put(ToolUseBlock.METADATA_THOUGHT_SIGNATURE, signature);
ToolUseBlock toolUse = ToolUseBlock.builder()
        .id("call_1").name("search").metadata(metadata).build();
Msg msg = Msg.builder().name("assistant")
        .content(List.of(toolUse)).role(MsgRole.ASSISTANT).build();

// 2. What every JSON state store does (save + load):
Msg restored = JsonUtils.getJsonCodec()
        .fromJson(JsonUtils.getJsonCodec().toJson(msg), Msg.class);
// metadata value is now a Base64 String, not byte[]

// 3. What the converter does on the next request:
List<Content> contents = new GeminiMessageConverter()
        .convertMessages(List.of(restored));
// contents.get(0).parts().get().get(0).thoughtSignature() -> Optional.empty
// signature is gone; the API then returns 400
```

Root cause locations (at 6059782):

- `GeminiMessageConverter` (~L129): only `signature instanceof byte[]` is attached; a Base64 `String` is ignored.
- `OpenAIMessageConverter` (~L369, ~L396): only `signature instanceof String` is accepted (so the in-memory `byte[]` cross-provider path is dropped too), and `"reasoningDetail"` metadata is only accepted as a typed `OpenAIReasoningDetail`, not as the `LinkedHashMap` that comes back from persistence.

## Live API verification

Replaying the second turn of a real tool-calling conversation against `gemini-3-flash-preview`, with the two request payloads identical except for the signature:

- **without** `thoughtSignature`: HTTP `400 INVALID_ARGUMENT` — "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly for Gemini 3 models."
- **with** `thoughtSignature`: HTTP `200`, model answers normally.

## Environment

- agentscope-java v2.0.3-SNAPSHOT (6059782)
- `agentscope-extensions-model-gemini` with google-genai SDK 1.45.0
- JDK 21 (Temurin), macOS
- Model: `gemini-3-flash-preview`

A fix proposal is ready — decode Base64 strings back to `byte[]` (and normalize `byte[]` to Base64 on the OpenAI side) in the converters.

chaucerj:
Fixes #2909

## Problem

Gemini 3 requires `thought_signature` to be replayed with each `functionCall`. The parsers store signatures as `byte[]` (and reasoning details as typed `OpenAIReasoningDetail`) in block metadata, but any JSON persistence round-trip (agent session save/load through `JsonUtils`/Jackson) restores them as a Base64 `String` / `LinkedHashMap`. The converters' `instanceof` checks silently dropped the values, so the next request went out without signatures and the API rejected it:

```
400 INVALID_ARGUMENT — "Function call is missing a thought_signature in
functionCall parts. This is required for tools to work correctly for
Gemini 3 models."
```

## Fix

**`GeminiMessageConverter`** — besides `byte[]`, accept a non-empty Base64 `String` signature and decode it back to `byte[]`; non-Base64 values are skipped with a warning instead of failing.

**`OpenAIMessageConverter`** — extract two small normalizers:

- `toSignatureString`: accepts both the Base64 `String` form and the in-memory `byte[]` form (the latter also fixes the cross-provider path where a Gemini/OpenRouter parser produced `byte[]` and the OpenAI converter dropped it entirely).
- `toReasoningDetail`: accepts the typed `OpenAIReasoningDetail` and the `LinkedHashMap` form restored from persistence, converting back via `JsonUtils`.

## Tests

New `@Tag("unit")` test classes:

- `GeminiThoughtSignaturePersistenceTest` — byte[] attached directly; signature restored after `JsonUtils` toJson/fromJson round-trip (asserts equal bytes); non-Base64 string ignored.
- `OpenAIThoughtSignatureTest` — byte[] signature converted to Base64; tool signature + `reasoningDetail` kept after round-trip; `ThinkingBlock` reasoning details kept after round-trip.

6/6 new tests pass; the existing `GeminiMessageConverterTest` + `OpenAIMessageConverterTest` suites (83 tests) still pass against the patched classes.

## Verification

Reproduced end-to-end against the live `gemini-3-flash-preview` API with two second-turn payloads identical except for the signature:

- without `thoughtSignature`: HTTP `400` — "Function call is missing a thought_signature in functionCall parts..."
- with `thoughtSignature`: HTTP `200`, normal answer.

Before this fix, the round-tripped state produced exactly the first (400) payload.


oss-maintainer (approved):
LGTM. fix(model): preserve thought signatures across JSON persistence round-trips — change looks correct and clean.

---

PR #2914, opened 2026-08-31
fix(openai): remove redundant reasoningDetail from ToolUseBlock metadata

### LINKED ISSUE #2913: [Bug]: OpenAIMessageConverter drops ThinkingBlock reasoning details when the message also contains tool calls (tool-call details overwrite them)
### What happened?

When an assistant `Msg` contains **both** a `ThinkingBlock` (with `METADATA_REASONING_DETAILS` in its metadata) **and** tool call blocks, `OpenAIMessageConverter.convertToMessage` builds the outgoing `reasoningDetails` field **twice** — and the second assignment **silently overwrites** the first. The reasoning details restored from the `ThinkingBlock` are dropped whenever the message also contains tool calls.

`OpenAIMessage.reasoningDetails` is a single list (default `null`), so the last `builder.reasoningDetails(...)` call wins.

### Code path

`convertAssistantMessage` in
`agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/main/java/io/agentscope/extensions/model/openai/formatter/OpenAIMessageConverter.java` (current `main`, f96cdc7):

Site 1 — ThinkingBlock path (~line 349):

```java
if (detailsObj instanceof List<?> list && !list.isEmpty()) {
    List<OpenAIReasoningDetail> details = new ArrayList<>();
    for (Object item : list) {
        if (item instanceof OpenAIReasoningDetail rd) {
            details.add(rd);
        }
    }
    if (!details.isEmpty()) {
        builder.reasoningDetails(details);   // <-- set #1
    }
}
```

Site 2 — tool-call path (~line 430):

```java
builder.toolCalls(toolCalls);

if (!reasoningDetails.isEmpty()) {
    builder.reasoningDetails(reasoningDetails);   // <-- set #2 overwrites set #1
}
```

Site 2 uses a locally-declared list (`List<OpenAIReasoningDetail> reasoningDetails = new ArrayList<>();` at ~line 363) that is never seeded with the details collected at site 1, so **any tool call carrying a `reasoningDetail` in its metadata replaces the whole thinking-side list**.

### Reproduction

```java
OpenAIReasoningDetail thinkingDetail = new OpenAIReasoningDetail();
thinkingDetail.setId("rs_thinking");
thinkingDetail.setData("thinking-side-signature");

Map<String, Object> thinkingMeta = new HashMap<>();
thinkingMeta.put(ThinkingBlock.METADATA_REASONING_DETAILS,
        new ArrayList<>(List.of(thinkingDetail)));
ThinkingBlock thinking =
        ThinkingBlock.builder().thinking("thinking...").metadata(thinkingMeta).build();

OpenAIReasoningDetail toolDetail = new OpenAIReasoningDetail();
toolDetail.setId("rs_tool");
toolDetail.setData("tool-side-signature");

Map<String, Object> toolMeta = new HashMap<>();
toolMeta.put("reasoningDetail", toolDetail);
ToolUseBlock toolUse =
        ToolUseBlock.builder().id("call_1").name("search").metadata(toolMeta).build();

Msg msg = Msg.builder()
        .name("assistant")
        .content(List.of(thinking, toolUse))
        .role(MsgRole.ASSISTANT)
        .build();

OpenAIMessage result = new OpenAIMessageConverter(m -> "", b -> "").convertToMessage(msg, false);
System.out.println(result.getReasoningDetails());
// prints: [rs_tool]  -- rs_thinking is gone
```

**Actual:** `reasoningDetails = [rs_tool]`
**Expected:** both details present, e.g. `[rs_thinking, rs_tool]`

### Impact

This is the standard shape of an assistant turn from reasoning models behind OpenAI-compatible endpoints (e.g. Gemini 3 via OpenRouter): one assistant message carries the reasoning/thinking block *and* the tool calls it decided on. For such providers the reasoning details carry thought signatures that the API requires to be replayed with subsequent turns — dropping the thinking-side entries silently degrades or breaks multi-turn tool-calling with interleaved thinking (the same class of failure as missing thought signatures on tool calls, which can surface as `400 INVALID_ARGUMENT` on replay).

Note the asymmetry: a message with only a `ThinkingBlock` keeps its details, and a message with only tool calls keeps its details — the loss only happens when both appear together, which is exactly the normal tool-calling turn.

### Suggested fix

Accumulate both sources into a single list and call `builder.reasoningDetails(...)` exactly once, after the tool-call section (falling back to the thinking-only list when there are no tool calls). Happy to submit a PR.

### Environment

- agentscope-java `main` (f96cdc7), module `agentscope-extensions-model-openai`
- Found while investigating #2909 (related converter area, independent defect)

chaucerj:
Fixes #2913


chaucerj:
The two red ubuntu runs are both unrelated to this change:

1. Run [33346235574](https://github.com/agentscope-ai/agentscope-java/actions/runs/33346235574) failed in `agentscope-harness` `JsonSessionDefaultLocationTest` — a `@TempDir` teardown `DirectoryNotEmptyException` (async session writer leaves `bob`/`s1` dirs behind); the assertions themselves passed.
2. Run [33346777378](https://github.com/agentscope-ai/agentscope-java/actions/runs/33346777378) fails to compile `agentscope-extensions-rag-simple` test sources: `apiKey(any())` is ambiguous because `main` now pins `openai-java` 4.52.0, which added `apiKey(Optional<String>)`. `main` itself is currently red for the same reason (run 33321234498) — fixed by #2912.

This branch only touches `OpenAIMessageConverter` + its test. Once #2912 lands I'll re-trigger.

chaucerj:
The latest failures on this PR are all the same flaky-test class in the **harness** module, unrelated to this change (which only touches `OpenAIMessageConverter` and its tests):

| Run | OS | Test | Error |
|---|---|---|---|
| [33356275214](https://github.com/agentscope-ai/agentscope-java/actions/runs/33356275214) | windows | `HarnessAgentDynamicHookBuilderTest.disableDynamicSkills_skillsEnabledFalseLeavesCatalogEmpty` | suppressed `DirectoryNotEmptyException` |
| [33357579563](https://github.com/agentscope-ai/agentscope-java/actions/runs/33357579563) | ubuntu | `HarnessAgentIntegrationExampleTest.example_fullWorkspace_singleTurn_seesSessionSubagentsAndWorkspaceContext` | suppressed `DirectoryNotEmptyException` (leftover `integration-session-1/`) |
| [33358003309](https://github.com/agentscope-ai/agentscope-java/actions/runs/33358003309) | ubuntu | `HarnessAgentDynamicHookBuilderTest.disableDynamicSkills_skillsEnabledFalseLeavesCatalogEmpty` | suppressed `DirectoryNotEmptyException` |

In every case: **assertions pass** (`Failures: 0`), and the error is thrown from JUnit `@TempDir` teardown as a *suppressed* `DirectoryNotEmptyException` — a background session writer appears to still hold/leave subdirectories when the temp dir is being cleaned up.

Evidence it's not this PR: the sibling PR #2910 (identical tree except this PR's 17-line converter change, which no harness test routes through) passed the same suite green on both OSes in run 33356276237.

I lack admin rights to re-run failed jobs, so I re-triggered via empty commits twice; the flake hit a third time. Maintainers with admin should be able to get it green with a simple *"Re-run failed jobs"*. Longer term it may be worth making the session writer await termination before test teardown.

oss-maintainer (approved):
LGTM. fix(openai): merge ThinkingBlock and tool reasoning details instead of overwriting — change looks correct and clean.

---

PR #2919, opened 2026-08-31
feat(agent): support structured output in streamEvents

### LINKED ISSUE #1717: 2.0版本streamEvents增加结构化输出参数
在1.0版本，流式输出可以定义结构化输出，在2.0，目前还不支持流式输出JSON结构化数据

Letter2025:
## Problem

The fine-grained streaming API (`ReActAgent#streamEvents(...)`) — the recommended replacement for the deprecated coarse-grained `stream(...)` — has **no structured-output overloads**. Only `call(msgs, schema/class, ctx)` (non-streaming) and the deprecated `stream(msgs, options, schema/class, ctx)` support `outputSchema`.

`buildAgentStream(...)` is private, so downstream code cannot inject a schema variant of `streamEvents` without forking the framework or relying on the deprecated API (which returns the legacy `io.agentscope.core.agent.Event` instead of the fine-grained `io.agentscope.core.event.AgentEvent`).

Closes #1717.

## Change

Add the missing fine-grained overloads, reusing the existing `buildAgentStream` core so streaming and blocking calls share identical lifecycle / middleware / hook semantics:

- `ReActAgent#streamEvents(List<Msg>, Class<?>, RuntimeContext)`
- `ReActAgent#streamEvents(List<Msg>, JsonNode, RuntimeContext)`
- `HarnessAgent#streamEvents(List<Msg>, Class<?>, RuntimeContext)`
- `HarnessAgent#streamEvents(List<Msg>, JsonNode, RuntimeContext)`

Each overload simply routes `m -> doCall(m, schema/class)` through `buildAgentStream`, exactly mirroring `call(msgs, schema/class, ctx)` → `callInternal(msgs, ctx, m -> doCall(m, schema/class))`. This means:

- **native path**: the model's `response_format` (strict `json_schema`) is preferred when `supportsNativeStructuredOutput()` is true; the JSON text surfaces as `TextBlockDeltaEvent`s.
- **fallback path**: when native is unsupported, the synthetic `generate_response` tool is injected and its `ToolCallStartEvent` / `ToolCallDeltaEvent` / tool-result events surface in the stream.
- the final `AgentResultEvent` carries a `Msg` with `hasStructuredData() == true` plus aggregated token `usage`, semantically identical to the blocking `call(...)` result.

## Tests

- `ReActAgentStreamEventsStructuredOutputTest` (core): class/JsonNode overloads, fallback tool events + final structured result + usage, native `TextBlockDeltaEvent` JSON deltas, and semantic equivalence with `call(...)`.
- `HarnessAgentStructuredOutputStreamEventsTest` (harness): both overloads delegate structured output through the harness sandbox-lifecycle wrapper.

## Notes

- No change to the deprecated `stream(...)` API; those overloads remain as-is.
- Interface-level placement of `streamEvents` is intentionally left for a follow-up (see #1894); this PR only adds the concrete `ReActAgent` / `HarnessAgent` overloads requested in #1717.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -1140,6 +1140,41 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
     }
 
     /**
+     * Stream fine-grained {@link AgentEvent}s with structured output driven by a target class.
+     *
+     * <p>Semantically equivalent to {@code call(List, Class, RuntimeContext)}: the model's native
+     * {@code response_format} path is preferred when supported, otherwise the synthetic
+     * {@code generate_response} tool is used as a fallback. The final {@link AgentResultEvent}
+     * carries a {@link Msg} whose {@code hasStructuredData()} is {@code true}.
+     *
+     * @param msgs input messages
+     * @param structuredOutputClass class defining the expected structure
+     * @param context runtime context to propagate into the call
+     * @return event stream covering the full agent invocation lifecycle
+     */
+    public Flux<AgentEvent> streamEvents(
+            List<Msg> msgs, Class<?> structuredOutputClass, RuntimeContext context) {
+        return buildAgentStream(msgs, context, m -> doCall(m, structuredOutputClass));
+    }
+
+    /**
+     * Stream fine-grained {@link AgentEvent}s with structured output driven by a JSON schema.
+     *
+     * <p>Semantically equivalent to {@code call(List, JsonNode, RuntimeContext)}: the model's
+     * native {@code response_format} path is preferred when supported, otherwise the synthetic
+     * {@code generate_response} tool is used as a fallback.
+     *
+     * @param msgs input messages
+     * @param outputSchema JSON schema defining the expected structure
+     * @param context runtime context to propagate into the call
+     * @return event stream covering the full agent invocation lifecycle
+     */
+    public Flux<AgentEvent> streamEvents(
+            List<Msg> msgs, JsonNode outputSchema, RuntimeContext context) {
+        return buildAgentStream(msgs, context, m -> doCall(m, outputSchema));
+    }
+
+    /**
      * Stream fine-grained {@link AgentEvent}s for a single input message with a caller-supplied
      * {@link RuntimeContext}.
      *
--- a/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentStreamEventsStructuredOutputTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentStreamEventsStructuredOutputTest.java
@@ -1,0 +1,283 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.agent;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.agentscope.core.ReActAgent;
+import io.agentscope.core.agent.test.MockModel;
+import io.agentscope.core.event.AgentEvent;
+import io.agentscope.core.event.AgentResultEvent;
+import io.agentscope.core.event.TextBlockDeltaEvent;
+import io.agentscope.core.event.ToolCallDeltaEvent;
+import io.agentscope.core.event.ToolCallStartEvent;
+import io.agentscope.core.message.Msg;
+import io.agentscope.core.message.MsgRole;
+import io.agentscope.core.message.TextBlock;
+import io.agentscope.core.message.ToolUseBlock;
+import io.agentscope.core.model.ChatResponse;
+import io.agentscope.core.model.ChatUsage;
+import io.agentscope.core.tool.Toolkit;
+import io.agentscope.core.util.JsonUtils;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Verifies that the fine-grained {@code streamEvents(...)} API supports structured output —
+ * mirroring the non-streaming {@code call(msgs, schema/class, ctx)} semantics — across both the
+ * native {@code response_format} path and the synthetic {@code generate_response} fallback path.
+ */
+class ReActAgentStreamEventsStructuredOutputTest {
+
+    private static final String NATIVE_JSON =
+            "{\"location\":\"San Francisco\",\"temperature\":\"72°F\",\"condition\":\"Sunny\"}";
+
+    private static final Map<String, Object> FALLBACK_TOOL_INPUT =
+            Map.of(
+                    "response",
+                    Map.of(
+                            "location", "San Francisco",
+                            "temperature", "72°F",
+                            "condition", "Sunny"));
+
+    static class WeatherResponse {
+        public String location;
+        public String temperature;
+        public String condition;
+    }
+
+    private static Msg userMsg() {
+        return Msg.builder()
+                .name("user")
+                .role(MsgRole.USER)
+                .content(TextBlock.builder().text("What's the weather in San Francisco?").build())
+                .build();
+    }
+
+    private static JsonNode weatherSchema() {
+        String json =
+                """
+                {
+                  "type": "object",
+                  "properties": {
+                    "location": {"type": "string"},
+                    "temperature": {"type": "string"},
+                    "condition": {"type": "string"}
+                  },
+                  "required": ["location", "temperature", "condition"],
+                  "additionalProperties": false
+                }
+                """;
+        return JsonUtils.getJsonCodec().fromJson(json, JsonNode.class);
+    }
+
+    /**
+     * A model that drives the fallback (synthetic {@code generate_response} tool) path: first call
+     * emits a {@code generate_response} tool use, second call emits a terminating text response.
+     */
+    private static MockModel fallbackModel() {
+        return new MockModel(
+                msgs -> {
+                    boolean hasToolResults =
+                            msgs.stream().anyMatch(m -> m.getRole() == MsgRole.TOOL);
+                    if (!hasToolResults) {
+                        return List.of(
+                                ChatResponse.builder()
+                                        .id("msg_1")
+                                        .content(
+                                                List.of(
+                                                        ToolUseBlock.builder()
+                                                                .id("call_123")
+                                                                .name("generate_response")
+                                                                .input(FALLBACK_TOOL_INPUT)
+                                                                .content(
+                                                                        JsonUtils.getJsonCodec()
+                                                                                .toJson(
+                                                                                        FALLBACK_TOOL_INPUT))
+                                                                .build()))
+                                        .usage(new ChatUsage(10, 20, 0.5))
+                                        .build());
+                    }
+                    return List.of(
+                            ChatResponse.builder()
+                                    .id("msg_2")
+                                    .content(List.of(TextBlock.builder().text("Done").build()))
+                                    .usage(new ChatUsage(5, 10, 0.1))
+                                    .build());
+                });
+    }
+
+    /**
+     * A model that advertises native structured output support and returns the structured JSON as
+     * plain text content.
+     */
+    private static MockModel nativeModel() {
+        return new MockModel(
+                msgs ->
+                        List.of(
+                                ChatResponse.builder()
+                                        .id("msg_native")
+                                        .content(
+                                                List.of(
+                                                        TextBlock.builder()
+                                                                .text(NATIVE_JSON)
+                                                                .build()))
+                                        .usage(new ChatUsage(10, 20, 0.5))
+                                        .build())) {
+            @Override
+            public boolean supportsNativeStructuredOutput() {
+                return true;
+            }
+        };
+    }
+
+    private static ReActAgent agent(MockModel model) {
+        return ReActAgent.builder()
+                .name("weather-agent")
+                .sysPrompt("You are a weather assistant")
+                .model(model)
+                .toolkit(new Toolkit())
+                .build();
+    }
+
+    private static AgentResultEvent lastResult(List<AgentEvent> events) {
+        return events.stream()
+                .filter(AgentResultEvent.class::isInstance)
+                .map(AgentResultEvent.class::cast)
+                .reduce((first, second) -> second)
+                .orElseThrow(() -> new AssertionError("No AgentResultEvent emitted"));
+    }
+
+    private static void assertWeather(Msg result) {
+        assertTrue(result.hasStructuredData(), "final result should carry structured data");
+        WeatherResponse weather = result.getStructuredData(WeatherResponse.class);
+        assertNotNull(weather);
+        assertEquals("San Francisco", weather.location);
+        assertEquals("72°F", weather.temperature);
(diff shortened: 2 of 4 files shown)


dailingtao (approved):
LGTM

oss-maintainer (approved):
LGTM. feat(agent): support structured output in streamEvents — change looks correct and clean.

---

dargoner:
跟进一下这个 PR 的状态，顺带把 API 形状的问题摆到台面上。

现状核对：1 commit / 4 files，两个人工 approve（dailingtao、oss-maintainer），CI 5/5 通过，自 8-31 提交后没有新动作；分支落后 main 53 个提交。

同时 issue #1717（2.0 版本 streamEvents 增加结构化输出参数）是同一个 API 面诉求，目前两处并行挂着。建议维护者在这条 PR 上定个调子：

1. 形状以哪一处为准：ReActAgent.streamEvents(List<Msg>, Class<?>, RuntimeContext) 与 (List<Msg>, JsonNode, RuntimeContext) 是否就是期望的签名；
2. 是否希望补齐 Msg / String 形式的便捷重载（下游实际使用中这两类重载的调用频率更高，只有 List 形式的话上层还要自己包一层）；
3. core 与 harness 是否统一（HarnessAgent 侧同样需要，否则 harness 用户拿不到这个能力）。

之所以追问形状：我们下游已经按同样语义实现了 core + harness 的这套重载（含 Msg / String 便捷形式），如果官方签名定下来，我们可以直接对齐，避免出现「官方一套、下游一套」的长期分裂；#1717 里的提问者也能得到一个明确答复。

如果需要先 rebase 到最新 main 才能进入合并队列，也请说一下。

Letter2025:
Thanks @dargoner for the thorough follow-up — all three questions are addressed in the latest push (`ce329fe6`, rebased onto `main` @ `283f216d`):

1. **API shape** — `streamEvents(List<Msg>, Class<?>, RuntimeContext)` and `streamEvents(List<Msg>, JsonNode, RuntimeContext)` are the canonical shapes, mirroring the existing `call(msgs, class/schema, ctx)` semantics (native `response_format` first, synthetic `generate_response` fallback, final `AgentResultEvent` carrying validated structured data + usage). On top of that, the `Msg` / `String` convenience overloads you mentioned are now added — on **both** `ReActAgent` and `HarnessAgent` — so downstream callers no longer need to wrap into a `List` themselves (`streamEvents(msg|text, class/schema, ctx)`).

2. **core + harness alignment** — both layers now expose the full overload set through the same `buildAgentStream` / `wrappedStreamEvents` cores as their non-structured siblings, with identical sandbox-lifecycle semantics.

3. **Rebase** — rebased onto latest `main` (`283f216d`), no conflicts, spotless clean; targeted core (5 tests) and harness (3 tests) suites pass locally, CI is running on the new commit.

@oss-maintainer PTAL when you get a chance — thanks for the earlier approval.

（中文小结：形状以 `List + Class/JsonNode + RuntimeContext` 为准、与 `call()` 对齐；已补上 `Msg`/`String` 四组便捷重载，core 与 harness 两层都齐；已 rebase 到最新 main 无冲突，扩展测试本地全绿。）

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -925,6 +986,40 @@ public Flux<AgentEvent> streamEvents(List<Msg> msgs, RuntimeContext ctx) {
...
+     * caller-supplied {@link RuntimeContext}. Mirrors {@code ReActAgent#streamEvents(List, Class,
+     * RuntimeContext)} with the same sandbox-lifecycle acquire/release semantics.
+     *
+     * @param msgs input messages
+     * @param structuredModel class defining the expected structure
+     * @param ctx runtime context to propagate into the call
+     * @return event stream covering the full agent invocation lifecycle
+     */
+    public Flux<AgentEvent> streamEvents(
+            List<Msg> msgs, Class<?> structuredModel, RuntimeContext ctx) {
```
oss-maintainer, line 1004: **[Warning]** The Javadoc here promises "the same sandbox-lifecycle acquire/release semantics" as the plain overload, and the delegation (`ensureSessionDefaults` + `wrappedStreamEvents`) does read identical to `streamEvents(List, RuntimeContext)` — but nothing pins that promise. The three new harness tests all build the agent with `LocalFilesystem`, so the acquire/release path is never entered. A regression that routes the structured overloads around `wrappedStreamEvents` (i.e. leaks a sandbox lease per structured streaming call under a real sandbox backend) would still leave this file green. Worth one test against the sandbox-backed harness that asserts acquire+release exactly once per subscription — `HarnessAgentSubagentStreamEventsTest` shows the pattern.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1140,6 +1140,41 @@ public Flux<AgentEvent> streamEvents(List<Msg> msgs, RuntimeContext context) {
...
+     * {@code response_format} path is preferred when supported, otherwise the synthetic
+     * {@code generate_response} tool is used as a fallback. The final {@link AgentResultEvent}
+     * carries a {@link Msg} whose {@code hasStructuredData()} is {@code true}.
+     *
+     * @param msgs input messages
+     * @param structuredOutputClass class defining the expected structure
+     * @param context runtime context to propagate into the call
+     * @return event stream covering the full agent invocation lifecycle
+     */
+    public Flux<AgentEvent> streamEvents(
```
oss-maintainer, line 1184: **[Info]** The plain family offers context-less entries (`streamEvents(List<Msg>)` at :1114, `streamEvents(Msg)` at :1124, `streamEvents(String)` at :1223), but every structured variant requires an explicit `RuntimeContext`. `agent.streamEvents(msgs, WeatherResponse.class)` therefore does not compile while `agent.streamEvents(msgs)` does — a small asymmetry users will trip over. Either add the two ctx-less overloads (delegating with `null`, exactly like :1114 does) or call the requirement out in the class-level docs, so "mirror the call() family" is true in shape and not only in semantics.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1140,6 +1140,41 @@ public Flux<AgentEvent> streamEvents(List<Msg> msgs, RuntimeContext context) {
...
+     * <p>Semantically equivalent to {@code call(List, JsonNode, RuntimeContext)}: the model's
+     * native {@code response_format} path is preferred when supported, otherwise the synthetic
+     * {@code generate_response} tool is used as a fallback.
+     *
+     * @param msgs input messages
+     * @param outputSchema JSON schema defining the expected structure
+     * @param context runtime context to propagate into the call
+     * @return event stream covering the full agent invocation lifecycle
+     */
+    public Flux<AgentEvent> streamEvents(
```
oss-maintainer, line 1205: **[Info]** `streamEvents(List, Class<?>, RuntimeContext)` and `streamEvents(List, JsonNode, RuntimeContext)` make the second argument genuinely ambiguous for a `null` literal: `streamEvents(msgs, null, ctx)` is a compile error (reference ambiguous) rather than a clean failure. `doStructuredCall` already returns `Mono.error(IllegalArgumentException)` for "both null", so the runtime guard exists; a short Javadoc note (`pass an explicit (Class<?>) null cast, or use the plain overload`) plus one test asserting the error when a casted-null reaches `doStructuredCall` would make the intent explicit. Not blocking — `call()` at :851/:855 has the same shape, so this is consistent-by-inheritance.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentStreamEventsStructuredOutputTest.java
```diff
@@ -0,0 +1,326 @@
...
+                        .block(Duration.ofSeconds(10));
+        assertNotNull(msgClassEvents);
+        assertWeather(lastResult(msgClassEvents).getResult());
+
+        List<AgentEvent> msgSchemaEvents =
+                agent.streamEvents(userMsg(), weatherSchema(), RuntimeContext.empty())
+                        .collectList()
+                        .block(Duration.ofSeconds(10));
+        assertNotNull(msgSchemaEvents);
+        assertEquals(
```
oss-maintainer, line 301: **[Info]** Nice to see `streamEventsStructuredResultMatchesCallResult` — asserting the streamed `AgentResultEvent` and the `call()` result agree is the check that actually protects the "semantically equivalent to call()" claim in the Javadoc. Extending that same equivalence assertion to the *native* path (currently only the fallback path is cross-checked against `call()`) would close the loop, since the native branch is the one where `wrapNativeStructuredResult` post-processes the `Msg` differently.

oss-maintainer (approved):
### Summary
Re-reviewed after the rebase + the new convenience overloads (`f2aca34b` → `ce329fe6`). The change is a faithful mirror of the `call()` structured family: `streamEvents(msgs, Class|JsonNode, ctx)` routes through the same `buildAgentStream(msgs, context, m -> doCall(m, ...))` core that `callInternal()` uses, so "semantically equivalent to `call()`" holds by construction, and the harness side keeps `ensureSessionDefaults` + `wrappedStreamEvents` identical to the plain overload. 600 lines of tests covering both the native `response_format` path and the synthetic `generate_response` fallback, including a `streamEvents` vs `call` equivalence check. CLA signed, CI green. **Approving** — nothing below is blocking.

### Findings
- **[Warning]** `HarnessAgent.java:1000` — the "same sandbox-lifecycle acquire/release semantics" claim is not pinned by any test; all three new harness cases use `LocalFilesystem`, so a regression that bypassed `wrappedStreamEvents` would still be green.
- **[Info]** `ReActAgent.java:1156` — the plain family has context-less entries (`streamEvents(List)`, `(Msg)`, `(String)`), every structured variant requires a `RuntimeContext`; small asymmetry users will trip over.
- **[Info]** `ReActAgent.java:1173` — `Class<?>` vs `JsonNode` in the same position makes `streamEvents(msgs, null, ctx)` an ambiguous-call compile error; the runtime guard exists in `doStructuredCall`, a Javadoc note would help.
- **[Info]** `ReActAgentStreamEventsStructuredOutputTest.java:300` — the `call()` equivalence assertion currently covers the fallback path; extending it to the native path closes the loop, since `wrapNativeStructuredResult` post-processes the `Msg` only there.

### Suggestions
One sandbox-backed harness test asserting acquire+release exactly once per subscription for a structured overload is the highest-value addition; `HarnessAgentSubagentStreamEventsTest` already shows the shape.

---

Letter2025:
Thanks for the detailed notes @oss-maintainer — all four addressed in `1c9caff8`:

1. **[Warning] sandbox acquire/release not pinned** — added `sandboxBackedStreamEvents_acquiresAndReleasesExactlyOncePerSubscription` in `HarnessAgentStructuredOutputStreamEventsTest`: a sandbox-backed HarnessAgent (`InMemorySandboxFilesystemSpec`) runs two structured streams against the same session and asserts `InMemorySandboxClient.createCount == 1` then `resumeCount == 1` — i.e. acquire exactly once per subscription, and release persists state exactly once (a regression that routes the structured overloads around `wrappedStreamEvents` would never enter the sandbox path and fails the create/resume counts).
2. **[Info] context-less asymmetry** — added `streamEvents(List, Class<?>)` and `streamEvents(List, JsonNode)` on `ReActAgent`, delegating with a null context exactly like the plain `streamEvents(List)` entry; covered by `contextlessClassAndSchemaOverloadsYieldStructuredResults`.
3. **[Info] null-literal ambiguity** — Javadoc note added to both 3-arg structured overloads (bare `null` is ambiguous between the `Class`/`JsonNode` siblings; cast explicitly or use the plain overload), plus `castedNullStructuredArgumentFailsFastClearly` asserting a casted-null surfaces the existing `IllegalArgumentException` from `doStructuredCall` when nothing is provided.
4. **[Info] native-path equivalence** — added `streamEventsNativePathMatchesCallResult`, cross-checking the `AgentResultEvent` against `call(...)` on the native `response_format` path (where `wrapNativeStructuredResult` post-processes the `Msg`), mirroring the existing fallback-path equivalence test.

Local: core `ReActAgentStreamEventsStructuredOutputTest` 8/8 and harness `HarnessAgentStructuredOutputStreamEventsTest` 4/4 green, spotless clean. @oss-maintainer PTAL when convenient — thanks again for the review.

oss-maintainer (approved):
### Summary
Delta re-review of `ce329fe6` → `1c9caff8` — the commit carrying my four non-blocking items from the previous approval. All four are addressed, and the fixes are the right kind: two of them are tests that pin behaviour previously asserted only in prose, so the claim cannot rot silently.

- **Sandbox acquire/release** — `sandboxBackedStreamEvents_acquiresAndReleasesExactlyOncePerSubscription` drives a real `InMemorySandboxFilesystemSpec` and asserts exactly one create on the first subscription and exactly one resume on the second, i.e. `wrappedStreamEvents`' acquire/release is now pinned per subscription instead of being inherited by analogy from the plain overload. That was my main item; it is closed.
- **Casted-null `structuredOutputClass` / `outputSchema`** — javadoc on both 3-arg overloads now names the ambiguity and points at the explicit cast or the plain overload, and `castedNullStructuredArgumentFailsFastClearly` pins the exception message (`targetClass or schemaDesc`) rather than merely that it throws.
- **Context-less form** — `streamEvents(msgs, Class)` / `streamEvents(msgs, JsonNode)` delegate to the 3-arg version with `(RuntimeContext) null`, so they cannot diverge from the default-context path, and `contextlessClassAndSchemaOverloadsYieldStructuredResults` covers both.
- **Native-vs-`call` equivalence** — `streamEventsNativePathMatchesCallResult` compares the `streamEvents(...)` and `call(...)` structured payloads on the same fixture model, which is the check the "semantically equivalent to `call()`" claim actually needs.

### Findings
- **[Info]** `ReActAgent.java:1126` / `:1138` — the context-less structured pair exists only on `ReActAgent`. `HarnessAgent` offers `streamEvents(msgs, Class, ctx)` (:999) and `(msgs, JsonNode, ctx)` (:1017) but no `(msgs, Class)` / `(msgs, JsonNode)`, so callers who start from `HarnessAgent` — the documented entry point — lose exactly the convenience this commit adds, and the asymmetry reads as a missing feature rather than a choice. Mirroring the two delegating one-liners on the harness side, or noting the difference in the javadoc, keeps the two surfaces consistent. Not blocking.

### Verification
The delta is confined to `ReActAgent.java` (+32 lines: javadoc plus two delegating overloads) and the two test classes; no change on the already-approved path, so the earlier reasoning still holds. CLA signed (`license/cla=success` on `1c9caff8`). CI on the new head was still `PENDING` when this review was written (build ubuntu/windows, validate, license) — this approves the code, not the pipeline; if a check goes red the fix is the author's and I will re-check on the next sweep.

---

Letter2025:
Thanks — addressed the remaining [Info] with `edaed31e`.

On the harness side I went with your second option (javadoc note) rather than mirroring the context-less pair: `HarnessAgent`'s existing context-less plain `streamEvents` entries (`streamEvents(Msg)`, `streamEvents(List<Msg>)`, `streamEvents(String)`) are all `@Deprecated(since = "2.2.0")` in favor of an explicit `RuntimeContext`, so adding new non-deprecated context-less structured overloads would fight that API direction. Both 3-arg structured overloads now carry the note recording the deliberate asymmetry vs `ReActAgent`.

Everything else on this PR stands as previously reviewed: CI on the current head is running (Module Sync + License already green). @oss-maintainer thanks again for the thorough passes.

Letter2025:
Heads-up on the red Windows check on `edaed31e` — it's a pre-existing flake, unrelated to the diff:

- The only failure is `LocalFilesystemPersonalAssistantExampleTest.localFilesystem_filesPersistAcrossCalls` (`MEMORY.md content should be unchanged after call 2`) — 1043 of 1044 harness tests passed, and every test added by this PR passed.
- The `edaed31e` diff is javadoc-only (8 insertions, no executable code).
- The same class (including that exact test) passes **3/3 locally on Windows**.
- The failure mode is the familiar harness background memory-flush race: the flush/maintenance scheduler writes `MEMORY.md` asynchronously after the call completes, so an immediate "unchanged" assertion can catch a late write. We've reported the same family of flakes before on other PRs (e.g. `Failed to close extension context` at `@TempDir` teardown on unrelated main runs).

Since the fork account cannot rerun the workflow on this repo, I pushed `d0363392` — an empty commit with no file changes — purely to trigger a fresh CI run. @oss-maintainer if the rerun comes back green (as expected), this should be ready to merge.

oss-maintainer (approved):
### Summary
Delta since my previous approval is documentation-only (+8 lines of javadoc on `HarnessAgent#streamEvents(..., structuredModel, ctx)` / `(…, schema, ctx)` explaining why the context-less structured overloads are deliberately absent) plus a CI re-run commit with no code change. Nothing behavioral — the earlier approval stands on the new head. CLA signed, `MERGEABLE/CLEAN`.

---

PR #2925, opened 2026-09-01
fix(harness): preserve Docker exec commands on Windows

### LINKED ISSUE #2924: [Bug]:DockerSandbox.doExec() strips nested double quotes on Windows
**Describe the bug**

On Windows, `DockerSandbox.doExec()` fails for any shell command containing nested double quotes (e.g. `mkdir -p "$(dirname ...)"`, `python3 -c "..."`). This breaks `write_file` and `edit_file` tools entirely.

Root cause: `ProcessBuilder` on Windows goes through `cmd.exe /c`, which strips inner double quotes from the `sh -c` argument. Linux/macOS are unaffected (no `cmd.exe`).

**To Reproduce**

```java
// Exact command from BaseSandboxFilesystem.write() line ~165
String cmd = "if [ -e 'test.txt' ]; then echo 'EXISTS'; exit 1; fi; "
    + "mkdir -p \"$(dirname 'test.txt')\" 2>&1";

// Current doExec() approach
ProcessBuilder pb = new ProcessBuilder(
    "docker", "exec", "-w", "/workspace", containerId, "sh", "-c", cmd);
Process p = pb.start();
p.waitFor(30, TimeUnit.SECONDS);
System.out.println(p.exitValue()); // Windows: 2, Linux: 0
```

Run on Windows 11 + Docker Desktop → exit code 2, stderr: `'test.txt') 2>&1: 1: Syntax error: end of file unexpected (expecting ")")`

**Expected behavior**

Exit code 0 on all platforms.

**Error messages**

Windows exit code 2, stderr:

```
'test.txt') 2>&1: 1: Syntax error: end of file unexpected (expecting ")")
```

`edit_file` (requires python3 image) stderr:

```
File "<string>", line 1
    import
          ^
SyntaxError: invalid syntax
```

**Environment:**

- Java Version: 21
- OS: Windows 11
- Docker: Docker Desktop with Linux containers

**Additional context**

- `doHydrateWorkspace()` already uses `docker exec -i` (for tar stdin) and is unaffected.
- Only `doExec()` is affected because it's the only method that passes a shell command via `-c`.
- Suggested fix: pipe command through stdin instead of `-c` argument (see reproduction code above).

guslegend0510:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #2924

`DockerSandbox.doExec()` previously passed the complete shell command as the final `docker.exe` argument (`sh -c <command>`). On Windows, nested double quotes can be altered while the Win32 command line is serialized and parsed, breaking filesystem commands such as `write_file` and `edit_file`.

This PR:

- runs `docker exec -i ... sh -s` and streams the command through UTF-8 stdin instead of placing it in argv;
- wraps the streamed command so command-level stdin remains isolated while heredocs and explicit redirections continue to work;
- preserves existing stdout/stderr capture, output truncation, non-zero exit, and timeout behavior;
- adds regression tests for nested quotes, Unicode, multiline heredocs, non-zero exits, timeouts, and output truncation.

## Testing

- `mvn -pl agentscope-harness -am -Dtest=DockerSandboxExecTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -pl agentscope-harness -am test`
- `mvn spotless:check`
- `git diff --check`
- Docker Linux-container verification covering the Issue #2924 reproduction, file write, Python heredoc edit, and stdin isolation

## Checklist

- [x] Related documentation is not required; the implementation Javadoc was updated


agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
```diff
@@ -173,15 +173,29 @@ protected ExecResult doExec(RuntimeContext runtimeContext, String command, int t
...
         Future<String> stderrFuture =
-                drainer.submit(() -> readStream(process.getErrorStream(), OUTPUT_TRUNCATE_BYTES));
-        drainer.shutdown();
+                ioExecutor.submit(
+                        () -> readStream(process.getErrorStream(), OUTPUT_TRUNCATE_BYTES));
+        Future<?> stdinFuture =
+                ioExecutor.submit(
+                        () -> {
+                            try (OutputStream stdin = process.getOutputStream()) {
+                                stdin.write(
```
oss-maintainer, line 185: The stdin writer thread performs blocking I/O on a real `OutputStream` which is not interruptible by `Thread.interrupt()`. If the pipe buffer fills (unlikely for typical sandbox commands, but possible for very large generated scripts) and the process stalls, `stdinFuture.cancel(true)` in the timeout path will not unblock the writer. Consider adding a deadline to `stdinFuture.get()` in the success path (e.g. `stdinFuture.get(5, TimeUnit.SECONDS)`) to avoid an indefinite hang if the writer is stuck, or use a socket-based stream that supports async close.

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
```diff
@@ -194,11 +208,48 @@ protected ExecResult doExec(RuntimeContext runtimeContext, String command, int t
...
+    Process startExecProcess(List<String> command) throws IOException {
+        return new ProcessBuilder(command).start();
+    }
+
+    /**
+     * Wraps the command in a compound command before streaming it to {@code sh -s}.
+     *
+     * <p>The shell parses the complete compound command before applying the {@code /dev/null}
+     * redirection. Commands such as {@code read} therefore cannot consume the remaining script,
+     * while heredocs and explicit stdin redirections continue to work.
```
oss-maintainer, line 228: The `{...} </dev/null` wrapper intentionally severs stdin for inner commands to prevent them from consuming the script text fed to `sh -s`. This is a behavioral change from the old `sh -c` approach: commands like bare `cat` or `read` that previously would hang waiting for stdin will now silently receive EOF from `/dev/null`. Consider documenting this constraint in the class-level Javadoc or the `wrapCommandForStdin` Javadoc so callers are aware.

oss-maintainer (commented):
### Summary
Re-review of PR #2925 at its current head (automated review of the changes in this diff). 2 item(s) raised below; nothing here blocks the approach itself — details inline.

### Findings
- **[Warning]** `agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java:185` — The stdin writer thread performs blocking I/O on a real `OutputStream` which is not interruptible by `Thread.interrupt()`. If the pipe buffer fills (unlikely for typical sandbox commands, but possible for very large generated scripts) and the process stalls, `stdinFuture.cancel(true)` in the timeout path will not unblock the writer. Consider adding a deadline to `stdinFuture.get()` in the success path (e.g. `stdinFuture.get(5, TimeUnit.SECONDS)`) to avoid an indefinite hang if the writer is stuck, or use a socket-based stream that supports async close.
- **[Info]** `agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java:228` — The `{...} </dev/null` wrapper intentionally severs stdin for inner commands to prevent them from consuming the script text fed to `sh -s`. This is a behavioral change from the old `sh -c` approach: commands like bare `cat` or `read` that previously would hang waiting for stdin will now silently receive EOF from `/dev/null`. Consider documenting this constraint in the class-level Javadoc or the `wrapCommandForStdin` Javadoc so callers are aware.

### Notes

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
```diff
@@ -194,11 +208,48 @@ protected ExecResult doExec(RuntimeContext runtimeContext, String command, int t
                         || stderr.length() >= OUTPUT_TRUNCATE_BYTES;
         ExecResult result = new ExecResult(exitCode, stdout, stderr, truncated);
         if (!result.ok()) {
+            awaitWriterIgnoringFailure(stdinFuture);
             throw new SandboxException.ExecException(exitCode, stdout, stderr);
         }
+        stdinFuture.get();
```
oss-maintainer, line 214: This unconditional `get()` turns a *writer* failure into an exec failure. If the child exits before draining the script (a large command, or one that ends early), the write fails with EPIPE, the `Future` completes exceptionally, and a command that exited 0 is reported as an exception instead of an `ExecResult`. The non-zero path already calls `awaitWriterIgnoringFailure(stdinFuture)` for exactly this reason — consider the same treatment here, since the exit code is the authoritative signal.

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandbox.java
```diff
@@ -194,11 +208,48 @@ protected ExecResult doExec(RuntimeContext runtimeContext, String command, int t
...
+    /**
+     * Wraps the command in a compound command before streaming it to {@code sh -s}.
+     *
+     * <p>The shell parses the complete compound command before applying the {@code /dev/null}
+     * redirection. Commands such as {@code read} therefore cannot consume the remaining script,
+     * while heredocs and explicit stdin redirections continue to work.
+     */
+    static String wrapCommandForStdin(String command) {
+        String trailingNewline = command.endsWith("\n") ? "" : "\n";
+        return "{\n:\n" + command + trailingNewline + "} </dev/null\n";
```
oss-maintainer, line 232: Worth a note in the javadoc that the wrapper also means the *inner* commands never see host stdin (it is `/dev/null` for the whole group). Today nothing pipes data in from the host so this is a feature, but it is the kind of invariant that a future `execWithStdin` caller would trip over.

agentscope-harness/src/test/java/io/agentscope/harness/agent/sandbox/impl/docker/DockerSandboxExecTest.java
```diff
@@ -0,0 +1,226 @@
...
+                        () -> sandbox.execute("echo broken >&2; exit 7", 30));
+
+        assertEquals(7, error.getExitCode());
+        assertEquals("partial", error.getStdout());
+        assertEquals("failed", error.getStderr());
+        assertTrue(process.stdin.closed);
+    }
+
+    @Test
+    void destroysProcessAndClosesStdinOnTimeout() {
```
oss-maintainer, line 94: Nice seam with `startExecProcess`. One gap: no case where the process exits while the writer still has bytes pending (the EPIPE scenario above). A `FakeProcess` whose stdin sink rejects writes after `exitValue()` becomes available would pin down whatever behaviour you decide on here.

oss-maintainer (commented):
### Summary
Streaming the script through `sh -s` instead of `sh -c <argv>` is the right fix for Windows argv mangling, and wrapping it in `{ : ...; } </dev/null` is a neat way to stop an inner `read` from swallowing the rest of the script while heredocs and explicit redirects keep working. Tests are thorough (nested quotes/unicode/heredoc byte-fidelity/exit-code/timeout/truncation).

### Findings
- **[Warning]** `stdinFuture.get()` on the success path can convert a successful exec into a thrown exception on EPIPE.
- **[Info]** the group redirect also denies host stdin to inner commands — document the invariant.
- **[Info]** no test for "process exits while the script is still being written".

### Verdict
Nothing here is architecturally wrong; the [Warning] is a small robustness fix. Not approving in this round because the PR still has a merge conflict with `main` — rebase and ping for a re-review.

---

PR #2935, opened 2026-09-02
test(harness): drain fire-and-forget memory flush before @TempDir teardown to stop flaky temp-dir deletion

jujn:
### Problem

Two CI runs fail intermittently with JUnit errors that are not test-logic failures but teardown failures:

- `Failed to delete temp directory` (Windows, PR #2923)
- `Failed to close extension context` / `Failed to delete temp directory` (Ubuntu, PR #2926)

Both surface only when a test builds a transient `HarnessAgent`, drives it to completion via `.block()` / `.stream()...block()`, and uses `@TempDir` for its workspace/state home.

### Root cause

This is a long-standing race, not a regression of any single commit. The harness memory flush has always been asynchronous: when an agent stream completes, `MemoryFlushMiddleware#onAgent` dispatches the flush on `Schedulers.boundedElastic()` via `subscribe(...)` — i.e. fire-and-forget. The calling test's `.block()` only waits for the business stream, **not** for that background flush.

So the timeline is:

1. Test calls `.block()` and returns.
2. JUnit begins `@TempDir` teardown and deletes the temp directory.
3. The async flush on `boundedElastic` is **still writing session/transcript mirror files into that same `@TempDir`** (or still holds open file handles).

The result: directory/file deletion fails with `IOException` → wrapped as `JUnitException`. It is timing-dependent (depends on IO speed, scheduler, and how many files are written), which is why it flakes rather than failing deterministically, and why Windows (stricter file locking) fails more readily than Linux.

The normal production path avoids this because `HarnessAgent#close()` calls `SessionTree.awaitMirrorQuiescence(...)` + `MemoryBackgroundTasks.awaitQuiescence(...)`. The flaky tests never call `close()` on their transient agent.

### Fix

Add test-side quiescence that mirrors what `HarnessAgent#close()` already does, run **after each test method but before the `TempDir` extension deletes the directory**:

- `HarnessBackgroundTaskQuiescenceExtension` — an `AfterEachCallback` that calls `SessionTree.awaitMirrorQuiescence(5s)` + `MemoryBackgroundTasks.awaitQuiescence(5s)`. When nothing is in flight both calls return immediately, making it a no-op for tests that never trigger a flush.
- `@HarnessQuiescence` — a composed meta-annotation (`@ExtendWith(HarnessBackgroundTaskQuiescenceExtension.class)`) so at-risk tests only need a one-line annotation.
- Applied `@HarnessQuiescence` to 16 harness test classes that match the at-risk pattern (`HarnessAgent.builder()` + `.call()`/`.stream()` + `@TempDir`), including the three classes that flaked in CI: `JsonSessionDefaultLocationTest`, `HarnessAgentIntegrationExampleTest`, `HarnessAgentDynamicHookBuilderTest`.

Production code is unchanged — the flush remains fire-and-forget so conversation completion is never blocked.

### DIFF
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentDistributedSandboxTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/HarnessAgentDistributedSandboxTest.java
@@ -31,6 +31,7 @@ import io.agentscope.harness.agent.filesystem.remote.store.BaseStore;
 import io.agentscope.harness.agent.filesystem.spec.RemoteFilesystemSpec;
 import io.agentscope.harness.agent.sandbox.impl.docker.DockerFilesystemSpec;
 import io.agentscope.harness.agent.sandbox.snapshot.LocalSnapshotSpec;
+import io.agentscope.harness.agent.testing.HarnessQuiescence;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
@@ -38,6 +39,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 import reactor.core.publisher.Flux;
 
+@HarnessQuiescence
 class HarnessAgentDistributedSandboxTest {
 
     @TempDir Path workspace;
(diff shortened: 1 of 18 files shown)


Buktal (approved):
LGTM

PR #2936, opened 2026-09-02
fix(harness): drain local shell output while process runs

### LINKED ISSUE #2838: [Bug]: LocalFilesystemWithShell.execute deadlocks when child process output exceeds pipe buffer (waitFor before draining stdout/stderr)

CryoThrust:
Fixes #2838.

`LocalFilesystemWithShell.execute` waited for the process before consuming stdout and stderr. Commands producing more than the OS pipe buffer could deadlock and be reported as timeouts.

This change starts daemon stream readers immediately after process startup, waits for the process with the existing timeout, and then collects both outputs. It preserves output truncation and timeout response behavior and adds a regression test covering approximately 108 KB of stdout.

Signed-off-by: Yohanes <CryoThrust@users.noreply.github.com>


CryoThrust:
Added focused regression coverage for the previously uncovered paths: large stderr output and command timeout handling. The tests use platform-native commands and assert both the timeout exit code and truncation behavior. The commit is now pushed to the PR branch; CI is running.

oss-maintainer:
**CLA Not Signed**

@CryoThrust please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.

---

oss-maintainer (commented):
### Summary
This PR fixes the pipe-buffer deadlock in `LocalFilesystemWithShell.execute` by draining stdout/stderr via daemon reader threads while the process runs, but it introduces a few correctness and resource-management issues that should be addressed before merging.

### Findings
- **[warning]** `LocalFilesystemWithShell.java:451` — `ExecutionException` is swallowed alongside `TimeoutException`, so reader I/O errors can be reported as successful output.
- **[warning]** `LocalFilesystemWithShell.java:60` — the static `STREAM_READER_POOL` is an unbounded cached thread pool that is never shut down.
- **[warning]** `LocalFilesystemWithShell.java:361` — streams are still fully buffered with `readAllBytes()` before `maxOutputBytes` truncation, preserving the OOM risk for huge outputs.
- **[warning]** `LocalFilesystemWithShell.java:449` — `InterruptedException` is consumed inside `getOutput`, returning empty output instead of propagating the cancellation signal.

> Note: this PR cannot be approved until the CLA is signed (see CLA reminder).

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -408,6 +443,19 @@ public ExecuteResponse execute(
         }
     }
 
+    private byte[] getOutput(Future<byte[]> future, long timeout, TimeUnit unit) {
+        try {
+            return future.get(timeout, unit);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            future.cancel(true);
```
oss-maintainer: `getOutput` catches `ExecutionException` together with `TimeoutException` and returns an empty byte array, so an I/O failure while reading the process stream is silently swallowed. In the original code such an exception propagated to the caller and produced an error response; now the command can be reported as successful with missing output. Split the handling: propagate `ExecutionException` (e.g., wrap its cause as `IOException`) or at least log it at error level and surface the failure in the returned output.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -49,6 +56,23 @@ public class LocalFilesystemWithShell extends LocalFilesystem implements Abstrac
 
     private static final Logger log = LoggerFactory.getLogger(LocalFilesystemWithShell.class);
 
+    private static final ExecutorService STREAM_READER_POOL =
+            Executors.newCachedThreadPool(
```
oss-maintainer: The new `STREAM_READER_POOL` is a static, unbounded cached thread pool that is never shut down. Under concurrent shell load this can create a large number of threads, and without a lifecycle hook the pool outlives dynamic class loaders. Replace it with a bounded executor (or a virtual-thread executor if the project runs on Java 21+) and provide a shutdown path via `Closeable` or a framework shutdown hook.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -335,11 +359,22 @@ public ExecuteResponse execute(
 
             Process proc = pb.start();
```
oss-maintainer: The reader tasks still buffer each stream fully with `readAllBytes()` before the `maxOutputBytes` truncation is applied, so a command emitting huge output can still exhaust the heap, and both stdout and stderr are now buffered concurrently. Use bounded stream readers that stop at `maxOutputBytes` (plus a small margin for stderr formatting) to avoid OOM while preserving the truncation behavior.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShell.java
```diff
@@ -408,6 +443,19 @@ public ExecuteResponse execute(
         }
     }
 
+    private byte[] getOutput(Future<byte[]> future, long timeout, TimeUnit unit) {
+        try {
+            return future.get(timeout, unit);
+        } catch (InterruptedException e) {
```
oss-maintainer: In `getOutput`, `InterruptedException` sets the thread's interrupt flag and then returns an empty result, so a caller that uses interruption to cancel agent execution will see a normal response instead of an error. Either propagate the interrupt as an exception or handle it explicitly in the caller rather than swallowing it in the helper.

CryoThrust:
Addressed the review findings in c15e255b:\n\n- Replaced the static cached reader pool with per-execution daemon readers that are shut down in a finally block.\n- Bound each stdout/stderr capture while continuing to drain the underlying stream, so large output cannot be fully retained in memory.\n- Propagate reader failures to the command error response instead of treating them as empty output, and preserve interruption while terminating the active process.\n- Added focused coverage for bounded capture and reader I/O failures.\n\nVerified locally with:\n\n`mvn -pl agentscope-harness -am -Dtest=LocalFilesystemWithShellTest -DfailIfNoTests=false test`\n\nSpotless also passes for the harness module.

CryoThrust:
Follow-up in 6d5bf576: Linux CI exposed a timeout edge case where a shell descendant can retain an output pipe after the parent process is forcibly terminated. The timeout branch now cancels both readers and returns the established 124 response immediately, instead of letting reader collection convert a timeout into an execution failure. The focused harness test suite and Spotless pass locally.

CryoThrust:
The CLA check is now green, and the Linux/Windows builds, module sync, license check, and Codecov checks are all passing on the latest commit. The PR is ready for maintainer review.

jujn:
fixed in: https://github.com/agentscope-ai/agentscope-java/pull/2839

PR #2940, opened 2026-09-02
fix(a2a): suppress cancellation after final event

### LINKED ISSUE #2927: [Bug]:JSON-RPC streaming client reports "Request cancelled" after receiving final TaskStatusUpdateEvent
## Description

  When using the JSON-RPC streaming client with `a2a-java-sdk` version `0.3.3.Final`, the error callback is invoked with:

  ```text
  Request cancelled

  after the client has already received a valid terminal TaskStatusUpdateEvent with final=true.

  The remote A2A task completes successfully, but the client reports the normal end of the stream as an error.

  ## Environment

  - a2a-java-sdk: 0.3.3.Final
  - Transport: JSON-RPC over SSE
  - HTTP client: JdkA2AHttpClient
  - Java: 21
  - A2A method: message/stream
  - Protocol version: A2A 0.3.x

  Relevant dependencies:

  <dependency>
      <groupId>io.github.a2asdk</groupId>
      <artifactId>a2a-java-sdk-client</artifactId>
      <version>0.3.3.Final</version>
  </dependency>

  <dependency>
      <groupId>io.github.a2asdk</groupId>
      <artifactId>a2a-java-sdk-client-transport-jsonrpc</artifactId>
      <version>0.3.3.Final</version>
  </dependency>

  <dependency>
      <groupId>io.github.a2asdk</groupId>
      <artifactId>a2a-java-sdk-http-client</artifactId>
      <version>0.3.3.Final</version>
  </dependency>

  ## Steps to reproduce

  Create an A2A client with JSON-RPC streaming enabled:

  ClientConfig config = ClientConfig.builder()
          .setStreaming(true)
          .setPolling(false)
          .build();

  Client client = Client.builder(agentCard)
          .withTransport(
                  JSONRPCTransport.class,
                  new JSONRPCTransportConfig(new JdkA2AHttpClient()))
          .clientConfig(config)
          .build();

  Send a streaming message:

  client.sendMessage(
          message,
          List.of((event, card) -> {
              System.out.println("EVENT: " + event);
          }),
          error -> {
              System.out.println("ERROR: " + error);
          },
          null);

  The server returns normal A2A SSE events and eventually sends:

  {
    "jsonrpc": "2.0",
    "id": "request-id",
    "result": {
      "taskId": "task-id",
      "status": {
        "state": "completed",
        "timestamp": "2026-09-01T06:19:29.2982548Z"
      },
      "contextId": "context-id",
      "final": true,
      "kind": "status-update"
    }
  }

  The server logs confirm that the task completed successfully:

  Agent execution completed successfully
  Task completed: signal=onComplete

  However, immediately after the final event, the client error callback receives:

  Request cancelled

  ## Actual behavior

  The callback sequence is effectively:

  TaskStatusUpdateEvent(final=true)
  errorHandler(Request cancelled)

  The application therefore sees a successfully completed A2A task as a failed streaming request.

  The issue does not occur when the same endpoint is called with curl -N. curl receives the terminal event and exits normally.

  ## Expected behavior

  After receiving:

  {
    "kind": "status-update",
    "final": true,
    "status": {
      "state": "completed"
    }
  }

  the client should complete normally.

  Expected callback behavior:

  TaskStatusUpdateEvent(final=true)
  completion callback

  The error callback should not receive Request cancelled for an internally initiated stream shutdown after a valid final event.

  ## Suspected root cause

  SSEEventListener.handleMessage() appears to cancel the HTTP future after delivering a final status event:

  eventHandler.accept(streamingEvent);

  if (streamingEvent instanceof TaskStatusUpdateEvent statusUpdate
          && statusUpdate.isFinal()) {
      future.cancel(true);
  }

  Cancelling the JDK HTTP client's future causes the HTTP layer to report:

  Request cancelled

  The resulting exception is then forwarded by SSEEventListener.onError():

  public void onError(Throwable error, Future<Void> future) {
      if (errorHandler != null) {
          errorHandler.accept(error);
      }
      future.cancel(true);
  }

  The client therefore does not distinguish between:

  1. an internal cancellation after receiving final=true;
  2. a user-initiated cancellation;
  3. an actual network or transport failure.

  The completed flag also appears to be set only from onComplete(), not when a final TaskStatusUpdateEvent is received.

  ## Suggested fix

  Mark the listener as successfully completed before cancelling the HTTP future, and suppress the cancellation exception caused by this internal
  cleanup.

  For example:

  private volatile boolean completed;

  private void handleMessage(JsonNode message, Future<Void> future) {
      StreamingEventKind event = parseEvent(message);

      eventHandler.accept(event);

      if (event instanceof TaskStatusUpdateEvent statusUpdate
              && statusUpdate.isFinal()) {
          completed = true;
          future.cancel(true);
          signalSuccessfulCompletion();
      }
  }

  public void onError(Throwable error, Future<Void> future) {
      if (completed && isCancellation(error)) {
          return;
      }

      if (errorHandler != null) {
          errorHandler.accept(error);
      }
  }

  Alternatively, the client could wait for the server to close the SSE connection naturally instead of cancelling the future after receiving
  final=true.

  ## Workaround

  At the application layer, we currently ignore Request cancelled only when:

  - a final event has already been received; or
  - valid response artifacts have already been received.

  A cancellation received before any valid response is still treated as a real error.

  This workaround prevents successful A2A streaming tasks from being reported as failures, but the behavior would be better fixed inside the SDK.

wmh123wmh1:
## Summary

- Keep `a2a-java-sdk 0.3.3.Final` and A2A Protocol v0.3 compatibility.
- Wrap the default JSON-RPC HTTP client to suppress the internal `CancellationException` emitted after a final status update.
- Continue forwarding cancellations received before a final event and real transport errors.
- Add regression tests for the callback behavior.

## Testing

- `mvn -pl agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client -am test`

Fixes #2927

### DIFF
--- a/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/A2aAgent.java
+++ b/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/A2aAgent.java
@@ -188,7 +188,9 @@ public class A2aAgent extends AgentBase {
         ClientBuilder builder = Client.builder(this.agentCardResolver.getAgentCard(name));
         if (this.a2aAgentConfig.clientTransports().isEmpty()) {
             // Default Add The Basic JSON-RPC Transport
-            builder.withTransport(JSONRPCTransport.class, new JSONRPCTransportConfig());
+            builder.withTransport(
+                    JSONRPCTransport.class,
+                    new JSONRPCTransportConfig(new CancellationSuppressingA2aHttpClient()));
         } else {
             this.a2aAgentConfig.clientTransports().forEach(builder::withTransport);
         }
--- a/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/CancellationSuppressingA2aHttpClient.java
+++ b/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/CancellationSuppressingA2aHttpClient.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.agentscope.core.a2a.agent;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.a2a.client.http.A2AHttpClient;
+import io.a2a.client.http.A2AHttpResponse;
+import io.a2a.client.http.JdkA2AHttpClient;
+import io.a2a.util.Utils;
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
+
+/**
+ * Applies the A2A Java SDK 1.1 SSE cancellation fix to the v0.3 client API.
+ *
+ * <p>The v0.3.3 SDK cancels its HTTP future after delivering a final status update. The JDK HTTP
+ * client can report that internal cleanup as a {@link CancellationException}, which is then passed
+ * to the streaming error callback. This decorator keeps the v0.3 SDK/API and suppresses only that
+ * cancellation after a final status-update event has been received.
+ *
+ * <p>This class is intentionally package-private. It is an implementation detail of the default
+ * JSON-RPC transport and can be removed when AgentScope moves to a v0.3-compatible SDK release
+ * containing the upstream fix.
+ */
+final class CancellationSuppressingA2aHttpClient implements A2AHttpClient {
+
+    private final A2AHttpClient delegate;
+
+    CancellationSuppressingA2aHttpClient() {
+        this(new JdkA2AHttpClient());
+    }
+
+    CancellationSuppressingA2aHttpClient(A2AHttpClient delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public GetBuilder createGet() {
+        return delegate.createGet();
+    }
+
+    @Override
+    public PostBuilder createPost() {
+        return new CancellationSuppressingPostBuilder(delegate.createPost());
+    }
+
+    @Override
+    public DeleteBuilder createDelete() {
+        return delegate.createDelete();
+    }
+
+    private static boolean isCancellation(Throwable throwable) {
+        for (Throwable current = throwable; current != null; current = current.getCause()) {
+            if (current instanceof CancellationException) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static boolean isFinalStatusUpdate(String message) {
+        try {
+            JsonNode root = Utils.OBJECT_MAPPER.readTree(message);
+            JsonNode result = root.get("result");
+            return result != null
+                    && "status-update".equals(result.path("kind").asText())
+                    && result.path("final").asBoolean(false);
+        } catch (IOException | RuntimeException ignored) {
+            return false;
+        }
+    }
+
+    private static final class CancellationSuppressingPostBuilder implements PostBuilder {
+
+        private final PostBuilder delegate;
+
+        private final AtomicBoolean completed = new AtomicBoolean();
+
+        private CancellationSuppressingPostBuilder(PostBuilder delegate) {
+            this.delegate = delegate;
+        }
+
+        @Override
+        public PostBuilder url(String url) {
+            delegate.url(url);
+            return this;
+        }
+
+        @Override
+        public PostBuilder addHeaders(Map<String, String> headers) {
+            delegate.addHeaders(headers);
+            return this;
+        }
+
+        @Override
+        public PostBuilder addHeader(String name, String value) {
+            delegate.addHeader(name, value);
+            return this;
+        }
+
+        @Override
+        public PostBuilder body(String body) {
+            delegate.body(body);
+            return this;
+        }
+
+        @Override
+        public A2AHttpResponse post() throws IOException, InterruptedException {
+            return delegate.post();
+        }
+
+        @Override
+        public CompletableFuture<Void> postAsyncSSE(
+                Consumer<String> messageConsumer,
+                Consumer<Throwable> errorConsumer,
+                Runnable completeRunnable)
+                throws IOException, InterruptedException {
+            completed.set(false);
+            return delegate.postAsyncSSE(
+                    message -> {
+                        if (isFinalStatusUpdate(message)) {
+                            completed.set(true);
+                        }
+                        messageConsumer.accept(message);
+                    },
+                    error -> {
+                        if (!completed.get() || !isCancellation(error)) {
+                            errorConsumer.accept(error);
+                        }
+                    },
+                    completeRunnable);
+        }
+    }
+}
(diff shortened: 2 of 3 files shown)


dailingtao (approved):
LGTM

oss-maintainer (commented):
### Summary
Package-private `A2AHttpClient` decorator that suppresses the spurious `CancellationException` the a2a-java-sdk 0.3.3 emits after a final `status-update`, wired into `A2aAgent`'s default JSON-RPC transport. Direction looks right and the callback matrix is well tested (final vs non-final, malformed JSON, non-cancellation errors, builder delegation, forward of `post()`).

The three substantive findings below are all the same gap: the decorator filters `errorConsumer`, but not the two other ways a stream terminates — the returned `CompletableFuture` and `completeRunnable`. If the SDK signals the cancellation through either of those instead, a caller either still sees the error or sees a stream that never terminates. CI is green and the CLA is signed, so nothing here blocks the change — I'd just like confirmation on the completion-signal question.

---

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/CancellationSuppressingA2aHttpClient.java
```diff
@@ -0,0 +1,152 @@
...
+            completed.set(false);
+            return delegate.postAsyncSSE(
+                    message -> {
+                        if (isFinalStatusUpdate(message)) {
+                            completed.set(true);
+                        }
+                        messageConsumer.accept(message);
+                    },
+                    error -> {
+                        if (!completed.get() || !isCancellation(error)) {
```
oss-maintainer, line 145: **[Warning]** Suppressing the error here drops the *only* terminal signal the consumer gets: when the SDK reports the post-final cancellation through `errorConsumer`, `completeRunnable` is not invoked either, so a caller that closes its stream / releases resources in `onComplete` will see a stream that never terminates (it just goes quiet). Please confirm the SDK's behaviour for this path and, if it holds, run `completeRunnable` (or otherwise signal completion) when the error is suppressed, e.g.

```java
error -> {
    if (completed.get() && isCancellation(error)) {
        completeRunnable.run();   // terminal signal, exactly once
    } else {
        errorConsumer.accept(error);
    }
}
```

If `postAsyncSSE` can call both callbacks, guard with an `AtomicBoolean` compare-and-set so the consumer still sees exactly one terminal event.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/CancellationSuppressingA2aHttpClient.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        @Override
+        public CompletableFuture<Void> postAsyncSSE(
+                Consumer<String> messageConsumer,
+                Consumer<Throwable> errorConsumer,
+                Runnable completeRunnable)
+                throws IOException, InterruptedException {
+            completed.set(false);
+            return delegate.postAsyncSSE(
```
oss-maintainer, line 137: **[Warning]** The decorator only filters the three callbacks; the `CompletableFuture` returned by the delegate is passed through untouched. If `JdkA2AHttpClient` completes that future exceptionally with the same `CancellationException` (rather than / in addition to invoking `errorConsumer`), any caller that joins or composes on the returned future will still observe the spurious failure, i.e. the fix would be incomplete for those call sites. Could you verify how the 0.3.3 SDK surfaces this cancellation and, if the future is also affected, wrap the return value as well (`.exceptionally(t -> completed.get() && isCancellation(t) ? null : rethrow(t))`)? A test that inspects the returned future's state in the suppressed case would pin this down — all current tests drive the callbacks directly and use the mock's never-completed future, so this path is untested.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/CancellationSuppressingA2aHttpClient.java
```diff
@@ -0,0 +1,152 @@
...
+            if (current instanceof CancellationException) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static boolean isFinalStatusUpdate(String message) {
+        try {
+            JsonNode root = Utils.OBJECT_MAPPER.readTree(message);
```
oss-maintainer, line 81: **[Warning]** This runs a full Jackson tree parse for **every** SSE chunk, on top of the parse the SDK itself already does, so a long `message/stream` response pays double deserialization on the hot path (artifact/token chunks can be many and large). A cheap pre-filter keeps the behaviour identical while making the non-matching case nearly free:

```java
if (message == null || message.indexOf('"') < 0
        || !message.contains("status-update")) {
    return false;
}
```

Alternative: only try detection while `completed` is still `false`, since once a final status update has been seen there is nothing left to record.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/CancellationSuppressingA2aHttpClient.java
```diff
@@ -0,0 +1,152 @@
...
+    CancellationSuppressingA2aHttpClient() {
+        this(new JdkA2AHttpClient());
+    }
+
+    CancellationSuppressingA2aHttpClient(A2AHttpClient delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public GetBuilder createGet() {
```
oss-maintainer, line 56: **[Info]** `createGet()` is delegated unwrapped, so only the POST streaming path gets the fix. If the same SDK bug can surface on the GET-based SSE path (e.g. `tasks/resubscribe` / polling a running task), that path is still exposed. Worth a one-line comment stating the intended scope ("only `message/stream` over JSON-RPC POST is affected") so the next reader does not treat the missing GET wrapper as an oversight — or wrap `GetBuilder` symmetrically if it can also stream.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/CancellationSuppressingA2aHttpClient.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    private static boolean isFinalStatusUpdate(String message) {
+        try {
+            JsonNode root = Utils.OBJECT_MAPPER.readTree(message);
+            JsonNode result = root.get("result");
+            return result != null
+                    && "status-update".equals(result.path("kind").asText())
+                    && result.path("final").asBoolean(false);
+        } catch (IOException | RuntimeException ignored) {
```
oss-maintainer, line 86: **[Info]** The parse failure is swallowed silently, which is the right default for non-JSON keep-alives, but it also means a genuinely final event that does not match the expected shape (different envelope, or `final` nested differently) silently disables the suppression and the user keeps seeing the confusing `CancellationException`. A one-line `debug`/`trace` log here (and one when an error is actually suppressed at line 145) would make field diagnosis of "why is this workaround not working" much cheaper.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-a2a/agentscope-extensions-a2a-client/src/main/java/io/agentscope/core/a2a/agent/A2aAgent.java
```diff
@@ -188,7 +188,9 @@ private Client buildA2aClient(String name) {
         ClientBuilder builder = Client.builder(this.agentCardResolver.getAgentCard(name));
         if (this.a2aAgentConfig.clientTransports().isEmpty()) {
             // Default Add The Basic JSON-RPC Transport
-            builder.withTransport(JSONRPCTransport.class, new JSONRPCTransportConfig());
+            builder.withTransport(
+                    JSONRPCTransport.class,
+                    new JSONRPCTransportConfig(new CancellationSuppressingA2aHttpClient()));
```
oss-maintainer, line 193: **[Info]** This changes the default client for every `A2aAgent` built without explicit transports, with no opt-out, and the suppression is a behaviour change (errors disappear) rather than a pure fix. Two small things would make it easier to live with: (1) link the upstream SDK issue/PR in the Javadoc of `CancellationSuppressingA2aHttpClient` together with the version that fixes it, so the workaround can be deleted mechanically later; (2) consider gating it behind an `A2aAgentConfig` flag (default on) so users who *want* to observe cancellations are not stuck.

PR #2945, opened 2026-09-02
feat(core): add returnDirect terminal tool short-circuit

### LINKED ISSUE #1408: [Feature]: Can Tool Annotation support returnRedirect?

dragondyt:
## AgentScope-Java Version

2.0.3-SNAPSHOT

Closes #1408
Related to #2362

## Description

Adds first-class `returnDirect` / `returnResult` on `@Tool`, `AgentTool`, and `ToolBase` so a tool can end the ReAct turn after execution without another model call or `summarizing()` step.

- `returnDirect=true`: persist `tool_use` / `tool_result`, then stop with `GenerateReason.RETURN_DIRECT`
- `returnResult=false`: silent stop (empty reply)
- Concurrent `returnDirect` tools in one batch: all surfaced results are returned in original tool-call order (`TextBlock`s plus `MessageMetadataKeys.RETURN_DIRECT_RESULTS`)

This differs from #2891, which short-circuits only when every successful tool in the batch is `returnDirect`. This PR short-circuits when any successful tool is `returnDirect`.

## Why this is a draft (not ready to land)

Do **not** merge yet.

After a `returnDirect` tool finishes, AG-UI clients with server-side memory (CopilotKit and similar) still replay the full transcript:

`user -> assistant(tool_calls) -> tool -> next user`

`AguiRequestProcessor.extractLatestUserMessage` treats everything after the last assistant message as a follow-up. The already-persisted `tool` result is extracted again and appended to server memory, so the next turn receives an extra tool message.

HITL resume still depends on trailing tool-only messages after the last assistant, so a correct fix has to drop the completed `returnDirect` tool turn without breaking interrupt approval. That follow-up is out of scope here.

## Checklist

- [ ] Code is ready for review

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -218,14 +218,16 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
     private static final GracefulShutdownManager shutdownManager =
             GracefulShutdownManager.getInstance();
 
-    /** Tool name used for the per-call structured-output {@code generate_response} tool. */
+    /**
+     * Tool name used for the per-call structured-output {@code generate_response} tool.
+     */
     public static final String STRUCTURED_OUTPUT_TOOL_NAME = "generate_response";
 
     /**
      * @deprecated Permission HITL no longer uses a Reactor Sink. Confirm results are now
-     *     delivered via a second {@code agent.call(msgs)} carrying a {@link ConfirmResult}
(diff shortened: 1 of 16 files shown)


dragondyt:
与 #2891  重复

PR #2947, opened 2026-09-02
fix(harness): bound hung memory background tasks without losing final flushes

birdie7761:
## Problem

Since #2777 the per-call memory flush and maintenance run fire-and-forget, but a hung model call has no bound and no escape:

- `runFlush`'s pipeline has no timeout, so a hung flush keeps the conversation's coalescing slot in `FLUSH_QUEUES` **running forever** — every later flush of that conversation queues behind the dead slot and memory extraction permanently stops for it;
- `MemoryBackgroundTasks.inFlight` never returns to zero, so every subsequent `close()` in the JVM waits its full 5s budget forever after;
- maintenance consolidation runs `consolidator.consolidate(rc).block()` inside `Mono.fromRunnable`, so a hung consolidation **leaks a boundedElastic worker for the rest of the process**: the inner block subscription is unreachable from any outer dispose, and once the model eventually returns, the same worker keeps writing retention/prune results into the workspace.

## Fix

- Both pipelines get a **5-minute timeout** — the only cancellation mechanism. The budget runs from subscription (including any `boundedElastic` pickup wait): under scheduler saturation a healthy-but-slow run may be skipped and logged, the safe direction for fire-and-forget work.
- `doMaintenance` becomes a **reactive composition** (`expire → consolidator.consolidate → prune`), so the timeout's cancel propagates down the chain into the consolidation model stream and the worker is actually freed. Consolidation failures log and let the retention steps still run, matching the previous try/catch semantics.

## Deliberately NOT changed: quiescence keeps its abandon semantics

Cancelling whatever is still in flight when `close()`'s 5s budget elapses would kill **healthy** flushes: with the pipeline timeout in place, anything that survives to that moment is by definition still within its 5-minute bound, and an LLM flush routinely outlives 5s — with fire-and-forget memory work, a close-during-flush is the common case, not the exception. A close-time sweep would therefore routinely drop the last turn's memory extraction. Instead, an abandoned task keeps running and typically completes; only work hung past the pipeline timeout dies. `MemoryBackgroundTasks` is unchanged apart from a javadoc note documenting this division of responsibilities.

## Boundary

The #2935-style window of a background task writing against an already-torn-down workspace is **not eliminated** — it is bounded from *unbounded* down to at most 5 minutes (the pipeline timeout). That is the inherent cost of the abandon semantics, and strictly better than before this change.

## Test evidence

New `MemoryBackgroundHardeningTest` (3 cases): a short quiescence budget gives up **without cancelling** a healthy in-flight flush, which then completes on its own (the abandon invariant, locked); a hung flush times out, releases the quiescence count and the conversation queue recovers; a hung consolidation times out and the cancellation is asserted to reach the consolidation subscription itself (the old `.block()` variant released the counter while the worker stayed stuck). Timeouts are injectable via package-private test hooks; production uses the 5-minute default. Regression: `MemoryFlushMiddlewareAsyncFlushTest` / `MemoryMaintenanceMiddlewareAsyncGateTest` green; full harness suite 888/888.

## Checklist

- [x] Related documentation has been updated (class javadocs document the abandon/timeout division; no user-facing behavior or config changed)

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
@@ -29,6 +29,7 @@ import io.agentscope.harness.agent.memory.MemoryBackgroundTasks;
 import io.agentscope.harness.agent.memory.MemoryConfig;
 import io.agentscope.harness.agent.memory.MemoryFlushManager;
 import io.agentscope.harness.agent.workspace.WorkspaceManager;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
@@ -85,6 +86,13 @@ import reactor.core.scheduler.Schedulers;
 public class MemoryFlushMiddleware implements HarnessRuntimeMiddleware {
 
     private static final Logger log = LoggerFactory.getLogger(MemoryFlushMiddleware.class);
+
+    /**
+     * Upper bound on the flush LLM call, so a hung model provider cannot pin the conversation's
+     * coalescing slot forever (every later flush of that conversation would queue behind it) or
+     * keep the JVM-wide quiescence count above zero for the rest of the process lifetime.
+     */
+    static final Duration FLUSH_TIMEOUT = Duration.ofMinutes(5);
 
     private final WorkspaceManager workspaceManager;
     private final Model model;
@@ -149,6 +157,14 @@ public class MemoryFlushMiddleware implements HarnessRuntimeMiddleware {
         this.periodicGate = periodicGate != null ? periodicGate : new LocalPeriodicGate();
     }
 
+    /** Timeout applied to the flush pipeline; production builds use {@link #FLUSH_TIMEOUT}. */
+    private Duration flushTimeout = FLUSH_TIMEOUT;
+
+    /** Test hook to shrink the flush timeout; keeps timeout behaviour unit-testable. */
+    void setFlushTimeoutForTests(Duration timeout) {
+        this.flushTimeout = timeout != null ? timeout : FLUSH_TIMEOUT;
+    }
+
     @Override
     public Flux<AgentEvent> onAgent(
             Agent agent,
@@ -203,6 +219,7 @@ public class MemoryFlushMiddleware implements HarnessRuntimeMiddleware {
     private void runFlush(String key, Agent agent, RuntimeContext rc) {
         Mono.defer(() -> doFlush(agent, rc))
                 .subscribeOn(Schedulers.boundedElastic())
+                .timeout(flushTimeout)
                 .doFinally(
                         signal -> {
                             MemoryBackgroundTasks.end();
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryMaintenanceMiddleware.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryMaintenanceMiddleware.java
@@ -130,6 +130,24 @@ public class MemoryMaintenanceMiddleware implements HarnessRuntimeMiddleware {
         this.periodicGate = periodicGate != null ? periodicGate : new LocalPeriodicGate();
     }
 
+    /**
+     * Upper bound on the maintenance run (gate claim + retention sweep + consolidation LLM
+     * call), so a hung model cannot keep the JVM-wide quiescence count above zero for the rest
+     * of the process lifetime. Mirrors {@code MemoryFlushMiddleware#FLUSH_TIMEOUT}.
+     */
+    static final Duration MAINTENANCE_TIMEOUT = Duration.ofMinutes(5);
+
+    /**
+     * Timeout applied to the maintenance pipeline; production uses {@link
+     * #MAINTENANCE_TIMEOUT}.
+     */
+    private Duration maintenanceTimeout = MAINTENANCE_TIMEOUT;
+
+    /** Test hook to shrink the maintenance timeout; keeps timeout behaviour unit-testable. */
+    void setMaintenanceTimeoutForTests(Duration timeout) {
+        this.maintenanceTimeout = timeout != null ? timeout : MAINTENANCE_TIMEOUT;
+    }
+
     public MemoryMaintenanceMiddleware(
             WorkspaceManager workspaceManager, MemoryConsolidator consolidator) {
         this(workspaceManager, consolidator, 90, 180, DEFAULT_MIN_GAP);
@@ -152,6 +170,7 @@ public class MemoryMaintenanceMiddleware implements HarnessRuntimeMiddleware {
                             MemoryBackgroundTasks.begin();
                             Mono.defer(() -> doMaintenance(rc))
                                     .subscribeOn(Schedulers.boundedElastic())
+                                    .timeout(maintenanceTimeout)
                                     .doFinally(signal -> MemoryBackgroundTasks.end())
                                     .subscribe(
                                             null,
@@ -163,12 +182,37 @@ public class MemoryMaintenanceMiddleware implements HarnessRuntimeMiddleware {
     }
 
     private Mono<Void> doMaintenance(RuntimeContext rc) {
-        if (!periodicGate.tryClaim(compositeTimerKey(rc), minGap)) {
-            // Throttled out; the in-flight slot acquired at dispatch is released when this
-            // Mono completes.
+        return Mono.defer(
+                () -> {
+                    if (!periodicGate.tryClaim(compositeTimerKey(rc), minGap)) {
+                        // Throttled out; the in-flight slot acquired at dispatch is released
+                        // when this Mono completes.
+                        return Mono.empty();
+                    }
+                    // The pipeline stays reactive end-to-end: the pipeline timeout's cancel
+                    // propagates into the consolidation model stream instead of stranding the
+                    // worker on an inner .block() that no outer dispose can reach.
+                    return Mono.fromRunnable(() -> expireDailyFiles(rc))
+                            .then(consolidateReactive(rc))
+                            .then(Mono.fromRunnable(() -> pruneOldSessions(rc)));
+                });
+    }
+
+    /**
+     * Consolidation segment of the maintenance pipeline. Reactive so cancellation reaches the
+     * underlying model stream; a consolidation failure logs and lets the retention steps still
+     * run, matching the previous try/catch semantics around {@code consolidate().block()}.
+     */
+    private Mono<Void> consolidateReactive(RuntimeContext rc) {
+        if (consolidator == null) {
             return Mono.empty();
         }
-        return Mono.fromRunnable(() -> runMaintenance(rc));
+        return Mono.defer(() -> consolidator.consolidate(rc))
+                .onErrorResume(
+                        e -> {
+                            log.warn("Memory consolidation failed: {}", e.getMessage());
+                            return Mono.empty();
+                        });
     }
 
     /**
@@ -193,14 +237,6 @@ public class MemoryMaintenanceMiddleware implements HarnessRuntimeMiddleware {
                     MemoryFlushMiddleware.blankToEmpty(rc != null ? rc.getSessionId() : null);
             case AGENT, GLOBAL -> "";
         };
-    }
-
-    private void runMaintenance(RuntimeContext rc) {
-        log.debug("Running memory maintenance...");
-        expireDailyFiles(rc);
-        consolidateMemory(rc);
-        pruneOldSessions(rc);
-        log.debug("Memory maintenance completed");
     }
 
     private void expireDailyFiles(RuntimeContext rc) {
@@ -237,17 +273,6 @@ public class MemoryMaintenanceMiddleware implements HarnessRuntimeMiddleware {
             } catch (Exception e) {
                 // not a date-named file, skip
             }
-        }
-    }
-
-    private void consolidateMemory(RuntimeContext rc) {
-        if (consolidator == null) {
-            return;
-        }
-        try {
-            consolidator.consolidate(rc).block();
-        } catch (Exception e) {
-            log.warn("Memory consolidation failed: {}", e.getMessage());
         }
     }
 
--- a/agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/MemoryBackgroundHardeningTest.java
+++ b/agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/MemoryBackgroundHardeningTest.java
@@ -1,0 +1,239 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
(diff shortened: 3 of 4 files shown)


oss-maintainer (commented):
test probe

oss-maintainer:
test-probe-1789289717

oss-maintainer:
dummy-skip

oss-maintainer (commented):
### Summary

Bounds the fire-and-forget memory flush/maintenance pipelines with a 5-minute timeout and converts `doMaintenance` from a `Mono.fromRunnable` + inner `.block()` into a reactive composition, so a hung model call can no longer pin the conversation's coalescing slot or the JVM-wide quiescence counter. The direction is right, the reasoning in the javadoc is unusually clear, and `MemoryBackgroundHardeningTest` encodes the exact invariants (abandon-don't-cancel, queue recovery after a timeout, cancel reaching the consolidation subscription) — that is the part I most want to see land.

Findings are all in the "the fix is narrower than the claim" category rather than objections: the reactive rewrite moves blocking work onto the model client's thread (F1), and the timeout cannot interrupt the blocking filesystem segments it also runs through (F2). Both are cheap to address.

### Verdict

**COMMENT** — no blocking issue; the thread-placement point (F1) is worth resolving before merge since it is a regression introduced by this refactor, and configurable + distinguishable timeouts (F2/F4) would make the bound operationally usable.

Nice touch: asserting `flushCancelled` stays `false` in the healthy-flush test locks in the "never drop the last turn's memories" property that the doc comment promises.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryMaintenanceMiddleware.java
```diff
@@ -163,12 +182,37 @@ public Flux<AgentEvent> onAgent(
...
+                        // Throttled out; the in-flight slot acquired at dispatch is released
+                        // when this Mono completes.
+                        return Mono.empty();
+                    }
+                    // The pipeline stays reactive end-to-end: the pipeline timeout's cancel
+                    // propagates into the consolidation model stream instead of stranding the
+                    // worker on an inner .block() that no outer dispose can reach.
+                    return Mono.fromRunnable(() -> expireDailyFiles(rc))
+                            .then(consolidateReactive(rc))
+                            .then(Mono.fromRunnable(() -> pruneOldSessions(rc)));
```
oss-maintainer: **[WARNING]** Thread-placement regression risk: before this change all three stages ran inside a single `Mono.fromRunnable(() -> runMaintenance(rc))` on `boundedElastic`. After the reactive composition, the `then(...)` continuation subscribes to `Mono.fromRunnable(() -> pruneOldSessions(rc))` on whichever thread emitted `consolidate()`'s terminal signal — i.e. the model HTTP client's I/O thread. `pruneOldSessions` does blocking workspace/file I/O, so retention pruning can now run on an event-loop thread and stall it.

Suggest pinning the blocking segments back to the background scheduler, e.g.
```java
return Mono.fromRunnable(() -> expireDailyFiles(rc))
        .then(consolidateReactive(rc))
        .then(Mono.fromRunnable(() -> pruneOldSessions(rc)).subscribeOn(Schedulers.boundedElastic()));
```

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryMaintenanceMiddleware.java
```diff
@@ -163,12 +182,37 @@ public Flux<AgentEvent> onAgent(
...
+        return Mono.defer(
+                () -> {
+                    if (!periodicGate.tryClaim(compositeTimerKey(rc), minGap)) {
+                        // Throttled out; the in-flight slot acquired at dispatch is released
+                        // when this Mono completes.
+                        return Mono.empty();
+                    }
+                    // The pipeline stays reactive end-to-end: the pipeline timeout's cancel
+                    // propagates into the consolidation model stream instead of stranding the
+                    // worker on an inner .block() that no outer dispose can reach.
```
oss-maintainer: **[WARNING]** The comment (and the PR description) state the worker is "actually freed" by the timeout. That holds for a hang inside the *reactive* model stream, but not for the blocking segments: `expireDailyFiles` / `pruneOldSessions` here, and the eager workspace reads inside `consolidate()` (`readWatermark` / `readMemoryMd` / `readDailyEntries`, which run at subscribe time in the `Mono.defer`). A hung filesystem (NFS, K8s-backed sandbox volume) still strands the `boundedElastic` worker indefinitely — the timeout only releases the in-flight counter via `doFinally` and unblocks the caller's quiescence wait.

Either narrow the claim in the javadoc ("cancellation reaches the consolidation model stream; blocking FS segments are not interruptible") or bound the FS segments separately so a hung volume cannot exhaust the shared `boundedElastic` pool.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryMaintenanceMiddleware.java
```diff
@@ -163,12 +182,37 @@ public Flux<AgentEvent> onAgent(
...
+                    return Mono.fromRunnable(() -> expireDailyFiles(rc))
```
oss-maintainer, line 209: **[INFO]** Two small behaviour losses while refactoring away `runMaintenance`:

1. The `log.debug("Running memory maintenance...")` / `"Memory maintenance completed"` bookends are gone, which were handy when diagnosing why a maintenance run did nothing.
2. `periodicGate.tryClaim` happens *before* the stages run, so a run that is killed by the timeout still consumes the gate slot — the next attempt waits a full `minGap` even though nothing was expired/consolidated/pruned. Worth a comment at minimum, or re-claiming on timeout.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
```diff
@@ -203,6 +219,7 @@ private void scheduleFlush(Agent agent, RuntimeContext rc) {
     private void runFlush(String key, Agent agent, RuntimeContext rc) {
         Mono.defer(() -> doFlush(agent, rc))
                 .subscribeOn(Schedulers.boundedElastic())
+                .timeout(flushTimeout)
```
oss-maintainer: **[WARNING]** `.timeout(...)` is applied after `subscribeOn`, so the 5-minute budget also counts the `boundedElastic` queue wait (acknowledged in the PR description). The consequence for the *flush* path specifically is silent memory loss: when the budget elapses the flush is cancelled, `onErrorResume` in `doFlush` is bypassed, `doFinally` drains the queue, and that conversation's extraction for the turn is simply gone — with the same `"Memory flush failed"` warn as a genuine provider error.

Two suggestions:
* make the budget configurable (via `MemoryConfig`/a system property) rather than a hardcoded `Duration.ofMinutes(5)`, so long-context / slow-provider deployments can raise it;
* distinguish a `TimeoutException` in the log message (and ideally increment a counter) so operators can tell "provider hung" apart from "we timed out a healthy-but-slow flush".

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
```diff
@@ -149,6 +157,14 @@ public MemoryFlushMiddleware(
         this.periodicGate = periodicGate != null ? periodicGate : new LocalPeriodicGate();
     }
 
+    /** Timeout applied to the flush pipeline; production builds use {@link #FLUSH_TIMEOUT}. */
+    private Duration flushTimeout = FLUSH_TIMEOUT;
+
+    /** Test hook to shrink the flush timeout; keeps timeout behaviour unit-testable. */
+    void setFlushTimeoutForTests(Duration timeout) {
```
oss-maintainer, line 167: **[INFO]** `flushTimeout` is a non-final, non-volatile instance field written from the test thread and read from the `boundedElastic` worker inside `runFlush` — technically an unsafe publication (same for `maintenanceTimeout` in `MemoryMaintenanceMiddleware`). In practice the `subscribeOn` hand-off happens to establish the edge, but it is fragile.

Prefer either folding the value into the existing package-private constructor overload, or declaring the field `private volatile Duration flushTimeout` — that also removes the need for a mutable test-only setter on a production class.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryMaintenanceMiddleware.java
birdie7761: Confirmed and fixed — thank you for this one, it's a real regression the refactor introduced. The prune stage now re-pins itself with .subscribeOn(boundedElastic) because then(...) subscribes on the thread that emitted the previous stage's terminal signal — the model HTTP client's I/O thread for consolidate(). The new test completes consolidate() on a dedicated model-io thread and asserts both retention sweeps observe boundedElastic; it was written red first and failed exactly as you predicted (prune sweep ... ran on: model-io-1).

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryMaintenanceMiddleware.java
birdie7761: The comment is now scoped to what cancellation actually reaches: the consolidation model stream (the realistic hang), while the retention sweeps are short blocking filesystem calls a cancel cannot interrupt mid-call — the worker returns when the call returns. The PR description claim should read the same way (keeping this reply as the record since the description predates the commit).

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
birdie7761: Fixed by moving the timeout inside the deferred pipeline — it now bounds the work, not the scheduler wait, for BOTH pipelines (flush and maintenance). The queue-wait-cancels-a-run-that- never-ran failure mode you described is gone; a timeout that fires on actual work is now also logged for what it is: "Memory flush for timed out after ; the conversation's extraction for this turn is dropped". The existing hung-flush/hung-maintenance tests confirm in-work cancellation semantics are unchanged.

agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/MemoryBackgroundHardeningTest.java
```diff
@@ -0,0 +1,367 @@
...
+        RecordingThreadFs fs = new RecordingThreadFs();
+        WorkspaceManager wsm = mock(WorkspaceManager.class);
+        when(wsm.getFilesystem()).thenReturn(fs);
+        MemoryConsolidator consolidator = mock(MemoryConsolidator.class);
+        when(consolidator.consolidate(any()))
+                .thenAnswer(
+                        inv ->
+                                Mono.<Void>empty()
+                                        .subscribeOn(
+                                                reactor.core.scheduler.Schedulers.newSingle(
```
oss-maintainer: This stub creates a fresh `Schedulers.newSingle("model-io")` on every `consolidate(...)` invocation and never disposes it, so the test leaks a scheduler thread per call (the assertion only needs the completion thread, not the scheduler identity). Hoist it into a field created in the test body and `dispose()` it in a `finally` / `@AfterEach`, otherwise this class slowly accumulates live `model-io` threads as the suite grows.

agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/MemoryBackgroundHardeningTest.java
```diff
@@ -0,0 +1,367 @@
...
+        AtomicBoolean consolidationCancelled = new AtomicBoolean(false);
+        MemoryConsolidator consolidator = mock(MemoryConsolidator.class);
+        when(consolidator.consolidate(any()))
+                .thenAnswer(
+                        inv ->
+                                Mono.<Void>never()
+                                        .doOnCancel(() -> consolidationCancelled.set(true)));
+
+        MemoryMaintenanceMiddleware mw = new MemoryMaintenanceMiddleware(wsm, consolidator);
+        mw.setMaintenanceTimeoutForTests(Duration.ofMillis(200));
```
oss-maintainer, line 204: 200 ms is the *timeout under test* here, and the assertions around it are wall-clock polls (`pollQuiescence(5s)`, `Thread.sleep(10)` loops). That is the right shape, but on a contended CI runner a 200 ms budget can be crossed by scheduling noise before the timeout branch is even reached, which shows up as a flake in `hungMaintenance_timesOut_releasesQuiescence` rather than a real regression. If this test turns out flaky, prefer driving the timeout through a `TestPublisher`/virtual clock instead of shrinking the real one.

oss-maintainer (commented):
### Summary
Re-review after the author's follow-up commit `a7c51b8808` (memory background work moved off the model I/O thread). The core change is right: `FLUSH_TIMEOUT` / `MAINTENANCE_TIMEOUT` are applied to the reactive pipeline with a `timeout(...)` that cancels the subscription itself (the old `.block()` variant released the in-flight counter while leaving the worker stuck — `hungMaintenance_timesOut_releasesQuiescence` pins exactly that regression), the retention sweeps are asserted to stay on `boundedElastic`, and the timeouts remain production constants with test-only setters that null-guard. The new `MemoryBackgroundTasks` quiescence bookkeeping is the piece that makes `HarnessAgent.close()` bounded, and the queue-recovery assertion in `hungFlush_timesOut_releasesQuiescenceAndQueueRecovers` covers the deadlock the fix is for.

No correctness or compatibility blocker found; two test-hygiene points are inline (one undisposed `Schedulers.newSingle` leak, one note on the 200 ms wall-clock budget).

---

oss-maintainer (approved):
### Summary
Re-reviewed the increment since my last pass (`a7c51b88` -> `ba56cf80`). It is test-only and fixes the one item I asked for, so this is an **approve**. All checks on this head (`build (ubuntu-latest)`, `build (windows-latest)`, Check License, Check Module Sync, codecov/patch) are green and the CLA is signed.

### What the increment does
- `maintenanceRetentionSweeps_neverRunOnTheModelEmittingThread` now creates a single `Schedulers.newSingle("model-io")` outside the `thenAnswer`, subscribes every `consolidate(...)` stub on it, and disposes it in a `finally`. The previous version built one undisposed single-thread scheduler per invocation, leaking a model-I/O thread per call even though only the completion thread mattered for the assertion.
- The assertions themselves are unchanged, so the thread-placement hardening from `a7c51b8808` is still what is under test; only the fixture lifecycle changed. Keeping the scheduler identity stable also removes a source of nondeterminism, since the stub no longer hops to a fresh thread per call.

### Carried over from the previous round (informational, not a blocker)
The 200 ms timeout under test next to wall-clock polls (`pollQuiescence(5s)`, the `Thread.sleep(10)` loops) is still tight on a contended runner. If it ever goes flaky, widen the polling window rather than the timeout under test, so the assertion keeps covering the real budget.

---

PR #2955, opened 2026-09-03
fix(agui): dedupe resume tool results already present in messages

### LINKED ISSUE #2954: [Bug]: AguiMessageConverter duplicates ToolResultBlock when CopilotKit sends tool message + resume
### Description

When using CopilotKit v2 `useInterrupt.resolve()` with AgentScope schema-only HITL tools (e.g. `toolkit.registerSchema(...)` / `TOOL_SUSPENDED` → AG-UI interrupt with `reason: "tool_call"`), the resumed `RunAgentInput` commonly contains **both**:

1. a `role: "tool"` message with the user payload (`toolCallId = call_xxx`)
2. a `resume[]` entry for the same interrupt / tool call

`AguiMessageConverter.toMsgList` currently converts **both** into `ToolResultBlock` with the same id. `ReActAgent` then throws:

```text
Duplicate tool result ID: call_xxx
```

### Why this is not fixed on the CopilotKit side alone

CopilotKit [#6201](https://github.com/CopilotKit/CopilotKit/issues/6201) / [#6470](https://github.com/CopilotKit/CopilotKit/pull/6470) only stops synthesizing client tool results for **non-`tool_call`** interrupts. AgentScope schema-only / permission-bound interrupts intentionally use `reason: "tool_call"`, so CopilotKit still dual-sends tool message + `resume[]` by design (keeps chat history well-formed).

The official `agentscope-copilotkit` example primarily demos permission ASK / frontend `useHumanInTheLoop`, which do not hit this path. Schema-only `requestHumanApproval` on the `default` agent **does**.

### Repro (against `agentscope-copilotkit`, agent=`default`)

1. Run until `requestHumanApproval` emits `RUN_FINISHED` with interrupt (`reason=tool_call`).
2. Resume with CopilotKit-shaped payload: same `toolCallId` in `messages` **and** `resume[]`.
3. Observe `RUN_ERROR` / `Duplicate tool result ID`.

We reproduced this against local `2.0.3-SNAPSHOT`.

### Expected

Converter should be idempotent for ordinary tool resumes: if a `ToolResultBlock` for that id already exists from `messages`, skip generating another from `resume`.

Permission-confirm resumes must still produce USER `ConfirmResult` (they are not `ToolResultBlock` duplicates).

### Proposed fix

In `AguiMessageConverter.toMsgList(RunAgentInput, Map)`:

- collect existing `ToolResultBlock` ids from converted messages
- for non-permission-confirm resumes, skip when id already present
- prefer the message-history result (CopilotKit already wrote the payload there)

resomi:
## AgentScope-Java Version

`2.0.3-SNAPSHOT` (main)

### Background

CopilotKit v2 `useInterrupt.resolve()` for `reason: "tool_call"` interrupts attaches:

- `messages`: `role:"tool"` result (for UI / history)
- `resume[]`: protocol resume payload

AgentScope schema-only HITL (`registerSchema` → `TOOL_SUSPENDED` → interrupt) uses `reason: "tool_call"`. `AguiMessageConverter.toMsgList` previously converted both into the same `ToolResultBlock` id → `ReActAgent` `Duplicate tool result ID`.

Related CopilotKit issue [#6201](https://github.com/CopilotKit/CopilotKit/issues/6201) does not remove dual-send for `tool_call` interrupts.

### Changes

- `AguiMessageConverter.toMsgList`: skip ordinary tool resumes whose `toolCallId` already has a `ToolResultBlock` from `messages` (messages win; no payload comparison).
- Permission-confirm resumes (`agentscope.interruptKind=permission_confirm`) still always append USER `ConfirmResult`.
- Unit tests for dual-path dedupe + permission-confirm keep behavior.

### How to test

```bash
mvn -pl agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui \
  -Dtest=AguiMessageConverterTest#testCopilotKitDualPathSkipsDuplicateToolResultFromResume+testPermissionConfirmResumeKeptWhenToolMessageAlreadyPresent \
  test
```

Optional E2E: `agentscope-copilotkit` agent `default`, schema-only `requestHumanApproval`, resume with tool message + `resume[]` — should continue without `Duplicate tool result ID`.

## Checklist

- [x] Focused unit tests are passing

Fixes #2954


oss-maintainer (approved):
LGTM. This PR records ToolResultBlock IDs produced from incoming AG-UI messages and skips ordinary resume results when the ID already exists, while retaining permission-confirm results. I verified the converter call path, the ReActAgent duplicate-result validation, and null-list handling; the focused new tests and the full 48-test AguiMessageConverterTest suite pass, so the change is localized, correct, and adequately covered.

---

PR #2963, opened 2026-09-03
fix(transport): unblock backpressured SSE streams

### LINKED ISSUE #2957: [Bug]: OkHttpTransport stalls backpressured SSE streams after the first token until [DONE]

**Describe the bug**

When `OkHttpTransport` is consumed with incremental downstream demand, as in the streaming pipeline used by `ReActAgent.modelCallStream()`, only the first SSE content chunk is delivered promptly. Subsequent chunks are buffered and are not delivered downstream until the upstream response reaches `[DONE]`.

This makes an AG-UI response display its first character and then appear frozen for several seconds, after which the rest of the answer arrives in a burst.

The blocking SSE source currently ends with:

```java
Flux.create(sink -> {
    // Blocking BufferedReader.readLine() loop
    // ...
    sink.next(data);
})
.subscribeOn(Schedulers.boundedElastic());
```

The default overload is equivalent to `subscribeOn(scheduler, true)`. It schedules downstream request signals on the same worker that is occupied by the blocking `readLine()` loop. When the downstream requests items incrementally, later requests cannot run until the read loop exits, so `Flux.create` buffers the remaining chunks.

**To Reproduce**

1. Configure an OpenAI-compatible streaming model that returns SSE chunks over several seconds.
2. Use AgentScope's `OkHttpTransport` through a `ReActAgent` streaming call.
3. Consume the stream through the normal AgentScope/AG-UI streaming pipeline.
4. Observe the processing timestamps for SSE content chunks and the `[DONE]` marker.

Observed before the change:

```text
+0 ms     first content chunk processed
+2959 ms  [DONE] received by OkHttpTransport
+2960 ms  second content chunk processed
```

The second chunk is processed only after `[DONE]`, even though the server streamed it earlier.

**Expected behavior**

Every SSE content chunk should be delivered downstream as it is read, before `[DONE]`, while still keeping the blocking HTTP read off the caller thread.

**Proposed fix**

Disable scheduling downstream requests on the subscription worker:

```java
.subscribeOn(Schedulers.boundedElastic(), false);
```

After applying only this change, the same request produced:

```text
+0 ms     first content chunk processed
+82 ms    second content chunk processed
+3297 ms  [DONE] received by OkHttpTransport
```

All subsequent chunks were delivered progressively before `[DONE]`.

**Error messages**

There is no exception. The symptom is a multi-second pause between the first and second content chunks.

- AgentScope-Java Version: 2.0.2
- Spring Boot Version: 3.5.6
- Java Version: Eclipse Temurin 21.0.11+10
- OS: Windows
- Transport: `OkHttpTransport`
- Model protocol: OpenAI-compatible SSE streaming
- Integration: `agentscope-agui-spring-boot-starter`

**Additional context**

This may be related to the switch from `publishOn` to `subscribeOn` in #410. The issue is specifically the default `requestOnSeparateThread=true` behavior combined with a blocking source and incremental downstream demand.

wmh123wmh1:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #2957

`OkHttpTransport` runs its blocking SSE reader on `boundedElastic`, but the single-argument `subscribeOn` overload also schedules downstream request signals on that same worker. When a downstream operator requests items incrementally, those requests can remain queued behind the blocking `readLine()` loop until `[DONE]` is received.

This change:

- uses `subscribeOn(Schedulers.boundedElastic(), false)` so the blocking subscription remains off the caller thread without queueing downstream demand on the reader worker;
- adds a regression test with a streaming response and cross-thread incremental demand, verifying that the second SSE event is delivered while the reader is still waiting for `[DONE]`.

The regression test times out with the previous implementation and passes with this change.

### Testing

- `mvn -pl agentscope-core -Dtest=OkHttpTransportTest test`
- `mvn -pl agentscope-core test` (2317 tests passed, 9 skipped)

## Checklist

- [x] All relevant tests are passing
- [x] Javadoc comments are complete and follow project conventions (no public API changes)
- [x] Related documentation has been updated (not applicable; internal bug fix)

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/model/transport/OkHttpTransport.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/transport/OkHttpTransport.java
@@ -307,7 +307,7 @@ public class OkHttpTransport implements HttpTransport {
                                 closeQuietly(response);
                             }
                         })
-                .subscribeOn(Schedulers.boundedElastic());
+                .subscribeOn(Schedulers.boundedElastic(), false);
     }
 
     @Override
--- a/agentscope-core/src/test/java/io/agentscope/core/model/transport/OkHttpTransportTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/model/transport/OkHttpTransportTest.java
@@ -21,16 +21,25 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.io.IOException;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import okhttp3.OkHttpClient;
+import okhttp3.Protocol;
+import okhttp3.Response;
+import okhttp3.ResponseBody;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.RecordedRequest;
+import okio.BufferedSink;
+import okio.Okio;
+import okio.Pipe;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import reactor.core.scheduler.Schedulers;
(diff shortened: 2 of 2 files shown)


dailingtao (approved):

jujn (approved):

jujn:
Assign to @dailingtao (Review)

PR #2968, opened 2026-09-03
fix(core): fail model empty-completion before retry

### LINKED ISSUE #2962: [Bug]: Zero-chunk model completion ends the ReAct run as silent success - no error, no retry, no fallback
**Describe the bug**

When a model provider (or an LLM gateway in front of it) returns HTTP 200 with a stream that completes without any content-bearing chunk — e.g. only the `[DONE]` marker, or chunks that all fail to parse — the whole chain treats it as a normal successful completion:

1. `OpenAIClient` streaming: a `[DONE]`-only stream completes the Flux normally (`takeWhile` on the done marker); malformed chunks are parsed to null and silently skipped. No error is ever raised.
2. `ModelUtils.applyTimeoutAndRetry` — the shared wrapper all model implementations route through (openai / gemini / dashscope / ollama / anthropic) — has no empty-completion detection, so the error never enters the existing `retryWhen` / fallback machinery.
3. `ReasoningContext.buildFinalMessage()` returns `null` when no blocks were accumulated.
4. `ReActAgent.reasoning()`: `finalMsg == null` → `Mono.justOrEmpty(null)` → the subsequent `.flatMap` is skipped entirely, so `runPostReasoningPipeline` never executes.
5. The run ends "normally": no WARN, no error event, no fallback model selected, no assistant message persisted. To the caller it looks like the model simply chose to say nothing.

The empty-response reminder loop added for "message with no visible content" (#2750 / #2755) cannot catch this case: it lives inside `runPostReasoningPipeline`, which is unreachable when the final message is null. The two conditions are also semantically different — a thinking-only message is a model-behavior problem (the reminder loop is the right response), while a zero-chunk completion is an upstream transport/service anomaly (error + retry is the right response).

**To Reproduce**

1. Point an `OpenAIChatModel` at a stub endpoint that returns HTTP 200 with body `data: [DONE]\n\n` (or unit-level: invoke a `ReActAgent` whose model returns `Flux.empty()`).
2. Ask a plain question: `agent.call(...)`.
3. The run completes successfully — no assistant message, no error logged, no retry (even with `maxAttempts > 1`), and no fallback model engaged (`switchOnFirst` only switches on an error signal).

Observed in production (AgentScope 2.0.1, OpenAI-compatible endpoint behind an LLM gateway): a run whose chat call returned 200 + zero content chunks in 0.211s ended "successfully" with no reply; a retry two minutes later also got an empty stream (0.146s); a third attempt six minutes later produced a normal 5.26s response — a transient upstream fault, silently swallowed twice with no signal at any layer.

**Expected behavior**

A stream that completes without any content-bearing `ChatResponse` (no `TextBlock` / `ThinkingBlock` / `ToolUseBlock`) is an upstream anomaly, not a valid "the model said nothing" outcome. It should surface as a `ModelException` so the existing retry (`retryWhen`) and fallback chains can engage and the caller sees an error instead of a silent empty reply. `ModelUtils.applyTimeoutAndRetry` is the natural single point for this detection — shared by all model implementations — chained before `retryWhen`.

**Error messages**

None — that is the bug. The only trace is the INFO-level harness log (`AgentTraceMiddleware`: "empty completion (no text, no tool call) — ReAct loop will terminate") acknowledging that the loop terminates silently.

- AgentScope-Java Version: 2.0.1 (production); verified still present on main @ `ea511ec2`
- Java Version: 17
- OS: any
- Provider: OpenAI-compatible endpoint (MiniMax) behind an LLM gateway

**Additional context**

- Distinct from #2750 / #2755: that fix handles "has message, no visible content" and requires a non-null final message; zero-chunk completions bypass `runPostReasoningPipeline` entirely, so the protection has a blind spot by construction.
- Contributing gap, likely worth its own issue: `OpenAIResponse.isError()` does not recognize MiniMax-style `base_resp.status_code != 0` / `input_sensitive` markers, so provider-side filtered responses are indistinguishable from normal empty streams. Related: #2641 asks for status exposure in middleware events.
- Keeping the two empty-outcome conditions separate matters for the fix: folding zero-chunk into the reminder loop would waste a full LLM call re-prompting what is actually a transport fault.

Buktal:
## AgentScope-Java Version

main @ `ea511ec2` (defect observed in production on 2.0.1)

## Description

Fixes #2962

When a model provider (or an LLM gateway in front of it) returns HTTP 200 with a stream that completes without any content-bearing chunk — e.g. only the `[DONE]` marker, or only well-formed chunks whose content parses to empty — the entire chain treated it as a normal successful completion: no error, no retry, no fallback, and no assistant message persisted. To the caller it looked like the model simply chose to say nothing.

### Changes

**`ModelUtils.applyTimeoutAndRetry`** — the shared path all model implementations route through — now detects empty completions and converts them to a `ModelException` before `retryWhen`:

- **Detection**: `ensureNonEmptyCompletion` (new private helper) rejects a stream that completes without any chunk carrying a non-null, non-empty `getContent()`. Any content block counts (multimodal blocks included); only a fully contentless stream is rejected.
- **Withholding**: contentless chunks are withheld until the first content-bearing chunk arrives, then released in order. This guarantees the error is the *first* downstream signal on an empty completion: `switchOnFirst`-based fallback commits on the first signal, and the common OpenAI-compatible empty shape (a role-only first chunk, then `[DONE]`) would otherwise emit a contentless chunk first and bypass the fallback model entirely. Healthy streams are unaffected — withheld chunks are released the moment content arrives, with order and metadata preserved.
- **Placement**: after timeout (a timed-out request goes straight to error) and before retry (an empty completion triggers retry like any other model error).
- **Unconditional**: the check applies regardless of retry configuration, so a bare `agent.call(...)` without an `ExecutionConfig` still surfaces the error instead of ending as silent success.
- **Hardening**: `applyTimeoutAndRetry` no longer NPEs when both `options` and `defaultOptions` are null — `GenerateOptions.mergeOptions` documents that it can return null, and both parameters are documented as nullable.

### Tests

**`ModelTimeoutRetryTest`** — 6 tests:

1. Zero chunks (`Flux.empty()`, the `[DONE]`-only case) → error.
2. Two contentless chunks → error **as the first downstream signal** (no chunk is emitted before it, so a `switchOnFirst`-based fallback still engages).
3. Empty first attempt, success on retry → retry engages, `attemptCount == 2`.
4. Content-then-mid-stream-error followed by an empty retry attempt → still fails (`RetryExhaustedException` wrapping `ModelException`).
5. Image-only (multimodal) chunks → completes normally.
6. Leading contentless chunks are released in order once content arrives; trailing contentless chunks pass through.

**`ReActAgentTest`**: primary model returns an empty completion (contentless chunks, then complete) → fallback model takes over and answers.

**Provider extension tests**: mocks that stubbed contentless responses (`choices: []`, or streaming bodies without any content-bearing delta) were updated to return realistic content-bearing responses, since a contentless completion is now an error by design (`OpenAIChatModelTest`, `DashScopeChatModelTest`, `DashScopeNonStreamingBlockingBehaviorTest`).

## Checklist

- [x] All tests are passing (`mvn test -pl agentscope-core`, plus the full `agentscope-extensions-model` reactor)
- [x] Related documentation has been updated (n/a — internal behavior change, covered by Javadoc)

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
@@ -16,10 +16,14 @@
 package io.agentscope.core.model;
 
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Predicate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 import reactor.util.retry.Retry;
 
 /**
@@ -42,6 +46,17 @@ public final class ModelUtils {
      * <p>This method wraps the original Flux with timeout and retry operators based on the
      * configuration in GenerateOptions. Both timeout and retry are optional and only applied if
      * configured.
+     *
+     * <p><b>Empty-completion detection:</b> A stream that completes without any content-bearing
+     * {@code ChatResponse} (no chunk carries a non-null, non-empty content list) is converted to
+     * a {@code ModelException} after the timeout check and before retry. Contentless chunks are
+     * withheld until the first content-bearing chunk arrives (then released in order), so an
+     * empty completion surfaces the error as the first downstream signal and switchOnFirst-based
+     * fallback models engage. The check is deliberately type-agnostic: any content block counts,
+     * so multimodal completions carrying only image/video/audio blocks are not misclassified as
+     * empty. This detection is unconditional (applies regardless of retry configuration), since
+     * an empty completion is an upstream transport anomaly that should always surface as an
+     * error.
      *
      * <p><b>Timeout Behavior:</b>
      * <ul>
@@ -77,9 +92,11 @@ public final class ModelUtils {
         GenerateOptions effectiveOptions = GenerateOptions.mergeOptions(options, defaultOptions);
 
         // Extract execution config
-        ExecutionConfig execConfig = effectiveOptions.getExecutionConfig();
+        ExecutionConfig execConfig =
+                effectiveOptions != null ? effectiveOptions.getExecutionConfig() : null;
+
+        // Apply timeout if configured
         if (execConfig != null) {
-            // Apply timeout if configured
             Duration timeout = execConfig.getTimeout();
             if (timeout != null) {
                 responseFlux =
@@ -92,8 +109,12 @@ public final class ModelUtils {
                                                 provider)));
                 LOG.debug("Applied timeout: {} for model: {}", timeout, modelName);
             }
+        }
 
-            // Apply retry if configured (maxAttempts > 1 means retry is enabled)
+        responseFlux = ensureNonEmptyCompletion(responseFlux, modelName, provider);
+
+        // Apply retry if configured
+        if (execConfig != null) {
             Integer maxAttempts = execConfig.getMaxAttempts();
             if (maxAttempts != null && maxAttempts > 1) {
                 Duration initialBackoff = execConfig.getInitialBackoff();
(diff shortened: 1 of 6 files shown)


oss-maintainer:

```bash
git fetch origin
git checkout fix/2962-empty-completion-silent-success
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

Code review is not blocked by this — I will re-review the new head once it is pushed. This is a one-time reminder; @mention me if you want a re-review sooner.

---

agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
```diff
@@ -92,8 +109,12 @@ public static Flux<ChatResponse> applyTimeoutAndRetry(
                                                 provider)));
                 LOG.debug("Applied timeout: {} for model: {}", timeout, modelName);
             }
+        }
 
-            // Apply retry if configured (maxAttempts > 1 means retry is enabled)
+        responseFlux = ensureNonEmptyCompletion(responseFlux, modelName, provider);
```
oss-maintainer, line 114: **[Warning]** This guard is applied unconditionally to *every* caller of `applyTimeoutAndRetry`, i.e. all five chat adapters (dashscope / openai / anthropic / gemini / ollama) plus anything that reuses the helper. A completion that used to terminate the stream normally now surfaces as a `ModelException`, and when `maxAttempts > 1` it also consumes retry budget (extra billed calls) for what may be a deterministic provider response — e.g. a content-filter refusal that legitimately returns no content blocks.

Consider making it opt-out-able via `GenerateOptions`/`ExecutionConfig` (defaulting to the new behaviour is fine) so callers that treat an empty completion as a valid outcome can keep it, and so the behaviour change is discoverable per call site rather than global-only.

agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
```diff
@@ -138,6 +159,83 @@ public static Flux<ChatResponse> applyTimeoutAndRetry(
...
+                                                if (released.get()) {
+                                                    return Mono.empty();
+                                                }
+                                                LOG.warn(
+                                                        "Model {} ({}) returned an empty completion"
+                                                                + " (zero content-bearing chunks)",
+                                                        modelName,
+                                                        provider);
+                                                return Mono.error(
+                                                        new ModelException(
```
oss-maintainer, line 216: **[Warning]** When the stream is rejected here, every withheld chunk is dropped, so any `usage` / `finishReason` / metadata the provider did send (OpenAI-compatible endpoints commonly send a final usage-only chunk) is lost: the caller cannot account tokens for the attempt and cannot see the provider's finish reason, which is exactly the diagnostic you want when debugging an empty completion.

Suggest keeping the last non-null `usage` / `finishReason` observed while withholding and either logging it alongside the warn or exposing it on the exception (e.g. a dedicated `EmptyCompletionException` carrying the finish reason), so triage does not require a provider-side replay.

agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
```diff
@@ -138,6 +159,83 @@ public static Flux<ChatResponse> applyTimeoutAndRetry(
...
+    /**
+     * Returns whether the chunk carries any content block.
+     *
+     * <p>Type-agnostic by design: multimodal completions may carry only image/video/audio
+     * blocks, which are valid completions, not empty ones.
+     *
+     * @param chunk the response chunk to inspect
+     * @return true if the chunk carries at least one content block
+     */
+    private static boolean isContentBearing(ChatResponse chunk) {
```
oss-maintainer, line 235: **[Info]** The type-agnostic rule (`any content block counts`) is the right default, but it makes an implicit assumption per adapter: that a chunk carrying only a *partial* tool-call argument fragment still exposes a non-empty `getContent()`. A provider chunk that surfaces tool-call fragments through metadata rather than a `ToolUseBlock` would be classified as empty and the error would fire mid-stream.

Worth a one-line note (or a parameterized test) per adapter covering "tool-call-only streaming response" so the assumption is pinned down — `DashScopeResponseParser` / `OpenAIResponseParser` build `ToolUseBlock`s, but the empty-completion path here is shared by all of them.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/test/java/io/agentscope/extensions/model/dashscope/DashScopeChatModelTest.java
```diff
@@ -556,19 +556,16 @@ void testDoStreamWithAdditionHeadersAndParams() throws Exception {
         MockWebServer mockServer = new MockWebServer();
         mockServer.start();
 
+        // The response must carry content: contentless completions are rejected as upstream
```
oss-maintainer, line 559: **[Info]** This converts the existing mock from a non-streaming JSON body to an SSE body to make the chunk content-bearing, so the previous non-streaming-with-`application/json` code path is no longer exercised by this case, and the `choices: []` shape is no longer asserted at all.

Suggest keeping both: (a) a non-streaming JSON case that returns real content, and (b) an explicit `choices: []` case asserting the new `ModelException`, so the intentional behaviour change is pinned by a test instead of being sidestepped by the fixture rewrite. Same for the `OpenAIChatModelTest` / `DashScopeNonStreamingBlockingBehaviorTest` updates.

oss-maintainer (commented):
### Summary
Guards model completions that carry zero content-bearing chunks by withholding contentless chunks and surfacing a `ModelException` before retry, so switchOnFirst-based fallback engages and empty completions retry like any other model error. The per-subscription state via `Flux.defer` is the right call (retry resubscribes cleanly, no leakage between attempts), the type-agnostic content check avoids misclassifying multimodal completions, and the test matrix (zero chunks / all-empty / retry-then-succeed / mid-stream failure then empty / multimodal / leading-contentless ordering) covers the interesting cases well. Not approving yet mostly for the scope question below: the guard is unconditional across every adapter, which makes an empty completion a billed-retry-and-error for callers that may treat emptiness as a legitimate result.

Two smaller points: usage/finish-reason data is dropped on the rejected path (hurts triage of issue #2962), and the existing non-streaming fixtures were rewritten to carry content rather than asserting the new failure shape.

Also flagging: this PR currently has merge conflicts with `main` and cannot be merged as-is — see the earlier conflict reminder.

---

oss-maintainer:

```bash
git fetch origin
git checkout fix/2962-empty-completion-silent-success
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #2965, opened 2026-09-03
fix(rag): preserve blank lines between paragraphs in WordReader

pure11011:
## AgentScope-Java Version

2.0.0 (also reproduced on 1.0.12; `WordReader.java` is unchanged on `main`)

## Description

#2964.

**Problem.** An empty `<w:p>` element is how Word represents a blank line. `getDataBlocks()` discarded it: `extractTextFromParagraph()` returns `""` for such a paragraph (there are no runs, so both the `para.getText()` path and the run-concatenation fallback yield an empty string), and the `if (text != null && !text.isEmpty())` guard then skipped the whole branch without writing a character or setting a flag. The fact that a paragraph existed there became unrecoverable, so `A / blank line / B` came out as `"A\nB"` with the blank line silently gone.

**Change.** Append a `"\n"` when an empty paragraph is encountered. The following paragraph is merged with another `"\n"`, so the two together reproduce the blank line the user typed, and consecutive blank paragraphs are preserved one for one.

**Testing**

Fixtures are authored in memory with `XWPFDocument` and written to `@TempDir`, so no binary test resource is added to the repository.

1. One blank paragraph between two text paragraphs, under the default `PARAGRAPH` strategy → `"Paragraph one.\n\nParagraph two."`
2. Two consecutive blank paragraphs, under `CHARACTER` → `"Paragraph one.\n\n\nParagraph two."`

The second case uses `CHARACTER` deliberately. `chunkByParagraph()` splits on `\n\s*\n` and re-joins the pieces with a fixed `"\n\n"`, so under `PARAGRAPH` any run of blank lines is normalised back to a single one and the preserved count cannot be observed through the public API. `CHARACTER` returns the extracted text verbatim, which is what makes the assertion meaningful.

Full module suite after the change: 446 tests, 0 failures, 0 errors.

## Checklist

Please check the following items before code is ready to be reviewed.

- [x]  Code has been formatted with `mvn spotless:apply`
- [x]  All tests are passing (`mvn test`)
- [x]  Javadoc comments are complete and follow project conventions
- [x]  Related documentation has been updated (e.g. links, examples, etc.) — n/a, no public API or documented behaviour surface changed
- [x]  Code is ready for review

PR #2973, opened 2026-09-03
fix(rag): preserve paragraph boundaries across empty paragraphs in WordReader

### LINKED ISSUE #2964: [Bug]: WordReader collapses paragraph boundaries into a single \n, making SplitStrategy.PARAGRAPH silently degrade to character chunking
**Describe the bug**

`WordReader.getDataBlocks()` joins consecutive paragraphs with a single `"\n"` and skips empty paragraphs entirely.

In WordprocessingML every `<w:p>` is a paragraph in its own right, and a blank line is an empty `<w:p>` element with no `<w:r>` child. Pressing Enter twice in Word produces:

```xml
<w:p><w:r><w:t>Paragraph one.</w:t></w:r></w:p>
<w:p><w:pPr>...</w:pPr></w:p>          <!-- blank line: no <w:r> at all -->
<w:p><w:r><w:t>Paragraph two.</w:t></w:r></w:p>
```

For the middle element `extractTextFromParagraph()` returns `""` (there are no runs, so both the `para.getText()` path and the run-concatenation fallback yield an empty string). The `if (text != null && !text.isEmpty())` guard then skips the whole branch — no characters are written and no flag is set. **The fact that a paragraph existed here is destroyed at that point and cannot be recovered downstream.**

The result is `"Paragraph one.\nParagraph two."`: the blank line is gone, and a paragraph boundary has become indistinguishable from an in-paragraph line break.

**Why this matters beyond formatting**

`TextChunker` recognises paragraph boundaries only via:

```java
private static final Pattern PARAGRAPH_SEPARATOR = Pattern.compile("\\n\\s*\\n");
```

This requires **two** newlines. Since `getDataBlocks()` never emits `\n\n`, `PARAGRAPH_SEPARATOR.split(text)` always returns a single element covering the whole document. `chunkByParagraph()` therefore loops once, inevitably falls into the `currentChunk.length() > chunkSize` branch, and delegates to `chunkByCharacter()`.

In other words, **`SplitStrategy.PARAGRAPH` — the default for `WordReader` — has never actually performed paragraph-aware chunking.** The degradation is silent: no exception, no warning, and a non-empty chunk list is still returned.

**To Reproduce**

```java
List<Document> docs = new WordReader()   // defaults: chunkSize=512, PARAGRAPH, overlap=50
        .read(ReaderInput.fromPath(Paths.get("your.docx")))
        .block();
for (Document d : docs) {
    System.out.println(d.getMetadata().getContentText().length());
}
```

Measured on two real-world documents (Chinese corporate policy documents, 53 and 36 body paragraphs):

| Document | `\n\s*\n` matches | chunk sizes | sample cut point |
|---|---|---|---|
| A (53 paragraphs, 8 blank, 1 table) | 1 | 512 / 512 / 512 / 286 | mid-token |
| B (36 paragraphs, 0 blank) | 1 | 512 / 512 / 432 | mid-sentence |

With the paragraph separator changed to `"\n\n"`:

| Document | `\n\s*\n` matches | chunk sizes |
|---|---|---|
| A | 46 (= 45 non-empty paragraphs + 1 table) | 506 / 509 / 495 / 357 |
| B | 36 | 512 / 498 / 481 |

**Expected behavior**

A `<w:p>` boundary should be encoded as a blank line (`\n\n`), so that the paragraph structure of the source document survives into the chunking stage and `SplitStrategy.PARAGRAPH` actually takes effect.

**Error messages**

None. No exception is thrown — that is precisely the problem: the degradation is entirely silent.

**Environment**

- AgentScope-Java Version: 2.0.0 (also reproducible on 1.0.12). `WordReader.java` is byte-identical across 1.0.12, 2.0.0 and current `main` (md5 `397960db581a057292110d352b40bab7`, 20557 bytes), so `main` is affected as well
- Java Version: 21
- OS: macOS

lvcccc-sys:
Fixes #2964

## Problem

`WordReader.getDataBlocks()` skips empty `<w:p>` paragraphs (blank lines in Word) entirely:

```java
String text = extractTextFromParagraph(para);
if (text != null && !text.isEmpty()) {   // empty paragraph -> skipped, blank line lost
```

Consecutive non-empty paragraphs are therefore joined with a single `"\n"`. Meanwhile `TextChunker` detects paragraph boundaries only via `Pattern.compile("\\n\\s*\\n")` (two newlines). Since the reader never emits `\n\n`, `SplitStrategy.PARAGRAPH` — the default for `WordReader` — silently degrades to character chunking: the document becomes one giant "paragraph", exceeds the chunk size, and falls back to `chunkByCharacter()`.

## Fix

When an empty paragraph is encountered, preserve the boundary by appending a line feed to the current text block, so adjacent paragraphs yield `\n\n` and paragraph-aware chunking actually works. Image and table handling are untouched.

```java
} else {
    // Empty paragraph -> keep the <w:p> boundary as "\n\n"
    if ("text".equals(lastType)) {
        TextBlock lastBlock = (TextBlock) blocks.get(blocks.size() - 1);
        blocks.set(blocks.size() - 1,
                TextBlock.builder().text(lastBlock.getText() + "\n").build());
    }
}
```

## Test

Adds `WordReaderTest.testParagraphBoundaryPreservedAcrossBlankLine`, which builds a docx containing `paragraph -> blank line -> paragraph`, reads it with `SplitStrategy.PARAGRAPH`, and asserts the output is `"First paragraph.\n\nSecond paragraph."`. This fails on the previous code (which yields a single `\n`).

### DIFF
--- a/agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/reader/WordReader.java
+++ b/agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/reader/WordReader.java
@@ -243,6 +243,18 @@ public class WordReader extends AbstractChunkingReader {
                             blocks.add(TextBlock.builder().text(text).build());
                         }
                         lastType = "text";
+                    } else {
+                        // Empty paragraph (e.g. a blank line in Word, represented by an empty
+                        // <w:p> element). Preserve the paragraph boundary by appending a line
+                        // feed; otherwise adjacent non-empty paragraphs are joined with a single
+                        // "\n" and TextChunker's PARAGRAPH strategy (separated by "\n\s*\n")
+                        // cannot detect the boundary, degrading to character splitting.
+                        if ("text".equals(lastType)) {
+                            TextBlock lastBlock = (TextBlock) blocks.get(blocks.size() - 1);
+                            blocks.set(
+                                    blocks.size() - 1,
+                                    TextBlock.builder().text(lastBlock.getText() + "\n").build());
+                        }
                     }
 
                 } else if (element.getElementType() == BodyElementType.TABLE) {
--- a/agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/test/java/io/agentscope/core/rag/reader/WordReaderTest.java
+++ b/agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/test/java/io/agentscope/core/rag/reader/WordReaderTest.java
@@ -21,7 +21,13 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import io.agentscope.core.rag.exception.ReaderException;
+import io.agentscope.core.rag.model.Document;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
 import java.util.List;
+import org.apache.poi.xwpf.usermodel.XWPFDocument;
+import org.apache.poi.xwpf.usermodel.XWPFParagraph;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


lvcccc-sys (commented):
Reviewed the fix — it correctly addresses #2964 and the logic is sound.

**What it does**
\texttt{WordReader} was skipping empty `<w:p>` paragraphs entirely, so adjacent non-empty paragraphs were joined with a single `\n`, which `TextChunker`'s `PARAGRAPH` strategy (separated by `\n\s*\n`) cannot detect — silently degrading word text splitting to character chunking. Appending a `\n` to the previous `TextBlock` on an empty paragraph correctly yields `\n\n`, restoring the paragraph boundary.

**Correctness check against the surrounding code**
- Empty paragraph at the very start of the document: `lastType` is `null`, so the branch is a no-op — no leading blank/dangling newline. Good.
- Empty paragraph following a table (with `separateTable=false`): the append is guarded by `"text".equals(lastType)`, so tables/images are untouched. Good.
- `lastBlock.getText()` cannot be null here: the block was only created with a non-empty `text` in the preceding branch.

**One minor observation (not blocking)**
Multiple consecutive blank lines each append one `\n` (e.g. `First\n\n\nSecond`), losing the distinction between one blank line and several. This is acceptable for chunking — `\n\s*\n` still matches — but worth noting if preserving exact line spacing ever matters.

**Test**
`testParagraphBoundaryPreservedAcrossBlankLine` is valid: POI's `createParagraph()` writes a real empty `<w:p>`, `extractTextFromParagraph` returns `""` for it, and the assertion fails against the old code (single `\n`). Confirmed the test exercises the new branch.

Overall: fix and test look correct. Left a detailed analysis above; main external blocker is the `license/cla` check (pending) — please recheck the CLA so CI can run.

jujn:
fixed int: https://github.com/agentscope-ai/agentscope-java/pull/2965

PR #2976, opened 2026-09-04
fix(sandbox-e2b): reliable exec timeout with server-side kill

### LINKED ISSUE #2974: [sandbox-e2b] exec timeout is unreliable: readTimeout preempts callTimeout, no server-side kill, cancellation misreported
## Problem

`E2bEnvdProcessClient.runShellCapture()` timeout handling (`agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java`) has three defects:

1. **Premature false timeout.** The per-call client only sets `callTimeout(timeoutSeconds)` and inherits the base `readTimeout`:

```java
http.newBuilder().callTimeout(timeoutSeconds, TimeUnit.SECONDS).build()
```

If base `readTimeout < timeout` (e.g. custom client `readTimeout=1s`, exec `timeout=3s`), the idle `readTimeout` fires first. The resulting `InterruptedIOException` is caught and wrapped as `ExecTimeoutException("timed out after 3s")`, but only ~1s elapsed.

2. **Timed-out remote process is left running.** `buildEnvdRequest()` sends no `Connect-Timeout-Ms` header, so timeout only disconnects the client HTTP stream. The sandbox process (e.g. `sleep 1000`) keeps running — leak / unexpected billing.

3. **External cancellation misreported as timeout.** `catch (InterruptedIOException)` unconditionally throws `ExecTimeoutException`, even when the thread was interrupted by outer cancellation. The interrupt bit is swallowed.

## Repro

- Base client `readTimeout=1s`, `runShell(..., timeoutSeconds=3)` against a hanging envd stream (`NO_RESPONSE`).
- Expected: `ExecTimeoutException` after ~3s.
- Actual: `ExecTimeoutException` after ~1s with message claiming 3s.

## Expected

- `readTimeout` for the call must stay above `callTimeout` so total-duration semantics wins.
- Send `Connect-Timeout-Ms` when `timeoutSeconds > 0` so envd kills the process server-side.
- If `Thread.currentThread().isInterrupted()`, restore the bit and rethrow instead of wrapping as timeout.

chcodex:
Fixes #2974.

- Per-call client disables the idle read timeout (`readTimeout=0`) so total-duration semantics is owned solely by `callTimeout`; previously a small base `readTimeout` preempted `callTimeout` and misreported a short idle stall as a full exec timeout (both surface as `InterruptedIOException`).
- Send `Connect-Timeout-Ms` so envd kills the remote process server-side; previously a client timeout only dropped our HTTP stream and leaked the process.
- External cancellation (`Thread.interrupt()`) is rethrown as-is with the interrupt bit restored instead of being wrapped as `ExecTimeoutException`.
- `timeoutSeconds <= 0` now fails fast with `IllegalArgumentException`.
- Tests: `E2bEnvdProcessClientTest` 10/10 pass (4 new: header carried end-to-end, interruption rethrown with bit restored, fail-fast without request; plus existing codec/drain cases). Full reactor `test` BUILD SUCCESS.

### DIFF
--- a/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
+++ b/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
@@ -119,10 +119,19 @@ final class E2bEnvdProcessClient {
     private ShellCapture runShellCapture(
             E2bSandboxState state, String cwd, String shellCommand, int timeoutSeconds)
             throws Exception {
+        if (timeoutSeconds <= 0) {
+            throw new IllegalArgumentException(
+                    "timeoutSeconds must be positive: " + timeoutSeconds);
+        }
         OkHttpClient callClient =
-                timeoutSeconds > 0
-                        ? http.newBuilder().callTimeout(timeoutSeconds, TimeUnit.SECONDS).build()
-                        : http;
+                http.newBuilder()
+                        .callTimeout(timeoutSeconds, TimeUnit.SECONDS)
+                        // Disable the idle read timeout: it fires on gaps between bytes and
+                        // would preempt callTimeout (both surface as InterruptedIOException),
+                        // misreporting a short idle stall as a full exec timeout (#2974).
+                        // Total-duration semantics is owned solely by callTimeout.
+                        .readTimeout(0, TimeUnit.SECONDS)
+                        .build();
         String host = envdHost(state);
         String url = host + "/process.Process/Start";
         byte[] envelope = encodeStartRequestEnvelope(shellCommand, cwd);
@@ -134,7 +143,10 @@ final class E2bEnvdProcessClient {
                         .addHeader("User-Agent", "agentscope-java-e2b")
                         .addHeader("E2b-Sandbox-Id", state.getSandboxId())
                         .addHeader("E2b-Sandbox-Port", Integer.toString(ENVD_PORT))
-                        .addHeader("Authorization", basicAuthUser(opt.getRunUser()));
+                        .addHeader("Authorization", basicAuthUser(opt.getRunUser()))
+                        // Ask envd to kill the remote process when this lapses; otherwise a
+                        // client-side timeout only drops our HTTP stream and leaks the process.
+                        .addHeader("Connect-Timeout-Ms", String.valueOf(timeoutSeconds * 1000L));
         if (state.getEnvdAccessToken() != null && !state.getEnvdAccessToken().isBlank()) {
             rb.addHeader("X-Access-Token", state.getEnvdAccessToken());
         }
@@ -145,7 +157,7 @@ final class E2bEnvdProcessClient {
         int exit;
         try (Response res = callClient.newCall(req).execute()) {
             if (!res.isSuccessful()) {
-                String err = res.body() != null ? res.body().string() : "";
+                String err = res.body().string();
                 throw new SandboxException.SandboxRuntimeException(
                         SandboxErrorCode.WORKSPACE_START_ERROR,
                         "envd Start failed HTTP " + res.code() + ": " + err);
@@ -154,6 +166,12 @@ final class E2bEnvdProcessClient {
                 exit = drainStartStream(in, stdout, stderr);
             }
         } catch (InterruptedIOException e) {
+            // External cancellation surfaces here too; don't misreport it as a timeout
+            // and don't swallow the interrupt bit.
+            if (Thread.currentThread().isInterrupted()) {
+                Thread.currentThread().interrupt();
+                throw e;
+            }
             throw new SandboxException.ExecTimeoutException(shellCommand, timeoutSeconds);
         }
         return new ShellCapture(exit, stdout, stderr);
@@ -179,7 +197,7 @@ final class E2bEnvdProcessClient {
                 break;
             }
             int len = ByteBuffer.wrap(lenB).order(ByteOrder.BIG_ENDIAN).getInt() & 0x7FFFFFFF;
-            if (len < 0 || len > 64 * 1024 * 1024) {
+            if (len > 64 * 1024 * 1024) {
                 throw new IOException("Invalid connect frame length: " + len);
             }
             byte[] data = in.readNBytes(len);
--- a/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/test/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClientTest.java
+++ b/agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/test/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClientTest.java
@@ -17,23 +17,30 @@ package io.agentscope.extensions.sandbox.e2b;
 
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.DynamicMessage;
+import io.agentscope.harness.agent.sandbox.SandboxException;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.net.SocketTimeoutException;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
+import java.util.concurrent.atomic.AtomicReference;
+import okhttp3.Interceptor;
 import okhttp3.MediaType;
+import okhttp3.OkHttpClient;
 import org.junit.jupiter.api.Test;
 
 class E2bEnvdProcessClientTest {
@@ -306,5 +313,75 @@ class E2bEnvdProcessClientTest {
         E2bSandboxClientOptions options = new E2bSandboxClientOptions();
         options.setCodec(codec);
         return options;
+    }
+
+    @Test
+    void requestCarriesConnectTimeoutMsHeader() throws Exception {
+        AtomicReference<String> header = new AtomicReference<>();
+        Interceptor capture =
+                chain -> {
+                    header.set(chain.request().header("Connect-Timeout-Ms"));
+                    throw new SocketTimeoutException("timeout");
+                };
+        E2bEnvdProcessClient client = clientWithInterceptor(capture);
+
+        assertThrows(
+                SandboxException.ExecTimeoutException.class,
+                () -> client.runShell(state(), "/workspace", "sleep 1000", 3));
+        assertEquals("3000", header.get());
+    }
+
+    @Test
+    void interruptionIsRethrownNotWrappedAsTimeout() throws Exception {
+        Interceptor interrupting =
+                chain -> {
+                    Thread.currentThread().interrupt();
+                    throw new SocketTimeoutException("read timed out");
+                };
+        E2bEnvdProcessClient client = clientWithInterceptor(interrupting);
+        try {
+            SocketTimeoutException thrown =
+                    assertThrows(
+                            SocketTimeoutException.class,
+                            () -> client.runShell(state(), "/workspace", "sleep 1000", 3));
+            assertEquals("read timed out", thrown.getMessage());
+            assertTrue(Thread.currentThread().isInterrupted(), "interrupt bit must be restored");
+        } finally {
+            // Do not leak the interrupt bit into other tests.
+            Thread.interrupted();
+        }
+        assertFalse(Thread.currentThread().isInterrupted());
+    }
+
+    @Test
+    void nonPositiveTimeoutFailsFastWithoutRequest() throws Exception {
+        AtomicReference<String> header = new AtomicReference<>();
+        Interceptor capture =
+                chain -> {
+                    header.set(chain.request().header("Connect-Timeout-Ms"));
+                    throw new SocketTimeoutException("must not be called");
+                };
+        E2bEnvdProcessClient client = clientWithInterceptor(capture);
+
+        assertThrows(
+                IllegalArgumentException.class,
+                () -> client.runShell(state(), "/workspace", "echo hi", 0));
+        assertThrows(
+                IllegalArgumentException.class,
+                () -> client.runShell(state(), "/workspace", "echo hi", -5));
+        assertNull(header.get(), "no HTTP request must be issued for invalid timeout");
+    }
+
+    private static E2bEnvdProcessClient clientWithInterceptor(Interceptor interceptor)
+            throws Exception {
+        E2bSandboxClientOptions opt = options(E2bCodec.PROTO);
+        opt.setHttpClient(new OkHttpClient.Builder().addInterceptor(interceptor).build());
+        return new E2bEnvdProcessClient(opt);
+    }
+
+    private static E2bSandboxState state() {
+        E2bSandboxState state = new E2bSandboxState();
+        state.setSandboxId("test-sandbox");
+        return state;
     }
 }


CryoThrust (commented):
One small robustness concern: `res.body().string()` is now unconditional in the non-success response path. OkHttp permits a response with a null body (for example a 204/transport edge or a test double), so this can replace the intended `SandboxRuntimeException` with an NPE. Keeping the previous null-safe fallback (`res.body() != null ? res.body().string() : ""`) would preserve the error contract while still including the server body when present. A regression test with a non-success response and no body would pin this.

chcodex:
@CryoThrust Thanks for flagging this. The null guard was removed deliberately:

- `Response.body` is a non-null contract in OkHttp (Kotlin `val body: ResponseBody`). For a real `Call.execute()` response — the only path that reaches this code — an empty body (204/304, truncated gateway reply) arrives as `EmptyResponseBody`, so `.string()` returns `""` rather than throwing NPE.
- A null body can only come from a hand-built `Response.Builder.build()` without setting a body, which cannot occur on this production path. IDEA's nullability inference (from OkHttp's annotations) agrees and reports `res.body() != null` as always-true, i.e. the guard is dead code.
- The suggested regression test (non-success response with no body) would only exercise such a hand-built double, not a real server reply, so it would pin behavior that cannot happen in production.

oss-maintainer (commented):
### Summary
The timeout model is right: `readTimeout(0)` hands total-duration ownership to `callTimeout`, `Connect-Timeout-Ms` stops leaking the remote process, and the interrupt bit is no longer swallowed. CI is green, CLA signed, head mergeable — nothing blocking.

Five points, in priority order:
1. **[Warning]** `Thread.currentThread().isInterrupted()` (line 171) reads ambient thread state, so a stale interrupt bit can turn a genuine exec timeout into an unmapped `InterruptedIOException`. Snapshot/clear the bit around the call instead.
2. **[Warning]** `timeoutSeconds <= 0` now throws `IllegalArgumentException` where it previously meant "unbounded"; that type is not mapped by `SandboxBackedFilesystem.execute()` and diverges from the daytona/agentrun backends, which clamp.
3. **[Warning]** Client `callTimeout` and the server `Connect-Timeout-Ms` deadline are identical, so on expiry the client still wins the race and the partial stdout/stderr already drained is discarded.
4. **[Info]** On the `res.body()` null-guard exchange with @CryoThrust: the production-path argument is convincing for real `Call.execute()` responses — `Objects.requireNonNull(res.body(), ...)` documents that without reintroducing the ternary, and survives a NullAway/ErrorProne rollout or a synthesising interceptor.
5. **[Info]** Good coverage on the header and the fail-fast path; a `MockWebServer` stall case is the one thing that would actually pin `readTimeout(0)` itself, since the app-level interceptors short-circuit before OkHttp timeouts engage.

---

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
```diff
@@ -119,10 +119,19 @@ byte[] runShellBinaryStdout(
     private ShellCapture runShellCapture(
             E2bSandboxState state, String cwd, String shellCommand, int timeoutSeconds)
             throws Exception {
+        if (timeoutSeconds <= 0) {
+            throw new IllegalArgumentException(
+                    "timeoutSeconds must be positive: " + timeoutSeconds);
```
oss-maintainer, line 124: **[Warning]** Fail-fast on `timeoutSeconds <= 0` changes the effective contract, and it diverges from the sibling sandbox implementations.

`AbstractBaseSandbox.exec()` only substitutes a default when the argument is `null` (`int timeout = timeoutSeconds != null ? timeoutSeconds : getDefaultExecTimeoutSeconds()`), so an explicit `0` used to mean "no `callTimeout`" and now aborts before the request is issued. `DaytonaHttp.execute()` (`Math.max(1, timeoutSeconds)`) and `AgentRunMcpChannel.exec()` (`Math.max(1, ...)`) clamp instead of throwing, so the same input now behaves differently per backend.

It also escapes the tool-layer error mapping: `SandboxBackedFilesystem.execute()` catches `SandboxException.ExecTimeoutException` / `ExecException` only, so an `IllegalArgumentException` surfaces as an unmapped failure rather than an `ExecuteResponse` with a non-zero exit code.

Either clamp to keep the backends consistent, or throw a `SandboxException` so the harness renders it as a tool error:
```java
if (timeoutSeconds <= 0) {
    timeoutSeconds = getDefaultExecTimeoutSeconds(); // or throw SandboxException.ExecException
}
```
If throwing is deliberate, it is worth a line in the PR description plus a javadoc note on `runShell`/`runShellBinaryStdout`, since both are called with hardcoded constants today.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
```diff
@@ -134,7 +143,10 @@ private ShellCapture runShellCapture(
                         .addHeader("User-Agent", "agentscope-java-e2b")
                         .addHeader("E2b-Sandbox-Id", state.getSandboxId())
                         .addHeader("E2b-Sandbox-Port", Integer.toString(ENVD_PORT))
-                        .addHeader("Authorization", basicAuthUser(opt.getRunUser()));
+                        .addHeader("Authorization", basicAuthUser(opt.getRunUser()))
+                        // Ask envd to kill the remote process when this lapses; otherwise a
+                        // client-side timeout only drops our HTTP stream and leaks the process.
+                        .addHeader("Connect-Timeout-Ms", String.valueOf(timeoutSeconds * 1000L));
```
oss-maintainer, line 149: **[Warning]** The server-side deadline and the client-side `callTimeout` are set to exactly the same instant, so on timeout the two abort races and the client usually wins — which means the process *is* reaped server-side (good, that is #2974) but the terminal frames from envd are still lost, and the caller sees a generic `ExecTimeoutException` without the partial output/exit status the server already flushed.

Give the server a small head start so its close frame can arrive first:
```java
long timeoutMs = TimeUnit.SECONDS.toMillis(timeoutSeconds);
rb.addHeader("Connect-Timeout-Ms", String.valueOf(timeoutMs));
// client waits slightly longer than envd's own deadline
OkHttpClient callClient = http.newBuilder()
        .callTimeout(timeoutMs + 5_000L, TimeUnit.MILLISECONDS)
        .readTimeout(0, TimeUnit.SECONDS)
        .build();
```
Also `timeoutSeconds * 1000L` overflows silently for very large values from a caller-supplied timeout — `TimeUnit.SECONDS.toMillis(...)` saturates instead. Note `TAR_TIMEOUT_SECONDS` and the 120s workspace calls are unaffected; this only matters for the LLM-supplied path.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
```diff
@@ -154,6 +166,12 @@ private ShellCapture runShellCapture(
                 exit = drainStartStream(in, stdout, stderr);
             }
         } catch (InterruptedIOException e) {
+            // External cancellation surfaces here too; don't misreport it as a timeout
+            // and don't swallow the interrupt bit.
+            if (Thread.currentThread().isInterrupted()) {
```
oss-maintainer, line 171: **[Warning]** Restoring the interrupt bit and rethrowing is the right call, but the exception type now leaves the `SandboxException` vocabulary. `runShellCapture` is reached from `E2bSandbox.doExec`, and `SandboxBackedFilesystem.execute()` maps only `ExecTimeoutException` / `ExecException`; a raw `InterruptedIOException` propagates past both handlers.

Two knock-on effects worth deciding on explicitly:

1. A cancelled exec produces an unstructured failure for the agent instead of an `ExecuteResponse` that says the command was interrupted.
2. The check is a heuristic — any stale interrupt bit on a pooled thread (e.g. after `shutdownNow()`, or an interceptor that interrupts and does not clear it) turns a genuine `callTimeout` into "cancellation". The `finally` in the new test hints at how easy that is to leak. If the goal is to distinguish the two reliably, an explicit `call.isExecuted()`/`canceled` flag or a wrapper exception (`SandboxException.ExecInterruptedException`) is more robust than reading global thread state.

The second `catch` clause of `DockerSandbox.doExec` has the same shape, so a shared convention would keep the backends aligned.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
```diff
@@ -179,7 +197,7 @@ private int drainStartStream(
                 break;
             }
             int len = ByteBuffer.wrap(lenB).order(ByteOrder.BIG_ENDIAN).getInt() & 0x7FFFFFFF;
-            if (len < 0 || len > 64 * 1024 * 1024) {
+            if (len > 64 * 1024 * 1024) {
                 throw new IOException("Invalid connect frame length: " + len);
```
oss-maintainer, line 201: **[Info]** Dropping `len < 0` is correct — the `& 0x7FFFFFFF` on line 199 already makes the value non-negative, so the branch was dead. But that same mask is why a corrupt header with the high bit set is silently truncated rather than rejected: `0x80000001` becomes `1`, i.e. a plausible one-byte frame instead of an error.

If you want the length guard to be meaningful, validate the unsigned value before masking:
```java
int raw = ByteBuffer.wrap(lenB).order(ByteOrder.BIG_ENDIAN).getInt();
long len = Integer.toUnsignedLong(raw);
if (len > 64L * 1024 * 1024) {
    throw new IOException("Invalid connect frame length: " + len);
}
```
Pre-existing behaviour, not introduced here — mentioning it only because this hunk touches the validation.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
```diff
@@ -119,10 +119,19 @@ byte[] runShellBinaryStdout(
...
-                timeoutSeconds > 0
-                        ? http.newBuilder().callTimeout(timeoutSeconds, TimeUnit.SECONDS).build()
-                        : http;
+                http.newBuilder()
+                        .callTimeout(timeoutSeconds, TimeUnit.SECONDS)
+                        // Disable the idle read timeout: it fires on gaps between bytes and
+                        // would preempt callTimeout (both surface as InterruptedIOException),
+                        // misreporting a short idle stall as a full exec timeout (#2974).
+                        // Total-duration semantics is owned solely by callTimeout.
+                        .readTimeout(0, TimeUnit.SECONDS)
```
oss-maintainer, line 133: **[Info]** `readTimeout(0)` is the actual fix for #2974, but nothing pins it: the new tests use application interceptors, which throw before OkHttp's timeouts engage, so the assertions really cover "header is set" and "interrupt is not swallowed". If someone restores the inherited `readTimeout`, the suite still passes.

A `MockWebServer` case that writes one frame, stalls longer than `opt.getReadTimeoutSeconds()`, and asserts `ExecTimeoutException` would lock the behaviour in; asserting `callClient.readTimeoutMillis() == 0 && callClient.callTimeoutMillis() == timeoutSeconds * 1000` via a small package-private seam is the cheap version.

chcodex:
@oss-maintainer Thanks for the thorough review — all five points addressed in ec8cf440 (19/19 tests green, spotless clean). Point by point:

1. **Stale interrupt bit — fixed.** Snapshot/clear with `Thread.interrupted()` before the call, restore in `finally`; the catch only treats a bit set *during* the call as cancellation.

2. **`<=0` fail-fast — kept deliberately, diverges from daytona/agentrun.** Clamping to 1s would silently convert a caller bug (0/-5 is never intentional) into a 1-second execution — masking real bugs in production. fail-fast stays, but the message is now actionable and survives the generic-catch surfacing path (`"Internal sandbox error: ..."`): `[e2b] timeoutSeconds must be positive, got 0 for command "...". This is a caller bug, not a sandbox failure; check the timeout argument.` Normal agent paths always pass positive values (default 120), so this only fires on direct API misuse.

3. **Identical deadlines — kept, with evidence, plus deterministic attribution.** Live-verified against a real sandbox: `sleep 30` + `timeout=5` → `ExecTimeoutException` at ~5s, and `ps` confirms the server killed the process (no leak, no retry overlap). Rationale for identical values: the client timer starts at call initiation while the server timer starts at handler-begin (~1 RTT later), so the client wins by construction; no slack constant needed in either direction. For the residual case (server wins — client-timer delivery jitter past the ~RTT margin), the client now reads the kill reason instead of guessing:
   - end-stream error frames are parsed (JSON + PROTO via minimal `CodedInputStream` read, no new descriptors); `deadline_exceeded` → `ExecTimeoutException`, other codes → stream error carrying the code;
   - no error frame (#2828 truncation path) → new `MissingExitCodeException extends IOException` (upstream `assertThrows(IOException)` still passes) mapped by elapsed time: `elapsed >= timeout` → `ExecTimeoutException`, earlier → stream error as before.
   Partial stdout/stderr attached to the timeout exception needs a harness exception change — left as follow-up.

4. **`requireNonNull` — accepted**, one line documenting the non-null body contract instead of the ternary.

5. **Socket-level stall test — agreed but blocked.** `envdHost` is hardcoded https with no port/host injection, so MockWebServer cannot be pointed at it; app-level interceptors short-circuit before OkHttp timeouts engage (the old sleep-based test proved nothing and was removed). Making the host injectable is a separate refactor — follow-up.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
```diff
@@ -143,25 +162,160 @@ private ShellCapture runShellCapture(
         ByteArrayOutputStream stdout = new ByteArrayOutputStream();
         ByteArrayOutputStream stderr = new ByteArrayOutputStream();
         int exit;
+        // Snapshot the interrupt bit: a stale bit set before this call must not turn a
+        // genuine exec timeout into a fake cancellation below.
+        boolean preInterrupted = Thread.interrupted();
```
oss-maintainer: Snapshot-and-restore fixes the misclassification from my previous point 1, but it also removes the only fast-abort a caller has: `Thread.interrupted()` clears the bit, so a thread that was **already** cancelled before entering `runShellCapture` now issues the HTTP call and blocks in `execute()` for up to `timeoutSeconds` (default is a full exec timeout) before the bit is re-set in `finally`. OkHttp's blocking `execute()` does not poll the interrupt bit, so restoring it afterwards is not an early return — it's a delayed one.

Suggest failing fast on a pre-existing bit instead of absorbing it:

```java
boolean preInterrupted = Thread.interrupted();
if (preInterrupted) {
    Thread.currentThread().interrupt();
    throw new InterruptedIOException("caller thread already interrupted; aborting exec before Start");
}
```

That keeps the guarantee you documented (a stale bit must not fake a cancellation of *this* call) while not making an already-cancelled caller wait out the timeout. Either way, a short test asserting the pre-existing-bit path's behaviour would be good — `interruptionIsRethrownNotWrappedAsTimeout` only covers an interrupt raised from inside the call.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
```diff
@@ -143,25 +162,160 @@ private ShellCapture runShellCapture(
...
+            // External cancellation surfaces here too; don't misreport it as a timeout
+            // and don't swallow the interrupt bit. Only a bit set during this call
+            // counts (pre-existing stale bits were cleared above).
+            if (Thread.currentThread().isInterrupted()) {
+                throw e;
+            }
             throw new SandboxException.ExecTimeoutException(shellCommand, timeoutSeconds);
+        } catch (ConnectStreamException e) {
+            if (CONNECT_DEADLINE_EXCEEDED.equals(e.code())) {
+                throw new SandboxException.ExecTimeoutException(shellCommand, timeoutSeconds);
```
oss-maintainer, line 210: The `deadline_exceeded` mapping is the right shape, but as long as client `callTimeout` (line 135) and server `Connect-Timeout-Ms` (line 156) are the *same* duration, this branch is mostly unreachable: the client aborts the stream at `T` and the server writes the end-stream frame at `T` too, so in practice the connection is closed before the frame is read and `InterruptedIOException` (or the new `MissingExitCodeException` heuristic) wins the race. Two consequences:

1. The error code you are mapping — and the `mapPrematureEnd` fallback below — will fire rarely in production and mostly in tests (the canned-body tests bypass the race entirely, since the frame is already in memory).
2. The partial `stdout`/`stderr` already drained at that point is still discarded: `ExecTimeoutException` carries only the command and the timeout, so the operator sees no output for a long-running command that produced megabytes before being killed. That was the substance of my previous point 3 and it is not yet addressed.

Cheap improvement: give the server a slightly shorter deadline (e.g. `timeoutSeconds * 1000L - 500`) or extend `callTimeout` by a small grace (`timeoutSeconds + 1s`) so the terminal frame can arrive, and consider a `ExecTimeoutException(cmd, timeout, stdout, stderr)` variant (or attaching the captured output to the existing type) so the killed run's partial output survives to the harness error path.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
```diff
@@ -119,10 +121,24 @@ byte[] runShellBinaryStdout(
     private ShellCapture runShellCapture(
             E2bSandboxState state, String cwd, String shellCommand, int timeoutSeconds)
             throws Exception {
+        if (timeoutSeconds <= 0) {
+            throw new IllegalArgumentException(
+                    "[e2b] timeoutSeconds must be positive, got "
+                            + timeoutSeconds
+                            + " for command \""
+                            + truncateCommand(shellCommand)
```
oss-maintainer: Fair resolution of the previous point 2 — the `[e2b]` prefix plus "this is a caller bug, not a sandbox failure" makes the intent unmistakable, and pinning "no HTTP request must be issued" in the test is the part that actually matters.

One thing to note for whoever reads this later: the `IllegalArgumentException` type is still not mapped by `SandboxBackedFilesystem.execute()`, and daytona/agentrun clamp instead of throwing, so the cross-backend divergence you and @CryoThrust discussed is now documented in the message rather than removed. If that is the deliberate call, a one-line comment at the throw site saying why clamping was rejected would save the next reviewer re-litigating it.

Also mind that the message now embeds up to 200 chars of the raw command. Shell commands in this path are model-generated and routinely contain inline tokens (`curl -H "Authorization: Bearer ..."`), and an `IllegalArgumentException` message is exactly what lands in logs and exception metrics. If the goal is just operator context, consider logging the command at `warn` behind the existing redaction helper (or quoting only the first argv word) instead of putting it in the throwable.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-e2b/src/main/java/io/agentscope/extensions/sandbox/e2b/E2bEnvdProcessClient.java
```diff
@@ -143,25 +162,160 @@ private ShellCapture runShellCapture(
...
+                String message = "";
+                int inner;
+                while ((inner = in.readTag()) != 0) {
+                    int field = WireFormat.getTagFieldNumber(inner);
+                    if (WireFormat.getTagWireType(inner) != WireFormat.WIRETYPE_LENGTH_DELIMITED) {
+                        in.skipField(inner);
+                        continue;
+                    }
+                    if (field == 1) {
+                        code = in.readString();
```
oss-maintainer, line 308: Two small robustness notes on the hand-rolled wire decode:

1. `in.skipField(tag)` on an *outer* tag of unknown wire type, combined with the `WIRETYPE_LENGTH_DELIMITED` guard, is safe for `EndStreamResponse{ Error error = 1 }`, but if the field is ever reused with a different type the loop silently `skipField`s and returns `null`, which turns a real server error back into `MissingExitCodeException`. A `logger.debug("unparsable connect end-stream error ({} bytes)", data.length)` before `return null` would make a server-side format change diagnosable.
2. `readRawVarint32()` as the sub-message limit followed by `pushLimit` is correct, but a hostile/truncated `len` yields `NegativeRuntimeLengthException` — an `InvalidProtocolBufferException` that is *not* an `IOException` subclass here... actually it is, but it escapes through `parseProtoEndStreamError`'s `catch (Exception e) { return null; }` only because the catch is broad. It is fine today; please keep the broad catch, since a future narrowing to `InvalidProtocolBufferException` would let a malformed frame abort the drain loop.

oss-maintainer (commented):
### Summary
Re-review after the force-push that addresses my previous five points. All five landed: the interrupt bit is snapshotted and restored instead of being read ambiently, the fail-fast path now explains itself and is pinned by a "no HTTP request issued" assertion, `Objects.requireNonNull` settles the `res.body()` debate, and the new `Connect-Timeout-Ms` → `deadline_exceeded` mapping plus `mapPrematureEnd` close the "server killed it but we call it a stream error" gap with real coverage on both codecs (JSON and hand-decoded proto). CI green, CLA signed, mergeable; the two points below are not blockers for me, they are the race and the pre-existing-bit case that the new logic still leaves open.

- **[Warning]** `E2bEnvdProcessClient.java:167` — clearing a pre-existing interrupt bit means an already-cancelled caller now blocks in `execute()` for the full `timeoutSeconds` before the bit is restored; suggest failing fast when `preInterrupted` is true, plus a test for that path.
- **[Warning]** `E2bEnvdProcessClient.java:190` — client `callTimeout` and the server deadline are the same duration, so the `deadline_exceeded` branch will rarely win the race in production (the canned-body tests bypass it), and partial `stdout`/`stderr` drained before the kill is still discarded by `ExecTimeoutException`. Offsetting one side by ~500 ms and carrying the captured output would make the new mapping actually observable.
- **[Info]** `:129` — good resolution of the `IllegalArgumentException` debate, but the cross-backend divergence (daytona/agentrun clamp) is now documented rather than removed, and the message embeds raw model-generated command text that may carry inline tokens; worth a redaction pass or a why-not-clamp comment.
- **[Info]** `:281` — the hand-rolled proto end-stream decode is correct for `EndStreamResponse{error=1}`; keep the broad catch and consider a debug log when an unparsable frame falls through to `MissingExitCodeException`.

---

chcodex:
@oss-maintainer Re-review addressed in d9951910 (23/23 tests green, spotless clean). Point by point:

1. **Pre-cancelled caller no longer blocks — fail fast.** `Thread.interrupted()` snapshot is now followed by an immediate throw (`InterruptedIOException("exec cancelled before start")`, bit restored) when set, so an already-cancelled caller never enters `execute()`. Pinned by `alreadyCancelledCallFailsFastWithoutBlocking`: asserts immediate return, restored bit, non-timeout message, and no HTTP request issued. The now-dead `finally` restore was removed.

2. **Race made deterministic: server wins by 500ms.** `Connect-Timeout-Ms` stays exactly `timeout*1000` (kill lands on time, timeout promise stays accurate); client `callTimeout` is `timeout*1000+500ms` as pure backup. Since the server timer starts ~1 RTT after the client timer, the server now deterministically kills first and the client observes it via `deadline_exceeded` — so a retry can never overlap a still-running process (the exact failure mode from the earlier discussion). The client backup only fires on server misbehavior. Partial stdout/stderr is no longer silently dropped: both timeout-mapping paths emit a debug log with elapsed time, sizes, and 512-char tails. Carrying output *in* `ExecTimeoutException` needs a harness exception change — left as follow-up.

3. **Command text removed from the fail-fast message.** IAE now carries only the value + caller-bug guidance, plus a why-not-clamp comment. (`ExecTimeoutException` itself still embeds the full command — pre-existing behavior, out of scope here.)

4. **Debug log on unparsable proto end-stream frames — added.**

Coverage follow-ups included: malformed JSON end frame, proto frame with unknown top-level fields (skip path), empty proto end frame, and both `mapPrematureEnd` branches were already pinned. The socket-level stall test remains blocked on hardcoded `envdHost` (separate refactor).

oss-maintainer (commented):
### Summary
Delta re-review of `d9951910` (my previous review was `COMMENTED` on `ec8cf440`). Both non-blocking warnings from that review are now closed, and the fixes match what was asked:

- **Pre-cancel fail fast (`:167`)** — the interrupt-bit snapshot moved ahead of the call; it re-arms the bit and throws `InterruptedIOException("exec cancelled before start")` instead of letting an already-cancelled caller block for the full `timeoutSeconds`. `alreadyCancelledCallFailsFastWithoutBlocking` pins both the fast return and that no HTTP request is issued.
- **Server-wins race (`:190`)** — `callTimeout` is now `timeoutSeconds*1000 + 500ms` while `Connect-Timeout-Ms` keeps the exact server deadline, so `deadline_exceeded` deterministically wins and a retry never overlaps a still-running process. `timeoutSeconds * 1000L` promotes before the add, so no int overflow at realistic values.
- **`truncateCommand` removal** — command text is gone from the invalid-timeout error, with the rationale (model-generated commands can carry inline tokens) documented at the call site; grep over head confirms no dangling references.
- **Debug-only partial-stream logging** — `logPartialStreams` is guarded by `isDebugEnabled()` and capped at a 512-byte tail, so the hot path allocates nothing when debug is off, and the added `parseProtoEndStreamError` debug line makes the silent fall-through observable.

### Findings
- **[Info]** `E2bEnvdProcessClient.java:204` — only the `deadline_exceeded` and `MissingExitCodeException` paths log partial stdout/stderr tails; the generic `ConnectStreamException` path still surfaces code + message only. Reasonable scope choice, noting it just so the asymmetry is deliberate.
- **[Info]** CI: `build (ubuntu-latest)` and `build (windows-latest)` are still queued on `d9951910` as of this review, so I am not approving on this commit yet. `license/cla`, `Check License`, `Check Module Sync` are green and the branch is `MERGEABLE/CLEAN`.

The remaining check is just CI going green; nothing here needs another code change from my side.

---

PR #2978, opened 2026-09-04
fix(aistio): create agent_shares table and expose tierForCurrentUser (#2977)

### LINKED ISSUE #2977: [Bug]:fix(aistio): agent_shares table is never created — Agent Share API fails with "relation agent_shares does not exist", and console Agent detail tabs are hidden because tierForCurrentUser is never  returned
Describe the bug

  In agentscope-service (the aistiod Go control plane), the Agent sharing feature reads and writes a table named agent_shares, but the startup schema migration (aistio/internal/product/migrate.go) never creates
  that table — it only creates a legacy, unused resource_shares table. As a result, every Share operation fails with a Postgres error.

  There is a closely related second defect: the console frontend (frontend/src/components/AgentLayout.tsx) gates every Agent detail tab (Workspace / Skills / Tools / Subagents / Channels / Settings, all
  minTier: 'RUN') on a field called tierForCurrentUser, but the Go agent API never returns that field. With tierForCurrentUser absent, the tab strip is not rendered at all — even for the agent owner — so
  Tools/Skills/Subagents/etc. are unreachable from the UI.

  To Reproduce

  1. You code / environment: check out main and bring up the AgentScope Service stack (the control plane is the Go aistiod binary; schema is created by aistio/internal/product/migrate.go on startup).
  2. How to execute:
    either via docker compose
  docker compose -f agentscope-service/docker-compose.yml up --build -d
    or via the local dev stack
  cd agentscope-service && scripts/dev-down.sh && BUILDER_REBUILD=1 scripts/dev-up.sh
  2. Then open the console at http://localhost:8080, log in (admin/admin), and create an Agent.
  3. See error:
    - (a) Open the Agent's detail page — the whole tab strip (Workspace / Skills / Tools / Subagents / Channels / Settings) is missing.
    - (b) Click the ↗ Share button (or call GET /api/agents/{id}/shares / POST /api/agents/{id}/shares) — the request fails with a 500.

  Expected behavior

  - Opening an Agent's detail page should show all detail tabs for the owner (owner implicitly holds EDIT), and show read-only/clone-only views for users an agent was shared with.
  - Listing/adding/revoking shares should succeed and persist grants, supporting both per-user grants (grantee_type=USER) and workspace-wide grants (grantee_type=WORKSPACE, grantee_id='*') with the
  CLONE/RUN/EDIT tiers.

  Error messages

  GET /api/agents/{id}/shares  →  500
  {"error":"ERROR: relation \"agent_shares\" does not exist (SQLSTATE 42P01)"}
  (The same SQLSTATE 42P01 is raised by POST .../shares, by the admin user-deletion cascade DELETE FROM agent_shares ..., and by any agent-list visibility query that joins agent_shares.)

  For the hidden-tabs defect there is no server error; the frontend simply receives an agent object without tierForCurrentUser, so tierImplies(null, 'RUN') is false and no tab is rendered.

  - AgentScope-Java Version: 2.0.3-SNAPSHOT — agentscope-service, aistiod control plane (Go module github.com/spring-ai-alibaba/aistio)
  - Java Version: 17 (Temurin 17.0.19)
  - OS: macOS (Darwin 25.6.0); also reproduced in the Linux docker-compose stack
  - Database: PostgreSQL 17, schema cp

  Additional context

  Root cause:

  - handlers_agent_extras.go (listShares/addShare/revokeShare) and handlers_admin.go (user-deletion cascade) all operate on agent_shares, but migrate.go has no CREATE TABLE IF NOT EXISTS agent_shares — only the
  structurally different, unused resource_shares (which has grantee_user_id/permission and cannot represent grantee_type or the CLONE/RUN/EDIT tiers).
  - handlers_agents.go never populates tierForCurrentUser in the agent JSON, while AgentLayout.tsx requires it to render tabs.

  Suggested fix (already prepared on a branch based on main):

  1. Add the agent_shares migration (columns owner_id, agent_id, grantee_type, grantee_id, tier, created_at, with a unique constraint on (owner_id, agent_id, grantee_type, grantee_id) matching the addShare ON 
  CONFLICT clause, plus lookup indexes). The migration is idempotent (CREATE TABLE IF NOT EXISTS), so existing deployments pick it up on aistiod restart — no data volume reset required.
  2. Return tierForCurrentUser from the agent API: owners resolve to EDIT; other users get their highest CLONE/RUN/EDIT grant from agent_shares (a direct USER grant or a WORKSPACE '*' grant), and scope list/get
  to agents the caller owns or has a share for.
  3. Mark resource_shares as a legacy/unused table.

  ---

shynemo:
## Summary

Fixes #2977.

The Agent sharing feature in the `agentscope-service` aistiod control plane reads/writes a table named **`agent_shares`**, but the startup schema migration never created it — only the legacy, unused `resource_shares` table was. This causes Share operations (and any agent-list visibility query that joins `agent_shares`) to fail with `relation "agent_shares" does not exist (SQLSTATE 42P01)`.

A closely related defect: the console `AgentLayout` gates every Agent detail tab (Workspace/Skills/Tools/Subagents/Channels/Settings, all `minTier: 'RUN'`) on `tierForCurrentUser`, but the Go agent API never returned that field, so the whole tab strip was hidden — even for the agent owner.

## Changes

- **`aistio/internal/product/migrate.go`**
  - Add the `agent_shares` `CREATE TABLE` migration (`owner_id, agent_id, grantee_type, grantee_id, tier, created_at`), with a unique constraint on `(owner_id, agent_id, grantee_type, grantee_id)` matching the `addShare` `ON CONFLICT` clause, plus lookup indexes. The migration is idempotent (`CREATE TABLE IF NOT EXISTS`), so existing deployments pick it up on aistiod restart — no data-volume reset required.
  - Mark `resource_shares` as a legacy table with no readers/writers (doc comment only).
- **`aistio/internal/product/handlers_agents.go`**
  - Return `tierForCurrentUser`: owners resolve to `EDIT`; other users get their highest CLONE/RUN/EDIT grant from `agent_shares` (a direct `USER` grant or a `WORKSPACE '*'` grant).
  - Scope `listAgents`/`getAgent` to agents the caller owns or has a share for; load cross-owner shared agents via `loadSharedAgent` with a tier check.
- **`service-scheduler/.../worker/HandsWorkerMain.java`**
  - Fix the javadoc usage example: the self-hosted Hands worker authenticates with an environment key issued as `ek_<id>` (`product.shortID("ek_")`), not `ebk_`.

## Verification

- `go build ./... && go vet ./internal/product/ && go test ./internal/product/` — pass.
- The added table columns/constraint were checked against every SQL site that touches `agent_shares`: `listShares`/`addShare`/`revokeShare` (`handlers_agent_extras.go`), the admin user-deletion cascade (`handlers_admin.go`), and the visibility/tier queries (`handlers_agents.go`).
- After deploying, restarting aistiod runs the idempotent migration (`running cp schema migration` → `cp schema migration complete`); the Agents page then loads, the detail tabs render, and Share list/add/revoke succeed.

## Additional notes

- `resource_shares` is left in place (not dropped) for compatibility with existing databases; it has no code readers or writers and is now documented as legacy.

### DIFF
--- a/agentscope-service/aistio/internal/product/handlers_agents.go
+++ b/agentscope-service/aistio/internal/product/handlers_agents.go
@@ -17,6 +17,7 @@ package product
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -79,6 +80,10 @@ type agentRow struct {
 	ArchivedAt                 *int64
 	CreatedAt                  int64
 	UpdatedAt                  int64
+
+	// EffectiveTier 是当前请求用户对该 Agent 持有的控制台 ShareTier
+	// （CLONE/RUN/EDIT），不持久化；空串表示"无权限/未知"。
+	EffectiveTier string
 }
 
 func (a agentRow) toJSON() gin.H {
@@ -127,7 +132,38 @@ func (a agentRow) toJSON() gin.H {
 	if a.MultiagentJSON != nil && *a.MultiagentJSON != "" && *a.MultiagentJSON != "null" {
 		out["multiagent"] = parseJSONRaw(*a.MultiagentJSON)
 	}
+	// tierForCurrentUser 控制控制台 Agent 详情标签页的渲染：EDIT 全部可编辑，
+	// RUN 标签页只读，CLONE 仅显示克隆卡片；无权限时不输出该字段。
+	if a.EffectiveTier != "" {
+		out["tierForCurrentUser"] = a.EffectiveTier
+	}
 	return out
+}
+
+// agentTierForUser 解析当前用户对某 Agent 持有的控制台 ShareTier：
+// owner 本人 → EDIT；否则取 agent_shares 中匹配的最高档授权（授权给该用户的
+// USER 授权，或共享给所有人的 WORKSPACE '*' 授权）。无任何授权时返回空串。
+// 档位排序需与控制台保持一致（CLONE < RUN < EDIT）。
+func (s *Server) agentTierForUser(ctx context.Context, ownerID, agentID, userID string) string {
+	if userID != "" && userID == ownerID {
+		return "EDIT"
+	}
+	if userID == "" {
+		return ""
+	}
+	var tier string
+	err := s.db.Pool.QueryRow(ctx,
+		`SELECT tier FROM agent_shares
+		 WHERE owner_id=$1 AND agent_id=$2
+		   AND ((grantee_type='WORKSPACE' AND grantee_id='*')
+		     OR (grantee_type='USER' AND grantee_id=$3))
+		 ORDER BY CASE tier WHEN 'EDIT' THEN 3 WHEN 'RUN' THEN 2 WHEN 'CLONE' THEN 1 ELSE 0 END DESC
+		 LIMIT 1`,
+		ownerID, agentID, userID).Scan(&tier)
+	if err != nil {
+		return ""
+	}
+	return tier
 }
 
 func deref(p *string) string {
@@ -157,6 +193,26 @@ const agentSelect = `SELECT owner_id, agent_id, workspace_path, workspace_id, na
 	default_environment_id, default_vault_ids_json, default_memory_store_ids_json,
 	head_version, archived_at, created_at, updated_at FROM agents`
 
+// agentVisibilityClause 把可见范围限定为：用户自己拥有的 Agent（$1），
+// 或通过 agent_shares 分享给该用户的 Agent（授权给该用户的 USER 授权，
+// 或共享给所有人的 WORKSPACE '*' 授权）。
+const agentVisibilityClause = `(a.owner_id=$1
+	OR EXISTS (SELECT 1 FROM agent_shares s
+		WHERE s.owner_id=a.owner_id AND s.agent_id=a.agent_id
+		  AND ((s.grantee_type='WORKSPACE' AND s.grantee_id='*')
+		    OR (s.grantee_type='USER' AND s.grantee_id=$1))))`
+
+// agentTierExpr 在 SQL 中按行计算当前用户的有效 ShareTier：owner 为 EDIT；
+// 否则取匹配授权中的最高档（CLONE < RUN < EDIT）。
+const agentTierExpr = `COALESCE(
+	CASE WHEN a.owner_id=$1 THEN 'EDIT' END,
+	(SELECT s.tier FROM agent_shares s
+		WHERE s.owner_id=a.owner_id AND s.agent_id=a.agent_id
+		  AND ((s.grantee_type='WORKSPACE' AND s.grantee_id='*')
+		    OR (s.grantee_type='USER' AND s.grantee_id=$1))
+		ORDER BY CASE s.tier WHEN 'EDIT' THEN 3 WHEN 'RUN' THEN 2 WHEN 'CLONE' THEN 1 ELSE 0 END DESC
+		LIMIT 1))`
+
 func (s *Server) listAgents(c *gin.Context) {
 	owner := currentUserID(c)
 	limit, offset, ok := pageParams(c)
@@ -166,12 +222,16 @@ func (s *Server) listAgents(c *gin.Context) {
 	}
 	var total int64
 	if err := s.db.Pool.QueryRow(c.Request.Context(),
-		`SELECT COUNT(*) FROM agents WHERE owner_id=$1 AND archived_at IS NULL`, owner).Scan(&total); err != nil {
+		`SELECT COUNT(*) FROM agents a WHERE a.archived_at IS NULL AND `+agentVisibilityClause,
+		owner).Scan(&total); err != nil {
 		writeErr(c, http.StatusInternalServerError, err.Error())
 		return
 	}
 	writeTotalCount(c, total)
-	q := agentSelect + ` WHERE owner_id=$1 AND archived_at IS NULL ORDER BY updated_at DESC`
+	q := `SELECT a.*, ` + agentTierExpr + ` AS effective_tier
+		FROM (` + agentSelect + `) a
+		WHERE a.archived_at IS NULL AND ` + agentVisibilityClause + `
+		ORDER BY a.updated_at DESC`
 	args := []any{owner}
 	q, args = appendPage(q, limit, offset, args)
 	rows, err := s.db.Pool.Query(c.Request.Context(), q, args...)
@@ -182,11 +242,18 @@ func (s *Server) listAgents(c *gin.Context) {
 	defer rows.Close()
 	list := []gin.H{}
 	for rows.Next() {
-		a, err := s.scanAgent(rows)
-		if err != nil {
+		var a agentRow
+		var tier *string
+		if err := rows.Scan(
+			&a.OwnerID, &a.AgentID, &a.WorkspacePath, &a.WorkspaceID, &a.Name, &a.Description,
+			&a.SysPrompt, &a.Model, &a.MaxIters, &a.ToolsJSON, &a.McpServersJSON,
+			&a.SkillsJSON, &a.MultiagentJSON,
+			&a.DefaultEnvironmentID, &a.DefaultVaultIDsJSON, &a.DefaultMemoryStoreIDsJSON,
+			&a.HeadVersion, &a.ArchivedAt, &a.CreatedAt, &a.UpdatedAt, &tier); err != nil {
 			writeErr(c, http.StatusInternalServerError, err.Error())
 			return
 		}
+		a.EffectiveTier = deref(tier)
 		list = append(list, a.toJSON())
 	}
 	c.JSON(http.StatusOK, list)
@@ -290,6 +357,7 @@ func (s *Server) createAgent(c *gin.Context) {
 		writeErr(c, http.StatusInternalServerError, err.Error())
 		return
 	}
+	a.EffectiveTier = "EDIT"
 	c.JSON(http.StatusOK, a.toJSON())
 }
 
@@ -313,11 +381,11 @@ func (s *Server) agentSnapshot(owner, id, name, desc, system, model string, maxI
 		"id": id, "name": name, "description": desc, "system": system, "model": model,
 		"maxIters": maxIters, "tools": tools, "mcpServers": mcp, "skills": skills,
 		"multiagent": multi, "scope": "user", "ownerId": owner, "workspacePath": ws,
-		"workspaceId":             nullStr(workspaceID),
-		"defaultEnvironmentId":    nullStr(defaultEnv),
-		"defaultVaultIds":         defaultVault,
-		"defaultMemoryStoreIds":   defaultMem,
-		"version":                 version, "createdAt": created, "updatedAt": updated,
+		"workspaceId":           nullStr(workspaceID),
+		"defaultEnvironmentId":  nullStr(defaultEnv),
+		"defaultVaultIds":       defaultVault,
+		"defaultMemoryStoreIds": defaultMem,
+		"version":               version, "createdAt": created, "updatedAt": updated,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


oss-maintainer:

```bash
git fetch origin
git checkout 2977-fix-agent-shares-tier
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

oss-maintainer (commented):
### Summary

The root cause is correctly diagnosed and the fix is in the right place: `agent_shares` was read and written by `handlers_agent_extras.go` but never created by the migration, and the console's tab strip gates on a `tierForCurrentUser` field the Go API never returned. Computing the tier inside the list query instead of per row is the right call, the unique constraint matches the `ON CONFLICT` target in `addShare` exactly, and collapsing the `loadAgent`-then-`scanAgent` duplication into a shared fragment keeps `agentSelect` as the single source of truth for column order.

What I am not comfortable signing off on is the identity model around the new grants: `listAgents` matches a grant against `currentResourceOwner`, `getAgent` matches it against `currentUserID`, and `loadSharedAgent` looks agents up by `agent_id` alone even though `agent_id` is only unique per owner. Those three disagreements decide who can see an agent, so they need an explicit answer rather than a follow-up. The change also lands with no tests in a package that has an established Go access-control test convention.

### Findings
- **[Warning]** `handlers_agents.go:482` — `loadSharedAgent` filters on `agent_id` alone; with two owners using the same id, the first row Postgres returns decides the outcome and a real grant can 404.
- **[Warning]** `handlers_agents.go:199` — `agentVisibilityClause` matches grants with the same `$1` it uses for ownership (`currentResourceOwner`), while `getAgent` uses `currentUserID`; the two endpoints can disagree about the same share.
- **[Warning]** `migrate.go:204` — no `CHECK` on `tier` / `grantee_type` although `addShare` inserts client input verbatim; `ORDER BY CASE … ELSE 0` ranks an unrecognised tier but still returns it as `tierForCurrentUser`, and a typo'd `grantee_type` yields a share row that grants nothing forever.
- **[Warning]** `handlers_agents.go:207` — `('WORKSPACE','*')` means "every authenticated user", not "members of this workspace"; a `('WORKSPACE','ws_…')` row stores fine and matches nothing.
- **[Info]** `handlers_agents.go:250` — hand-written 21-column `Scan` duplicates `scanAgent`'s contract, so a new column shifts it silently.
- **[Info]** `handlers_agents.go:163` — all query errors become `""`, so an outage reads as "not shared"; that is precisely the `relation does not exist` failure this PR fixes.

### Missing
- No test coverage for any of it. `resource_access_test.go` and `managed_resources_test.go` already establish the pattern for exactly this kind of check, and the cases worth pinning are: owner reads own agent → `EDIT`; a `RUN` grantee sees it in the list *and* can open it; the same grantee after `revokeShare` disappears from both; `('WORKSPACE','*')` reaches a second, unrelated user; two owners sharing the same `agent_id` each resolve only their own; and an unrecognised `tier` string.
- Nothing guards the migration itself. The fix is "the table was never created", so a test that boots the schema and exercises a share round-trip is the cheapest way to make sure the next table added to `handlers_agent_extras.go` cannot regress the same way.

### Questions
1. Is a grant keyed on the authenticated account (`currentUserID`) or on the namespace resource owner (`currentResourceOwner`)? Whichever it is, both `agentVisibilityClause` and `getAgent` should use it, and the comment should say so — the two identifiers are deliberately different in this package.
2. Should `agent_id` be treated as globally unique here? If yes, the `agents` unique key is the wrong constraint; if no, `loadSharedAgent` needs an owner predicate.

### Strengths
- Correct diagnosis of a latent schema omission, fixed idempotently: `CREATE TABLE IF NOT EXISTS` plus `CREATE INDEX IF NOT EXISTS` means an existing deployment heals on aistiod restart without a data-volume reset, and the comment marking `resource_shares` as legacy stops the next reader from repeating the mistake.
- Both indexes map onto the two real access patterns (by agent, by grantee) rather than being speculative.
- `tierForCurrentUser` is omitted rather than emitted as `""` when there is no grant, so the console's "no tier" case and its "unknown tier" case cannot diverge.
- Hard-coding `EffectiveTier = "EDIT"` in `createAgent`/`updateAgent`/`archiveAgent` is correct: those flows resolve the agent through `loadAgent(owner, …)`, so the caller is necessarily the owner.

---

agentscope-service/aistio/internal/product/handlers_agents.go
```diff
@@ -390,14 +460,38 @@ func (s *Server) loadAgent(ctx context.Context, owner, agentID string) (agentRow
...
 		return
 	}
 	c.JSON(http.StatusOK, a.toJSON())
 }
 
+// loadSharedAgent 加载他人所有、但已通过 agent_shares 分享给当前用户的
+// Agent。Agent 不存在或用户无任何授权时返回错误。
+func (s *Server) loadSharedAgent(ctx context.Context, user, agentID string) (agentRow, error) {
+	row := s.db.Pool.QueryRow(ctx,
+		agentSelect+` WHERE agent_id=$1 AND archived_at IS NULL`, agentID)
```
oss-maintainer, line 482: **[Warning]** `agent_id` is not globally unique — `agents` is only `UNIQUE (owner_id, agent_id)` (migrate.go:53), and `agent_shares` carries the same two-column key (migrate.go:206). This statement filters on `agent_id` alone, so when two owners have the same id, `QueryRow` returns whichever row Postgres happens to emit first and the tier check runs against *that* owner. The consequence is a false 404: the user is granted on owner B's `agent_x`, but owner A's row comes back first, `agentTierForUser` finds no grant for A, and the handler reports `agent not found` (line 472) even though the grant exists. Requiring the grant row to name the owner removes the ambiguity without any schema change:

```go
row := s.db.Pool.QueryRow(ctx,
    `SELECT `+agentSelectColumns+` FROM agents a
     WHERE a.agent_id=$1 AND a.archived_at IS NULL
       AND EXISTS (SELECT 1 FROM agent_shares s
                   WHERE s.owner_id=a.owner_id AND s.agent_id=a.agent_id
                     AND ((s.grantee_type='WORKSPACE' AND s.grantee_id='*')
                       OR (s.grantee_type='USER' AND s.grantee_id=$2)))
     LIMIT 1`, agentID, user)
```

so the row is only returned when it is actually shared with this user.

agentscope-service/aistio/internal/product/handlers_agents.go
```diff
@@ -157,6 +193,26 @@ const agentSelect = `SELECT owner_id, agent_id, workspace_path, workspace_id, na
 	default_environment_id, default_vault_ids_json, default_memory_store_ids_json,
 	head_version, archived_at, created_at, updated_at FROM agents`
 
+// agentVisibilityClause 把可见范围限定为：用户自己拥有的 Agent（$1），
+// 或通过 agent_shares 分享给该用户的 Agent（授权给该用户的 USER 授权，
+// 或共享给所有人的 WORKSPACE '*' 授权）。
+const agentVisibilityClause = `(a.owner_id=$1
```
oss-maintainer, line 199: **[Warning]** The two read paths match a grant against different principals. `listAgents` passes `owner := currentResourceOwner(c)` as `$1`, and this clause reuses that one parameter for both `a.owner_id=$1` and `s.grantee_id=$1`; `getAgent` (line 470) instead resolves grants through `currentUserID(c)`. `currentResourceOwner` returns `resourceOwner` when the control plane set one and only otherwise falls back to `currentUserID` (middleware.go:90-95), and `SetResourceOwner` is explicitly documented as *not* replacing the authenticated account identity (middleware.go:87-88). So whenever those differ, the endpoints disagree about the same share row — an agent can appear in the list with a `tierForCurrentUser` and still 404 when opened, or open fine and never appear in the list. This is the same class of bug as #2977 (one path knowing about a grant, the other not), so it seems worth pinning now: take the grantee from `currentUserID` in both statements and pass owner and grantee as separate parameters, e.g. `(a.owner_id=$1 OR EXISTS (… s.grantee_id=$4))`.

agentscope-service/aistio/internal/product/migrate.go
```diff
@@ -189,6 +191,23 @@ CREATE TABLE IF NOT EXISTS resource_shares (
...
+-- grantee_type='USER' 时 grantee_id 为被授权用户；'WORKSPACE' 时 grantee_id='*'
+-- 表示全员共享。handlers_agent_extras.go 的 shares 接口与 handlers_agents.go
+-- 的可见性/tier 解析均依赖此表。
+CREATE TABLE IF NOT EXISTS agent_shares (
+    row_id       BIGSERIAL PRIMARY KEY,
+    owner_id     TEXT NOT NULL,
+    agent_id     TEXT NOT NULL,
+    grantee_type TEXT NOT NULL,
+    grantee_id   TEXT NOT NULL DEFAULT '',
+    tier         TEXT NOT NULL DEFAULT 'RUN',
```
oss-maintainer, line 204: **[Warning]** This is the moment to constrain the two enum-ish columns: `addShare` (handlers_agent_extras.go:585-588) only requires `granteeType` and `tier` to be non-empty and inserts them verbatim, and the readers here assume a closed set. Because the tier ranking is `ORDER BY CASE … ELSE 0`, the `ELSE` only *demotes* an unrecognised value — it never filters it — so a row with `tier='ADMIN'` is still handed back as `tierForCurrentUser`, which is exactly what `AgentLayout` gates tabs on. A typo'd `grantee_type` is worse: it is writable, listable and revokable, but permanently invisible to visibility and tier resolution.

```sql
    grantee_type TEXT NOT NULL CHECK (grantee_type IN ('USER','WORKSPACE')),
    grantee_id   TEXT NOT NULL DEFAULT '',
    tier         TEXT NOT NULL DEFAULT 'RUN' CHECK (tier IN ('CLONE','RUN','EDIT')),
```

Since the table is brand new, a `CHECK` costs nothing today and is a painful, lock-taking migration later.

agentscope-service/aistio/internal/product/handlers_agents.go
```diff
@@ -157,6 +193,26 @@ const agentSelect = `SELECT owner_id, agent_id, workspace_path, workspace_id, na
...
+	OR EXISTS (SELECT 1 FROM agent_shares s
+		WHERE s.owner_id=a.owner_id AND s.agent_id=a.agent_id
+		  AND ((s.grantee_type='WORKSPACE' AND s.grantee_id='*')
+		    OR (s.grantee_type='USER' AND s.grantee_id=$1))))`
+
+// agentTierExpr 在 SQL 中按行计算当前用户的有效 ShareTier：owner 为 EDIT；
+// 否则取匹配授权中的最高档（CLONE < RUN < EDIT）。
+const agentTierExpr = `COALESCE(
```
oss-maintainer, line 207: **[Warning]** Worth making the `WORKSPACE` semantics explicit, because the clause as written means "published to every authenticated user in this deployment", not "members of this workspace": `grantee_id='*'` is the only value that matches, no workspace membership is consulted, and `a.workspace_id` is not referenced anywhere in this predicate. `addShare` will happily store `('WORKSPACE', 'ws_123')`, and such a row then grants nothing at all — the opposite failure. If the console's intent is workspace-scoped sharing, the predicate needs `s.grantee_id = a.workspace_id` plus a membership check; if `'*'` really means "public to all", that belongs in the comment above `agentVisibilityClause` and in the table comment in `migrate.go`, since the tier is the console's only edit/run gate.

agentscope-service/aistio/internal/product/handlers_agents.go
```diff
@@ -183,11 +245,18 @@ func (s *Server) listAgents(c *gin.Context) {
 	defer rows.Close()
 	list := []gin.H{}
 	for rows.Next() {
-		a, err := s.scanAgent(rows)
-		if err != nil {
+		var a agentRow
+		var tier *string
+		if err := rows.Scan(
```
oss-maintainer, line 250: **[Info]** `SELECT a.*` plus a hand-written 21-destination `Scan` duplicates the column contract that `scanAgent` (line 176-189) exists to own: `agentSelect` is the single source of truth today, and adding a column to it silently shifts this list instead of failing where it is declared. Since `scanAgent` already accepts any `interface{ Scan(dest ...any) error }`, a `scanAgentWithTier` helper (the 20 fields plus `&tier`) keeps one source of truth while preserving the single-query shape you chose to avoid an N+1.

agentscope-service/aistio/internal/product/handlers_agents.go
```diff
@@ -127,9 +132,40 @@ func (a agentRow) toJSON() gin.H {
...
+	var tier string
+	err := s.db.Pool.QueryRow(ctx,
+		`SELECT tier FROM agent_shares
+		 WHERE owner_id=$1 AND agent_id=$2
+		   AND ((grantee_type='WORKSPACE' AND grantee_id='*')
+		     OR (grantee_type='USER' AND grantee_id=$3))
+		 ORDER BY CASE tier WHEN 'EDIT' THEN 3 WHEN 'RUN' THEN 2 WHEN 'CLONE' THEN 1 ELSE 0 END DESC
+		 LIMIT 1`,
+		ownerID, agentID, userID).Scan(&tier)
+	if err != nil {
```
oss-maintainer, line 163: **[Info]** Swallowing every error as `""` is fail-closed, which is the right direction for an authorization lookup, but it makes an outage indistinguishable from "no grant" and surfaces it as a 404 at line 472. It also hides the exact failure mode this PR fixes: before the migration landed, every statement here would have errored on `relation "agent_shares" does not exist` and looked like "nobody is shared with you". `errors.Is(err, pgx.ErrNoRows)` for the empty case, and logging the rest, would keep a transient failure from reading like lost permissions. Same for the `SELECT` in `loadSharedAgent` (line 483), where a real error is returned as "not found" to the client.

PR #2980, opened 2026-09-04
fix(sandbox-kubernetes): support claim lifecycle retention

### LINKED ISSUE #2979: [Bug]:2.0.2版本 k8s sandbox  每次调用后都会删除claim，下一次调用resume不到卡30秒后会重新创建
**Describe the bug**
claimOwned被设置为true，会走到删除claim的逻辑。 下次call时根据AgentState保存的claimName去resume时还要卡30秒，然后走create逻辑。

**To Reproduce**
采用k8s sandbox，现象必现。

**Expected behavior**
希望能在下个版本中进行修复。

我尝试解的决方式：
起初以为可以用agent-sandbox的ttlSecondsAfterFinished机制，但由于是运行时服务，没有Finished条件，所以claim释放不了。
对sdkSandbox增加个renewShutdownTime方法，而不是调用sdkSandbox的terminate方法。

guslegend0510:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #2979.

Kubernetes sandbox claims were always treated as client-owned. Releasing a sandbox therefore deleted its `SandboxClaim`, while the persisted agent state still referenced that claim. The next call attempted to resume the deleted claim and waited for the watch timeout before creating a replacement.

This PR:

- Exposes `claimOwned(boolean)` for explicitly controlling deletion on release.
- Adds `shutdownAfterSeconds(long)` as a hard controller-managed expiry.
- Adds `ttlSecondsAfterFinished(int)` and serializes it to `SandboxClaim.spec.lifecycle`.
- Keeps the compatibility default for claims without a hard shutdown deadline.
- Avoids retaining TTL-only claims implicitly because long-running runtimes may never reach `Finished`.
- Fails immediately when a persisted claim no longer exists instead of waiting for the 30-second watch timeout.
- Preserves the existing public `createSandboxClaim` overload for source and binary compatibility.
- Adds unit tests and English/Chinese documentation.

### Lifecycle behavior

- Default: the client deletes the claim on release.
- With `shutdownAfterSeconds`: release only disconnects; the controller deletes the claim at the configured deadline.
- With `ttlSecondsAfterFinished` only: client ownership remains enabled by default.
- `claimOwned(false)` can explicitly retain a claim when the runtime is guaranteed to reach `Finished`.

### Verification

- Kubernetes module test suite: 44 tests passed.
- Final focused lifecycle tests: 31 tests passed, 0 failures.
- Spotless check passed.
- `git diff --check` passed.

## Checklist

- [x] All affected-module tests are passing
- [x] Javadoc comments follow project conventions


CryoThrust (commented):
I traced the lifecycle paths against the reported resume failure. The ownership resolution and the preserved seven-argument `createSandboxClaim` overload look consistent with the compatibility goal.

Two points that may be worth making explicit before merge:

1. For an already-persisted `KubernetesSandboxState`, `claimOwned` remains whatever was serialized previously unless the caller supplies an explicit `claimOwned` option. That is a reasonable migration choice, but it means an existing state created with the old default still deletes its claim on release. Could the documentation or a regression test state this migration behavior explicitly?
2. `ttlSecondsAfterFinished` is serialized into `spec.lifecycle` alongside `shutdownTime` and `shutdownPolicy`. Please confirm that the deployed agent-sandbox CRD/controller version accepts this field and ignores it safely when unsupported; a small serialization assertion or compatibility note would make the rollout contract clearer.

The immediate missing-claim failure is a useful improvement for this issue because it removes the 30-second watch wait. I did not find a competing change in the open PR list beyond this PR, so I am leaving this as review feedback rather than opening a duplicate implementation.

oss-maintainer:

```bash
git fetch origin
git checkout codex/fix-2979-kubernetes-claim-lifecycle
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/main/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxClient.java
```diff
@@ -150,6 +152,7 @@ public Sandbox resume(SandboxState state) {
                     "Expected KubernetesSandboxState but got: " + state.getClass().getName());
         }
         KubernetesSandboxClientOptions merged = merge(null);
+        applyConfiguredClaimOwnership(k8s, merged);
```
oss-maintainer, line 155: `resume()` applies `merge(null)` — i.e. only the client's *default* options — so an explicit `claimOwned` on the default options silently overwrites the ownership that was decided when this particular claim was created. Two claims created under different per-call options and resumed by the same client end up with the same release policy, and the persisted `KubernetesSandboxState.claimOwned` (the record of who created it) is ignored. Suggest the persisted state win on resume and only fall back to configured defaults when the state carries no decision (e.g. make the persisted field nullable, or gate on "state was written before this field existed"), so resume never changes a claim's ownership semantics under the caller's feet.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/main/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxClient.java
```diff
@@ -163,6 +166,26 @@ public Sandbox resume(SandboxState state) {
...
+     */
+    static boolean resolveClaimOwned(KubernetesSandboxClientOptions options) {
+        if (options.getClaimOwned() != null) {
+            return options.getClaimOwned();
+        }
+        return options.getShutdownAfterSeconds() == null;
+    }
+
+    /** Applies an explicit release policy to persisted state. */
+    static void applyConfiguredClaimOwnership(
```
oss-maintainer, line 182: The `shutdownAfterSeconds != null => claimOwned = false` inference is a behaviour change for existing users: today configuring a hard shutdown deadline still leaves the client deleting the claim on release, and after this PR release only disconnects. If the deadline elapses without the runtime reaching `Finished`, the claim (and its pod) now outlives the session unless the controller enforces `shutdownTime`. That is defensible, but it belongs in the release notes and ideally behind an explicit option; at minimum please call it out in the PR description so nobody reads it as a no-op config addition. Also worth documenting that `resolveClaimOwned` is consulted only at `create()` — changing the option later has no effect on already-created claims.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/main/java/io/agentscope/extensions/sandbox/kubernetes/client/internal/K8sHelper.java
```diff
@@ -201,21 +264,19 @@ public SandboxResource getSandbox(String name, String namespace) {
...
-            if (resolved != null) {
-                log.debug(
-                        "[sandbox-client] Sandbox name already resolved: {} -> {}",
-                        claimName,
-                        resolved);
-                return resolved;
-            }
+        SandboxClaim current = getSandboxClaim(claimName, namespace);
+        if (current == null) {
+            throw new SandboxNotFoundException(
```
oss-maintainer, line 269: This changes the fast path from "claim missing → fall through to the watch" to "claim missing → hard failure". `SandboxClient.getOrCreate` calls `resolveSandboxName` immediately after a create, and with an informer-cached or eventually-consistent client a just-created `SandboxClaim` can be briefly invisible; that window now surfaces as `SandboxNotFoundException` and the caller tears down the entry from `activeSandboxes`. The previous behaviour tolerated it by waiting inside the watch. If failing fast is intended, keep it but only after one watch cycle, or distinguish "claim never existed" from "claim not yet visible" in the message.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/main/java/io/agentscope/extensions/sandbox/kubernetes/client/internal/K8sHelper.java
```diff
@@ -96,14 +157,22 @@ public SandboxClaim createSandboxClaim(
...
-            String shutdownTime =
+            shutdownTime =
                     DateTimeFormatter.ISO_INSTANT.format(shutdownAt.atOffset(ZoneOffset.UTC));
-            spec.setLifecycle(new SandboxClaimSpec.Lifecycle(shutdownTime, "Delete"));
+        }
+        if (ttlSecondsAfterFinished != null && ttlSecondsAfterFinished < 0) {
+            throw new IllegalArgumentException("ttlSecondsAfterFinished must not be negative");
+        }
+        if (shutdownTime != null || ttlSecondsAfterFinished != null) {
+            spec.setLifecycle(
```
oss-maintainer, line 173: Extracting `buildSandboxClaim` so the CR can be asserted without a live client is a nice testability win. Two small notes: `shutdownAfterSeconds` is now validated in three places (options setter, `CreateSandboxOptions` ctor, here) — keeping one owner of the rule would avoid drift; and `shutdownTime` is computed from `Instant.now()` at build time, so a retry of a failed create moves the deadline. Fine today because create is not retried, but worth a comment.

oss-maintainer (commented):
### Summary
Adds claim lifecycle retention (`claimOwned`, `shutdownAfterSeconds`, `ttlSecondsAfterFinished`) through options → state → SDK, wires the CRD `spec.lifecycle.ttlSecondsAfterFinished`, and extracts `buildSandboxClaim` for real unit coverage. The direction is right: releasing a sandbox should not always mean destroying the cluster-side claim, and the finished-only vs hard-deadline distinction is called out explicitly.

### Findings
- **[Warning]** `resume()` lets the client default options override persisted per-claim ownership.
- **[Warning]** `shutdownAfterSeconds != null` now implies \"client does not delete\" — a behaviour change for existing configurations that needs a release note.
- **[Warning]** `resolveSandboxName` turns \"claim not yet visible\" into a hard `SandboxNotFoundException`; the old code waited inside the watch.
- **[Info]** validation triplicated; `shutdownTime` is computed at build time.

### Suggestions
`KubernetesSandboxClientTest` covering `resolveClaimOwned` combinations is exactly the right table. Please add one case for \"resume after create with different default options\" so the ownership-precedence decision is pinned down either way.

### Verdict
Comment only. The precedence question and the resume/visibility behaviour changes need a decision from a maintainer, and the PR currently has a merge conflict with `main`.

---

Issue #2983, opened 2026-09-04
[Feature]: Tool-level circuit breaker — stop offering a repeatedly failing tool instead of letting the ReAct loop keep retrying it


pure11011:
## Motivation

A ReAct loop is reason → call tool → read result → reason again. When a tool returns an error the model commonly calls it again, because from the model's point of view one failure looks incidental. Nothing in the framework remembers that a tool is broken.

Against a dependency that is genuinely down — provider outage, revoked key, network partition, rate-limit wall — that costs one model call, one outbound request and seconds of latency *per retry*, and still fails. Ten consecutive failures are ten wasted reasoning rounds. There is no mechanism that says "this tool is broken right now, stop advertising it".

What exists today covers a different problem:

- `ExecutionConfig` timeout (`ToolExecutor.applyTimeout`, default 5 min) bounds one call.
- `ExecutionConfig` retry (`ToolExecutor.applyRetry`) recovers a transient blip within one call, and defaults to `maxAttempts=1` for tools (`ExecutionConfig.TOOL_DEFAULTS`). It is also currently bypassed for tool exceptions — see #2829.
- `maxIters` caps loop length, but counts iterations, not failures, so a broken tool is still called every round until the cap.
- `Toolkit.setActiveGroups` / `ToolGroup.addTool` can disable a tool, but only when some external logic decides to.

None of these carry state across calls, so none can express "this tool has failed N times in a row; leave it alone for a while". Model-side failover has an open proposal along these lines (#2863, per-candidate cooldown), but there is no equivalent for tools.

## Proposal

An opt-in, per-tool circuit breaker with the usual three states and exponential backoff:

```
CLOSED --failureThreshold consecutive failures--> OPEN
OPEN --cooldown elapsed--> HALF_OPEN     (tool advertised again, as a probe)
HALF_OPEN --probe succeeds--> CLOSED
HALF_OPEN --probe fails--> OPEN          (next generation, longer cooldown)
```

Two decisions are worth stating up front, because they are what make this fit an agent rather than a service mesh:

**1. Withhold the tool, do not reject the call.** A classic breaker sits between caller and dependency and fails fast once open. An agent has a strictly better option: drop the tool from the schema list sent to the model. A tool the model cannot see is a tool it cannot ask for, so the retry loop disappears at the source instead of being absorbed. No prompt has to tell the model to avoid the tool, and the model cannot argue with the decision.

[...]

PR #2984, opened 2026-09-04
feat(tool): add tool-level circuit breaker to stop offering repeatedly failing tools

pure11011:
## AgentScope-Java Version

2.0.3-SNAPSHOT (branch rebased onto `main` at `7eec9e8`)

## Description

Implements #2983.

**Problem.** A ReAct loop is reason → call tool → read result → reason again. When a tool returns an error the model usually calls it again, because one failure looks incidental from where it stands, and nothing in the framework remembers that a tool is broken. Against a dependency that is genuinely down each retry costs a model call, an outbound request and seconds of latency, and still fails; ten consecutive failures are ten wasted reasoning rounds. The existing controls address a different problem — `applyTimeout` bounds a single call, `applyRetry` recovers a transient blip within a single call (and defaults to `maxAttempts=1` for tools), and `maxIters` counts iterations rather than failures — so none of them can express "this tool has failed N times in a row, leave it alone for a while".

**Change.** An opt-in, per-tool breaker with `CLOSED → OPEN → HALF_OPEN` transitions and exponential backoff, added as new classes only. Nothing existing changes behaviour.

New in `io.agentscope.core.tool.circuitbreaker`:

- `ToolCircuitBreaker` — the state machine and backoff policy. A plain object with no reactive or agent dependencies, so it is unit-testable against an injected `Clock`.
- `ToolCircuitBreakerConfig` — supervised/excluded tool sets, threshold, cooldown bounds, multiplier and probe timeout, with validation in `build()`.
- `ToolCircuitBreakerStore` — persistence SPI (`snapshot` / `compareAndSet` / `reset`), plus `InMemoryToolCircuitBreakerStore` as the default.
- `ToolCircuitSnapshot` / `ToolCircuitState` — all persisted state for one tool as a single immutable value, and the derived lifecycle state.
- `ToolCircuitBreakerMiddleware` — the adapter onto `MiddlewareBase`.

New in `agentscope-extensions-redis`: `RedisToolCircuitBreakerStore`, so replicas share one view of a broken tool instead of each rediscovering the outage and sending N times the failing traffic.

Four design points reviewers will want to weigh:

**1. The breaker withholds the tool rather than rejecting the call.** A classic breaker sits between caller and dependency and fails fast once open. Here there is a better option: drop the tool from the schema list the model receives. A tool the model cannot see is a tool it cannot ask for, so the retry loop disappears at the source instead of being absorbed, and no prompt has to argue the model out of it.

`onReasoning` filters `ReasoningInput.tools()` into a per-turn copy; the `Toolkit` is never touched. Mutating `ToolGroup` membership is the other way to hide a tool, and I deliberately avoided it: group membership is shared mutable state, so a circuit tripped for one session would remove the tool from every concurrent session of that agent and would overwrite the registrations the application declared. Per-turn filtering also means recovery needs no repair step — once the breaker stops withholding, the unfiltered list is already correct — which removes a whole class of state-thrashing bugs.

**2. Every transition is one compare-and-set, so the documented semantics actually hold under concurrency.** A breaker is shared by every concurrent turn, session and (with a distributed store) replica that can reach the tool. All state for a tool therefore lives in one `ToolCircuitSnapshot` and each transition is published by replacing exactly the value the caller observed. A caller that loses the race re-reads and recomputes. Two consequences worth stating explicitly:

- The threshold really counts *consecutive* failures. Counting the final failure and opening the circuit happen in the **same** compare-and-set, so the counter is never persisted at the threshold and no caller can observe or act on that intermediate state. A success that clears the streak in between invalidates the failure's decision, which then recounts from the cleared state.
- Backoff cannot be inflated. If several failures race, the first commits the next generation and the rest re-read, see a freshly opened circuit and are ignored — so the generation advances at most once per cooldown window.

**3. HALF_OPEN admits exactly one recovery probe.** Recovery is a claim, not merely an elapsed timestamp. `tryAcquireProbe` compare-and-sets a token plus lease into the snapshot; only the winner advertises the tool and the other turns keep withholding it. Outcomes are conditional on the token, so a late result from a superseded probe cannot close or re-open a newer one. The lease bounds the damage when a probe never reports an outcome — a turn that never calls the advertised tool, a lost execution, a crashed replica — after which another caller may claim it. The claim rides on the per-call `RuntimeContext`, bound to the tool-call id the model chooses; a turn that never calls the tool, a refused call and a cancelled call all hand it back early rather than waiting out the lease, while a suspended call keeps it because its outcome is still pending.

**4. Failures are classified by typed state, and supervision is opt-in.** `onActing` reads the `ToolResultState` already carried on `ToolResultEndEvent`; only `ERROR` counts. `DENIED` is a permission refusal and `INTERRUPTED` a cancellation — neither is evidence about the dependency, and counting `DENIED` would let a user who declines a confirmation prompt trip the circuit. Only tools named via `monitorTools(...)` are watched unless `monitorAllTools(true)` is set, and `excludeTools(...)` overrides both; the default configuration is inert. Withholding a flaky weather API degrades an agent gracefully, but withholding the database or filesystem tool cripples it, and opt-out supervision would also cover framework-supplied per-call tools such as the structured-output tool — losing that one stops the agent completing at all.

Cooldown is `min(initialCooldown * multiplier^(generation-1), maxCooldown)`, derived from the stored generation rather than persisted, so changing the policy takes effect immediately and never needs migrating. Defaults of 60s / x2 / 600s give 60s, 120s, 240s, 480s, 600s… `cooldownFor` clamps non-finite and over-ceiling values, so a large generation cannot isolate a tool indefinitely. `probeTimeout` defaults to 5 minutes to match `ExecutionConfig.TOOL_DEFAULTS`; the javadoc tells callers with longer tool timeouts to raise it.

State is derived from the stored snapshot plus the current time, so a cooldown or lease that elapsed while the agent was idle is recognised on the next read: no scheduler, no timer, no background thread, and nothing to leak.

The Redis store keeps the whole snapshot in **one** key and compare-and-sets it in one Lua script, so every script touches a single key, the store needs no hash tag and works unchanged on Redis Cluster. A missing key is the encoding of the closed state, so a recovered tool leaves nothing behind.

## How to test

- `mvn -pl agentscope-core clean verify` — passes locally: **2371 tests, 0 failures, 0 errors, 9 pre-existing skips**. 52 are new: 28 in `ToolCircuitBreakerTest` (supervision scope, every state transition, probe claim/release/expiry, backoff and overflow clamping, config validation), 16 in `ToolCircuitBreakerMiddlewareTest` (a tripped tool is removed from the offered schemas, messages/options survive filtering, the acting stream is forwarded unchanged, `DENIED`/`INTERRUPTED`/`RUNNING` do not trip the circuit, probe binding and release), 8 in `InMemoryToolCircuitBreakerStoreTest` (compare-and-set contract, including that exactly one of sixteen concurrent swaps wins).
- `mvn -pl agentscope-extensions/agentscope-extensions-redis clean verify` — passes locally: **13 tests, 0 failures**.

**Concurrency tests use deterministic coordination, not stress loops.** A delegating store suspends the compare-and-set that would trip the circuit, so a success can be interleaved into the exact window; and eight reasoning turns are released together after a cooldown to assert only one is offered the tool.

**I verified the tests fail without the fix**, rather than only that they pass with it. Two mutations reproduce the pre-fix behaviour:

| Mutation reproducing the old behaviour | Tests that failed |
|---|---|
| `onReasoning` advertises to every half-open caller without claiming the probe | `onlyOneConcurrentReasoningTurnReceivesTheHalfOpenProbe` (the only failure) |
| `compareAndSet` ignores `expected` and commits unconditionally | `successDuringThresholdFailurePreventsAStaleOpen`, plus 3 compare-and-set contract tests |

Both were reverted afterwards.

## Checklist

- [x] All tests are passing (`mvn test`) — `agentscope-core` 2371 tests and `agentscope-extensions-redis` 13 tests pass locally; full reactor build runs in CI
- [x] Javadoc comments are complete and follow project conventions — `mvn -pl agentscope-core javadoc:javadoc` reports no warnings
- [ ] Related documentation has been updated (e.g. links, examples, etc.) — no docs page added yet; the API is documented in class-level javadoc with a usage example on `ToolCircuitBreakerMiddleware`. Happy to add a page under `docs/` if maintainers want one, once the API shape is agreed.


CryoThrust (commented):
One concurrency edge case to consider before merge: the implementation documents HALF_OPEN as a single recovery probe, but `state()` only derives `HALF_OPEN` from the elapsed timestamp and does not claim a probe. Once the cooldown elapses, concurrent reasoning turns can all observe `HALF_OPEN`, advertise the tool, and invoke it before any result is recorded. Each successful call will close the circuit, while concurrent failures can call `open()` repeatedly and advance the generation/backoff more than once. The same race exists across replicas with the Redis store.

Could the store/middleware expose an atomic half-open permit (for example, compare-and-set OPEN -> HALF_OPEN with a lease/token) so only one in-flight probe is allowed, while other turns continue withholding the tool until that probe succeeds or its lease expires? If the intended policy is to allow concurrent probes, the state-machine/Javadoc should say so and the tests should cover the behavior, since it differs from the stated “single probe” semantics.

CryoThrust (commented):
A second concurrency issue affects the “consecutive failures” contract, independently of the HALF_OPEN probe race above. `recordFailure()` performs `recordFailure()` -> checks the returned count -> `open()` -> `resetFailures()` as separate store operations, while `recordSuccess()` can call `resetFailures()` between those steps. A possible interleaving is: (1) failure A increments the closed counter to the threshold; (2) a successful call B observes CLOSED and clears the counter; (3) failure A still calls `open()` using its stale threshold result and trips the circuit even though the failures were not consecutive. The same window exists with the Redis store because the increment and open scripts are separate commands. Please make the threshold transition conditional/atomic (for example, a store operation that records the outcome and opens only if the same counter version still reaches the threshold, or a CAS/version token), or document and test a weaker policy. A deterministic concurrent test with a barrier between the threshold check and `open()` would expose this regression.

pure11011:
Thanks for both reviews — I could reproduce each one, and they turned out to share a root cause, so I fixed them together in e7c82dc rather than patching each site.

**Root cause.** The store made individual methods atomic, but a policy transition spanned several of them and the SPI had no conditional write. Every decision was therefore a read-modify-write split across calls, and both races lived in the gaps.

**What changed.** All state for a tool now lives in one immutable `ToolCircuitSnapshot` (failure count, generation, opened-at, probe token, probe lease), and every transition is published as a single compare-and-set against the exact value the caller observed. The SPI is now `snapshot` / `compareAndSet` / `reset`; a caller that loses the race re-reads and recomputes.

**On the consecutive-failure contract (second comment).** Counting the final failure and opening the circuit are now the same compare-and-set: the update is computed from the pre-increment snapshot and goes straight to the open state, so the counter is never persisted at the threshold. There is no longer an intermediate "threshold reached but not open" state to act on, and if a success clears the streak first, the failure's CAS fails and it recounts from the cleared state.

**On the half-open probe (first comment).** I took your first option rather than documenting the weaker policy, because the generation inflation was not only a documentation mismatch: N concurrent failing probes advanced the backoff N times, so the 60→120→240→480 ramp was skipped and the circuit jumped straight to the ceiling — and it got worse the more replicas you added, which is backwards. `tryAcquireProbe` now compare-and-sets a token plus lease into the snapshot, so exactly one turn advertises the tool while the others keep withholding it. Outcomes are conditional on the token, so a late result from a superseded probe cannot close or re-open a newer one, and the lease (`probeTimeout`, default 5 minutes to match `ExecutionConfig.TOOL_DEFAULTS`) lets another caller take over when a holder never reports. The claim rides on the per-call `RuntimeContext`, bound to the tool-call id the model chooses; a turn that never calls the advertised tool, a refused call and a cancelled call all hand it back early, while a suspended (`RUNNING`) call keeps it because its outcome is still pending.

Worth noting that the CAS retry bounds the inflation independently of the permit: if several failures race, the first commits the next generation and the rest re-read, see a freshly opened circuit and are ignored — so the generation advances at most once per cooldown window.

agentscope-core/src/main/java/io/agentscope/core/tool/circuitbreaker/ToolCircuitBreaker.java
```diff
@@ -0,0 +1,445 @@
...
+        while (true) {
+            ToolCircuitSnapshot current = store.snapshot(toolName);
+            long now = clock.millis();
+            ToolCircuitState state = stateOf(current, now);
+            if (state == ToolCircuitState.OPEN) {
+                // Withheld yet still executed: tolerated, see recordFailure.
+                return;
+            }
+            if (state == ToolCircuitState.HALF_OPEN) {
+                if (current.hasActiveProbe(now)
```
oss-maintainer, line 304: HALF_OPEN transitions are guarded only when the stored probe is *active* and the token differs. If the probe lease expires before a stale/non-probe result arrives (and no new probe is claimed), `hasActiveProbe` is false and the result is allowed to close or reopen the circuit. The check should verify that the result token matches the stored probe token, so only genuine probe results transition HALF_OPEN. Same issue exists in `recordFailure` at line 364.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/circuitbreaker/RedisToolCircuitBreakerStore.java
```diff
@@ -0,0 +1,214 @@
...
+    @Override
+    public void reset(String toolName) {
+        client.deleteKeys(circuitKey(toolName));
+    }
+
+    /**
+     * Encode a snapshot, mapping CLOSED to the absent-key marker so "missing" and "closed" compare
+     * equal.
+     */
+    private static String encode(ToolCircuitSnapshot snapshot) {
```
oss-maintainer, line 153: The delimiter encoding concatenates `probeToken` with unescaped `:` separators. The current generator emits UUID hex, but any future token (or externally constructed snapshot) containing `:` produces six fields and is decoded as malformed, causing the circuit to fail open and lose state. Use length-prefixed, JSON, or escaped encoding for the persisted snapshot.

oss-maintainer (commented):
### Summary

Introduces a per-tool circuit breaker (`CLOSED → OPEN → HALF_OPEN`) with a pluggable store (in-memory + Redis), middleware integration into the ReAct loop, exponential cooldown, and probe-lease tokens, plus ~1.1k lines of tests. This is a well-structured feature — the CAS loop, injected `Clock`, and withheld-but-tolerated OPEN handling are all deliberate and mostly correct.

Verdict: comment, not request-changes — the issues below are real but contained, and the test suite is unusually thorough for a first pass on this area.

### Findings
8 inline comments — see the diff anchors. Headlines:
- **[Warning]** `ToolCircuitBreaker.java:304` — the HALF_OPEN token guard only applies while `hasActiveProbe(now)` is true; once a probe lease expires, a stale or non-probe result can close (or re-open) the circuit. Compare the token against the stored probe unconditionally in HALF_OPEN. Same shape in `recordFailure`.
- **[Warning]** `RedisToolCircuitBreakerStore.java:153` — the `:`-delimited snapshot encoding has no escaping; any `probeToken` containing `:` decodes as malformed and the state is dropped.
- **[Warning]** `RedisToolCircuitBreakerStore.java:116` — default 24h TTL vs. an unbounded `maxCooldown`: a longer cooldown silently fails open when the key expires mid-cooldown. Validate `stateTtl > maxCooldown` at construction.
- **[Warning]** `ToolCircuitBreakerMiddleware.java:269` — with a null `RuntimeContext` outcomes are recorded without a probe token, so a successful probe cannot close HALF_OPEN.
- **[Warning]** `RedisToolCircuitBreakerStoreTest.java:40` — the fake adapter records script arguments but never executes the Lua, so the atomic compare-and-set semantics (the whole point of the Redis path) are untested.

### Suggestions
Add a null-guard for `reset(null)` on the Redis store (`circuitKey(null)` NPEs while the in-memory store tolerates it), document the middleware's `order()` expectations relative to retry/caching middleware, and cover the expired-lease probe case in `ToolCircuitBreakerTest`.

CLA `signed` (`license/cla=success`), `mergeable=CLEAN`. Not reviewed for the `distribution` multi-tenant isolation angle: if a breaker key is only scoped by tool name, confirm that two tenants sharing one Redis instance cannot open each other's circuits.

---

agentscope-core/src/main/java/io/agentscope/core/tool/circuitbreaker/ToolCircuitBreakerMiddleware.java
```diff
@@ -0,0 +1,303 @@
...
+        String toolName = result.getToolCallName();
+        ToolResultState state = result.getState();
+        if (toolName == null || state == null) {
+            return;
+        }
+        if (state == ToolResultState.RUNNING) {
+            return;
+        }
+        String token = consumeProbeToken(ctx, result.getToolCallId(), toolName);
+        switch (state) {
```
oss-maintainer, line 269: When `RuntimeContext` is null the middleware records outcomes without a probe token. In `ToolCircuitBreaker.recordSuccess`, a HALF_OPEN circuit with an active probe ignores a null-token success, so a successful probe can never close the circuit without a context. Either document that null-context usage cannot complete probes, or short-circuit the probe claim when `ctx == null`.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/circuitbreaker/RedisToolCircuitBreakerStore.java
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * Create a store with an explicit key prefix and TTL.
+     *
+     * @param client Redis client adapter
+     * @param keyPrefix prefix for every key, letting environments share one Redis instance
+     * @param stateTtl how long non-closed state is retained; must be positive and should exceed the
+     *     breaker's maximum cooldown
+     */
+    public RedisToolCircuitBreakerStore(
```
oss-maintainer, line 116: The default TTL is 24h while `ToolCircuitBreakerConfig` lets users set `maxCooldown` to any positive duration. If `maxCooldown` exceeds the TTL, the Redis key expires mid-cooldown and the tool is offered again early (fail-open). Add a constructor-time check or warning requiring `stateTtl` to be greater than the configured `maxCooldown`.

agentscope-extensions/agentscope-extensions-redis/src/test/java/io/agentscope/extensions/redis/circuitbreaker/RedisToolCircuitBreakerStoreTest.java
```diff
@@ -0,0 +1,305 @@
...
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Client-side behaviour of {@link RedisToolCircuitBreakerStore}: key layout, the arguments handed to
+ * the compare-and-set script, and the encoding of persisted state.
+ *
+ * <p>Scope note: the Lua script is executed by Redis, so a fake client cannot run it. These tests
+ * cover the Java side — which key is addressed, what the script receives, and how stored values are
+ * encoded and decoded, including values a healthy writer would never produce. The script's
```
oss-maintainer, line 40: The unit tests use a fake `RedisClientAdapter` that records script arguments but never executes the Lua. The actual atomicity logic (missing-key to `''` conversion, absent-vs-closed comparison, conditional `DEL` vs `SET EX`) is therefore untested. Add an integration test against a live or embedded Redis to verify server-side behavior.

agentscope-core/src/main/java/io/agentscope/core/tool/circuitbreaker/ToolCircuitBreaker.java
```diff
@@ -0,0 +1,445 @@
...
+    /**
+     * Force a tool back to {@link ToolCircuitState#CLOSED}, discarding its failure streak,
+     * accumulated backoff and any probe claim.
+     *
+     * <p>Intended for operators who know a dependency is healthy again and do not want to wait out
+     * the cooldown.
+     *
+     * @param toolName tool to reset
+     */
+    public void reset(String toolName) {
```
oss-maintainer, line 410: `reset(null)` delegates to the store unconditionally. `InMemoryToolCircuitBreakerStore` handles a null key safely, but `RedisToolCircuitBreakerStore` will throw NPE inside `circuitKey(null)`. Guard the call with a null check or align with `supervises(null)` semantics.

agentscope-core/src/main/java/io/agentscope/core/tool/circuitbreaker/ToolCircuitBreakerMiddleware.java
```diff
@@ -0,0 +1,303 @@
...
+ *         .toolkit(toolkit)
+ *         .middleware(new ToolCircuitBreakerMiddleware(new ToolCircuitBreaker(config)))
+ *         .build();
+ * }</pre>
+ *
+ * <p>Share one breaker (and therefore one store) across the agents that call the same dependency, so
+ * they learn from each other's failures instead of each discovering the outage separately.
+ */
+public class ToolCircuitBreakerMiddleware implements MiddlewareBase {
+
```
oss-maintainer, line 95: The middleware relies on seeing the final tool list in `onReasoning` and the actual outcome in `onActing`, but it does not override `order()` and uses the default of 1. Without guidance, retry/caching middleware placed differently will distort failure counts or hide real outcomes. Document a recommended `order()` or provide an override.

agentscope-core/src/main/java/io/agentscope/core/tool/circuitbreaker/ToolCircuitBreaker.java
```diff
@@ -0,0 +1,445 @@
...
+            return Optional.empty();
+        }
+        while (true) {
+            ToolCircuitSnapshot current = store.snapshot(toolName);
+            long now = clock.millis();
+            if (stateOf(current, now) != ToolCircuitState.HALF_OPEN
+                    || current.hasActiveProbe(now)) {
+                return Optional.empty();
+            }
+            String token = UUID.randomUUID().toString().replace("-", "");
```
oss-maintainer, line 221: Every half-open probe creates a `UUID` and strips dashes, which allocates and draws entropy on a per-tool/per-recovery basis. Consider a cheaper unique token (e.g. `ThreadLocalRandom` bytes encoded base62) since tokens only need uniqueness within one tool's probe history.

PR #2987, opened 2026-09-04
fix(models): route Qwen3.8 variants to DashScope multimodal API

### LINKED ISSUE #2981: [Bug] DashScopeHttpClient.isMultimodalModel() misses qwen3.8-flash (and other 3.8 variants except -max), causing AUTO endpoint routing to the wrong API
## Describe the bug

`DashScopeHttpClient.isMultimodalModel(String)` only routes models starting with
`qwen3.8-max` to the multimodal-generation endpoint. As of the Qwen3.8 release, the
whole commercial Qwen3.8 series (same as 3.5 / 3.6 / 3.7) is served through the
**multimodal-generation** API on DashScope native endpoints — but `qwen3.8-flash`
does not match any rule, so with the default `EndpointType.AUTO` it is routed to
`/api/v1/services/aigc/text-generation/generation` and the request fails
(URL/endpoint error from the server side).

## Affected version

- `io.agentscope:agentscope-extensions-model-dashscope:2.0.2`
- Also present on `master` as of today (verified against source), so it's not yet fixed.

## Relevant code

`io.agentscope.extensions.model.dashscope.DashScopeHttpClient` — `isMultimodalModel()`
(sources jar of 2.0.2, lines 389–407):

```java
// Reverse exclusions: text-only models whose prefix would otherwise match.
if (lowerModelName.equals("qwen3.6-max-preview")
        || lowerModelName.startsWith("qwen3.7-max")) {
    return false;
}
return lowerModelName.startsWith("qvq")
        || lowerModelName.contains("-vl")
        || lowerModelName.contains("-asr")
        || lowerModelName.startsWith("qwen3.5")   // entire 3.5 series → multimodal
        || lowerModelName.startsWith("qwen3.6")   // entire 3.6 series → multimodal
        || lowerModelName.startsWith("qwen3.7")   // entire 3.7 series (except -max) → multimodal
        || lowerModelName.startsWith("qwen3.8-max") // ← 3.8: only "-max" covered, "-flash" missed
        || lowerModelName.contains("kimi-k2.5")
        || lowerModelName.contains("kimi-k2.6");
Note the inconsistency: for 3.5/3.6/3.7 the rule covers the whole series prefix, while for 3.8 it was narrowed to -max only.

Steps to reproduce
DashScopeChatModel model = DashScopeChatModel.builder()
        .apiKey(System.getenv("DASHSCOPE_API_KEY"))
        .modelName("qwen3.8-flash")
        .stream(true)
        .build();
// AUTO detection → text-generation endpoint → request fails against DashScope,
// because qwen3.8-flash is only available on the multimodal-generation API.
Symptom: call fails with an endpoint/url error from the server (the model is not served by the text-generation API), while qwen3.8-max works fine with the same code.

Expected behavior
isMultimodalModel("qwen3.8-flash") (and other non-max Qwen3.8 commercial variants) should return true, so EndpointType.AUTO routes them to /api/v1/services/aigc/multimodal-generation/generation.

Suggested fix
Align the 3.8 rule with the 3.5/3.6/3.7 convention:

- || lowerModelName.startsWith("qwen3.8-max")
+ || lowerModelName.startsWith("qwen3.8")
(If some future 3.8 text-only variant needs the old text endpoint, add it to the reverse-exclusion block at the top instead, mirroring qwen3.7-max.)

Workaround (current releases)
Explicitly override the routing:

DashScopeChatModel.builder()
        .apiKey(...)
        .modelName("qwen3.8-flash")
        .endpointType(EndpointType.MULTIMODAL)
        .build();

### PR DESCRIPTION
**Pull request by ZHCHJ888**
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #2981.

With the default `EndpointType.AUTO`, `qwen3.8-flash` is not recognized as multimodal, so requests use the text-generation endpoint and simple text message formatting. Match the Qwen3.8 family to select the multimodal endpoint and content-part formatting, while explicitly preserving text-generation routing for `qwen3.8-2.4t-a95b` and its variants.

The [DashScope documentation](https://help.aliyun.com/zh/model-studio/text-generation) lists `qwen3.8-max`, `qwen3.8-flash`, and `qwen3.8-27b` as requiring the multimodal API, and `qwen3.8-2.4t-a95b` as requiring the text API. The exclusion avoids changing the latter when widening the prefix match.

Add regression coverage for model names, case and snapshot suffixes, the text-only exclusion, and the actual HTTP endpoint and message content format for streaming and non-streaming Flash requests.

### Validation

Java 17 on Windows:

```sh
mvn -pl agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope -am test -Dtest=DashScope*Test
```

- Before the fix: 6 of the 13 targeted regression cases fail, including both Flash HTTP routing cases.
- After the fix: all 13 regression cases pass. The DashScope suite reports 460 tests, 0 failures, 0 errors, and 2 existing disabled tests (458 passed).
- Spotless checks bound to compilation pass, and `git diff --check` passes.
- Tests use local mock services; no live model API was called. Broader repository results are recorded below.

### Full repository validation (local)

A subsequent run covered the repository's default Maven test scope on Windows with Java 17: **8,417 tests, 8,244 passed, 0 failures, 2 errors, and 171 skipped**. Counts were checked against both XML test cases and Maven module summaries.

```sh
mvn -B -ntp -fae -Dmaven.test.failure.ignore=true test
```

The initial 89-module reactor stalled at the examples' npm audit requests. Its completed test reports were retained, and all 28 test classes in the remaining examples were run with `-pl` / `-am` and a complete `-Dtest` class list. The retry used process-local `npm_config_fetch_timeout=15000` and `npm_config_fetch_retries=0`, retaining npm auditing. Paw, DataAgent, CodingAgent, and CopilotKit then completed successfully, including their frontend builds. Thus the results combine the root run and the remaining-example retry; the root command did not finish successfully in a single invocation.

The two errors are in the existing Core `DangerousPathBypassTest.symlinkToSshIsDetected` and `symlinkToDotEnvIsDetected` tests. Both fail during setup at `Files.createSymbolicLink` because the Windows account lacks the required privilege, before reaching their assertions. All other executed tests passed. Existing skip conditions, including disabled model E2E tests, were retained; no live model API was called.

`maven.test.failure.ignore` was used only to continue collecting test results after errors. The all-tests-passing checkbox remains unchecked.

## Checklist

- [ ] All tests are passing (`mvn test`) — 2 existing Core tests error on local Windows symlink privileges; details above
- [x] Related documentation has been updated (routing Javadoc)

### DIFF
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeHttpClient.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeHttpClient.java
@@ -372,7 +372,8 @@ public class DashScopeHttpClient {
      *   <li>Models starting with "qwen3.5" (e.g., qwen3.5-plus, qwen3.5-flash)</li>
      *   <li>Models starting with "qwen3.6" (e.g., qwen3.6-plus, qwen3.6-flash)</li>
      *   <li>Models starting with "qwen3.7" where not "qwen3.7-max" (e.g., qwen3.7-plus)</li>
-     *   <li>Models starting with "qwen3.8-max" (e.g., qwen3.8-max)</li>
+     *   <li>Models starting with "qwen3.8" where not "qwen3.8-2.4t-a95b"
+     *       (e.g., qwen3.8-max, qwen3.8-flash, qwen3.8-27b)</li>
      *   <li>Models containing "kimi-k2.5"/"kimi-k2.6" (e.g., kimi-k2.6, kimi/kimi-k2.5)</li>
      * </ul>
      *
@@ -381,6 +382,8 @@ public class DashScopeHttpClient {
      *   <li>"qwen3.6-max-preview" — text-only preview model.</li>
      *   <li>Models starting with "qwen3.7-max" (e.g., qwen3.7-max, qwen3.7-max-2026-05-20)
      *       — text-only models that use the text-generation endpoint.</li>
+     *   <li>Models starting with "qwen3.8-2.4t-a95b" — text-only models that use the
+     *       text-generation endpoint.</li>
      * </ul>
      *
      * @param modelName the model name
@@ -393,7 +396,8 @@ public class DashScopeHttpClient {
         String lowerModelName = modelName.toLowerCase();
         // Reverse exclusions: text-only models whose prefix would otherwise match.
         if (lowerModelName.equals("qwen3.6-max-preview")
-                || lowerModelName.startsWith("qwen3.7-max")) {
+                || lowerModelName.startsWith("qwen3.7-max")
+                || lowerModelName.startsWith("qwen3.8-2.4t-a95b")) {
             return false;
         }
         return lowerModelName.startsWith("qvq")
@@ -402,7 +406,7 @@ public class DashScopeHttpClient {
                 || lowerModelName.startsWith("qwen3.5")
                 || lowerModelName.startsWith("qwen3.6")
                 || lowerModelName.startsWith("qwen3.7")
-                || lowerModelName.startsWith("qwen3.8-max")
+                || lowerModelName.startsWith("qwen3.8")
                 || lowerModelName.contains("kimi-k2.5")
                 || lowerModelName.contains("kimi-k2.6");
     }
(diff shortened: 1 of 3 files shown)


agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeHttpClient.java
```diff
@@ -393,7 +396,8 @@ public static boolean isMultimodalModel(String modelName) {
         String lowerModelName = modelName.toLowerCase();
         // Reverse exclusions: text-only models whose prefix would otherwise match.
         if (lowerModelName.equals("qwen3.6-max-preview")
-                || lowerModelName.startsWith("qwen3.7-max")) {
+                || lowerModelName.startsWith("qwen3.7-max")
+                || lowerModelName.startsWith("qwen3.8-2.4t-a95b")) {
```
zouyx, line 400: 这种写死是不是不太好，需不需要有一个统一配置地方更好？

@jujn

zouyx (approved):

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeHttpClient.java
jujn: > 这种写死是不是不太好，需不需要有一个统一配置地方更好？
> 
我先合了，后面有时间看看

Issue #2970, opened 2026-09-03
[Bug]: Lightweight compaction results are discarded when summarization is skipped


yyyzl:
## Describe the bug

`ConversationCompactor.compactIfNeeded()` performs argument truncation and tool-result pruning before deciding whether to summarize. Both lightweight passes return a replacement list when they change messages.

However, the method returns `Optional.empty()` in either of these cases:

1. The reduced conversation is below the summarization trigger.
2. The trigger fires, but the safe cutoff is zero because the conversation fits within the preserved tail.

`CompactionMiddleware` interprets the empty result as a no-op and forwards the original `ReasoningInput`. The reduced messages therefore reach neither the immediate model request nor the working `AgentState`. This can leave oversized tool content in subsequent requests even though the lightweight pass successfully reduced it.

## Reproduction

Reproduced on upstream `main` at `ea511ec2f68ca929a6bbc4e92cd0c0952a38c592` (`2.0.3-SNAPSHOT`), using Java 17 and a mocked model; no provider calls are required.

- Build a conversation containing a user message, an assistant tool call, and its matching tool result.
- Put `"x".repeat(10_000)` in the tool result.
- Configure `triggerTokens(1_000)`, `keepTokens(0)`, `keepMessages(20)` and pruning with `protectTokens(0)`, `minimumTokens(1)`, `maxOutputChars(100)`.
- Invoke `CompactionMiddleware.onReasoning()` and capture the input received by `next`.
- The tool result is pruned internally, but `next` receives the original input and state is unchanged.

Setting `triggerMessages(3)` reproduces the second early-return path (zero cutoff). Enabling argument truncation instead of result pruning reproduces both paths with a large `write_file` content argument.

A regression test exercises all four combinations at the real middleware boundary: all four fail on the baseline, while two unchanged-conversation controls pass.

## Expected behavior

Return the reduced conversation whenever a lightweight pass changed it, even when summarization is skipped. Reserve `Optional.empty()` for a genuine no-op. This lets existing callers update both the request and working state without adding an LLM call or changing the public method signature.

## Related work and scope

This is distinct from #2267 / #2319, which fixed `ToolResultEvictionMiddleware` updating state without updating the copied reasoning input. The issue here is the two early returns in `ConversationCompactor`.

This report does not propose changing pruning thresholds, protected windows, default argument truncation settings, or summary-error handling. I am preparing a small fix with regression tests.

oss-maintainer:
[...]

PR #2966, opened 2026-09-03
fix(core): preserve usage in structured output messages

xihongshichaojidan8:
## AgentScope-Java Version

2.0.3-SNAPSHOT (current main)

## Description

Fixes #2960.

Structured-output responses were rebuilt without preserving the canonical `Msg.usage` field:

- The native structured-output path copied content, metadata, and timestamp, but omitted `result.getUsage()`.
- The fallback path exposed aggregated usage only through metadata, leaving `Msg.getUsage()` null.

This change propagates usage through both message builders and adds regression assertions for the native and fallback structured-output paths.

Validation:

- `mvn -pl agentscope-core test` — 2318 tests run, 0 failures, 0 errors, 9 skipped.
- `mvn -pl agentscope-core spotless:apply` — successful, with no formatting changes required.

## Checklist

- [x] Code has been formatted with `mvn spotless:apply`
- [x] All tests are passing (`mvn test`)
- [x] Javadoc comments are complete and follow project conventions (no public API changes)
- [x] Related documentation has been updated (not applicable; behavior fix with regression coverage)
- [x] Code is ready for review

PR #2990, opened 2026-09-05
fix(harness): prevent missed compaction for multilingual contexts

### LINKED ISSUE #2985: [Bug]:上下文压缩失效导致agent运行失败
`"metadata": {
      "_chat_usage": {
        "inputTokens": 76789,
        "outputTokens": 1344,
        "cachedTokens": 0,
        "time": 76.132,
        "totalTokens": 78133
      }
    },
    "timestamp": "2026-09-04 17:16:02.922",
    "usage": {
      "inputTokens": 76789,
      "outputTokens": 1344,
      "cachedTokens": 0,
      "time": 76.132,
      "totalTokens": 78133
    }`
从studio上观测到最后一条totalTokens已经达到了78133，设置得triggerTokens=70000，通过断点调试`ConversationCompactor.compactIfNeeded,TokenCounterUtil.calculateToken(messages)`最终计算出来的token=3.8万，汉字居多，导致上下文超限，模型是本地部署Qwen3_5-27B-FP8

guslegend0510:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes https://github.com/agentscope-ai/agentscope-java/issues/2985

Chinese-heavy conversations could exceed the model's context window before compaction triggered. The estimator used the same character ratio for every language, while the trigger excluded system prompts and tool definitions.

This change:
- Counts non-ASCII text more conservatively while preserving the existing ASCII estimate, and includes thinking and hint text.
- Includes system prompts, tool definitions and response-format schemas in the trigger budget.
- Uses the primary reasoning model's context window for dynamic thresholds, even when a separate summary model is configured.
- Applies argument truncation and tool-result pruning even when summarization is skipped.
- Preserves original history when summarization fails or returns empty text.
- Retains existing public method signatures and updates English and Chinese documentation.

Token counts remain estimates; exact Qwen tokenizer accuracy and live deployment behavior have not been validated.

## Validation

- 40 focused compaction tests passed.
- Full Core suite: 2,318 tests, 0 failures/errors, 9 skipped.
- Full Harness suite: 942 tests, 0 failures/errors, 5 skipped.
- The 15 new budget regression cases produced 14 failures on unchanged main and all passed with this fix.

Commands:
- mvn -pl agentscope-harness -am -Dtest=TokenCounterUtilTest,CompactionBudgetTest,ConversationCompactorTest,CompactionMiddlewareTest -Dsurefire.failIfNoSpecifiedTests=false test
- mvn -pl agentscope-harness -am test
- mvn -pl agentscope-harness -am spotless:check -DskipTests

## Related links

- Issue: https://github.com/agentscope-ai/agentscope-java/issues/2985
- Commit: https://github.com/guslegend0510/agentscope-java/commit/408e0644413803fc140d72e75450424245821a41
- Changes: https://github.com/agentscope-ai/agentscope-java/compare/main...guslegend0510:codex/fix-2985-compaction-token-budget

## Checklist

- [x] Full Core/Harness reactor tests pass; other modules were not run
- [x] Javadoc comments have been updated


CryoThrust (commented):
The new request-overhead accounting is used to decide *whether* compaction starts, but it is not applied when choosing the preserved tail. `compactIfNeeded(..., requestOverheadTokens)` calls `determineCutoffIndex(messages, totalTokens, config)` with conversation-only tokens, and `findTokenBasedCutoff` still keeps up to the full `config.getKeepTokens()`. The rebuilt request then adds the same system/tool/response-schema overhead back on top of that tail.\n\nFor example, with `keepTokens=800`, a preserved 800-token conversation suffix and 400 fixed system/tool/schema tokens, compaction can report success but send roughly 1200 input tokens before the summary itself is counted. This is especially likely for large tool schemas, which is one of the cases this PR explicitly adds to the trigger budget.\n\nCould the cutoff budget be `max(0, keepTokens - requestOverheadTokens)` (with a documented behavior when the fixed overhead alone exceeds the budget), and could there be a regression test asserting that the post-compaction request estimate, including system/tools/response format, stays within the resolved budget? Otherwise the change detects the overflow earlier but cannot reliably make the final request fit.

birdie7761:
The conservative non-ASCII counting and the expanded budget scope look right, and the red-on-main budget tests are strong evidence.

One enhancement worth considering as a follow-up layer: anchor the estimate to the provider-reported usage of the last successful call. Msg usage is now reliably preserved (#2966), so the middleware can hold (lastUsage, surfaceFingerprint) and compute `effective = max(lastReportedInputTokens + heuristicDelta, pureHeuristic)` — the heuristic then only prices the delta between anchors, so systematic bias (CJK ratios, chat templates, tool schemas, multimodal) is absorbed by the ground truth on every call. DeepSeek's open-source harness (deepseek-harness, packages/llm/token-meter) ships exactly this anchored-measurement design in production, including a conservative guard (usage accepted only when ≥ the heuristic price of the same surface) and route-consistency (anchor reused only under the same provider/model).

This would complement rather than replace the estimator fix here — the fixed heuristic remains the bootstrap before the first anchored call and the delta layer afterwards.

Separately: the "prune even when summarization is skipped" and "preserve history on failed summary" parts overlap with #2971 (fixes #2970) — worth sequencing with the maintainers so the two PRs don't churn the same ConversationCompactor paths. And since explicit `triggerTokens` users (especially Chinese-heavy ones) will see compaction fire noticeably earlier after this lands, a release-notes line would help.

guslegend0510:
> The conservative non-ASCII counting and the expanded budget scope look right, and the red-on-main budget tests are strong evidence.
> 
> One enhancement worth considering as a follow-up layer: anchor the estimate to the provider-reported usage of the last successful call. Msg usage is now reliably preserved (#2966), so the middleware can hold (lastUsage, surfaceFingerprint) and compute `effective = max(lastReportedInputTokens + heuristicDelta, pureHeuristic)` — the heuristic then only prices the delta between anchors, so systematic bias (CJK ratios, chat templates, tool schemas, multimodal) is absorbed by the ground truth on every call. DeepSeek's open-source harness (deepseek-harness, packages/llm/token-meter) ships exactly this anchored-measurement design in production, including a conservative guard (usage accepted only when ≥ the heuristic price of the same surface) and route-consistency (anchor reused only under the same provider/model).
> 
> This would complement rather than replace the estimator fix here — the fixed heuristic remains the bootstrap before the first anchored call and the delta layer afterwards.
> 
> Separately: the "prune even when summarization is skipped" and "preserve history on failed summary" parts overlap with #2971 (fixes #2970) — worth sequencing with the maintainers so the two PRs don't churn the same ConversationCompactor paths. And since explicit `triggerTokens` users (especially Chinese-heavy ones) will see compaction fire noticeably earlier after this lands, a release-notes line would help.

Thanks @birdie7761, the usage-anchored estimate makes sense as a follow-up, with the heuristic retained for bootstrap and delta estimation. I’d keep that separate from this fix so the usage guard and provider/model consistency can be tested independently.

I’ve merged the latest main, resolved the conflicts, and added English and Chinese release notes explaining that explicit triggerTokens may now trigger compaction earlier, especially for Chinese-heavy conversations.

Thanks also for flagging the overlap with #2971. I’m happy to coordinate the merge order with the maintainers and adjust the overlapping changes accordingly.

guslegend0510:
> The conservative non-ASCII counting and the expanded budget scope look right, and the red-on-main budget tests are strong evidence.
> 
> One enhancement worth considering as a follow-up layer: anchor the estimate to the provider-reported usage of the last successful call. Msg usage is now reliably preserved (#2966), so the middleware can hold (lastUsage, surfaceFingerprint) and compute `effective = max(lastReportedInputTokens + heuristicDelta, pureHeuristic)` — the heuristic then only prices the delta between anchors, so systematic bias (CJK ratios, chat templates, tool schemas, multimodal) is absorbed by the ground truth on every call. DeepSeek's open-source harness (deepseek-harness, packages/llm/token-meter) ships exactly this anchored-measurement design in production, including a conservative guard (usage accepted only when ≥ the heuristic price of the same surface) and route-consistency (anchor reused only under the same provider/model).
> 
> This would complement rather than replace the estimator fix here — the fixed heuristic remains the bootstrap before the first anchored call and the delta layer afterwards.
> 
> Separately: the "prune even when summarization is skipped" and "preserve history on failed summary" parts overlap with #2971 (fixes #2970) — worth sequencing with the maintainers so the two PRs don't churn the same ConversationCompactor paths. And since explicit `triggerTokens` users (especially Chinese-heavy ones) will see compaction fire noticeably earlier after this lands, a release-notes line would help.

Thanks @CryoThrust, you’re right—the fixed request overhead also needs to reduce the retained-tail budget. Updated in dbf5704 to use max(0, keepTokens - requestOverheadTokens), with regression tests for overhead consuming the budget and overhead exceeding it.

I also clarified the boundary in the English and Chinese docs: the final message and complete assistant/tool groups are preserved even when they exceed the remaining budget, and the generated summary adds tokens separately.

This addresses the tail-budget accounting issue, but the current tests do not establish a hard cap on the complete post-compaction request, including the summary. That part of your suggestion remains open.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/ConversationCompactor.java
```diff
@@ -108,14 +136,15 @@ public Mono<Optional<List<Msg>>> compactIfNeeded(
                         config.getPruneConfig());
```
oss-maintainer, line 137: The system prompt is double-counted in the compaction budget: `requestOverheadTokens` already includes the system message computed in CompactionMiddleware, but `messages` also has it prepended via `prependSystemMessage`. This inflates `requestTokens`, causing premature compaction and a smaller retained conversation tail than configured.

oss-maintainer (commented):
### Summary
Re-review of PR #2990 at its current head (automated review of the changes in this diff). 1 item(s) raised below; nothing here blocks the approach itself — details inline.

### Findings
- **[Critical]** `agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/ConversationCompactor.java:137` — The system prompt is double-counted in the compaction budget: `requestOverheadTokens` already includes the system message computed in CompactionMiddleware, but `messages` also has it prepended via `prependSystemMessage`. This inflates `requestTokens`, causing premature compaction and a smaller retained conversation tail than configured.

### Notes

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/ConversationCompactor.java
```diff
@@ -243,10 +276,18 @@ private static boolean shouldCompact(
...
+            CompactionConfig config,
+            int requestOverheadTokens) {
         int rawCutoff;
         if (config.getKeepTokens() > 0) {
-            rawCutoff = findTokenBasedCutoff(messages, totalTokens, config.getKeepTokens());
+            // Reserve fixed system/tool/schema overhead from keepTokens before selecting
+            // the conversation tail; the generated summary is additional. If the overhead
+            // exhausts the budget, findTokenBasedCutoff still preserves the final message as the
+            // minimum viable conversational tail.
+            int tailBudget = Math.max(0, config.getKeepTokens() - requestOverheadTokens);
```
oss-maintainer, line 289: When `requestOverheadTokens >= keepTokens` — realistic for a small-context model with a large tool catalogue, since `keepTokens` is itself derived from the same context window in dynamic mode (`contextWindow - trigger` headroom) — `tailBudget` becomes 0 and `findTokenBasedCutoff` keeps only the final message. Because the overhead alone already exceeds the trigger, `shouldCompact` stays true on the next reasoning step too, so the loop rewrites a one-message tail every turn: constant summarization calls, constant context churn, and effectively no conversational memory. Suggest a floor (e.g. keep at least `keepMessages`/the last N messages regardless of the token budget) or a guard that skips token-based compaction and logs once when `requestOverheadTokens >= keepTokens`, so the failure is visible instead of silently degenerate. A regression test for that configuration would be valuable.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/ConversationCompactor.java
```diff
@@ -372,15 +413,8 @@ private Mono<String> summarizePrefix(List<Msg> prefix, CompactionConfig config)
...
-                .defaultIfEmpty("(Summary unavailable)")
-                .onErrorResume(
-                        e -> {
-                            if (ExceptionUtils.containsInterruptedException(e)) {
-                                return Mono.error(e);
-                            }
-                            log.warn("Summarization LLM call failed: {}", e.getMessage());
-                            return Mono.just("(Summarization failed: " + e.getMessage() + ")");
-                        });
+                .switchIfEmpty(
```
oss-maintainer, line 416: Removing the placeholder + `onErrorResume` here means a transient summarization error now aborts compaction for that turn (the middleware logs and continues with an uncompacted context). That is better than writing a `(Summarization failed)` marker into the history, but it removes the only bound on growth: if the summarizing model keeps failing, context grows until the reasoning call itself is rejected. Consider a last-resort degrade that does not need the LLM — e.g. apply the existing prune/truncate steps and drop the oldest prefix without a summary — or at minimum surface a metric/WARN so operators can tell "compaction permanently disabled" from "compaction not needed yet".

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/TokenCounterUtil.java
```diff
@@ -75,13 +74,39 @@ public static int calculateToken(List<Msg> messages) {
...
+     * @param messages all request messages, including system messages
+     * @param tools tool definitions, or null
+     * @param options request options, or null
+     * @return estimated input tokens, saturated at {@link Integer#MAX_VALUE}
+     */
+    public static int calculateToken(
+            List<Msg> messages, List<ToolSchema> tools, GenerateOptions options) {
+        long tokens = calculateToken(messages);
+        if (tools != null && !tools.isEmpty()) {
+            tokens += estimateTextTokens(JsonUtils.getJsonCodec().toJson(tools));
```
oss-maintainer, line 102: `CompactionMiddleware` recomputes this on every reasoning step, and `input.tools()` is effectively constant for the life of an agent, so the full tool-catalogue JSON serialization + code-point scan is repeated work on the hot path (same comment applies to `estimateTextTokens`, which no longer has an O(1) path). Caching the estimate against the tools list identity, or computing it once when the middleware is constructed, keeps the accuracy without the per-step cost.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/compaction/ConversationCompactor.java
```diff
@@ -210,12 +239,16 @@ public Mono<Optional<List<Msg>>> compactIfNeeded(
                                                 }));
     }
 
+    private static Optional<List<Msg>> preprocessedResult(List<Msg> original, List<Msg> processed) {
+        return processed == original ? Optional.empty() : Optional.of(processed);
```
oss-maintainer, line 243: Identity comparison as the "did preprocessing change anything" signal is a bit fragile: it depends on every preprocessing step returning the *same* instance when it is a no-op. A future step that returns a fresh but equal list turns "nothing to do" into "replace the whole context" on every turn. An explicit `changed` flag threaded through preprocessing (or `List.equals`) would encode the intent directly.

oss-maintainer (commented):
### Summary
Real bug, well diagnosed: a flat 2.5 chars/token ratio under-counts CJK and other non-Latin scripts, so the compaction trigger can stay below threshold and the context is never compacted. Per-code-point widths (ASCII 0.4, BMP non-ASCII 1.0, supplementary 2.0), `long` accumulation with saturating casts, `HintBlock`/`ThinkingBlock` coverage, and folding system prompt + tool schemas + response format into the decision are all improvements, and the added `CompactionBudgetTest`/`TokenCounterUtilTest` coverage is substantial. Using the reasoning model rather than the compaction model for dynamic thresholds is also the correct call.

### Findings
- **[Warning]** `tailBudget = max(0, keepTokens - requestOverhead)` can collapse to 0 and then thrash every turn when prompt overhead alone exceeds the keep budget.
- **[Warning]** summarization failures now propagate, which removes the previous (ugly but bounding) fallback and can leave the context unbounded.
- **[Info]** tool-schema JSON is re-serialized on every reasoning step — cacheable.
- **[Info]** `processed == original` identity check encodes \"did preprocessing change anything\" implicitly.

### Suggestions
The `msg.withContent(...)` substitutions in `pruneToolResults`/`truncateToolUseArgs` are a genuine cleanup — the hand-rolled `Msg.builder()` copies dropped nothing today only because the field list happens to match.

### Verdict
Comment only, not approval: the two [Warning]s are behavioural edges in the core memory path (which cascades into harness/distribution), and the PR still has a merge conflict with `main`.

---

oss-maintainer:

```bash
git fetch origin
git checkout codex/fix-2985-compaction-token-budget
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

Issue #2982, opened 2026-09-04
[Bug]:2.0.0版本HarnessAgent配置了MemoryConfig.FlushTrigger和consolidationMinGap，但是第一次运行时总会触发consolidation


2325381878-droid:

oss-maintainer:
**Issue Evaluation**

Category: `bug` | Status: **Evaluated — behaviour is by design; needs the reporter's expectation confirmed**

The issue body is empty, so this evaluation is built from the title alone (2.0.0, `HarnessAgent` with `MemoryConfig.FlushTrigger` and `consolidationMinGap`, first run still flushing/consolidating immediately). Verified against `main` @ `7399ed6`.

**Why the first call always runs:** both periodic gates seed the "last successful claim" instant to the epoch when no prior entry exists, so `Duration.between(EPOCH, now)` exceeds any configured `minGap`.
- `agentscope-harness/src/main/java/io/agentscope/harness/agent/coordination/LocalPeriodicGate.java:34,46` — `EPOCH` constant and `computeIfAbsent(name, k -> new AtomicReference<>(EPOCH))`.
- `.../coordination/StoreBackedPeriodicGate.java:80-82` — missing store item returns `EPOCH`.
- `.../middleware/MemoryFlushMiddleware.java:291-302` — `shouldFlushNow()` delegates to `periodicGate.tryClaim(key, minGap)` in THROTTLED mode.
- `.../middleware/MemoryMaintenanceMiddleware.java:165-171` — consolidation/maintenance uses the same pattern under its own key prefix.
- `.../memory/MemoryConsolidator.java:307-342` — watermark `EPOCH` means "first run", and `:224` then admits every daily ledger, including the one flush just wrote in the same call.

[...]

PR #2993, opened 2026-09-05
fix(harness): isolate memory flush and maintenance throttles

guslegend0510:
THROTTLED memory flush and background maintenance share the same periodic gate key, so either operation can suppress the other even though they have separate configured intervals.

This change gives each operation a distinct key prefix and adds regression coverage for local/store-backed gates, all isolation scopes, and both execution orders.

It also clarifies first-call behavior in MemoryConfig and the English/Chinese documentation: minimum gaps limit subsequent runs; they do not delay the first eligible call. FlushTrigger.never() only disables per-call flush.

Validation:
- 16 new regression cases fail before the fix and pass afterward.
- Core: 2,318 tests, 0 failures/errors, 9 skipped.
- Harness: 935 tests, 0 failures/errors, 5 skipped.
- Spotless and git diff --check pass.

Upgrade note: existing store-backed entries use the old shared key. Each new operation key starts a fresh throttle window after upgrade.

Related to #2982.

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/coordination/PeriodicGate.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/coordination/PeriodicGate.java
@@ -28,7 +28,8 @@ public interface PeriodicGate {
     /**
      * Attempts to claim the named periodic slot.
      *
-     * @param name logical slot identifier (e.g. {@code "USER:alice"} or {@code "skill-curator"})
+     * @param name logical slot identifier (e.g. {@code "memory-flush:USER:alice"} or
+     *     {@code "skill-curator"})
      * @param minGap minimum elapsed time since the last successful claim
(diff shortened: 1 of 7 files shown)


CryoThrust:
I reviewed the current diff and the throttle contract looks coherent. Prefixing the operation into the composite key isolates flush and maintenance for both local and store-backed gates, while preserving scope isolation; the parameterized matrix covers both execution orders and middleware reconstruction.

The upgrade note is important: old shared keys are intentionally not migrated, so each operation gets a fresh window after upgrade. That matches the documented semantics and avoids treating a historical shared timestamp as either operation's successful run. The first-call behavior and `FlushTrigger.never()` boundary are also now explicit in both language docs.

I did not find a blocking issue in the current patch. The remaining merge gate is maintainer review.

jujn (approved):

Issue #2994, opened 2026-09-05
[Feature]:  add dashscope_image_to_image (image editing) tool to DashScopeMultiModalTool


tanzzj:
**Is your feature request related to a problem? Please describe.**

DashScopeMultiModalTool currently exposes dashscope_text_to_image, dashscope_image_to_text, dashscope_text_to_video, dashscope_image_to_video, dashscope_first_and_last_frame_image_to_video, dashscope_text_to_audio / dashscope_audio_to_text, and dashscope_video_to_text — but there is no image → image (image editing) tool.

**Describe the solution you'd like**

As a result, when a user uploads an image and asks for an edit that must preserve the original composition ("colorize this sketch", "change the background", "turn my line drawing into a finished illustration"), the agent has no tool that accepts an input image plus a prompt. The only available path is dashscope_text_to_image, which redraws from a textual description of the image: the composition, line work and layout are silently lost, and the tool still returns succeeded, so neither the agent nor the user can tell that the original image was never used. In our production deployment this consistently produced plausible-looking but entirely unrelated results, and the agent had to be explicitly instructed not to degrade to text-to-image.

**Describe alternatives you've considered**
A new @Tool method on DashScopeMultiModalTool, symmetric with the existing ones:

```java
@Tool(
    name = "dashscope_image_to_image",
    description =
        "Edit or transform an existing image based on a text prompt and an input image."
            + " Use dashscope_text_to_image when there is no input image.")
public Mono<ToolResultBlock> dashscopeImageToImage(
        @ToolParam(name = "image_url", ...) String imageUrl,
        @ToolParam(name = "prompt", ...) String prompt,
        @ToolParam(name = "model", required = false) String model, // default: qwen-image-edit
        @ToolParam(name = "size", required = false) String size,
        @ToolParam(name = "use_base64", required = false) Boolean useBase64) {
}
```
Behaviour:
- Calls MultiModalConversation (the multimodal-generation/sync endpoint) with [{image: ref}, {text: prompt}], model defaults to qwen-image-edit; other edit models (qwen-image-edit-plus, qwen-image-edit-max, qwen-image-2.0-pro) selectable per call.
- Returns an ImageBlock with URLSource, or Base64Source when use_base64=true — i.e. the exact same result shape as dashscope_text_to_image, so nothing downstream (formatters, frontends, media guards) needs to change.
[...]

PR #2995, opened 2026-09-05
feat(dashscope): add dashscope_image_to_image (image editing) tool to DashScopeMultiModalTool

tanzzj:
## AgentScope-Java Version

`2.0.3-SNAPSHOT` (branch base `origin/main` @ `c32de522`; local `<revision>` in the root `pom.xml`)

### Background

`DashScopeMultiModalTool` covers text-to-image, image-to-text, text-to-video,
image-to-video, first-last-frame-to-video, TTS/ASR and video-to-text, but has no
**image-to-image (image editing)** tool. An agent asked to edit a user-provided image
("colorize this sketch", "replace the background") therefore has no tool that takes
an input image plus a prompt, and falls back to `dashscope_text_to_image`, which
redraws from a *textual description* of the image. The original composition is lost and
the call still reports success, so the degradation is silent.

### Changes made

One file, additive: `agentscope-extensions-model-dashscope/.../dashscope/tool/DashScopeMultiModalTool.java`

- New `@Tool` method `dashscopeImageToImage(...)` registered as `dashscope_image_to_image`,
  parameters `image_url`, `prompt`, `model` (optional), `size` (optional), `use_base64` (optional).
- Calls `MultiModalConversation` (the `multimodal-generation` endpoint) with content
  `[{image: ref}, {text: prompt}]` and defaults to `qwen-image-edit`; returns an `ImageBlock`
  with `URLSource`, or `Base64Source` when `use_base64=true` — the same result shape as
  `dashscope_text_to_image`, so formatters and downstream consumers need no change.
- Guards for behaviours observed against the live API:
  - `rejectsImageInput(model)`: the `wan*` family and the text-to-image members of the
    `qwen-image` family cannot consume an image on this endpoint, so they are rejected
    locally with an actionable message instead of an opaque service-side parameter error
    (without this the model retries with a text-to-image model and the silent degradation
    comes back).
  - The request carries exactly one message with role `user`; a leading `system` message or
    a second message is rejected by the endpoint.
  - `resolveOutputSize` / `supportsOutputSize`: only `qwen-image-edit-plus`,
    `qwen-image-edit-max` and the `qwen-image-2.0` series accept an explicit resolution, so
    `size` is dropped with a `log.warn` for other models rather than failing the whole call
    (output resolution then follows the input image).
  - `toEditableImageRef`: an existing local file is inlined as a Base64 data URL, because this
    endpoint rejects the `file://` protocol that the vision tools use via
    `MediaUtils.urlToProtocolUrl()`.
  - When no image comes back, the `text` field of the same content map is surfaced as the
    error message (that is where a model-side refusal lands).
- Class-level Javadoc and the tool `description` updated to state the input formats and to
  steer back to `dashscope_text_to_image` when there is no input image.

### How to test

Static / unit (no key required):

```bash
mvn spotless:apply -pl agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope
mvn test -pl agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope
```

Manual, against the real service:

```bash
export DASHSCOPE_API_KEY=sk-xxx
```

then register the tool on an agent and ask "edit this image: <public image url>, make it
night-time". Expected: one `ImageBlock(URLSource)` in the tool result and a visibly edited
image. Negative cases worth checking: `model=wanx2.1-t2i-turbo` -> local rejection with the
hint message; `model=qwen-image-edit` + `size=1024*1024` -> `size` dropped with a warning and
a successful call; `image_url` pointing at a local file -> inlined as Base64 and accepted.

Also verified in a production deployment (an application that consumes this module via
`agentscope-extensions-model-dashscope`), where the tool replaced silent text-to-image
fallbacks for attachment-based edits.

``

### Not included

The tool tables under `docs/v1/en/**/tool.md` and `docs/v1/zh/**/tool.md` do
not list `dashscope_image_to_image` yet. Happy to add them here if you'd
prefer that over a follow-up PR.

## Checklist


CryoThrust:
The additive tool shape and local model guards look well-scoped. One security/privacy concern: the new DEBUG log records `imageUrl` and `prompt` verbatim. `imageUrl` may be a large `data:` URL containing user-provided image bytes, and prompts can contain sensitive content; this can leak payloads into application logs and create very large log entries. Could the log be reduced to safe metadata (for example, input-reference kind/length, model, size, and a bounded prompt length or no prompt), with a regression test or documented logging contract? Also, the PR description notes the tool docs are not updated; adding the tool to the generated/user-facing DashScope tool tables would make the feature discoverable or explicitly track that as a follow-up.

tanzzj:
> The additive tool shape and local model guards look well-scoped. One security/privacy concern: the new DEBUG log records `imageUrl` and `prompt` verbatim. `imageUrl` may be a large `data:` URL containing user-provided image bytes, and prompts can contain sensitive content; this can leak payloads into application logs and create very large log entries. Could the log be reduced to safe metadata (for example, input-reference kind/length, model, size, and a bounded prompt length or no prompt), with a regression test or documented logging contract? Also, the PR description notes the tool docs are not updated; adding the tool to the generated/user-facing DashScope tool tables would make the feature discoverable or explicitly track that as a follow-up.

removed the unnecessary logs and added example javadoc already, pls take a check

tanzzj:
#2994

jujn (approved):

PR #2996, opened 2026-09-05
fix(harness): inherit parent hooks in declared subagents

### LINKED ISSUE #1578: [Bug]:声明的子 Agent 不会继承父 Agent hooks
## 问题
  当前 AgentScope Harness 中，`general-purpose` 子 Agent 会继承父 Agent 的 hooks，但通过 `workspace/subagents/*.md` 声明的子 Agent 不会继承父 Agent hooks。
这会导致父 Agent 上配置的工具治理、HITL 人工确认、审计、预算等 hook 只对父 Agent 生效，对声明式子 Agent 不生效。

  ## 例子

  父 Agent 配置了工具确认 hook：

  ```java
  HarnessAgent.builder()
      .hook(new ToolGovernanceHook())
      .toolkit(toolkit)
      .workspace(workspace)
      .build();

  子 Agent 声明了受控工具：

  ---
  description: API 排错专家
  tools: [query_logs]
  ---

  预期：query_logs 在子 Agent 中调用时，也会触发父 Agent 的 ToolGovernanceHook。

  实际：声明式子 Agent 没继承 hook，query_logs 可能直接执行。

  ## 期望

  声明式子 Agent 应该像 general-purpose 子 Agent 一样继承父 Agent hooks，同时继续保留 tools allowlist。

  也就是：

  - 父 Agent hooks 继续作为全局治理能力
  - 子 Agent 仍然只能使用自己声明的工具
  - 工具调用、HITL、审计、预算等逻辑在父 Agent 和子 Agent 中行为一致

  ## 可能的修复方向

  在构建声明式子 Agent 时，把父 Agent hooks 传入子 Agent builder：

  .hooks(List.copyOf(parentHooks))

  另外，子 Agent 执行时的 RuntimeContext 也最好继承父上下文中的必要字段，否则 hook 即使继承了，也可能拿不到租户、权限、runId 等信息。

  ## 影响

  这应该是更安全的默认行为。否则用户以为 hook 是全局治理，但声明式子 Agent 实际绕过了它。

ordinary-s:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #1578

Automatically constructed declared subagents omitted Hooks explicitly configured on their parent, so child events and tool argument interception never reached those Hooks. Snapshot the explicit parent Hook collection and install it on the child builder, matching the general-purpose factory. Programmatic declarations, static workspace Markdown and dynamic Markdown discovery share this factory.

Apply the declaration allowlist and the child's existing `tools.json` filter to Hook-contributed tools before installing them. This prevents forbidden tools from being reintroduced when runtime tool deletion is disabled or the declaration lists a workspace-denied tool. Child-owned Harness tools remain available, and empty declaration lists still honor workspace policy.

Add `ReActAgent.Builder.build(Consumer<Toolkit>)` to filter resolved Hook tools in a temporary toolkit before transfer to the configured agent toolkit. The zero-argument build path is unchanged. Reuse the original Hook instances, priority ordering and deduplication; preserve the agent toolkit's execution configuration, callbacks and reflective tool metadata. No Hook inheritance switch is added. General-purpose behavior, middleware/configuration inheritance, remote subagents and custom factories remain unchanged.

Deterministic local tests run parent `agent_spawn` and child tool execution, including real Markdown created before and after parent construction. They cover child event identity, tool events and argument rewriting, Hook-contributed AgentTool and annotated methods, actual denied execution, model-visible schemas, order/deduplication, empty Hooks, builtin preservation and general-purpose compatibility.

## Validation

Windows, Maven 3.9.11, Java 17.0.20.1. Final branch is based on main `0f0924a1b086cd11ac2a176f3c13825b429dc682`.

- `mvn -pl agentscope-harness -am "-Dtest=HarnessAgentTest#subagentHooks_*,HookToolsRegistrationTest" -Dsurefire.failIfNoSpecifiedTests=false -DforkCount=0 test`: 22 Harness regressions and 4 core Hook tests passed. Original missing-Hook assertions failed on unchanged production; the two effective-tool-restriction cases also ran red before their production correction and green afterward.
- `mvn -pl agentscope-harness -am "-Dtest=HarnessAgent*Test,Subagent*Test,*Subagents*Test,DefaultAgentManager*Test,AgentSpawnTool*Test,PlanModeSubagentPropagationTest,HookToolsRegistrationTest,ToolkitTest" -Dsurefire.failIfNoSpecifiedTests=false -DforkCount=0 test`: 234 passed after rebasing (193 Harness, 41 core).
- `mvn -pl agentscope-harness -am "-Dtest=io.agentscope.harness.**" -Dsurefire.failIfNoSpecifiedTests=false -DforkCount=0 test`: 945 cases, 0 failures/errors, 6 skipped.
- `mvn -pl agentscope-core -am test`: 2320 cases, 0 assertion failures, 2 errors, 8 skipped. Both errors are Windows symlink privilege failures in `DangerousPathBypassTest`; running that class on a clean worktree at the same main SHA reproduced both errors.
- `mvn spotless:check` and `git diff origin/main...HEAD --check`: passed.

Harness uses `forkCount=0` because default Surefire fork communication failed locally, including for an untouched existing test. No plugin configuration was changed. Full-reactor green status is not claimed, and skips are not counted as passes.

## Checklist

- [x] Changed tests formatted with scoped `spotless:apply`; full `spotless:check` passed.
- [x] All tests are passing (`mvn test`) — Windows limitations documented above.
- [x] Public build overload documented in Javadoc.
- [x] Regression tests cover the real declared-subagent execution paths.


agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -5280,6 +5281,23 @@ private void configureSkillBox(Toolkit agentToolkit) {
...
+         * contributions, including individual annotated methods. It can remove excluded tools
+         * before they are installed on the agent's toolkit, independently of that toolkit's
+         * runtime deletion policy. Existing tools, execution configuration and Hook instances are
+         * preserved. Passing {@code null} retains normal hook tool registration.
+         *
+         * @param hookToolFilter optional construction-time filter for hook-contributed tools
+         * @return a new ReActAgent instance
+         * @throws IllegalArgumentException if required parameters are missing or invalid
+         */
+        public ReActAgent build(Consumer<Toolkit> hookToolFilter) {
```
xiaoyuuuuuupeng: Non-blocking suggestion: since this callback is intended to decide which hook-contributed tools to retain, could the public API accept a `Predicate<String>` instead? `Consumer<Toolkit>` also allows callers to add or replace tools in the temporary toolkit, which is broader than the filtering contract described by the Javadoc. A predicate would express the intended keep/drop behavior more directly.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
```diff
@@ -517,7 +518,8 @@ static SubagentFactory buildDeclaredFactory(
             }
 
             sub.middlewares(capturedMiddlewares);
-            return sub.build();
+            sub.hooks(capturedHooks);
```
xiaoyuuuuuupeng, line 594: Non-blocking suggestion: could we document that declared subagents inherit all explicitly configured parent hooks, and how parent-only hooks should skip child events? This would help users avoid unexpected duplicate reporting or other side effects.

xiaoyuuuuuupeng (approved):
LGTM

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
ordinary-s: Thanks for the suggestion. Updated the overload to build(Predicate<String>). The predicate receives the resolved registered tool name (the same name exposed in the tool schema), with true retaining the Hook-contributed tool and false excluding it.

Filtering is applied after Hook tool resolution and only to Hook contributions. Existing tool registration, ordering/deduplication, toolkit configuration, callbacks, and reflective metadata remain unchanged. I also added regression coverage for these boundaries.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
ordinary-s: Thanks for pointing this out. I added public Javadoc to HarnessAgent.Builder.hook() and hooks() documenting that automatically constructed local declared subagents inherit the same explicitly configured parent Hook instances.

The documentation also explains that parent-only Hooks can compare event.getAgent().getAgentId() with the parent agent’s ID to skip child events and avoid duplicate reporting or other side effects. I added a regression test covering this behavior, including the case where the parent and child share the same name.

xiaoyuuuuuupeng:
Thanks for the update. I reviewed the follow-up changes. The Predicate<String> API now matches the intended filtering contract, and the added tests cover the important boundaries: resolved tool names, Hook ordering/deduplication, existing and built-in tools, toolkit configuration/metadata, and effective child tool restrictions.

The Hook inheritance Javadoc and parent-only filtering guidance also address my documentation concern. No further changes requested from my side.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
```diff
@@ -538,7 +539,9 @@ static SubagentFactory buildDeclaredFactory(
             }
 
             var childTools = childToolsConfig(capturedToolsConfig, decl.getTools());
-            if (childTools != null) sub.toolsConfig(childTools);
+            // Without an explicit parent config, apply the declaration after the child loads
+            // its workspace policy; a synthetic override would hide workspace allow/deny rules.
+            if (capturedToolsConfig != null) sub.toolsConfig(childTools);
```
oss-maintainer, line 544: **[Warning]** This guard change has a side effect beyond hook inheritance that is worth calling out explicitly: when `capturedToolsConfig == null` and the declaration lists tools, the child no longer receives a synthetic strict override, so `resolvedToolsConfig` is now loaded from the child's own `workspace/tools.json` at build time. Since `childToolsConfig(...)` copies `parent.getMcpServers()` into the child config, a declared subagent that previously got **no MCP servers registered** (the synthetic override carried `mcpServers == null`) will now materialize the MCP servers from its workspace file.

That is arguably the intended fix, but it changes external connectivity for existing declared subagents whose declaration has a `tools:` list while its workspace also declares `mcpServers`. Could you confirm that is intended, and either note it in the PR description / release note or add a regression test that pins the MCP-server set before/after for that shape? I could not exercise the build locally, so this is a read-only inference.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -5337,7 +5362,17 @@ public ReActAgent build() {
                 }
             }
 
-            registerToolsFromHooks(agentToolkit);
+            if (hookToolFilter == null) {
+                registerToolsFromHooks(agentToolkit);
+            } else {
+                Toolkit hookTools = new Toolkit();
```
oss-maintainer, line 5368: **[Warning]** The throwaway `Toolkit` is constructed with `ToolkitConfig.defaultConfig()` while `agentToolkit` carries the caller's config (custom `executorService`, `allowToolDeletion(false)`, chunk callbacks). Retained tools are then copied out with `registerAgentTool(AgentTool)`, i.e. through the overload that passes `groupName = extendedModel = mcpClientName = presetParameters = null`.

For today's hook path (`registerToolsFromHooks` -> `registerTool(Object)`, which also registers ungrouped with null presets) the two paths are equivalent — and `filterRunsAfterToolNameResolutionWithoutChangingHookOrderOrDeduplication` plus the metadata assertions prove the observable result matches, which is the part I most wanted pinned. Two residual concerns for this shape:

1. the temp toolkit is never closed (`Toolkit.closeMcpClients()` exists), so any future hook contribution whose registration allocates a resource per toolkit would leak it here;
2. retained tools get registered twice (temp + real), so a tool that performs one-shot binding in `registerTool` would observe two owning toolkits.

Could you add a short comment stating that hook contributions must be plain value objects with no registration-time resource ownership? That turns the current equivalence into a documented contract instead of an accident of the call sites.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -5337,7 +5362,17 @@ public ReActAgent build() {
...
+                registerToolsFromHooks(hookTools);
+                for (String toolName : hookTools.getToolNames()) {
```
oss-maintainer, line 5370: **[Info]** `Toolkit.getToolNames()` returns `new HashSet<>(tools.keySet())`, so the copy order into `agentToolkit` is not deterministic. It does not affect the resulting tool set (names are unique by construction here, and the duplicate collapse already happened inside `hookTools`), but it makes the per-tool `Registered tool ...` INFO log order vary between builds, which is annoying when diffing logs across replicas. Iterating an insertion-ordered view (or sorting the names) would keep the logs stable at no cost.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -5326,6 +5327,30 @@ private void configureSkillBox(Toolkit agentToolkit) {
...
+         * builder are not tested. Retained contributions use normal tool registration semantics,
+         * including replacement of existing tools with the same name. Toolkit execution
+         * configuration, callbacks, tool metadata and Hook instances are preserved. Passing
+         * {@code null} retains normal hook tool registration, as does {@link #build()}.
+         *
+         * @param hookToolFilter optional construction-time filter for hook-contributed tools
+         * @return a new ReActAgent instance
+         * @throws IllegalArgumentException if required parameters are missing or invalid
+         */
+        public ReActAgent build(Predicate<String> hookToolFilter) {
```
oss-maintainer, line 5353: **[Info]** This adds a new public overload on `ReActAgent.Builder`, which is core API and cascades to harness/distribution/extensions. Two API-shape questions while it is still free to change:

- `null` meaning "no filtering" is documented, but `build(null)` at a call site reads as a bug; a builder setter (`hookToolFilter(Predicate<String>)`) or a distinct name (`buildFilteringHookTools(...)`) would be self-describing, and would also let `HarnessAgent` pass the predicate without reaching into `build()` overloads.
- The predicate only sees the tool **name**. The declaration policy it serves (`ToolFilter` + allow list) is also name-based, so this is sufficient today — just confirming that name-only is a deliberate boundary rather than an interim limitation, since widening it later would break the new signature.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
```diff
@@ -588,7 +591,8 @@ static SubagentFactory buildDeclaredFactory(
             }
 
             sub.middlewares(capturedMiddlewares);
-            return sub.build();
+            sub.hooks(capturedHooks);
```
oss-maintainer, line 594: **[Info]** Inheriting the parent's `Hook` **instances** (not copies) into every declared child matches `buildGeneralPurposeFactory`, so this closes #1578 consistently rather than inventing a second mechanism — good.

The consequence is that one hook instance now receives events from the parent and from N children, so any hook that keeps per-agent or per-session mutable state has to be reentrant and must not capture its owning agent. Is that requirement written down anywhere in the `Hook` javadoc? If not, it is worth a sentence there as part of this PR, because #1578 makes shared-instance inheritance the framework norm.

oss-maintainer (commented):
### Summary
Declared subagents now inherit the parent's explicit `Hook` collection, and hook-contributed tools are filtered by the declaration allowlist plus the child's workspace policy **before installation** — implemented by a new `ReActAgent.Builder.build(Predicate<String>)` hook and a package-private `HarnessAgent.Builder.build(List<String>)` variant. Fixes #1578.

**Verdict: COMMENT** — the approach is sound and the test coverage is unusually thorough, but I want two points resolved before this is treated as ready:

1. The `capturedToolsConfig != null` guard in `buildDeclaredFactory` quietly changes **MCP server materialization** for declared subagents that have a `tools:` declaration and their own `workspace/tools.json` (details inline). If that is intended, it belongs in the PR description / release note, because it widens what a declared child can reach externally.
2. The new `build(Predicate<String>)` overload is core API that cascades to harness, distribution and every extension; the name-only, `null`-means-no-op contract should be a deliberate decision (inline).

### Strengths
- Filtering before installation is the right layer: it keeps working when the toolkit forbids runtime deletion (`allowToolDeletion(false)`), which a post-registration `ToolFilter.apply` could not do. The test that asserts a rejected contribution leaves the child's same-name Harness tool intact pins exactly that.
- `filterRunsAfterToolNameResolutionWithoutChangingHookOrderOrDeduplication` is the correct regression to write here: it proves the throwaway-toolkit rewrite preserves hook priority ordering, one-shot `tools()` invocation, and duplicate-name resolution instead of just asserting the happy path.
- Preserving `ToolBase` metadata (readOnly / concurrencySafe), the caller's executor and the chunk callback through the copy is verified by executing the kept tool, not by inspecting fields.
- Making `childToolsConfig` package-private rather than duplicating its strict-allow synthesis keeps one definition of the policy.
- The empty-list case is handled deliberately: an empty declaration still applies the child's workspace policy instead of being read as "allow everything".

### Caveat
Read-only review: nothing was compiled or executed here beyond the CI evidence below, and the 727-line `HarnessAgentTest` addition was reviewed for what it pins, not run.

CI on head `8d5687ba`: ubuntu + windows build, Check License, Check Module Sync and codecov/patch all pass; CLA signed. `MERGEABLE/BLOCKED` with one human `APPROVED` already on record.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2924,6 +2962,16 @@ public HarnessAgent build() {
 
             // ---- Build inner ReActAgent ----
             inner.toolkit(agentToolkit);
+            if (inheritedHookToolAllowlist != null) {
+                ToolsConfig childToolsConfig = resolvedToolsConfig;
+                // Filter before installation, even when the toolkit forbids runtime deletion.
+                // Rejected contributions leave the child's own same-name Harness tools intact.
+                inner.hookToolFilter(
```
oss-maintainer, line 3008: The child's hook-tool policy is now the intersection of two allow policies (the non-empty `decl.getTools()` and the effective `resolvedToolsConfig`), but the parent has no equivalent restriction: `ToolFilter.apply(agentToolkit, ...)` runs before `inner.build()`, so hook-contributed tools are never filtered for the parent agent. Consequence: a hook tool the user configured on the parent silently disappears from declared children unless the same name also appears in `tools:` / workspace `allow`. That is defensible, but it is a real parent/child asymmetry that will read as a silent no-op. Could the `hook()` / `hooks()` Javadoc state explicitly that in declared children hook tools are additionally gated by the declaration allowlist and the workspace policy?

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2924,6 +2962,16 @@ public HarnessAgent build() {
...
+                        toolName ->
+                                (inheritedHookToolAllowlist.isEmpty()
+                                                || inheritedHookToolAllowlist.contains(toolName))
+                                        && ToolFilter.isAllowed(toolName, childToolsConfig));
```
oss-maintainer, line 3012: Nothing is logged when a predicate rejects an inherited hook tool, and there are three candidate reasons (non-empty declaration allowlist, workspace `tools.json` allow/deny, or the config synthesized at line ~2765), so "my declared subagent lost my hook tool" has no evidence trail. `HarnessAgentTest#subagentHooks_*` already pins the matrix — please add one INFO line at build time listing the dropped names and which policy caused the drop. Cheap to add here because the rejected set is exactly `hookTools - allowed`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -2726,6 +2759,11 @@ public HarnessAgent build() {
                     resolvedToolsConfig = ToolsConfigLoader.load(wsManager).orElse(null);
                 }
             }
+            if (inheritedHookToolAllowlist != null) {
+                resolvedToolsConfig =
+                        HarnessAgentBuilderSupport.childToolsConfig(
+                                resolvedToolsConfig, inheritedHookToolAllowlist);
```
oss-maintainer, line 2804: For a child whose parent set no explicit `ToolsConfig`, this is the workspace config from `ToolsConfigLoader.load(wsManager)`, and `childToolsConfig(...)` then yields `strictAllow=true` + `defaultToolsEnabled=false` with `allow = decl.tools filtered by workspace isAllowed` — and `ToolFilter.isAllowed` short-circuits on `deny` first. The precedence (workspace deny > declaration allow > platform exemption) is correct but hard to reconstruct at the call site. One sentence here naming the winner for hook tools would save the next reader a round-trip through three classes.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgentBuilderSupport.java
```diff
@@ -450,6 +450,7 @@ static SubagentFactory buildDeclaredFactory(
         final Toolkit capturedParentToolkit =
                 b.toolkit != null ? b.toolkit.copy() : HarnessAgent.Builder.newDefaultToolkit();
         final Function<String, Model> capturedResolver = b.modelResolver;
+        final List<Hook> capturedHooks = List.copyOf(b.hooks);
```
oss-maintainer, line 455: `capturedHooks` is snapshotted when `buildDeclaredFactory` runs, so hooks added to the parent builder afterwards never reach an existing child factory, and every child reuses the identical shared instances. Both behaviours match the general-purpose path, so this is consistent — just confirming the snapshot point is intentional rather than lazy. If so, the `hook()` Javadoc could add "hooks registered after a declared-subagent entry is built are not inherited", so the exclusion is documented instead of discovered.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -5337,7 +5366,21 @@ public ReActAgent build() {
...
 
-            registerToolsFromHooks(agentToolkit);
+            if (hookToolFilter == null) {
+                registerToolsFromHooks(agentToolkit);
+            } else {
+                // Hook tools are plain value objects with no registration-time resource ownership
+                // or one-shot toolkit binding (see Hook.tools()). Resolving them here acquires no
+                // MCP clients; registering retained values again needs no temporary-toolkit
+                // cleanup.
+                Toolkit hookTools = new Toolkit();
```
oss-maintainer, line 5611: Two consequences of resolving hook tools into a throwaway `Toolkit` and copying them through the metadata-less `registerAgentTool(AgentTool)` overload:

- `Hook.tools()` is now invoked for a second toolkit as well, so a hook that builds resources or mutates state inside `tools()` observes two registration passes. The new `Hook.tools()` Javadoc contracts contributions to plain value objects, which is the right call, but the contract is unenforceable and the harness-provided hooks (`SkillHook`, long-term-memory hooks) are the natural suspects. A regression test that fails when `tools()` is non-idempotent would make the contract real rather than aspirational.
- `new Toolkit()` uses `ToolkitConfig.defaultConfig()`, so when the agent toolkit carries a custom `executorService` or `allowToolDeletion(false)`, anything bound at registration time binds against the throwaway defaults. The inline comment asserts registration acquires no resources; worth naming the executor explicitly in the `Hook.tools()` contract so implementers know not to capture it.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -5337,7 +5366,21 @@ public ReActAgent build() {
...
+                registerToolsFromHooks(hookTools);
+                for (String toolName : hookTools.getToolNames().stream().sorted().toList()) {
+                    if (hookToolFilter.test(toolName)) {
```
oss-maintainer, line 5614: Sorted iteration removes the log-ordering nondeterminism from the previous round — good. One residual: the temporary toolkit keeps only the last contribution for a duplicated name, so the predicate effectively tests "the winner", while a retained tool can still replace a same-name tool already present on `agentToolkit` (skills, memory tools, `skill_manage`). The Javadoc says "including replacement of existing tools with the same name", so the behaviour is documented; my suggestion is only to state the ordering explicitly — hook-contributed tools win over builder-registered same-name tools in the filtered path — since that is the case a user is most likely to trip over.

oss-maintainer (commented):
﻿### Summary
Re-review after the follow-up push (`8d5687ba` → `a6ad007e`). The previous round's blocking concerns are addressed: contributed tools are copied in sorted name order, the new `hookToolFilter` setter replaces the `build(Predicate)` shape, the temporary-toolkit contract is now documented on both `Hook.tools()` and `ReActAgent.Builder`, and the workspace-MCP / declaration-allowlist interaction is pinned by `DeclaredSubagentMcpConfigTest` plus the `subagentHooks_*` matrix. Hook inheritance for declared subagents now matches the general-purpose path, which is the right way to close #1578.

### Findings
No correctness bug found in the new commits. The remaining items are about an undocumented parent/child asymmetry and the silence of policy-driven drops:
- **[Warning]** `HarnessAgent.java:2969` — inherited hook tools are policy-gated in declared children but never in the parent (`ToolFilter.apply` runs before `inner.build()`); the asymmetry should be documented.
- **[Warning]** `HarnessAgent.java:2973` — a dropped hook tool leaves no log evidence of which policy rejected it.
- **[Warning]** `ReActAgent.java:5376` — the throwaway toolkit makes `Hook.tools()` a second-pass call and binds against `ToolkitConfig.defaultConfig()`; tighten the contract wording.
- **[Info]** `HarnessAgent.java:2765`, `HarnessAgentBuilderSupport.java:453`, `ReActAgent.java:5379` — precedence / snapshot / same-name-replacement clarity.

### Verdict
Comment only, not an approval: `build (windows-latest)` is still pending on the new head and `codecov/patch` reports a failure, so coverage on the added filter branches should be confirmed green before this is approved. None of the findings above block a merge once CI is green — they are documentation and observability follow-ups.

---

oss-maintainer (approved):
### Summary
Re-review at `65af113f`. The only change since the previous pass (`a6ad007e`) is a merge of upstream `main`: `Hook.java`, `HookToolsRegistrationTest` and `DeclaredSubagentMcpConfigTest` are untouched by the merge, so the earlier analysis still holds.

CI on this head is green — `build (ubuntu-latest)`, `build (windows-latest)`, `Check License`, `Check Module Sync`, `codecov/patch`, `license/cla` — which was the only reason the previous round stopped short of an approval.

### Findings
No new issues in the merge. Open items from the previous round stay non-blocking follow-ups:
- **[Warning]** `HarnessAgent.java` — inherited hook tools are policy-gated in declared children but not in the parent (`ToolFilter.apply` runs before `inner.build()`); the asymmetry deserves a doc note.
- **[Warning]** `HarnessAgent.java` — a policy-dropped hook tool leaves no log evidence of which policy rejected it; a debug log would help operators.
- **[Info]** `ReActAgent.java` — the throwaway toolkit makes `Hook.tools()` a second-pass call bound to `ToolkitConfig.defaultConfig()`; tighten the contract wording.

### Verdict
Approving: hook inheritance for declared subagents is correct, pinned by tests, and CI + CLA are green on the current head. The items above are documentation/observability follow-ups, not merge blockers.

---*Automated review by github-manager-bot*

oss-maintainer:

This PR currently conflicts with `main` and cannot be merged (`mergeable: CONFLICTING`). Please rebase or merge `main` into `fix/1578-declared-subagent-hooks` and resolve the conflicts:

```bash
git fetch origin
git checkout fix/1578-declared-subagent-hooks
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #3012, opened 2026-09-06
feat(permission): model-requested escalation with approval-gated widening

### LINKED ISSUE #3011: [Feature]: Model-requested permission escalation — sandbox_permissions + justification, approval-gated
## Motivation

Today's permission model is caller-side only: the integrator picks a mode (`DEFAULT` / `ACCEPT_EDITS` / `BYPASS`) once at build time, and the agent lives inside that choice for the whole session. This forces an all-or-nothing trade-off:

- `DEFAULT` (or ASK rules) asks for nearly everything → confirmation fatigue; the practical escape is switching to `BYPASS`, which drops the safety story entirely (see the friction behind #2870-class requests for finer-grained modes).
- `BYPASS` removes the human from the loop with no way back per-call.

What's missing is the third path used by production agent CLIs: **the agent starts locked down and requests exactly the widening it needs, when it needs it, with a human-auditable reason** — while the approval pipeline itself stays fail-closed. The confirm pipeline this builds on is already load-bearing in the field (#2956 shows it in active use under SSE + `ChatUiChannel`, and its fragility matters here too).

## Proposal

Add a model-facing escalation vocabulary shared by the executing tool families (shell first), following the design DeepSeek's open-source harness ships in production (`deepseek-harness`, `packages/sandbox/sandbox/src/escalation.ts`):

1. **Strictly-wider ladder**: `read-only → workspace-write → danger-full-access`. A call may only escalate along the table; nothing escalates down, `read-only` is the floor.
2. **Argument pairing**: tools accept optional `sandbox_permissions` (target mode) and `justification` (non-empty string sentence). Either without the other is a malformed ask, rejected before any approval prompt.
3. **Approval before execution**: a valid request resolves through the existing user-confirmation channel (`RequireUserConfirmEvent`) BEFORE anything executes; deny/timeout/cancel are all fail-closed — no silent grant, no partial execution.
4. **Execution-time validation, never schema-baked**: allowed targets are checked against the call's effective mode at execution time; schemas advertise only the closed target vocabulary (registry-global schemas cannot express per-call truth).
5. **One home**: the ladder, pairing validation, and model-facing denial texts live in a single module so the tool families' approval ordering and error wording cannot drift apart.
6. **Posture preservation**: enabling escalation must not change how any other tool call is evaluated — only calls that actually carry the escalation arguments engage the permission engine; everything else keeps its previous evaluation path.
7. **Model-correctable denials**: rejection reasons (malformed pairing, unknown or non-wider target, feature disabled) reach the model in the DENIED tool result, so it can self-correct instead of blindly retrying.

## Compatibility

Opt-in: a builder flag (e.g. `permissionEscalation(true)`) enables the tool arguments and approval choreography. Default remains exactly today's behavior — no schema, state, or evaluation change.

## Non-goals

- Local OS-level confinement (separate feature; this composes with remote sandboxes and the local shell alike).
- Changing the existing caller-side modes — escalation composes with them.

## Open questions

1. Should the ladder vocabulary reuse `PermissionMode` names or keep the sandbox-specific targets (`read-only` / `workspace-write` / `danger-full-access`) with an internal rank mapping?
2. Tool families beyond shell — filesystem is the natural follow-up; only a family that advertises the arguments in its schema should see the model use them.
3. Should an approved escalation persist for the session or stay per-call? (Current implementation in the accompanying PR: per-call.)

Prior art: DeepSeek Harness's `escalation.ts` (open source) implements exactly this choreography in production; happy to link the file in the design discussion.

birdie7761:
Fixes #3011

## Description

Adds an opt-in mechanism for the model to request a strictly wider permission mode for a single tool call, with the user approving before anything executes — the third path between confirmation fatigue (`DEFAULT` asks for everything) and dropping the safety story (`BYPASS`). Design follows the choreography DeepSeek's open-source harness ships in production (`deepseek-harness`, `escalation.ts`).

**Core pieces:**

- **`PermissionEscalation`** (one home shared by tool families): the closed target vocabulary `read-only` / `workspace-write` / `danger-full-access`, the strictly-wider ladder, argument-pairing validation (`sandbox_permissions` requires a non-empty string `justification`; either alone is a malformed ask; non-string values are rejected rather than coerced; the justification is trimmed and capped at 500 characters with a visible ellipsis), and fail-closed resolution. Validation happens at **execution time** from the raw tool-call input, never baked into a tool schema — schemas are registry-global while the effective mode is per-call truth. A key explicitly mapped to `null` carries no escalation intent.
- **`PermissionEngine`**: escalation requests resolve **after deny rules** (a hard integrator ceiling escalation can never override) and before everything else. A valid strictly-wider request becomes ASK carrying the target and justification, riding the existing user-confirmation loop (`RequireUserConfirmEvent` → confirm → `ALLOWED` → execute). `DONT_ASK` denies: no approver, no escalation. The branch evaluates every call that carries the arguments (not just the shell tool — only shell advertises them today; legacy non-`ToolBase` tools bypass permission evaluation entirely, as before).
- **`ReActAgent`**: denial **reasons** now reach the model — `PermissionVerdict` carries the decision message and the auto-denied gate is a call-id → message map, so `writeAutoDeniedResults` / `runToolBatch` write the engine's (or tool self-check's) own denial text into the DENIED `ToolResultBlock` and its delta event instead of a generic placeholder. A model facing a malformed escalation ask learns exactly what to fix instead of blindly retrying. Note this also improves plain deny-rule and tool-self-check denials — an intentional, deliberate behavior improvement beyond the escalation scope. Calls carrying escalation arguments engage the engine **per call** (`useEngine || hasEscalationArgs(input)`); every other call keeps exactly its previous evaluation path.
- **`PermissionContextState`**: new `escalationEnabled` flag, serialized as `escalation_enabled`, default `false` — pre-feature JSON deserializes unchanged. The flag deliberately does NOT affect `isTrivial()`: enabling escalation must not flip an otherwise default-configured agent from auto-execution to ask-per-call.
- **`HarnessAgent.Builder.permissionEscalation(boolean)`** (default `false`): merges the flag onto the effective permission context (preserving user-supplied modes and rules, call-order independent) and registers `EscalatingShellExecuteTool` in place of the plain shell tool — same tool name, execution body shared via `ShellExecuteTool.executeAndFormat` (the working-directory validation is a security check and must not drift between variants), plus the two optional escalation arguments in the schema.

**Compatibility**: default builds observe no change at all — no schema, state, or evaluation difference; enabling the flag changes only calls that actually carry the escalation arguments.

## Test evidence (TDD, compile-level red on main)

- `PermissionEscalationTest` (14): ladder, pairing, type rejection, capped truncation with marker, null-valued keys, normalization, `DONT_ASK`, ladder-phrased denials
- `PermissionEngineEscalationTest` (8): deny-rule supremacy, malformed pairing, disabled agent, no-args-unchanged evaluation, serialization round-trip + legacy JSON, posture-preserving triviality, `withMode` preservation
- `ReActAgentEscalationHitlTest` (3, end-to-end on a trivial-context agent): plain call auto-executes after the flag is enabled; escalation call pauses → confirms → executes; hallucinated args on a disabled agent deny closed **with the specific reason asserted in the DENIED tool result**; a message-less tool self-check DENY still denies (fail-closed regression)
- `HarnessAgentPermissionEscalationWiringTest` (3): default build unchanged, flag travels onto context + schema, user context preserved on merge

Full core suite **2344/0**, full harness suite **926/0**.

## Checklist

- [x] Related documentation has been updated (builder/class javadocs document the posture-preservation and pairing contracts)


CryoThrust:
I reviewed the proposal and the opt-in posture-preserving design is a good fit. Two details are worth making explicit before merge:\n\n1. Please add a test for approval-to-execution drift: after an approval is recorded, mutate or replay the same toolCallId with different canonical arguments and assert it cannot reuse the approval. The approval must be bound to the exact call snapshot, not only the id.\n2. Please add a concurrent duplicate-resume test asserting that only one decision wins and the tool executes at most once. This is especially important because the approval path is asynchronous and the same callback can be retried by a transport.\n\nThe current tests cover malformed pairing, disabled escalation, deny precedence and ordinary calls, but these two cases protect the runtime security boundary described in the issue. I will avoid a competing implementation and am happy to review the follow-up.

birdie7761:
> I reviewed the proposal and the opt-in posture-preserving design is a good fit. Two details are worth making explicit before merge:\n\n1. Please add a test for approval-to-execution drift: after an approval is recorded, mutate or replay the same toolCallId with different canonical arguments and assert it cannot reuse the approval. The approval must be bound to the exact call snapshot, not only the id.\n2. Please add a concurrent duplicate-resume test asserting that only one decision wins and the tool executes at most once. This is especially important because the approval path is asynchronous and the same callback can be retried by a transport.\n\nThe current tests cover malformed pairing, disabled escalation, deny precedence and ordinary calls, but these two cases protect the runtime security boundary described in the issue. I will avoid a competing implementation and am happy to review the follow-up.

Thank you for the sharp review — both requests are now covered in `ReActAgentApprovalBoundaryTest` (f258df79), plus two extra cases:

1. **Approval binding**: investigation surfaced the actual semantics — `applyConfirmResults` intentionally lets the approver EDIT the confirmed call's arguments before approving (the confirm channel is the trusted human channel), so the approval binds to the id of a currently-ASKING call. What must not happen — and is now pinned by tests — is execution without a fresh approval for that id: a replayed confirm after execution is processed as a normal turn (no re-execution; the idempotency contract is now documented on `Msg.METADATA_CONFIRM_RESULTS` — "no error" ≠ "the confirmation took effect"), and a tool-call id reused by the model with different arguments never inherits the approval. One pre-existing core quirk found along the way: the reused-id call is silently dropped by the id-based tool-result correlation (fail-closed, but the model gets no feedback) — that deserves its own follow-up issue.
2. **Concurrent duplicate resume**: two threads submitting the same confirmation execute the tool exactly once (same-session serialization + no ASKING remains for the duplicate) — test included.
3. Extra: a confirm referencing a non-ASKING id is rejected outright with the "non-ASKING" message, and the rejection is asserted side-effect free — the genuinely ASKING call is untouched and a correct confirm for it still executes afterwards.

Full core suite 2347/0. Would love your review on the follow-up as well.

guslegend0510:
1. **[P1] 提权会跳过工具自身的硬拒绝。**  
   [PermissionEngine.java:169]直接返回 ASK，没有执行工具自检；用户确认后又直接执行。已复现：普通调用返回 `DENY`，加入提权参数并确认后执行成功，**自检调用次数为 0**。应区分可放宽的模式限制与不可覆盖的业务拒绝，保留后者。

2. **[P2] 未开启功能，也会误拒绝普通工具。**  
   [PermissionEscalation.java:69]把任何工具的 `justification` 参数视为提权意图。已用普通 `@Tool` 复现：合法业务参数在改动前成功，PR 中变成“提权未启用”的拒绝。应给支持提权的工具明确标记，避免全局占用普通参数名。

3. **[P2] 复用 Builder 后无法真正关闭提权。**  
   [HarnessAgent.java:2303]只处理开启分支。同一 Builder 先 `true → build()`，再 `false → build()`，第二个 Agent 仍然启用提权，只是 schema 隐藏了参数。每次构建都应重新计算有效权限上下文。

guslegend0510 (changes requested):
需要检查一下这些问题

birdie7761:
> 1. **[P1] 提权会跳过工具自身的硬拒绝。**
>        [PermissionEngine.java:169]直接返回 ASK，没有执行工具自检；用户确认后又直接执行。已复现：普通调用返回 `DENY`，加入提权参数并确认后执行成功，**自检调用次数为 0**。应区分可放宽的模式限制与不可覆盖的业务拒绝，保留后者。
> 
>     2. **[P2] 未开启功能，也会误拒绝普通工具。**
>        [PermissionEscalation.java:69]把任何工具的 `justification` 参数视为提权意图。已用普通 `@Tool` 复现：合法业务参数在改动前成功，PR 中变成“提权未启用”的拒绝。应给支持提权的工具明确标记，避免全局占用普通参数名。
> 
>     3. **[P2] 复用 Builder 后无法真正关闭提权。**
>        [HarnessAgent.java:2303]只处理开启分支。同一 Builder 先 `true → build()`，再 `false → build()`，第二个 Agent 仍然启用提权，只是 schema 隐藏了参数。每次构建都应重新计算有效权限上下文。

感谢 @guslegend0510 的细致评审，所有意见均已验证成立并修复完毕，最新提交 `7001bb52`（core 全套 2351/0，harness 全套 927/0）。按轮次汇总如下：

**一、CHANGES_REQUESTED 三条（全部复现、全部修复）**

1. **P1（提权跳过工具自检硬拒绝）**：确认属实。修复：escalation 的 ASK 分支现在会先执行工具自身的 `checkPermissions`——工具的业务性 DENY 永远压过提权（不可放宽），而模式类限制照旧跳过（那正是请求用户批准要解除的东西，两类拒绝语义已分开）。回归测试：自检返回业务 DENY 时，带提权参数的调用最终仍为 DENIED 且携带工具自身的理由。
2. **P2（未开启功能时 `justification` 业务参数被误拒）**：确认是真实回归。修复：**提权意图按 schema 声明限定**——引擎与 ReActAgent 的按调用门控都只对参数 schema 中声明了 `sandbox_permissions` 的工具生效，普通业务工具自己的 `justification` 参数在开关两种状态下均保持原有求值路径（两种状态均已测试）。这也让 issue 中 open question 2 的约定从“惯例”升格为“结构强制”。
3. **P3（Builder 复用无法关闭提权）**：确认状态泄漏。修复：有效权限上下文改为**每次构建都从用户 context 重新计算**，`permissionEscalation(true) → build() → false → build()` 得到完全关闭的 agent（标志为 false 且 schema 无提权参数），已加测试。

**二、后续评审轮的加固（均来自对修复本身的复查）**

- **空流 fail-open**：escalation ASK 分支的自检调用补充了 `.switchIfEmpty` 兜底——空流自检的调用现在回落到提权 ASK（fail-closed），与常规路径同一防御姿态，不会再静默漏出确认门直接执行；
- **500 字符截断升级为真边界保护**：原实现只约束了日志文案，审批界面看到的仍是未截断原文；现在通过 `updatedInput`（此前全仓无消费者的字段）重写进入确认事件的 pending 块，`RequireUserConfirmEvent` 携带的即为带省略号的截断文本（e2e 断言打在事件面）；
- **`ToolUseBlock.withInput` 字段保真**：改用与 `withState` 相同的全参构造器，`content` 与 `metadata`（如 Gemini 的 thought signature）不再丢失，javadoc 同步归位；
- **单一家园原则**：schema 嗅探逻辑下沉为 `PermissionEscalation.advertisesEscalation` 一处（引擎与门控共同委托），签名去除了无用参数。

此外按 CryoThrust 的评审补齐了批准边界测试（重放确认不复用、id 复用不继承批准、并发重复恢复至多执行一次、拒绝无副作用），并在 `Msg.METADATA_CONFIRM_RESULTS` 的 javadoc 中写明了幂等契约。欢迎继续复审。

guslegend0510 (approved):
LGTM

agentscope-core/src/main/java/io/agentscope/core/permission/PermissionEscalation.java
```diff
@@ -0,0 +1,242 @@
...
+        return Outcome.ask(target, reason);
+    }
+
+    /**
+     * Ladder rank of an escalation target, {@code null} when unknown. Ranks are ordered
+     * floor-to-ceiling and comparable with {@link #rankOfMode}.
+     */
+    private static Integer rankOfTarget(String target) {
+        return switch (target) {
+            case TARGET_READ_ONLY -> 0;
```
oss-maintainer, line 212: **[Warning]** `read-only` (rank 0) can never be a *strictly wider* target: `rankOfMode` maps the lowest mode (`EXPLORE`) to 0 as well, so `targetRank <= modeRank` rejects it under every mode. Yet both the `@Tool` description and the `sandbox_permissions` param doc advertise `read-only` as a valid value, so a model that honestly asks to narrow a call gets a hard DENY whose text ("it is not strictly wider than the current mode") reads like a formatting error — one wasted round trip every time it tries.

Suggest either dropping `read-only` from the advertised vocabulary, or mapping a narrower/no-op request to `NOT_PRESENT` (strip the args, evaluate normally) instead of `DENY`. If the rejection is deliberate, a note on `rankOfTarget` that rank 0 is intentionally unreachable would save the next reader this derivation.

agentscope-core/src/main/java/io/agentscope/core/permission/PermissionEscalation.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+        String reason = ((String) justification).trim();
+        if (reason.isEmpty()) {
+            return Outcome.deny(
+                    "Invalid justification: expected a non-empty sentence explaining why this "
+                            + "call needs wider permissions.");
+        }
+        if (reason.length() > MAX_JUSTIFICATION_LENGTH) {
+            // Keep the approved prompt honest: the approver must see that the text was cut.
+            reason = reason.substring(0, MAX_JUSTIFICATION_LENGTH - 1) + "…";
```
oss-maintainer: **[Warning]** This cap counts UTF-16 code units, so `substring(0, MAX_JUSTIFICATION_LENGTH - 1)` can split a surrogate pair (emoji, non-BMP CJK) and leave a lone high surrogate in front of the `…`. That string then reaches the approver's prompt *and* `cappedInput`, i.e. it is persisted on the pending `ToolUseBlock` and re-serialized on resume, where a lone surrogate tends to surface as a replacement character or as a JSON encode failure depending on mapper configuration.

Code-point aware cut:
```java
int cut = Character.isHighSurrogate(reason.charAt(MAX_JUSTIFICATION_LENGTH - 1))
        ? MAX_JUSTIFICATION_LENGTH - 2
        : MAX_JUSTIFICATION_LENGTH - 1;
reason = reason.substring(0, cut) + "\u2026";
```
The 500-char ceiling itself is the right call — it just needs to land on a character boundary.

agentscope-core/src/main/java/io/agentscope/core/permission/PermissionEngine.java
```diff
@@ -140,19 +149,87 @@ public Mono<PermissionDecision> checkPermission(ToolBase tool, Map<String, Objec
...
+                // unapproved).
+                return tool.checkPermissions(input, context)
+                        .flatMap(
+                                toolDecision -> {
+                                    if (toolDecision != null
+                                            && toolDecision.getBehavior()
+                                                    == PermissionBehavior.DENY) {
+                                        return Mono.just(toolDecision);
+                                    }
+                                    return Mono.just(escalationAsk);
```
oss-maintainer, line 220: **[Warning]** Absorbing the tool self-check into the escalation ASK keeps only its DENY, which silently discards two things the normal path preserves:

1. `updatedInput` — `cappedInput` is rebuilt from the raw `input`, so a self-check that sanitised arguments reaches neither the approver nor the executed call.
2. `suggestedRules` — the ordinary ASK path enriches via `tool.generateSuggestions(input)` so the confirmation UI can offer "always allow"; an escalation ASK carries none, so approving one can never become a durable rule through the same flow.

If the absorption is intentional (the comment says it is), consider carrying `toolDecision.getUpdatedInput()` forward when non-null, and appending the tool's own message when its self-check returned ASK rather than PASSTHROUGH/ALLOW — today the approver sees only the model's justification for a call the tool *also* flagged.

agentscope-core/src/main/java/io/agentscope/core/permission/PermissionEscalation.java
```diff
@@ -0,0 +1,242 @@
...
+    /** Upper bound on the justification text carried into the approval prompt. */
+    public static final int MAX_JUSTIFICATION_LENGTH = 500;
+
+    /**
+     * Whether a tool's parameter schema advertises the escalation argument. Escalation intent
+     * is scoped to advertising tools so that ordinary business tools with (for example) a
+     * legitimate {@code justification} parameter keep their previous evaluation untouched —
+     * the single home for this sniff so engine and agent-side gating cannot drift apart.
+     */
+    public static boolean advertisesEscalation(io.agentscope.core.tool.ToolBase tool) {
```
oss-maintainer, line 69: **[Info]** Escalation handling is keyed on the *argument name* `sandbox_permissions` appearing in the advertised schema. `justification` gets an explicit exemption in the comments ("that tool's own business argument"), but `sandbox_permissions` does not — an integrator tool that legitimately takes a parameter by that name would silently gain escalation semantics, and with `escalationEnabled=false`, a hard DENY from `resolve` in place of its own validation.

A marker would be sturdier than name sniffing: an `EscalationAware` interface or a `@Tool(escalation = true)` flag implemented by `EscalatingShellExecuteTool`, so the behaviour is opted into by the tool. Cheap now, awkward to change once third-party tools start advertising the same argument name.

oss-maintainer (commented):
### Summary
Model-requested permission escalation for the shell tool family: a closed strictly-wider ladder (`sandbox_permissions` + `justification`) resolved in `PermissionEngine` between deny rules and ask rules, valid requests routed through the existing HITL confirmation flow, and every malformed / non-wider / no-approver case failing closed. Plus a real usability win on the rejection path — `PermissionGate.autoDenied` became id → message, so a DENIED tool result now tells the model *why* instead of "Permission denied by rules".

The design reads sound and the intent comments in `PermissionEngine` pre-empt most reviewer questions. Two things I specifically checked and liked: `isTrivial()` deliberately ignoring the flag keeps every non-escalating call on its previous evaluation path byte-for-byte, and recomputing the merged context on every `build()` closes the builder-reuse leak (`true → build → false → build` yields escalation OFF, not a hidden engine-on/schema-off split).

Findings are refinements, not blockers: the unreachable `read-only` rung is still advertised to the model, the justification cap can cut a surrogate pair, and an escalation ASK discards the tool self-check's own `updatedInput` / `suggestedRules`.

### Checked
- `ToolBase.checkPermissions` returns `Mono.just(passthrough(name))` — never empty — so the `switchIfEmpty` fallback is belt-and-braces rather than load-bearing.
- The widened mode is never persisted (`withMode` is untouched by the approval path), so an approved escalation cannot leak into later calls of the session.
- `AgentState` replaces a null permission context with `PermissionContextState.builder().build()`, so `inner.permissionContext(null)` in the new `else` branch is safe and preserves the pre-PR path.
- Prior P1 from review (escalation overriding a tool's hard business DENY) is fixed: the self-check DENY short-circuits ahead of `escalationAsk`.

---

agentscope-core/src/main/java/io/agentscope/core/permission/PermissionEngine.java
birdie7761: Intentional, documented in-line at this branch: the escalation cappedInput must own updatedInput — it is the "what the approver actually sees" contract, and the 500-char cap has to bind the persisted pending ToolUseBlock, not just log text; merging a second input rewrite on top is ill-defined. The self-check's ASK is absorbed into the escalation ASK because both paths end at the same human approver. suggestedRules has no consumer in the confirmation flow — adding a path for it here would be dead weight. Only the self-check's DENY is preserved, which is the security-relevant half.

agentscope-core/src/main/java/io/agentscope/core/permission/PermissionEscalation.java
birdie7761: Agreed on the direction — but a marker has to cover both engagement paths, not just the engine: ReActAgent.evaluateOne also engages the engine via PermissionEscalation.hasEscalationArgs(input) for trivial-context agents, so a name-sniffing→marker switch is an API-shape decision (interface vs @Tool flag) touching core and harness. Filed as a follow-up rather than a post-approval bolt-on; today the only tool advertising sandbox_permissions is the escalation shell itself, and the justification business-argument exemption is already documented at the engine site.

oss-maintainer (approved):
### Summary
Re-reviewed the new head `db34de31`. The single commit since my last pass is exactly the fix I asked for, and it is the right shape: `read-only` is kept in the vocabulary for validation completeness but is no longer named in the unknown-target denial, so the model is not taught to retry with the floor. The floor now gets its own ladder-specific reason, which also fixes the case where the generic "not strictly wider" text read like a format error when coming from `EXPLORE`.

All three findings from my previous round are resolved on this head:
- `PermissionGate.autoDenied` is now `id -> message`, so a DENIED tool result carries the engine's / tool self-check's own text instead of a placeholder, and the `ToolResultStartEvent` path uses the same map with a `getOrDefault` fallback.
- The ASK branch routes `v.effectiveUse()` (justification-capped input) into the gate, so the approver sees exactly what will execute and the tool self-check's `updatedInput` is no longer discarded.
- The unreachable `read-only` rung is no longer advertised in model-facing denial text.

Verified: `LinkedHashMap` keeps denial ordering stable for batched tool calls, the `getOrDefault("Permission denied")` fallback covers an id present in the event stream but absent from the map, `Character.isHighSurrogate` still guards the cap against splitting a surrogate pair, and `ToolUseBlock.withInput` preserves `content` / `metadata` / `state` as documented. `effectiveUse()` is the single rewrite point (`updatedInput == null ? use : use.withInput(updatedInput)`), so the ASK path can no longer lose a self-check's rewritten arguments. 1485 added lines across five new test classes cover the approval boundary, the HITL escalation path, the engine resolution and the wiring. CLA signed, CI green on ubuntu and windows, mergeable.

LGTM — approving.

---

oss-maintainer:

```bash
git fetch upstream
git checkout feat/permission-escalation
git rebase upstream/main
# resolve conflicts, then:
git push --force-with-lease
```

---

birdie7761:
> ⚠️ **Merge conflict detected**
> 
> This PR currently conflicts with `main` and cannot be merged. Please rebase (or merge `main` in) and resolve the conflicts:
> 
> ```shell
> git fetch upstream
> git checkout feat/permission-escalation
> git rebase upstream/main
> # resolve conflicts, then:
> git push --force-with-lease
> ```
> 
> This is a one-time reminder. Once the conflicts are resolved, comment `@oss-maintainer please re-review` and I will pick it up again.
> 
> _Automated notification by github-manager-bot_

@oss-maintainer  please re-review

oss-maintainer (commented):
### Summary
Re-review after your conflict fix (`@oss-maintainer please re-review`). The merge resolution itself is correct — one blocker remains, and it is CI, not the code.

**Merge resolution verified against `main@70686e4a`:**
- `ReActAgent.java` still carries the whole `#3104` auto-deny lifecycle on your head: `gate.autoDeniedIds()`, `writeAutoDeniedResults(...)` (both call site and definition), `deniedToolResultEvents(...)` and the `PERMISSION_DENIED_BY_RULES` string — 11 `autoDenied` references versus 10 on main, i.e. your additions on top of upstream, nothing reverted.
- No conflict markers left in `ReActAgent.java` or `HarnessAgent.java`; the only lines that differ from main are your own escalation logic.
- All 4 earlier review threads are resolved, and this head is the only change since my previous approval — a merge of `origin/main`, no new feature code.

**Blocker:** `build (windows-latest)` fails on `ff237eda` (https://github.com/agentscope-ai/agentscope-java/actions/runs/35216699925/job/105186766440) while ubuntu, License, Module Sync and codecov pass. Since this PR adds a permission-gated path that is exercised by the platform-independent test suite, a windows-only failure is most likely environment or timing, but it needs to be green before this can merge — please re-run the failed job (or push the fix if it is real). Once windows passes I will convert my earlier approval to this head without another full review.

---

birdie7761:
@oss-maintainer please re-review

oss-maintainer (approved):
### Summary
Re-review at `972f07fb` — approving. The only thing that changed since my previous review (`ff237eda`) is a merge of `main`, and it is resolved correctly: 13 of the 14 files in this PR are byte-identical to the reviewed head, and the one that differs, `HarnessAgent.java`, picks up upstream's `WorkspacePathNormalizer.of(path, nsFactory)` signature change while keeping the full escalation wiring (`permissionEscalation` builder flag, `withEscalationEnabled(true)` merge on the effective context, the `else` branch that restores the user's own context verbatim so builder reuse can't leak the flag, and the `EscalatingShellExecuteTool` / `ShellExecuteTool` switch at registration). No conflict markers anywhere in the touched files.

The blocker I raised last round is gone: `build (windows-latest)` is now green on this head, alongside `build (ubuntu-latest)`, `Check License`, `Check Module Sync` and `codecov/patch`. CLA is signed, the PR is `MERGEABLE`/`clean` against `main`, and all 4 earlier review threads are resolved.

Nothing new to ask for. Behaviour remains gated exactly as reviewed: default builds observe no schema, state or evaluation change, and only calls that actually carry `sandbox_permissions` + `justification` engage the engine — with deny rules still winning over any escalation request. Ready for a maintainer to merge.

---

PR #3015, opened 2026-09-07
fix(core): prevent shutdown state saver retention for transient agents

### LINKED ISSUE #3014: [Bug]: GracefulShutdownManager::bindStateSaver导致内存溢出
原因：
HarnessAgent底层使用ReActAgent，HarnessAgent.Builder.agentId("default"),不能生效于底层。导致每次构建请求都会新增一条数据到GracefulShutdownManager::stateSavers。触发OOM。

Image

Image

Image

guslegend0510:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Closes #3014.

Repeatedly constructing stateful HarnessAgent/ReActAgent instances without calling close() leaves agents and their state stores retained by the global shutdown callback registry.

Use weak owner references and a reference queue to remove collected registrations. The ReActAgent saver no longer strongly captures the agent. Active requests still retain the agent and saver, preserving per-session shutdown persistence and CAS version updates. Explicit close() still unbinds immediately, and per-instance runtime IDs remain distinct from the Harness namespace ID.

Regression tests cover abandoned agents and stores, stale registration cleanup, active-request retention, custom callback lifetime, and same-namespace HarnessAgent isolation.

## Validation

- Two retention regressions fail on the original main and pass with this fix.
- 77 focused tests passed.
- `mvn -pl agentscope-harness -am spotless:check test`: Core 2,322 tests, Harness 957 tests; zero failures/errors, 14 skipped.
- `git diff --check` passed.

## Checklist

- [x] Code formatted with Spotless
- [x] Full Core and Harness test suites passed (other modules not run)
- [x] Registry callback lifecycle documented in Javadoc
- [x] Code ready for review


oss-maintainer (commented):
### Summary
Fixes a real retention bug: the shutdown saver lambda registered by `ReActAgent` captured `this` strongly, so `GracefulShutdownManager.stateSavers` kept every transient agent (and its state store) alive for the JVM's lifetime. The change is well structured — a `WeakReference<ReActAgent>` created before the lambda, the persistence logic extracted into `saveShutdownState`, and a `WeakReference`-based registration in the manager that is expunged via a shared `ReferenceQueue`. The two new retention tests assert the actual collection outcome rather than just map size, which is the right level of rigour for this class of bug.

### Findings
- **[Warning]** `GracefulShutdownManager.java:145` — reclamation is still driven only by `bind` / `unbind` / `registerRequest`, so the map keys and saver lambdas can accumulate when the manager stops being touched after agents die.
- **[Info]** `ReActAgent.java:361` — once the agent is collectable, a shutdown save for a still-live session can be skipped silently; a debug log would make that diagnosable.
- **[Info]** `GracefulShutdownManager.java:138` — the identity-based two-arg `remove(key, value)` deserves a comment so it is not "simplified" later.

The first item is about how much of the retention gap this closes, not about correctness of what is implemented — no functional objection.

---

agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownManager.java
```diff
@@ -128,9 +138,30 @@ public void unbindStateSaver(Agent agent) {
         if (agent == null) {
             return;
         }
+        removeCollectedStateSavers();
         stateSavers.remove(agent.getAgentId());
     }
 
+    private void removeCollectedStateSavers() {
```
oss-maintainer, line 145: The weak-registration design is the right fix for the leak this PR targets (an agent pinned by the global manager can no longer pin its state store), but the *map itself* is still only ever pruned from inside `bindStateSaver` / `unbindStateSaver` / `registerRequest`. Those are all agent-lifecycle paths, so in a deployment that creates transient agents and then stops touching the manager, `stateSavers` keeps growing with entries whose `WeakReference` has already been enqueued: the keys (`String agentId`) and the `ShutdownStateSaver` lambdas stay strongly reachable forever.

Not a regression relative to today's code, but the PR claims to remove retention, so it is worth closing the last part of the gap. Cheap options: drain the queue from the place that actually needs the memory bound (e.g. when `stateSavers.size()` crosses a threshold), or run `removeCollectedStateSavers()` from the existing shutdown monitor / a single daemon tick so reclamation stops being coincidental with register/bind.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -357,39 +358,43 @@ private ReActAgent(Builder builder, Toolkit agentToolkit) {
         this.hookDispatcher = new LegacyHookDispatcher(this);
 
         if (this.stateStore != null) {
+            // The global manager must not keep an otherwise unreachable agent alive through
```
oss-maintainer, line 362: Correct treatment of the `this`-escape: `new WeakReference<>(this)` before building the lambda, with the null-check falling through to "agent is gone, nothing to persist". The extraction into `saveShutdownState(agentState)` keeps the CAS/slot-version logic unchanged and readable, and it makes the lambda small enough that reviewers can see there is no strong `this` capture left.

One thing to double-check: if the agent is collected but its session is still mid-request during shutdown, the state is now silently *not* persisted (previously it was, because the agent was pinned). The javadoc added on `bindStateSaver` says the saver "must not strongly capture the agent if collection ... is desired", which implies the trade-off is intentional — but a `log.debug` in the `agent == null` branch would make a missing shutdown save diagnosable instead of invisible.

agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownManager.java
```diff
@@ -128,9 +138,30 @@ public void unbindStateSaver(Agent agent) {
         if (agent == null) {
```
oss-maintainer, line 138: `stateSavers.remove(registration.agentId, registration)` is the right guard — a stale queued registration from a replaced bind must not evict the new one, and `remove(key, value)` on `ConcurrentHashMap` uses identity equality, which is exactly what is wanted here since `StateSaverRegistration` does not override `equals`. Worth a one-line comment saying that the two-arg form is deliberately identity-based, so nobody "optimises" it into `remove(key)` later.

agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownManager.java
```diff
@@ -107,12 +110,19 @@ public Mono<Void> getShutdownTimeoutSignal() {
...
+     * agent if collection without explicit {@link #unbindStateSaver(Agent)} is desired.
      */
     public void bindStateSaver(Agent agent, ShutdownStateSaver saver) {
         if (agent == null || saver == null) {
             return;
         }
-        stateSavers.put(agent.getAgentId(), saver);
+        removeCollectedStateSavers();
+        stateSavers.put(
+                agent.getAgentId(), new StateSaverRegistration(agent, saver, collectedAgents));
```
oss-maintainer, line 125: Cleanup of dead registrations (`removeCollectedStateSavers()`) is only triggered by `bindStateSaver`, `unbindStateSaver`, and `registerRequest`. The shutdown iteration path over `stateSavers` (not shown in this diff) should also call `removeCollectedStateSavers()` before processing savers. Without it, a process that creates and abandons many agents then sits idle will carry dead map entries into shutdown. The saver lambda's null-check on `agentReference.get()` prevents functional errors, but invoking the cleanup first is consistent with the other access paths and avoids processing known-dead entries during the time-sensitive shutdown window.

oss-maintainer (commented):
### Summary
Re-review of PR #3015 at its current head (automated review of the changes in this diff). 2 item(s) raised below; nothing here blocks the approach itself — details inline.

### Findings
- **[Warning]** `agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownManager.java:125` — Cleanup of dead registrations (`removeCollectedStateSavers()`) is only triggered by `bindStateSaver`, `unbindStateSaver`, and `registerRequest`. The shutdown iteration path over `stateSavers` (not shown in this diff) should also call `removeCollectedStateSavers()` before processing savers. Without it, a process that creates and abandons many agents then sits idle will carry dead map entries into shutdown. The saver lambda's null-check on `agentReference.get()` prevents functional errors, but invoking the cleanup first is consistent with the other access paths and avoids processing known-dead entries during the time-sensitive shutdown window.
- **[Info]** `agentscope-core/src/test/java/io/agentscope/core/shutdown/ShutdownStateSaverRetentionTest.java:68` — The discarded-agent test relies on `System.gc()` actually collecting 16 unreachable agents and their state stores within a 5-second deadline. On memory-pressure CI runners or JVMs with non-cooperative GC (e.g., ZGC/Shenandoah with different `System.gc()` semantics), this can become flaky. Consider adding `-XX:+ExplicitGCInvokesConcurrent` or a JVM-arg comment documenting the expectation, and verify the test passes under `-XX:+UseZGC` if that is a supported runtime configuration.

### Notes

---

agentscope-core/src/test/java/io/agentscope/core/shutdown/ShutdownStateSaverRetentionTest.java
```diff
@@ -0,0 +1,188 @@
...
+            for (int i = 0; i < 16; i++) {
+                // Intentionally omit close: a forgotten lifecycle call must not leave the
+                // entire agent graph rooted in the process-wide shutdown manager.
+                createDiscardedAgent(agents, stores);
+            }
+            awaitCollected(agents);
+            awaitCollected(stores);
+            // Normal registry traffic must also reclaim the dead UUID/registration entries.
+            try (ReActAgent live = newAgent(new InMemoryAgentStateStore())) {
+                long deadline = System.nanoTime() + Duration.ofSeconds(5).toNanos();
```
oss-maintainer, line 68: The discarded-agent test relies on `System.gc()` actually collecting 16 unreachable agents and their state stores within a 5-second deadline. On memory-pressure CI runners or JVMs with non-cooperative GC (e.g., ZGC/Shenandoah with different `System.gc()` semantics), this can become flaky. Consider adding `-XX:+ExplicitGCInvokesConcurrent` or a JVM-arg comment documenting the expectation, and verify the test passes under `-XX:+UseZGC` if that is a supported runtime configuration.

agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownManager.java
```diff
@@ -107,12 +110,19 @@ public Mono<Void> getShutdownTimeoutSignal() {
      *
      * <p>The saver is invoked during shutdown to persist the agent's {@link AgentState}
      * (with {@code shutdownInterrupted} set to {@code true}).
+     *
+     * <p>The registration weakly references its agent and is removed on subsequent registry
+     * access after the agent is collected. The saver itself is retained strongly so inline
+     * callbacks remain available while the agent is alive; it must not strongly capture the
+     * agent if collection without explicit {@link #unbindStateSaver(Agent)} is desired.
```
oss-maintainer, line 117: **[Warning]** The contract is now silently violable by callers. `this.saver` is held strongly by the registration, so any `ShutdownStateSaver` that captures its agent — which is the natural thing to write, and what `ReActAgent` itself did until this commit — keeps the agent alive forever and the weak reference never enqueues. `ReActAgent`'s own lambda is now weak-only, but `bindStateSaver` is public. Two options that cost little: (a) wrap the incoming saver so the agent is reached through the registration's own `WeakReference` (single place where the capture rule is enforced), or (b) at minimum detect the capture at bind time and `log.warn` once, so an extension module that reintroduces a strong capture is visible instead of turning into a leak report three months later.

agentscope-core/src/main/java/io/agentscope/core/shutdown/GracefulShutdownManager.java
oss-maintainer, line 145: **[Info]** Reclamation is driven entirely by registry traffic: `removeCollectedStateSavers()` is only reached from `bindStateSaver`, `unbindStateSaver`, `registerRequest` and `resetForTesting`. In the exact scenario this change targets — a transient agent created, used once and dropped, with nothing else happening afterwards — its entry stays in the map (with its strong `saver`) until some other agent touches the registry, and on a quiescent JVM that can be until shutdown. Also worth noting that a registration whose referent is already cleared is still returned by `stateSavers.get(...)` at line 218 (only the queue drain removes it), so `registerRequest` can hand out a `saver` for a half-dead agent. Draining at the top of `saveActiveRequestStates`/the shutdown path (or simply skipping a registration where `get() == null` at lookup) would make the two paths agree.

oss-maintainer (commented):
### Summary
Re-review at `546dac62`; the new author commit `bea9b484` replaces the strong `agentId -> ShutdownStateSaver` map with a `WeakReference<Agent>` registration on a `ReferenceQueue`, drains it on registry access, and extracts `ReActAgent#saveShutdownState` so the bound lambda captures only a weak reference. The leak being fixed is real, the two-arg `remove(agentId, registration)` correctly avoids evicting a replacement registration, and 348 lines of retention tests across core and harness show the author verified both directions (collected agent disappears from the registry; in-flight agent does not). Commenting rather than approving because of the strong-`saver` contract below and the GC-hint dependence in the new tests.

### Findings
- **[Warning]** GracefulShutdownManager.java:117 — the registration keeps `saver` strongly, so the "must not capture the agent" rule is now an unenforced requirement on every public implementor; wrap or detect it
- **[Info]** GracefulShutdownManager.java:145 — reclamation only happens on registry traffic, and `stateSavers.get()` at :218 can still return a registration whose referent is gone
- **[Warning]** ShutdownStateSaverRetentionTest.java:180 — `System.gc()` is a hint (`-XX:+DisableExplicitGC`), so the failure message would misattribute a JVM-config difference to a registry leak
- **[Info]** ReActAgent.java:364 — callback is published from the constructor; safe today, worth a comment

### Suggestions
The cleanest single change for the first point: let the registration own the weak reference and pass the agent to the saver (or wrap the saver internally), so callers cannot accidentally re-introduce the capture this commit removes.

### Verification
CLA signed on `546dac62`, CI green (`validate`, `build` ubuntu/windows, `Check License`, `Check Module Sync`, `codecov/patch`), `MERGEABLE/BLOCKED` only because a review is still required. I did not execute the new tests locally; the CI run is the evidence.

---

agentscope-core/src/test/java/io/agentscope/core/shutdown/ShutdownStateSaverRetentionTest.java
```diff
@@ -0,0 +1,188 @@
...
+        Field field = GracefulShutdownManager.class.getDeclaredField("stateSavers");
+        field.setAccessible(true);
+        return ((Map<?, ?>) field.get(manager)).size();
+    }
+
+    private static void awaitCollected(List<? extends WeakReference<?>> references)
+            throws Exception {
+        long deadline = System.nanoTime() + Duration.ofSeconds(5).toNanos();
+        do {
+            System.gc();
```
oss-maintainer, line 180: **[Warning]** GC-timing assertion risk. `awaitCollected` treats `System.gc()` as a command, but it is only a hint: under `-XX:+DisableExplicitGC` (used in some CI and production JVM profiles) the loop only exits via the 5 s deadline and then `assertTrue(...refersTo(null))` fails with a message that blames the registry ("must not retain discarded agents") for what is actually a GC configuration. `assertNotNull(reference.get())` at line 92/120 has the mirror-image risk — a strong local kept live by the interpreter frame. Cheap hardening: create allocation pressure in the loop (a throwaway `byte[]` churn) alongside the `System.gc()` hint, and phrase the failure message to mention that explicit GC may be disabled so a future flake report is diagnosable. Given this repo has already been chasing flaky shutdown/memory tests, worth settling now.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -358,39 +359,43 @@ private ReActAgent(Builder builder, Toolkit agentToolkit) {
...
+            // this callback. ActiveRequestContext holds the agent strongly while a call runs.
+            WeakReference<ReActAgent> agentReference = new WeakReference<>(this);
```
oss-maintainer, line 364: **[Info]** The registration is published from inside the constructor, so a shutdown arriving on another thread can invoke `saveShutdownState` before the constructor returns. The fields the extracted method touches (`stateStore`, `slotVersions`, `stateConflictCount`) are assigned earlier, so this is fine today — flagging it only so that a future field added to `saveShutdownState` is not initialised after this `bindStateSaver` call. A one-line comment next to the bind would be enough.

PR #3019, opened 2026-09-07
perf(harness): batch heartbeat task-record persistence per session store

### LINKED ISSUE #3018: [Enhancement]: heartbeat rewrites the whole task-record store once per running task every 30s
## Description

`WorkspaceTaskRepository`'s heartbeat refreshes `lastUpdatedAt` for every live local task every 30 seconds. Each refresh goes through `updateStatus` → `WorkspaceManager.writeTaskRecord`, which performs a **full read-modify-write of the session's task-record store** (`agents/<agentId>/tasks/<sessionId>.json`):

1. `updateStatus` reads the whole store to evaluate its terminal-state guards (`readTaskRecord`),
2. `writeTaskRecord` reads it **again** inside the per-path lock,
3. then serializes and rewrites the **whole** map, pretty-printed.

So with `R` running tasks sharing one store file that holds `T` accumulated records (finished tasks keep their ~KB-sized results in the same file), every heartbeat cycle costs `R × (2 parses + 1 serialization + 1 whole-file write)` — `O(R × T)` work and bytes every 30 seconds, executed on the `ws-task-maint-*` scheduler thread while holding the per-path `ReentrantLock` for each pass (a task completing concurrently queues behind each full cycle).

## Measured impact

Standalone replica of the store mechanics (JDK 17, Jackson 2.21.1 as in the project BOM, `TaskRecord` mirrored field-for-field, finished records carrying ~2 KB result text; one "cycle" = the heartbeat's R sequential refreshes; median of 5 runs):

| Store (total / running tasks) | Cycle time | Bytes written per cycle | Per-update lock hold |
|---|---|---|---|
| 20 / 5 | ~5 ms | 283 KB | ~1 ms |
| 100 / 25 | ~76 ms | 7.1 MB | ~4 ms |
| 400 / 100 | ~1.1 s | **114 MB** | up to ~25 ms |

At the stress point this is ≈3.8 MB/s of sustained rewrite traffic (≈328 GB/day while tasks are in flight) for bookkeeping that conceptually appends 100 timestamps.

## What already works (verified on main)

To scope this correctly: the store's durability and robustness basics are already in place — `writeLocalFile` writes via temp file + `ATOMIC_MOVE`, and `readTaskMapLocked` isolates per-file corruption (an unreadable store logs an error and yields an empty view without breaking the sweeper's listing). This issue is purely about the **write amplification**, not about correctness.

## Proposed direction

Batch the heartbeat: all refreshes that resolve to the **same store file** can share one lock acquisition, one parse, one merge and one whole-file write — a format-free change that turns the per-cycle cost from `O(R × T)` into `O(T)`. Grouping must be by resolved store path (per-task `RuntimeContext` instances can differ while mapping to the same file). The per-record guards (terminal immutability, cancellation blocking) are preserved and become evaluated against the same in-lock snapshot that is written back.

With batching, the measured cycle at 400/100 drops to ~7.5 ms and ~1.1 MB (one rewrite), and 100/25 to ~2.9 ms / 285 KB.

A more invasive alternative — an append-only journal per store with fold-on-read — would additionally reduce bytes ~5× further and make single-record events (task completion) O(1), but given the numbers above it does not look necessary once batching is in; mentioning it only for completeness.

## Environment

- AgentScope-Java version: 2.0.3-SNAPSHOT (main, 1048fdc3)
- Module: agentscope-harness (`WorkspaceTaskRepository`, `WorkspaceManager`)

birdie7761:
Fixes #3018

## Description

Every 30-second heartbeat cycle refreshed each running local task with its own full read-modify-write of the session's task-record store: with `R` running tasks sharing `agents/<id>/tasks/<sid>.json`, one cycle performed `R` parses, `R` pretty-print serializations and `R` whole-file writes of the same file — each update even parses the store twice (`updateStatus`'s terminal-state guard read, then `writeTaskRecord`'s in-lock read). On the per-path `ReentrantLock` this is `O(R x store)` work and bytes per cycle, and it grows with both the running count and the accumulated history of finished tasks in the same session.

The heartbeat now groups the refreshes that resolve to the same store file and persists them in a single read-modify-write:

- `WorkspaceManager.updateTaskRecordStatuses(rc, agentId, sessionId, statusByTaskId)` — one lock acquisition, one parse, one whole-file persist regardless of how many records are refreshed. The per-record guards mirror `WorkspaceTaskRepository.updateStatus` (terminal records are immutable; a pending cancellation request blocks non-terminal refreshes), but are evaluated against the same in-lock snapshot that is written back — so a terminal transition that landed earlier can no longer be clobbered mid-cycle, a slightly *stronger* guarantee than the per-task path.
- `WorkspaceManager.taskRecordStorePath(rc, agentId, sessionId)` — the resolved store-path identity used for grouping. Tasks captured with different `RuntimeContext` instances that map to the same store file still share one batch.
- `WorkspaceTaskRepository.heartbeat()` builds the batches and calls the new method once per store file; failure isolation moves from per-task to per-batch (grouping keeps per-task isolation, and both levels surface at `warn`).

No persistence format change, no API signature change; `writeTaskRecord` / `updateStatus` keep their exact semantics for single-record paths (task completion, cancellation, remote delivery).

## Benchmark

Standalone replica of the store mechanics (JDK 17, Jackson 2.21.1 — the project's BOM version —, `TaskRecord` mirrored field-for-field, finished records carrying ~2 KB result text, one cycle = R sequential refreshes of a store holding T records, median of 5 runs):

| Scenario (total / running) | Before: cycle / bytes written | After: cycle / bytes written |
|---|---|---|
| 20 / 5 | 4.8 ms / 283 KB | 0.7 ms / 57 KB |
| 100 / 25 | 76 ms / 7.1 MB | 2.9 ms / 285 KB |
| 400 / 100 | 1113 ms / 114 MB | 7.5 ms / 1.1 MB |

At the stress point that is ~114 MB written **per 30-second cycle** (≈328 GB/day while tasks are in flight) reduced to one whole-file rewrite per cycle, with the per-path lock hold going from one full read-modify-write per task to one per store file.

## Failure isolation

The pre-batching heartbeat could not throw (per-task try/catch); the batched version preserves that contract and widens it where the blast radius grew:

- grouping keeps a **per-task** try/catch — an exception escaping the grouping loop on the maintenance scheduler would permanently suppress every later heartbeat execution (`scheduleAtFixedRate` semantics), so a pluggable `NamespaceFactory` throwing for one task only skips that task's refresh for the cycle; a **grouping** failure logs at `warn`, because a task whose grouping persistently fails is never refreshed and is eventually orphan-swept;
- a **batch** failure (whole session for one cycle) also logs at `warn`, aligned with `persistRecord`;
- a batched write whose store cannot be parsed **aborts without writing** — a malformed store is never overwritten with partial data (same stance as `writeTaskRecord`);
- `updateTaskRecordStatuses` skips entries with a null/blank task ID or a null status instead of failing the batch (defensive input handling mirroring `writeTaskRecord`);
- the grouping-key contract — the resolved runtime-data path must be the write target, i.e. backends must not derive storage location from the `RuntimeContext` itself — is documented on `taskRecordStorePath`.

## Test evidence (TDD red → green)

New `WorkspaceTaskRepositoryHeartbeatBatchTest` (8 cases, counting backend writes through a routed in-memory filesystem):

| Case | Before (main) | After |
|---|---|---|
| 3 running tasks in one session → one store write per heartbeat | ❌ 3 writes | ✅ 1 write, all records refreshed |
| tasks across two sessions → one write per session file | ❌ 2 + 1 writes | ✅ 1 + 1 writes |
| terminal record never clobbered while siblings refresh (incl. no `touch`) | — (behaviour lock) | ✅ |
| record with pending cancellation not refreshed (incl. no `touch`) | — (behaviour lock) | ✅ |
| corrupt store: batch aborts, content untouched, no backend write | — (behaviour lock) | ✅ |
| grouping exception (throwing `NamespaceFactory`) skips one task, heartbeat survives, other task still refreshed | ❌ exception escapes `heartbeat()` | ✅ |
| `updateTaskRecordStatuses` skips entries with null/blank task ID or null status | ❌ NPE on null status | ✅ valid entries still persist |
| same-session tasks under two namespaces → two batches on the same physical store, both refreshed | — (pins the documented split) | ✅ |

The grouping test deliberately uses distinct `RuntimeContext` instances per task to pin the resolved-path grouping contract; the namespace test pins the documented one-cycle-per-namespace split (task-record IO is namespace-agnostic, so both batches target one physical file).

Regression: full harness suite green (963 tests, 6 skipped pre-existing), `spotless:check` clean.

## Known pre-existing limitation (not addressed here)

The single-record path (`updateStatus` → `writeTaskRecord`) still evaluates its terminal/cancellation guards against a read taken **outside** the write lock, so a cancellation landing between guard-read and write can in principle be overwritten by a concurrent completion. The batched path closes this window for heartbeats; converging the single-record path on the same in-lock guard evaluation is planned as a follow-up.

## Checklist

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/AbstractFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/AbstractFilesystem.java
@@ -170,6 +170,26 @@ public interface AbstractFilesystem {
     // ==================== Path validation utility ====================
 
     /**
+     * Returns an identity object for the storage location that workspace IO for {@code path}
+     * under {@code rc} resolves to. Two operations whose keys compare equal are guaranteed to
+     * target the same physical storage, which lets callers safely batch them into one
+     * read-modify-write.
+     *
+     * <p>The conservative default isolates per context instance: it only reports equal keys for
+     * the very same {@code RuntimeContext} and path, which is always safe for backends whose
+     * routing may depend on the context. Backends that locate content purely by path should
+     * override this to return the normalized path, and backends that derive the location from
+     * the context (e.g. a per-user namespace) must include that derived location in the key.
+     *
+     * @param runtimeContext per-call agent runtime; {@link RuntimeContext#empty()} when none
+     * @param path the workspace-relative path of the operation
+     * @return an object with value equality reflecting the storage identity
+     */
+    default Object storageKey(RuntimeContext runtimeContext, String path) {
+        return java.util.Arrays.asList(runtimeContext, path);
+    }
+
+    /**
      * Validates that {@code path} is safe (non-null, non-blank, no {@code ..} traversal).
      *
      * @param path the path to validate
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/OverlayFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/OverlayFilesystem.java
@@ -299,4 +299,9 @@ public class OverlayFilesystem implements AbstractFilesystem {
     public AbstractFilesystem getLower() {
         return lower;
     }
+
+    @Override
+    public Object storageKey(RuntimeContext runtimeContext, String path) {
+        return upper.storageKey(runtimeContext, path);
+    }
 }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
@@ -42,6 +42,7 @@ import io.agentscope.harness.agent.filesystem.model.WriteResult;
 import io.agentscope.harness.agent.filesystem.remote.store.NamespaceFactory;
 import io.agentscope.harness.agent.filesystem.sandbox.SandboxBackedFilesystem;
 import io.agentscope.harness.agent.subagent.task.TaskRecord;
+import io.agentscope.harness.agent.subagent.task.TaskStatus;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -491,6 +492,126 @@ public class WorkspaceManager implements AutoCloseable {
             record.touch();
             map.put(record.getTaskId(), record);
             persistTaskMap(rc, rel, map);
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    /**
+     * Returns the storage identity of a session's task-record store under the given context.
+     * This is the identity callers use to group per-record refreshes into one batched
+     * read-modify-write: keys that compare equal are guaranteed to hit the same physical
+     * storage.
+     *
+     * <p>The identity follows the routing the store IO actually takes: the host-disk fallback
+     * of the per-call sandbox route serves the workspace-relative path regardless of context,
+     * and every other backend is asked through
+     * {@link AbstractFilesystem#storageKey(RuntimeContext, String)}. A backend whose resolved
+     * storage location does not depend on the context batches every context of a session
+     * together (e.g. a {@code LocalFilesystem} without a {@code NamespaceFactory}); a backend
+     * that derives the location from the context splits per derived location, each batch
+     * written back under the context that produced it — a {@code RemoteFilesystem} with a
+     * per-user {@code NamespaceFactory}, or a {@code LocalFilesystem} configured with one,
+     * keeps per-user stores in separate batches.
+     *
+     * @param rc per-call agent runtime; {@link RuntimeContext#empty()} when none
+     * @param agentId the parent agent identifier
+     * @param sessionId the session identifier
+     * @return an object with value equality reflecting the store's storage identity
+     */
+    public Object taskRecordStoreKey(RuntimeContext rc, String agentId, String sessionId) {
+        String rel = taskRecordPath(agentId, sessionId);
+        if (taskRecordsRouteToLiveSandbox(rel)) {
+            return rel;
+        }
+        return filesystem != null ? filesystem.storageKey(rc, rel) : rel;
+    }
+
+    /**
+     * Refreshes the status of several {@link TaskRecord}s of one session store in a single
+     * read-modify-write cycle.
+     *
+     * <p>Batched sibling of {@link #writeTaskRecord} for callers that refresh many records of
+     * the same session — the heartbeat of
+     * {@link io.agentscope.harness.agent.subagent.task.WorkspaceTaskRepository} touches every
+     * running task every cycle. One lock acquisition, one parse and one whole-file persist
+     * regardless of how many records are refreshed, instead of a full cycle per record.
+     *
+     * <p>The per-record guards mirror {@code WorkspaceTaskRepository.updateStatus}: terminal
+     * records are immutable, and a record with a pending cancellation request refuses
+     * non-terminal refreshes. Unlike the single-record path they are evaluated against the same
+     * in-lock snapshot that is written back, so a terminal transition that landed earlier can
+     * never be clobbered by the batch.
+     *
+     * <p>Entries with a null or blank task ID or a null status are skipped, mirroring the
+     * defensive input handling of {@link #writeTaskRecord}.
+     *
+     * @param rc per-call agent runtime; {@link RuntimeContext#empty()} when none
+     * @param agentId the parent agent identifier
+     * @param sessionId the session identifier
+     * @param statusByTaskId the status to persist per task ID
+     */
+    public void updateTaskRecordStatuses(
+            RuntimeContext rc,
+            String agentId,
+            String sessionId,
+            Map<String, TaskStatus> statusByTaskId) {
+        if (agentId == null
+                || agentId.isBlank()
+                || sessionId == null
+                || sessionId.isBlank()
+                || statusByTaskId == null
+                || statusByTaskId.isEmpty()) {
+            return;
+        }
+        String rel = taskRecordPath(agentId, sessionId);
+        ReentrantLock lock = pathLocks.computeIfAbsent(rel, k -> new ReentrantLock());
+        lock.lock();
+        try {
+            Map<String, TaskRecord> map;
+            try {
+                map = readTaskMap(rc, rel); // already holding lock
+            } catch (IOException e) {
+                // Never overwrite a malformed store with partial data from a failed parse.
+                log.error(
+                        "Failed to parse task record store {}, aborting batched status update to"
+                                + " avoid data loss.",
+                        rel,
+                        e);
+                return;
+            }
+            boolean changed = false;
+            for (Map.Entry<String, TaskStatus> entry : statusByTaskId.entrySet()) {
+                String taskId = entry.getKey();
+                if (taskId == null || taskId.isBlank() || entry.getValue() == null) {
+                    continue;
+                }
+                TaskRecord existing = map.get(taskId);
+                if (existing != null
+                        && existing.getStatus() != null
+                        && existing.getStatus().isTerminal()) {
+                    continue;
+                }
+                if (!entry.getValue().isTerminal()
+                        && existing != null
+                        && existing.isCancelRequested()) {
+                    continue;
+                }
+                TaskRecord record = existing;
+                if (record == null) {
+                    record = new TaskRecord();
+                    record.setTaskId(taskId);
+                    record.setParentAgentId(agentId);
+                    record.setParentSessionId(sessionId);
+                }
+                record.setStatus(entry.getValue());
+                record.touch();
+                map.put(taskId, record);
+                changed = true;
+            }
+            if (changed) {
+                persistTaskMap(rc, rel, map);
+            }
         } finally {
             lock.unlock();
         }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/BakedContextFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/BakedContextFilesystem.java
@@ -110,4 +110,11 @@ public final class BakedContextFilesystem implements AbstractFilesystem {
     public boolean exists(RuntimeContext runtimeContext, String path) {
         return delegate.exists(bakedRc, path);
     }
+
+    @Override
+    public Object storageKey(RuntimeContext runtimeContext, String path) {
+        // Storage identity follows the substituted context, not the caller's: every delegated
+        // operation (and therefore its write target) is resolved under bakedRc.
+        return delegate.storageKey(bakedRc, path);
+    }
 }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/CompositeFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/CompositeFilesystem.java
@@ -105,6 +105,12 @@ public class CompositeFilesystem implements AbstractFilesystem {
         return routeForPath(path).backend();
     }
 
+    @Override
+    public Object storageKey(RuntimeContext runtimeContext, String path) {
+        RouteResult route = routeForPath(path);
+        return route.backend().storageKey(runtimeContext, route.backendPath());
+    }
+
     private RouteResult routeForPath(String path) {
         // Canonicalize both sides by stripping any leading slash before matching so callers can
         // pass either "/skills/foo" (the {@link AbstractFilesystem} contract) or "skills/foo"
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/RoutedSandboxFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/RoutedSandboxFilesystem.java
@@ -135,4 +135,9 @@ public final class RoutedSandboxFilesystem implements AbstractSandboxFilesystem 
     public boolean exists(RuntimeContext runtimeContext, String path) {
         return composite.exists(runtimeContext, path);
     }
+
+    @Override
+    public Object storageKey(RuntimeContext runtimeContext, String path) {
+        return composite.storageKey(runtimeContext, path);
+    }
 }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystem.java
@@ -922,4 +922,12 @@ public class LocalFilesystem implements AbstractFilesystem {
 
         return matches;
     }
+
+    @Override
+    public Object storageKey(RuntimeContext runtimeContext, String path) {
+        // The exact path this backend's IO resolves to — namespace prefix (when a
+        // NamespaceFactory is configured) and mode included — so contexts that store in
+        // different namespace directories never share a batch.
+        return resolvePath(runtimeContext, path).normalize();
+    }
 }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/remote/RemoteFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/remote/RemoteFilesystem.java
@@ -725,4 +725,9 @@ public class RemoteFilesystem implements AbstractFilesystem {
         }
         return key.equals(normalizedPath) || key.startsWith(normalizedPath + "/");
     }
+
+    @Override
+    public Object storageKey(RuntimeContext runtimeContext, String path) {
+        return List.of(getNamespace(runtimeContext), path);
+    }
 }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/subagent/task/WorkspaceTaskRepository.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/subagent/task/WorkspaceTaskRepository.java
@@ -26,6 +26,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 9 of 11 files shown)


guslegend0510 (changes requested):
在 WorkspaceTaskRepository.java 第 713–717 行，核心是 第 715 行的 batches.computeIfAbsent(...) 分组。

[P1] 合批可能跨命名空间写入任务记录

分组键来自 WorkspaceManager 推导的本地路径，但内置 RemoteFilesystem 可以按 RuntimeContext 独立选择远端命名空间。使用二参 WorkspaceManager 构造器、后端按用户分区时，同 session 的不同用户任务会被合批，并统一使用首个任务的上下文写入：另一用户的真实记录不再刷新，其 taskId 反而被写进首个用户的命名空间。应按后端实际存储身份分组；无法确认身份时保留上下文隔离，并补充真实 RemoteFilesystem 的跨用户回归测试。

guslegend0510:
[P1] LocalFilesystem.storageKey() 忽略了用户命名空间。
第 927–929 行直接返回 path，而实际读写通过 resolvePath(rc, path) 加上了 NamespaceFactory 对应的目录。因此，不同用户使用相同 agent/session 时，心跳仍会错误合批，并统一使用首个用户的上下文写入。
已通过真实 repo.heartbeat() 复现：Alice 的记录刷新成功，Bob 的记录未刷新，且 Bob 的任务 ID 被写入 Alice 的文件。
建议使用与实际 IO 一致的规范化绝对路径作为分组键，并补充 namespaced LocalFilesystem 的双用户 heartbeat 回归测试。临时验证这个修法后，两边均正常刷新且没有串写。

birdie7761:
> [P1] LocalFilesystem.storageKey() 忽略了用户命名空间。 第 927–929 行直接返回 path，而实际读写通过 resolvePath(rc, path) 加上了 NamespaceFactory 对应的目录。因此，不同用户使用相同 agent/session 时，心跳仍会错误合批，并统一使用首个用户的上下文写入。 已通过真实 repo.heartbeat() 复现：Alice 的记录刷新成功，Bob 的记录未刷新，且 Bob 的任务 ID 被写入 Alice 的文件。 建议使用与实际 IO 一致的规范化绝对路径作为分组键，并补充 namespaced LocalFilesystem 的双用户 heartbeat 回归测试。临时验证这个修法后，两边均正常刷新且没有串写。

感谢指出，这个问题完全成立——这是一个真实的正确性缺陷，而不是文档缺口。manager 推导的本地路径隐含假设了后端不感知 context，而内置的 `RemoteFilesystem` 配合用户级 `NamespaceFactory` 恰恰按 context 分区存储：合批后整个 store（两个用户的记录）会写进首个任务的命名空间，另一用户自己的命名空间持续停滞，直到其任务被孤儿清扫置为 FAILED。`6c5d2ef7` 按您建议的方向重构了分组——改为向后端询问真实存储身份，而不是假设一个：

- 新增 `AbstractFilesystem.storageKey(rc, path)`，返回操作实际落点的存储身份——键相等即保证命中同一物理存储。**默认实现就是您说的保守回退**：按 context 实例隔离，任何未覆写的后端都不可能被不安全地合批。
- 与 context 无关的后端（`LocalFilesystem` 及其 shell 子类）返回路径——同 session 的所有 context 共享一批。
- `RemoteFilesystem` 把推导出的命名空间折进键：按用户分区的存储每个用户一批，各自用产生该批的 context 写回。
- 路由包装类（`CompositeFilesystem`、`RoutedSandboxFilesystem`、`OverlayFilesystem`）沿其 IO 实际经过的路由取键；`BakedContextFilesystem` 按被替换的（baked）context 取键；sandbox 宿主盘回退按 workspace 相对路径分组，与它的写路由一致。

您要的回归测试：`heartbeat_userNamespacedRemoteStore_keepsRecordsPerUser` 使用**真实 `RemoteFilesystem`** 挂 `InMemoryStore`、以 `userId` 分区的 `NamespaceFactory` 构造；alice/bob 各一个同 session 任务，一次心跳后逐命名空间断言——自己的记录已刷新、零跨用户泄漏。该测试在修复前的版本上恰好以您描述的方式失败（整个 store 落进其中一个用户的命名空间）。

回归：harness 全量 997 个测试通过（6 个预存 skip），service-dataplane 22 个通过，spotless 干净。欢迎继续指正。

birdie7761:
> [P1] LocalFilesystem.storageKey() 忽略了用户命名空间。 第 927–929 行直接返回 path，而实际读写通过 resolvePath(rc, path) 加上了 NamespaceFactory 对应的目录。因此，不同用户使用相同 agent/session 时，心跳仍会错误合批，并统一使用首个用户的上下文写入。 已通过真实 repo.heartbeat() 复现：Alice 的记录刷新成功，Bob 的记录未刷新，且 Bob 的任务 ID 被写入 Alice 的文件。 建议使用与实际 IO 一致的规范化绝对路径作为分组键，并补充 namespaced LocalFilesystem 的双用户 heartbeat 回归测试。临时验证这个修法后，两边均正常刷新且没有串写。

完全成立，感谢再次抓住要害——我们把 `LocalFilesystem` 归类为 context-free 是错的：它带 `NamespaceFactory` 构造时，IO 经 `resolvePath(rc, path)` → `applyNamespacePrefix` 落在按用户分区的目录下，裸 `path` 作为键会跨用户合批。您描述的三个症状（Alice 刷新成功、Bob 未刷新、Bob 的 taskId 串写进 Alice 的文件）我们用真实 `repo.heartbeat()` 在修复前逐条复现了。

`01693cc0` 按您的建议修复：`LocalFilesystem.storageKey` 现在返回 `resolvePath(runtimeContext, path).normalize()`——即该后端 IO 实际解析出的规范化绝对路径，namespace 前缀（若配置了 factory）和 mode（SANDBOXED/ROOTED/UNRESTRICTED）都包含在内。未配置 factory 时退化为同一 resolved 路径，同 session 各 context 仍然合批，批量收益不受影响；`LocalFilesystemWithShell` 继承同一实现。

您要的回归测试已补：`heartbeat_userNamespacedLocalStore_keepsRecordsPerUser`——真实 `LocalFilesystem`（ROOTED + 按 userId 分区的 `NamespaceFactory`），alice/bob 各一个同 session 任务，一次心跳后逐文件断言：各自的 namespace 目录下自己的记录已刷新，且两个文件里都没有对方的 taskId。该测试在裸 `path` 键的版本上以您复现的方式失败。

顺带说明：这个错误也纠正了我们 PR 描述里的归类——"LocalFilesystem 返回路径"的表述已改为"按实际解析路径取键（含 namespace 前缀与 mode）"。`MemoryStoreFilesystem` 我们复核过：其存储位置完全由构造参数 `(ownerId, storeId)` 决定，各方法签名中的 rc 未参与任何路由，按 `(ownerId, storeId, path)` 取键是准确的。

回归：harness 全量 998 个测试通过（6 个预存 skip），spotless 干净。欢迎继续指正。

guslegend0510 (approved):
LGTM

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
```diff
@@ -496,6 +497,126 @@ public void writeTaskRecord(
...
+     * keeps per-user stores in separate batches.
+     *
+     * @param rc per-call agent runtime; {@link RuntimeContext#empty()} when none
+     * @param agentId the parent agent identifier
+     * @param sessionId the session identifier
+     * @return an object with value equality reflecting the store's storage identity
+     */
+    public Object taskRecordStoreKey(RuntimeContext rc, String agentId, String sessionId) {
+        String rel = taskRecordPath(agentId, sessionId);
+        if (taskRecordsRouteToLiveSandbox(rel)) {
```
oss-maintainer, line 534: **[Warning]** `taskRecordStoreKey` resolves the live-sandbox branch once, at grouping time, and bakes the answer into the batch key: `rel` when `taskRecordsRouteToLiveSandbox(rel)` is true (context-free host disk, merging *all* contexts of the session), otherwise `filesystem.storageKey(rc, rel)` (context-partitioned, one batch per derived location). But `readTaskMap` / `persistTaskMap` re-evaluate `taskRecordsRouteToLiveSandbox(rel)` independently at write time, and `backendFor(rel)` is exactly the thing that flips during sandbox acquire/release.

If the route changes between grouping and the batched write, the batch was formed under the host-disk assumption (`rel`, one merged batch, keyed by the first task's `rc`) while the write lands on the context-partitioned path — i.e. every other context's taskIds get persisted under the first task's namespace, and their own stores are never touched. That is the same failure mode as the two P1s already resolved in this thread, just reached through a lifecycle race instead of a missing namespace.

Could the branch decision travel with the batch (pass the resolved key/branch into `updateTaskRecordStatuses` and skip or re-split on mismatch), or the whole grouping+write run under the same `pathLocks` entry so the route cannot change underneath it? The 30s cycle makes the window small, but the failure is silent and the sweeper will report the affected tasks as orphaned rather than pointing here.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/AbstractFilesystem.java
```diff
@@ -169,6 +169,26 @@ List<FileUploadResponse> uploadFiles(
...
+     * the very same {@code RuntimeContext} and path, which is always safe for backends whose
+     * routing may depend on the context. Backends that locate content purely by path should
+     * override this to return the normalized path, and backends that derive the location from
+     * the context (e.g. a per-user namespace) must include that derived location in the key.
+     *
+     * @param runtimeContext per-call agent runtime; {@link RuntimeContext#empty()} when none
+     * @param path the workspace-relative path of the operation
+     * @return an object with value equality reflecting the storage identity
+     */
+    default Object storageKey(RuntimeContext runtimeContext, String path) {
```
oss-maintainer, line 201: **[Warning]** Returning a bare `Object` from a public interface method makes the batching contract unverifiable at compile time. Grouping only works if implementations return something with value equality — but a backend that returns `new Object()`, a `char[]`, or any identity-equality type gets *silently* per-context batches, so the optimisation quietly disappears (or, worse, degrades after a refactor with no test signal, since the heartbeat tests here pin the concrete backends). The `java.util.Arrays.asList(...)` default is safe, and `LocalFilesystem` / `RemoteFilesystem` / `MemoryStoreFilesystem` all return proper value types, so today's behaviour is correct — the risk is for the implementations this PR cannot see.

Since `AbstractFilesystem` is public API and changes cascade to the extensions modules, a `List<Object>` return type (or a small `record StorageKey(Object location, String path)`) would make the intent self-enforcing, and would also let `WorkspaceTaskRepository` group on a typed key instead of `Map<Object, StoreBatch>`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/OverlayFilesystem.java
```diff
@@ -299,4 +299,9 @@ public AbstractFilesystem getUpper() {
     public AbstractFilesystem getLower() {
         return lower;
     }
+
+    @Override
+    public Object storageKey(RuntimeContext runtimeContext, String path) {
+        return upper.storageKey(runtimeContext, path);
```
oss-maintainer: **[Info]** The overlay keys on `upper` only. That is right for the *write* target, but a task store whose current content lives in `lower` (a read-through overlay) means two contexts can share an upper key while their merged read view differs. `updateTaskRecordStatuses` snapshots under the first task's `rc` and then persists the whole map to upper, so the second context's lower-only records simply aren't in that map — no data is deleted (upper has no entry for them, so lower stays visible to that context), but the record they expected refreshed is not refreshed, and it will age into the orphan sweep with nothing pointing at the overlay as the cause.

Worth either documenting overlay routing as unsupported for task-record batching, or folding the lower identity into the key (`List.of("overlay", upper.storageKey(...), lower.storageKey(...)")`). If task stores are never routed through an `OverlayFilesystem` in practice, a line in the javadoc here saying so is enough — I could not tell from the call sites.

oss-maintainer (commented):
### Summary
Batches the sub-agent task-record heartbeat from one full read-modify-write of the session store *per running task* into one per resolved storage location, removing the `O(running tasks x store size)` parse/serialize/write cost every 30s cycle. Batching correctness hinges on a new `AbstractFilesystem.storageKey(rc, path)` identity contract, implemented per backend (`LocalFilesystem` via `resolvePath`, `RemoteFilesystem` via namespace, `BakedContextFilesystem` via the substituted context, `MemoryStoreFilesystem` per store instance) with a conservative identity default.

Solid change, and the review history in this thread shows it: the two earlier P1s (context-free merging across user namespaces, `LocalFilesystem` ignoring its namespace) are both fixed by making the key the *resolved* location rather than the raw path. The in-lock re-evaluation of the terminal/cancelRequested guards against the snapshot that gets written back is the correct place for those checks — it closes a race the single-record path had — and the parse-failure bail-out (`return` before persist) is the right instinct. Test coverage is excellent for a perf refactor: write counting, per-namespace split, terminal-record protection, and lifecycle race.

Main open question is the grouping-time vs write-time evaluation of `taskRecordsRouteToLiveSandbox` (inline below); the rest is API-hardening detail.

### Checked
- Distinct sessions cannot merge into one batch: every key derivation includes the store path, which embeds the sessionId, so `StoreBatch` keeping the first task's `rc`/`sessionId` stays consistent.
- The `sid == null || sid.isBlank()` addition and `getOrDefault(key, RuntimeContext.empty())` are behaviour-preserving against the old explicit null check.
- Per-batch `log.warn` replaces per-task `log.debug`, so a persistently failing backend now logs one line per session per cycle instead of one per task — less volume, higher severity; fine, but worth confirming operators are not alerting on warn-level in this module.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/AbstractFilesystem.java
birdie7761: Kept the `Object` shape and hardened the contract into the interface docs, for two reasons:

- The failure mode you describe is silent loss of the optimization, never of correctness:
  an identity-equality key degrades to one batch per key instance, which is exactly the
  pre-batching behaviour — every refresh pays its own full read-modify-write. The javadoc now
  states this degradation semantics outright.
- A narrower type doesn't actually enforce value equality: `List.of(new Object())` and a
  record holding an identity-equal part degrade identically. The compile-time guarantee you're
  after doesn't exist for arbitrary parts; what can be enforced is the documented contract
  plus the conservative default (which never merges). Every in-repo implementation already
  returns a value type, and the heartbeat tests pin the concrete backends.

Happy to move to `List<Object>` or a `StorageKey` record as a follow-up if maintainers prefer
the shape — it does read better at the `Map<Object, StoreBatch>` call site — it just doesn't
buy the enforcement.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/OverlayFilesystem.java
birdie7761: Took the key fold, with one honest correction to the suggested fix's scope —
`List.of("overlay", upper.storageKey(...), lower.storageKey(...))` is in 9213fc2d, because the
principle is right: a batch is a read-modify-write and its snapshot reads through to lower
while upper lacks the path, so the key must capture the **read** identity, not just the write
target. Contexts whose lower views differ now land in different batches (new key-level test,
red on the upper-only key); when both layers resolve the path identically for every context
the composed key is equal and batching merges exactly as before.

But folding alone cannot fully fix the corner you described, and the javadoc now says so:
`OverlayFilesystem.read` is upper-exists-else-lower, so with sequential whole-file
read-modify-writes, the first batch's persist creates the upper file and shadows every other
context's lower-only copies for all subsequent reads — no grouping rule can merge a
context-partitioned lower into one upper file. In live flows this corner is unreachable
through the repository: `putTask` writes the task record through the overlay at registration,
so upper holds the union from the start and every later read — single-task or batched — sees
the same file. The residual limitation (externally-seeded partitioned lower content) is
documented on the method rather than papered over.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/AbstractFilesystem.java
```diff
@@ -169,6 +169,33 @@ List<FileUploadResponse> uploadFiles(
...
+     * identity-equal key silently degrades batching to one batch per key instance — every
+     * refresh pays its own full read-modify-write, exactly the pre-batching behaviour. No
+     * correctness is lost, only write coalescing.
+     *
+     * @param runtimeContext per-call agent runtime; {@link RuntimeContext#empty()} when none
+     * @param path the workspace-relative path of the operation
+     * @return an object with value equality reflecting the storage identity
+     */
+    default Object storageKey(RuntimeContext runtimeContext, String path) {
+        return java.util.Arrays.asList(runtimeContext, path);
```
oss-maintainer, line 202: The default key is `Arrays.asList(runtimeContext, path)`, but `RuntimeContext` does not override `equals`/`hashCode` (main@`39cd304a` has no value equality, and `RuntimeContext.empty()` builds a new instance per call), so this default is identity-equal — exactly the case the contract above warns about. Any filesystem that inherits the default therefore keeps the pre-PR behaviour silently: one batch per task instead of one per store. `LocalFilesystem` and `RemoteFilesystem` avoid it by returning the resolved physical path. Could the default do the same (derive from the backend's own path/namespace resolution) instead of embedding the context object, so "forgot to override" degrades to correct-but-shared rather than to no batching at all?

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
```diff
@@ -496,6 +497,136 @@ public void writeTaskRecord(
...
+                || agentId.isBlank()
+                || sessionId == null
+                || sessionId.isBlank()
+                || statusByTaskId == null
+                || statusByTaskId.isEmpty()) {
+            return;
+        }
+        String rel = taskRecordPath(agentId, sessionId);
+        ReentrantLock lock = pathLocks.computeIfAbsent(rel, k -> new ReentrantLock());
+        lock.lock();
```
oss-maintainer, line 579: The per-path lock is now held across a full read-parse-serialize-write of the whole store, so its hold time scales with the number of refreshed tasks in the batch rather than one record. That is required for the read-modify-write to stay atomic, and `writeTaskRecord` already uses the same `pathLocks`, so no new blocking is introduced — just worth remembering for sessions with many concurrent sub-agent tasks, where an interactive `writeTaskRecord` for the same session now queues behind the whole batch.

oss-maintainer (commented):
### Summary
Re-review after the author's follow-up commit `9213fc2da3` ("include the overlay lower layer's read identity in the batching key"), which was a response to the previous review round. The batching itself is correct and the risky part is preserved: `updateTaskRecordStatuses` re-implements both guards that `WorkspaceTaskRepository.updateStatus` had — a terminal status is never overwritten, and a non-terminal write is refused when `cancelRequested` is set — and it evaluates them against a map re-read **inside** the per-path lock, so it cannot act on a pre-cancellation snapshot. Grouping is per store key with the context that produced it, a blank `sessionId` is skipped, each task's grouping call is individually try/caught so an exception cannot kill the maintenance scheduler thread (and the log level was correctly raised from debug to warn, since a batch failure's blast radius is the whole session), and invalid entries are skipped. `WorkspaceTaskRepositoryHeartbeatBatchTest` covers context-partitioned vs context-free backends and the cross-user case.

Two points inline: the `AbstractFilesystem.storageKey` default is identity-equal because `RuntimeContext` has no `equals`/`hashCode`, which silently disables coalescing for any backend that inherits it; and a note on the per-path lock hold time now scaling with batch size.

Coverage caveat: +1344/-8 over 11 files; this pass read the main-source changes in full and the new test file in summary. No build/test run was possible on this host (no JDK 17 / Maven).

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/AbstractFilesystem.java
birdie7761: Sharp observation — `RuntimeContext` indeed overrides no `equals`/`hashCode`, so the default
key is identity-equal by construction. That is deliberate rather than an oversight, and the
javadoc now says so explicitly (commit on top of 9213fc2d): a context-blind default (the bare
path) would merge contexts a backend routes apart — exactly the cross-namespace failure this
thread already fixed once — and no proxy of context fields can stand in for the backend's own
resolution, since a `NamespaceFactory` may key on any context property. Deriving from "the
backend's own path/namespace resolution" is precisely the information the interface default
does not have; that is what overrides provide.

For reach: every in-repo backend overrides the method (Local, Remote, Composite, Routed,
Overlay, BakedContext, MemoryStore), so the default only reaches third-party implementations
— where it degrades to correct-but-uncoalesced, the documented floor.

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
birdie7761: Acknowledged, no change: the hold time scales with the batch by necessity (the read-modify-
write must stay atomic against `writeTaskRecord`, which shares the same `pathLocks`), and the
cadence bounds it in practice — batches run on the 30s heartbeat and a full store
re-serialization is milliseconds at the sizes the sweeper sees, so an interactive
`writeTaskRecord` queues behind at most one such cycle, not behind per-task writes as before.

oss-maintainer (commented):
### Summary

Delta re-review of the one commit pushed since my last comment (`9213fc2d..6101f8d9`, one file, javadoc only). It answers the point I raised about `AbstractFilesystem.storageKey`: the identity-equal default is now stated as intentional rather than an oversight, with the reason spelled out (`RuntimeContext` overrides no `equals`/`hashCode`, so a context-blind default would merge contexts that a backend routes apart, and no proxy of context fields can stand in for the backend's own resolution), plus the consequence for implementations that inherit it — they degrade to one batch per key instance, i.e. exactly the pre-batching behaviour. The existing `@return` contract text about value equality is untouched, and no key computation changed.

Because there is no functional change I did not re-read the batching logic: the two guards that matter (a terminal status is never overwritten; a non-terminal write is refused once `cancelRequested` is set, both evaluated against a map re-read inside the per-path lock) are the same code reviewed in full last round, and all review threads from earlier rounds are now resolved with zero outstanding.

### Findings

None — doc-only delta, and the added text is accurate against the implementation it documents.

### Verdict

**COMMENT (no change request).** All checks are green on `6101f8d9` (build ubuntu + windows, license, module sync, codecov patch, source), `mergeable=MERGEABLE`, and `mergeStateStatus=BLOCKED` only because a review decision is still outstanding. I am not posting `APPROVE` this round for one reason only: the commit status for `6101f8d9` carries no `license/cla` context, so CLA state is `unknown` rather than `signed` for this head, and an approve would read as "safe to merge" while that piece is unverified. Once a `license/cla` status appears on this head (or a maintainer confirms it is not enforced here), this is good to go as-is.

---

oss-maintainer:
**Correction to the review comment above:** the thread count was misquoted. This PR has **5** review threads, all resolved; the '21 threads' figure belongs to a different PR's re-review request and was pasted in error. Nothing else in that review changes — the delta is still javadoc-only, CI is green on `6101f8d9`, and the only reason it was posted as `COMMENT` instead of `APPROVE` is the missing `license/cla` status on this head.

---
*Automated correction by github-manager-bot*

oss-maintainer (approved):
Re-reviewed on head `1cb91f3a` and approving.

The only change since my previous review (`6101f8d9`, which had no findings) is the `Merge branch 'main'` sync: I compared every file in the PR between `6101f8d9` and `1cb91f3a` and all 11 blobs are byte-identical, so the batching logic, the `storageKey` identity-equal default (with the javadoc note explaining why, which was my round-1 point) and the overlay read-identity fix are unchanged. All 5 review threads are resolved, and the remaining reason I could not approve last time is now cleared — the head carries `license/cla: success`.

Checks on `1cb91f3a`: `build (ubuntu-latest)`, `build (windows-latest)`, `Check License`, `Check Module Sync`, `source`, `codecov/patch` all green.

`mergeStateStatus` is still `blocked`, which is branch protection (an approving review from a required owner), not something on my side — I am leaving the merge to the maintainers.

---

PR #2952, opened 2026-09-03
fix(harness): sanitize session ids used in workspace file names

wylovelyi:
## AgentScope-Java Version

`2.0.3-SNAPSHOT` (current `main`)

## Description

Closes #2937.

**Background.** `RuntimeContext.sessionId` is caller-supplied and carries no documented
character constraint, yet `WorkspaceManager` uses it verbatim to build file names
(`agents/{agentId}/sessions/{sessionId}.jsonl`, `agents/{agentId}/tasks/{sessionId}.json`).
A namespaced id such as `agent:abc-123:main:main-9f3c` is perfectly legal as an identifier,
but the colon is reserved by Windows/NTFS, so path construction fails on Windows with a bare
`InvalidPathException` thrown from deep inside the workspace layer — with no hint that the
caller-supplied id is the problem.

Reproduced locally on Windows 11 / JDK 17:

```
java.nio.file.InvalidPathException: Illegal char <:> at index 5: agent:abc-123:main:main-9f3c.json
```

**Change.** Reserved characters are replaced with `-` in the file-name segment rather than
rejected, so existing ids keep working unchanged on every platform:

- new `UNSAFE_SEGMENT_CHARS` pattern `[<>:"/\\|?*]` plus a small `safeSegment(String)` helper
- applied at the four places where `sessionId` becomes part of a file name:
  `resolveSessionFile`, `resolveSessionContextFile`, `resolveSessionLogFile`, `taskRecordPath`

`/` and `\` are included deliberately: `sessionId` is a single file-name segment here, so either
one would otherwise be read as a directory separator and silently write outside the session
directory. This follows the same approach as the already-merged #1031 and #2921, which stripped
Windows-reserved characters from skill source paths.

**No on-disk data changes shape.** Writes (`SessionTranscriptWriter`) and reads
(`SessionSearchTool`) both go through these same helpers, so the sanitized name is symmetric —
there is no way to write a transcript that can no longer be found. Ids that work today contain
no reserved character and are returned unchanged; ids that do contain one could not produce a
file on Windows in the first place, since they threw before any file was created.

Scope is kept to `sessionId`, as the issue describes. `agentId` reaches path construction the
same way and would benefit from the same treatment — happy to extend this PR if you would
prefer the two handled together.

**How to test.** `WorkspaceManagerPathSafetyTest` gains two cases, both platform-independent
(they assert the replacement behaviour rather than relying on a Windows-only crash):

[...]

PR #3020, opened 2026-09-07
fix(harness): expose session workspace in prompt

### LINKED ISSUE #2941: [Feature]: 无沙箱环境运行时，运行skill脚本技能的产物读取不到
/workspace 初始化智能体时工作目录配置的路径，session隔离权限，观测到调用llm时Middleware会在提示词新增工作目录路径，会以/workspace为工作目录，llm执行脚本命令时就会以/workspace为根目录去执行脚本文件，脚本执行出来的产物也是在/workspace ，list_file工具只会读取/workspace/session_xxx 的文件夹下的文件，导致agent获取不到skill的产物

ningmao-hlyz:
## Summary

- derive the effective per-call workspace path from the existing namespace factory for local overlay filesystems
- advertise that path consistently in both the AgentStateStore context and Workspace prompt sections
- keep the base workspace for path-policy roots and leave sandbox, remote-filesystem, and shell behavior unchanged

## Problem

With session isolation enabled, relative file-tool paths resolve under the session namespace (for example, `/workspace/session-1`), while the prompt currently advertises the base workspace (`/workspace`). Models can therefore run skill scripts against the wrong path, write artifacts outside the session namespace, and then fail to find those artifacts through file tools.

This change only corrects the path shown to the model. It reuses `WorkspaceManager.resolveRuntimeDataPath` so the advertised path follows the same namespace factory already used by runtime data.

## Testing

- `mvn -pl agentscope-harness spotless:check`
- `mvn -pl agentscope-harness -Dtest="WorkspaceContextMiddlewarePathBoundsTest,WorkspaceContextMiddlewareMemoryPromptTest,WorkspaceContextMiddlewareSandboxPromptTest,LocalFilesystemModeTest,LocalFilesystemWithShellTest" test` (38 tests passed)

Fixes #2941


agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddleware.java
```diff
@@ -215,7 +215,12 @@ private String buildWorkspaceSection(RuntimeContext rc) {
                 includeMemoryContext ? workspaceManager.readMemoryMd(rc).strip() : "";
         String knowledgeContent = workspaceManager.readKnowledgeMd(rc).strip();
         Path workspace = workspaceManager.getWorkspace();
-        String sessionContext = buildSessionContextSection(workspace, rc);
+        AbstractFilesystem filesystem = workspaceManager.getFilesystem();
+        Path effectiveWorkspace =
+                detectLocalUpper(filesystem) != null
+                        ? workspaceManager.resolveRuntimeDataPath(rc, "")
```
xiaoyuuuuuupeng, line 221: This advertises /workspace/session-1 under SESSION isolation, but HarnessAgent still wires WorkspacePathNormalizer to the base /workspace. So if the model follows the prompt and calls write_file with /workspace/session-1/artifact.txt, the normalizer strips that down to session-1/artifact.txt, then LocalFilesystem prefixes the session again — and you land on /workspace/session-1/session-1/artifact.txt. Shell using the advertised absolute path won't find that file (same story when reading a shell-written artifact by absolute path).

Could you make absolute and relative paths hit the same place, and add a FilesystemTool regression with HarnessAgent's normalizer so we don't get a duplicated session directory? The new prompt-string checks don't cover this. Also noting: this gap was already there; the new prompt just makes it easier to hit.

xiaoyuuuuuupeng (changes requested):

oss-maintainer:

This PR currently conflicts with the `main` branch and cannot be merged. Please rebase (or merge `main`) and resolve the conflicts:

```bash
git fetch origin
git checkout fix/2941-session-workspace-context
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddleware.java
ningmao-hlyz: Thanks for catching this. The issue was valid: under SESSION isolation, the prompt could advertise a session-scoped absolute workspace path, while the filesystem namespace was applied again after path normalization.
I updated the existing PR to normalize paths with the current RuntimeContext namespace, so the advertised absolute path and the relative path now resolve to the same file without producing a duplicated session directory such as session-1/session-1.

ningmao-hlyz:
> ⚠️ **检测到合并冲突**
> 
> 该比例原则目前与该分支存在冲突，无法合并。请重新基底（或合并）并解决冲突：`main``main`
> 
> ```shell
> git fetch origin
> git checkout fix/2941-session-workspace-context
> git rebase origin/main
> # resolve conflicts, then:
> git push --force-with-lease
> ```
> 
> 这是一次性提醒。冲突解决后，欢迎随时@mention我重新审核。
> 
> _由 github-manager-bot 自动通知_

@oss-maintainer

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspacePathNormalizer.java
```diff
@@ -75,9 +97,38 @@ public static WorkspacePathNormalizer of(String... workspacePrefixes) {
      * @return workspace-relative path, or the original path if no registered prefix matched
      */
     public String normalize(String path) {
+        return normalize(path, RuntimeContext.empty());
```
oss-maintainer, line 100: The retained 1-arg overload now silently resolves the namespace from `RuntimeContext.empty()`. For a normalizer built **with** a `namespaceFactory`, an empty context yields an empty namespace, so the namespaced branch is skipped and the plain base prefix is stripped instead: `/workspace/session-1/artifact.txt` -> `session-1/artifact.txt` -> the filesystem re-applies the namespace -> `/workspace/session-1/session-1/artifact.txt`. That is precisely the double-nesting failure the new test asserts must not happen, but it becomes reachable through any caller that keeps (or later adds) a 1-arg call.

Both production callers were migrated here, so nothing is broken today — could the overload be deprecated, or made to throw / log when `namespaceFactory != null`, so the context is always explicit? Otherwise the invariant "always normalise with the operation's RuntimeContext" is held only by convention, and it is a very quiet failure mode when it slips.

Minor related point: the Javadoc on `normalize(String)` still describes only prefix stripping, without mentioning that the result now depends on a namespace resolved from an empty context.

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspacePathNormalizer.java
```diff
@@ -75,9 +97,38 @@ public static WorkspacePathNormalizer of(String... workspacePrefixes) {
...
+    public String normalize(String path, RuntimeContext runtimeContext) {
         if (path == null || path.isBlank()) {
             return path;
         }
+        if (namespaceFactory != null) {
+            List<String> namespace =
+                    namespaceFactory.getNamespace(
+                            runtimeContext != null ? runtimeContext : RuntimeContext.empty());
+            if (namespace != null && !namespace.isEmpty()) {
+                String namespacePrefix = String.join("/", namespace);
```
oss-maintainer, line 123: Composing the prefix as `prefix + "/" + String.join("/", namespace)` treats namespace segments as trusted path components. If a session id (or any other namespace segment) can contain `/`, `..`, or an empty segment, the composed prefix can match a path outside the intended session — e.g. a `..` segment makes `prefix + "/.."` match `/workspace/../etc`, which then strips down to a bare relative name and escapes the session scoping rather than containing it.

Is session-id sanitisation guaranteed upstream? #2952 (`fix(harness): sanitize session ids used in workspace file names`) looks like it addresses the same surface for file names, so this may already be covered — but the two changes compose right here. Worth adding an explicit test with a hostile id (`a/../../b`, absolute-looking or empty segments) asserting the path is **not** stripped, to pin the contract between id sanitisation and prefix composition.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddleware.java
```diff
@@ -215,7 +215,12 @@ private String buildWorkspaceSection(RuntimeContext rc) {
                 includeMemoryContext ? workspaceManager.readMemoryMd(rc).strip() : "";
         String knowledgeContent = workspaceManager.readKnowledgeMd(rc).strip();
         Path workspace = workspaceManager.getWorkspace();
-        String sessionContext = buildSessionContextSection(workspace, rc);
+        AbstractFilesystem filesystem = workspaceManager.getFilesystem();
+        Path effectiveWorkspace =
```
oss-maintainer, line 219: Scope question rather than a defect: `effectiveWorkspace` is substituted into the session-context section and the Workspace paragraph, but `buildKnowledgeBlock(...)` two lines below is still passed the base `workspace`, and `WorkspaceManager.listKnowledgeFiles` relativises its local-disk half against the base workspace while `getKnowledgeDir()` returns the un-namespaced `workspace.resolve(KNOWLEDGE_DIR)`.

So under session isolation the model sees "Workspace (your home base): /workspace/session-1" while `Knowledge files:` entries remain relative to `/workspace`, and session-scoped knowledge files may not appear in the listing at all (the local walk looks in the base dir; only the `filesystem.glob(rc, ...)` half is context-aware). Since the symptom reported in #2941 is exactly models resolving advertised paths against the wrong root, this looks like the same bug class in an adjacent section.

If keeping knowledge base-scoped is deliberate (shared read-only knowledge across sessions), could you note that in a short comment so the next reader does not "fix" it? Otherwise passing `effectiveWorkspace` (and namespacing `getKnowledgeDir()`) would complete the correction. Happy to see that as a follow-up PR rather than blocking this one.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddleware.java
```diff
@@ -235,7 +240,7 @@ private String buildWorkspaceSection(RuntimeContext rc) {
 
         String workspaceParagraph =
                 buildWorkspaceParagraph(
-                        workspace, workspaceManager.getFilesystem(), artifactDeliveryEnabled);
+                        workspace, effectiveWorkspace, filesystem, artifactDeliveryEnabled);
```
oss-maintainer, line 243: Nit: threading `workspace` and `effectiveWorkspace` side by side through the private helpers makes it easy to pick the wrong one at a call site later, and the two are only distinguishable by name. Since `extraRootsOf(localUpper, project, workspace)` and the path-policy roots genuinely want the base while the prompt text wants the effective root, consider a small value holder (e.g. `WorkspaceRoots(Path base, Path effective)`) or resolving both once inside the method. Not blocking.

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspacePathNormalizer.java
```diff
@@ -75,9 +97,38 @@ public static WorkspacePathNormalizer of(String... workspacePrefixes) {
...
+     *
+     * @param path the raw path (absolute or relative)
+     * @param runtimeContext current operation context
+     * @return workspace-relative path, or the original path if no registered prefix matched
+     */
+    public String normalize(String path, RuntimeContext runtimeContext) {
         if (path == null || path.isBlank()) {
             return path;
         }
+        if (namespaceFactory != null) {
```
oss-maintainer, line 118: Suggestion for completeness: `String.join("/", namespace)` supports multi-segment namespaces, but the new tests only exercise a single `sessionId` segment. If any isolation scope composes two dimensions (e.g. user + session) the prefix becomes two levels deep, which is exactly where a mistake would hide — one test with a 2-segment namespace would cover it cheaply.

Also worth confirming the overlay-without-namespace case (`namespaceFactory != null` but an empty namespace) stays byte-identical to the pre-PR prompt, since `HarnessAgent` now always passes `nsFactory` for local overlays.

oss-maintainer (commented):
### Summary

Re-review after the merge conflict was resolved (thanks for the quick rebase, and for the `@mention`). Conflicts are gone (`mergeable: MERGEABLE`), CLA is signed, and all checks pass on the current head — including `build (windows-latest)`, which matters for a path-normalisation change.

The fix is the right shape: rather than patching the prompt string in isolation, it derives the advertised root from the same `NamespaceFactory` the filesystem already applies at operation time, and threads the operation's `RuntimeContext` through `norm(...)` so absolute and relative paths converge on one file. Reusing `resolveRuntimeDataPath` keeps a single source of truth, and the end-to-end `sessionIsolation_absoluteAndRelativeWorkspacePathsResolveToSameFile` test is the right way to pin that contract — asserting the negative (`session-1/session-1/...` must not exist) is especially good.

Verdict: **no blocking issues; left as COMMENT so the three questions below can be answered or explicitly deferred.** Two are about how safely this new dual-mode normaliser resists future misuse, one is a scope question about an adjacent prompt section that appears to have the same bug class as #2941.

### Findings
- **[Warning]** `WorkspacePathNormalizer.java:100` — the 1-arg `normalize(String)` overload now silently resolves the namespace from an empty context, which reproduces the double-nesting bug for any caller that does not pass a `RuntimeContext`.
- **[Warning]** `WorkspacePathNormalizer.java:123` — namespace segments are treated as trusted path components; worth pinning the contract with id sanitisation (#2952) via a hostile-id test.
- **[Info]** `WorkspaceContextMiddleware.java:219` — `buildKnowledgeBlock` still receives the base workspace, so `Knowledge files:` stays relative to the un-namespaced root: same bug class as #2941, or intentional?
- **[Info]** `WorkspaceContextMiddleware.java:243` — two adjacent `Path` params distinguished only by name; a small holder type would prevent future mix-ups.
- **[Info]** `WorkspacePathNormalizer.java:118` — multi-segment namespaces and the empty-namespace overlay case are untested.

### Suggestions
- For the overload: prefer making the safe path the only path — either `@Deprecated` the 1-arg method with a note, or have it throw (or at least `log.warn`) when `namespaceFactory != null`. A compile-time/deprecation warning is much cheaper than a misplaced artifact at runtime.
- For the knowledge block: if base-scoped is deliberate, a one-line comment saying so is enough; otherwise it likely wants `effectiveWorkspace` too.
- Consider a test asserting the rendered prompt is unchanged when isolation is disabled, so the refactor of `buildWorkspaceParagraph` is provably behaviour-preserving for the common single-tenant configuration.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspacePathNormalizer.java
```diff
@@ -65,19 +69,71 @@ public static WorkspacePathNormalizer of(String... workspacePrefixes) {
...
+     * <p>Callers using a normalizer with a namespace factory should use
+     * {@link #normalize(String, RuntimeContext)} so the operation context is available. This
+     * overload is retained for compatibility and uses an empty context.
+     *
      * @param path the raw path (absolute or relative)
      * @return workspace-relative path, or the original path if no registered prefix matched
      */
+    @Deprecated
     public String normalize(String path) {
+        return normalize(path, RuntimeContext.empty());
```
oss-maintainer, line 105: **[Warning]** Good step — but the deprecation is currently descriptive rather than structural. The javadoc only *tells* callers with a `namespaceFactory` to prefer `normalize(String, RuntimeContext)`; nothing prevents a future call site from taking the 1-arg overload and silently falling back to the base prefix (the exact `session-1/session-1/...` mis-write this PR fixes).

Since there are no remaining unscoped call sites in main today (`FilesystemTool` / `ArtifactDeliveryTool` are both migrated), consider making the safe path the only path:
```java
// option A — no unscoped escape hatch on a namespaced normalizer
public static WorkspacePathNormalizer unscoped(String... prefixes) { ... }
public static WorkspacePathNormalizer namespaced(String prefix, NamespaceFactory nsf) { ... }

// option B — keep the overload, but make the silent case observable
if (namespaceFactory != null) {
    log.debug("normalizing without RuntimeContext; namespace prefix not applied to {}", path);
}
```
Option A also lets the deprecation be retired later without a second API break.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddleware.java
```diff
@@ -215,7 +215,12 @@ private String buildWorkspaceSection(RuntimeContext rc) {
                 includeMemoryContext ? workspaceManager.readMemoryMd(rc).strip() : "";
         String knowledgeContent = workspaceManager.readKnowledgeMd(rc).strip();
         Path workspace = workspaceManager.getWorkspace();
-        String sessionContext = buildSessionContextSection(workspace, rc);
+        AbstractFilesystem filesystem = workspaceManager.getFilesystem();
+        Path effectiveWorkspace =
+                detectLocalUpper(filesystem) != null
+                        ? workspaceManager.resolveRuntimeDataPath(rc, "")
+                        : workspace;
```
oss-maintainer, line 222: **[Info]** The `effectiveWorkspace` gate is `detectLocalUpper(filesystem) != null`, i.e. local-overlay only. A plain (non-shell, non-overlay) `LocalFilesystem` built with the same `nsFactory` still gets namespaced relative paths at operation time, but falls into the generic `"Your working directory is: "` branch and keeps advertising the unscoped root — the same mismatch as #2941 for that topology, and a regression test for it would not catch the drift.

If that is deliberate (only the overlay topology can advertise a host path), a one-line comment here stating why the gate is `detectLocalUpper` would be enough. Otherwise it may be worth a follow-up issue so remote / composite topologies are tracked explicitly.

agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/FilesystemToolTest.java
```diff
@@ -98,6 +104,31 @@ void listFiles_normalizesWindowsAbsoluteWorkspacePath() {
...
+
+        assertTrue(
+                tool.writeFile(runtimeContext, absolutePath.toString(), "artifact")
+                        .startsWith("Written to "));
+        assertTrue(Files.exists(absolutePath));
+        assertFalse(Files.exists(workspace.resolve("session-1/session-1/artifact.txt")));
+        assertEquals("artifact", tool.readFile(runtimeContext, "artifact.txt", null, null));
+        assertEquals(
+                "artifact", tool.readFile(runtimeContext, absolutePath.toString(), null, null));
+    }
```
oss-maintainer, line 130: **[Info]** Nice that the negative (`session-1/session-1/artifact.txt` must not exist) is asserted. One adjacent case is still unpinned: a model that echoes a *foreign* namespace, e.g. `/workspace/session-2/x.txt` while the active namespace is `session-1`. With the current ordering that path does not match the namespaced prefix, so it strips to `session-2/x.txt`, which the filesystem then re-namespaces to `session-1/session-2/x.txt` — a new directory inside the current session rather than an error.

Not introduced here (it is pre-existing relative-path semantics), and the absolute-path bounds check may already reject it — if so, a one-assert test would make that guarantee explicit and cheap to keep.

oss-maintainer (commented):
### Summary

Re-review after `443c3502` (the only author commit since my last review — the other new commits are upstream resync merges). Thanks for the follow-up: conflict resolution, `@Deprecated` on the unscoped overload, and the mention ping.

What I verified this round:
- **Scope of the new commit**: `git diff 1f5b9b7f..443c3502` touches only `WorkspacePathNormalizer` javadoc + `@Deprecated`. No behaviour change, so the previously reviewed logic (single source of truth via `resolveRuntimeDataPath`, `RuntimeContext` threaded through `norm(...)`) still stands.
- **Migration is complete in main**: `FilesystemTool.norm` and `ArtifactDeliveryTool.norm` are the only production call sites and both pass the `RuntimeContext`, so the deprecated overload has no in-repo users. That is what makes the finding below a hardening note rather than a defect.
- **Gate correctness**: the namespaced strip is only wired in for the `OverlayFilesystem` + `LocalFilesystemWithShell` branch (`HarnessAgent.java:2692`), and the new prompt test asserts both the positive (`session-1` advertised) and the negative (`Additional roots` does not absorb the workspace) — good.
- **CI / CLA / conflicts**: `check`, `check windows build`, `check codeql`, `check license/cla` green on head `443c3502`; `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (blocked on review/merge settings, not on conflicts).

**Verdict: no new blocking issues; nothing here changes my previous assessment.** Not approving because the five threads from my previous review are still open and unresolved — I have not restated them; the three notes below are the *new* ones only, all non-blocking. If you consider the open questions addressed or deferred, resolve the threads (or reply "deferred: …") and @mention me for an approval pass.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspacePathNormalizer.java
ningmao-hlyz: Following up on this point: the one-argument overload is deprecated in 443c3502 to preserve API compatibility. All production callers now pass the RuntimeContext overload, so there is no remaining in-repo call site using the unsafe fallback.

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspacePathNormalizer.java
ningmao-hlyz: Agreed that namespace segments need an explicit validation contract. I am keeping that out of this PR because sanitizing only in WorkspacePathNormalizer could diverge from the namespace used by the filesystem. #2952 handles session IDs in file-name segments, not namespace composition. I will track namespace validation separately.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddleware.java
ningmao-hlyz: This gate is intentional for this PR: only the local overlay with a host shell advertises a host-visible effective workspace, which is the path used by the skill and script workflow addressed here. Plain LocalFilesystem and remote or sandbox topologies use different prompt and path semantics; extending this behavior there should be tracked separately.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddleware.java
ningmao-hlyz: Thanks, agreed this is a reasonable maintainability improvement. Since these are private helpers and the current names distinguish the base and effective roots, I am leaving the signatures unchanged to keep this PR focused.

agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/FilesystemToolTest.java
ningmao-hlyz: Agreed this is worth documenting, but it is pre-existing relative-path behavior rather than a cross-session access path. An unmatched absolute namespace is normalized as a child path under the active namespace, so it cannot address session-2 data. I will keep this PR focused and track an explicit rejection policy separately if needed.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/WorkspaceContextMiddleware.java
ningmao-hlyz: Knowledge is intentionally base-scoped here as shared/static domain context; per-user or per-session overrides are selected by the filesystem overlay. The related shared-knowledge fallback and listing work is tracked in #2687, so I am keeping this PR focused on workspace prompt/path normalization.

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspacePathNormalizer.java
ningmao-hlyz: Thanks for the follow-up. The one-argument overload is deprecated in 443c3502 to preserve source and binary compatibility. All production callers now pass the RuntimeContext overload. Making this overload throw would be an API behavior break, so I would keep the deprecated compatibility path for now and remove it only in a future API cleanup.

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspacePathNormalizer.java
ningmao-hlyz: The implementation already joins an arbitrary number of namespace segments, and the built-in IsolationScope factories currently produce either one segment or an empty namespace. An empty namespace intentionally falls back to the existing unscoped path. I will leave broader hardening tests for a follow-up.

ningmao-hlyz:
@oss-maintainer

oss-maintainer (approved):
### Summary

Re-review on head `11d3a165` in response to the @mention. The only new commits since my last review (`443c3502`) are upstream `main` resync merges — no further changes to the PR's own code, and the merge resolved the previous conflict cleanly.

Verified this round:
- All 8 review threads from previous rounds are resolved.
- The namespaced-strip path in `WorkspacePathNormalizer.normalize(path, rc)` is still correct after the resync: namespaced prefix checked before the unscoped prefix, `RuntimeContext` threaded through `FilesystemTool`/`ArtifactDeliveryTool`, and the deprecated single-arg overload delegates with an empty context only.
- `WorkspaceContextMiddleware` advertises the effective (session-scoped) workspace consistently in both the session-context section and the workspace paragraph; `knowledgeBlock` intentionally keeps the unscoped root.
- Prompt-bound test (`localOverlay_sessionIsolationAdvertisesEffectiveWorkspace`) and round-trip test (`sessionIsolation_absoluteAndRelativeWorkspacePathsResolveToSameFile`) cover both the positive and negative assertions.
- CI: 6/6 checks passing on `11d3a165`; CLA signed; `mergeable: MERGEABLE`.

LGTM — approving.

### Suggestions
- [non-blocking] If a path carries a *different* session's namespace segment (e.g. `/workspace/session-2/x.txt` under a `session-1` context), the unscoped fallback yields `session-2/x.txt`, which the filesystem then resolves under the active namespace. Current callers can't hit this via the prompt, but a defensive note in the javadoc would make the contract explicit.

---

xiaoyuuuuuupeng (approved):
LGTM.本地复测过了

ningmao-hlyz:
@jujn 打扰了，可以看看这个pr嘛，目前处于approve状态

jujn (approved):

jujn:

PR #3024, opened 2026-09-08
feat(tool): add execution-guarded circuit breaker and schema filtering

### LINKED ISSUE #3023: Feature: execution-guarded circuit breaker for AgentTool
### Problem

Per-call retries do not prevent a model from choosing the same unavailable tool again on subsequent reasoning turns. Applications need an opt-in way to stop those calls temporarily, without modifying shared Toolkit registrations.

### Proposed change

Add an `AgentTool` decorator that counts consecutive failures, opens for a bounded exponential cooldown, and admits one recovery probe at execution time. A companion reasoning middleware filters unavailable decorated tools from the model's schema list. Direct or already-generated calls still encounter the execution guard.

The initial implementation would use per-decorator in-memory state, preserve the delegate's schema and permission attributes, and support a custom result failure predicate. Cancellation and non-terminal results should not count as dependency failures. Tests would cover concurrency, stale completions, cooldown recovery, and schema filtering.

This proposal is limited to opt-in tool execution protection; it does not change retry defaults or introduce distributed state.

q141041-dot:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Repeated tool failures can consume subsequent reasoning turns even after per-call retries are exhausted. This adds an opt-in `CircuitBreakerTool` decorator and a companion `CircuitBreakerMiddleware`.

The decorator guards actual execution, counts consecutive failures, applies a capped exponential cooldown, and admits one recovery probe. The middleware filters only the current model request; it never edits Toolkit registrations. Admission happens at subscription time, so a model that ignores an advertised tool cannot leave a probe claim behind. Existing calls may finish, but completions from an older circuit generation cannot reset a newer circuit.

State is local to the decorator instance. There are no new dependencies, background threads, distributed storage, or changes to default retry behavior. The decorator preserves schema and read-only permission attributes. Applications can classify business failures with a predicate; denied, interrupted and suspended results are not treated as dependency failures.

Closes #3023.

## Validation

- `mvn -pl agentscope-core -am test`: 2,334 tests, 0 failures, 0 errors, 9 skips (13 new tests).
- `mvn -pl agentscope-core spotless:check`: passed.
- `git diff --check`: passed.

The full multi-module reactor was not tested locally. Tests use an injected clock and coordinated concurrent subscriptions, without sleeps or external services. The documentation states the scope of in-memory sharing, timeout/retry interaction, and suspended-call limitations.

## Checklist

- [x] Relevant tests pass
- [x] Public API has Javadoc
- [x] Documentation includes a usage example

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/middleware/CircuitBreakerMiddleware.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/middleware/CircuitBreakerMiddleware.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.middleware;
+
+import io.agentscope.core.agent.Agent;
+import io.agentscope.core.agent.RuntimeContext;
+import io.agentscope.core.event.AgentEvent;
+import io.agentscope.core.tool.CircuitBreakerTool;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import reactor.core.publisher.Flux;
+
+/**
+ * Hides unavailable circuit-breaking tools from each reasoning request without mutating Toolkit.
+ *
+ * <p>Register the same decorator instances with the Toolkit and this middleware. Probe admission
+ * happens in the decorator at execution time: advertising a recovered tool does not reserve it or
+ * leave a lease behind when the model chooses a different tool. Concurrent requests can see the
+ * recovered schema, but only one probe can execute.
+ */
+public final class CircuitBreakerMiddleware implements MiddlewareBase {
+    private final Map<String, CircuitBreakerTool> tools;
+
+    /**
+     * @param tools the exact decorated tools registered with the agent's toolkit
+     * @throws IllegalArgumentException if tool names are duplicated
+     */
+    public CircuitBreakerMiddleware(List<CircuitBreakerTool> tools) {
+        Map<String, CircuitBreakerTool> byName = new HashMap<>();
+        for (CircuitBreakerTool tool : tools) {
+            if (byName.putIfAbsent(tool.getName(), tool) != null) {
+                throw new IllegalArgumentException(
+                        "Duplicate circuit breaker tool: " + tool.getName());
+            }
+        }
+        this.tools = Map.copyOf(byName);
+    }
+
+    @Override
+    public Flux<AgentEvent> onReasoning(
+            Agent agent,
+            RuntimeContext ctx,
+            ReasoningInput input,
+            Function<ReasoningInput, Flux<AgentEvent>> next) {
+        return Flux.defer(
+                () -> {
+                    if (input.tools() == null) {
+                        return next.apply(input);
+                    }
+                    return next.apply(
+                            new ReasoningInput(
+                                    input.messages(),
+                                    input.tools().stream()
+                                            .filter(
+                                                    schema ->
+                                                            !tools.containsKey(schema.getName())
+                                                                    || tools.get(schema.getName())
+                                                                            .isAvailable())
+                                            .toList(),
+                                    input.options()));
+                });
+    }
+}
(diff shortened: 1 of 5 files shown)


oss-maintainer:
**CLA Not Signed**

@q141041-dot please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.

---

agentscope-core/src/main/java/io/agentscope/core/tool/CircuitBreakerTool.java
```diff
@@ -0,0 +1,284 @@
...
+        if (!isAvailable()) {
+            return -1;
+        }
+        if (retryAt != null) {
+            probing = true;
+        }
+        return epoch;
+    }
+
+    private synchronized void complete(long admittedEpoch, Boolean failed) {
```
oss-maintainer, line 91: The probe slot is not reserved atomically. `isAvailable()` is only false while a probe is running if the probe already *started* (i.e. `probing` was set by a previous `acquire()`), so N callers that all subscribe inside the same window can each pass the availability check. The first one to fail opens the circuit and bumps `epoch`, and every later probe's outcome is then dropped by the `admittedEpoch != epoch` guard below — including a genuinely successful call, which is discarded and leaves `probing=true` until the first probe settles. The class Javadoc promises "exactly one subscription may probe recovery"; could `acquire()` claim the probe inside the synchronized block (e.g. reject when `retryAt != null && probing`) so only one caller is ever admitted, and so a successful half-open call can always close the circuit?

agentscope-core/src/main/java/io/agentscope/core/tool/CircuitBreakerTool.java
```diff
@@ -0,0 +1,284 @@
...
+        long millis =
+                cooldown == null
+                        ? initialCooldown.toMillis()
+                        : (long)
+                                Math.min(
+                                        maxCooldown.toMillis(),
+                                        Math.ceil(cooldown.toMillis() * multiplier));
+        cooldown = Duration.ofMillis(millis);
+        retryAt = clock.instant().plus(cooldown);
+        probing = false;
```
oss-maintainer, line 127: Worth a Javadoc note on the reset policy: `failures` is zeroed in `open()`, so a half-open probe that fails re-opens the circuit immediately (one failure per window) while `cooldown` keeps growing. That is a reasonable design, but it means a single failing probe after each cooldown extends the outage exponentially, which callers will hit in practice — please make the contract explicit above `open()` and in the docs section.

oss-maintainer (commented):
### Summary
Adds an opt-in in-memory `CircuitBreakerTool` decorator plus a `CircuitBreakerMiddleware` that withholds unavailable tool schemas from the current reasoning request. Well-structured, nicely documented, and the 13 new tests cover the tricky parts (backoff cap, cancellation/empty publisher, stale generations, suspended calls). Verdict: the design looks sound — one concurrency nit in the half-open probe path is worth resolving, the rest is optional polish.

### Findings
- **[Warning]** `agentscope-core/.../tool/CircuitBreakerTool.java:84-91` — the recovery probe is not reserved atomically, so several concurrent subscriptions can all be admitted in the half-open window and all but the first result are dropped by the epoch guard (see inline comment).
- **[Info]** `CircuitBreakerTool.java:118-127` — failing half-open probes re-open immediately with a growing cooldown; worth making that contract explicit in Javadoc/docs.
- **[Info]** `CircuitBreakerMiddleware.java:64` — positional `ReasoningInput` construction is fragile for a core hook; a stale tool call for a withheld schema deserves an explicit test.

### What I liked
- Epoch/generation handling so completions from an older circuit cannot mutate a newer one.
- Settling exactly once per subscription via `AtomicBoolean` + `doFinally`, including cancellation and empty publishers.
- Preserving schema and read-only/permission attributes on the decorator, and never mutating the registered `Toolkit`.

### Note
This PR cannot be approved or merged until the CLA is signed (see the reminder comment above), so this review is posted as a comment only.

---

agentscope-core/src/main/java/io/agentscope/core/middleware/CircuitBreakerMiddleware.java
```diff
@@ -0,0 +1,78 @@
...
+    public Flux<AgentEvent> onReasoning(
+            Agent agent,
+            RuntimeContext ctx,
+            ReasoningInput input,
+            Function<ReasoningInput, Flux<AgentEvent>> next) {
+        return Flux.defer(
+                () -> {
+                    if (input.tools() == null) {
+                        return next.apply(input);
+                    }
```
oss-maintainer, line 64: Two small hardening suggestions for a core hook: (1) `ReasoningInput` is constructed positionally here, so adding a component to the record later breaks this class silently at compile time — consider a `withTools(...)`-style copy helper on the record so filtering stays source-compatible for all middleware implementers. (2) A tool withheld by `isAvailable()` can still be named by an already-emitted assistant message, in which case the call is rejected before reaching the delegate; adding a test for that (withheld schema + stale tool call) would pin down the intended behaviour.

PR #3026, opened 2026-09-08
fix(harness): clarify skill shell path handling

### LINKED ISSUE #2788: ## Code Execution says "alwasy use absolute paths from <File-root>" ,but execute tool's working diretory rejects them
**Code Execution prompt:**

Image

---

**tool define:**

Image

ZHCHJ888:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #2788.

Skill execution guidance previously referred to a non-existent `execute_shell_command` tool and could lead agents to pass an absolute `<files-root>` path as `working_directory`. However, `working_directory` intentionally accepts only workspace-relative paths, so this caused skill script execution to be rejected.

This PR:

- Uses the actual `execute` tool name in the generated skill prompt and documentation.
- Clarifies that absolute skill paths belong in `command`, while `working_directory` must remain workspace-relative or be omitted.
- Rejects POSIX absolute paths, Windows drive-letter paths, UNC/rooted paths, `~`, and `..` in `working_directory`.
- Shares the working-directory requirement across the tool schema, runtime error, and generated skill prompt.
- Adds regression coverage for the generated prompt, exposed tool schema, and cross-platform invalid working-directory forms, including `/d/...`, `C:\...`, `C:/...`, and UNC paths.
- Updates the English and Chinese skill and workspace documentation.

How to test:

```bash
mvn -pl agentscope-harness -am -Dtest=SkillRuntimeTest,ShellExecuteToolTest -Dsurefire.failIfNoSpecifiedTests=false test
mvn -pl agentscope-harness -am -Dtest='io.agentscope.harness.**.*Test' -Dsurefire.failIfNoSpecifiedTests=false clean verify
```

Local results:

- Targeted tests: 41 passed, 0 failed, 0 skipped.
- Clean Harness verification: 1049 tests, 0 failures, 0 errors, 15 environment-dependent tests skipped.
- The skipped tests consist of 5 Windows-disabled shell integration tests, 9 Docker integration tests requiring `-Ddocker.it`, and 1 POSIX-permissions test.

## Checklist


oss-maintainer (commented):
### Summary
Clarifying the `working_directory` contract in three places (tool description, prompt bullets, runtime error) plus a rename of the stale tool name in the harness skill docs — the direction is right, and the added tests pin both the schema description and the prompt wording, which is what makes this kind of guidance change safe to land. Two things worth a second look: the rejection guard still only recognises POSIX-shaped absolute paths (so a Windows path passes and the new message over-promises), and the `execute_shell_command` → `execute` rename stops at `skill.md` while the sibling `workspace.md` harness pages still name the old tool. Non-blocking, so leaving as comments.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
```diff
@@ -66,7 +68,8 @@ public String execute(
             String wd = workingDirectory.strip();
             if (wd.startsWith("/") || wd.startsWith("~") || wd.contains("..")) {
```
oss-maintainer, line 69: The message now promises that "absolute paths, '~', and '..' are not allowed", but the guard only catches POSIX-shaped absolute paths. A Windows-shaped one (`C:\workspace\skills\alpha`, `\\server\share\x`, or `/d/workspace/x`) passes this check and reaches `commandWithWorkingDirectory`, which emits `cd /d "C:\workspace\skills\alpha" && ...` resolved against the sandbox process cwd — exactly the confusing failure this PR is removing, and the repo does build on `windows-latest` and already has a Windows branch here. Could the check also reject a drive-letter / leading-`\` form? Then the new rejection test would have a Windows counterpart.

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillPromptBuilder.java
```diff
@@ -86,8 +86,10 @@ public final class SkillPromptBuilder {
             1. After loading a skill, look at its <files-root> in <available_skills>
             2. List its files:    ls <files-root>/
             3. Run scripts:       python3 <files-root>/scripts/<script-name>
-            4. Always use absolute paths derived from <files-root>; never invent paths
-            5. If a script exists for the task, run it directly — do not rewrite its logic inline
+            4. In the command, always use absolute paths derived from <files-root>; never invent paths
+            5. The working_directory parameter accepts only workspace-relative paths; never pass <files-root> to it
```
oss-maintainer, line 90: This bullet says `working_directory` accepts "only workspace-relative paths", but the tool also rejects a leading `~` and *any* path containing `..`. A model that reads only the prompt will still emit `working_directory="skills/../shared"` and burn a tool call on the error. Since the stated goal of the change is to stop wasted calls, could this line state the same three rejections as the tool message (or just defer to it), so prompt and `@ToolParam` description express one rule instead of two partial ones?

agentscope-harness/src/main/java/io/agentscope/harness/agent/skill/runtime/SkillPromptBuilder.java
```diff
@@ -86,8 +86,10 @@ public final class SkillPromptBuilder {
             1. After loading a skill, look at its <files-root> in <available_skills>
             2. List its files:    ls <files-root>/
             3. Run scripts:       python3 <files-root>/scripts/<script-name>
-            4. Always use absolute paths derived from <files-root>; never invent paths
-            5. If a script exists for the task, run it directly — do not rewrite its logic inline
+            4. In the command, always use absolute paths derived from <files-root>; never invent paths
```
oss-maintainer, line 89: The same policy now lives in four places that must move together: this bullet list, the `working_directory` description in `ShellExecuteTool`, the runtime error string, and `docs/v2/{en,zh}/docs/harness/skill.md`. The two new tests pin the prompt lines and the schema description independently, so any future rewording needs several coordinated edits. Would extracting the rule into one shared constant (consumed by both the prompt and the tool description) work here, or is the duplication deliberate because this prompt text is intentionally frozen?

docs/v2/en/docs/harness/skill.md
```diff
@@ -307,15 +307,15 @@ The agent doesn't see this difference — `load_skill_through_path` always works
 
 ### `<files-root>` and shell execution
 
-When a skill ships scripts (e.g. `scripts/run-checks.sh`), the agent needs an absolute path to invoke them via `execute_shell_command`. That path comes from the `<files-root>` element on each skill entry. Resolution depends on the filesystem mode:
+When a skill ships scripts (e.g. `scripts/run-checks.sh`), the agent needs an absolute path to invoke them with the `execute` tool. That path comes from the `<files-root>` element on each skill entry. Resolution depends on the filesystem mode:
```
oss-maintainer, line 310: Good catch renaming the tool here, but the rename is incomplete in the same doc set: `docs/v2/en/docs/harness/workspace.md:391` and `docs/v2/zh/docs/harness/workspace.md:386` still say the agent invokes skill scripts "via `execute_shell_command`". On the harness path the registered shell tool is `execute` (`ShellExecuteTool.NAME`), so a reader following those harness docs ends up calling a tool that the harness does not register — the same class of mistake this PR exists to prevent. (`execute_shell_command` is still legitimate for users who register core `ShellCommandTool` themselves, and `AgentSkillPromptProvider` keeps that wording for the `SkillBox` path, so a scope note would be enough if that split is intentional.) Could you update the two workspace.md lines here, or file a follow-up so the harness docs name one tool consistently?

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ShellExecuteTool.java
ZHCHJ888: Thanks for catching this. `working_directory` now rejects drive-letter paths (`C:\...` and `C:/...`) plus leading-backslash and UNC paths. I also added parameterized regression cases for those forms and for `/d/...` (covered by the leading-slash branch), so the cross-platform contract is explicit.

oss-maintainer (approved):
### Summary
Re-reviewed after the force-push: both non-blocking points from the previous round are now resolved. The rejection guard no longer under-covers Windows-shaped inputs — `isInvalidWorkingDirectory` adds `\`-prefixed, UNC (`\\server\share`) and drive-letter (`C:` / `C:/`) forms, and the message, the tool schema description and the skill prompt bullets are all assembled from one `WORKING_DIRECTORY_REQUIREMENT` constant, so the three places can no longer drift. The rename also reached `workspace.md` this time, not just `skill.md`. Tests pin the schema description, the prompt wording and the new reject cases including "no command was dispatched" assertions. Approving.

### One nit (non-blocking)
- `workingDirectory.contains("..")` still rejects legitimate relative names such as `a..b` or `skills/run..check.sh`. Pre-existing behaviour, and the safe direction to err, but a segment-wise check (`Path.of(wd).getName(i).equals("..")` style) would keep the guard without the false positives if you ever want it.

---

PR #3027, opened 2026-09-08
test(harness): expand shell pipe deadlock regression coverage

guslegend0510:
Refs #3022
Follow-up to #2839.

The wait-before-drain deadlock reported in #3022 is already fixed on main by #2839. This PR strengthens regression coverage for LocalFilesystemWithShell.execute():

- Exercise more than 1 MiB of stdout, stderr, and alternating writes to both streams.
- Verify complete output capture and successful command completion.
- Verify that both pipes continue draining when output exceeds the capture limit.

Restoring the original wait-before-drain ordering makes all four output scenarios fail with exit code 124, confirming that the tests detect the reported regression.

Validation on macOS / JDK 21:
- LocalFilesystemWithShellTest: 7 tests, no failures or errors.
- Full Core suite: 2321 tests, no failures or errors, 9 skipped.
- Full Harness suite: 990 tests, no failures or errors, 5 skipped.

Windows execution has not been verified locally.


agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShellTest.java
```diff
@@ -46,28 +54,59 @@ void outputCharset_fallsBackToDefaultWhenWindowsNativeEncodingIsUnavailable() {
                 LocalFilesystemWithShell.outputCharset("Windows 10", null));
     }
 
+    @ParameterizedTest
```
oss-maintainer, line 58: Nice coverage upgrade — parameterising over `stdout` / `stderr` / `both` also catches a sequential reader, and dropping the OS-specific shell `for /l` / `while` branching in favour of `largeOutputCommand(...)` with `Locale.ROOT` is a clear improvement over the old single-stream test.

One robustness question: the new cases write ~1 MiB per stream but pass `timeoutSeconds = 10` (the old test used 60). A bash/`cmd` loop of 8192 iterations on a cold Windows CI runner is not that far away from 10s, which would surface as an intermittent exit-124 failure rather than a real regression. Worth considering a more generous budget (e.g. 30-60s) since the point of the assertion is deadlock, not speed.

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShellTest.java
```diff
@@ -46,28 +54,59 @@ void outputCharset_fallsBackToDefaultWhenWindowsNativeEncodingIsUnavailable() {
...
-        String command =
-                windows
-                        ? "for /l %i in (1,1," + lines + ") do @echo " + payload
-                        : "i=0; while [ \"$i\" -lt "
-                                + lines
-                                + " ]; do echo "
-                                + payload
-                                + "; i=$((i+1)); done";
+    void execute_outputBeyondCaptureLimitStillDrainsBothPipes(@TempDir Path tempDir) {
+        LocalFilesystemWithShell fs =
```
oss-maintainer, line 85: `execute_outputBeyondCaptureLimitStillDrainsBothPipes` pins three implementation details of `LocalFilesystemWithShell#execute` at once: that stdout is appended before stderr, that the `\n` separator between streams is not counted towards the cap, and that `maxOutputBytes` truncates by *characters* (`outputStr.length()`), not bytes.

All three are true today and the assertion is exactly right for the current code, so this is not a blocker — but `substring(0, maxOutputBytes)` combined with a raw byte-count message means non-ASCII output is already mis-reported by the production code, and this test would lock the ASCII-only behaviour in. If the intent is to assert the *draining* behaviour (both pipes consumed, no deadlock, `truncated == true`), asserting `resp.truncated()` plus `output.startsWith(STDOUT_LINE.substring(0, 64))` would keep the regression value while leaving room to fix the byte/char inconsistency separately.

oss-maintainer (commented):
### Summary
Test-only hardening of the shell pipe-drain deadlock regression: the single stdout case becomes a `stdout` / `stderr` / `both` parameterised matrix with alternating writes (which also catches a sequential reader), plus a new case asserting both pipes are still drained when the capture cap is hit. Logic verified against `LocalFilesystemWithShell#execute` — the expected `stdout + "\n" + "[stderr] ..."` layout and the exact 128-byte truncation string both match the current implementation, and `@ParameterizedTest` is already used elsewhere in this module.

### Findings
- **[Warning]** `LocalFilesystemWithShellTest.java:82` — the truncation case pins stream ordering, the separator not counting towards the cap, and character-based (not byte-based) truncation; see inline comment for a lower-coupling alternative.
- **[Info]** `LocalFilesystemWithShellTest.java:57` — 10s budget for ~1 MiB per stream through a shell loop is tight on Windows CI; the old case used 60s.

Neither is a blocker; LGTM overall.

---

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShellTest.java
```diff
@@ -17,16 +17,25 @@
...
 import java.nio.file.Path;
+import java.util.Locale;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 class LocalFilesystemWithShellTest {
 
+    private static final int OUTPUT_LINES = 8192;
```
oss-maintainer, line 34: **[Warning]** This scales the payload from ~64 KB to ~1 MB per stream and keeps one 60 s budget for all three parameterisations, but the Windows leg of `largeOutputCommand` is a `cmd.exe` `for /l` loop running two builtins per iteration — that is per-iteration interpreter overhead, not a syscall, and it is the path most likely to eat the timeout on a loaded CI runner. The previous 4000-line/16-char case existed partly because a smaller loop was enough to exceed the ~4 KB Windows pipe buffer.

The deadlock being pinned is "child output exceeds the OS pipe buffer while the parent waits", which 512 lines at 132 chars (~68 KB) already exceeds by a wide margin on both Linux (64 KB) and Windows (4 KB). Suggest scaling `OUTPUT_LINES` per platform (e.g. 8192 on POSIX, ~1024 on Windows) so the Linux job keeps the heavy 1 MiB case and the Windows job stops racing its own timeout. Fine to leave as-is if the Windows leg is already measured green here — a note in the PR with the observed wall time would settle it.

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShellTest.java
```diff
@@ -46,28 +55,60 @@ void outputCharset_fallsBackToDefaultWhenWindowsNativeEncodingIsUnavailable() {
...
 
+    @ParameterizedTest
+    @ValueSource(strings = {"stdout", "stderr", "both"})
+    void execute_outputLargerThanOsPipeBufferCompletesWithoutDeadlock(
+            String streams, @TempDir Path tempDir) {
+        // More than 1 MiB per stream, with room to capture both streams without truncation.
+        // Alternating writes to both pipes also catches sequential stdout/stderr readers.
+        LocalFilesystemWithShell fs =
+                new LocalFilesystemWithShell(
+                        tempDir, false, OUTPUT_TIMEOUT_SECONDS, 4 * 1024 * 1024, null, false);
```
oss-maintainer, line 66: **[Info]** Six positional constructor arguments (`tempDir, false, OUTPUT_TIMEOUT_SECONDS, 4 * 1024 * 1024, null, false`) in a test whose whole point is a specific capture ceiling — the `4 * 1024 * 1024` and the trailing `false`s are unreadable without opening the production signature, and the second test's `128` is the value that actually matters there.

Could a small local factory keep the intent visible? e.g.
```java
private LocalFilesystemWithShell fsWithCapture(Path dir, int maxOutputBytes) {
    return new LocalFilesystemWithShell(dir, false, OUTPUT_TIMEOUT_SECONDS, maxOutputBytes, null, false);
}
```
Then the two cases read as `fsWithCapture(tempDir, 4 * 1024 * 1024)` (must not truncate) and `fsWithCapture(tempDir, 128)` (must truncate). Minor, and the assertion pairing (one truncated=false, one truncated=true against the same command) is a nice way to pin both sides of the ceiling.

oss-maintainer (commented):
### Summary
Re-review of `d6f466cd` since my last pass: the single stdout-only pipe-buffer case becomes a three-way parameterisation (stdout / stderr / both) at ~1 MB per stream with alternating writes, and a second case pins that output beyond the capture ceiling still drains both pipes instead of hanging. Purely test-side; no production behaviour change in this commit.

The added coverage is worth having — a stderr-only producer deadlocking while the parent reads stdout is a distinct failure from the stdout-only case, and the sequential-reader bug it mentions is exactly the kind of regression that silently returns after a refactor of the drainer.

Verified against the implementation, so the exact-equality assertions look sound rather than lucky: the drainers write into two separate buffers and are joined before the combined string is assembled, so `stdout` then `\n[stderr] …` per line is deterministic despite the interleaved writes, and ~2.2 MB total sits under the 4 MB ceiling used in the non-truncating cases.

Two test-quality items inline (per-platform payload sizing for the Windows `for /l` leg, and a local factory to make the capture ceiling visible). Neither blocks.

---

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShellTest.java
```diff
@@ -46,28 +55,60 @@ void outputCharset_fallsBackToDefaultWhenWindowsNativeEncodingIsUnavailable() {
...
+        ExecuteResponse resp =
+                fs.execute(null, largeOutputCommand(streams), OUTPUT_TIMEOUT_SECONDS);
+
+        assertEquals(0, resp.exitCode());
+        assertFalse(resp.truncated());
```
oss-maintainer, line 71: The expected stderr output assumes ExecuteResponse.output() prefixes stderr lines with '[stderr] ' and that stdout entirely precedes stderr in the 'both' case; these implementation details are not visible in the diff and make the regression tests brittle.

oss-maintainer (commented):
### Summary
Re-review of PR #3027 at its current head (automated review of the changes in this diff). 1 item(s) raised below; nothing here blocks the approach itself — details inline.

### Findings
- **[Warning]** `agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShellTest.java:71` — The expected stderr output assumes ExecuteResponse.output() prefixes stderr lines with '[stderr] ' and that stdout entirely precedes stderr in the 'both' case; these implementation details are not visible in the diff and make the regression tests brittle.

### Notes

---

oss-maintainer (commented):
### Summary
Re-review of the new commit `387ac88`: the pipe-drain regression test is now parameterized over stdout / stderr / both, the expected output is built from the real `[stderr] ` prefix and stdout-then-stderr ordering (which I verified against `LocalFilesystemWithShell.execute()`), and a second case covers the beyond-capture-limit path. The earlier brittleness warning is resolved — the expectations now match the implementation instead of guessing at it. Two non-blocking points inline: payload size / hard 60 s wall-clock cap, and the byte-for-byte ordering assertion. CLA signed.

---

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShellTest.java
```diff
@@ -17,16 +17,25 @@
...
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 class LocalFilesystemWithShellTest {
 
+    private static final int OUTPUT_LINES = 8192;
+    private static final int OUTPUT_TIMEOUT_SECONDS = 60;
+    private static final String STDOUT_LINE = "out-" + "0123456789abcdef".repeat(8);
```
oss-maintainer, line 36: `OUTPUT_LINES = 8192` at ~144 bytes per line means each of the three parameterized cases now captures ~1.2 MB per stream (vs. ~64 KB for the test it replaces), and the whole suite relies on a hard 60 s wall-clock cap (`proc.waitFor` returns exit 124 on timeout). On a slow Windows CI runner `cmd.exe /c "for /l %i in (1,1,8192) do @(...)"` spawns two redirections per iteration, so a timeout here would fail as an infrastructure flake rather than as the deadlock regression this test guards. Consider deriving the line count from a target payload size (e.g. keep ~1 MiB total) or lowering `OUTPUT_LINES` for the Windows branch, so the deadlock coverage stays while the runtime stays bounded.

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/local/LocalFilesystemWithShellTest.java
```diff
@@ -46,28 +55,60 @@ void outputCharset_fallsBackToDefaultWhenWindowsNativeEncodingIsUnavailable() {
...
+        assertEquals(0, resp.exitCode());
+        assertFalse(resp.truncated());
+        String stdout = (STDOUT_LINE + "\n").repeat(OUTPUT_LINES);
+        String stderr = ("[stderr] " + STDERR_LINE + "\n").repeat(OUTPUT_LINES).stripTrailing();
+        String expected =
+                switch (streams) {
+                    case "stdout" -> stdout;
+                    case "stderr" -> stderr;
+                    default -> stdout + "\n" + stderr;
+                };
```
oss-maintainer, line 79: The `default ->` case asserts byte-for-byte equality of `stdout + "\n" + stderr`, which encodes three implementation details of `LocalFilesystemWithShell.execute()` at once: the `[stderr] ` line prefix, stdout-before-stderr ordering, and the single separator newline. That is fine as a characterization test as long as those three are intentional contract (they are already asserted by the older `[stderr]` cases elsewhere), but if the prefix/order is ever treated as internal detail this will break loudly for an unrelated reason. A short comment naming them as the asserted contract would keep the intent explicit.

PR #3031, opened 2026-09-08
feat(harness): allow direct artifact delivery without host buffering

### LINKED ISSUE #3030: feat(harness): support sandbox-side artifact delivery without host byte buffering
### Problem

`ArtifactDeliveryTool.deliverArtifact` unconditionally calls `filesystem.downloadFiles(...)` before invoking `ArtifactDeliveryTarget`. `ArtifactDeliveryRequest` carries a `byte[]`, so the entire artifact must pass through the host JVM even when the sandbox could upload it directly to object storage or WebDAV.

For large reports, archives, and videos this adds host heap pressure and a sandbox-to-host transfer. A target cannot opt out because it is called only after the download. This follows up on the delivery tool introduced for #2663.

### Proposed change

Add an opt-in path-based delivery SPI that receives the active filesystem, per-call runtime context, and normalized source/destination metadata before any download. Applications can then implement sandbox-side uploads through their backend SDK or shell, without host byte materialization.

Keep existing byte-based target implementations and builder configuration compatible. Preserve input validation, destination conflict handling, and `force`. A direct-upload failure should be returned, not silently retried through a full in-memory download.

### Acceptance criteria

- Direct delivery never invokes `downloadFiles`, including failure/conflict paths.
- Normalized source path, active filesystem, runtime context, destination name, description, and `force` reach the target.
- Existing byte-based targets retain their behavior.
- Document application responsibility for backend path resolution, upload execution, source checks, and destination conflicts.

q827505931:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

`deliver_artifact` currently downloads every artifact into a host JVM `byte[]` before calling the application target. A sandbox capable of uploading directly must therefore send large files through the host first.

Add an opt-in `DirectArtifactDeliveryTarget` and metadata-only `ArtifactDeliverySource`. The tool dispatches direct targets with the active filesystem and runtime context before any download. Applications can perform the upload inside their sandbox using a backend SDK or shell. Existing byte-based targets, lambdas, and builder configuration remain compatible.

Both paths share filename validation and success/error/conflict rendering. Direct failures never trigger an in-memory fallback. Source checks, backend path resolution (including overlays/routing), destination conflict enforcement, and actual transport remain application responsibilities; this PR does not add a provider-specific uploader. English and Chinese docs include an integration example and explain these boundaries.

Closes #3030

### Validation

- `mvn -pl agentscope-harness -am spotless:apply test -Dtest=ArtifactDeliveryToolTest,HarnessAgentTest -Dsurefire.failIfNoSpecifiedTests=false`
- 67 tests passed (18 artifact delivery tests, 49 HarnessAgent tests), zero failures/errors/skips. The compile phase also runs Spotless check.
- New tests cover direct delivery without filesystem reads, normalized metadata/context propagation, defaults, rejected input, and failure/conflict/null results without fallback. Existing byte-delivery regressions pass.
- `git diff --check` passed.
- No live sandbox/object-store integration test was run; the transport is application-defined.

## Checklist

- [ ] All tests are passing (`mvn test`) — focused reactor tests above passed; full repository suite not run


agentscope-harness/src/main/java/io/agentscope/harness/agent/artifact/DirectArtifactDeliveryTarget.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param runtimeContext per-call runtime, possibly {@code null}; use it for sandbox resolution
+     * @param filesystem the active agent filesystem, potentially an overlay or routed filesystem
+     * @param source validated destination metadata and normalized source path; contains no bytes
+     * @return delivery result, never {@code null}
+     */
+    ArtifactDeliveryResult deliverFromFilesystem(
+            RuntimeContext runtimeContext,
+            AbstractFilesystem filesystem,
+            ArtifactDeliverySource source);
+
```
oss-maintainer, line 55: The `deliver(RuntimeContext, ArtifactDeliveryRequest)` default returns a failure string rather than throwing, so a `DirectArtifactDeliveryTarget` wired into any call site that only knows the base `ArtifactDeliveryTarget` type degrades to a runtime error that is easy to miss in tests. Consider making the dispatch contract explicit at registration time (e.g. log once in `HarnessAgent.Builder#artifactDeliveryTarget` when the target is direct-only), or document that direct targets are supported **only** by `deliver_artifact`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ArtifactDeliveryTool.java
```diff
@@ -122,27 +124,37 @@ public String deliverArtifact(
...
-        if (!response.isSuccess()) {
-            return "Error: failed to read '"
-                    + filePath
-                    + "' from the workspace: "
-                    + response.error();
-        }
+        ArtifactDeliveryResult result;
+        if (target instanceof DirectArtifactDeliveryTarget directTarget) {
+            result =
+                    directTarget.deliverFromFilesystem(
```
oss-maintainer, line 130: Nice dispatch shape. One asymmetry worth documenting: the byte path resolves the file through `filesystem.downloadFiles`, so a missing source surfaces as a tool error, while the direct path trusts the target entirely for existence, `force`/conflict and size limits. Since the agent sees only the returned string, it would help to state in the `@param source` javadoc that `filePath` is not guaranteed to exist.

oss-maintainer (commented):
### Summary
Introduces an opt-in `DirectArtifactDeliveryTarget` SPI plus a metadata-only `ArtifactDeliverySource`, so a sandbox can upload large artifacts without pulling bytes through the host JVM. Clean, backwards-compatible design: the existing byte path is untouched, the tool keeps `result == null` handling, and `no fallback on failure` is the right default. Tests cover the no-download guarantee, defaults, and the invalid-input short-circuit. Left two non-blocking notes on SPI discoverability and the source-existence contract.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/artifact/DirectArtifactDeliveryTarget.java
q827505931: Addressed in ce02173a. The SPI and builder Javadocs now explicitly state that direct targets are supported only by the registered deliver_artifact tool because the inherited byte-based entry point has no source filesystem. I also added a regression test asserting that accidental use of the base entry point returns the documented failure.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/ArtifactDeliveryTool.java
q827505931: Addressed in ce02173a. The source parameter Javadoc now says that filePath is not guaranteed to exist and that the implementation must check existence and size limits and enforce overwrite/conflict behavior.

q827505931:
Addressed the two review notes in 399986ca: direct targets are now explicitly documented as supported only through the registered `deliver_artifact` tool, and the source contract now states that `filePath` is not guaranteed to exist and that implementations must validate existence, size limits, and overwrite/conflict behavior. I also added a regression test for accidental use of the inherited byte-based entry point. The focused suite passes with 68 tests.

oss-maintainer (approved):
### Summary
Re-review after the follow-up commits — both points from the previous pass are resolved, and resolved in the way that fits an SPI best: the contract is now written down instead of being left to inference. `DirectArtifactDeliveryTarget`'s type javadoc states that direct targets are supported only by the framework's `deliver_artifact` tool and that the inherited byte-based `deliver` returns a failure because it has no source filesystem, the `@param source` javadoc says the normalized path is not guaranteed to exist and puts existence/size/`force`-conflict enforcement on the implementation, `HarnessAgent.Builder#artifactDeliveryTarget` repeats the restriction where a target is wired, and both the EN and ZH `sandbox.md` sections carry the same caveats (no automatic byte fallback after a failure, normalized path ≠ sandbox-native path, quote paths for shell uploads, keep credentials out of tool output). The added `directTarget_byteBasedEntryPointReturnsFailure` test pins the documented behavior rather than just the happy path.

### Findings
None blocking. Nothing in the new delta changes the dispatch logic itself, so the previously-reviewed behavior (instanceof dispatch, no fallback, validation before either branch) stands as reviewed.

### Suggestions
Two optional follow-ups, neither worth holding the PR for:
- If the direct path turns out to need a size cap in practice, the framework has no lever today since it never sees the bytes — a `long maxBytesHint()` on the SPI (defaulting to unlimited, documented as advisory) would give a later enforcement point without breaking this contract.
- A `//noinspection`-style pointer from `ArtifactDeliveryTarget`'s javadoc to `DirectArtifactDeliveryTarget` would make the two implementations discoverable from either side; right now only the direct interface cross-references both.

CI is green on the current head and the CLA is signed. Thanks for the quick, thorough turnaround on the feedback, and for keeping the docs and ZH/EN guides in step with the code.

---

q827505931:
All review feedback has been addressed, the PR is approved, and I manually re-ran the CI checks to confirm the current head. Could a maintainer please confirm whether this PR is ready to merge, or let me know if there are any remaining concerns? Thanks!

PR #3038, opened 2026-09-08
feat(anthropic): support bearer token authentication

### LINKED ISSUE #3029: [Feature]: AnthropicChatModel.Builder 缺少 authToken() 方法，无法支持自定义 Bearer Token 鉴权
问题描述

  当前 AnthropicChatModel.Builder 只暴露了 .apiKey() 方法，但没有暴露 .authToken() 方法。底层 Anthropic Java
  SDK（AnthropicOkHttpClient.Builder）实际上同时支持两种鉴权方式：

  // AnthropicOkHttpClient.Builder (SDK 原生)
  fun apiKey(apiKey: String?)       // 设置 X-Api-Key header
  fun authToken(authToken: String?) // 设置 Authorization: Bearer <token> header

  但 AgentScope 的 AnthropicChatModel 只转发了 apiKey，没有转发 authToken，也没有调用 .fromEnv() 来读取环境变量。

  影响

  对于使用企业 AI 网关（非 Anthropic 官方地址）的场景，通常需要通过 Authorization: Bearer <token> 鉴权，而不是 Anthropic SDK 默认的 X-Api-Key。目前 AgentScope
  无法在代码层面传递这个 token。

  原因分析

  尝试过用 GenerateOptions.additionalHeaders 在请求级别添加 Authorization header，但 Anthropic SDK 的 AnthropicBackend.authorizeRequest()
  方法会在发请求前检查，如果请求上已经存在 auth header 就抛出异常：

  // AnthropicBackend.kt
  override fun authorizeRequest(request: HttpRequest): HttpRequest {
      require(
          !request.headers.names().contains(HEADER_API_KEY) &&
              !request.headers.names().contains(HEADER_AUTHORIZATION)
      ) { "Request already authorized for Anthropic." }

      // SDK 想自己加 auth header
      return request.toBuilder().apply {
          apiKey?.let { putHeader(HEADER_API_KEY, it) }
          authToken?.let { putHeader(HEADER_AUTHORIZATION, "Bearer $it") }
      }.build()
  }

  所以 additionalHeaders 这条路走不通，必须由 SDK 的 client 级别来设置 authToken。

  建议修复

  在 AnthropicChatModel.Builder 中增加 .authToken(String) 方法，并转发给 SDK：

  public Builder authToken(String authToken) {
      this.authToken = authToken;
      return this;
  }

  构造函数中追加：

  if (authToken != null) {
      clientBuilder.authToken(authToken);
  }

  修复后业务代码可以这样使用：

  AnthropicChatModel.builder()
          .baseUrl("https://gateway.example.com/anthropic/v1")
          .authToken("my-api-token")  // → Authorization: Bearer my-api-token
          .modelName("qwen3.7-plus")
          .build()

qiyu-lu:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #3029.

Anthropic 兼容网关需要 `Authorization: Bearer <token>` 时，现有模型无法通过 Builder 配置该凭据。本 PR 增加 SDK 客户端级别的 Bearer Token 支持，并保留原有 API Key 用法和公开构造方法。

- 增加 `AnthropicChatModel.Builder.authToken` 和 Spring Boot 配置 `agentscope.anthropic.auth-token`，由底层 SDK 设置认证头。
- API Key 和 Bearer Token 互斥：同时配置时，在创建 SDK 客户端前抛出 `IllegalArgumentException`，异常只包含配置项名称，不包含凭据值。校验覆盖 Builder 和直接构造入口。
- Spring Starter 保留空白值处理及 customizer 覆盖顺序；customizer 可以在校验前通过 `apiKey(null)` 或 `authToken(null)` 清除其中一种凭据。
- 更新中英文文档，并补充认证请求、冲突拒绝和凭据不出现在 `toString()` / 校验异常中的回归测试。

## Validation

- 基于 main `c5db8f72` 完成本地变基，并解决两处文档链接冲突。
- `mvn -B -pl agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic,agentscope-extensions/agentscope-spring-boot-starters/agentscope-anthropic-spring-boot-starter -am '-Dtest=Anthropic*Test' -Dsurefire.failIfNoSpecifiedTests=false spotless:apply test`：198 个用例，0 failures、0 errors、1 skipped。
- 上述模块及其 reactor 依赖的 `mvn -o -B ... -am spotless:check`：通过。
- `git diff --check`：通过。

HTTP 测试使用真实 Anthropic SDK 请求本地 MockWebServer，覆盖流式／非流式调用、Token 单独配置、API Key 单独配置、旧构造方法，以及 Starter 的空白值和 customizer 覆盖行为。新增测试覆盖双凭据拒绝、customizer 消除冲突，以及模型、Builder、配置对象和校验异常的凭据保护。未执行真实企业网关／付费模型联调或全仓测试；这些检查不代表所有 SDK 异常和日志路径均已验证。

### DIFF
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
@@ -96,6 +96,44 @@ public class AnthropicChatModel extends ChatModelBase {
             AnthropicBaseFormatter formatter,
             ProxyConfig proxyConfig,
             String cacheTtl) {
+        this(
+                baseUrl,
+                apiKey,
+                null,
+                modelName,
+                streamEnabled,
+                defaultOptions,
+                formatter,
+                proxyConfig,
+                cacheTtl);
+    }
+
+    /**
+     * Creates an Anthropic chat model with optional bearer token authentication.
+     *
+     * <p>When both {@code apiKey} and {@code authToken} are configured, the SDK sends both
+     * {@code X-Api-Key} and {@code Authorization} headers.
+     *
+     * @param baseUrl the base URL for the Anthropic API (null for default)
+     * @param apiKey the API key for authentication (null to omit)
+     * @param authToken the bearer token without the {@code Bearer } prefix (null to omit)
+     * @param modelName the model name to use
+     * @param streamEnabled whether streaming should be enabled
+     * @param defaultOptions default generation options
+     * @param formatter the message formatter to use (null for the default formatter)
+     * @param proxyConfig the proxy configuration (null for no proxy)
+     * @param cacheTtl the TTL for prompt-caching markers (null for default 5m)
+     */
+    public AnthropicChatModel(
+            String baseUrl,
+            String apiKey,
+            String authToken,
+            String modelName,
+            boolean streamEnabled,
+            GenerateOptions defaultOptions,
+            AnthropicBaseFormatter formatter,
+            ProxyConfig proxyConfig,
+            String cacheTtl) {
         this.baseUrl = baseUrl;
         this.apiKey = apiKey;
         this.modelName = modelName;
@@ -110,6 +148,10 @@ public class AnthropicChatModel extends ChatModelBase {
 
         if (apiKey != null) {
             clientBuilder.apiKey(apiKey);
+        }
+
+        if (authToken != null) {
+            clientBuilder.authToken(authToken);
         }
 
         if (baseUrl != null) {
@@ -282,6 +324,7 @@ public class AnthropicChatModel extends ChatModelBase {
     public static class Builder {
         private String baseUrl;
         private String apiKey;
+        private String authToken;
         private String modelName = "claude-sonnet-4-5-20250929";
         private boolean streamEnabled = true;
         private GenerateOptions defaultOptions;
@@ -309,6 +352,20 @@ public class AnthropicChatModel extends ChatModelBase {
          */
         public Builder apiKey(String apiKey) {
             this.apiKey = apiKey;
+            return this;
+        }
+
+        /**
+         * Sets the bearer token for authentication with an Anthropic-compatible gateway.
+         *
+         * <p>The SDK adds the {@code Bearer } prefix to the {@code Authorization} header. If an
+         * API key is also configured, the SDK sends both authentication headers.
+         *
+         * @param authToken the token without the {@code Bearer } prefix (null to omit)
+         * @return this builder
+         */
+        public Builder authToken(String authToken) {
+            this.authToken = authToken;
             return this;
         }
 
@@ -397,6 +454,7 @@ public class AnthropicChatModel extends ChatModelBase {
                     new AnthropicChatModel(
                             baseUrl,
                             apiKey,
+                            authToken,
                             modelName,
                             streamEnabled,
                             defaultOptions,
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
@@ -111,8 +111,7 @@ public class AnthropicChatModel extends ChatModelBase {
     /**
      * Creates an Anthropic chat model with optional bearer token authentication.
      *
-     * <p>When both {@code apiKey} and {@code authToken} are configured, the SDK sends both
-     * {@code X-Api-Key} and {@code Authorization} headers.
+     * <p>{@code apiKey} and {@code authToken} are mutually exclusive.
      *
      * @param baseUrl the base URL for the Anthropic API (null for default)
      * @param apiKey the API key for authentication (null to omit)
@@ -123,6 +122,7 @@ public class AnthropicChatModel extends ChatModelBase {
      * @param formatter the message formatter to use (null for the default formatter)
      * @param proxyConfig the proxy configuration (null for no proxy)
      * @param cacheTtl the TTL for prompt-caching markers (null for default 5m)
+     * @throws IllegalArgumentException if both API key and bearer token are configured
      */
     public AnthropicChatModel(
             String baseUrl,
@@ -134,6 +134,10 @@ public class AnthropicChatModel extends ChatModelBase {
             AnthropicBaseFormatter formatter,
             ProxyConfig proxyConfig,
             String cacheTtl) {
+        if (apiKey != null && authToken != null) {
+            throw new IllegalArgumentException(
+                    "apiKey and authToken are mutually exclusive; configure only one credential");
+        }
         this.baseUrl = baseUrl;
         this.apiKey = apiKey;
         this.modelName = modelName;
@@ -358,8 +362,8 @@ public class AnthropicChatModel extends ChatModelBase {
         /**
          * Sets the bearer token for authentication with an Anthropic-compatible gateway.
          *
-         * <p>The SDK adds the {@code Bearer } prefix to the {@code Authorization} header. If an
-         * API key is also configured, the SDK sends both authentication headers.
+         * <p>The SDK adds the {@code Bearer } prefix to the {@code Authorization} header.
+         * Configuring both an API key and a bearer token causes model construction to fail.
          *
          * @param authToken the token without the {@code Bearer } prefix (null to omit)
          * @return this builder
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
@@ -72,21 +72,6 @@ public class AnthropicChatModel extends ChatModelBase {
     private final GenerateOptions defaultOptions;
     private final AnthropicBaseFormatter formatter;
 
-    /**
-     * Creates a new Anthropic chat model instance.
-     *
-     * @param baseUrl        the base URL for Anthropic API (null for default)
-     * @param apiKey         the API key for authentication (null to load from
-     *                       ANTHROPIC_API_KEY env var)
-     * @param modelName      the model name to use (e.g.,
-     *                       "claude-sonnet-4-5-20250929")
-     * @param streamEnabled  whether streaming should be enabled
-     * @param defaultOptions default generation options
-     * @param formatter      the message formatter to use (null for default
-     *                       Anthropic formatter)
-     * @param proxyConfig    the proxy configuration (null for no proxy)
-     * @param cacheTtl       the TTL for prompt-caching markers (null for default 5m)
-     */
     public AnthropicChatModel(
             String baseUrl,
             String apiKey,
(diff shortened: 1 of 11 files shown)


oss-maintainer:

@qiyu-lu This PR currently conflicts with `main` and cannot be merged. Please rebase or merge `main` into your branch:

```bash
git fetch origin
git checkout feat/anthropic-bearer-auth
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

This is a one-time reminder. Feel free to @mention me for a re-review once conflicts are resolved.

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
```diff
@@ -312,6 +355,20 @@ public Builder apiKey(String apiKey) {
             return this;
         }
 
+        /**
+         * Sets the bearer token for authentication with an Anthropic-compatible gateway.
+         *
```
oss-maintainer, line 360: The javadoc says that if both `apiKey` and `authToken` are set the SDK sends both headers, and the builder just forwards them. Silently sending two conflicting credentials is the kind of thing that produces a confusing 401 from the gateway; for an Anthropic-compatible proxy with a static `Authorization` header this is also a footgun in the Spring starter wiring. Consider rejecting the combination in `build()` (or ignoring `apiKey` when `authToken` is present and logging a warning) so the effective credential is unambiguous.

oss-maintainer (commented):
### Summary
Adds bearer-token auth for Anthropic-compatible gateways on top of the existing API-key path, threaded through the builder and the Spring auto-configuration, with MockWebServer assertions on the actual `Authorization` header. That is the right way to test this. Two things before this can move: the PR is in conflict with `main`, and the both-credentials-configured case currently sends two conflicting auth headers silently.

---

- **[INFO]** `agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java:1` — Two follow-ups: (1) this PR currently conflicts with `main` — see the reminder comment, please rebase; (2) please confirm the bearer token never reaches `toString()`/`equals()`/logging on the model or its config object, per the project rule that credentials must not leak into logs or events. The MockWebServer assertion in the new test is good evidence for the header itself; a `toString()` redaction check would close the loop. _(line outside diff)_

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
```diff
@@ -96,6 +96,48 @@ public AnthropicChatModel(
...
+            String baseUrl,
+            String apiKey,
+            String authToken,
+            String modelName,
+            boolean streamEnabled,
+            GenerateOptions defaultOptions,
+            AnthropicBaseFormatter formatter,
+            ProxyConfig proxyConfig,
+            String cacheTtl) {
+        if (apiKey != null && authToken != null) {
```
oss-maintainer, line 122: The guard is null-based, but `Builder.apiKey("")` / `authToken("   ")` are stored verbatim, so a blank credential slips through: `apiKey=""` + `authToken="x"` reaches the SDK with both headers, and `authToken=""` sends a malformed `Authorization: Bearer `. The starter trims blanks (`trimToNull`) before calling the builder, but this class is public API, so the same policy should be enforced here too — please normalize blank → null in the setters/constructor and only reject when two *present* credentials conflict, so the documented guarantee holds for direct builder users as well.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
```diff
@@ -112,6 +154,10 @@ public AnthropicChatModel(
             clientBuilder.apiKey(apiKey);
         }
 
+        if (authToken != null) {
```
oss-maintainer, line 142: Feature-scope gap: the SPI / model-registry path (`AnthropicModelProvider.create()` → `firstNonBlank(context.getApiKey(), ANTHROPIC_API_KEY)`) still has no way to pass a bearer token — `ModelCreationContext` exposes only `apiKey`/`baseUrl`/`options`, and `AnthropicCredential` requires a non-null `apiKey`. So gateway users who configure models through the registry (non-Spring) get no `authToken`, while the docs page added here presents bearer auth as a general model capability. Could you either extend provider + credential + context with `auth_token`, or state in the doc that bearer auth currently covers the builder and Spring starter paths only?

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
```diff
@@ -312,6 +359,20 @@ public Builder apiKey(String apiKey) {
...
+        /**
+         * Sets the bearer token for authentication with an Anthropic-compatible gateway.
+         *
+         * <p>The SDK adds the {@code Bearer } prefix to the {@code Authorization} header.
+         * Configuring both an API key and a bearer token causes model construction to fail.
+         *
+         * @param authToken the token without the {@code Bearer } prefix (null to omit)
+         * @return this builder
+         */
+        public Builder authToken(String authToken) {
```
oss-maintainer, line 356: `shouldNotExposeBearerTokenInModelOrBuilderToString` passes only because neither `Builder` nor `AnthropicChatModel` overrides `toString()` today. That makes the guarantee accidental — a future `@ToString`/`record` refactor on the builder would silently start printing `authToken` (and `apiKey`) into logs. Since this PR is exactly about credential hygiene, it would be nice to add an explicit redacting `toString()` here (e.g. print `authToken=***`) so the test pins real behaviour.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
```diff
@@ -96,6 +96,48 @@ public AnthropicChatModel(
...
+                cacheTtl);
+    }
+
+    /**
+     * Creates an Anthropic chat model with optional bearer token authentication.
+     *
+     * <p>{@code apiKey} and {@code authToken} are mutually exclusive.
+     *
+     * @param baseUrl the base URL for the Anthropic API (null for default)
+     * @param apiKey the API key for authentication (null to omit)
```
oss-maintainer, line 102: Minor javadoc drift: the pre-existing constructor documents `apiKey` as "null to load from ANTHROPIC_API_KEY env var", while this one says "null to omit", and the new `@throws` line doesn't mention that the two-credential check is what changed. Aligning the wording (and noting explicitly that env-var fallback happens in `AnthropicModelProvider`, not in this constructor) would keep the two overloads from reading as different contracts.

oss-maintainer (approved):
### Summary
Re-reviewed after conflict resolution and the follow-up commits. The dual-credential concern from my earlier review is now fixed at the API level (`apiKey`/`authToken` are mutually exclusive and the message never echoes the secret values), the Spring starter trims blanks before validation, and the new MockWebServer tests assert the actual `X-Api-Key` / `Authorization` headers on both the streaming and non-streaming paths. CI is green and the CLA is signed — LGTM, with two non-blocking points below about places where the new guarantee is not yet enforced (blank strings via the public builder, and the SPI/registry path that still cannot express a bearer token).

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicChatModel.java
```diff
@@ -72,30 +72,57 @@ public class AnthropicChatModel extends ChatModelBase {
     private final GenerateOptions defaultOptions;
     private final AnthropicBaseFormatter formatter;
 
+    public AnthropicChatModel(
```
oss-maintainer, line 75: **[Info]** This commit dropped the Javadoc from the 8-arg convenience constructor. It is still a public constructor and it now silently encodes a contract ("delegates with `authToken = null`", i.e. API-key / `X-Api-Key` auth). A one-line comment plus `@param`/`@see` to the 9-arg overload would keep that visible to callers who pick it from IDE completion; the module publishes javadoc for these public types.

oss-maintainer (commented):
### Summary

Re-review after `6bf1a3a4`, which changed credential resolution *after* my previous approval, so the approval no longer covers the current head. The new precedence logic reads correctly: explicit `apiKey` / `authToken` from the context win, environment variables are consulted only when neither explicit credential is present, and `AnthropicChatModel` still rejects the both-set case without echoing the values. Two follow-ups below, plus a CI note that is not your fault.

### Findings
- **[Warning]** `agentscope-extensions/.../anthropic/AnthropicModelProvider.java:71` — the new "explicit credential beats environment" precedence (and the `ANTHROPIC_API_KEY` > `ANTHROPIC_AUTH_TOKEN` order) has no test, and it is exactly the rule a later refactor is likely to invert.
- **[Info]** `agentscope-extensions/.../anthropic/AnthropicChatModel.java:75` — the 8-arg public constructor lost its Javadoc in this commit; it now silently means "API-key auth, `authToken = null`".

### CI
`build (ubuntu-latest)` fails on this head in `agentscope-harness` → `AguiPermissionResumeTest.approvedToolResultSurvivesNewStreamContext`, which none of this PR's files touch, and the same test is red on #3108 as well. That points at shared state / ordering in the harness test suite rather than at this change. Please re-run once it is confirmed unrelated; I am keeping this as a comment rather than an approval until the head is green.

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/AnthropicModelProvider.java
```diff
@@ -55,10 +63,22 @@ public Model create(String modelId, ModelCreationContext context) {
             throw new IllegalArgumentException("Unsupported Anthropic model id: " + modelId);
         }
         String modelName = modelId.substring(PREFIX.length());
-        String apiKey = firstNonBlank(context.getApiKey(), System.getenv("ANTHROPIC_API_KEY"));
+        String apiKey = trimToNull(context.getApiKey());
+        String authToken = stringOption(context, OPTION_AUTH_TOKEN);
+        if (apiKey == null && authToken == null) {
+            // No explicit credential: fall back to the environment, keeping the historical
+            // precedence of ANTHROPIC_API_KEY over ANTHROPIC_AUTH_TOKEN.
+            apiKey = trimToNull(System.getenv("ANTHROPIC_API_KEY"));
```
oss-maintainer, line 71: The env fallback is only consulted when *neither* explicit credential is present, so an explicit `authToken` context option correctly wins over a stale `ANTHROPIC_API_KEY` in the environment. That precedence is the part easy to break in a later refactor and there is no test pinning it: `AnthropicModelProviderTest` covers explicit apiKey + explicit authToken, non-string authToken, and the bearer round-trip, but not "authToken option set while `ANTHROPIC_API_KEY` is set in the environment" (nor the reverse). Since the environment is process-wide state, consider extracting the resolution into a package-private helper that takes the two env values as parameters and unit-testing the matrix there, instead of relying on `System.getenv` in the test.

PR #3039, opened 2026-09-08
fix(sandbox): avoid session mirrors after sandbox release

### LINKED ISSUE #3036: [Bug]: SessionTree mirror NPEs when it outlives the sandbox — PinnedSandboxFilesystem cannot know the pinned sandbox is closed (regression from #2946, v2.0.3)
**Versions:** `agentscope-harness:2.0.3`, `agentscope-extensions-sandbox-kubernetes:2.0.3`
**Regressed by:** #2946 (merged 2026-09-06, released in v2.0.3 on 2026-09-07)
**Same root cause as:** #2147 (still open) — different teardown point
**Context:** #2490 / #2675 (per-call sandbox binding), #2771 (per-call K8s pod termination), #2777 / #2935 (fire-and-forget flush)

### Summary

`SessionTree` mirrors the session log to the agent's filesystem fire-and-forget. When that filesystem is a sandbox, the queued mirror routinely runs *after* the sandbox has been shut down and dies with a `NullPointerException` rather than any meaningful error.

Before #2946 this same race degraded to a `No active sandbox` warning, because the async mirror resolved the sandbox lazily through the shared slot, which release had already cleared. #2946 pinned the `Sandbox` instance to fix *which* sandbox the mirror targets — correct in itself — but the pinned reference is now a hard one, so the mirror walks into a dead sandbox instead of failing soft.

### Observed

```
WARN session-tree-mirror i.a.h.a.filesystem.sandbox.SandboxBackedFilesystem
  [sandbox-fs] native upload failed for path: agents/<agent>/sessions/<sessionId>.log.jsonl
java.lang.NullPointerException: Cannot invoke
  "io.agentscope.extensions.sandbox.kubernetes.client.CommandExecutor.run(String)"
  because the return value of
  "io.agentscope.extensions.sandbox.kubernetes.client.Sandbox.commands()" is null
    at io.agentscope.extensions.sandbox.kubernetes.KubernetesSandbox.uploadFile(KubernetesSandbox.java:272)
    at io.agentscope.harness.agent.filesystem.sandbox.SandboxBackedFilesystem.uploadFiles(...)
    at io.agentscope.harness.agent.memory.session.SessionTree.mirrorToFilesystem(SessionTree.java:733)
    at io.agentscope.harness.agent.memory.session.SessionTree.lambda$scheduleMirror$6(SessionTree.java:593)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
```

One run, three log lines 78 ms apart: the trace middleware logs POST_CALL, then `[sandbox-client] Connection to sandbox claim '<claim>' has been closed.`, then the mirror fails.

> The `SandboxBackedFilesystem` frame's line number is omitted because we shadow that class locally for an unrelated reason. The frame is the unmodified native-transfer branch of `uploadFiles` that calls `SandboxFileTransfer.uploadFile`.

### Root cause

1. `SessionTree.flush()` → `scheduleMirror()` wraps the filesystem via `pinIfSandbox(...)` and hands the work to the static single-threaded `MIRROR_EXECUTOR` (thread `session-tree-mirror`). `flush()` returns as soon as the task is queued.
2. `PinnedSandboxFilesystem` pins the concrete `Sandbox` in its constructor (`setSandbox(sandbox)`) **and overrides `clearSandboxIfCurrent` to a no-op**. The override is deliberate — it stops release from un-pinning the instance — but `clearSandboxIfCurrent` is also the only channel through which this object could learn the sandbox was released (#2675). It therefore holds a hard reference to a sandbox it has no way of knowing is dead.
3. `SandboxLifecycleMiddleware.releaseForCall` → `SandboxManager.release` calls `stop()` then `shutdown()` for every sandbox with `isSelfManaged() == true`, at the end of *every* call.
4. `KubernetesSandbox.shutdown()` reaches the SDK's `Sandbox.closeConnection()`, which nulls `commands` and `files` and sets `closed = true`.
5. The queued mirror then calls `KubernetesSandbox.uploadFile`, which does `sdkSandbox.commands().run("mkdir -p ...")` with no liveness check → NPE.

Steps 1 and 3 are not synchronised in any way.

### This is not limited to user-managed sandboxes

#2771 documents the same step 3 for the stock client: *"`SandboxLifecycleMiddleware` … releases the sandbox after **every** call: `SandboxManager.release()` (self-managed path) → `stop()` → `shutdown()`. `KubernetesSandbox.shutdown()` sees `claimOwned == true` and calls `terminate()`, deleting the claim and the pod — every call."* #2771 was closed on the acquire/resume side; `release` still shuts the sandbox down per call in 2.0.3. So a plain framework-managed Kubernetes sandbox hits this race too.

### Why the existing quiescence API doesn't cover it

`SessionTree.awaitMirrorQuiescence` exists, but across the whole harness jar it is referenced only by `SessionTree` itself and `HarnessAgent.close()` — i.e. container shutdown. Nothing on the sandbox lifecycle path drains the queue, and there is no way to wait for one session's mirror; the executor is static and global, so the only available barrier is "wait for every session".

That is consistent with how the API arrived: #2777 made the flush fire-and-forget on purpose, to unblock conversation completion, and #2935 describes `awaitMirrorQuiescence` as something `HarnessAgent#close()` already did, reusing it test-side to stop `@TempDir` teardown flakiness. It was never introduced as a lifecycle contract for resource owners.

**#2147 is the same root cause at a different teardown point** and has been open since 2026-07-12: there the late `session-tree-mirror` write lands on a closed Jedis pool (`Pool not open`) after `HarnessAgent.close()` returned. Its "Expected" list — stop accepting new mirror jobs on close, drain or cancel accepted ones, close the executor before returning — is exactly what the sandbox path needs as well. Filing separately because the fix location differs (pinned filesystem / sandbox release rather than agent close), but the two should probably be solved together.

### Secondary defects

- **`KubernetesSandbox.uploadFile` / `downloadFile` don't check liveness.** The SDK's `Sandbox` already exposes `isActive()`. A closed sandbox should produce a `SandboxException` naming the cause, not an NPE from dereferencing a nulled field.
- **`isRunning()` lies after shutdown.** `KubernetesSandbox.shutdown()` overrides the interface default and only calls `terminate()` / `closeConnection()`; it never clears `AbstractBaseSandbox`'s `running` flag, so `Sandbox.isRunning()` still returns `true` on a shut-down sandbox. Callers wanting to guard defensively have no reliable check on the `Sandbox` interface and must downcast to `KubernetesSandbox` for `getSdkSandbox().isActive()`.

### Suggested fixes, roughly in order of value

1. Have `PinnedSandboxFilesystem` (or `SessionTree.mirrorToFilesystem`) skip the write and log at debug when the pinned sandbox is no longer alive — the mirror is best-effort by design.
2. Make `KubernetesSandbox.uploadFile` / `downloadFile` fail with a `SandboxException` on a closed sandbox instead of an NPE.
3. Clear the `running` flag in `KubernetesSandbox.shutdown()` so `isRunning()` is trustworthy.
4. Drain the mirror queue in `SandboxManager.release` before `shutdown()` for self-managed sandboxes, or expose a per-session `awaitMirror(sessionId, timeout)` so lifecycle owners can order the two correctly. This would also address #2147.

`SessionTreeMirrorTest`, added by #2946, is the natural home for a regression test: schedule a mirror against a sandbox that is shut down before the executor picks the task up.

### Workaround

We own our sandbox lifecycle (`SandboxContext.externalSandbox`), so we call `SessionTree.awaitMirrorQuiescence(5, TimeUnit.SECONDS)` immediately before shutting the sandbox down. It covers the normal path deterministically — `TranscriptMiddleware` triggers the flush inside the middleware chain, which completes before the `Flux.using` cleanup that releases the sandbox — but it relies on internal ordering rather than a contract, it blocks on unrelated sessions queued on the shared executor, and it isn't available at all to anyone using a framework-managed sandbox, since that shutdown happens inside `SandboxManager.release`.

larry-zy:
## Summary
- skip best-effort SessionTree mirrors when their pinned sandbox has been released
- report a clear sandbox exception when Kubernetes file transfer uses a closed connection
- add regression coverage for both cases

## Testing
- mvn -pl agentscope-harness -am -DskipITs -Dtest=SessionTreeMirrorTest -Dsurefire.failIfNoSpecifiedTests=false test
- mvn -pl agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes -am -DskipTests package

Fixes #3036


agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystem.java
```diff
@@ -28,18 +36,85 @@
...
+            return !mirrorGate.released && pinnedSandbox.isRunning();
+        } finally {
+            mirrorGate.lock.readLock().unlock();
+        }
+    }
+
+    @Override
+    public List<FileUploadResponse> uploadFiles(
+            RuntimeContext runtimeContext, List<Map.Entry<String, byte[]>> files) {
+        mirrorGate.lock.readLock().lock();
```
oss-maintainer, line 109: **[Critical]** `uploadFiles` holds the read lock across the entire remote transfer (line 81 → 92) while `markReleased()` takes the write lock — so `markMirrorSandboxReleased(result)` on the teardown path now *blocks until every in-flight mirror upload returns*. `releaseForCall` is invoked from `ReActAgent.afterAgentExecution()`, i.e. in-band with the agent's completion path, and neither `super.uploadFiles` nor the file API below it has a timeout here. One wedged upload (stalled file API, pod being evicted) therefore turns "skip a best-effort mirror" into "hang the call release and the lease close", which is a worse failure mode than the race being fixed.

Could the exclusive acquisition be bounded? e.g. keep the gate semantics but don't wait forever:

```java
private void markReleased() {
    if (lock.writeLock().tryLock(RELEASE_GATE_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS)) {
        try {
            released = true;
        } finally {
            lock.writeLock().unlock();
        }
    } else {
        released = true;   // needs volatile: stop new mirrors, don't wait on the in-flight one
        log.warn("Mirror upload still in flight for {}; releasing sandbox without waiting", sandboxId);
    }
}
```

Marking the gate is what prevents *new* mirrors; forcing teardown to wait for an unbounded network write looks like an accidental side effect of the read/write-lock choice.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystem.java
```diff
@@ -28,18 +36,85 @@
...
  *
  * <p>Safe for DataAgent-style <em>user-managed</em> sandboxes that stay alive across
- * acquire/release. Self-managed sandboxes that stop on release may still fail if the async upload
- * races past shutdown.
+ * acquire/release. Self-managed sandbox release takes an exclusive gate before shutdown, so an
+ * upload either completes before release or is skipped without touching released resources.
  */
 public final class PinnedSandboxFilesystem extends SandboxBackedFilesystem {
 
+    private static final Map<Sandbox, MirrorGate> MIRROR_GATES =
```
oss-maintainer: **[Warning]** Two related concerns with a static gate map keyed on the `Sandbox` instance:

1. `released` is a one-way latch with no reset path. If the same `Sandbox` object is ever handed to a later call — resume/reconnect paths rebind sandboxes from persisted state (see #3025 for the same area), and `pinIfSandbox` creates a *new* `PinnedSandboxFilesystem` per mirror rather than per call — every future mirror for that instance is silently dropped at `debug` level, which is very hard to diagnose in a long-running service. Can the gate be cleared when a sandbox is re-bound (or keyed on the `SandboxAcquireResult`/`SandboxLease` instead, whose lifetime actually matches the release being modelled)?
2. `Collections.synchronizedMap(new WeakHashMap<>())` compares keys with `equals`/`hashCode`. That is fine only while no `Sandbox` implementation overrides them; a value-equality implementation would either share a gate between distinct sandboxes or fail to find the gate created in the constructor. Identity-based keying removes the trap.

oss-maintainer (commented):
### Summary
Guards the best-effort session mirror against races with self-managed sandbox release: a new static `Sandbox → MirrorGate` (`ReentrantReadWriteLock`) in `PinnedSandboxFilesystem` blocks release until in-flight uploads finish and rejects new ones afterwards, `SandboxLifecycleMiddleware` marks the gate before `sandboxManager.release(...)`, `SessionTree` skips mirrors for a released sandbox, and `ObjectStoreTranscriptStore.appendSegment` / `mirrorToFilesystem` now fail loudly instead of ignoring a returned failure response. The problem is real (#3036) and the "stop reporting a mirror as done when the upload actually failed" part is a genuine improvement — `compact()` in particular was previously free to delete the input segments after a lost upload.

### Findings
- **[Critical]** `PinnedSandboxFilesystem.java:81` — the read lock spans the whole remote upload while release takes the write lock, so call teardown now waits on an unbounded network transfer; one stalled upload hangs `releaseForCall` (and the lease close) on the agent's completion path.
- **[Warning]** `PinnedSandboxFilesystem.java:44` — `released` is a one-way latch keyed on the `Sandbox` instance with no reset path (resume/reconnect rebinds instances), and `WeakHashMap` key equality is not identity-based.
- **[Warning]** `SessionTree.java:745` — the new early return also skips the local `index.upsertFromLocalFile`, so a failed mirror leaves the workspace index stale for a file that exists on disk.
- **[Warning]** `ObjectStoreTranscriptStore.java:91` — `appendSegment` now throws, which `scheduleSegmentMirror` logs and `compact()` propagates; the interface documents no failure contract, and the "released" case becomes a WARN storm instead of a distinguishable skip.
- **[Warning]** `SessionTreeMirrorTest.java:213` — both new tests exercise `isRunning() == false` (the pre-existing liveness check); `markSandboxReleased` and the gate interaction are untested.
- **[Info]** `KubernetesSandbox.java:289` — solid addition; `exec`/`persistWorkspace`/`hydrateWorkspace` remain unguarded and `WORKSPACE_STOP_ERROR` is reused for a closed connection.

### Suggestions
The blocking-teardown item is the one I'd like resolved before merge — a bounded `tryLock` on the gate keeps the intended ordering for new mirrors without letting an un-timed remote write stall call release. The rest are contract/test hardening: name the "skipped because released" outcome so callers can treat it as normal, state the throws contract on `TranscriptStore#appendSegment`, keep the local index update independent of the remote copy, and add one test that drives the gate itself.

Also related for the maintainer's eye: #3094 (defer self-managed sandbox shutdown via a retain/release counter) and #3106 (move transcript persistence off the completion path) are attacking the same lifecycle window from three directions with three different mechanisms — worth converging on one owner for "async write vs. sandbox release" before all three land.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTree.java
```diff
@@ -728,9 +732,19 @@ private void mirrorToFilesystem(AbstractFilesystem fs, Path file, String relativ
...
+        }
         try {
             byte[] bytes = Files.readAllBytes(file);
-            fs.uploadFiles(fsRc, List.of(Map.entry(relativePath, bytes)));
+            List<FileUploadResponse> uploads =
+                    fs.uploadFiles(fsRc, List.of(Map.entry(relativePath, bytes)));
+            if (uploads.size() != 1 || !uploads.get(0).isSuccess()) {
+                String error =
+                        uploads.size() == 1 ? uploads.get(0).error() : "missing upload response";
+                log.warn("Failed to mirror session file {} to filesystem: {}", file, error);
```
oss-maintainer, line 764: **[Warning]** This early `return` also skips `index.upsertFromLocalFile(relativePath, file)` a few lines below, whose own comment says the index tracks the *local* file ("the local file already exists — update index with its current stats"). After this change a failed or skipped mirror leaves the workspace index stale for a file that is present on disk, so anything resolving sessions through the index (`SessionSearchTool`, compaction references, restore) can miss it even though the local write succeeded. If the goal is only to stop pretending the remote copy succeeded, consider logging the failure and still updating the index, or moving the index update ahead of the upload.

agentscope-harness/src/main/java/io/agentscope/harness/agent/transcript/ObjectStoreTranscriptStore.java
```diff
@@ -83,7 +84,12 @@ public String appendSegment(
             TranscriptRef ref, long seqStart, long seqEnd, String writerId, byte[] jsonl) {
         String name = seqStart + "-" + seqEnd + "-" + sanitize(writerId) + ".jsonl";
         String key = rootPrefix + ref.prefix() + "/events/" + name;
-        filesystem.uploadFiles(rc, List.of(Map.entry(key, jsonl)));
+        List<FileUploadResponse> uploads =
+                filesystem.uploadFiles(rc, List.of(Map.entry(key, jsonl)));
+        if (uploads.size() != 1 || !uploads.get(0).isSuccess()) {
+            String error = uploads.size() == 1 ? uploads.get(0).error() : "missing upload response";
+            throw new IllegalStateException("segment upload failed for " + key + ": " + error);
```
oss-maintainer, line 91: **[Warning]** Converting a returned failure response into a thrown `IllegalStateException` is the right instinct — previously a lost segment upload let `compact(ref)` continue and delete the input segments — but the contract is now implicit and shared by two very different callers: `scheduleSegmentMirror` catches and logs a WARN, while `compact` (line ~161) will abort mid-way. `TranscriptStore#appendSegment` documents only `@return storage key of the written segment`, so other implementations have no signal about whether throwing is expected. Worth stating the throws contract on the interface.

Also: with `PinnedSandboxFilesystem.uploadFiles` reporting a released sandbox as `FileUploadResponse.fail("Sandbox has been released")`, the race between the `skipMirrorForReleasedSandbox` pre-check and the actual upload now surfaces as an exception plus a per-segment WARN, for what this PR defines as a normal outcome. If "skipped because released" were distinguishable (a dedicated reason/enum on the response, or a `isReleased()` probe checked inside the same gate), the mirror path could log it once at DEBUG instead of producing a WARN storm for busy sessions during shutdown.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/session/SessionTreeMirrorTest.java
```diff
@@ -203,6 +209,39 @@ void flush_localWriteCompletesImmediately() throws Exception {
                 "local log file must exist immediately after flush()");
     }
 
+    @Test
+    void flush_skipsMirrorWhenPinnedSandboxWasReleased() throws Exception {
```
oss-maintainer, line 215: **[Warning]** Both new tests reach the skip through `isRunning() == false` on a `StoppedTransferSandbox`, so they cover the pre-existing liveness check but never the actual new mechanism: `markSandboxReleased` is never called, `MirrorGate.released` is never set, and the read/write-lock interaction (the part that can block teardown) is untested. A test that keeps `isRunning()` true, blocks the first upload on a latch, then calls `PinnedSandboxFilesystem.markSandboxReleased(sandbox)` from another thread — asserting (a) the mark does not return until the in-flight upload finishes or times out, and (b) a later mirror upload is skipped — would pin down the ordering this design depends on, and would surface the unbounded-wait concern raised on `uploadFiles`.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/main/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandbox.java
```diff
@@ -281,9 +282,18 @@ public void uploadFile(String absolutePath, byte[] content) throws Exception {
 
     @Override
     public byte[] downloadFile(String absolutePath) throws Exception {
+        requireActiveConnection();
         return sdkSandbox.files().read(requireFileApiRelative(absolutePath));
     }
 
+    private void requireActiveConnection() {
```
oss-maintainer, line 292: **[Info]** Good defensive addition, and the tests that assert `sdkSandbox.isActive()` is consulted before any transfer make it clear. Two nits: `exec` / `persistWorkspace` / `hydrateWorkspace` still dereference `sdkSandbox` unguarded — worth either guarding them for the same reason or noting in the method javadoc why the file API is the only racy entry point; and reusing `SandboxErrorCode.WORKSPACE_STOP_ERROR` for "connection closed" conflates two conditions for callers that branch on the code (retry vs. give up).

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystem.java
```diff
@@ -28,18 +40,171 @@
...
  * <p>Safe for DataAgent-style <em>user-managed</em> sandboxes that stay alive across
- * acquire/release. Self-managed sandboxes that stop on release may still fail if the async upload
- * races past shutdown.
+ * acquire/release. Self-managed sandbox release takes an exclusive gate before shutdown, so an
+ * upload either completes before release or is skipped without touching released resources.
  */
 public final class PinnedSandboxFilesystem extends SandboxBackedFilesystem {
 
+    private static final Logger log = LoggerFactory.getLogger(PinnedSandboxFilesystem.class);
+    private static final long RELEASE_GATE_TIMEOUT_MILLIS = 1_000;
```
oss-maintainer, line 49: `RELEASE_GATE_TIMEOUT_MILLIS = 1_000` is a fixed, non-configurable data-loss budget: the `released` latch is published first, so once the write-lock `tryLock` gives up, release proceeds while a mirror upload is still in flight and that workspace/transcript mirror is silently abandoned. For a large archive tar over a slow or congested network 1 s is easily crossed in normal operation, not just in a wedged transfer. Would you consider making it configurable (system property or the sandbox settings the manager already carries) and logging at the point where a mirror is skipped rather than only where the gate is abandoned, so operators can distinguish "bounded release skipped my mirror" from a real upload failure?

oss-maintainer (commented):
### Summary
Re-review after the author's follow-up commit `e0fe07df01` (address sandbox mirror review findings). The `MirrorGate` design holds up: `released` is published **before** waiting for the write lock, so new mirror uploads are rejected even when an existing transfer is wedged, and release stays bounded because mirrors are best-effort. The per-sandbox generation is replaced (not reopened) on `markSandboxAcquired`, which correctly stops a previous call's mirror from attaching to the new call's lifecycle, and `gateFor` keeps the shared `MIRROR_GATES` map under `synchronized` with weak-key expunge so gates do not outlive their sandbox. The added tests (`flush_skipsMirrorWhenPinnedSandboxWasReleased`, `flush_skipsTranscriptSegmentWhenPinnedSandboxWasReleased`, and the "released sandbox must not execute / persist / hydrate / download" guards) target the use-after-release failure the PR exists for.

One concern inline: the hardcoded 1 s gate cap.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystem.java
```diff
@@ -28,18 +40,171 @@
...
+        try {
+            return !mirrorGate.released && pinnedSandbox.isRunning();
+        } finally {
+            mirrorGate.lock.readLock().unlock();
+        }
+    }
+
+    /** Returns whether this pinned filesystem's call generation has been released. */
+    public boolean isSandboxReleased() {
+        return mirrorGate.released;
```
oss-maintainer, line 103: `isSandboxReleased()` reads `MirrorGate.released` without holding the read lock, so it is the only read of that field that is not fenced by the lock — and it is the single carrier of the "this failure happened because the sandbox was released" signal. `PinnedSandboxFilesystem.uploadFiles()` returns a generic `FileUploadResponse.fail(path, "Sandbox has been released")` instead of propagating the reason, and `SessionTree.mirrorToFilesystem` decides warn-vs-debug purely from `isReleasedSandbox(fs)`. Under the JMM the plain read may keep observing `false` after another thread's `markReleased()` set it, which turns every post-release flush into a `log.warn` "Failed to mirror session file ..." for a sandbox that is intentionally gone — noisy and, worse, misleading during an incident.

Either take the same read lock as `isSandboxRunning()`:

```java
public boolean isSandboxReleased() {
    mirrorGate.lock.readLock().lock();
    try {
        return mirrorGate.released;
    } finally {
        mirrorGate.lock.readLock().unlock();
    }
}
```

or — probably better — put the reason into the response (`FileUploadResponse` error code / a `SANDBOX_RELEASED` sentinel) so `SessionTree` does not have to re-derive it from a second source of truth.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
```diff
@@ -169,10 +172,17 @@ public void releaseForCall(RuntimeContext ctx) {
...
             sandboxManager.release(result);
         } catch (Exception e) {
             log.warn("[sandbox-mw] Failed to release sandbox session: {}", e.getMessage(), e);
         }
         result.getLease().close();
     }
+
+    private static void markMirrorSandboxReleased(SandboxAcquireResult result) {
+        if (result.isSelfManaged()) {
+            PinnedSandboxFilesystem.markSandboxReleased(result.getSandbox());
```
oss-maintainer, line 185: This is the hard-coded gate cap I flagged last round, and it is still a fixed `1_000` ms paid synchronously on the release path: `markSandboxReleased` takes the write lock with `tryLock(1s)` before `sandboxManager.release(result)`, so a wedged mirror upload now adds up to a full second to **every** self-managed call release (and `releaseForCall` also runs on the `acquireForCall` error-unwind path at line 131, where the caller is already handling a failure). Two asks:

1. Make the cap configurable from the sandbox/starter properties (with the current value as the default) — an arbitrary constant buried in a `final class` is the kind of thing that has to be tuned in production, and the only escape hatch today would be a `log.warn`.
2. Note the interaction with `PinnedSandboxFilesystem.uploadFiles`, which holds the **read** lock across the whole remote transfer: a slow-but-successful upload keeps the writer waiting, so the 1 s is a per-release latency floor, not an upper bound on how long a release is delayed across consecutive calls.

Not a blocker for correctness — release stays bounded and the `released = true` publication before waiting is the right order — but the value should not be a compile-time constant.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystem.java
```diff
@@ -28,18 +40,171 @@
...
+    /**
+     * Starts a fresh mirror-gate generation for a newly acquired sandbox.
+     *
+     * <p>A manager may reuse the same {@link Sandbox} object for a later call. Replacing rather
+     * than reopening the old gate keeps mirrors from the previous call permanently released while
+     * ensuring they cannot attach themselves to the new call's lifecycle.
+     */
+    public static void markSandboxAcquired(Sandbox sandbox) {
+        if (sandbox == null) {
+            return;
```
oss-maintainer, line 71: `markSandboxAcquired` replaces the gate for a `Sandbox` **object identity**, which is what makes a new call generation safe. The assumption that is not written down (or enforced) is one-live-generation-per-instance: if the same `Sandbox` instance is ever acquired by two concurrent calls — a `NONE`/`SELF_MANAGED` instance handed out twice, or an object pooled by a future manager — the second acquire drops the first call's gate, and in-flight mirrors pinned to it are no longer reachable from `markSandboxReleased` for that generation. Worth a one-line `@implNote` next to the method plus, if cheap, a debug-level assertion that the previous gate was already released before being replaced. Fine to leave as-is if single-generation-per-instance is already a manager invariant.

oss-maintainer (commented):
### Summary
Re-review of the new commit `560861f`. The release-gate design is sound: `released` is published before the write-lock wait, a gate is replaced (not reopened) per acquire generation, `KubernetesSandbox.requireActiveConnection()` now turns use-after-close into a clear `SANDBOX_CONNECTION_CLOSED` instead of an opaque SDK failure, and moving `index.upsertFromLocalFile` ahead of the upload is right — the index describes the authoritative local file, not the best-effort mirror. The new tests cover both the released-sandbox guards and the never-execute/never-persist contract.

Three points inline: the plain `released` read in `isSandboxReleased()` is the only carrier of the release reason and can mis-warn on every post-release flush; the 1 s gate cap should be configurable rather than a compile-time constant (this was the open item from last round); and the identity-keyed gate map deserves a note on its one-generation-per-instance assumption. CLA signed.

---

oss-maintainer:

This PR conflicts with `main` and cannot be merged (`mergeable_state=dirty` on head `560861f5`). `main` has advanced substantially since your last commit (2026-09-14), so a rebase should pick up the sandbox and session-mirror changes it now depends on:

```bash
git fetch origin
git checkout fix/session-tree-sandbox-mirror-lifecycle
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

Once the conflict is cleared, comment `@oss-maintainer please re-review` and this PR goes back into the review queue. This is a one-time reminder.

---

oss-maintainer (commented):
### Summary
Re-review of the two commits since `560861f`: `dab48dba` (configurable mirror drain budget) and the merge of `main`. The drain-budget knob does what I asked for and the javadoc is candid about the remaining hole ("an upload still in flight when the wait expires may fail during shutdown") — good. Two things need attention before this is ready:

1. **CI is red on this head.** The `requireActiveConnection()` guard added to `doHydrateWorkspace` breaks `KubernetesSandboxColdRestoreTest.persistedSnapshotSurvivesAcquireStartAndStop` (3/3 parameterised cases) because that test's mocked SDK sandbox never stubs `isActive()` — see the inline comment for the stack trace. `build (ubuntu-latest)` was cancelled, so only the Windows job surfaced it.
2. **The new system property is undocumented and process-global.** `agentscope.state.home` sets the precedent for a *root-path* override and is documented in both the en and zh docs; a timing tunable read on every release arguably belongs on the spec/builder with the property as a fallback.

Non-blocking note on the stall test's wall-clock bands — on the same runner class that just failed, a 900 ms ceiling against a 1000 ms default is a thin margin.

Everything from the previous round that I asked about (release latch ordering, gate replacement per generation, `index.upsertFromLocalFile` ahead of the upload) still holds after the rebase.

---

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/main/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandbox.java
```diff
@@ -168,6 +170,7 @@ private InputStream persistViaExec(String root, String tarArgs) throws Exception
 
     @Override
     protected void doHydrateWorkspace(InputStream archive) throws Exception {
+        requireActiveConnection();
```
oss-maintainer, line 173: This guard turns `KubernetesSandboxColdRestoreTest.persistedSnapshotSurvivesAcquireStartAndStop` red — that is currently the only failing check on the PR head (`build (windows-latest)` = failure, `build (ubuntu-latest)` = cancelled):

```
SandboxException$WorkspaceStartException: Failed to start workspace at: \workspace
Caused by: SandboxRuntimeException: Kubernetes sandbox connection has been closed
    at KubernetesSandbox.requireActiveConnection(KubernetesSandbox.java:299)
    at KubernetesSandbox.doHydrateWorkspace(KubernetesSandbox.java:173)
    at KubernetesSandboxColdRestoreTest.persistedSnapshotSurvivesAcquireStartAndStop(...:147)
```

The cold-restore test builds `sdkSandbox` with `mock(...)` and stubs only `commands()` / `files()`, so the new `isActive()` call inside the `start()` path returns Mockito's default `false` and branch C (`snapshot.restore()` -> `doHydrateWorkspace`) now throws. `KubernetesSandboxTest` was updated with `when(sdkSandbox.isActive()).thenReturn(true)`; the cold-restore test needs the same. Worth grepping every test that drives `start()`/`stop()` on a mocked SDK sandbox, since this guard is now on the hydrate path too, not just on user-facing exec/transfer calls.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystem.java
```diff
@@ -28,18 +40,212 @@
...
  */
 public final class PinnedSandboxFilesystem extends SandboxBackedFilesystem {
 
+    private static final Logger log = LoggerFactory.getLogger(PinnedSandboxFilesystem.class);
+
+    /**
+     * JVM property controlling the per-release mirror drain budget in milliseconds (default 1000).
+     * Zero disables waiting. Invalid or negative values fall back to the default.
+     */
+    public static final String RELEASE_GATE_TIMEOUT_PROPERTY =
```
oss-maintainer, line 54: Making the drain budget configurable is the right call (the hard-coded 1s was the main thing I flagged last round), but a JVM-global system property is the weakest of the available config surfaces here:

* it is process-wide, so a host app that runs agents with different sandbox lifetimes cannot tune one without affecting the other;
* it is read on **every** `markSandboxReleased` (`System.getProperty` + `Long.parseLong` on the shutdown path), which is fine cost-wise but means the effective value can change underneath an in-flight release;
* it is invisible to the builder — `HarnessAgent` / the filesystem spec have no `.sandboxMirrorReleaseTimeout(...)`-style knob, so discovery is only via grepping the constant.

The precedent in this repo (`agentscope.state.home`, `HarnessAgent.java:1137`) is a *deployment-time root path*, and it is documented in `docs/v2/en/docs/building-blocks/context.md` and the `zh` counterpart. This new property is documented nowhere outside the javadoc. Suggest at minimum adding it to the sandbox/workspace docs pages (en + zh) next to `agentscope.state.home`, and ideally threading it through the spec/builder with the system property kept as the fallback override — that matches how every other tunable in this module reaches users.

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystemTest.java
```diff
@@ -0,0 +1,235 @@
...
+            assertTrue(sandbox.uploadStarted.await(2, TimeUnit.SECONDS));
+
+            long started = System.nanoTime();
+            PinnedSandboxFilesystem.markSandboxReleased(sandbox);
+            long elapsedMillis = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - started);
+
+            assertTrue(
+                    elapsedMillis >= timeoutMillis, "release should allow the configured budget");
+            assertTrue(
+                    elapsedMillis < 900, "release must use the configured budget, not the default");
```
oss-maintainer, line 112: The lower bound `elapsedMillis >= timeoutMillis` is vacuous for the `0` case and only 50 ms wide for the second one, while the upper bound is fixed at 900 ms. So the assertion that actually carries the regression risk ("did we silently fall back to the 1000 ms default?") sits 100 ms below the default on a shared runner that is already known to be slow here — `windows-latest` is the job that just failed. Two cheap hardening options:

* assert on the *observed* wait rather than a wall-clock band, e.g. expose the timeout the gate used (package-private `lastDrainBudgetMillis` or a `markReleased(sandbox, timeout)` overload used only by tests) and assert equality;
* or keep the timing check but widen the band relative to the configured value (`elapsedMillis < timeoutMillis + 400`) and add a separate explicit case asserting that an unset property still yields ~1000 ms, which is the number the band is really protecting.

The `@ResourceLock(Resources.SYSTEM_PROPERTIES)` addition and the `previous` restore in `finally` are correct — without them these global-property tests would race the rest of the class.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/test/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxColdRestoreTest.java
```diff
@@ -104,6 +104,7 @@ void persistedSnapshotSurvivesAcquireStartAndStop(boolean claimGone, boolean wor
         RuntimeContext runtime = RuntimeContext.builder().sessionId("conversation").build();
 
         var sdkSandbox = mock(io.agentscope.extensions.sandbox.kubernetes.client.Sandbox.class);
+        when(sdkSandbox.isActive()).thenReturn(true);
```
oss-maintainer, line 107: Confirmed fixed on this head. Stubbing `isActive()` here is exactly what the previous round asked for: the new `requireActiveConnection()` guard made this parameterised case throw, which is why `build (windows-latest)` was red on `50c7c93d`. On `7a192fe5` all five checks (`build ubuntu-latest`, `build windows-latest`, `Check License`, `Check Module Sync`, `codecov/patch`) pass and `license/cla` reports signed. Blocking item 1 is closed.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystem.java
```diff
@@ -28,18 +40,212 @@
...
+            "agentscope.sandbox.mirror.release-timeout-millis";
```
oss-maintainer, line 56: Blocking item 2 from the previous round is still open. `agentscope.sandbox.mirror.release-timeout-millis` is a new *public, process-global* knob, and `git grep 'release-timeout'` across `docs/**` (en and zh trees) returns nothing, so an operator hitting a wedged mirror cannot discover it exists. `agentscope.state.home` is the in-repo precedent: a root-path override documented in the configuration reference in both languages. Either of these closes it:

1. document the key next to `agentscope.state.home` (en + zh) — unit, the `1000` ms default, that `0` disables waiting, and that the budget is *per release* rather than a global or per-upload timeout; or
2. put the drain budget on the filesystem spec/builder and keep the system property only as a fallback, so it stops being a JVM-wide ambient value.

The javadoc on the constant is accurate, but javadoc is not what an operator greps when teardown is stalling.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystem.java
```diff
@@ -28,18 +40,212 @@
...
+     * Prevents new mirror uploads and briefly waits for an in-flight upload before shutdown.
+     *
+     * <p>The budget is read from {@link #RELEASE_GATE_TIMEOUT_PROPERTY} on each release. Uploads
+     * hold the read lock throughout remote transfer, so even a healthy slow upload can consume
+     * the entire budget. Consecutive releases each incur their own wait; this is not a global
+     * timeout or an upload timeout. Shutdown proceeds when the budget expires.
+     */
+    public static void markSandboxReleased(Sandbox sandbox) {
+        if (sandbox != null) {
+            gateFor(sandbox).markReleased(releaseGateTimeoutMillis());
```
oss-maintainer, line 101: Still worth a sentence in the javadoc (and it is the part of the design I would like a maintainer to eyeball): this wait runs synchronously on the release path before `sandboxManager.release(...)`, so with the default budget a wedged mirror costs up to 1 s **per release**, and consecutive releases each pay it again — the property is a per-release cap, not a bound on total shutdown time. That is a reasonable trade for keeping remote I/O off the mirror hot path, but the class doc currently reads as if the knob is a global timeout; `Consecutive releases each incur their own wait` says it, and I would make that the headline of the property description too.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystem.java
```diff
@@ -28,18 +40,212 @@
...
+        }
+    }
+
     @Override
     public synchronized void clearSandboxIfCurrent(Sandbox expected) {
         // Keep the pin for out-of-call mirror uploads.
     }
+
+    static long releaseGateTimeoutMillis() {
+        String configured = System.getProperty(RELEASE_GATE_TIMEOUT_PROPERTY);
```
oss-maintainer, line 154: Non-blocking: this is a process-wide property read plus `Long.parseLong` on every release, with a `NumberFormatException` branch that exists only because the value has no other home. If the budget moves onto the spec/builder as above, both disappear and the effective value becomes visible from the object graph while debugging. If the system property stays, that is defensible — just make it documented so the trade-off is deliberate.

oss-maintainer (commented):
### Summary

Re-review on head `7a192fe5`. The single new commit is the test stub the previous round asked for, and it does close the CI finding: `build (ubuntu-latest)`, `build (windows-latest)`, `Check License`, `Check Module Sync` and `codecov/patch` are all green and `license/cla` reports signed. Fail-closed on a released sandbox is the right shape for #3036, and the `generate`-side ordering (release latch before drain, gate replaced per generation, `index.upsertFromLocalFile` ahead of the upload) still holds after the rebase.

Not converting to an approval, for one reason only: item 2 of the previous round — the new `agentscope.sandbox.mirror.release-timeout-millis` knob is process-global and not documented anywhere operators will look — is unchanged by this push. Everything in the code itself is fine; this is a one-file docs/config decision. If you document it next to `agentscope.state.home` (en + zh), or move the budget onto the filesystem spec/builder with the property as fallback, this is an approval on the next tick without any further code changes.

### Findings

- **[Warning]** `PinnedSandboxFilesystem.java:55` — new public, process-global JVM property with no presence in `docs/**` (en or zh); discoverability gap for the exact failure mode this PR fixes.
- **[Warning]** `PinnedSandboxFilesystem.java:100` — the wait is synchronous on the release path and is *per release*, not a global bound; worth stating in the property doc so nobody sets it expecting a shutdown deadline.
- **[Info]** `PinnedSandboxFilesystem.java:153` — property read + `Long.parseLong` on every release, with a `NumberFormatException` branch that only exists because the value has no other home.
- **[Info]** `KubernetesSandboxColdRestoreTest.java:107` — previous round's CI blocker confirmed fixed on this head.

### Notes for a maintainer

`mergeable` is `true` but `mergeStateStatus` is `blocked`, which here means branch protection still wants a *human* approval, not that CI is failing. 8 of 13 review threads are resolved; the other 5 are all from this bot's earlier passes and the two design questions among them (unlocked read of `MirrorGate.released`, and one-live-generation-per-`Sandbox`-instance being an assumption rather than an enforced invariant) are worth a maintainer's judgement before merge.

---

agentscope-harness/src/test/java/io/agentscope/harness/agent/filesystem/sandbox/PinnedSandboxFilesystemTest.java
```diff
@@ -0,0 +1,233 @@
...
+                    executor.submit(
+                            () ->
+                                    filesystem.uploadFiles(
+                                            RuntimeContext.empty(),
+                                            List.of(Map.entry("session.jsonl", new byte[] {1}))));
+            assertTrue(sandbox.uploadStarted.await(2, TimeUnit.SECONDS));
+
+            // Check configuration directly; do not distinguish 50 ms from the 1000 ms default
+            // using a narrow wall-clock band on shared CI runners.
+            assertEquals(timeoutMillis, PinnedSandboxFilesystem.releaseGateTimeoutMillis());
```
oss-maintainer: **[Info]** Non-blocking, and the flakiness trade-off is the right call — recording it so the intent is on the record. This assertion now pins the *parser* (`releaseGateTimeoutMillis()` returns what the property says), while the previous version also asserted the release path stayed inside `[timeoutMillis, 900ms)`, i.e. that `markSandboxReleased` actually consumed the configured budget rather than the `1000` default. After this change the "uses the configured value" property is covered only by the single call site `gateFor(sandbox).markReleased(releaseGateTimeoutMillis())`, and the `release.get(5, TimeUnit.SECONDS)` bound is too wide to notice a wrong-but-plausible budget.

Fine to leave as is given the single call site; if we ever want it pinned without a wall-clock band, capturing the value handed to `markReleased(...)` on the gate (or a `MirrorGate` latch count) would assert the behaviour without depending on runner latency.

oss-maintainer (approved):
### Summary
Re-review on head `32911482`. Both open findings from the previous round are resolved by this push, so this is the approval that round promised.

- **Process-global knob now documented (en + zh).** `docs/v2/en/docs/building-blocks/context.md` and its `zh` twin gained a "Sandbox mirror release wait" / "沙箱镜像的释放等待" section next to the `agentscope.state.home` discussion: property name, process-wide JVM scope, read at each release, `1000` default, `0` = no wait, malformed/negative fallback, the exact warning string to grep for, and the tuning trade-off. The claims match the implementation in `releaseGateTimeoutMillis()` — the fallback branches were checked, not just the prose.
- **Per-release vs. total-deadline ambiguity removed.** The `RELEASE_GATE_TIMEOUT_PROPERTY` javadoc now states "Each consecutive release has its own budget; this is not a total shutdown deadline", which is what the code does (`gateFor(sandbox).markReleased(releaseGateTimeoutMillis())` per release), and both doc languages repeat it.
- **Timing-band assertions dropped.** `releaseDoesNotWaitIndefinitelyForStalledUpload` pins the configuration directly and bounds the release with `release.get(5, TimeUnit.SECONDS)` instead of measuring wall-clock inside `[50ms, 900ms)`, so the release path no longer depends on scheduler latency. Right trade-off — no request to bring the band back; one non-blocking note is left on the test about what the weaker assertion no longer covers.

CI is green on this head (`SUCCESS=7`), `license/cla` is signed, and `mergeable` is `MERGEABLE`.

**Still for a human maintainer, unchanged from last round:** `mergeStateStatus` is `blocked` because branch protection wants a human approval, and the two design questions from earlier threads are judgement calls rather than defects — the unlocked read of `MirrorGate.released`, and one-live-generation-per-`Sandbox` being an assumption rather than an enforced invariant. Related: #3036's fail-closed expectation is met.

---

PR #3041, opened 2026-09-08
perf(core): cache generated JSON schemas to remove lock from hot path

jnduan:
## Context

Follow-up to #2796 (now merged as 6ba5fefe), which made `JsonSchemaUtils`
thread-safe by serializing every schema generation on `SCHEMA_LOCK`.

That fix is correct, but it re-runs reflective schema generation on *every*
structured-output and tool call: `ReActAgent.doStructuredCall` regenerates the
schema for the same class on each call, so under load the lock is taken
repeatedly for identical, deterministic work.

## What this PR does

Cache the generated schema per `Class` / `Type`, so `SCHEMA_LOCK` is taken
**only on a cache miss** (the first time a class or type is seen). A cache hit
converts a fresh, independently mutable `Map` from the cached, never-mutated
node — no lock required.

Entries are held in `ClassValue` slots, so a slot lives on the class it
describes instead of in a static map that strongly references the class as a
key. The `Type` cache groups entries by the class that declares them
(`ClassValue<Map<Type, JsonNode>>`), so the parametrizations of one class share
a slot and follow class unloading. A type with no raw class rises to the class
that declares it — a type variable to its declaring class, or to the class
declaring its method, and a generic array to whichever class its component type
resolves to — so it is cached too, rather than regenerated under the lock on
every call.

## Why it's safe

- **Independent maps per call.** Each call still returns a freshly-converted
  mutable `Map`, so callers that mutate the result in place (e.g.
  `ToolSchemaGenerator`, which hoists `$defs` out of nested schemas) cannot
  corrupt the cache or interfere with one another. Covered down to nested
  structures, not just the top level.
- **No classloader pinning.** Entries live on the class they describe rather
  than in a static `Map<Class, JsonNode>`, which strongly references the key
  class and therefore pins it — and its classloader — for the JVM's life. That
  matters because structured-output and tool-parameter classes are not always
  compile-time-fixed: `agentscope-extensions` loads skills and tools at
  runtime. Scoping entries to the class removes that edge rather than capping it.
  Attributed types follow the same rule, and deliberately do not use a static
  `Map<Type, JsonNode>` fallback: a type variable holds its declaration, so such
  a map would re-introduce exactly the reference this scoping removes.
- **Bounded entries.** A class's `Type` map grows with the distinct generic
  signatures written against that class *in code*: every `Type` reaching this
  utility originates in a `TypeReference` literal or a reflective method
  signature, and structurally equal signatures share one entry. Loading a class
  at runtime adds a class with its own cache rather than another entry on an
  existing one, and synthesized types are released with the class that owns the
  cache.
- **No `computeIfAbsent`.** It runs the generation while holding the map's bin
  lock, which serializes unrelated classes hashing to the same bin behind a
  whole generation, and violates the contract that a remapping function must not
  modify its own map. Both caches now publish with a re-read inside
  `SCHEMA_LOCK`, which also preserves "generated exactly once per key".
- **No deadlock.** Lock order is one-directional (cache slot read →
  `SCHEMA_LOCK`); while holding `SCHEMA_LOCK` we only call `generateSchema`,
  never touching the cache slots, so there is no reverse path.
- **Null behavior unchanged.** `generateSchemaFromType` rejects a null type with
  `Objects.requireNonNull` before it reaches the cache, matching the pre-cache
  path.

Known boundary: a dynamically loaded class appearing as a *type argument* (e.g.
`List<TenantModel>`) is still held by `List`'s slot for as long as `List` lives,
because the inner map's keys are the parametrized types themselves. Closing
that would cost a lock on the hit path, or permanent misses with weak keys.
Only a type with no attributing class at all — a wildcard synthesized by a
caller — is generated without caching, and no reflective signature can reach it.

## Tests

- Repeated calls return equal but independently mutable maps, for both `Class`
  and `Type` entry points, including mutations below the top level.
- Null-argument parity, the shapes without a raw class that are now attributed,
  and the uncached path are all covered.
- The 12-thread concurrency test from #2796 now races two variants of the *same*
  raw class, so it exercises the shared slot map, and asserts the two variants
  yield different schemas.
- Full `agentscope-core` suite: 2345 tests, 0 failures, 9 skipped.

No public API signature changes.


agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
```diff
@@ -106,10 +131,14 @@ public class JsonSchemaUtils {
      */
     public static Map<String, Object> generateSchemaFromClass(Class<?> clazz) {
         try {
-            JsonNode schemaNode;
-            synchronized (SCHEMA_LOCK) {
-                schemaNode = schemaGenerator.generateSchema(clazz);
-            }
+            JsonNode schemaNode =
+                    CLASS_SCHEMA_CACHE.computeIfAbsent(
```
oss-maintainer: `computeIfAbsent` runs the remapping function while holding the map's bin lock, and here that function then blocks on the process-wide `SCHEMA_LOCK`. Two consequences: (1) `ConcurrentHashMap` forbids a remapping function that mutates the same map — if anything reachable from `schemaGenerator.generateSchema(...)` (a Jackson module/serializer callback, or a future nested `generateSchemaFromType` call) touches this map, the second update blocks on the bin lock held by the first: self-deadlock on the same key, `IllegalStateException` on the same bin. (2) Unrelated classes hashing into the same bin are blocked for the whole generation, so the lock is only truly off the path on cache hits. Equivalent shape without nesting the locks:

```java
JsonNode cached = CLASS_SCHEMA_CACHE.get(clazz);
if (cached == null) {
    synchronized (SCHEMA_LOCK) {
        cached = CLASS_SCHEMA_CACHE.get(clazz);
        if (cached == null) {
            cached = schemaGenerator.generateSchema(clazz);
            CLASS_SCHEMA_CACHE.put(clazz, cached);
        }
    }
}
```

`TYPE_SCHEMA_CACHE` (line 177) has the same issue.

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
```diff
@@ -68,10 +71,32 @@ public class JsonSchemaUtils {
...
+     * converts a fresh, independently mutable {@link Map} from the cached node, so callers that
+     * mutate the returned map (e.g. {@code ToolSchemaGenerator}) cannot corrupt the cache or
+     * interfere with one another.
+     *
+     * <p>Unbounded by design: keys are the compile-time-fixed structured-output and
+     * tool-parameter classes declared by application code, so the entry count is bounded by the
+     * small, finite set of such classes the JVM loads for that purpose — not by request volume
+     * or untrusted input.
+     */
+    private static final Map<Class<?>, JsonNode> CLASS_SCHEMA_CACHE = new ConcurrentHashMap<>();
```
oss-maintainer: The key is a strong `Class<?>` reference held in a `static` map that is never evicted, so each entry pins the class **and its classloader** for JVM lifetime. The comment argues the key space is bounded by compile-time-fixed classes, which holds for `ReActAgent.generateSchemaFromClass(targetClass)` and reflective `@ToolParam` types — but not for the paths this framework itself supports: `agentscope-extensions` loads skill/tool classes at runtime and `agentscope-distribution` runs multi-tenant, so per-tenant/per-skill classloaders mint fresh keys with no upper bound. That turns a hot-path optimization into a classloader leak that survives tenant unload. `ClassValue<JsonNode>` fits the class-keyed cache exactly (identity-scoped, classloader-safe, no eviction bookkeeping); the `Type`-keyed one needs a bound (e.g. Caffeine `maximumSize` with soft values). Worth one javadoc sentence naming which callers may pass dynamically loaded classes.

agentscope-core/src/test/java/io/agentscope/core/util/JsonSchemaUtilsTest.java
```diff
@@ -168,6 +169,55 @@ void testGenerateSchemaFromType() {
...
+    void testGenerateSchemaFromClassRepeatedCallsReturnEqualIndependentMaps() {
+        Map<String, Object> first = JsonSchemaUtils.generateSchemaFromClass(SimpleModel.class);
+        Map<String, Object> second = JsonSchemaUtils.generateSchemaFromClass(SimpleModel.class);
+
+        // A repeated class must yield an equal schema, so caching cannot change the result.
+        assertEquals(first, second);
+
+        // Each call must return a fresh, independently mutable map: mutating one must not leak
+        // into another, matching in-place-mutating callers such as ToolSchemaGenerator.
+        first.put("description", "mutated");
```
oss-maintainer, line 182: Independence is only asserted at the top level (`first.put("description", ...)`). The real mutating callers reach into nested structures — `ToolSchemaGenerator`/`ReActAgent` hoist `$defs`/`definitions` out of inner schemas and merge them into the params root, removing and replacing nested maps. Please add one case that mutates a nested entry (e.g. remove a key under `properties`, or perform a `$defs` hoist) and re-reads from a second call. Jackson's `convertValue` deep-copies, so the code is almost certainly fine — but that deep-copy is exactly the invariant this cache now silently depends on.

oss-maintainer (commented):
### Summary
Caching the generated `JsonNode` per class/type and taking `SCHEMA_LOCK` only on a miss is the right fix for the hot-path serialization, and the "every call converts a fresh mutable Map" invariant is backed by tests. Filed as COMMENT rather than APPROVE: two robustness points are worth settling before merge (locks nested inside `computeIfAbsent`, and an unbounded static cache keyed by `Class`); neither is a correctness bug on today's call sites.

### Findings
- **[Warning]** `JsonSchemaUtils.java:135` — remapping function blocks on a global lock while the CHM bin lock is held; breaks the "must not mutate this map" contract and blocks unrelated keys in the same bin.
- **[Warning]** `JsonSchemaUtils.java:92` — static, never-evicted cache with strong `Class` keys pins classloaders; unsafe under runtime-loaded skills / multi-tenant distribution.
- **[Info]** `JsonSchemaUtilsTest.java:182` — independence tested only at the top level; nested `$defs`/properties mutation is what callers actually do.

### Suggestions
Move to get -> double-check under lock -> put, and to `ClassValue` (or a bounded cache) for the key space. Both keep the public API and the existing tests green.

---

jnduan:
Thanks for the review — both findings hold up, and I've reworked the caching along those lines.

**1. `computeIfAbsent` / nested lock**

My "no deadlock" reasoning was too narrow. I checked lock ordering (bin lock → `SCHEMA_LOCK`) and that `SCHEMA_LOCK` never touches the cache map, but missed the two points that make `computeIfAbsent` the wrong tool here:

- `ConcurrentHashMap` forbids the remapping function from modifying its own map. Nothing does that today, but the contract only holds by accident of what `generateSchema` currently reaches.
- Unrelated classes hashing to the same bin were serialized behind the whole generation — a real regression on the hot path, not just a theoretical one.

`computeIfAbsent` is gone. Both caches now re-read the slot inside the lock:

```java
JsonNode cached = slot.get();
if (cached != null) {
    return cached;
}
synchronized (SCHEMA_LOCK) {
    cached = slot.get();              // double-check
    if (cached == null) {
        cached = schemaGenerator.generateSchema(type);
        slot.put(type, cached);
    }
}
return cached;
```

(The class path is the same shape, with an `AtomicReference` slot instead of a map.)

**2. Classloader pinning**

Adopted `ClassValue` for both caches: entries are held on the class they describe rather than in a static map that strongly references the class as a key, so the pinning edge is removed rather than capped.

One detail worth calling out: I do **not** generate inside `ClassValue#computeValue`. Its javadoc allows the JDK to invoke it more than once for the same class ("several racing threads have computed values, one is chosen"), which would re-generate whenever several threads first reach a class at once — a step back from the current guarantee of generating exactly once per key. So `computeValue` only allocates an empty slot (`ClassValue<AtomicReference<JsonNode>>`), and the value is published by the double-checked read above. Net: the same once-per-key guarantee as before, plus the classloader-safe scope.

For the `Type` cache I grouped keys by raw class — `ClassValue<Map<Type, JsonNode>>`, the inner map holding the parametrizations of one raw class (`List<String>` vs `List<Integer>`). Those entries are classloader-scoped in the same way, and unrelated raw classes no longer share a lock. Types with no raw class (type variables, wildcards, generic arrays) now generate without caching: correctness is unchanged, they only give up the caching benefit. That is a deliberate behavior difference from this PR's first revision, which cached them in the `Type`-keyed map.

That also made a separate bound unnecessary: the key space now follows class unloading instead of growing for JVM life, and `agentscope-core` has no Caffeine today, so I kept the dependency set untouched.

One boundary I did not close: the inner map's keys are the parametrized types themselves, so a dynamically loaded class appearing as a type argument (e.g. `List<TenantModel>`) is still held alive by `List`'s slot for as long as `List` lives. Closing that too would need the inner map to track type-argument classes as well, at the cost of taking a lock on the hit path — or of never hitting, if keys were weak, since reflection hands out a fresh `Type` instance per call. Since the extension paths in the finding load skill/tool classes as the target class itself, I went with the raw-class grouping and left that corner open. Happy to revisit if you'd rather close it.

**3. Nested mutation coverage**

Added two cases that mutate below the top level — removing a key under `properties`, and writing into a nested property schema — then read the same class/type again and assert the pristine shape. This pins down the deep-copy invariant that `ToolSchemaGenerator`'s `$defs` hoisting depends on. Also added a case for the no-raw-class fallback above, and updated the null-argument comment that still referred to `computeIfAbsent`.

Verified: `spotless:check` clean, `JsonSchemaUtilsTest` 16/16, full `agentscope-core` suite 2342 tests / 0 failures / 9 skipped.

Please take another look.

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
```diff
@@ -156,6 +199,84 @@ public static Map<String, Object> generateSchemaFromType(Type type) {
...
+     *
+     * <p>The miss path re-reads the slot inside the lock rather than calling {@code
+     * ConcurrentHashMap#computeIfAbsent}, which runs its mapping function while holding the map's
+     * bin lock: nesting the schema lock inside it would block unrelated types hashing to the same
+     * bin for the whole generation, and would break the contract that a mapping function must not
+     * modify its own map.
+     */
+    private static JsonNode cachedSchemaNode(Type type) {
+        Class<?> rawType = rawTypeOf(type);
+        if (rawType == null) {
```
oss-maintainer: **[Warning]** Uncacheable types now regenerate on **every call while holding the global `SCHEMA_LOCK`**.

Type variables, wildcards and generic arrays have no raw class to hang a slot on, so they bypass caching entirely and take the process-wide lock for the full duration of `schemaGenerator.generateSchema(type)`. Previously `TYPE_SCHEMA_CACHE` was a plain `Map<Type, JsonNode>` and cached these types like any other, so this is a latency regression on the fallback path — and because it holds the shared lock, it also stalls schema generation for every *other* class in the JVM.

The added test (`testGenerateSchemaFromTypeVariableStillGenerates`) proves the path is correct but asserts nothing about cost. Two options:
1. key a second `ClassValue`-free `ConcurrentHashMap<Type, JsonNode>` fallback off the `Type` itself (these types are interned/`equals`-stable, so they are fine as keys), or
2. confirm with a benchmark that no reachable structured-output path produces a `TypeVariable`/`WildcardType`, and say so in the javadoc so the next reader does not have to re-derive it.

Option 1 is cheap and removes the cliff.

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
```diff
@@ -68,10 +79,48 @@ public class JsonSchemaUtils {
...
+    /**
+     * Same caching strategy as {@link #CLASS_SCHEMA_SLOT}, keyed by generic {@link Type} to support
+     * parameterized structured-output and tool-parameter types. The variants of one raw class (e.g.
+     * {@code List<String>} versus {@code List<Integer>}) share the map held on that raw class, so
+     * these slots are scoped to a classloader in the same way.
+     *
+     * <p>Because unrelated raw classes never share a map, a miss takes {@link #SCHEMA_LOCK} without
+     * grouping unrelated types behind the same lock.
+     */
+    private static final ClassValue<Map<Type, JsonNode>> TYPE_SCHEMA_SLOT =
```
oss-maintainer, line 149: **[Warning]** The `TYPE_SCHEMA_SLOT` map has no size bound or eviction, and the javadoc's original bound argument no longer applies to it.

The old `CLASS_SCHEMA_CACHE` javadoc justified being unbounded because keys were compile-time-fixed classes. Entries here are now keyed by `Type` *variant* within a raw class, so a raw class that is commonly parameterized (`List`, `Map`, `Optional`) accumulates one entry per distinct generic signature ever passed to `generateSchemaFromType`. In the multi-tenant / runtime-loaded scenarios this PR explicitly motivates — where the classloader-leak fix is the whole point — those signatures can be produced per tenant or per dynamically loaded skill rather than being fixed at compile time.

The leak direction is genuinely improved (a slot no longer pins its class), but the per-class entry count is now the unbounded dimension. Worth either a stated bound or a sentence in the javadoc recording that this is deliberate and why the variant count stays small in practice.

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
```diff
@@ -156,6 +199,84 @@ public static Map<String, Object> generateSchemaFromType(Type type) {
...
+
+    /**
+     * Returns the raw class whose cache a type's schema belongs to, or {@code null} when the type
+     * has no raw class and therefore cannot be cached.
+     *
+     * @throws NullPointerException if the type is null
+     */
+    private static Class<?> rawTypeOf(Type type) {
+        if (type == null) {
+            throw new NullPointerException("type must not be null");
```
oss-maintainer: **[Info]** `rawTypeOf` throws `NullPointerException` for a null argument, and `generateSchemaFromType(null)` relies on that to keep the pre-cache contract (covered by `testGenerateSchemaFromTypeNullThrows`).

That works, but the public method's null contract is now enforced by a private helper's exception message (`"type must not be null"`) rather than at the entry point. `Objects.requireNonNull(type, "type")` as the first statement of `generateSchemaFromType` expresses the same guarantee where a caller reading the public signature would look, and keeps the test from silently passing if the helper is ever inlined away.

oss-maintainer (commented):
### Summary

Replaces the two static `ConcurrentHashMap` schema caches with `ClassValue` slots, so a cached schema is held on the class it describes instead of being strongly referenced by a static map key. That stops an entry from keeping a class — and its defining classloader — reachable, which is the right fix for the runtime-loaded-skill and per-tenant-classloader cases. Left `COMMENT` over the uncached-fallback cost and the unbounded per-class `Type` map; both are judgement calls, not bugs.

The double-checked read is correct: `AtomicReference` and `ConcurrentHashMap` give the needed visibility, the slot is re-read under `SCHEMA_LOCK` so a contended class is generated exactly once, and the change also removes the pre-existing hazard of running a long lock-holding mapping function inside `ConcurrentHashMap.computeIfAbsent`'s bin lock. The new nested-mutation tests are a genuinely valuable addition — `testGenerateSchemaFromClassNestedMutationDoesNotAffectLaterCalls` pins the *nested* copy invariant the whole cache rests on, where the previous tests only checked the top level.

### Findings

- **[Warning]** `JsonSchemaUtils.java:236` — types with no raw class (type variables, wildcards, generic arrays) now regenerate under the global `SCHEMA_LOCK` on **every call**, where the old `Map<Type, JsonNode>` cached them. Correctness is covered by `testGenerateSchemaFromTypeVariableStillGenerates`; cost is not asserted anywhere. Because this path holds the *process-wide* lock for a full generation, it also stalls schema generation for unrelated classes. Keying a small `ConcurrentHashMap<Type, JsonNode>` fallback off the `Type` itself is cheap and removes the cliff; if instead you can confirm no reachable structured-output path produces such a `Type`, say so in the javadoc so the next reader does not have to re-derive it.
- **[Warning]** `JsonSchemaUtils.java:116` — `TYPE_SCHEMA_SLOT` is unbounded and unevicted, and the original "keys are compile-time-fixed classes" bound argument no longer holds for it: entries are keyed by `Type` *variant* within a raw class, so a commonly-parameterized raw class (`List`, `Map`, `Optional`) accumulates one entry per distinct generic signature ever seen. In exactly the multi-tenant / dynamically-loaded scenario that motivates this PR, those signatures may be produced per tenant or per runtime-loaded skill. The leak direction is improved; the per-class entry count is the new unbounded dimension and deserves either a stated bound or a documented rationale.
- **[Info]** `JsonSchemaUtils.java:268` — the public null contract of `generateSchemaFromType` is now enforced by a private helper's exception. `Objects.requireNonNull(type, "type")` as the first statement of the public method expresses the same guarantee where a caller reading the signature would look.

### CI note

`build (windows-latest)` is failing and `build (ubuntu-latest)` is `CANCELLED`. The windows failure is `LocalFilesystemPersonalAssistantExampleTest.localFilesystem_filesPersistAcrossCalls:103` ("MEMORY.md content should be persisted on disk ==> expected: <true> but was: <false>") in `agentscope-harness` — untouched by this diff, and recent `main` runs are green, so it reads as a windows-specific flake rather than a regression here. Still needs a green or maintainer-explained build before merge; and since this PR is specifically about lock contention, the `CANCELLED` ubuntu job is worth a re-run rather than ignoring.

### Tests

A concurrency test exercising `generateSchemaFromType` on two variants of the same raw class would document the shared-map behaviour the new javadoc describes.

---

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
oss-maintainer: **[Warning]** Uncacheable types now regenerate on **every call while holding the global `SCHEMA_LOCK`**.

Type variables, wildcards and generic arrays have no raw class to hang a slot on, so they bypass caching entirely and take the process-wide lock for the full duration of `schemaGenerator.generateSchema(type)`. Previously `TYPE_SCHEMA_CACHE` was a plain `Map<Type, JsonNode>` and cached these types like any other, so this is a latency regression on the fallback path — and because it holds the shared lock, it also stalls schema generation for every *other* class in the JVM.

The added test (`testGenerateSchemaFromTypeVariableStillGenerates`) proves the path is correct but asserts nothing about cost. Two options:
1. key a second `ClassValue`-free `ConcurrentHashMap<Type, JsonNode>` fallback off the `Type` itself (these types are interned/`equals`-stable, so they are fine as keys), or
2. confirm with a benchmark that no reachable structured-output path produces a `TypeVariable`/`WildcardType`, and say so in the javadoc so the next reader does not have to re-derive it.

Option 1 is cheap and removes the cliff.

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
oss-maintainer, line 149: **[Warning]** The `TYPE_SCHEMA_SLOT` map has no size bound or eviction, and the javadoc's original bound argument no longer applies to it.

The old `CLASS_SCHEMA_CACHE` javadoc justified being unbounded because keys were compile-time-fixed classes. Entries here are now keyed by `Type` *variant* within a raw class, so a raw class that is commonly parameterized (`List`, `Map`, `Optional`) accumulates one entry per distinct generic signature ever passed to `generateSchemaFromType`. In the multi-tenant / runtime-loaded scenarios this PR explicitly motivates — where the classloader-leak fix is the whole point — those signatures can be produced per tenant or per dynamically loaded skill rather than being fixed at compile time.

The leak direction is genuinely improved (a slot no longer pins its class), but the per-class entry count is now the unbounded dimension. Worth either a stated bound or a sentence in the javadoc recording that this is deliberate and why the variant count stays small in practice.

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
oss-maintainer: **[Info]** `rawTypeOf` throws `NullPointerException` for a null argument, and `generateSchemaFromType(null)` relies on that to keep the pre-cache contract (covered by `testGenerateSchemaFromTypeNullThrows`).

That works, but the public method's null contract is now enforced by a private helper's exception message (`"type must not be null"`) rather than at the entry point. `Objects.requireNonNull(type, "type")` as the first statement of `generateSchemaFromType` expresses the same guarantee where a caller reading the public signature would look, and keeps the test from silently passing if the helper is ever inlined away.

oss-maintainer (commented):
### Summary

Replaces the two static `ConcurrentHashMap` schema caches with `ClassValue` slots, so a cached schema is held on the class it describes instead of being strongly referenced by a static map key. That stops an entry from keeping a class — and its defining classloader — reachable, which is the right fix for the runtime-loaded-skill and per-tenant-classloader cases. Left `COMMENT` over the uncached-fallback cost and the unbounded per-class `Type` map; both are judgement calls, not bugs.

The double-checked read is correct: `AtomicReference` and `ConcurrentHashMap` give the needed visibility, the slot is re-read under `SCHEMA_LOCK` so a contended class is generated exactly once, and the change also removes the pre-existing hazard of running a long lock-holding mapping function inside `ConcurrentHashMap.computeIfAbsent`'s bin lock. The new nested-mutation tests are a genuinely valuable addition — `testGenerateSchemaFromClassNestedMutationDoesNotAffectLaterCalls` pins the *nested* copy invariant the whole cache rests on, where the previous tests only checked the top level.

### Findings

- **[Warning]** `JsonSchemaUtils.java:236` — types with no raw class (type variables, wildcards, generic arrays) now regenerate under the global `SCHEMA_LOCK` on **every call**, where the old `Map<Type, JsonNode>` cached them. Correctness is covered by `testGenerateSchemaFromTypeVariableStillGenerates`; cost is not asserted anywhere. Because this path holds the *process-wide* lock for a full generation, it also stalls schema generation for unrelated classes. Keying a small `ConcurrentHashMap<Type, JsonNode>` fallback off the `Type` itself is cheap and removes the cliff; if instead you can confirm no reachable structured-output path produces such a `Type`, say so in the javadoc so the next reader does not have to re-derive it.
- **[Warning]** `JsonSchemaUtils.java:116` — `TYPE_SCHEMA_SLOT` is unbounded and unevicted, and the original "keys are compile-time-fixed classes" bound argument no longer holds for it: entries are keyed by `Type` *variant* within a raw class, so a commonly-parameterized raw class (`List`, `Map`, `Optional`) accumulates one entry per distinct generic signature ever seen. In exactly the multi-tenant / dynamically-loaded scenario that motivates this PR, those signatures may be produced per tenant or per runtime-loaded skill. The leak direction is improved; the per-class entry count is the new unbounded dimension and deserves either a stated bound or a documented rationale.
- **[Info]** `JsonSchemaUtils.java:268` — the public null contract of `generateSchemaFromType` is now enforced by a private helper's exception. `Objects.requireNonNull(type, "type")` as the first statement of the public method expresses the same guarantee where a caller reading the signature would look.

### CI note

`build (windows-latest)` is failing and `build (ubuntu-latest)` is `CANCELLED`. The windows failure is `LocalFilesystemPersonalAssistantExampleTest.localFilesystem_filesPersistAcrossCalls:103` ("MEMORY.md content should be persisted on disk ==> expected: <true> but was: <false>") in `agentscope-harness` — untouched by this diff, and recent `main` runs are green, so it reads as a windows-specific flake rather than a regression here. Still needs a green or maintainer-explained build before merge; and since this PR is specifically about lock contention, the `CANCELLED` ubuntu job is worth a re-run rather than ignoring.

### Tests

A concurrency test exercising `generateSchemaFromType` on two variants of the same raw class would document the shared-map behaviour the new javadoc describes.

---

jnduan:
Thanks for the review — all three points are addressed, and the `rawTypeOf` miss was the sharpest catch.

**On the classifier miss (no raw class → recompute under the lock)**

Confirmed reachable rather than theoretical: `ToolSchemaGenerator:129` passes `Parameter.getParameterizedType()`, which yields a `TypeVariable` for a generic method parameter such as `<T> void tool(@ToolParam T input)`.

Instead of the suggested static fallback map, I made the attribution itself complete. `rawTypeOf` is now `scopeClassOf`, which resolves a scope class for every type a signature can declare:

- `Class` → itself
- `ParameterizedType` → its raw class
- `GenericArrayType` → its component type, recursively
- `TypeVariable` → its declaring class, or the declaring class of its declaring method

A static `Map<Type, JsonNode>` fallback would have reintroduced the `Type` → declaration strong reference that the previous round removed, so I kept the fix inside the `ClassValue` scoping: these types are cached on the class that declares them and take the lock only on a miss. Only a type with no attributing class at all — a wildcard synthesized by a caller — still generates under the lock on every call, and that path is no longer reachable from a reflective signature. The javadoc on `cachedSchemaNode(Type)` and a test record it.

**On the unbounded-map concern**

The bound is now stated explicitly in the class javadoc: entries are bounded by the distinct generic signatures written against a class *in code*, since every `Type` reaching this utility originates in a `TypeReference` literal or a reflective method signature, and structurally equal signatures share one entry. Loading a class at runtime adds a class with its own cache rather than another entry on an existing one. A caller synthesizing `Type` instances can exceed that bound, but those entries are released together with the owning class.

**On the public null contract**

`generateSchemaFromType` now opens with `Objects.requireNonNull(type, "type")`, and the javadoc documents `@throws NullPointerException`.

**Tests**

- `testGenerateSchemaFromTypeConcurrently` now uses two variants of the *same* raw class (`List<ConcurrentClassC>` / `List<ConcurrentClassD>`) instead of two different ones, so it actually exercises the shared slot map, and asserts the two variants yield different schemas.
- Added coverage for the three newly attributed shapes (class-level type variable, method-level type variable, generic array) and for the uncached path.
- `JsonSchemaUtilsTest`: 19 tests, 0 failures. Full `agentscope-core`: 2345 tests, 0 failures, 0 errors, 9 skipped.

**On the failing check**

`build (windows-latest)` fails in `agentscope-harness`, at `LocalFilesystemPersonalAssistantExampleTest.localFilesystem_filesPersistAcrossCalls:103` — a file-persistence assertion in a module this diff does not touch. `build (ubuntu-latest)` shows cancelled 17s later, i.e. fail-fast cancellation rather than an independent failure.

I could not reproduce it locally: the class passes 5 consecutive runs (3 tests each, 0 failures) on macOS/JDK 17 with the same Maven build. It reads as a windows-specific flake. Submitted from a fork, so I can't re-run the check myself — the new push will trigger a fresh run.

Would appreciate a re-run of the cancelled ubuntu job, or a maintainer's read on the windows failure.

oss-maintainer (commented):
### Summary

Re-review after the two new commits (`e5c404c7`, `90ebf59e`). `rawTypeOf` → `scopeClassOf` is the right shape of fix for the generic-method-parameter case (`ToolSchemaGenerator:129` really does hand over `TypeVariable` instances), the `computeIfAbsent` removal is a genuine improvement, and the new type-variable / method-type-variable / null-contract tests cover exactly the paths my previous review flagged. All CI is green on this head (`build (ubuntu-latest)`, `build (windows-latest)`, `Check License`, `Check Module Sync`), so the windows flake I flagged earlier is gone.

What I still cannot sign off is the claim in the new class javadoc that a cached schema cannot outlive the class it describes. Scoping a parameterized type to its **raw** class keeps the entry on a bootstrap class (`java.util.List`, `java.util.Map`, `java.util.Optional`), which is never unloaded, so the entry — and the app class captured in the `Type` key — is reachable for the lifetime of the JVM. That is the same classloader-pinning the `ClassValue` change was meant to eliminate, just moved one level down. The four inline notes are all cheap to address; the fourth is a nit.

### Findings
- **[Warning]** `JsonSchemaUtils.java:302` — raw-class scoping pins app classes behind JDK collections; prefer the deepest non-bootstrap class in the type.
- **[Warning]** `JsonSchemaUtils.java:82` — the "released together with the owning class" bound does not hold for entries scoped to bootstrap classes; state the exception or fix the scoping.
- **[Info]** `JsonSchemaUtils.java:312` — `Constructor` is a `GenericDeclaration` but not a `Method`; match on `Executable` so constructor-declared type variables are not pushed onto the uncached global-lock path.
- **[Info]** `JsonSchemaUtils.java:266` — a bare `Class` reaching `generateSchemaFromType` is cached twice (once per slot map).

### Not blocking
A concurrency test that hits `generateSchemaFromType` with two variants of the *same* raw class (which the author's note says now exists) is the right regression guard for the shared-map behaviour.

---

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
```diff
@@ -156,6 +219,103 @@ public static Map<String, Object> generateSchemaFromType(Type type) {
...
+     * @param type the type to attribute; must not be {@code null}
+     * @return the class to cache the schema under, or {@code null} if the type has none
+     */
+    private static Class<?> scopeClassOf(Type type) {
+        if (type instanceof Class<?> clazz) {
+            return clazz;
+        }
+        if (type instanceof ParameterizedType parameterizedType
+                && parameterizedType.getRawType() instanceof Class<?> rawClass) {
+            return rawClass;
```
oss-maintainer: **[Warning]** Scoping a parameterized type to its **raw** class defeats the classloader-pinning guarantee this PR exists to provide. For `List<TenantDto>` / `Map<String, SkillDto>` the scope is `java.util.List` / `java.util.Map`, and a `ClassValue` entry on a bootstrap class is only reclaimed when that class is unloaded — i.e. never. So the `Map<Type, JsonNode>` owned by `java.util.List` accumulates one permanent entry per distinct generic signature ever seen, and each key (`ParameterizedTypeImpl`) strongly references the app class, keeping it and its classloader reachable.

This is exactly the runtime-loaded-skill / per-tenant scenario the new class javadoc argues about: such a class does not only "add a class with its own cache", it also adds an entry to the JDK collection's cache that is never released. It is also not hypothetical — `ToolSchemaGenerator:129` feeds `Parameter#getParameterizedType()` straight into `generateSchemaFromType`, so a `@ToolParam List<Foo>` lands here.

Could `scopeClassOf` prefer the deepest non-bootstrap class it can find (raw type plus `getActualTypeArguments()`, recursing through arrays/parameterized types) and fall back to the raw class only when everything is bootstrap-loaded? That keeps `List<Foo>` on `Foo`'s `ClassValue`, where it dies with `Foo`, and leaves genuinely JDK-only signatures (`List<String>`) on the JDK class where the unbounded growth does not matter.

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
```diff
@@ -53,9 +61,25 @@
...
+ * not always compile-time-fixed: extensions can load skills and tools at runtime, and
+ * multi-tenant deployments may load classes per tenant.</p>
+ *
+ * <p>The number of entries a single class's cache can hold is bounded by the distinct generic
+ * signatures the code produces against it: every {@link Type} reaching this utility originates in
+ * a {@link TypeReference} literal or a reflective method signature, and two structurally equal
+ * signatures share one entry. Loading a class at runtime therefore adds a class with its own
+ * cache rather than another entry on an existing one. A caller that synthesizes {@code Type}
+ * instances at runtime can add entries beyond that bound, but those entries are released together
+ * with the class that owns the cache.</p>
```
oss-maintainer: **[Warning]** The bound argument here holds for *classes* but not for the shared caches of JDK raw classes: "entries are released together with the class that owns the cache" is true for an application class, but `java.util.List`, `java.util.Map` and `java.util.Optional` are never unloaded, so entries scoped to them are permanent, and their keys pin the app classes appearing in the type arguments (see the note on `scopeClassOf`). Either scope such types to an app-owned class, or state the exception explicitly so the next reader does not take the javadoc as a general no-leak guarantee.

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
```diff
@@ -156,6 +219,103 @@ public static Map<String, Object> generateSchemaFromType(Type type) {
...
+        }
+        if (type instanceof GenericArrayType arrayType) {
+            return scopeClassOf(arrayType.getGenericComponentType());
+        }
+        if (type instanceof TypeVariable<?> typeVariable) {
+            GenericDeclaration declaration = typeVariable.getGenericDeclaration();
+            if (declaration instanceof Class<?> declaringClass) {
+                return declaringClass;
+            }
+            if (declaration instanceof Method method) {
```
oss-maintainer: **[Info]** `GenericDeclaration` has three implementations, and `Constructor` is not a `Method` (both are `Executable`). A type variable declared by a constructor therefore falls through to `null` and takes the uncached path, i.e. the process-wide `SCHEMA_LOCK` on every call — the cliff this commit set out to remove. Matching on `Executable` instead of `Method` closes it for free: `if (declaration instanceof Executable e) return e.getDeclaringClass();`

agentscope-core/src/main/java/io/agentscope/core/util/JsonSchemaUtils.java
```diff
@@ -156,6 +219,103 @@ public static Map<String, Object> generateSchemaFromType(Type type) {
...
+            // No class to hang a slot on, so this type cannot be cached. Every type a method
+            // signature can declare is attributed, so this path is only reachable for a type a
+            // caller synthesizes: a wildcard is only ever a type argument, never a parameter or
+            // return type.
+            synchronized (SCHEMA_LOCK) {
+                return schemaGenerator.generateSchema(type);
+            }
+        }
+
+        Map<Type, JsonNode> slot = TYPE_SCHEMA_SLOT.get(scope);
```
oss-maintainer, line 287: **[Info]** When `type` is a bare `Class`, `scopeClassOf` returns it, so `generateSchemaFromType(Foo.class)` caches into `TYPE_SCHEMA_SLOT.get(Foo)` while `generateSchemaFromClass(Foo.class)` caches into `CLASS_SCHEMA_SLOT.get(Foo)`. The same schema is then generated and stored twice for the same class. Delegating the `Class` case of `generateSchemaFromType` to the `Class` overload (or routing both through one lookup) keeps one entry per class.

jnduan:
Thanks — the javadoc claim does not survive your counterexample, and raw-class scoping was the wrong placement rather than an acceptable trade-off. I fixed the scoping instead of restating the bound.

**On raw-class scoping (`:302`)**

`scopeClassOf` now takes the deepest application class a type mentions, as you suggested: type arguments are searched before the raw type, recursing through nested parameterized types, generic arrays and wildcard bounds, and the raw class is only the fallback for a signature whose classes all come from the bootstrap loader. `List<TenantDto>` is held on `TenantDto` and dies with it, `Map<String, SkillDto>` on `SkillDto`, `List<? extends Foo>` on `Foo`, while `List<String>` stays on `java.util.List` — the growth there is genuinely unbounded, but nothing application-owned appears in its key. A type variable declared inside a parameterized signature (`List<T>` on a generic class) is attributed the same way instead of falling back to the container.

**On the bound statement (`:82`)**

Rewritten rather than qualified. The class javadoc now says what holds: a signature built only from JDK classes has no application class to be scoped to and stays on its raw class, so those entries live as long as the JVM, and they cannot pin an application class because none appears in them. The remaining honest limit is written down too: when a type mentions several application classes, its single key is held on the first one and all of them are released together with it. `TYPE_SCHEMA_SLOT`'s javadoc points at the same argument.

**On `Constructor` (`:312`)**

Fixed via `Executable`, with the resolution moved into `declaringClassOf(TypeVariable)`, so class, method and constructor declarations are one path. You were right that this is a cliff rather than a corner: `<T> Foo(T value)` is a signature the tool path can hand over directly.

**On the double cache (`:266`)**

`generateSchemaFromType` delegates a bare `Class` to `generateSchemaFromClass`, so a class reached through both entry points is generated once and stored in one slot. The error message on that path is now the class one, which is also the more useful of the two.

**Tests**

Four added, and I checked they are worth having: with the production file reverted to `90ebf59e` each of them fails, and with the fix restored all 23 pass. They assert the slot maps directly (`TYPE_SCHEMA_SLOT` / `CLASS_SCHEMA_SLOT` by reflection, with a comment saying why: every schema the public API returns is a fresh copy, so the entry behind it is not observable in behaviour).

One consequence you should know about: the new scoping moves `List<ConcurrentClassC>` and `List<ConcurrentClassD>` onto different slots, so the concurrency test no longer exercised a shared map. It now runs `List<String>` / `List<Integer>` — same raw class, both JDK-only, both on the `java.util.List` slot — alongside those two variants, and asserts the shared pair really are on one slot and stay independent.

`JsonSchemaUtilsTest`: 23 tests, 0 failures. Full `agentscope-core`: 2349 tests, 0 failures, 0 errors, 9 skipped. `spotless:apply` clean; the diff is confined to the two files.

Pushed as `602c3c42`.

oss-maintainer (approved):
### Summary
Re-review after the new commit `602c3c42` (`refactor(core): scope schema cache to a class that can be unloaded`). It resolves exactly the concern raised on the previous head: a parameterized type is no longer parked on its raw class, so `List<TenantDto>` is now cached on `TenantDto` and dies together with that class instead of pinning the tenant classloader through `java.util.List`.

### What was verified
- `applicationClassIn` walks type arguments before the raw type, and covers `GenericArrayType`, `WildcardType` (upper then lower bounds) and `TypeVariable` via its declaring class; JDK/bootstrap classes (`getClassLoader() == null`) are correctly skipped, and the JDK-only fallback to the raw class is documented, including why that entry cannot pin an application class.
- The `TypeVariable` case now matches `Executable`, not just `Method`, so constructor-declared type variables stop falling through to the uncached path and re-taking the global lock.
- `generateSchemaFromType(Class)` delegates to `generateSchemaFromClass`, so both entry points share one slot instead of generating and storing the same schema twice.
- The three new tests assert slot placement positively *and* negatively (`typeSlot(TenantElement)` contains the entry, `typeSlot(List)` does not), which is the assertion that actually guards the classloader-pinning property.
- Docs in the class javadoc were updated in step with the behaviour, including the one unbounded case (JDK-only signatures) — the previous objection was documentation drift, and it is gone.

CI: build (ubuntu + windows), license/CLA and module sync all green.

---

PR #3044, opened 2026-09-08
fix(harness): preserve bytes in cross-backend file moves

### LINKED ISSUE #3042: [Bug]: Cross-backend CompositeFilesystem.move rewrites file bytes and deletes the source
**Describe the bug**

When paths are routed to different backends, `CompositeFilesystem.move()` copies `read().fileData().content()` through string `write()`, then deletes the source. Local reads return Base64 for binary files and a reminder for empty/blank text, so a successful move can change the file bytes.

**To Reproduce**

Run with `io.agentscope:agentscope-harness:2.0.3`. Only local temporary directories are used.

```java
import io.agentscope.core.agent.RuntimeContext;
import io.agentscope.harness.agent.filesystem.CompositeFilesystem;
import io.agentscope.harness.agent.filesystem.local.LocalFilesystem;
import java.nio.file.Files;
import java.nio.file.Path;
import java.util.Arrays;
import java.util.Map;

public class MoveRepro {
    public static void main(String[] args) throws Exception {
        Path source = Files.createTempDirectory("move-source-");
        Path target = Files.createTempDirectory("move-target-");
        var fs = new CompositeFilesystem(
                new LocalFilesystem(source, true, 10),
                Map.of("/target/", new LocalFilesystem(target, true, 10)));
        byte[] expected = new byte[0];
        Files.write(source.resolve("empty.txt"), expected);
        var result = fs.move(RuntimeContext.empty(),
                "/empty.txt", "/target/empty.txt");
        byte[] actual = Files.readAllBytes(target.resolve("empty.txt"));
        System.out.println("success=" + result.isSuccess());
        System.out.println("sourceExists=" + Files.exists(source.resolve("empty.txt")));
        System.out.println("outputBytes=" + actual.length);
        System.out.println("bytesPreserved=" + Arrays.equals(expected, actual));
    }
}
```

Observed:

```text
success=true
sourceExists=false
outputBytes=51
bytesPreserved=false
```

The target contains `System reminder: File exists but has empty contents`. Also reproduced with synthetic files in 2.0.3 and relevant unmodified main classes:

| Input | Cross-backend output |
| --- | --- |
| 68-byte PNG | 92 bytes of Base64 text |
| Empty text | 51-byte reminder |
| Whitespace-only text | 51-byte reminder |
| Ordinary UTF-8 / CRLF text | Original bytes preserved |

Moving the PNG within the same LocalFilesystem preserves its bytes.

**Expected behavior**

Successful file moves preserve original bytes, including binary and empty/blank files. Copy failures must retain the source.

**Environment**

- AgentScope-Java: 2.0.3; main checked at `6ba5fefe0484ea3dda37cbeeb97b754f471ce7df`
- Java: 17.0.19
- OS: macOS arm64

**Additional context**

The implementation dates back to #1353. I am preparing a fix using raw byte transfer while retaining create-only destination behavior, with byte-for-byte regression tests.

This is separate from #2447 / #2456, which address `RemoteFilesystem.uploadFiles()` encoding. This reproduction uses only LocalFilesystem, and the current move path calls string `write()` instead of `uploadFiles()`.

LinJinl:
## AgentScope-Java Version

2.0.3-SNAPSHOT, synchronized with main at `97696a3a`.

## Description

Fixes #3042.

Cross-backend `CompositeFilesystem.move` copies the display-oriented `read` result, which can contain Base64 or an empty-file reminder instead of the original bytes. A 68-byte PNG becomes 92 bytes of Base64 text, and an empty file becomes 51 bytes of reminder text; the source is then deleted.

This change downloads the original bytes, uploads them with `UploadMode.CREATE_NEW`, and deletes the source only after a successful upload. It reports download, upload, and deletion failures. If source deletion fails after the copy, the destination is retained and the result identifies the partial failure. Cross-backend moves remain non-atomic, and the original upload overload keeps its existing behavior.

The API Javadoc and English and Chinese filesystem docs describe the supported backends and limits:

- Local, including LocalWithShell, and Remote preserve arbitrary bytes. Remote uses reversible UTF-8 or Base64 encoding and requires `BaseStore.putIfVersion(..., 0)` for conditional creation.
- Composite, BakedContext, and RoutedSandbox forward the selected backend's capability. Overlay and SandboxBacked, including PinnedSandbox, currently use the strict UTF-8 default for create-only uploads and reject other byte sequences.
- Concurrent-writer protection depends on the backend's atomic creation support. Overlay may create an upper-layer file that shadows a lower-layer file without changing the lower-layer original.
- A backend without raw downloads, such as the service's `MemoryStoreFilesystem`, fails the move without deleting the source.

Includes 60 regression cases for byte preservation, empty and whitespace-only text, existing destinations, competing writers, transfer and deletion failures, legacy uploads, wrapper routing, and runtime context. The branch includes the upstream truncated-download safeguard from #2923, with three additional cases that verify source retention, unchanged destinations, and continuation of later downloads in a batch.

## Validation

- Before the original fix, the initial 27-case regression suite had 16 expected failures. The three sandbox truncation cases also failed on `c54be700`, reproducing a 400,000-byte source reduced to 388,110 bytes and then deleted; they pass with the upstream safeguard.
- `mvn -B -pl agentscope-harness -am spotless:apply clean verify` passed on `5b0d6102`: Core 2,344 tests / 9 skipped; AG-UI 509 tests / 0 skipped; Harness 1,100 tests / 14 skipped; zero failures or errors. This included Spotless checks and Javadoc generation.
- The final commit `59299437` changes only the two filesystem docs. A fresh targeted run of `CompositeFilesystemMoveTest`, `FilesystemUploadModeTest`, `BackendUploadModeFailureTest`, and `SandboxBackedFilesystemTest` passed all 77 cases, including all 60 PR regression cases, with `spotless:check`.
- `git diff --check` passed. All seven upstream checks passed for the final head `59299437`: Linux and Windows builds, Codecov patch coverage, documentation validation, license, module sync, and CLA. See [CI](https://github.com/agentscope-ai/agentscope-java/actions/runs/34975957830).

## Checklist

- [x] All tests are passing (`mvn test`) — local validation and upstream Linux and Windows builds passed for the final head.

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/AbstractFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/AbstractFilesystem.java
@@ -25,9 +25,15 @@ import io.agentscope.harness.agent.filesystem.model.GlobResult;
 import io.agentscope.harness.agent.filesystem.model.GrepResult;
 import io.agentscope.harness.agent.filesystem.model.LsResult;
 import io.agentscope.harness.agent.filesystem.model.ReadResult;
+import io.agentscope.harness.agent.filesystem.model.UploadMode;
 import io.agentscope.harness.agent.filesystem.model.WriteResult;
+import java.nio.ByteBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * Abstract filesystem API for agents: list, read, write, edit, grep, glob, upload, download.
@@ -122,6 +128,56 @@ public interface AbstractFilesystem {
             RuntimeContext runtimeContext, List<Map.Entry<String, byte[]>> files);
 
     /**
+     * Upload files using the requested destination write policy.
+     *
+     * <p>{@link UploadMode#OVERWRITE} delegates to the existing upload method. The default
+     * {@link UploadMode#CREATE_NEW} implementation accepts strictly valid UTF-8 bytes and uses
+     * {@link #write}, preserving that backend's creation and layer semantics. Backends must override
+     * this method to support other byte sequences in create-only mode. Atomic creation depends on
+     * the backend; this default does not add locking or a transaction.
+     *
+     * @param runtimeContext per-call agent runtime
+     * @param files path-to-content mappings; an empty byte array represents an empty file
+     * @param mode destination write policy
+     * @return one response per input file, in the same order
+     */
+    default List<FileUploadResponse> uploadFiles(
+            RuntimeContext runtimeContext, List<Map.Entry<String, byte[]>> files, UploadMode mode) {
+        Objects.requireNonNull(mode, "mode");
+        if (mode == UploadMode.OVERWRITE) {
+            return uploadFiles(runtimeContext, files);
+        }
+        List<FileUploadResponse> responses = new ArrayList<>(files.size());
+        for (Map.Entry<String, byte[]> file : files) {
+            if (file.getValue() == null) {
+                responses.add(
+                        FileUploadResponse.fail(file.getKey(), "File content must not be null"));
+                continue;
+            }
+            String content;
+            try {
+                content =
+                        StandardCharsets.UTF_8
+                                .newDecoder()
+                                .decode(ByteBuffer.wrap(file.getValue()))
+                                .toString();
+            } catch (CharacterCodingException e) {
+                responses.add(
+                        FileUploadResponse.fail(
+                                file.getKey(),
+                                "Backend does not support create-only binary uploads"));
+                continue;
+            }
+            WriteResult result = write(runtimeContext, file.getKey(), content);
+            responses.add(
+                    result.isSuccess()
+                            ? FileUploadResponse.success(file.getKey())
+                            : FileUploadResponse.fail(file.getKey(), result.error()));
+        }
+        return responses;
+    }
+
+    /**
      * Download multiple files.
      *
      * @param runtimeContext per-call agent runtime; {@link RuntimeContext#empty()} when none
@@ -145,8 +201,9 @@ public interface AbstractFilesystem {
      * Move (rename) a file or directory from {@code fromPath} to {@code toPath}.
      *
      * <p>Implementations that span multiple stores (e.g. {@code CompositeFilesystem}) may
-     * fall back to a read + write + delete sequence when source and destination live in
-     * different backend filesystems.
+     * copy the original bytes before deleting the source. Cross-backend file moves require raw
+     * download support and create-only upload support for the source content. They are not atomic;
+     * a source deletion failure can leave both copies and must be reported as a failure.
      *
      * @param runtimeContext per-call agent runtime; {@link RuntimeContext#empty()} when none
      * @param fromPath absolute source path
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/BakedContextFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/BakedContextFilesystem.java
@@ -23,6 +23,7 @@ import io.agentscope.harness.agent.filesystem.model.GlobResult;
 import io.agentscope.harness.agent.filesystem.model.GrepResult;
 import io.agentscope.harness.agent.filesystem.model.LsResult;
 import io.agentscope.harness.agent.filesystem.model.ReadResult;
+import io.agentscope.harness.agent.filesystem.model.UploadMode;
 import io.agentscope.harness.agent.filesystem.model.WriteResult;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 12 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/AbstractFilesystem.java
```diff
@@ -121,6 +127,56 @@ EditResult edit(
...
+     * this method to support other byte sequences in create-only mode. Atomic creation depends on
+     * the backend; this default does not add locking or a transaction.
+     *
+     * @param runtimeContext per-call agent runtime
+     * @param files path-to-content mappings; an empty byte array represents an empty file
+     * @param mode destination write policy
+     * @return one response per input file, in the same order
+     */
+    default List<FileUploadResponse> uploadFiles(
+            RuntimeContext runtimeContext, List<Map.Entry<String, byte[]>> files, UploadMode mode) {
```
oss-maintainer, line 145: The default `uploadFiles(..., CREATE_NEW)` accepts only strictly valid UTF-8 and otherwise fails, which is the right safe behaviour — but it makes "binary cross-backend move" silently unsupported for every backend that does not override it. Since `LocalFilesystem` and `RemoteFilesystem` are the only overriders here, could you add a one-line list to the javadoc of which built-in backends preserve raw bytes today, so callers do not have to read implementations to find out? Sandbox-backed filesystems in particular look like the next candidates.

oss-maintainer (commented):
### Summary
Fixes genuine data corruption in cross-backend `CompositeFilesystem.move`, which round-tripped the display-oriented `read` string (Base64 / empty-file reminder text) instead of raw bytes. Downloading bytes and pushing them through a create-only upload, deleting the source only after success, is the right ordering, and the failure-path coverage is unusually thorough. Main ask is discoverability of which built-in backends actually preserve raw bytes with the new default `uploadFiles(.., CREATE_NEW)`.

---

- **[INFO]** `agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/CompositeFilesystem.java:1` — Confirmed the move sequence is upload-then-delete and remains explicitly non-atomic (stated in the javadoc) — good that a failed source delete now returns a partial-copy error and retains the destination. Follow-up worth an issue rather than this PR: on partial failure the copy is left behind at the destination with no marker, so a retried move now hits the create-only path and reports "already exists". A `force`/overwrite recovery hint in the error text would make this self-service. _(line outside diff)_

oss-maintainer (approved):
### Summary

`5b0d6102` and `59299437` close the only thing I was holding this back on: the new javadoc block on `uploadFiles(.., CREATE_NEW)` now names exactly which built-in backends preserve raw bytes (`LocalFilesystem` / `LocalFilesystemWithShell` / `RemoteFilesystem`), which ones forward the capability (`CompositeFilesystem`, `BakedContextFilesystem`, `RoutedSandboxFilesystem`) and which ones are still on the UTF-8-only default (`OverlayFilesystem`, `SandboxBackedFilesystem` incl. `PinnedSandboxFilesystem`). An operator sizing a cross-backend move now knows upfront whether it will work instead of finding out from a failure. LGTM.

The `move(...)` javadoc rewording is a genuine improvement too: the old "An existing destination is not overwritten" was a guarantee that the default `CREATE_NEW` path does not actually provide on a backend whose `write()` has overwrite semantics — the new "uses its backend's create-only semantics" plus the explicit non-atomicity paragraph states what callers can really rely on. Same correction applied to both the English and Chinese docs, and the new Overlay paragraph (creating an upper-layer file shadows a lower-layer file at the same path without touching the original) documents a distinction that was previously implicit.

The added tests are the right ones: `downloadFiles_rejectsTruncatedOutputAndContinuesBatch` pins that a truncated download is per-file and does not poison the rest of the batch, and `crossBackendMove_rejectsTruncatedDownloadAndPreservesFiles` is parameterised over destination-exists so a failed transfer cannot delete the source or clobber the destination. Both assert on the actual retained bytes rather than just the returned status.

### Notes for follow-up (not blocking)

- `AbstractFilesystem.java` — the default `CREATE_NEW` path still reports `"Backend does not support create-only binary uploads"` for non-UTF-8 bytes without naming the backend. When a `CompositeFilesystem` route fans out to several backends, that message alone is hard to act on; including the backend class would make the failure self-explanatory.
- The per-file loop in the default method is sequential, unlike the batched path in `CompositeFilesystem.uploadFiles`. Fine for current callers, but worth remembering if `uploadFiles` ever becomes hot for large bundles.

### CI / Merge state

- All checks green on `59299437` (build ubuntu + windows, Check License, Check Module Sync, validate, codecov/patch).
- CLA signed. `mergeable=MERGEABLE`; `mergeStateStatus` still reflects outstanding required review, which a maintainer needs to clear.

Thanks for the thorough turnaround on this one, and for keeping the code, the javadoc and both doc languages in step.

---

oss-maintainer:

```bash
git fetch origin
git checkout codex/fix-cross-backend-move
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

---

oss-maintainer (approved):
### Summary

Re-review of the new head `1b77856` after the `chore: sync main and resolve remote upload conflict` push. The sync brings in `main`'s strict UTF-8 decoding (`CodingErrorAction.REPORT`) for the existing 2-arg `RemoteFilesystem.uploadFiles` OVERWRITE path; nothing in this PR's own logic changed since my previous approval on `5929943` (the byte-preserving download -> create-only upload -> delete ordering, the create-only capability matrix in the Javadoc, and the EN/ZH doc updates are all still intact). The PR's `CREATE_NEW` path uses a `newDecoder()` whose defaults are already `REPORT`, so both paths agree on rejecting malformed input rather than silently replacing it.

Verified on the new head:
- `license/cla` = success, CI = 7/7 passing (Java CI with Maven, Mintlify docs).
- Tests still pin the important failure semantics: `moveDoesNotOverwriteExistingTarget`, `moveReportsSourceDeletionFailureAndPreservesBothCopies`, `movePreservesSourceWhen*`, `concurrentMovesOnlyDeleteTheWinningSource`, and the Overlay UTF-8-only rejection path.

No new findings. LGTM — the stale-approval risk from the rebased head is now cleared, so this is good to go from a review standpoint.

---

PR #3046, opened 2026-09-08
fix(agui): reconcile replayed tool results with session state

### LINKED ISSUE #3043: [Bug]:: Message: Invalid tool result IDs: [call_01_w7wlz4ch9Hiw4Um74zwv8117]. Expected: [call_00_OugZmGqpB71ykFO5IjxM8342]
（大模型调试了一天发现的问题）
Describe the bug Resuming a run with external-execution ("client/frontend") tool results can fail with IllegalStateException: Invalid tool result IDs: [...] Expected: [...], permanently wedging the session.

ReActAgent.CallExecution.validateAndAddToolResults fails the whole run whenever any submitted tool-result ID is not in the pending tool-call IDs of the agent's last assistant message. With AG-UI clients that replay their full transcript (e.g. CopilotKit headless + browser-side frontend tools), a later user message can carry stale tool results from an earlier assistant round (the round whose result was already consumed or whose required result lives mid-transcript). After the first such submission the run dies with a RUN_ERROR, and because the stale result stays in the client transcript, every subsequent user message in the thread fails the same validation forever.

Observed across multiple reproductions:

Invalid tool result IDs: [call_01_7icKHvTMy32ReIAnjkgu2856]. Expected: [call_00_EEGrx8VtxFIpkJarXNXD6198]
Invalid tool result IDs: [call_01_SxJ4HtkNkradgAcGZZBj3288]. Expected: [call_00_BibzeGa9V4O4xDeIvZju4111]
The "Expected" ID always refers to an assistant tool-call message that sits earlier in the client transcript (its result was trimmed away by AguiRequestProcessor.extractLatestUserMessage, which only forwards messages that trail the client's last assistant message), while the "Provided" ID belongs to a newer round that the server is no longer waiting for.

To Reproduce Steps to reproduce the behavior:

Stand up an AG-UI runtime (AguiMvcController) whose agent (ReActAgent, stateful per thread) exposes a mix of server tools and client-executed tools (schema-only / ToolSuspendException tools).
Drive it from a transcript-replaying client that executes client tools in the browser:
Java

// First run -> assistant message asks for a client tool "external_api" (id ext1), run suspends.
agent.streamEvents(List.of()).blockLast();
// Resume #1 -> client replies with the correct result for ext1; server resumes.
agent.streamEvents(List.of(toolResultMsg("ext1", "ok"))).blockLast();
// Later: agent produces ANOTHER assistant tool round, then the user sends a new message
// while the client transcript still replays the (already consumed) "ext1" result as trailing.
Msg stale = Msg.builder().role(MsgRole.TOOL)
    .content(ToolResultBlock.builder().id("ext1")...build()).build();
Msg fresh = Msg.builder().role(MsgRole.USER).textContent("please continue").build();
agent.streamEvents(List.of(stale, fresh)).collectList().block(); // throws
See error:

IllegalStateException: Invalid tool result IDs: [ext1]. Expected: [<current pending id>]
(In production the payload shape is the AG-UI run resume body; the stack surfaces on the client as a RUN_ERROR event, code agent_run_error_event.)

Trigger conditions observed in real traffic:

The model emits more than one tool call in a single assistant message (e.g. one client tool + one server tool in the same reply). The client splits them into separate local assistant rounds, submits them in separate resume requests, and later replays already-accepted results.
A new user message is sent while an older external-execution round is still "pending" server-side but already resolved on the client.
Expected behavior A resumed request that carries tool results should be tolerant of stale results:

Tool-result IDs that are not among the current pending IDs should be ignored (logged), not treated as a fatal protocol violation.
If the input only contains stale results plus a real user text message, the abandoned pending tool round should be cancelled (synthetic error result, same as the existing enablePendingToolRecovery behavior) and the new user message should proceed as a fresh instruction — instead of throwing forever on every later turn.
Duplicate IDs and "partial results + text" should still be rejected exactly as today.
Error messages

java.lang.IllegalStateException: Invalid tool result IDs: [call_01_7icKHvTMy32ReIAnjkgu2856]. Expected: [call_00_EEGrx8VtxFIpkJarXNXD6198]
    at io.agentscope.core.ReActAgent$CallExecution.validateAndAddToolResults(ReActAgent.java:2221)
Surfaced over AG-UI as:

JSON

{
  "type": "run-error",
  "code": "INVALID_INPUT_ERROR",
  "message": "Invalid tool result IDs: [call_01_...]. Expected: [call_00_...]"
}

AgentScope-Java Version: 2.0.3-SNAPSHOT (also confirmed affected in 2.0.2 — same ReActAgent.validateAndAddToolResults code)
Java Version: [17+ expected; reproduce under your JDK]
OS: Windows
Additional context Root-cause chain:

ReActAgent.validateAndAddToolResults (private CallExecution method) treats every provided result ID not in pendingIds as a hard error.
With server-side memory enabled, AguiRequestProcessor.extractLatestUserMessage forwards only the messages after the client's last assistant message; the earlier round's real result (which the server actually needs) sits mid-transcript and is dropped.
Because the client keeps replaying that stale round on every subsequent request, the session never recovers — one bad round bricks all later turns in the thread.
Proposed fix (minimal, in validateAndAddToolResults):

Partition provided results into valid (in pendingIds) and stale (not pending). Log stale IDs at warn and drop them instead of throwing.
If no valid results remain and the input carries a non-tool-result user message, synthesize an ToolResultBlock.error for each still-pending call (same pattern as maybePatchPendingToolCalls / the shutdown-interrupt recovery) and add only the cleaned messages, so the new user instruction is handled by a fresh reasoning step.
Only throw when no valid results remain and there is no user text.
Strip stale ToolResultBlocks from the messages before appending to context to avoid unbounded duplicate growth.
A regression test for this shape (stale result + new user text resumes successfully and cancels the abandoned pending call) was validated against a local patched build of ReActAgentNewLoopReplyTest; all related suites (ReActAgentHitlTest, ReActAgentSummarizingTest, ReActAgentNewLoopReplyTest) stayed green. Happy to open a PR with the fix and test if maintainers agree with the direction.

**Error messages**

Code: agent_run_error_event
Message: Invalid tool result IDs: [call_01_w7wlz4ch9Hiw4Um74zwv8117]. Expected: [call_00_OugZmGqpB71ykFO5IjxM8342]
Context: {
  "source": "onRunErrorEvent",
  "event": {
    "type": "RUN_ERROR",
    "timestamp": 1788838954967,
    "message": "Invalid tool result IDs: [call_01_w7wlz4ch9Hiw4Um74zwv8117]. Expected: [call_00_OugZmGqpB71ykFO5IjxM8342]",
    "code": "INVALID_INPUT_ERROR",
    "threadId": "476270bb-8bd3-4d2d-a96b-7519a0cce1b5",
    "runId": "937f2963-789c-4310-96c3-907c90d8c436"
  },
  "runtimeErrorCode": "INVALID_INPUT_ERROR",
  "agentId": "f_assistant"
}
Stack:
Error: Invalid tool result IDs: [call_01_w7wlz4ch9Hiw4Um74zwv8117]. Expected: [call_00_OugZmGqpB71ykFO5IjxM8342]
    at Object.onRunErrorEvent (http://localhost:5173/node_modules/.vite/deps/directive-CFNeHBg4.js?v=dab07746:20263:36)
    at http://localhost:5173/node_modules/.vite/deps/directive-CFNeHBg4.js?v=dab07746:9454:88
    at N (http://localhost:5173/node_modules/.vite/deps/directive-CFNeHBg4.js?v=dab07746:9032:17)
    at async http://localhost:5173/node_modules/.vite/deps/directive-CFNeHBg4.js?v=dab07746:9454:39

agentscope 2.0.3-snapsho
**Additional context**

guslegend0510:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #3043

Stateful AG-UI clients can replay tool results across multiple local assistant turns. Previously, transcript extraction could discard a result the server still needed, while forwarding already-consumed results that caused subsequent requests to fail.

This change:
- Selects replayed input after loading authoritative user/session state, preserving pending tool results anywhere in the transcript.
- Removes previously consumed results before recovery or context insertion.
- Preserves duplicate-ID, unknown-ID, partial-results-with-text, and permission confirmation checks.
- Keeps pending-tool recovery controlled by enablePendingToolRecovery.
- Adds 22 regression cases and updates English/Chinese documentation.

Orphan tool results are now rejected even when no calls are pending. Initial client-owned history can still include results paired with preceding assistant tool calls.

Validation:
- Reproduced failures on unmodified main; corresponding cases pass with this fix.
- Full Core, Harness, AG-UI, and AG-UI Spring Starter suites: 3,778 tests, 0 failures, 0 errors, 14 skipped.

## Checklist

- [x] Relevant module tests pass
- [x] Javadoc comments follow project conventions


agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -756,6 +756,12 @@ protected Object beforeAgentExecution(List<Msg> msgs, RuntimeContext rc) {
         return scope;
     }
 
+    @Override
+    protected List<Msg> prepareInputMessages(List<Msg> msgs, Object callScope) {
+        CallExecution scope = (CallExecution) callScope;
```
oss-maintainer, line 761: `prepareInputMessages` does an unchecked cast `(CallExecution) callScope` and then calls `selectReplayedMessages`. Two things worth hardening: (1) if `callScope` is ever null or a different type (a subclass of `ReActAgent` overriding `beforeAgentExecution`, or a future hook that passes a different scope) this becomes a raw `ClassCastException`/NPE inside the call path instead of a clear error; (2) the base-class contract for `prepareInputMessages` is not documented here. An `instanceof` check that returns `msgs` unchanged on mismatch (or making the hook signature generic/typed) would keep the replay optimisation strictly best-effort and non-breaking for subclasses.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2050,6 +2060,133 @@ private void applyToolUseBlockReplacements(Map<String, ToolUseBlock> replacement
...
+                return msgs;
+            }
+            int lastAssistant = -1;
+            for (int i = msgs.size() - 1; i >= 0; i--) {
+                if (msgs.get(i).getRole() == MsgRole.ASSISTANT) {
+                    lastAssistant = i;
+                    break;
+                }
+            }
+            if (lastAssistant < 0) {
```
oss-maintainer, line 2077: `selected.addAll(msgs.subList(lastAssistant + 1, msgs.size()))` keeps everything after the last assistant message, and in the `selected.isEmpty()` fallback the single last USER message is re-appended. Please confirm the ordering guarantees for the case where a tool result for a pending call appears *after* the last assistant message in the client transcript — with the fallback path, a mid-transcript pending result would be dropped and only the old user prompt replayed, which is exactly the failure mode the PR sets out to fix. If that transcript shape is possible with the AG-UI client, a scan over the whole transcript (not only indices `<= lastAssistant`) would be more robust; if it is guaranteed impossible, it would help to state that as an assumption in the javadoc of `RuntimeContext.REPLAYED_INPUT`.

agentscope-core/src/main/java/io/agentscope/core/agent/RuntimeContext.java
```diff
@@ -32,6 +32,15 @@
...
 
+    /**
+     * Boolean attribute indicating that input contains a replayed client transcript while the
+     * server owns conversation history. ReActAgent selects follow-up messages after the last
+     * client assistant turn, plus results for currently pending calls anywhere in the transcript.
+     * Selection happens after the authoritative session state is loaded. Omit this attribute for
+     * incremental input or stateless, client-owned history.
+     */
+    public static final String REPLAYED_INPUT = "agentscope_replayed_input";
+
```
oss-maintainer, line 43: Nit on the public API surface: this introduces a new cross-module contract (`agentscope-extensions-agui` writes the flag, `agentscope-core` reads it) as a raw string-keyed boolean attribute. Since it is a first-class behavioural switch rather than user data, consider a typed accessor on `RuntimeContext` (e.g. `isReplayedInput()` / `builder().replayedInput(true)`) so the flag cannot be misspelled or set to a non-Boolean value, and so it shows up in the typed-attribute API rather than in the free-form `get(String)` bag.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/processor/AguiRequestProcessor.java
```diff
@@ -163,6 +163,9 @@ public ProcessResult process(AguiRuntimeContextRequest<?> request) {
                             }
 
                             try {
+                                RuntimeContext effectiveRuntimeContext =
```
oss-maintainer, line 166: Moving `addResumeInterrupts(input, runtimeContext)` above the memory/`extractLatestUserMessage` branch changes which input the resume interrupts are computed from (full transcript instead of the extracted last user message, for the non-ReAct-with-memory path it is now built before truncation). It looks intentional and the tests cover the ReAct case; could you add (or point to) an assertion for the non-ReAct + server-memory path so the ordering change stays pinned? That is the only branch where the pre-existing behaviour could silently drift.

oss-maintainer (commented):
### Summary
Reconciles replayed AG-UI tool results with server-side session state: the AG-UI processor now keeps the full client transcript for ReAct-backed agents (`RuntimeContext.REPLAYED_INPUT`), and `ReActAgent` selects the follow-up messages *after* authoritative state is loaded, while consumed results are stripped before recovery/context insertion. Good, well-scoped fix for a genuinely subtle protocol bug, and 22 regression cases plus EN/ZH docs is a strong evidence base. I have two robustness questions about transcript shapes that the current selection window may not cover, plus API-surface nits. No blocking correctness issue found in the paths I could trace.

**CLA:** no `license/cla` status context was found on this PR's head commit (`unknown`), so this review is posted as a comment rather than an approval — approvals are reserved for PRs where CLA state is verifiable.

Findings:
- **[Warning]** `ReActAgent.java:2077` — selection only rescans indices `<= lastAssistant`; a pending result appearing after the last assistant message falls into the "replay the old user prompt" fallback.
- **[Warning]** `ReActAgent.java:761` — unchecked `(CallExecution) callScope` cast in the new `prepareInputMessages` override.
- **[Info]** `RuntimeContext.java:43` — new cross-module flag modelled as a raw string attribute.
- **[Info]** `AguiRequestProcessor.java:166` — `addResumeInterrupts` hoisting changes input for the non-ReAct + server-memory branch.

---

oss-maintainer (commented):
### Summary
Re-review of PR #3046 at its current head (automated review of the changes in this diff). 4 item(s) raised below; nothing here blocks the approach itself — details inline.

### Findings
- **[Warning]** `agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java:2120` — selectReplayedMessages replaces the content of messages containing pending results with only the matching ToolResultBlocks, dropping any TextBlock or other content that was in the same message. Keep the full message (or at least non-result blocks) so explanatory text is not lost when a replayed tool message carries both a pending result and text.
- **[Warning]** `agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java:761` — prepareInputMessages casts callScope to CallExecution without a null check. If beforeAgentExecution ever returns null (e.g., an early-exit subclass), this will NPE before the existing null-safe path in runLifecycleBody can handle it. Add a null guard and return msgs unchanged.
- **[Info]** `agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java:311` — runLifecycleBody now passes preparedMsgs to TracerRegistry.callAgent, notifyPreCall, and the error handler instead of the original caller input. Verify that tracers/pre-call hooks are expected to observe the filtered/replayed messages rather than the raw input.
- **[Info]** `agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java:2126` — selectReplayedMessages returns an empty list for a replay that contains only an assistant message with no preceding user message. Consider whether this should fail fast or fall back to the latest user message even when it appears after the assistant.

### Notes

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2076,6 +2086,133 @@ private void applyToolUseBlockReplacements(Map<String, ToolUseBlock> replacement
...
+            for (int i = 0; i <= lastAssistant; i++) {
+                Msg msg = msgs.get(i);
+                List<ContentBlock> results =
+                        msg.getContent().stream()
+                                .filter(
+                                        block ->
+                                                block instanceof ToolResultBlock result
+                                                        && pendingIds.contains(result.getId()))
+                                .toList();
+                if (!results.isEmpty()) {
```
oss-maintainer: selectReplayedMessages replaces the content of messages containing pending results with only the matching ToolResultBlocks, dropping any TextBlock or other content that was in the same message. Keep the full message (or at least non-result blocks) so explanatory text is not lost when a replayed tool message carries both a pending result and text.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
oss-maintainer, line 762: prepareInputMessages casts callScope to CallExecution without a null check. If beforeAgentExecution ever returns null (e.g., an early-exit subclass), this will NPE before the existing null-safe path in runLifecycleBody can handle it. Add a null guard and return msgs unchanged.

agentscope-core/src/main/java/io/agentscope/core/agent/AgentBase.java
```diff
@@ -307,18 +307,19 @@ private Mono<Msg> runLifecycleBody(
         // shutdown interrupts / saves the exact (userId, sessionId) session rather than the agent's
         // no-arg "most-recently-active" accessors.
         GracefulShutdownManager.getInstance().bindRequestState(requestId, stateForCall(scope));
+        List<Msg> preparedMsgs = prepareInputMessages(msgs, scope);
         Mono<Msg> body =
```
oss-maintainer, line 311: runLifecycleBody now passes preparedMsgs to TracerRegistry.callAgent, notifyPreCall, and the error handler instead of the original caller input. Verify that tracers/pre-call hooks are expected to observe the filtered/replayed messages rather than the raw input.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2076,6 +2086,133 @@ private void applyToolUseBlockReplacements(Map<String, ToolUseBlock> replacement
...
+                    selected.add(msg.withContent(results));
+                }
+            }
+            selected.addAll(msgs.subList(lastAssistant + 1, msgs.size()));
+            // Preserve regenerate/continue behavior, but never replay the old user prompt in
+            // place of a pending tool result found in the middle of the transcript.
```
oss-maintainer, line 2132: selectReplayedMessages returns an empty list for a replay that contains only an assistant message with no preceding user message. Consider whether this should fail fast or fall back to the latest user message even when it appears after the assistant.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2076,6 +2086,139 @@ private void applyToolUseBlockReplacements(Map<String, ToolUseBlock> replacement
...
+            for (int i = 0; i <= lastAssistant; i++) {
+                Msg msg = msgs.get(i);
+                List<ContentBlock> replayContent =
+                        msg.getContent().stream()
+                                .filter(
+                                        block ->
+                                                !(block instanceof ToolResultBlock result)
+                                                        || pendingIds.contains(result.getId()))
+                                .toList();
+                boolean hasPendingResult =
```
oss-maintainer: The pending-result check re-scans `replayContent` with the same predicate that just built it, so every replayed message is walked twice (and the second pass keeps `ToolResultBlock` instances that the filter already matched). Since the filter can only keep a result when it is pending, `hasPendingResult` is equivalent to "the original content had at least one pending result" — computing it during the first pass (e.g. a small loop that collects the kept blocks and sets a flag) removes the extra stream and makes the invariant explicit. Minor, no behavioural impact.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentToolResultReplayTest.java
```diff
@@ -0,0 +1,518 @@
...
+                    model.inputs.get(0).stream()
+                            .anyMatch(
+                                    m ->
+                                            "old prompt".equals(m.getTextContent())
+                                                    || "client system history"
+                                                            .equals(m.getTextContent())));
+        }
+    }
+
+    @Test
```
oss-maintainer, line 369: Good coverage for the new behaviour (pending result + text is preserved). One case worth adding while this code is fresh: a single message carrying **both** a pending result and a stale (already-consumed) result plus text — that is the path where `!(block instanceof ToolResultBlock) || pendingIds.contains(...)` actually filters something out, and today only the pure pending+text mix is asserted. It would pin the "stale results are still dropped" half of the contract, which is the part a future refactor of this filter could silently regress.

oss-maintainer (commented):
### Summary
Re-review after the new commit `d8652fdb` ("fix(core): preserve replay message context"). The previous **[Warning]** on `selectReplayedMessages` dropping non-result content is **fixed correctly**: the filter now keeps every block that is not a `ToolResultBlock`, plus the results whose ids are still pending, and the message is only replayed when it actually carries a pending result. The new `replaySelectionPreservesTextAlongsidePendingResult` test pins the exact regression, and CI is green on this head (build ubuntu/windows, validate, Check License, Check Module Sync, `license/cla` all pass).

Two open items from the earlier reviews are unchanged by this commit and still apply to the current head — not repeated in detail here, just flagged so they do not get lost:
- `ReActAgent.java:761` — `prepareInputMessages` still does an unchecked `(CallExecution) callScope` cast with no null guard.
- `AgentBase.java:311` — `runLifecycleBody` still feeds the *filtered* `preparedMsgs` to tracers / `notifyPreCall`, so hooks observe replayed input rather than the raw caller input; worth an explicit confirmation that this is intended.

The two points below are non-blocking nits on the new code itself; nothing here changes the verdict on the approach.

### Verdict
`COMMENT` — the replay-context fix looks right and is covered by a test. Remaining feedback is two carry-over items plus two minor nits; none of them are blockers from my side, so this is ready for a maintainer look.

---

oss-maintainer (commented):
### Summary
Follow-up review of the new commit `d26d43f5` ("tighten replay result filtering").

The change is a semantically equivalent rewrite of the previous stream-based filtering in `selectReplayedMessages` (single pass, same behavior: non-result blocks kept, ToolResultBlocks kept only when their id is still pending) and is strictly clearer. The added fixture line covers exactly the mixed `old + pending + text` case requested in the earlier review thread — good.

No new findings on the delta itself. The inline comments from the previous review round (unchecked `CallExecution` cast / null handling in `prepareInputMessages`, suffix-selection semantics in `selectReplayedMessages`, cross-module flag contract in `RuntimeContext`, and the `AgentBase` prepared-input propagation question) remain open — keeping those threads as the authoritative feedback to avoid duplication. CLA signed.

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -757,6 +757,12 @@ protected Object beforeAgentExecution(List<Msg> msgs, RuntimeContext rc) {
         return scope;
     }
 
+    @Override
+    protected List<Msg> prepareInputMessages(List<Msg> msgs, Object callScope) {
+        CallExecution scope = (CallExecution) callScope;
```
oss-maintainer, line 762: The `callScope` parameter arrives as `Object` from `AgentBase.runLifecycleBody`, and the previous review round already asked for this cast to be made safe. It is still unguarded: if `beforeAgentExecution` ever returns `null` (or a non-`CallExecution` scope, e.g. for a subclass/mixin that overrides `beforeAgentExecution`), this throws `ClassCastException`/`NullPointerException` from inside `prepareInputMessages` — before tracing and pre-call hooks — so the failure surfaces as an opaque run error rather than "replay handling not applicable". A cheap guard keeps the hook total:

```java
if (!(callScope instanceof CallExecution scope)) {
    return msgs;
}
return scope.selectReplayedMessages(msgs);
```

(plus a `msgs == null` short-circuit if the base class can pass `null`).

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2077,6 +2087,138 @@ private void applyToolUseBlockReplacements(Map<String, ToolUseBlock> replacement
...
+                    }
+                }
+                if (hasPendingResult) {
+                    selected.add(msg.withContent(replayContent));
+                }
+            }
+            selected.addAll(msgs.subList(lastAssistant + 1, msgs.size()));
+            // Preserve regenerate/continue behavior, but never replay the old user prompt in
+            // place of a pending tool result found in the middle of the transcript.
+            if (selected.isEmpty()) {
```
oss-maintainer, line 2133: This fallback can re-append an already-answered user prompt. `selected` is empty when (a) no message up to `lastAssistant` carries a result for a currently pending call, and (b) `msgs` has nothing after `lastAssistant` — i.e. a replayed transcript that ends on the client's assistant turn with no pending tools. In that case the `USER` message at `lastAssistant - 1` is added back and `removeConsumedToolResults` passes it through untouched (it only filters `ToolResultBlock`s), so the model is asked to answer the original instruction a second time. The comment above ("never replay the old user prompt") reads as if this path were excluded, and the no-pending-results case is exactly when the guarantee is not held.

For a "regenerate" request this is intended; for a stale/duplicate replay of a completed round it is a user-visible double execution (extra tool calls, extra model spend). Could you either confirm the intent and add a case for "REPLAYED_INPUT + transcript ends on assistant + `pendingIds` empty", or return the input unfiltered / reject when there is nothing to reconcile?

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1784,20 +1790,24 @@ private Mono<Msg> doCallInner(List<Msg> msgs) {
...
+            List<ToolUseBlock> asking = pendingIds.isEmpty() ? List.of() : askingToolCalls();
             if (!asking.isEmpty()) {
                 validateAndAcceptConfirmResults(msgs, asking);
                 return resumeAgent();
             }
 
+            // Transcript-replaying clients may resend results already consumed by this session.
+            // Normalize against the call-scoped state before either recovery or normal input
+            // handling, including calls with no pending tools. Unknown IDs remain invalid.
+            msgs = removeConsumedToolResults(msgs, pendingIds);
```
oss-maintainer, line 1804: `removeConsumedToolResults` is applied to *every* call, not just replays (only `selectReplayedMessages` is gated on `RuntimeContext.REPLAYED_INPUT`). For sessions with `pendingIds` empty, a client that echoes the last tool result together with its follow-up message — legal under the old `addToContext` path — now hits the `Invalid tool result IDs` throw at line 2185. The docs update covers it, and `unknownTrailingResultFailsWithoutPoisoningTheNextCorrectedRequest` pins the behaviour, but this is a behavior change for existing non-replay callers: would it be worth an explicit `@apiNote` on the `ReActAgent` class / a CHANGELOG entry mentioning the stricter contract (and that `REPLAYED_INPUT` is the opt-out path)?

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2077,6 +2087,138 @@ private void applyToolUseBlockReplacements(Map<String, ToolUseBlock> replacement
...
+                        throw new IllegalStateException("Duplicate tool result ID: " + id);
+                    }
+                    if (!pendingIds.contains(id)
+                            && !consumedIds.contains(id)
+                            && !inputToolIds.contains(id)) {
+                        invalidIds.add(id);
+                    }
+                }
+            }
+            if (!invalidIds.isEmpty()) {
```
oss-maintainer, line 2185: Minor typing note for the three new rejection paths here (`Duplicate tool result ID`, `Invalid tool result IDs`, `Input contains only previously consumed tool results`): these are all caller-input problems, but they are raised as `IllegalStateException`, which is also what `ObjectStoreTranscriptStore.appendSegment` now uses for an I/O failure. Through the AG-UI path both surface as `RUN_ERROR` (`unknownTrailingResultFailsWithoutPoisoningTheNextCorrectedRequest` pins that), so nothing is broken today — but a dedicated validation exception type (or an `IllegalArgumentException` for input-shaped failures) would let transports map 4xx-vs-5xx and would keep operator-facing "write failed" errors distinguishable from "client sent a bad transcript" in logs.

oss-maintainer (commented):
### Summary
Follow-up review of the new head `347c3047`. `prepareInputMessages` is the right seam — it runs after session activation and under the serialization gate, so replay reconciliation sees authoritative state instead of a cached one, and the default hook is a no-op for every other agent. `removeConsumedToolResults` is thorough: consumed-vs-pending-vs-initial-transcript are separated correctly, duplicate submission is caught before filtering, and the mixed message is rebuilt without dropping confirm metadata. Test coverage is unusually good (519 + 291 lines, including the no-poisoning-after-rejection case), and both docs pages were updated in lockstep. CLA signed.

Four points inline: the unchecked `CallExecution` cast from the previous round is still open; the `selected.isEmpty()` fallback can re-append an already-answered user prompt when a replayed transcript ends on the assistant turn with nothing pending; normalization now applies to non-replay calls as well, which is a stricter contract for existing clients and deserves an explicit note; and the three new caller-input rejections share `IllegalStateException` with genuine write failures. None of these look like blockers for the approach itself — the first two are worth resolving before merge.

---

oss-maintainer:

This PR currently conflicts with `main` and cannot be merged (head `codex/fix-3043-agui-tool-result-replay`). Please rebase (or merge `main` in) and resolve the conflicts:

```bash
git fetch upstream
git checkout codex/fix-3043-agui-tool-result-replay
git rebase upstream/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #3051, opened 2026-09-08
refactor(gemini): pass tool parameter schemas directly as native JSON Schema

### LINKED ISSUE #3050: [Bug]: Gemini tool conversion drops function declarations for nullable JSON Schema types
**Describe the bug**

When a Gemini tool parameter uses a nullable JSON Schema such as
`{"type": ["string", "null"]}`, the tool declaration can be dropped during
schema conversion.

If the same tool is forced with `ToolChoice.Specific`, Gemini rejects the
request because the forced tool name is not present in the submitted function
declarations.

**To Reproduce**

1. Define a tool parameter with the following schema:

   ```json
   {
     "type": "object",
     "properties": {
       "value": {
         "type": ["string", "null"]
       }
     }
   }
   ```

2. Use `ToolChoice.Specific("nullable_echo")`.
3. Send the request through `GeminiChatModel`.
4. Observe the Gemini API response.

**Expected behavior**

The nullable parameter should be converted successfully, and the forced tool
should remain present in the Gemini function declarations.

**Error messages**

```text
`allowed_function_names` should be a subset of the provided
`function_declarations` names. Found invalid function names: nullable_echo
```

**Environment**

- Java Version: 21
- OS: macOS

**Additional context**

The baseline implementation assumed that JSON Schema `type` was always a
string. A nullable schema uses an array, which caused conversion to fail and
the complete function declaration to be skipped.

## Actual reproduction evidence

This was reproduced with the real Gemini API path using
`GeminiNullableToolSchemaE2ETest`.

The test sends a tool named `nullable_echo` with the following parameter
schema:

```json
{
  "type": "object",
  "properties": {
    "value": {
      "type": ["string", "null"]
    }
  }
}
```

The request also uses:

```java
ToolChoice.Specific("nullable_echo")
```

Before the fix, the test produced:

```text
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0
Gemini API call failed: 400
allowed_function_names ... invalid function names: nullable_echo
```

The failure occurs in the following sequence:

1. `GeminiToolsHelper.convertParametersToSchema` assumes that `type` is a
   string.
2. The nullable schema provides `type` as a list.
3. The conversion raises a `ClassCastException`.
4. The surrounding tool conversion catches the exception and skips the
   complete function declaration.
5. The forced tool configuration still contains `nullable_echo`.
6. Gemini receives a forced tool name without a matching function declaration
   and returns HTTP 400.

This is not a mocked or local transport failure. It is a real API validation
error caused by the mismatch between the forced tool configuration and the
submitted function declarations.

The fix is implemented in PR #3051.

liugy789:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Fixes #3050

### Background

Gemini tool conversion previously assumed that JSON Schema `type` was always a
string. When a nullable parameter used `type: ["string", "null"]`, conversion
failed and the complete function declaration was dropped.

When `ToolChoice.Specific` still referenced that tool, Gemini returned HTTP 400
because the forced tool name was not included in the submitted function
declarations.

## Checklist

- [x] Code has been formatted with Spotless.
- [x] Relevant unit and module tests are passing.
- [x] Javadoc comments follow project conventions.
- [x] Related documentation was reviewed; no user-facing documentation update
      is required for this bug fix.

### DIFF
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-e2e-tests/src/test/java/io/agentscope/extensions/model/gemini/GeminiNullableToolSchemaE2ETest.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-e2e-tests/src/test/java/io/agentscope/extensions/model/gemini/GeminiNullableToolSchemaE2ETest.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.extensions.model.gemini;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+import io.agentscope.core.agent.test.TestUtils;
+import io.agentscope.core.e2e.E2ETestCondition;
+import io.agentscope.core.message.Msg;
+import io.agentscope.core.message.ToolUseBlock;
+import io.agentscope.core.model.ChatResponse;
+import io.agentscope.core.model.GenerateOptions;
+import io.agentscope.core.model.ToolChoice;
+import io.agentscope.core.model.ToolSchema;
+import io.agentscope.extensions.model.gemini.formatter.GeminiChatFormatter;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/** Provider-specific acceptance test for nullable Gemini tool schemas. */
+@Tag("e2e")
+@Tag("tools")
+@ExtendWith(E2ETestCondition.class)
+class GeminiNullableToolSchemaE2ETest {
+
+    private static final String GOOGLE_API_KEY = "GOOGLE_API_KEY";
+    private static final String GOOGLE_GEMINI_MODEL = "GOOGLE_GEMINI_MODEL";
+    private static final String DEFAULT_MODEL = "gemini-3.6-flash";
+    private static final String TOOL_NAME = "nullable_echo";
+    private static final Duration TEST_TIMEOUT = Duration.ofSeconds(90);
+
+    @Test
+    @DisplayName("Gemini accepts a nullable JSON Schema tool parameter")
+    void acceptsNullableToolSchema() {
+        String apiKey = System.getenv(GOOGLE_API_KEY);
+        assumeTrue(apiKey != null && !apiKey.isBlank(), "GOOGLE_API_KEY is required");
+        String modelName = System.getenv(GOOGLE_GEMINI_MODEL);
+        if (modelName == null || modelName.isBlank()) {
+            modelName = DEFAULT_MODEL;
+        }
+
+        GeminiChatModel model =
+                GeminiChatModel.builder()
+                        .apiKey(apiKey)
+                        .modelName(modelName)
+                        .formatter(new GeminiChatFormatter())
+                        .build();
+
+        Map<String, Object> parameters =
+                Map.of(
+                        "type",
+                        "object",
+                        "properties",
+                        Map.of("value", Map.of("type", List.of("string", "null"))));
+        ToolSchema toolSchema =
+                ToolSchema.builder()
+                        .name(TOOL_NAME)
+                        .description("Echo an optional value")
+                        .parameters(parameters)
+                        .build();
+
+        Msg request = TestUtils.createUserMessage("User", "Call nullable_echo with value hello.");
+        GenerateOptions options =
+                GenerateOptions.builder().toolChoice(new ToolChoice.Specific(TOOL_NAME)).build();
+
+        List<ChatResponse> responses =
+                model.stream(List.of(request), List.of(toolSchema), options)
+                        .collectList()
+                        .block(TEST_TIMEOUT);
+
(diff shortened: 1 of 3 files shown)


agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiToolsHelper.java
```diff
@@ -170,6 +187,128 @@ protected Schema convertParametersToSchema(Map<String, Object> parameters) {
...
+                }
+                nonNullSchemas.add(typedSchema);
+            }
+        }
+
+        if (nonNullSchemas.size() == anyOfSchemas.size()) {
+            return schema;
+        }
+
+        Map<String, Object> normalized = new HashMap<>(schema);
```
dailingtao, line 224: Removing the `null` branch here changes the schema semantics without marking the resulting Gemini schema as nullable. For example, `anyOf: [{type: "string"}, {type: "null"}]` becomes a non-nullable `STRING` schema, so valid `null` arguments are no longer accepted. Please preserve this by setting `nullable(true)` whenever a null branch is removed, and add assertions for both the single- and multi-branch nullable `anyOf` cases.

liugy789: Thank you for the careful review and for pointing out this nullable semantics issue.

Fixed in commit 83120227.

The nullable schema normalization now preserves nullable=true whenever null
branches are removed from anyOf. Added regression assertions for single-branch,
multi-branch, and all-null nullable anyOf cases, plus a guard against
nullable=false overriding type-array nullability.

Validation:
- GeminiToolsHelperTest: 20 tests passed.
- Gemini module tests: 148 tests passed.
- Real Gemini E2E tests: 2 passed, including nullable type-array and
  single-non-null-branch anyOf schemas with a JSON null argument.
- Spotless and git diff checks passed.

CryoThrust:
The nullable-schema normalization looks good, and the unit coverage is broad. One behavior worth making explicit in a follow-up test is the distinction between a nullable type array and an existing `anyOf`: the implementation intentionally rejects a schema that combines multiple non-null types in `type: [...]\` with `anyOf`. That is reasonable, but the exception contract should remain stable because callers may use this helper for provider-generated schemas.

I also confirmed the PR's current CI is green across Ubuntu/Windows, license, module sync, and CLA.

liugy789:
@CryoThrust Thank you for the additional review and for highlighting this boundary.

Added a focused regression assertion in commit a37e8e5b. The existing
testRejectsMultiTypeArrayCombinedWithAnyOf test now verifies that combining
multiple non-null types in type: [...] with anyOf consistently throws
IllegalArgumentException with the documented error message.

This keeps the accepted nullable type-array and existing anyOf paths distinct
from the unsupported conflicting combination. GeminiToolsHelperTest (20 tests)
and Spotless both pass.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiToolsHelper.java
```diff
@@ -170,6 +191,129 @@ protected Schema convertParametersToSchema(Map<String, Object> parameters) {
...
+            if ("null".equalsIgnoreCase(typeString)) {
+                nullable = true;
+            } else {
+                nonNullTypes.add(typeString);
+            }
+        }
+
+        if (nonNullTypes.isEmpty()) {
+            // Preserve the compatibility fallback for a null-only type array.
+            schemaBuilder.type(new Type(Type.Known.OBJECT));
```
dailingtao, line 297: `type: ["null"]` currently takes this branch and returns an `OBJECT` schema before `nullable(true)` is applied, so a JSON Schema that accepts only null is converted into one that does not accept null at all. Could we either mark the fallback `OBJECT` as nullable before returning, or reject this unsupported shape explicitly, and extend `testConvertsNullOnlyTypeArrayToObject` to assert the chosen contract?

liugy789: Thank you for catching this edge case. You are right that the null-only
type-array fallback returned an OBJECT schema before nullable could be applied.

Fixed in commit 847f3a0c by preserving nullable=true on that fallback, and
extended testConvertsNullOnlyTypeArrayToObject to assert the nullable contract.
The new regression test was RED before the fix and GREEN after it.

GeminiToolsHelperTest (20 tests), the full Gemini module tests (148 tests),
Spotless, and git diff checks all pass.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-e2e-tests/src/test/java/io/agentscope/extensions/model/gemini/GeminiNullableToolSchemaE2ETest.java
```diff
@@ -0,0 +1,162 @@
...
+    private static final String DEFAULT_MODEL = "gemini-3.6-flash";
+    private static final String TOOL_NAME = "nullable_echo";
+    private static final Duration TEST_TIMEOUT = Duration.ofSeconds(90);
+
+    @Test
+    @DisplayName("Gemini accepts a nullable JSON Schema tool parameter")
+    void acceptsNullableToolSchema() {
+        String apiKey = System.getenv(GOOGLE_API_KEY);
+        assumeTrue(apiKey != null && !apiKey.isBlank(), "GOOGLE_API_KEY is required");
+        String modelName = System.getenv(GOOGLE_GEMINI_MODEL);
```
oss-maintainer, line 57: `DEFAULT_MODEL = "gemini-3.6-flash"` — please double-check that identifier. If it is not a real deployable model name the opt-in E2E test fails on every run with a provider 404 instead of covering the bug, which is worse than skipping. If it is intentional, prefer resolving it from `GOOGLE_GEMINI_MODEL` only and skipping when unset, rather than keeping a hard-coded default in the source.

oss-maintainer (commented):
### Summary
Good, well-scoped fix: JSON Schema `type` as an array (nullable params) no longer drops the whole function declaration and then 400s on `ToolChoice.Specific`. The `anyOf` handling plus explicit rejection of conflicting multi-type/`anyOf` combinations at the conversion boundary is the right place to fail, and 20 unit tests plus opt-in E2E coverage is solid. One thing to confirm before merge: the hard-coded default model id in the new E2E test.

---

jujn (approved):

PR #3056, opened 2026-09-09
feat(harness): support multi-keyword search modes

### LINKED ISSUE #3054: [Feature] 建议为 memory_search / session_search 增加可选的多关键词匹配模式
模型调用 `memory_search` 和 `session_search` 时，有时会将多个关键词组合成一个 query。当前实现将整个 query 作为字面短语匹配，相关记录即使存在，也可能返回无匹配结果。

### 示例

记忆文件或历史消息包含：

```text
部署决定：使用蓝鲸方案，服务端口为 9580。
```

查询结果：

```text
query = "蓝鲸"       → 命中
query = "部署 蓝鲸"  → 无匹配
```

当前实现中：

- `memory_search` 使用 `Pattern.quote(query)` 进行字面匹配。
- `session_search` 使用完整 query 进行大小写不敏感的子串匹配。
- `memory_search` 的参数描述使用了 “Keywords”，容易让模型传入多个空格分隔的关键词。

对应源码：[MemorySearchTool](https://github.com/agentscope-ai/agentscope-java/blob/c549a5e0/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/MemorySearchTool.java)、[SessionSearchTool](https://github.com/agentscope-ai/agentscope-java/blob/c549a5e0/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SessionSearchTool.java)。

### 建议

保留现有短语匹配行为，增加可选参数 `matchMode`：

| 模式 | 含义 |
|---|---|
| `phrase` | 完整短语匹配，作为默认值 |
| `all` | 同一条记录包含所有关键词，不要求相邻或顺序一致 |
| `any` | 同一条记录包含任意关键词 |

第一版可以只按空白字符拆分关键词，不引入额外分词依赖。例如，模型可以传入 `"部署 蓝鲸"`，暂不处理连续中文句子的自动分词。

匹配范围保持现状：Memory 按行匹配，Session 按单条 entry 匹配，不跨记录组合关键词。同时在工具描述中说明不同模式的使用方式。

### 兼容性

- 不传新参数时保持当前行为。
- 保留现有 Java 调用方式。
- 关键词按字面值处理，不解释为正则表达式。
- 不涉及日期解析、语义检索或历史截断。

这个方向是否合适？如果有更适合的参数设计，也可以讨论调整。会附上一个实现 PR，便于结合代码评审。

CryoThrust:
Closes #3054

## What does this PR do?

Adds an optional `matchMode` parameter to `memory_search` and `session_search`.

- `phrase` (default) preserves the existing literal full-query behavior.
- `all` matches records containing every whitespace-separated keyword.
- `any` matches records containing at least one keyword.

Matching remains case-insensitive and literal. Memory searches remain line-scoped, session searches remain entry-scoped, and keywords are never combined across records.

## Compatibility

- Existing Java overloads remain available.
- Missing, blank, or unknown modes fall back to `phrase`.
- No tokenizer, semantic search dependency, or storage format change.

## Verification

`mvn -s /Users/yohanes/.m2/settings-public-central.xml -pl agentscope-harness -am -Dtest=SearchMatchModeTest -Dsurefire.failIfNoSpecifiedTests=false -Dcheckstyle.skip=true -Dspotless.check.skip=true test`

Result: 4 tests passed.

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/MemorySearchTool.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/MemorySearchTool.java
@@ -21,7 +21,6 @@ import io.agentscope.core.tool.ToolParam;
 import io.agentscope.harness.agent.workspace.WorkspaceManager;
 import java.util.List;
 import java.util.StringJoiner;
-import java.util.regex.Pattern;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -52,22 +51,30 @@ public class MemorySearchTool {
     public String memorySearch(
             RuntimeContext runtimeContext,
             @ToolParam(name = "query", description = "Keywords to search for in memory files")
-                    String query) {
+                    String query,
+            @ToolParam(
+                            name = "matchMode",
+                            description = "Matching mode: phrase (default), all, or any",
+                            required = false)
+                    String matchMode) {
         if (query == null || query.isBlank()) {
             return "No query provided";
         }
 
         RuntimeContext rc = runtimeContext != null ? runtimeContext : RuntimeContext.empty();
-        return keywordSearch(rc, query);
+        return keywordSearch(rc, query, SearchMatchMode.parse(matchMode));
     }
 
-    private String keywordSearch(RuntimeContext rc, String query) {
+    /** Preserves the pre-matchMode Java entry point for direct callers. */
+    public String memorySearch(RuntimeContext runtimeContext, String query) {
+        return memorySearch(runtimeContext, query, null);
+    }
+
+    private String keywordSearch(RuntimeContext rc, String query, SearchMatchMode matchMode) {
         StringJoiner results = new StringJoiner("\n");
         int matchCount = 0;
 
         List<String> memoryPaths = workspaceManager.listMemoryFilePaths(rc);
-        Pattern pattern = Pattern.compile(Pattern.quote(query), Pattern.CASE_INSENSITIVE);
-
         for (String relativePath : memoryPaths) {
             String content = workspaceManager.readManagedWorkspaceFileUtf8(rc, relativePath);
             if (content == null || content.isEmpty()) {
@@ -75,7 +82,7 @@ public class MemorySearchTool {
             }
             String[] lines = content.split("\n", -1);
             for (int i = 0; i < lines.length; i++) {
-                if (pattern.matcher(lines[i]).find()) {
+                if (matchMode.matches(lines[i], query)) {
                     results.add(String.format("Source: %s#%d: %s", relativePath, i + 1, lines[i]));
                     matchCount++;
                 }
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SearchMatchMode.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SearchMatchMode.java
@@ -6,6 +6,12 @@
  * You may obtain a copy of the License at
  *
  *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
 package io.agentscope.harness.agent.tool;
 
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SearchMatchMode.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SearchMatchMode.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.agentscope.harness.agent.tool;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.regex.Pattern;
+
+/** Matching modes shared by the memory and session search tools. */
+enum SearchMatchMode {
+    PHRASE,
+    ALL,
+    ANY;
+
+    static SearchMatchMode parse(String value) {
+        if (value == null || value.isBlank()) {
+            return PHRASE;
+        }
+        try {
+            return valueOf(value.trim().toUpperCase(Locale.ROOT));
+        } catch (IllegalArgumentException ignored) {
+            return PHRASE;
+        }
+    }
+
+    boolean matches(String text, String query) {
+        if (text == null || query == null) {
+            return false;
+        }
+        if (this == PHRASE) {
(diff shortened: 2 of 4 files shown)


CryoThrust:
All CI checks are now green, including Ubuntu and Windows builds, license, module sync, Codecov, and CLA. The PR preserves the existing phrase behavior and Java overloads, with focused coverage for phrase/all/any matching. Ready for maintainer review.

CryoThrust:
A parallel PR #3062 now proposes the same multi-keyword search modes for #3054. It was opened after this PR and follows a similar compatibility boundary. I am flagging the overlap so maintainers can choose one implementation path; I am happy to align with the preferred design or close this PR if #3062 is selected. The current PR remains ready for review with all required CI checks passing.

oss-maintainer (commented):
### Summary

Adds an optional `matchMode` (phrase / all / any) to `memory_search` and `session_search`, backed by a shared `SearchMatchMode` enum plus unit tests. The design is clean and the back-compatible Java overloads for existing callers are a nice touch. Two things hold this back from approval: a hot-path regression in the matcher, and a feature overlap with #3062.

### Findings

- **[Warning]** `SearchMatchMode.java:45` — the phrase `Pattern` is now compiled once per scanned line instead of once per search; previously `MemorySearchTool.keywordSearch` hoisted `Pattern.compile(Pattern.quote(query), CASE_INSENSITIVE)` out of the loop. Memory files are scanned line by line, so one compile becomes thousands.
- **[Warning]** `SearchMatchMode.java:50` — the keyword split is recomputed per line for the same reason. `\\s+` is also ASCII-only here, so an ideographic space (`　`) or non-breaking space (` `) inside a CJK query will not split into keywords; `Pattern.compile("\\s+", Pattern.UNICODE_CHARACTER_CLASS)` would.
- **[Warning]** `SearchMatchMode.java:36` — an unrecognised `matchMode` silently falls back to `PHRASE`. Since the parameter is filled in by the model, a value like `or`/`keywords` yields phrase semantics with no signal, so the agent sees "no results" and retries instead of correcting the argument. The tools already return error strings (`"No query provided"`, `"Error: query is required"`), so an explicit error would be consistent.
- **[Info]** `MemorySearchTool.java:85` — call site where the per-record cost lands; hoisting a prepared matcher out of the loop keeps identical behaviour at the original cost.

### Tests

`SearchMatchModeTest` covers the mode matrix well at the unit level. Gaps: no end-to-end coverage through `memory_search` / `session_search` via `Toolkit` (so schema generation and reflective invocation of the new optional parameter are untested), and no non-ASCII query — either would have surfaced the two issues above.

### Cross-PR Note

This overlaps with **#3062** (same `matchMode=phrase|all|any` parameter on the same two tools, same whitespace-split semantics, touching the same files), so the two will conflict. #3062 has already been approved by @dailingtao, hoists a prepared `Predicate<String>` out of the loop, errors on an invalid mode, and ships `docs/v2/{en,zh}` updates for the new parameter — which this PR does not.

If #3062 is the one going in, consider closing this to save the review effort; if you would rather keep this design, it is worth porting the non-ASCII split and the explicit-mode-error into it. Either way the tool schema and docs should land only once.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SearchMatchMode.java
```diff
@@ -0,0 +1,63 @@
...
+            return PHRASE;
+        }
+    }
+
+    boolean matches(String text, String query) {
+        if (text == null || query == null) {
+            return false;
+        }
+        if (this == PHRASE) {
+            return Pattern.compile(Pattern.quote(query), Pattern.CASE_INSENSITIVE)
```
oss-maintainer, line 45: **[Warning]** Hot-path regression: the phrase pattern is compiled on **every line**. Before this change `MemorySearchTool.keywordSearch` hoisted `Pattern.compile(Pattern.quote(query), CASE_INSENSITIVE)` out of the loop once per search; now `matches()` recompiles it per record (see `MemorySearchTool.java:85`). Memory files are scanned line-by-line, so this turns one compile into thousands.

Suggest resolving the matcher once per search instead of per record, e.g. make the mode produce a `Predicate<String>`:

```java
static Predicate<String> matcher(SearchMatchMode mode, String query) { ... }
```

so the loop body becomes `if (matcher.test(lines[i]))`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SearchMatchMode.java
```diff
@@ -0,0 +1,63 @@
...
+                    .matcher(text)
+                    .find();
+        }
+        List<String> keywords =
+                Arrays.stream(query.trim().split("\\s+"))
```
oss-maintainer, line 50: **[Warning]** Same issue as above — `query.trim().split("\\s+")` and the `keywords` list are rebuilt for every line scanned. Combined with the per-line `Pattern.compile` at line 45, the per-record cost is now dominated by work that is constant for the whole search.

Also `\\s+` is ASCII-only here, so an ideographic space (`\u3000`) or non-breaking space (`\u00a0`) inside a Chinese query will not split into keywords. If that matters, `Pattern.compile("\\s+", Pattern.UNICODE_CHARACTER_CLASS)` covers it.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SearchMatchMode.java
```diff
@@ -0,0 +1,63 @@
...
+    ANY;
+
+    static SearchMatchMode parse(String value) {
+        if (value == null || value.isBlank()) {
+            return PHRASE;
+        }
+        try {
+            return valueOf(value.trim().toUpperCase(Locale.ROOT));
+        } catch (IllegalArgumentException ignored) {
+            return PHRASE;
```
oss-maintainer, line 36: **[Warning]** Silently downgrading an unrecognised `matchMode` to `PHRASE` hides caller mistakes. `matchMode` is filled in by the model, so a value like `"ANY"` (handled) vs `"or"` / `"keywords"` (not handled) will produce phrase semantics with no signal — the agent sees "no results" and cannot tell that its parameter was ignored, which usually leads to retries rather than a fix.

Since the tool already returns error strings (`"No query provided"`, `"Error: query is required"`), returning an explicit error for an invalid mode would be consistent and much easier to debug.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/MemorySearchTool.java
```diff
@@ -52,30 +51,38 @@ public MemorySearchTool(WorkspaceManager workspaceManager) {
...
-
         for (String relativePath : memoryPaths) {
             String content = workspaceManager.readManagedWorkspaceFileUtf8(rc, relativePath);
             if (content == null || content.isEmpty()) {
                 continue;
             }
             String[] lines = content.split("\n", -1);
             for (int i = 0; i < lines.length; i++) {
-                if (pattern.matcher(lines[i]).find()) {
+                if (matchMode.matches(lines[i], query)) {
```
oss-maintainer, line 85: **[Info]** This is the call site that makes the per-record compilation visible: `matchMode.matches(lines[i], query)` runs the mode dispatch + pattern compile + keyword split inside the inner loop. Hoisting a prepared matcher out of the loop (as `#3062` does with a `Predicate<String>`) keeps the same behaviour at the original cost.

PR #3058, opened 2026-09-09
fix(core): retry streaming transport errors wrapped by ModelHttpException without status code

### LINKED ISSUE #3057: [Bug]: Model retry misclassifies streaming transport errors (Connection reset) as non-retryable when wrapped by OpenAIException with null status code
**Describe the bug**

Model calls that fail with a streaming transport error (e.g. `Connection reset` on a stale pooled connection) are **not retried**, even though the agent is configured with `ExecutionConfig.MODEL_DEFAULTS` (`maxAttempts=3`) and the log confirms the retry config was applied:

```
DEBUG io.agentscope.core.model.ModelUtils - Applied retry config: maxAttempts=3, initialBackoff=PT2S for model: qwen3.5-flash
... (1 ms later, no retry happened)
INFO  ...AgentTraceMiddleware - ERROR | OpenAIException: HTTP transport error during streaming: SSE/NDJSON stream failed: java.net.SocketException: Connection reset
```

The error propagates to the caller immediately. A manual retry seconds later succeeds on a fresh connection, so the error is transient and clearly retryable.

**To Reproduce**

1. Build a `HarnessAgent` (or `ReActAgent`) with an OpenAI-compatible model (`OpenAIChatModel` via `agentscope-extensions-model-openai`, default `JdkHttpTransport`) and `.modelExecutionConfig(ExecutionConfig.MODEL_DEFAULTS)`.
2. Make a model call so the HTTP connection pool holds a connection to the endpoint, then keep it idle long enough (e.g. 1–2 hours) for the server / intermediate device to close it.
3. Make another model call on the same JVM. The request is written to the stale pooled connection and receives a TCP RST within ~150 ms.

**Expected behavior**

A connection error during streaming (Connection reset / broken pipe on a stale pooled connection) should be retried according to the `ExecutionConfig` retry policy — the Javadoc of `ExecutionConfig` states that **"Network/IO errors"** are retryable.

**Error messages**

```
io.agentscope.extensions.model.openai.exception.OpenAIException:
    HTTP transport error during streaming: SSE/NDJSON stream failed: java.net.SocketException: Connection reset
  Caused by: io.agentscope.core.model.transport.HttpTransportException:
    SSE/NDJSON stream failed: java.net.SocketException: Connection reset      (statusCode = null)
  Caused by: java.util.concurrent.CompletionException: java.net.SocketException: Connection reset
  Caused by: java.net.SocketException: Connection reset
```

**Environment**

- AgentScope-Java Version: 2.0.3
- Java Version: 17
- OS: Windows 11 (server side), model endpoint: DashScope OpenAI-compatible mode
- Modules: `agentscope-extensions-model-openai`, default transport `JdkHttpTransport` via `HttpTransportFactory.getDefault()`

**Additional context**

Root cause — for streaming transport failures, `OpenAIClient.stream` maps the exception via `new OpenAIException(msg, ex)`, i.e. **no HTTP status code**. `OpenAIException implements ModelHttpException`, so in `ExecutionConfig.isRetryableError` (v2.0.3):

```java
if (error instanceof HttpTransportException hte) {
    return hte.isRetryable();            // (1) never reached — outer exception is OpenAIException
}
if (error instanceof ModelHttpException mhe) {
    return mhe.isRetryableHttpStatus();  // (2) OpenAIException matches here;
                                         //     statusCode == null → returns false immediately
}
if (error instanceof TimeoutException) { return true; }
if (error instanceof IOException) { return true; }   // (3) never reached
// cause-chain recursion                        // (4) never reached
```

Branch (2) returns `false` without consulting the cause chain, which bypasses:

- `HttpTransportException.isRetryable()` — which correctly returns `true` for `statusCode == null` ("Connection errors are usually retryable"), and
- the `IOException` rule (`SocketException` is an `IOException`).

So any streaming transport failure raised through the OpenAI extension is misclassified as a permanent client error and is never retried under `MODEL_DEFAULTS`.

Suggested fix (either works):

1. In `isRetryableError`, when a `ModelHttpException` has `getStatusCode() == null`, fall through to cause-chain inspection instead of returning `false` immediately.
2. Or make `OpenAIException` not report itself as a `ModelHttpException` when constructed without a status code.

Workaround (currently in use) — a custom `retryOn` predicate that first applies `ExecutionConfig.RETRYABLE_ERRORS`, then walks the cause chain for an `HttpTransportException` with a null status code:

```java
private static boolean isModelErrorRetryable(Throwable error) {
    if (ExecutionConfig.RETRYABLE_ERRORS.test(error)) {
        return true;
    }
    for (Throwable cause = error.getCause(); cause != null && cause != error; cause = cause.getCause()) {
        if (cause instanceof HttpTransportException transport && transport.getStatusCode() == null) {
            return true;
        }
    }
    return false;
}
```

Happy to submit a PR for fix option 1 if the approach is agreed.

wahllllll:
## Description

Model exceptions that implement `ModelHttpException` but carry **no HTTP status code** (e.g. `OpenAIException` wrapping a streaming transport failure) were classified as **non-retryable** by `ExecutionConfig.isRetryableError`, because the `ModelHttpException` branch returned `isRetryableHttpStatus()` — which is `false` for a null status — **without consulting the cause chain**. This bypassed both `HttpTransportException.isRetryable()` (which correctly returns `true` for connection errors without a status code) and the `IOException` rule, contradicting the documented *"Network/IO errors are retryable"* intent.

As a result, transient connection errors such as `Connection reset` on a stale pooled connection were never retried under `MODEL_DEFAULTS`, even though the log confirms `Applied retry config: maxAttempts=3`.

Fixes #3057

## Changes

**1. Retry classification fix (`ExecutionConfig.isRetryableError`)**
Only classify by HTTP status when a status code is present (`mhe.getStatusCode() != null`); otherwise fall through to the transport/IO and cause-chain checks.

**2. Emission guard for streaming retries (`ModelUtils.applyTimeoutAndRetry`)** (per review feedback)
Retrying a streaming call after partial output has already been emitted would reissue the request and downstream would observe the first partial response followed by the retried one, duplicating content. Retries are now allowed **only before the first `ChatResponse` of the current subscription is emitted**; once any attempt has emitted, further retries are rejected regardless of the error classification. All streaming models routed through `applyTimeoutAndRetry` benefit.

## Tests

- `ExecutionConfigTest`: null-status `ModelHttpException` wrapping a `HttpTransportException` / `SocketException` must be retryable; a null-status exception **without** a retryable cause stays non-retryable (existing behavior preserved).
- `OpenAIExceptionRetryClassificationTest` (openai extension module): the exact production exception chain from issue #3057 — `OpenAIException(msg, HttpTransportException(stream failed, SocketException))` → retryable; status-code based classification (429/5xx retryable, 400/401 not) unchanged.
- `ModelTimeoutRetryTest`: connection reset **before** the first chunk → retried and succeeds; reset **after** one emitted chunk → chunk delivered, error propagates, no retry.

## How Has This Been Tested?

- [x] `mvn -pl agentscope-core test -Dtest='ExecutionConfigTest,ModelTimeoutRetryTest'` — 21/21 pass
- [x] `mvn -pl agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai -am test -Dtest=OpenAIExceptionRetryClassificationTest` — 4/4 pass
- [x] `mvn spotless:check` passes
- [x] Verified against the production failure from issue #3057 (DashScope compatible-mode, default `JdkHttpTransport`, ~2h idle pooled connection): the same exception chain is now classified retryable and the retry succeeds on a fresh connection

## Notes

The existing test `shouldNotRetryModelHttpExceptionWithoutStatusCode` (null status, **no cause**) still passes unchanged — only exceptions whose cause chain contains a retryable transport/IO error change behavior.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ExecutionConfig.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ExecutionConfig.java
@@ -103,7 +103,11 @@ public class ExecutionConfig {
             return hte.isRetryable();
         }
 
-        if (error instanceof ModelHttpException mhe) {
+        // Only treat a ModelHttpException as an HTTP response error when a status code is
+        // present. Implementations without a status code (e.g. OpenAIException wrapping a
+        // streaming transport failure) must fall through to the transport/IO and cause-chain
+        // checks below instead of being classified as a permanent client error (issue #3057).
+        if (error instanceof ModelHttpException mhe && mhe.getStatusCode() != null) {
             return mhe.isRetryableHttpStatus();
         }
 
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
@@ -16,6 +16,7 @@
 package io.agentscope.core.model;
 
 import java.time.Duration;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Predicate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -110,25 +111,57 @@ public final class ModelUtils {
                 if (retryOn == null) {
                     retryOn = error -> true; // retry all errors by default
                 }
+                // Lambda captures below require effectively final variables
+                final int retryCount = maxAttempts - 1;
+                final Duration effectiveInitialBackoff = initialBackoff;
+                final Duration effectiveMaxBackoff = maxBackoff;
+                final Predicate<Throwable> effectiveRetryOn = retryOn;
 
-                Retry retrySpec =
-                        Retry.backoff(maxAttempts - 1, initialBackoff)
-                                .maxBackoff(maxBackoff)
-                                .jitter(0.5)
-                                .filter(retryOn)
-                                .doBeforeRetry(
-                                        signal ->
-                                                LOG.warn(
-                                                        "Retrying model request (attempt {}/{}) due"
-                                                                + " to: {}",
-                                                        signal.totalRetriesInARow() + 1,
-                                                        maxAttempts - 1,
-                                                        signal.failure().getMessage(),
-                                                        signal.failure()));
-
-                responseFlux = responseFlux.retryWhen(retrySpec);
+                // Retrying is only safe before the first response is emitted: resubscribing
+                // after partial output would reissue the request and downstream would observe
+                // the first partial response followed by the retried one, duplicating content.
+                // The emission flag is created inside defer so every subscription (model call)
+                // gets a fresh flag, while it persists across retry attempts of the same call.
+                final Flux<ChatResponse> source = responseFlux;
+                responseFlux =
+                        Flux.defer(
+                                () -> {
+                                    AtomicBoolean emittedAnyResponse = new AtomicBoolean(false);
+                                    return source.doOnNext(response -> emittedAnyResponse.set(true))
+                                            .retryWhen(
+                                                    Retry.backoff(
+                                                                    retryCount,
+                                                                    effectiveInitialBackoff)
+                                                            .maxBackoff(effectiveMaxBackoff)
+                                                            .jitter(0.5)
+                                                            .filter(
+                                                                    error ->
+                                                                            !emittedAnyResponse
+                                                                                            .get()
+                                                                                    && effectiveRetryOn
+                                                                                            .test(
+                                                                                                    error))
+                                                            .doBeforeRetry(
+                                                                    signal ->
+                                                                            LOG.warn(
+                                                                                    "Retrying model"
+                                                                                        + " request"
+                                                                                        + " (attempt"
+                                                                                        + " {}/{})"
+                                                                                        + " due to:"
+                                                                                        + " {}",
+                                                                                    signal
+                                                                                                    .totalRetriesInARow()
+                                                                                            + 1,
+                                                                                    retryCount,
+                                                                                    signal.failure()
+                                                                                            .getMessage(),
+                                                                                    signal
+                                                                                            .failure())));
+                                });
                 LOG.debug(
-                        "Applied retry config: maxAttempts={}, initialBackoff={} for model: {}",
+                        "Applied retry config: maxAttempts={}, initialBackoff={},"
+                                + " retryBeforeFirstResponse=true for model: {}",
                         maxAttempts,
                         initialBackoff,
                         modelName);
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
@@ -15,7 +15,9 @@
  */
 package io.agentscope.core.model;
 
+import io.agentscope.core.message.ContentBlock;
 import java.time.Duration;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Predicate;
 import org.slf4j.Logger;
@@ -117,17 +119,24 @@ public final class ModelUtils {
                 final Duration effectiveMaxBackoff = maxBackoff;
                 final Predicate<Throwable> effectiveRetryOn = retryOn;
 
-                // Retrying is only safe before the first response is emitted: resubscribing
+                // Retrying is only safe before user-visible content is emitted: resubscribing
                 // after partial output would reissue the request and downstream would observe
                 // the first partial response followed by the retried one, duplicating content.
-                // The emission flag is created inside defer so every subscription (model call)
-                // gets a fresh flag, while it persists across retry attempts of the same call.
+                // The visibility flag is created inside defer so every subscription (model
+                // call) gets a fresh flag, while it persists across retry attempts of the same
+                // call. Role-only or usage-only chunks carry no content blocks and do not
+                // disable retries — nothing user-visible has been delivered yet.
                 final Flux<ChatResponse> source = responseFlux;
                 responseFlux =
                         Flux.defer(
                                 () -> {
-                                    AtomicBoolean emittedAnyResponse = new AtomicBoolean(false);
-                                    return source.doOnNext(response -> emittedAnyResponse.set(true))
+                                    AtomicBoolean emittedVisibleContent = new AtomicBoolean(false);
+                                    return source.doOnNext(
+                                                    response -> {
+                                                        if (hasVisibleContent(response)) {
+                                                            emittedVisibleContent.set(true);
+                                                        }
+                                                    })
                                             .retryWhen(
                                                     Retry.backoff(
                                                                     retryCount,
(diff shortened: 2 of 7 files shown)


agentscope-core/src/main/java/io/agentscope/core/model/ExecutionConfig.java
```diff
@@ -103,7 +103,11 @@ private static boolean isRetryableError(Throwable error) {
             return hte.isRetryable();
         }
 
-        if (error instanceof ModelHttpException mhe) {
+        // Only treat a ModelHttpException as an HTTP response error when a status code is
+        // present. Implementations without a status code (e.g. OpenAIException wrapping a
+        // streaming transport failure) must fall through to the transport/IO and cause-chain
+        // checks below instead of being classified as a permanent client error (issue #3057).
+        if (error instanceof ModelHttpException mhe && mhe.getStatusCode() != null) {
```
dailingtao, line 110: [P1] Please avoid retrying after a streaming response has already emitted any chunks. With this change, a null-status ModelHttpException can fall through to a retryable transport/IO cause regardless of stream progress. ModelUtils applies retryWhen to the entire response Flux, so if the first attempt emits partial text or tool-call deltas and then the connection resets, resubscription reissues the request and downstream observes the first partial output followed by the second response, which can duplicate text or tool calls. AgentScope Python and the OpenAI Python SDK limit automatic retries to establishing the request/stream rather than consuming an already-started stream. Could we guard retries so they are allowed only before the first ChatResponse is emitted, and add tests covering both a reset before the first chunk (retries) and a reset after one chunk (does not retry)?

wahllllll: Thanks for the careful review — you're right, resubscribing after partial output would duplicate already-delivered chunks downstream.

I've pushed a guard in `ModelUtils.applyTimeoutAndRetry`: retries are now allowed only while the current subscription has not emitted any `ChatResponse` yet. The emission flag is created inside `Flux.defer` so each model call gets a fresh flag, while it persists across the retry attempts of the same call (once any attempt has emitted, further retries are rejected regardless of the error classification). All streaming models routed through `applyTimeoutAndRetry` benefit.

Added the two tests you asked for in `ModelTimeoutRetryTest`, both using the exact exception chain from the issue (`HttpTransportException` with null status wrapping `SocketException`):

- reset before the first chunk → retried, second attempt succeeds
- reset after one emitted chunk → the chunk is delivered, the error propagates, no retry

Commit history was rewritten to keep the two changes as separate commits.

agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
```diff
@@ -110,25 +111,57 @@ public static Flux<ChatResponse> applyTimeoutAndRetry(
...
+                                () -> {
+                                    AtomicBoolean emittedAnyResponse = new AtomicBoolean(false);
+                                    return source.doOnNext(response -> emittedAnyResponse.set(true))
+                                            .retryWhen(
+                                                    Retry.backoff(
+                                                                    retryCount,
+                                                                    effectiveInitialBackoff)
+                                                            .maxBackoff(effectiveMaxBackoff)
+                                                            .jitter(0.5)
+                                                            .filter(
```
oss-maintainer, line 137: The emission guard is the right call — retrying after partial output would duplicate content. But `emittedAnyResponse` is set by *any* `ChatResponse`, including providers that emit an initial empty/role-only chunk before real deltas. For those streams the guard disables retries for the whole call, which is exactly the case this PR fixes. Consider gating on "emitted user-visible content" (non-empty content blocks) instead of any response, or at least logging once when a retry is suppressed by the guard so this is diagnosable in production.

agentscope-core/src/main/java/io/agentscope/core/model/ExecutionConfig.java
```diff
@@ -103,7 +103,11 @@ private static boolean isRetryableError(Throwable error) {
             return hte.isRetryable();
         }
 
-        if (error instanceof ModelHttpException mhe) {
+        // Only treat a ModelHttpException as an HTTP response error when a status code is
```
oss-maintainer, line 106: Good fix — the status-code branch no longer short-circuits the cause-chain checks when `getStatusCode()` is `null`, and `shouldNotRetryModelHttpExceptionWithoutStatusCode` keeps the old contract. Since this changes retry behaviour for every provider implementing `ModelHttpException`, please confirm the non-streaming 4xx paths still land in `isRetryableHttpStatus()` in the extension-module tests (openai/anthropic/dashscope), not just the openai one.

oss-maintainer (commented):
### Summary
Two-part fix and both parts are needed. Classification: a `ModelHttpException` implementation with a null status code short-circuited `isRetryableError` to `false` without ever consulting `HttpTransportException.isRetryable()` or the `IOException` rule, so `Connection reset` on a stale pooled connection was never retried despite `maxAttempts=3`. Gating on `getStatusCode() != null` restores the documented intent. The emission guard is the right companion change, but see the note about providers that emit an initial empty chunk.

---

agentscope-core/src/main/java/io/agentscope/core/model/ModelUtils.java
wahllllll: Good catch — role-only/usage-only chunks do map to `ChatResponse`s with an empty content list (verified in `OpenAIResponseParser.parseChunkResponse`: text/thinking/tool-call deltas all land in content blocks, a role-only delta produces none). Pushed in 826495a1:

- the guard flag is now set only by chunks that carry content blocks (user-visible text/thinking/tool-call deltas), so an early reset after a role-only chunk is still retried;
- when the guard suppresses an otherwise-retryable error, a single `WARN` is logged with the model name and error so it is diagnosable in production;
- added `shouldRetryAfterEmptyContentChunks` covering exactly the shape you described: role-only chunk → connection reset → retried, second attempt succeeds.

agentscope-core/src/main/java/io/agentscope/core/model/ExecutionConfig.java
wahllllll: Checked the other model extensions: only the openai module has a `ModelHttpException` implementation (`OpenAIException`). The anthropic and dashscope models surface transport failures as plain `HttpTransportException` (no status code for connection errors, 4xx/5xx with status codes), so their classification goes through the `HttpTransportException` branch of `isRetryableError` and is unchanged by this fix. To pin that down, 826495a1 adds `AnthropicRetryClassificationTest` and `DashScopeRetryClassificationTest` asserting: connection error without status → retryable, 429/5xx → retryable, 400/401 → not retryable.

oss-maintainer (approved):
### Summary
Commit `826495a1` addresses the note from the previous review exactly: the retry gate now tracks *visible* content (`hasVisibleContent` — non-empty content blocks) instead of any emitted chunk, so role-only/usage-only preamble chunks no longer disable retries, while a retry that would duplicate delivered output is skipped with an explicit WARN naming the cause. New StepVerifier test covers the role-only-chunk-then-reset retry path, and the per-provider classification tests pin the `HttpTransportException` shapes. LGTM.

### CI
builds were still queued at review time; previous pushes were green.

---
*Automated review by ${BOT_SIG}*

oss-maintainer:

This PR currently conflicts with `main` and cannot be merged (head `fix/model-retry-null-status`). Please rebase (or merge `main` in) and resolve the conflicts:

```bash
git fetch upstream
git checkout fix/model-retry-null-status
git rebase upstream/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #3061, opened 2026-09-09
test(openai): cover ReAct streamEvents backpressure regression

guslegend0510:
## AgentScope-Java Version

2.0.3-SNAPSHOT

## Description

Refs #3055.
Related issue: #2957.
Follow-up to the transport fix in #2963.

The streaming stall described in #3055 matches the request-starvation issue already fixed in OkHttpTransport by #2963.

This PR adds regression coverage through OkHttpTransport, OpenAIChatModel, and ReActAgent.streamEvents():

- Require all 96 text deltas to arrive in order before sending `[DONE]`.
- Cover ordinary consumption and cross-thread incremental demand.
- Verify exactly one model request and one model-call end event.
- Document why the blocking transport uses `subscribeOn(Schedulers.boundedElastic(), false)`.

The existing runtime fix is preserved; this PR adds regression tests and an explanatory comment.

### Validation

- Temporarily restoring the old request scheduling makes both new tests time out.
- Both tests pass with the existing fix.
- Full Core suite: 2,323 tests, 9 skipped.
- Full Harness suite: 1,010 tests, 14 skipped.
- Full OpenAI module suite: 557 tests, 12 skipped.
- Zero failures or errors across these suites.
- Spotless, required focused checks, and `git diff --check` passed.

The reproduction uses a controlled streaming response body; no live model API was called.

## Checklist

- [x] Code has been formatted.
- [x] Relevant module test suites pass.
- [x] Comments explain the scheduling constraint.
- [x] No public API or user-facing documentation changes are needed.


oss-maintainer (commented):
### Summary
Regression coverage for the #3055 streaming stall: a new `OpenAIReActAgentStreamingBackpressureTest` that drives 96 SSE deltas through `OkHttpTransport` → `OpenAIChatModel` → `ReActAgent.streamEvents()` and asserts every delta is delivered before `[DONE]`, under both eager and incremental (`publishOn(parallel(), 1)`) demand, plus exactly one model request and one `ModelCallEndEvent`. The only production change is an explanatory comment on `subscribeOn(Schedulers.boundedElastic(), false)`. Clean, valuable test-first PR — the negative control (restoring the old scheduling makes both tests time out) is exactly the evidence a regression test needs, and the fake-`Pipe` harness avoids any live model call.

Two small notes, neither blocking:
- **[Info]** `agentscope-extensions-model-openai/src/test/.../OpenAIReActAgentStreamingBackpressureTest.java:78` — the response body is created with a `null` content type; passing `MediaType.parse("text/event-stream")` would make the fixture closer to what a real OpenAI-compatible endpoint returns, in case the decoder ever starts inspecting the content type.
- **[Info]** same file, `responseSink` is fully buffered *before* the request is subscribed. That is safe for 96 chunks (~8 KB against the 64 KB pipe) but will block the test thread (rather than fail cleanly) if someone scales `expected` up past the buffer. A one-line comment on that limit, or writing chunks from a separate task, would keep it future-proof.

**CLA:** signed ✅

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/test/java/io/agentscope/extensions/model/openai/OpenAIReActAgentStreamingBackpressureTest.java
```diff
@@ -0,0 +1,156 @@
...
+                responseSink.writeUtf8(
+                        "data: {\"id\":\"test\",\"choices\":[{\"index\":0,"
+                                + "\"delta\":{\"content\":\""
+                                + text
+                                + "\"}}]}\n\n");
+            }
+            responseSink.flush();
+
+            Msg input =
+                    Msg.builder()
```
oss-maintainer, line 111: All 96 SSE chunks (~15 KB) are written synchronously to the pipe sink before StepVerifier subscribes and the consumer starts reading. This works because the data fits comfortably in the 64 KB pipe buffer (line 68), but if a future change increases chunk count or size past ~64 KB, the write loop will block waiting for a reader that hasn't been subscribed yet — a test deadlock. Consider writing chunks on a background thread (e.g., via `Schedulers.boundedElastic().schedule(...)`) or reducing the buffer dependency by interleaving writes with subscription, so the test remains safe at larger data volumes.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/test/java/io/agentscope/extensions/model/openai/OpenAIReActAgentStreamingBackpressureTest.java
```diff
@@ -0,0 +1,156 @@
...
+        OkHttpClient client =
+                new OkHttpClient.Builder()
+                        .addInterceptor(
+                                chain -> {
+                                    requests.incrementAndGet();
+                                    return new Response.Builder()
+                                            .request(chain.request())
+                                            .protocol(Protocol.HTTP_1_1)
+                                            .code(200)
+                                            .message("OK")
```
oss-maintainer, line 75: The OkHttpClient created here is never explicitly shut down (no `client.dispatcher().executorService().shutdown()` or `client.connectionPool().evictAll()`). In a large test suite this can accumulate idle dispatcher threads. Consider wrapping it in a try-with-resources via a helper or calling `client.dispatcher().executorService().shutdown()` in the finally block alongside `transport.close()`.

oss-maintainer (commented):
### Summary
Re-review of PR #3061 at its current head (automated review of the changes in this diff). 2 item(s) raised below; nothing here blocks the approach itself — details inline.

### Findings
- **[Warning]** `agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/test/java/io/agentscope/extensions/model/openai/OpenAIReActAgentStreamingBackpressureTest.java:111` — All 96 SSE chunks (~15 KB) are written synchronously to the pipe sink before StepVerifier subscribes and the consumer starts reading. This works because the data fits comfortably in the 64 KB pipe buffer (line 68), but if a future change increases chunk count or size past ~64 KB, the write loop will block waiting for a reader that hasn't been subscribed yet — a test deadlock. Consider writing chunks on a background thread (e.g., via `Schedulers.boundedElastic().schedule(...)`) or reducing the buffer dependency by interleaving writes with subscription, so the test remains safe at larger data volumes.
- **[Info]** `agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai/src/test/java/io/agentscope/extensions/model/openai/OpenAIReActAgentStreamingBackpressureTest.java:75` — The OkHttpClient created here is never explicitly shut down (no `client.dispatcher().executorService().shutdown()` or `client.connectionPool().evictAll()`). In a large test suite this can accumulate idle dispatcher threads. Consider wrapping it in a try-with-resources via a helper or calling `client.dispatcher().executorService().shutdown()` in the finally block alongside `transport.close()`.

### Notes

---

oss-maintainer (commented):
### Summary
Delta re-review of `ce15336f7f` → `77ab186826` (current head). The 6 commits in between are `origin/main` merges only: neither file this PR touches (`OkHttpTransport.java`, `OpenAIReActAgentStreamingBackpressureTest.java`) differs from the state I reviewed on Sep 13, so the test code is unchanged and still reads well — the ordered 96-delta assertion plus the "exactly one model request / exactly one `ModelCallEndEvent`" invariants are the right things to pin for #3055, and the `subscribeOn(Schedulers.boundedElastic(), false)` comment documents the constraint #2963 introduced.

### Corrections and open items
- **[Info] Retracting one item from my previous pass.** I flagged the `OkHttpClient` built in `assertChunksArriveBeforeDone` as never shut down. That was wrong: the `finally { transport.close(); }` does reach `OkHttpTransport.close()`, which calls `dispatcher().executorService().shutdown()` and `connectionPool().evictAll()` (`OkHttpTransport.java:314-317`), and `ReActAgent` is already inside the try-with-resources. No leak — please disregard that comment.
- **[Info]** `OpenAIReActAgentStreamingBackpressureTest.java:63` — the `Pipe(64 * 1024)` sink has room for all 96 frames (~7 KB), so the pre-subscription write loop never blocks. That is what makes the test deterministic, but it also means the buffer size is load-bearing: all 96 `writeUtf8` calls happen on the test thread *before* `StepVerifier.create(deltas)` subscribes, so a future edit that raises the frame count past 64 KiB would hang the test thread on a full pipe rather than exercise the starvation path. Worth a one-line note next to the constant. Nothing blocking here.
- **[Warning]** CI on this head is not a usable signal yet: `build (ubuntu-latest)` ended `cancelled` (runner/resource cancellation, not a test failure) and `build (windows-latest)` failed in `LocalFilesystemPersonalAssistantExampleTest.localFilesystem_filesPersistAcrossCalls:110` — `MEMORY.md content should be unchanged after call 2 ==> expected: <# Notes\n- item 1> but was: <done>`. That is a harness memory-flush test in a module this PR does not touch, and `main` is green at `6c8ccc5c6`, so it looks unrelated to your change. A re-run should settle it.

Not approving while CI is red on the head; otherwise this is a low-risk, test-only diff and I have no further content objections.

---

PR #3062, opened 2026-09-09
feat(harness): 为记忆与会话检索增加可选多关键词匹配模式

### LINKED ISSUE #3054: [Feature] 建议为 memory_search / session_search 增加可选的多关键词匹配模式

wzq-xzwj:
Closes #3054

## 背景

模型有时会传入空格分隔的多个关键词，例如 `query="部署 蓝鲸"`。当前两个检索工具都把完整 query 当作字面短语，因此无法命中 `部署决定：使用蓝鲸方案`。

这个 PR 提供一个可选的多关键词匹配方案。默认行为保持不变，具体参数和语义可以根据评审意见调整。

## 方案

为 `memory_search` 和 `session_search` 增加可选字符串参数 `matchMode`：

| 值 | 语义 |
| --- | --- |
| `phrase`（默认） | 按完整 query 做字面子串匹配，保留原有空白和顺序 |
| `all` | 按 Unicode 空白拆分，所有关键词都要出现在同一条记录中，顺序不限 |
| `any` | 按 Unicode 空白拆分，同一条记录包含至少一个关键词即可 |

调用示例：

```json
{"query": "部署 蓝鲸", "matchMode": "all"}
```

`session_search` 仍可同时传入 `agentId` 和 `maxResults`。

### 边界与兼容性

- Memory 仍按行匹配，Session 仍按 entry 匹配，不跨记录组合关键词。
- 省略 `matchMode` 或传入 `null` 时使用 `phrase`。值区分大小写；非法值（包括空字符串）返回明确错误，不静默切换模式。
- 多关键词模式忽略多余空白、去重相同关键词；空查询不会因为 `allMatch` 的空集合语义而匹配所有记录。
- 关键词按字面值匹配，正则特殊字符不作为表达式执行。
- 保留两个工具各自现有的大小写匹配实现，避免在这个 PR 中额外改变 Unicode/Locale 行为。
- 保留旧 Java 方法签名，由旧方法转调新重载；只有新重载带 `@Tool`，避免重复注册。
- 保留查询范围、结果格式、顺序和数量限制；不新增相关性排序。

## 实现

- 新增包内辅助类 `KeywordMatcher`，共用模式校验、空白拆词和 all/any 组合逻辑；每次查询创建匹配器，不在每条记录上重新拆词或编译正则。
- `MemorySearchTool` 和 `SessionSearchTool` 接入匹配器，并补充工具参数描述。
- 更新中英文 Memory 文档，包含调用示例、默认值、记录边界和限制。
- 新增 `KeywordSearchModesTest`，同时验证两个工具的实际文件检索及 Toolkit 注册/调用。

## 测试

本地使用 JDK 21、Java release 17 编译，以下定向测试共 **83 项通过，0 失败、0 错误、0 跳过**，其中新增多关键词测试 18 项：

```bash
mvn -pl agentscope-harness -am test \
  -Dtest=KeywordSearchModesTest,SessionTranscriptWriterTest,SessionTreeMirrorTest,MemoryConsolidatorFilesystemTest,HarnessAgentTest \
  -Dsurefire.failIfNoSpecifiedTests=false -Dspotless.skip=true
```

另行对本次修改的四个 Java 文件执行 Spotless 格式化与检查，通过：

```bash
mvn -pl agentscope-harness spotless:check \
  '-DspotlessFiles=.*KeywordMatcher.java,.*MemorySearchTool.java,.*SessionSearchTool.java,.*KeywordSearchModesTest.java'
```

`git diff --cached --check` 通过。以上为定向回归，不是全仓测试，也没有调用真实 LLM 验证模型选择模式的概率。

覆盖：默认/旧 Java 入口、完整短语空白保留、中文关键词、英文大小写、不相邻及乱序关键词、all/any 无匹配、正则特殊字符、Unicode 空白、重复词、空查询、无效模式、不跨行/entry、daily ledger、agentId 过滤、maxResults，以及新参数的 schema 和反射调用。

## 不在本次范围内

- 连续中文句子的自动分词、同义词和语义检索。
- “昨天”等时间表达式解析。
- 历史内容截断、数据源迁移或记忆提取流程。
- 自动决定何时从 Memory 回退到 Session 检索。

没有增加第三方依赖，也没有改默认模式为 OR 匹配。该改动改善显式多关键词查询，不代表模型一定会选择新模式。


agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SessionSearchTool.java
```diff
@@ -68,15 +80,35 @@ public String sessionSearch(
...
-        String lowerQuery = query.toLowerCase();
+        Predicate<String> matcher;
+        try {
+            matcher =
+                    KeywordMatcher.compile(
+                            query,
+                            matchMode,
+                            term -> {
+                                String lowerTerm = term.toLowerCase();
+                                return text -> text.toLowerCase().contains(lowerTerm);
```
dailingtao, line 107: Non-blocking: each term predicate lowercases the full session entry independently, so an `all`/`any` query with N terms can allocate N lowercase copies per entry. Could we normalize `content` once per entry (while preserving the existing locale behavior) and run the predicates against that normalized string?

dailingtao (approved):
LGTM

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/MemorySearchTool.java
```diff
@@ -51,22 +57,46 @@ public MemorySearchTool(WorkspaceManager workspaceManager) {
...
-        return keywordSearch(rc, query);
+        Predicate<String> matcher;
+        try {
+            matcher =
+                    KeywordMatcher.compile(
+                            query,
+                            matchMode,
+                            term ->
+                                    Pattern.compile(Pattern.quote(term), Pattern.CASE_INSENSITIVE)
+                                            .asPredicate());
```
oss-maintainer, line 88: **[Warning]** The memory path matches via `Pattern.compile(Pattern.quote(term), CASE_INSENSITIVE).asPredicate()`, but `SessionSearchTool` (line 107) matches via `text.toLowerCase().contains(lowerTerm)`. Both are correct today, yet they encode *different* notions of case-insensitivity: `toLowerCase()` is locale-sensitive (Turkish-I dotless issue) while `CASE_INSENSITIVE` regex is not, so the same query can behave differently between the two tools on non-ASCII text.

Consider routing both through the same literal matcher, or at minimum `toLowerCase(Locale.ROOT)`, so `memory_search` and `session_search` cannot drift.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/SessionSearchTool.java
oss-maintainer, line 107: **[Warning]** Same point as the memory-side comment: `text.toLowerCase()` here uses the default locale, and is also evaluated per term per record (for `mode=all` with N keywords, the same `text` is lowercased N times). Resolving the text once per record and using `Locale.ROOT` fixes both the allocation and the locale sensitivity.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/KeywordMatcher.java
```diff
@@ -0,0 +1,54 @@
...
+    private KeywordMatcher() {}
+
+    static Predicate<String> compile(
+            String query, String matchMode, Function<String, Predicate<String>> literalMatcher) {
+        String mode = matchMode == null ? "phrase" : matchMode;
+        if (mode.equals("phrase")) {
+            return literalMatcher.apply(query);
+        }
+        if (!mode.equals("all") && !mode.equals("any")) {
+            throw new IllegalArgumentException("matchMode must be one of: phrase, all, any");
```
oss-maintainer, line 38: **[Info]** Returning an explicit error for an unknown mode is the right call for a model-supplied parameter (the alternative, silently falling back to `phrase`, produces "no results" that the agent cannot diagnose). Worth a line in the docs table, since #3056 chooses the silent-fallback behaviour for the same input — if both land, the flag would mean different things depending on which merged.

oss-maintainer (commented):
### Summary

Adds an optional `matchMode` (`phrase` | `all` | `any`) to `memory_search` and `session_search`, factored into a `KeywordMatcher` that builds one `Predicate<String>` per search. This is a well-judged change: the matcher is compiled once outside the scan loop, invalid modes return an explicit error instead of silently degrading, the original Java signatures are preserved as overloads, and both the en and zh docs are updated. The test coverage is unusually thorough for a change of this size — literal regex metacharacters, Unicode whitespace, per-record (non-cross-record) semantics, empty-query behaviour, limit/agentId preservation, and reflective invocation through `Toolkit`.

Left `COMMENT` rather than `APPROVE` only because of the duplicate-feature question below and two small nits; neither is a blocker on the code itself.

### Findings

- **[Warning]** `MemorySearchTool.java:88` — the memory path builds matchers with `Pattern.compile(Pattern.quote(term), CASE_INSENSITIVE).asPredicate()`, while `SessionSearchTool.java:107` uses `text.toLowerCase().contains(lowerTerm)`. Both are correct today, but the two tools now disagree on what "case-insensitive" means: `toLowerCase()` is locale-sensitive (the Turkish-I dotless issue) and the regex path is not, so the same query can behave differently between the two tools on non-ASCII input. Consider routing both through the same literal matcher, or at least `toLowerCase(Locale.ROOT)`.
- **[Info]** `KeywordMatcher.java:38` — an invalid mode throws and the tools convert it to `"Error: matchMode must be one of: phrase, all, any"`. Good. Note this diverges from #3056, which silently falls back to `phrase` for an unknown mode; keeping the explicit error is the better behaviour for a model-supplied parameter, and worth stating in the docs table so the two PRs do not land different semantics.
- **[Info]** `KeywordMatcher.java:48` — the "empty ALL query" guard throws `IllegalArgumentException` with a query-specific message. That reads as a query problem rather than a mode problem; the message is surfaced verbatim to the model, which is fine, but `"query must contain at least one keyword"` under a returned string prefixed `"Error: "` is slightly ambiguous next to the mode error above.

### Cross-PR Note

This implements the same feature as **#3056** (same `matchMode` parameter on the same two tools, same whitespace-split semantics, touching the same files) — the two will conflict. #3056 was opened first but recompiles the phrase pattern per scanned line and silently falls back to `phrase` on an unknown mode, and it has no docs changes; this PR does the hoisted-matcher and error handling correctly and documents the flag in both languages. Since @dailingtao has already approved this one, it is probably the natural survivor — but the maintainers of both should agree, so the effort in #3056 is not lost and the tool schema lands only once.

### Tests

Coverage is strong. One gap: no test pins the *memory* and *session* tools to identical matching semantics for a shared query, which is exactly where the two implementations currently diverge (see the locale note above).

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/MemorySearchTool.java
```diff
@@ -51,22 +57,46 @@ public MemorySearchTool(WorkspaceManager workspaceManager) {
...
-            @ToolParam(name = "query", description = "Keywords to search for in memory files")
-                    String query) {
+            @ToolParam(
+                            name = "query",
+                            description =
+                                    "Literal phrase, or whitespace-separated keywords when"
+                                            + " matchMode is all/any; no automatic Chinese word"
+                                            + " segmentation")
+                    String query,
+            @ToolParam(
```
oss-maintainer, line 67: This PR and #3056 both add an optional matchMode parameter to memory_search/session_search and both claim to close #3054. The two implementations use the same parameter name but conflicting semantics (#3056 silently falls back to phrase for blank/unknown modes, while this PR returns an error), so they cannot both land as-is.

oss-maintainer (commented):
### Summary
Well-engineered feature: default `phrase` behavior is preserved, invalid modes fail loudly instead of silently switching, keywords are Unicode-split/deduped/literal (`Pattern.quote`), the pre-existing Java overloads delegate to the new `@Tool` methods without double registration, and 18 new tests plus a module build pass locally. CI is green and the CLA is signed.

However, this PR **duplicates PR #3056** (also open): both add a `matchMode` parameter (`phrase`/`all`/`any`) to `memory_search`/`session_search` and both close #3054, touching the same two tool files with different helper designs (`KeywordMatcher` + string validation here vs. a `SearchMatchMode` enum there) and conflicting invalid-mode semantics (error here vs. lenient fallback there). They cannot both merge as-is — the maintainers need to pick one, and this needs author coordination before approval.

### Findings
- **[Critical]** `MemorySearchTool.java:67` / `SessionSearchTool.java:84` — same `matchMode` feature is also implemented in #3056 against the same files and issue; divergent invalid-mode semantics make the two mutually exclusive.
- **[Warning]** `KeywordMatcher.java:37` — lowercase-only, strict-throw mode parsing is reasonable on its own, but diverges from #3056's lenient/case-tolerant parsing; whichever PR wins should own the final contract.

### Suggestions
Ping @CryoThrust on #3056, agree on a single `matchMode` contract (invalid-value behavior + enum vs. string parsing), and one of the two PRs should be adapted or superseded. This PR's stricter validation + richer edge-case tests are a good baseline for the merge candidate.

---

PR #3070, opened 2026-09-09
fix(rag): normalize Milvus L2 retrieval scores

### LINKED ISSUE #3069: [Bug] MilvusStore returns inverted scores for the L2 metric
## Description

When `MilvusStore` is configured with `IndexParam.MetricType.L2`, Milvus returns Euclidean distances, where smaller values indicate greater similarity.

`MilvusStore` currently exposes the raw distance as `Document.score`, while AgentScope retrieval consumers and `scoreThreshold` use a higher-is-better score contract.

For example, distances `1.0` and `3.0` are exposed as scores `1.0` and `3.0`. This exposes a less similar result with a higher score, which can cause consumers that rank by higher-is-better scores to order it incorrectly and makes threshold filtering inconsistent.

## Expected behavior

`Document.score` should preserve AgentScope higher-is-better semantics, and `scoreThreshold` should filter results consistently.

liugy789:
## Summary

- Pass the configured metric type to the Milvus `SearchReq`.
- Normalize L2 distances into higher-is-better scores.
- Apply `scoreThreshold` after score normalization.
- Add regression tests for score conversion, metric propagation, and thresholding.

## Why

Milvus L2 search returns distances where smaller values indicate greater similarity. AgentScope retrieval uses a higher-is-better score contract, so raw L2 distances must be adapted before they are exposed as document scores.

This PR uses `1 / (1 + distance)` as an AgentScope-side normalization formula. This formula is not a Milvus-defined conversion.

## Validation

- `MilvusStoreTest`: 65 tests passed.
- RAG simple module: 449 tests passed.
- Spotless check passed.
- `git diff --check` passed.

## References

- [Milvus Metric Types](https://milvus.io/docs/v2.6.x/metric.md)
- [Milvus Basic Vector Search](https://milvus.io/docs/single-vector-search.md)
- [Milvus Java IndexParam API](https://milvus.io/api-reference/java/v2.6.x/v2/Management/IndexParam.md)

Fixes #3069

### DIFF
--- a/agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/store/MilvusStore.java
+++ b/agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/store/MilvusStore.java
@@ -588,6 +588,7 @@ public class MilvusStore implements VDBStoreBase, AutoCloseable {
                         .databaseName(databaseName)
                         .collectionName(collectionName)
                         .data(Collections.singletonList(queryVector))
+                        .metricType(metricType)
                         .limit(limit)
                         .outputFields(
                                 Arrays.asList(
@@ -609,8 +610,9 @@ public class MilvusStore implements VDBStoreBase, AutoCloseable {
         if (searchResults != null && !searchResults.isEmpty()) {
             for (SearchResp.SearchResult result : searchResults.get(0)) {
                 try {
-                    // Get score
-                    double score = result.getScore();
+                    // Milvus returns a distance for L2. Convert it to a higher-is-better score
+                    // before applying the shared threshold.
+                    double score = normalizeScore(result.getScore());
 
                     // Apply score threshold if specified
                     if (scoreThreshold != null && score < scoreThreshold) {
@@ -629,6 +631,23 @@ public class MilvusStore implements VDBStoreBase, AutoCloseable {
         }
 
         return results;
+    }
+
+    /**
+     * Converts a raw Milvus result to the score contract used by AgentScope.
+     *
+     * <p>L2 results are distances where lower values are more similar, so map them to a
+     * monotonically decreasing score in the [0, 1] range. Other supported metrics already return
+     * scores in the expected direction and are left unchanged.
+     *
+     * @param rawScore the raw score or distance returned by Milvus
+     * @return a higher-is-better score
+     */
+    private double normalizeScore(double rawScore) {
+        if (metricType == IndexParam.MetricType.L2) {
+            return 1.0 / (1.0 + rawScore);
+        }
+        return rawScore;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dailingtao (approved):
LGTM

agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/store/MilvusStore.java
```diff
@@ -588,6 +588,7 @@ private List<Document> searchDocumentsInMilvus(
                         .databaseName(databaseName)
                         .collectionName(collectionName)
                         .data(Collections.singletonList(queryVector))
+                        .metricType(metricType)
```
oss-maintainer, line 591: Passing `metricType` explicitly to `SearchReq` is new behaviour: previously Milvus searched with the metric recorded on the collection index, now it searches with the *configured* metric. When a store connects to an existing collection whose index metric differs from the builder default (`COSINE`), this turns a previously-working search into a provider-side error or wrong ranking. Could you validate the configured metric against `describeIndex`/`describeCollection` at connect time and fail fast with a message naming both values? The javadoc note alone will be easy to miss for people who configured the store before this change.

agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/store/MilvusStore.java
```diff
@@ -631,6 +633,23 @@ private List<Document> searchDocumentsInMilvus(
...
 
+    /**
+     * Converts a raw Milvus result to the score contract used by AgentScope.
+     *
+     * <p>L2 results are distances where lower values are more similar, so map them to a
+     * monotonically decreasing score in the [0, 1] range. Other supported metrics already return
+     * scores in the expected direction and are left unchanged.
+     *
+     * @param rawScore the raw score or distance returned by Milvus
+     * @return a higher-is-better score
```
oss-maintainer, line 644: `1 / (1 + distance)` is a reasonable monotone map into `(0, 1]`, and applying the threshold after normalization is the correct order. This is still a breaking change for anyone who already tuned `scoreThreshold` against raw L2 distances, and the resulting scores are not comparable with `COSINE`/`IP` scores. Worth a changelog entry plus a note in the RAG docs that thresholds must be re-tuned for L2 collections.

oss-maintainer (commented):
### Summary
Right problem, right direction: raw Milvus L2 distances were fed straight into the higher-is-better score contract and then compared against `scoreThreshold`, so L2 collections filtered essentially backwards. Passing the configured metric to `SearchReq` and normalizing before thresholding fixes it, with tests for conversion, propagation and ordering. Two compatibility asks: validate configured metric against the existing collection index, and call out the threshold retuning in a changelog.

---

jujn (approved):

jujn:
Assign to @dailingtao (Review)

PR #3077, opened 2026-09-09
fix(state): return unconditional write versions atomically

### LINKED ISSUE #3076: [Bug]: InMemoryAgentStateStore unconditional save can return another concurrent write's version
## Describe the bug

`InMemoryAgentStateStore.saveIfVersion(..., UNVERSIONED)` writes the state through `save()` and then separately reads the current entry's version. Another writer can commit between those operations, so the first caller receives the second writer's version rather than the version of its own write.

This violates the method's documented successful-write return contract and can defeat a subsequent compare-and-swap: a caller whose state was already overwritten can reuse the incorrectly returned newer version and overwrite that newer state without detecting a conflict.

The race is in the unconditional branch at [InMemoryAgentStateStore.java](https://github.com/agentscope-ai/agentscope-java/blob/2b378e158ce1e555cd98c67d114d43cbc316af73/agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java#L93). `setSingleState()` synchronizes the write, but the later version read occurs outside that critical section.

## To reproduce

Run the Java program below against the current `agentscope-core` source or put it on a classpath containing that module. It uses the unmodified store, no mocks or external services. The exact duplicate count depends on scheduling.

```java
import io.agentscope.core.state.*;
import java.util.concurrent.*;
import java.util.concurrent.atomic.AtomicInteger;

public class UnconditionalVersionRepro {
    record Value(String text) implements State {}

    public static void main(String[] args) throws Exception {
        var store = new InMemoryAgentStateStore();
        var versions = ConcurrentHashMap.<Long>newKeySet();
        var duplicates = new AtomicInteger();
        var start = new CountDownLatch(1);
        var pool = Executors.newFixedThreadPool(8);
        try {
            var tasks = new java.util.ArrayList<Future<?>>();
            for (int worker = 0; worker < 8; worker++) {
                int id = worker;
                tasks.add(pool.submit(() -> {
                    start.await();
                    for (int i = 0; i < 1000; i++) {
                        long version = store.saveIfVersion("u", "s", "k",
                                new Value(id + ":" + i), AgentStateStore.UNVERSIONED);
                        if (!versions.add(version)) duplicates.incrementAndGet();
                    }
                    return null;
                }));
            }
            start.countDown();
            for (var task : tasks) task.get(15, TimeUnit.SECONDS);
            System.out.printf("writes=8000 distinct_returned_versions=%d duplicate_versions=%d stored_version=%d%n",
                    versions.size(), duplicates.get(), store.getVersioned("u", "s", "k", Value.class).version());
            if (duplicates.get() != 0) throw new AssertionError("Successful writes returned duplicate versions");
        } finally {
            pool.shutdownNow();
        }
    }
}
```

Observed on the unmodified checkout:

```text
writes=8000 distinct_returned_versions=5437 duplicate_versions=2563 stored_version=8000
java.lang.AssertionError: Successful writes returned duplicate versions
```

The interleaving is also reproducible deterministically in a JUnit test by placing a two-party barrier immediately after `super.save()` in a test-only subclass: both unconditional callers then return version 2. This only controls thread timing; the stress reproduction above uses the ordinary store.

## Expected behavior

Each successful unconditional save returns the version assigned to that particular write. With two writes to an initially absent key, the returned versions must be 1 and 2; the caller with version 1 must then fail a CAS after version 2 has been written.

## Environment

- AgentScope Java: main at `2b378e158ce1e555cd98c67d114d43cbc316af73` (`2.0.3-SNAPSHOT`)
- Java: Eclipse Temurin 17.0.15
- OS: Windows 11
- No model API, Redis, or network service is required for reproduction.

## Proposed fix and validation

Return the assigned version within the same per-session synchronized operation that writes the entry. The existing CAS helper can also handle `UNVERSIONED` by skipping the expected-version comparison and incrementing the current version.

The existing `AgentStateStoreVersioningContractTest` passes all 6 tests on the baseline. A new controlled-interleaving regression fails on the baseline because both writes report version 2. I plan to contribute a focused fix and regression test for this issue.

This report and reproduction were prepared with AI assistance. The commands and results above were executed locally against the referenced checkout.

pengmoubuaixuexi:
## AgentScope-Java Version

`2.0.3-SNAPSHOT`, based on main at `2b378e158ce1e555cd98c67d114d43cbc316af73`.

## Description

Fixes #3076.

Concurrent unconditional `InMemoryAgentStateStore.saveIfVersion` calls can both return the later write's version: the method saves first, then reads the version separately. A caller whose state was overwritten can consequently use that newer version to pass a subsequent CAS incorrectly.

Route unconditional writes through the existing synchronized CAS helper, skipping the comparison for `UNVERSIONED` and returning the version assigned inside the same critical section. Plain `save()` delegates to the same helper with `UNVERSIONED`, keeping version increment and write logic in one synchronized operation. Conditional writes retain their existing behavior.

The Javadoc on both entry points explicitly documents that `saveIfVersion`, including unconditional writes, does not invoke the public `save` method. Subclasses that intercept single-state writes need to override both entry points.

The regression test forces the old write/read interleaving with a barrier after `super.save()` in a test-only subclass. It checks distinct versions, their association with the persisted state, and rejection of the earlier writer's stale version. The test fails on the unmodified implementation and passes with this patch. An additional local stress reproduction using the ordinary store produced 2,563 duplicate versions in 8,000 writes before the fix, and zero after it.

A second committed test uses an ordinary store with no method overrides: four workers each perform 200 unconditional writes, rendezvousing before every round. It checks all 800 returned versions are unique and contiguous, the final state belongs to the final returned version, and a stale-version write is rejected. This contention test complements the targeted old-interleaving regression; it does not rely on the public `save()` hook or guarantee a particular thread schedule.

## Validation

- Baseline existing versioning contract: 6 tests passed.
- Both committed concurrency tests fail against unmodified current main `c5db8f72dbea5c2c70de96885e4167ce5b1b24b0` in an isolated checkout: the targeted regression returns version 2 twice, and the 800-write test detects a duplicate returned version (2 failures, 0 errors).
- After the fix: 43 tests passed, zero failures/errors/skips:

```text
mvn -B -pl agentscope-core -am '-Dtest=*StateStore*Test,ReActAgentPerSessionStateTest' '-Dsurefire.failIfNoSpecifiedTests=false' '-Dspotless.skip=true' test
```

- Spotless apply/check passed for both changed files:

```text
mvn -B -pl agentscope-core '-DspotlessFiles=.*InMemoryAgentStateStore.java,.*InMemoryAgentStateStoreConcurrencyTest.java' spotless:apply spotless:check
```

- `git diff --cached --check` passed.
- Environment: Windows 11, Temurin 17.0.15, Maven 3.9.12. The full multi-module test suite was not run locally.

## Checklist

- [x] Changed code has been formatted with Spotless.
- [x] Relevant state-store and per-session regression tests pass.
- [ ] Full repository `mvn test` (not run locally).
- [x] Public method signatures and version semantics are preserved; the change to public `save()` delegation is documented for subclasses.

AI assistance was used to find, implement, and test this change. Validation results above are from commands executed against the local checkout.

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java
@@ -51,6 +51,13 @@ public class InMemoryAgentStateStore implements AgentStateStore {
     /** users → (sessionId → SessionData) */
     private final Map<String, Map<String, SessionData>> users = new ConcurrentHashMap<>();
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>{@link #saveIfVersion} writes directly through the internal atomic versioning operation,
+     * including unconditional writes; it does not invoke this method. Subclasses that intercept
+     * single-state writes should override both entry points.
+     */
     @Override
     public void save(String userId, String sessionId, String key, State value) {
         SessionData data = lookupOrCreate(userId, sessionId);
@@ -86,6 +93,13 @@ public class InMemoryAgentStateStore implements AgentStateStore {
         return new VersionedState<>(type.cast(state), entry.version());
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>The state write and its returned version are determined in the same critical section. This
+     * method does not delegate to {@link #save(String, String, String, State)}, even when {@code
+     * expectedVersion == UNVERSIONED}.
+     */
     @Override
     public long saveIfVersion(
             String userId, String sessionId, String key, State value, long expectedVersion) {
--- a/agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java
@@ -209,10 +209,8 @@ public class InMemoryAgentStateStore implements AgentStateStore {
         private final Map<String, VersionedEntry> singleStates = new ConcurrentHashMap<>();
         private final Map<String, List<State>> listStates = new ConcurrentHashMap<>();
 
-        synchronized void setSingleState(String key, State value) {
-            VersionedEntry prev = singleStates.get(key);
-            long next = prev == null ? 1L : prev.version() + 1L;
-            singleStates.put(key, new VersionedEntry(value, next));
+        void setSingleState(String key, State value) {
+            casSingleState(key, value, UNVERSIONED);
         }
 
         synchronized long casSingleState(String key, State value, long expectedVersion) {
--- a/agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java
@@ -89,12 +89,6 @@ public class InMemoryAgentStateStore implements AgentStateStore {
     @Override
     public long saveIfVersion(
             String userId, String sessionId, String key, State value, long expectedVersion) {
-        if (expectedVersion == UNVERSIONED) {
-            save(userId, sessionId, key, value);
-            SessionData data = lookup(userId, sessionId);
-            VersionedEntry entry = data != null ? data.getVersionedSingleState(key) : null;
-            return entry != null ? entry.version() : UNVERSIONED;
-        }
         SessionData data = lookupOrCreate(userId, sessionId);
         return data.casSingleState(key, value, expectedVersion);
     }
@@ -224,10 +218,10 @@ public class InMemoryAgentStateStore implements AgentStateStore {
         synchronized long casSingleState(String key, State value, long expectedVersion) {
             VersionedEntry prev = singleStates.get(key);
             long current = prev == null ? 0L : prev.version();
-            if (current != expectedVersion) {
+            if (expectedVersion != UNVERSIONED && current != expectedVersion) {
                 return UNVERSIONED;
             }
-            long next = expectedVersion + 1L;
+            long next = current + 1L;
             singleStates.put(key, new VersionedEntry(value, next));
             return next;
         }
--- a/agentscope-core/src/test/java/io/agentscope/core/state/InMemoryAgentStateStoreConcurrencyTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/state/InMemoryAgentStateStoreConcurrencyTest.java
@@ -17,7 +17,13 @@ package io.agentscope.core.state;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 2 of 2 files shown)


agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java
```diff
@@ -224,10 +218,10 @@ synchronized void setSingleState(String key, State value) {
         synchronized long casSingleState(String key, State value, long expectedVersion) {
             VersionedEntry prev = singleStates.get(key);
             long current = prev == null ? 0L : prev.version();
-            if (current != expectedVersion) {
+            if (expectedVersion != UNVERSIONED && current != expectedVersion) {
                 return UNVERSIONED;
```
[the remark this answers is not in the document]
pengmoubuaixuexi: Addressed in 693cf18. setSingleState now delegates to casSingleState(key, value, UNVERSIONED), so plain saves and unconditional versioned saves share the version increment/write logic. Synchronization stays in casSingleState; the delegating method no longer needs a second synchronized entry. Re-ran the 42 state-store/per-session tests and Spotless apply/check for the changed files: all passed.

dailingtao (approved):
LGTM

agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java
```diff
@@ -89,12 +89,6 @@ public <T extends State> VersionedState<T> getVersioned(
     @Override
     public long saveIfVersion(
             String userId, String sessionId, String key, State value, long expectedVersion) {
-        if (expectedVersion == UNVERSIONED) {
-            save(userId, sessionId, key, value);
-            SessionData data = lookup(userId, sessionId);
-            VersionedEntry entry = data != null ? data.getVersionedSingleState(key) : null;
-            return entry != null ? entry.version() : UNVERSIONED;
-        }
         SessionData data = lookupOrCreate(userId, sessionId);
```
oss-maintainer, line 106: Routing unconditional writes through `casSingleState` fixes the write-then-read race correctly, but it also means `save()` is no longer on the `saveIfVersion(..., UNVERSIONED)` path. `save` is public and overridable (the repo's own `SessionSandboxStateStoreTest.NoDeleteSession` subclasses this store), so a subclass hook that used to fire on an unconditional versioned write now silently does not. Worth either documenting the new delegation on `save()`, or keeping `save()` as the single funnel.

agentscope-core/src/test/java/io/agentscope/core/state/InMemoryAgentStateStoreConcurrencyTest.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+
+class InMemoryAgentStateStoreConcurrencyTest {
+
+    @Test
+    @Timeout(15)
+    void unconditionalWritersReturnTheirOwnVersions() throws Exception {
```
oss-maintainer, line 39: The barrier only trips on the buggy baseline (after the fix `save()` is never called from this path), so on `main` the test asserts distinct versions with effectively no concurrency stress — it guards the exact regression but is weak as a general stress test. The 8k-write stress reproduction you ran locally is the stronger signal; consider keeping a small deterministic loop here (e.g. 4 threads x 200 unconditional writes asserting all returned versions are distinct) so the guard survives refactors of the hook itself.

oss-maintainer (commented):
### Summary
Real race: `saveIfVersion(..., UNVERSIONED)` wrote and then re-read the version outside the critical section, so two concurrent unconditional writers could both return the later version and poison a subsequent CAS. Funneling unconditional writes through `casSingleState` and deriving `next` from `current` is the right fix, and the stale-version rejection is asserted. Two notes: the public `save()` hook is now bypassed on that path, and the regression test's barrier only engages the buggy shape.

---

agentscope-core/src/main/java/io/agentscope/core/state/InMemoryAgentStateStore.java
pengmoubuaixuexi: Addressed in 2f73def by taking the documentation option: the Javadoc on both `save()` and `saveIfVersion()` now explicitly states that versioned writes, including `UNVERSIONED`, go directly through the internal atomic operation and do not invoke the public `save()` override. It also tells subclasses that intercept single-state writes to override both entry points.

This retains the atomic write-and-return path without expanding the change to a new subclass hook API. I also checked `NoDeleteSession`: it overrides `delete()`, so that particular subclass does not depend on the old `save()` delegation. The relevant state-store/per-session suite passes all 43 tests, and Spotless passes for both changed files.

agentscope-core/src/test/java/io/agentscope/core/state/InMemoryAgentStateStoreConcurrencyTest.java
pengmoubuaixuexi: Addressed in 2f73def. I kept the targeted old-interleaving regression and added `concurrentUnconditionalWritesReturnUniqueVersions` using a plain `InMemoryAgentStateStore`, with no overridden hooks. Four workers each perform 200 writes, rendezvousing at a barrier before each round. The test checks all 800 returned versions are unique and contiguous, the final persisted state matches the write that returned the final version, and a stale-version write is rejected without changing the state.

I ran both tests against unmodified current main (`c5db8f72dbea5c2c70de96885e4167ce5b1b24b0`) in an isolated checkout: both failed on duplicate returned versions (2 failures, 0 errors). On the PR branch, all 43 relevant state-store/per-session tests pass, and Spotless passes. The new test adds contention on the real write path; the original test remains the deterministic guard for the specific old interleaving, since a contention loop cannot guarantee a particular thread schedule.

oss-maintainer (approved):
### Summary

Documents and tests the invariant that unconditional writes through `saveIfVersion(..., UNVERSIONED)` bypass the `save(...)` entry point and return a unique, monotonically increasing version per write, determined in the same critical section as the state write itself.

The delta since my previous review is javadoc plus a concurrency test — no production code change. The doc additions are the valuable part: they make explicit that a subclass intercepting single-state writes by overriding `save(...)` will **not** observe unconditional `saveIfVersion(...)` calls, which is a genuine footgun for the `AgentStateStore` SPI and exactly the kind of thing that otherwise gets discovered as a production bug in a custom store implementation.

The test is well constructed for its purpose:
- 4 writers × 200 writes contending on the same `(user, session, key)` through a `CyclicBarrier` releases them per round, so the versions really are produced under contention rather than sequentially;
- asserting `assertNull(writesByVersion.put(version, value))` on every collected write catches a duplicate version without needing to know which writer produced it, and the subsequent `1..800` contiguity check plus `latest.version() == 800` rules out both duplication and gaps;
- `HashMap` is used only on the main thread after `Future#get()`, so there is no unsynchronised access in the test itself;
- the final block pins the other half of the contract — a stale `expectedVersion` returns `UNVERSIONED` and leaves the stored state untouched, verified by re-reading.

### Findings

No defects found in this delta. One suggestion below.

### Suggestions

The barrier is awaited *inside* the per-iteration loop before each write, so it synchronises the start of each write rather than batching writes into rounds. That is fine for the property under test (each write still lands on whatever state the previous writes produced), but it means the assertion `assertEquals(totalWrites, writesByVersion.size())` depends on all 800 writes succeeding, and a `CyclicBarrier` timeout here would surface as a `BrokenBarrierException` inside `Future#get()` rather than as a clear test failure. Under a heavily loaded CI runner with `@Timeout(15)` this is the most plausible source of future flakiness; if it ever trips, the barrier is the first place to look rather than the store.

### Note

`InMemoryAgentStateStore.java:58` and the second javadoc block are documentation-only, so no behavioural risk. CLA is signed and CI is green (`MERGEABLE/BLOCKED` — the `BLOCKED` state is branch protection awaiting a maintainer approval, not a failure).

---

PR #3078, opened 2026-09-09
feat(openai-official): add OpenAI Responses API module

jujn:
### Summary

This PR adds a new model extension module, `agentscope-extensions-model-openai-official`, which integrates OpenAI models through the **official OpenAI Java SDK** (`com.openai:openai-java`) against the **Responses API**. This complements the existing `agentscope-extensions-model-openai` module, which uses a hand-rolled HTTP client against the Chat Completions API.

The Responses API is OpenAI's newer interface with built-in support for reasoning models, encrypted reasoning replay across turns, structured outputs, and server-side conversation state. This module brings first-class support for those capabilities into AgentScope.

### What's included

<details>
<summary>架构图</summary>
openai-official-architecture
</details>
<details>
<summary>流式时序图</summary>
openai-official-sequence
</details>

**Module: `agentscope-extensions-model-openai-official`** (14 production classes, ~2,500 LOC)

The module is structured around a clean separation of concerns:

- `OpenAIResponsesChatModel` — the `ChatModelBase` implementation. Owns the streaming/non-streaming dispatch, retry classification, and builder API. SDK retry is disabled (`maxRetries = 0`); AgentScope owns all retry logic via a module-level `retryOn` predicate that inspects `x-should-retry` headers and SDK exception types.
- `OpenAIOfficialModelProvider` — SPI provider registered via `ModelProvider` service loader. Supports the `openai-official:<model>` model-id convention and resolves `OPENAI_API_KEY` / `OPENAI_BASE_URL` from the environment.
- `ResponsesRequestMapper` — maps AgentScope `Msg` objects, `ToolSchema` definitions, and `GenerateOptions` to SDK `ResponseCreateParams`. Handles reasoning replay (encrypted content from previous turns), structured output (JSON object / JSON schema), tool-call mapping with strict mode, and fail-fast validation for unsupported fields.
- `ResponsesResponseParser` — parses a non-streaming `Response` into a single `ChatResponse`, assembling content blocks in fixed order (ThinkingBlock -> TextBlock -> ToolUseBlock) with full metadata extraction.
- `ResponsesStreamingAssembler` — assembles a `Flux<ResponseStreamEvent>` into incremental `ChatResponse` chunks. Routes text deltas, reasoning summary deltas, function-call argument deltas, and terminal events (completed/incomplete/failed/error) with proper resource cleanup via `doFinally`.
- `ResponsesMultiAgentFormatter` — multi-agent conversation formatter. Groups messages by type (SYSTEM, TOOL_SEQUENCE, AGENT_CONVERSATION, BYPASS), merges agent conversation messages into `<history>`-tagged user messages, and passes through tool sequences and system messages unchanged. Supports a customizable conversation history prompt.
- `OpenAIErrorTranslator` — normalizes SDK exceptions (`OpenAIServiceException`, `OpenAIIoException`, `OpenAIRetryableException`, `OpenAIInvalidDataException`, `TimeoutException`) into `OpenAIOfficialModelException` with HTTP status codes preserved.
- `OpenAISdkClientFactory` — the single production entry point for `OpenAIClient` creation. Sets `maxRetries = 0`, injects builder-level additional headers, and fail-fasts on missing API key.
- `OpenAIOfficialCredential` — JSON-serializable credential type (`openai_official_credential`) for use with AgentScope's credential system.
- `OpenAIOfficialConstants` — shared constants for metadata namespace keys (`openai.*`) and the `additionalBodyParams` whitelist.

**Key capabilities:**

- Streaming and non-streaming Responses API calls
- Tool calling with per-tool and builder-level strict mode
- Native structured output
- Reasoning effort control (`low` / `medium` / `high` / `minimal`) with encrypted reasoning content automatically replayed across turns via `Msg.metadata` — no manual management needed
- Responses-specific parameters via whitelisted `additionalBodyParams`: `reasoning.summary`, `reasoning.context`, `reasoning.mode`, `service_tier`, `prompt_cache_key`, `prompt_cache_options`, `max_tool_calls`, `safety_identifier`
- Builder-level additional headers (constant across all requests)
- Image input (URL and base64) in user messages and tool results
- Credential support and SPI auto-registration
- E2E test provider integration (`OpenAIOfficialResponsesProvider`)

### Supporting changes

Several improvements to core and harness modules were needed to support this integration:

- **`ModelContextWindows`** — added context window sizes for GPT-5.x, GPT-6, and GLM-5.3 models.
- **`ModelUtils`** — timeout exceptions now wrap a `TimeoutException` as cause, so module-level `retryOn` predicates can detect timeouts in the cause chain.
- **Distribution** — `agentscope-all` and `agentscope-bom` updated to include the new module.
- **Docs** — bilingual (English and Chinese) integration documentation added, including table-of-contents and overview updates.

### Testing

The module includes **13 test files** totaling ~5,600 lines, covering:

- Request mapping: history, tools, options, structured output, reasoning validation, rejected fields, prompt cache options, additional body params
- Response parsing: text, reasoning, tool calls, refusal, metadata/usage extraction
- Streaming assembly: text deltas, reasoning deltas, function-call deltas, terminal events, failed/error events, stream cleanup
- Multi-agent formatter: agent conversations, tool sequences, system messages, media handling, edge cases, bypass
- Error translation: all SDK exception types, cause-chain timeout, status-code classification
- Model provider: SPI support detection, context resolution, advanced options, stream defaults
- Cross-turn behavior: reasoning replay, tool-set changes across turns, option merging, streaming vs. non-streaming modes
- Client factory and credential validation

### Not yet supported

The following are intentionally out of scope for this initial PR. They are tracked for follow-up work:

1. **Spring Boot starter** — no dedicated starter is provided.
2. **`store` parameter** — hardcoded to `false` in `ResponsesRequestMapper`. OpenAI's server-side conversation storage is not used; AgentScope manages conversation history client-side. Making this configurable is a straightforward future enhancement.
3. **Multimodal input/output** — only `TextBlock`, `ImageBlock` (URL and base64), and image-type `DataBlock` are supported. `AudioBlock` and `VideoBlock` are not mapped; the request mapper fail-fasts on unsupported block types, and the multi-agent formatter silently skips them. The existing OpenAI and DashScope modules handle audio and video blocks.
4. **`ProxyConfig`** — not supported. This is a real gap for enterprise and restricted-network environments. The official SDK's `OpenAIOkHttpClient.Builder` supports proxy configuration, so wiring is feasible.
5. **MultiModalTool** — no `MultiModalTool` implementation. The OpenAI module provides `OpenAIMultiModalTool` (text-to-image, image-to-text, text-to-audio, audio-to-text) and the DashScope module provides `DashScopeMultiModalTool`. A corresponding tool for this module is optional but would improve feature parity.

### Usage

**Via explicit builder:**

```java
OpenAIResponsesChatModel model = OpenAIResponsesChatModel.builder()
    .apiKey(System.getenv("OPENAI_API_KEY"))
    .modelName("gpt-4o")
    .stream(true)
    .build();
```

**With reasoning and multi-agent formatter:**

```java
GenerateOptions options = GenerateOptions.builder()
    .reasoningEffort("high")
    .additionalBodyParam("reasoning.summary", "auto")
    .build();

OpenAIResponsesChatModel model = OpenAIResponsesChatModel.builder()
    .apiKey(apiKey)
    .modelName("gpt-5.4")
    .generateOptions(options)
    .formatter(new ResponsesMultiAgentFormatter())
    .build();
```

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/model/ModelContextWindows.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/model/ModelContextWindows.java
@@ -68,7 +68,14 @@ public final class ModelContextWindows {
                     Map.entry("o3-mini", 200_000),
                     Map.entry("o3", 200_000),
                     Map.entry("o1-mini", 128_000),
-                    Map.entry("o1", 200_000));
+                    Map.entry("o1", 200_000),
+                    Map.entry("gpt-5.4-mini", 400_000),
+                    Map.entry("gpt-5.4", 1_050_000),
+                    Map.entry("gpt-5.5", 1_050_000),
+                    Map.entry("gpt-5.6-luna", 1_050_000),
+                    Map.entry("gpt-5.6-terra", 1_050_000),
+                    Map.entry("gpt-5.6-sol", 1_050_000),
+                    Map.entry("gpt-6-astra", 1_050_000));
 
     public static final Map<String, Integer> DEEPSEEK =
             Map.ofEntries(
@@ -77,6 +84,7 @@ public final class ModelContextWindows {
 
     public static final Map<String, Integer> GLM =
             Map.ofEntries(
+                    Map.entry("glm-5.3", 1_000_000),
                     Map.entry("glm-5.2", 1_000_000),
                     Map.entry("glm-5.1", 200_000),
                     Map.entry("glm-5-turbo", 200_000),
--- a/agentscope-distribution/agentscope-all/pom.xml
+++ b/agentscope-distribution/agentscope-all/pom.xml
@@ -108,12 +108,12 @@
             <optional>true</optional>
         </dependency>
 
-<!--        <dependency>-->
-<!--            <groupId>io.agentscope</groupId>-->
-<!--            <artifactId>agentscope-extensions-model-openai-official</artifactId>-->
-<!--            <scope>compile</scope>-->
-<!--            <optional>true</optional>-->
-<!--        </dependency>-->
+        <dependency>
+            <groupId>io.agentscope</groupId>
+            <artifactId>agentscope-extensions-model-openai-official</artifactId>
+            <scope>compile</scope>
+            <optional>true</optional>
+        </dependency>
 
         <dependency>
             <groupId>io.agentscope</groupId>
--- a/agentscope-distribution/agentscope-all/pom.xml
+++ b/agentscope-distribution/agentscope-all/pom.xml
@@ -470,6 +470,9 @@
                                     <include>io.agentscope:agentscope-extensions-*</include>
                                 </includes>
                             </artifactSet>
+                            <transformers>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                            </transformers>
                         </configuration>
                     </execution>
                 </executions>
--- a/agentscope-distribution/agentscope-bom/pom.xml
+++ b/agentscope-distribution/agentscope-bom/pom.xml
@@ -355,11 +355,11 @@
                 <artifactId>agentscope-extensions-model-openai</artifactId>
                 <version>${project.version}</version>
             </dependency>
-<!--            <dependency>-->
-<!--                <groupId>io.agentscope</groupId>-->
-<!--                <artifactId>agentscope-extensions-model-openai-official</artifactId>-->
-<!--                <version>${project.version}</version>-->
-<!--            </dependency>-->
+            <dependency>
+                <groupId>io.agentscope</groupId>
+                <artifactId>agentscope-extensions-model-openai-official</artifactId>
+                <version>${project.version}</version>
+            </dependency>
             <dependency>
                 <groupId>io.agentscope</groupId>
                 <artifactId>agentscope-extensions-model-gemini</artifactId>
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-e2e-tests/src/test/java/io/agentscope/core/e2e/providers/OpenAIOfficialResponsesProvider.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-e2e-tests/src/test/java/io/agentscope/core/e2e/providers/OpenAIOfficialResponsesProvider.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.core.e2e.providers;
+
+import io.agentscope.core.ReActAgent;
+import io.agentscope.core.tool.Toolkit;
+import io.agentscope.extensions.model.openaiofficial.OpenAIResponsesChatModel;
+import io.agentscope.extensions.model.openaiofficial.ResponsesMultiAgentFormatter;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Provider for OpenAI models via the official OpenAI Java SDK (Responses API).
+ */
+@ModelCapabilities({ModelCapability.BASIC, ModelCapability.TOOL_CALLING})
+public class OpenAIOfficialResponsesProvider extends BaseModelProvider {
+
+    private static final String API_KEY_ENV = "OPENAI_API_KEY";
+    private static final String BASE_URL_ENV = "OPENAI_BASE_URL";
+
+    public OpenAIOfficialResponsesProvider(String modelName, boolean multiAgentFormatter) {
+        super(API_KEY_ENV, modelName, multiAgentFormatter);
+    }
+
+    @Override
+    protected ReActAgent.Builder doCreateAgentBuilder(String name, Toolkit toolkit, String apiKey) {
+        String baseUrl = System.getenv(BASE_URL_ENV);
+
+        OpenAIResponsesChatModel.Builder builder =
+                OpenAIResponsesChatModel.builder().apiKey(apiKey).modelName(getModelName());
+
+        if (baseUrl != null && !baseUrl.isEmpty()) {
+            builder.baseUrl(baseUrl);
+        }
+
+        if (isMultiAgentFormatter()) {
+            builder.formatter(new ResponsesMultiAgentFormatter());
+        }
+
+        return ReActAgent.builder().name(name).model(builder.build()).toolkit(toolkit);
+    }
+
+    @Override
+    public String getProviderName() {
+        return "OpenAI-Official";
+    }
+
+    @Override
+    public Set<ModelCapability> getCapabilities() {
+        Set<ModelCapability> caps = new HashSet<>(super.getCapabilities());
+        if (isMultiAgentFormatter()) {
+            caps.add(ModelCapability.MULTI_AGENT_FORMATTER);
+        }
+        return caps;
+    }
+
+    // ==========================================================================
+    // Provider Instances
+    // ==========================================================================
+
+    /** GPT-5.4 via OpenAI Official SDK (Responses API). */
+    @ModelCapabilities({
+        ModelCapability.BASIC,
+        ModelCapability.TOOL_CALLING,
+        ModelCapability.THINKING
+    })
+    public static class Gpt54 extends OpenAIOfficialResponsesProvider {
+        public Gpt54() {
+            super("gpt-5.4", false);
+        }
+
+        @Override
+        public String getProviderName() {
+            return "OpenAI-Official";
+        }
+    }
+
+    /** GPT-5.4 with Multi-Agent Formatter. */
+    @ModelCapabilities({
+        ModelCapability.BASIC,
+        ModelCapability.TOOL_CALLING,
+        ModelCapability.THINKING,
+        ModelCapability.MULTI_AGENT_FORMATTER
+    })
+    public static class Gpt54MultiAgent extends OpenAIOfficialResponsesProvider {
+        public Gpt54MultiAgent() {
+            super("gpt-5.4", true);
+        }
+
+        @Override
+        public String getProviderName() {
+            return "OpenAI-Official (Multi-Agent)";
+        }
+    }
+}
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/OpenAIOfficialModelProvider.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/OpenAIOfficialModelProvider.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.extensions.model.openaiofficial;
+
+import static io.agentscope.core.model.ModelProviderSupport.booleanOption;
+import static io.agentscope.core.model.ModelProviderSupport.findAssignableComponent;
+import static io.agentscope.core.model.ModelProviderSupport.firstNonBlank;
+import static io.agentscope.core.model.ModelProviderSupport.intOption;
+
+import io.agentscope.core.model.GenerateOptions;
+import io.agentscope.core.model.Model;
+import io.agentscope.core.model.ModelCreationContext;
+import io.agentscope.core.model.spi.ModelProvider;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+public final class OpenAIOfficialModelProvider implements ModelProvider {
+
+    private static final String PREFIX = "openai-official:";
+    private static final Pattern MODEL_ID = Pattern.compile("openai-official:.+");
+
+    @Override
+    public String providerId() {
+        return OpenAIOfficialConstants.PROVIDER_ID;
+    }
+
+    @Override
+    public boolean supports(String modelId) {
+        return modelId != null && MODEL_ID.matcher(modelId).matches();
+    }
+
+    @Override
+    public Model create(String modelId) {
+        return create(modelId, ModelCreationContext.empty());
+    }
+
+    @Override
+    public Model create(String modelId, ModelCreationContext context) {
+        if (!supports(modelId)) {
+            throw new IllegalArgumentException("Unsupported OpenAI model id: " + modelId);
+        }
+
+        String apiKey = firstNonBlank(context.getApiKey(), System.getenv("OPENAI_API_KEY"));
+        if (apiKey == null) {
+            throw new IllegalStateException(
+                    "Environment variable OPENAI_API_KEY is required to auto-create model: "
+                            + modelId);
+        }
+        String modelName = modelId.substring(PREFIX.length());
+        String baseUrl = firstNonBlank(context.getBaseUrl(), System.getenv("OPENAI_BASE_URL"));
+        boolean stream = context.getStream() != null ? context.getStream() : true;
+
+        OpenAIResponsesChatModel.Builder builder =
+                OpenAIResponsesChatModel.builder()
+                        .apiKey(apiKey)
+                        .baseUrl(baseUrl)
+                        .modelName(modelName)
+                        .stream(stream);
+
+        applyAdvancedOptions(builder, context);
+        return builder.build();
+    }
+
+    @SuppressWarnings("unchecked")
+    private static void applyAdvancedOptions(
+            OpenAIResponsesChatModel.Builder builder, ModelCreationContext context) {
+        GenerateOptions generateOptions = context.component(GenerateOptions.class);
+        if (generateOptions != null) {
+            builder.generateOptions(generateOptions);
+        }
+
+        Boolean strictTools = booleanOption(context, "strictTools");
+        if (strictTools != null) {
+            builder.strictTools(strictTools);
+        }
+
+        Boolean strictJsonSchema = booleanOption(context, "strictJsonSchema");
+        if (strictJsonSchema != null) {
+            builder.strictJsonSchema(strictJsonSchema);
+        }
+
+        Integer contextWindowSize = intOption(context, "contextWindowSize");
+        if (contextWindowSize != null) {
+            builder.contextWindowSize(contextWindowSize);
+        }
+
+        ResponsesMultiAgentFormatter formatter =
+                findAssignableComponent(context, ResponsesMultiAgentFormatter.class);
+        if (formatter != null) {
+            builder.formatter(formatter);
+        }
+
+        Object raw = context.option("additionalHeaders");
+        if (raw instanceof Map<?, ?> map) {
+            builder.additionalHeaders((Map<String, String>) map);
+        }
+    }
+}
--- a/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/OpenAIResponsesChatModel.java
+++ b/agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/OpenAIResponsesChatModel.java
@@ -1,0 +1,383 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.extensions.model.openaiofficial;
+
+import com.openai.client.OpenAIClient;
+import com.openai.core.http.StreamResponse;
+import com.openai.errors.OpenAIIoException;
+import com.openai.errors.OpenAIRetryableException;
+import com.openai.errors.OpenAIServiceException;
+import com.openai.models.responses.Response;
+import com.openai.models.responses.ResponseCreateParams;
+import com.openai.models.responses.ResponseInputItem;
+import com.openai.models.responses.ResponseStreamEvent;
+import io.agentscope.core.message.Msg;
+import io.agentscope.core.model.ChatModelBase;
+import io.agentscope.core.model.ChatResponse;
+import io.agentscope.core.model.ExecutionConfig;
+import io.agentscope.core.model.GenerateOptions;
+import io.agentscope.core.model.ModelContextWindows;
+import io.agentscope.core.model.ModelHttpException;
+import io.agentscope.core.model.ModelProviderSupport;
+import io.agentscope.core.model.ModelUtils;
+import io.agentscope.core.model.ToolSchema;
+import java.io.IOException;
+import java.time.Instant;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import reactor.core.publisher.Flux;
+import reactor.core.scheduler.Schedulers;
+
+public class OpenAIResponsesChatModel extends ChatModelBase {
+
+    private static final Logger log = LoggerFactory.getLogger(OpenAIResponsesChatModel.class);
+
+    private final OpenAIClient client;
+    private final GenerateOptions configuredOptions;
+    private final String apiKey;
+    private final String baseUrl;
+    private final Boolean strictTools;
+    private final Boolean strictJsonSchema;
+    private final ResponsesMultiAgentFormatter formatter;
+
+    OpenAIResponsesChatModel(
+            OpenAIClient client,
+            GenerateOptions configuredOptions,
+            String apiKey,
+            String baseUrl,
+            Boolean strictTools,
+            Boolean strictJsonSchema,
+            ResponsesMultiAgentFormatter formatter) {
+        this.client = client;
+        this.configuredOptions = configuredOptions;
+        this.apiKey = apiKey;
+        this.baseUrl = baseUrl;
+        this.strictTools = strictTools;
+        this.strictJsonSchema = strictJsonSchema;
+        this.formatter = formatter;
+    }
+
+    void applyNativeStructuredOutputDefaults() {
+        setNativeStructuredOutput(true);
+        setNativeStructuredOutputWithTools(true);
+    }
+
+    @Override
+    public String getModelName() {
+        return configuredOptions != null ? configuredOptions.getModelName() : null;
+    }
+
+    @Override
+    protected Flux<ChatResponse> doStream(
+            List<Msg> messages, List<ToolSchema> tools, GenerateOptions options) {
+        return ModelUtils.applyTimeoutAndRetry(
+                doStream0(messages, tools, options),
+                options,
+                configuredOptions,
+                configuredOptions.getModelName(),
+                OpenAIOfficialConstants.PROVIDER_ID);
+    }
+
+    private Flux<ChatResponse> doStream0(
+            List<Msg> messages, List<ToolSchema> tools, GenerateOptions options) {
+
+        GenerateOptions effectiveOptions = GenerateOptions.mergeOptions(options, configuredOptions);
+
+        validateConnectionFields(effectiveOptions);
+
+        boolean stream = effectiveOptions.getStream() != null ? effectiveOptions.getStream() : true;
+
+        String modelName = effectiveOptions.getModelName();
+        log.debug("OpenAI API call: model={}", modelName);
+
+        Function<List<Msg>, List<ResponseInputItem>> historyMapper =
+                formatter != null ? formatter::formatHistory : ResponsesRequestMapper::mapHistory;
+
+        ResponseCreateParams params =
+                ResponsesRequestMapper.map(
+                        messages,
+                        tools,
+                        effectiveOptions,
+                        strictTools,
+                        strictJsonSchema,
+                        historyMapper);
+
+        if (stream) {
+            return buildStreamingFlux(params, modelName);
+        } else {
+            return buildNonStreamingFlux(params, modelName);
+        }
+    }
+
+    private Flux<ChatResponse> buildNonStreamingFlux(
+            ResponseCreateParams params, String modelName) {
+        return Flux.defer(
+                        () -> {
+                            Instant start = Instant.now();
+                            try {
+                                Response response = client.responses().create(params);
+                                return Flux.just(
+                                        ResponsesResponseParser.parse(response, modelName, start));
+                            } catch (RuntimeException e) {
+                                return Flux.error(OpenAIErrorTranslator.translate(e, modelName));
+                            }
+                        })
+                .subscribeOn(Schedulers.boundedElastic());
+    }
+
+    private Flux<ChatResponse> buildStreamingFlux(ResponseCreateParams params, String modelName) {
+        return Flux.defer(
+                        () -> {
+                            Instant start = Instant.now();
+                            try {
+                                StreamResponse<ResponseStreamEvent> streamResponse =
+                                        client.responses().createStreaming(params);
+                                return ResponsesStreamingAssembler.assemble(
+                                        streamResponse, modelName, start);
+                            } catch (RuntimeException e) {
+                                return Flux.error(OpenAIErrorTranslator.translate(e, modelName));
+                            }
+                        })
+                .subscribeOn(Schedulers.boundedElastic());
+    }
+
+    private void validateConnectionFields(GenerateOptions effectiveOptions) {
+        String effectiveApiKey = ModelProviderSupport.trimToNull(effectiveOptions.getApiKey());
+        if (effectiveApiKey != null && !Objects.equals(effectiveApiKey, apiKey)) {
+            throw new OpenAIOfficialModelException(
+                    "apiKey override per request is not supported by the openai-official"
+                            + " provider. The apiKey must match the one configured at model"
+                            + " construction.",
+                    effectiveOptions.getModelName());
+        }
+
+        String effectiveBaseUrl = ModelProviderSupport.trimToNull(effectiveOptions.getBaseUrl());
+        if (effectiveBaseUrl != null && !Objects.equals(effectiveBaseUrl, baseUrl)) {
+            throw new OpenAIOfficialModelException(
+                    "baseUrl override per request is not supported by the openai-official"
+                            + " provider. The baseUrl must match the one configured at model"
+                            + " construction.",
+                    effectiveOptions.getModelName());
+        }
+    }
+
+    // Module retryOn predicate
+
+    static Predicate<Throwable> moduleRetryOn() {
+        return OpenAIResponsesChatModel::isModuleRetryable;
+    }
+
+    private static boolean isModuleRetryable(Throwable error) {
+        Boolean headerShouldRetry = checkXShouldRetry(error);
+        if (headerShouldRetry != null) {
+            return headerShouldRetry;
+        }
+
+        Throwable current = error;
+        while (current != null) {
+            if (current instanceof ModelHttpException mhe) {
+                Integer statusCode = mhe.getStatusCode();
+                if (statusCode != null) {
+                    int code = statusCode;
+                    if (code == 408 || code == 409 || code == 429 || (code >= 500 && code < 600)) {
+                        return true;
+                    }
+                }
+            }
+
+            if (current instanceof OpenAIIoException
+                    || current instanceof OpenAIRetryableException) {
+                return true;
+            }
+
+            if (current instanceof IOException) {
+                return true;
+            }
+
+            if (current instanceof TimeoutException) {
+                return true;
+            }
+
+            current = current.getCause();
+        }
+        return false;
+    }
+
+    private static Boolean checkXShouldRetry(Throwable current) {
+        while (current != null) {
+            if (current instanceof OpenAIServiceException svc) {
+                List<String> values = svc.headers().values("x-should-retry");
+                if (!values.isEmpty()) {
+                    String value = values.get(0);
+                    if ("true".equalsIgnoreCase(value)) {
+                        return true;
+                    }
+                    if ("false".equalsIgnoreCase(value)) {
+                        return false;
+                    }
+                }
+            }
+            current = current.getCause();
+        }
+        return null;
+    }
+
+    /** Package-private test seam for verifying Builder injection logic. */
+    GenerateOptions getConfiguredOptions() {
+        return configuredOptions;
+    }
+
+    // Builder
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+        private String apiKey;
+        private String baseUrl;
+        private String modelName;
+        private boolean stream = true;
+        private GenerateOptions generateOptions;
+        private int contextWindowSize = -1;
+        private Boolean strictTools;
+        private Boolean strictJsonSchema;
+        private ResponsesMultiAgentFormatter formatter;
+        private Map<String, String> additionalHeaders;
+
+        public Builder apiKey(String apiKey) {
+            this.apiKey = apiKey;
+            return this;
+        }
+
+        public Builder baseUrl(String baseUrl) {
+            this.baseUrl = baseUrl;
+            return this;
+        }
+
+        public Builder modelName(String modelName) {
+            this.modelName = modelName;
+            return this;
+        }
+
+        public Builder stream(boolean stream) {
+            this.stream = stream;
+            return this;
+        }
+
+        public Builder generateOptions(GenerateOptions generateOptions) {
+            this.generateOptions = generateOptions;
+            return this;
+        }
+
+        public Builder contextWindowSize(int contextWindowSize) {
+            this.contextWindowSize = contextWindowSize;
+            return this;
+        }
+
+        public Builder strictTools(Boolean strictTools) {
+            this.strictTools = strictTools;
+            return this;
+        }
+
+        public Builder strictJsonSchema(Boolean strictJsonSchema) {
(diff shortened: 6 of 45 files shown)


muranchenhui (approved):

muranchenhui:
LGTM

agentscope-distribution/agentscope-all/pom.xml
```diff
@@ -108,12 +108,12 @@
             <optional>true</optional>
         </dependency>
 
-<!--        <dependency>-->
```
zouyx, line 111: 为什么不是删除？

agentscope-distribution/agentscope-bom/pom.xml
```diff
@@ -348,11 +348,11 @@
                 <artifactId>agentscope-extensions-model-openai</artifactId>
                 <version>${project.version}</version>
             </dependency>
-<!--            <dependency>-->
```
zouyx, line 358: 为什么不是删除

agentscope-distribution/agentscope-all/pom.xml
```diff
@@ -108,12 +108,12 @@
...
-<!--            <groupId>io.agentscope</groupId>-->
-<!--            <artifactId>agentscope-extensions-model-openai-official</artifactId>-->
-<!--            <scope>compile</scope>-->
-<!--            <optional>true</optional>-->
-<!--        </dependency>-->
+        <dependency>
+            <groupId>io.agentscope</groupId>
+            <artifactId>agentscope-extensions-model-openai-official</artifactId>
+            <scope>compile</scope>
```
zouyx, line 114: 这种直接打包进去吗？还是应该由用户提供？

jujn: > 这种直接打包进去吗？还是应该由用户提供？

遵循了既有模式，现有的其它模型厂商模块也都在里面

Aias00 (approved):
已审。重点检查了新增 openai-official Responses 模块的请求映射、非流式解析、流式事件拼装、usage/metadata/finishReason、工具调用回放、错误转换、provider SPI、BOM/distribution 接入和相关测试覆盖。\n\n本地验证：`mvn -pl agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official -am test -DskipITs` 通过，274 tests, 0 failures/errors, 2 skipped。\n\n未发现需要阻塞合并的问题。

agentscope-distribution/agentscope-all/pom.xml
```diff
@@ -108,12 +108,12 @@
...
 
-<!--        <dependency>-->
-<!--            <groupId>io.agentscope</groupId>-->
-<!--            <artifactId>agentscope-extensions-model-openai-official</artifactId>-->
-<!--            <scope>compile</scope>-->
-<!--            <optional>true</optional>-->
-<!--        </dependency>-->
+        <dependency>
+            <groupId>io.agentscope</groupId>
+            <artifactId>agentscope-extensions-model-openai-official</artifactId>
```
Aias00, line 113: [HIGH] Adding this module to `agentscope-all` currently packages the classes but does not make the provider discoverable. The shade build warns that multiple model modules define `META-INF/services/io.agentscope.core.model.spi.ModelProvider`, and the generated `agentscope-2.0.3-SNAPSHOT.jar` keeps only the existing OpenAI entries: `unzip -p agentscope-distribution/agentscope-all/target/agentscope-2.0.3-SNAPSHOT.jar META-INF/services/io.agentscope.core.model.spi.ModelProvider` does not include `io.agentscope.extensions.model.openaiofficial.OpenAIOfficialModelProvider`, even though the class is present. Users depending on the advertised all-in-one `agentscope` artifact will therefore fail to resolve `openai-official:<model>` through `ModelRegistry`. Please configure the shade plugin to merge service descriptors, for example with `org.apache.maven.plugins.shade.resource.ServicesResourceTransformer`, and add a packaging assertion or smoke test that the shaded jar service file contains this provider.

Aias00 (changes requested):
Thanks for the large integration. I found one release-blocking issue in the all-in-one distribution: the new module is present in the shaded jar, but its ModelProvider service entry is not merged, so ModelRegistry cannot discover openai-official models when users depend on the `agentscope` artifact.

Verification run locally on PR head `0bcf01970ce985d0f973c81fefc9b30e241ed9fe`:
- `mvn -pl agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official -am test` passed: 274 tests, 0 failures, 2 skipped
- `mvn -pl agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-e2e-tests -am test-compile -DskipTests` passed
- `bash .github/scripts/check-shade-and-bom-sync.sh` passed
- `mvn -pl agentscope-distribution/agentscope-all -am package -DskipTests` passed, but emitted the overlapping `META-INF/services/io.agentscope.core.model.spi.ModelProvider` warning; inspecting the resulting jar confirmed the new provider is missing from the service descriptor.

Aias00 (changes requested):
Requesting changes based on two concrete correctness issues.

### [HIGH] Multi-agent formatter drops encrypted reasoning replay

**File**: `agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/ResponsesMultiAgentFormatter.java` (line 175)

When `OpenAIResponsesChatModel` is configured with `ResponsesMultiAgentFormatter`, all history mapping goes through `formatter::formatHistory`. For an ordinary assistant turn that contains `ThinkingBlock` plus `openai.reasoning.encrypted_content` metadata, `determineGroupType` classifies it as `AGENT_CONVERSATION`; `mergeAgentConversation` then rewrites it into a user `<history>` message and `processMessage` only appends the visible thinking/text content. It never calls `ResponsesRequestMapper.mapAssistantMessage`, so no `ResponseReasoningItem` with `encrypted_content` is sent on the next request.

This breaks the PR's documented reasoning replay behavior exactly for the sample configuration that combines reasoning with `ResponsesMultiAgentFormatter`. I verified with a minimal probe: the formatted history for `user -> assistant(thinking + encrypted metadata) -> user` produced `items=1 reasoning=0 easy=1`.

**Why tests miss it**: `CrossTurnTest` verifies encrypted replay only with the default mapper, and `ResponsesMultiAgentFormatterTest` covers thinking text but never asserts that encrypted reasoning metadata is preserved.

**Suggested fix**: make formatter preserve assistant messages with `openai.reasoning.encrypted_content` as Responses reasoning replay items, or classify them into a passthrough path that delegates to `ResponsesRequestMapper.mapAssistantMessage`. Add a regression test that builds `OpenAIResponsesChatModel` with `new ResponsesMultiAgentFormatter()` and asserts the second-turn `ResponseCreateParams.input` contains a reasoning item with the encrypted content.

### [MEDIUM] Per-schema strict mode is ignored for JSON schema response formats

**File**: `agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/ResponsesRequestMapper.java` (line 185)

`JsonSchema` already exposes `getStrict()`, but the Responses mapper only sets `schemaBuilder.strict(...)` from the builder-level `strictJsonSchema`. A caller using `ResponseFormat.jsonSchema(JsonSchema.builder().strict(true).build())` silently gets a request without `strict: true` unless they also set the model builder flag.

**Why tests miss it**: the tests only cover builder-level `strictJsonSchema`; there is no test where `JsonSchema.strict(true)` is set on the response format itself.

**Suggested fix**: resolve strict as `schema.getStrict() != null ? schema.getStrict() : strictJsonSchema`, then set it when non-null. Add a test that schema-level strict is propagated and that builder-level remains a fallback.

Validation run:
- `mvn -pl agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official -am test -DskipITs` passed: 274 tests, 0 failures, 2 skipped.
- Java LSP diagnostics were unavailable in this environment; the exposed diagnostic tool runs `npx tsc`, so Maven compile/tests were used as the Java diagnostic substitute.
- `ast-grep` was not installed; I used `rg` for hardcoded secret/debug/empty-catch/problem-pattern scanning.

Aias00 (approved):
已复核 PR head 0bcf01970ce985d0f973c81fefc9b30e241ed9fe。重点检查了 openai-official Responses 模块的请求映射、ReasoningContext metadata 持久化与后续 encrypted reasoning replay、流式 terminal metadata/usage 聚合、工具调用回放、provider SPI、BOM/distribution 接入和 retry 配置组合；未发现需要阻塞合并的问题。\n\nGitHub checks 当前通过：Java build ubuntu/windows、Check License、Check Module Sync、Codecov、CLA、Mintlify validate。\n\n结论：APPROVE。

Aias00 (changes requested):
Requesting changes after re-checking PR head 0bcf01970ce985d0f973c81fefc9b30e241ed9fe. I found release-blocking correctness issues that should be fixed before merge.

### [HIGH] Multi-agent formatter drops encrypted reasoning replay

`OpenAIResponsesChatModel` sends all history through `formatter::formatHistory` when a `ResponsesMultiAgentFormatter` is configured (`OpenAIResponsesChatModel.java:112-122`). In that path, ordinary assistant messages are classified as `AGENT_CONVERSATION` unless they contain tool blocks (`ResponsesMultiAgentFormatter.java:172-181`), and `mergeAgentConversation` rewrites them into a user `<history>` message (`ResponsesMultiAgentFormatter.java:209-237`). `processMessage` then serializes `ThinkingBlock` as visible text (`ResponsesMultiAgentFormatter.java:305-312`) and never delegates to `ResponsesRequestMapper.mapAssistantMessage`, which is the only code path that emits the Responses reasoning replay item from `openai.reasoning.encrypted_content` metadata (`ResponsesRequestMapper.java:433-445`, `ResponsesRequestMapper.java:477-493`).

This breaks stateless reasoning replay exactly for the documented reasoning + multi-agent formatter usage. Please preserve assistant messages with encrypted reasoning metadata as Responses reasoning input items, or route them through a passthrough/delegation path, and add a regression test that configures `OpenAIResponsesChatModel` with `new ResponsesMultiAgentFormatter()` and asserts second-turn input contains the encrypted `ResponseReasoningItem`.

### [MEDIUM] Schema-level strict mode is ignored for JSON schema structured output

`JsonSchema` exposes `strict` (`JsonSchema.java:91-96`), but `ResponsesRequestMapper` only writes `schemaBuilder.strict(...)` from the model-level `strictJsonSchema` flag (`ResponsesRequestMapper.java:178-187`). A caller setting `ResponseFormat.json_schema` with `JsonSchema.strict(true)` silently gets a request without `strict: true` unless they also configure the model builder flag.

Please resolve strict as schema-level first, then builder-level fallback, e.g. `schema.getStrict() != null ? schema.getStrict() : strictJsonSchema`, and add tests for schema-level override plus builder-level fallback.

### [HIGH] all-in-one shaded artifact can lose ModelProvider service registration

The new module is added to `agentscope-all` (`agentscope-all/pom.xml:111-116`) and has its own SPI file (`agentscope-extensions-model-openai-official/src/main/resources/META-INF/services/io.agentscope.core.model.spi.ModelProvider:1`). But the shade config only includes artifacts and has no `ServicesResourceTransformer` or equivalent service merge (`agentscope-all/pom.xml:449-466`). There are multiple model-provider service descriptors across extension modules, so the shaded jar can keep only one descriptor and drop `OpenAIOfficialModelProvider`, making `ModelRegistry` discovery fail for users depending on the all-in-one `agentscope` artifact.

Please add a service resource transformer to the shade plugin and a packaging regression check that the shaded jar's `META-INF/services/io.agentscope.core.model.spi.ModelProvider` contains `io.agentscope.extensions.model.openaiofficial.OpenAIOfficialModelProvider`.

CI is green, but these are behavioral/package correctness issues not covered by the current checks.

Aias00 (changes requested):
One additional blocking issue affects the default formatter path as well.

### [HIGH] The production ReActAgent path discards encrypted reasoning metadata before the next turn

`ResponsesResponseParser` stores `openai.reasoning.encrypted_content` only in the terminal `ChatResponse.metadata` (lines 105-106), and the streaming path does the same in `ResponsesStreamingAssembler` (lines 261-263). However, the production consumer, `ReasoningContext.processChunk`, never reads `chunk.getMetadata()`; `buildFinalMessage()` creates a fresh metadata map containing only `MessageMetadataKeys.CHAT_USAGE`. The resulting assistant `Msg` therefore cannot contain the key that `ResponsesRequestMapper.mapAssistantMessage` looks up at lines 434-444.

As a result, encrypted reasoning replay is unreachable through a normal `ReActAgent` call even without `ResponsesMultiAgentFormatter`. This is particularly important for stateless reasoning/tool-call loops, where the reasoning item from the prior response must be sent back with the function-call output. The current cross-turn tests miss this because they manually construct an `AssistantMessage` with `openai.reasoning.encrypted_content` already present instead of exercising `ChatResponse -> ReasoningContext -> Msg -> next request`.

Please preserve the terminal response metadata in the final assistant message (or place the replay data on a content block and ensure its accumulator preserves it), and add a ReActAgent-level two-iteration regression test that captures the second `ResponseCreateParams` and asserts it contains the prior encrypted reasoning item.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-e2e-tests/src/test/java/io/agentscope/core/e2e/providers/OpenAIOfficialResponsesProvider.java
```diff
@@ -0,0 +1,108 @@
...
+    private static final String API_KEY_ENV = "OPENAI_API_KEY";
+    private static final String BASE_URL_ENV = "OPENAI_BASE_URL";
+
+    public OpenAIOfficialResponsesProvider(String modelName, boolean multiAgentFormatter) {
+        super(API_KEY_ENV, modelName, multiAgentFormatter);
+    }
+
+    @Override
+    protected ReActAgent.Builder doCreateAgentBuilder(String name, Toolkit toolkit, String apiKey) {
+        String baseUrl = System.getenv(BASE_URL_ENV);
```
zouyx, line 40: 这个只支持通过系统环境变量获取吗？
为什么不是从 spring 生命周期中的配置获取？

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-e2e-tests/src/test/java/io/agentscope/core/e2e/providers/OpenAIOfficialResponsesProvider.java
```diff
@@ -0,0 +1,108 @@
...
+        if (isMultiAgentFormatter()) {
+            builder.formatter(new ResponsesMultiAgentFormatter());
+        }
+
+        return ReActAgent.builder().name(name).model(builder.build()).toolkit(toolkit);
+    }
+
+    @Override
+    public String getProviderName() {
+        return "OpenAI-Official";
```
zouyx, line 58: 为什么这个不是静态常量？

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-e2e-tests/src/test/java/io/agentscope/core/e2e/providers/OpenAIOfficialResponsesProvider.java
```diff
@@ -0,0 +1,108 @@
...
+        ModelCapability.MULTI_AGENT_FORMATTER
+    })
+    public static class Gpt54MultiAgent extends OpenAIOfficialResponsesProvider {
+        public Gpt54MultiAgent() {
+            super("gpt-5.4", true);
+        }
+
+        @Override
+        public String getProviderName() {
+            return "OpenAI-Official (Multi-Agent)";
```
zouyx, line 105: 为啥这不是静态常量呢？

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/OpenAIResponsesChatModel.java
```diff
@@ -0,0 +1,383 @@
...
+                return true;
+            }
+
+            if (current instanceof IOException) {
+                return true;
+            }
+
+            if (current instanceof TimeoutException) {
+                return true;
+            }
```
zouyx, line 218: 这几个不能整合吗？

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/OpenAISdkClientFactory.java
```diff
@@ -0,0 +1,92 @@
...
+                            + " builder.apiKey(...) or the OPENAI_API_KEY environment variable.",
+                    null,
+                    null);
+        }
+
+        try {
+            OpenAIOkHttpClient.Builder builder =
+                    OpenAIOkHttpClient.builder().apiKey(apiKey).maxRetries(0);
+
+            if (baseUrl != null && !baseUrl.isBlank()) {
```
zouyx, line 74: baseUrl.isNotBlank() 会不会好点？

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/OpenAIResponsesChatModel.java
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    private Flux<ChatResponse> buildStreamingFlux(ResponseCreateParams params, String modelName) {
+        return Flux.defer(
+                        () -> {
+                            Instant start = Instant.now();
+                            try {
+                                StreamResponse<ResponseStreamEvent> streamResponse =
+                                        client.responses().createStreaming(params);
+                                return ResponsesStreamingAssembler.assemble(
```
oss-maintainer, line 154: **[Warning]** `StreamResponse` leak on the assembly failure path.

`ResponsesStreamingAssembler.assemble(...)` evaluates `streamResponse.stream()` eagerly as an argument to `Flux.fromStream(...)`, and the `doFinally(signal -> closeQuietly(streamResponse))` guard is only attached to the flux *after* that call returns. If `stream()` (or anything else in `assemble` before `doFinally`) throws, the exception propagates out of `assemble`, this `catch (RuntimeException e)` converts it to `Flux.error(...)`, and the SDK stream is never closed — leaking the underlying HTTP connection for the remainder of the client's lifetime.

The window is narrow but it is exactly the path that fires when the upstream is already misbehaving, which is when connection hygiene matters most. Closing in the catch keeps the invariant local and obvious:

```java
StreamResponse<ResponseStreamEvent> streamResponse = null;
try {
    streamResponse = client.responses().createStreaming(params);
    Flux<ChatResponse> assembled =
            ResponsesStreamingAssembler.assemble(streamResponse, modelName, start);
    streamResponse = null;
    return assembled;
} catch (RuntimeException e) {
    if (streamResponse != null) {
        closeQuietly(streamResponse);
    }
    return Flux.error(OpenAIErrorTranslator.translate(e, modelName));
}
```

Alternatively, move the `doFinally` attach inside `assemble` onto a local `Flux` built before anything else can throw.

agentscope-core/src/main/java/io/agentscope/core/model/ModelContextWindows.java
```diff
@@ -68,7 +68,14 @@ private ModelContextWindows() {}
                     Map.entry("o3-mini", 200_000),
                     Map.entry("o3", 200_000),
                     Map.entry("o1-mini", 128_000),
-                    Map.entry("o1", 200_000));
+                    Map.entry("o1", 200_000),
+                    Map.entry("gpt-5.4-mini", 400_000),
```
oss-maintainer, line 72: **[Warning]** These context-window constants live in `agentscope-core`, but they describe models served by the new `agentscope-extensions-model-openai-official` module, and the existing entries in this `OPENAI` map are consumed by the OpenAI chat-completions extension too. Per the repo guidance, changes to core cascade to harness, distribution and every extension, so a wrong number here silently mis-sizes compaction for anyone resolving by model name.

Two asks:
1. Can the values be tied to a citable vendor limit in the PR description? `gpt-5.4-mini` at 400_000 against `gpt-5.4`/`5.5`/`5.6-*`/`6-astra` at 1_050_000 is a wide spread, and a too-large value fails late (provider rejection) rather than early.
2. A test asserting these model names resolve to the expected window (mirroring how the existing entries are covered) would keep the table honest.

agentscope-core/src/main/java/io/agentscope/core/model/ModelContextWindows.java
```diff
@@ -77,6 +84,7 @@ private ModelContextWindows() {}
 
     public static final Map<String, Integer> GLM =
             Map.ofEntries(
+                    Map.entry("glm-5.3", 1_000_000),
```
oss-maintainer, line 87: **[Info]** `glm-5.3` is unrelated to adding the OpenAI Responses API module. Bundling an edit to a different provider's table into a module-introduction PR makes it easy to miss in review and hard to revert on its own — worth splitting out or calling out explicitly in the description.

oss-maintainer (commented):
### Summary

Introduces a new `agentscope-extensions-model-openai-official` module implementing the OpenAI **Responses** API (as distinct from the existing chat-completions extension): SDK-backed client factory, request mapper, response parser, a streaming assembler over the SDK's `StreamResponse`, an error translator, credential type, `ModelProvider` SPI registration, an e2e `ProviderFactory` provider, and docs in both locales. 46 files, ~9.3k diff lines, of which the substantive new module is roughly 1.6k lines of `main` and the rest tests and docs.

This is a strong, unusually careful module. I specifically checked the areas this repo's review guidance calls out and they hold up:

- **Credential handling** — `OpenAIOfficialCredential.toString()` redacts the key to `apiKey=***`, and per-request `apiKey` overrides are explicitly rejected rather than silently honoured, which closes the usual leak-into-logs path.
- **Streaming reliability** — `doFinally(signal -> closeQuietly(streamResponse))` covers complete/error/cancel, terminal `failed`/`error`/`incomplete` events are translated rather than swallowed, and `maxRetries(0)` delegates retry to AgentScope with a `moduleRetryOn()` predicate rather than double-retrying.
- **Plugin isolation** — SPI registration via `META-INF/services` plus aggregation/BOM wiring, and `Check Module Sync` is green.

The one real defect is a narrow stream-leak path; the rest is a scoping question about core.

### Findings

- **[Warning]** `OpenAIResponsesChatModel.java:154` — `StreamResponse` is not closed if `assemble(...)` throws before its `doFinally` guard is attached (see inline comment for the mechanism and a fix).
- **[Warning]** `ModelContextWindows.java:72` — seven new model→context-window entries are added to `agentscope-core`, but they describe models served by this new extension and the same `OPENAI` table is consumed by the existing OpenAI extension. Core changes cascade to harness/distribution/extensions, so an inflated value mis-sizes compaction silently. A citable source per value plus a resolution test would make the table trustworthy.
- **[Info]** `ModelContextWindows.java:87` — the `glm-5.3` entry is unrelated to this feature and would be easier to review and revert as its own change.

### Suggestions

The stream cleanup is the only thing I would fix before merge; it is a few lines. If you prefer to keep the provider's model metadata with the provider, a per-extension context-window table that core consults through the SPI would also avoid the core edit entirely — not something to change in this PR, just worth a follow-up issue since every new provider module currently has to touch core.

### Tests

Coverage is comprehensive (error translation, credential, SDK client factory, request mapper, response parser, streaming assembler, cross-turn, plus e2e wiring), and CI is green on `ubuntu-latest` with 8 checks passing. Note that a 9.3k-line diff means I reviewed the new module's `main` sources and the core/pom edits closely and sampled the tests and docs rather than auditing every assertion — the maintainers should still read the assembler state machine end to end.

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/OpenAIResponsesChatModel.java
oss-maintainer, line 154: **[Warning]** `StreamResponse` leak on the assembly failure path.

`ResponsesStreamingAssembler.assemble(...)` evaluates `streamResponse.stream()` eagerly as an argument to `Flux.fromStream(...)`, and the `doFinally(signal -> closeQuietly(streamResponse))` guard is only attached to the flux *after* that call returns. If `stream()` (or anything else in `assemble` before `doFinally`) throws, the exception propagates out of `assemble`, this `catch (RuntimeException e)` converts it to `Flux.error(...)`, and the SDK stream is never closed — leaking the underlying HTTP connection for the remainder of the client's lifetime.

The window is narrow but it is exactly the path that fires when the upstream is already misbehaving, which is when connection hygiene matters most. Closing in the catch keeps the invariant local and obvious:

```java
StreamResponse<ResponseStreamEvent> streamResponse = null;
try {
    streamResponse = client.responses().createStreaming(params);
    Flux<ChatResponse> assembled =
            ResponsesStreamingAssembler.assemble(streamResponse, modelName, start);
    streamResponse = null;
    return assembled;
} catch (RuntimeException e) {
    if (streamResponse != null) {
        closeQuietly(streamResponse);
    }
    return Flux.error(OpenAIErrorTranslator.translate(e, modelName));
}
```

Alternatively, move the `doFinally` attach inside `assemble` onto a local `Flux` built before anything else can throw.

agentscope-core/src/main/java/io/agentscope/core/model/ModelContextWindows.java
oss-maintainer, line 72: **[Warning]** These context-window constants live in `agentscope-core`, but they describe models served by the new `agentscope-extensions-model-openai-official` module, and the existing entries in this `OPENAI` map are consumed by the OpenAI chat-completions extension too. Per the repo guidance, changes to core cascade to harness, distribution and every extension, so a wrong number here silently mis-sizes compaction for anyone resolving by model name.

Two asks:
1. Can the values be tied to a citable vendor limit in the PR description? `gpt-5.4-mini` at 400_000 against `gpt-5.4`/`5.5`/`5.6-*`/`6-astra` at 1_050_000 is a wide spread, and a too-large value fails late (provider rejection) rather than early.
2. A test asserting these model names resolve to the expected window (mirroring how the existing entries are covered) would keep the table honest.

agentscope-core/src/main/java/io/agentscope/core/model/ModelContextWindows.java
oss-maintainer, line 87: **[Info]** `glm-5.3` is unrelated to adding the OpenAI Responses API module. Bundling an edit to a different provider's table into a module-introduction PR makes it easy to miss in review and hard to revert on its own — worth splitting out or calling out explicitly in the description.

oss-maintainer (commented):
### Summary

Introduces a new `agentscope-extensions-model-openai-official` module implementing the OpenAI **Responses** API (as distinct from the existing chat-completions extension): SDK-backed client factory, request mapper, response parser, a streaming assembler over the SDK's `StreamResponse`, an error translator, credential type, `ModelProvider` SPI registration, an e2e `ProviderFactory` provider, and docs in both locales. 46 files, ~9.3k diff lines, of which the substantive new module is roughly 1.6k lines of `main` and the rest tests and docs.

This is a strong, unusually careful module. I specifically checked the areas this repo's review guidance calls out and they hold up:

- **Credential handling** — `OpenAIOfficialCredential.toString()` redacts the key to `apiKey=***`, and per-request `apiKey` overrides are explicitly rejected rather than silently honoured, which closes the usual leak-into-logs path.
- **Streaming reliability** — `doFinally(signal -> closeQuietly(streamResponse))` covers complete/error/cancel, terminal `failed`/`error`/`incomplete` events are translated rather than swallowed, and `maxRetries(0)` delegates retry to AgentScope with a `moduleRetryOn()` predicate rather than double-retrying.
- **Plugin isolation** — SPI registration via `META-INF/services` plus aggregation/BOM wiring, and `Check Module Sync` is green.

The one real defect is a narrow stream-leak path; the rest is a scoping question about core.

### Findings

- **[Warning]** `OpenAIResponsesChatModel.java:154` — `StreamResponse` is not closed if `assemble(...)` throws before its `doFinally` guard is attached (see inline comment for the mechanism and a fix).
- **[Warning]** `ModelContextWindows.java:72` — seven new model→context-window entries are added to `agentscope-core`, but they describe models served by this new extension and the same `OPENAI` table is consumed by the existing OpenAI extension. Core changes cascade to harness/distribution/extensions, so an inflated value mis-sizes compaction silently. A citable source per value plus a resolution test would make the table trustworthy.
- **[Info]** `ModelContextWindows.java:87` — the `glm-5.3` entry is unrelated to this feature and would be easier to review and revert as its own change.

### Suggestions

The stream cleanup is the only thing I would fix before merge; it is a few lines. If you prefer to keep the provider's model metadata with the provider, a per-extension context-window table that core consults through the SPI would also avoid the core edit entirely — not something to change in this PR, just worth a follow-up issue since every new provider module currently has to touch core.

### Tests

Coverage is comprehensive (error translation, credential, SDK client factory, request mapper, response parser, streaming assembler, cross-turn, plus e2e wiring), and CI is green on `ubuntu-latest` with 8 checks passing. Note that a 9.3k-line diff means I reviewed the new module's `main` sources and the core/pom edits closely and sampled the tests and docs rather than auditing every assertion — the maintainers should still read the assembler state machine end to end.

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-e2e-tests/src/test/java/io/agentscope/core/e2e/providers/OpenAIOfficialResponsesProvider.java
jujn: > 这个只支持通过系统环境变量获取吗？ 为什么不是从 spring 生命周期中的配置获取？

e2e 测试这里好像，不太重要吧

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/OpenAISdkClientFactory.java
jujn: > baseUrl.isNotBlank() 会不会好点？

差不多嘿嘿

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/ResponsesMultiAgentFormatter.java
```diff
@@ -0,0 +1,355 @@
...
+            case TOOL -> GroupType.TOOL_SEQUENCE;
+            case USER, ASSISTANT -> {
+                if (msg.hasContentBlocks(ToolUseBlock.class)
+                        || msg.hasContentBlocks(ToolResultBlock.class)) {
+                    yield GroupType.TOOL_SEQUENCE;
+                }
+                // Assistant messages with encrypted reasoning content must be
+                // passed through to mapAssistantMessage.
+                if (msg.getRole() == MsgRole.ASSISTANT && hasEncryptedReasoning(msg)) {
+                    yield GroupType.TOOL_SEQUENCE;
```
oss-maintainer, line 183: **[Info]** The behaviour is right — an assistant turn carrying `encrypted_content` must reach `mapAssistantMessage` rather than be folded into `<history>` — but getting there by claiming `GroupType.TOOL_SEQUENCE` depends on a side effect: that branch happens to filter on `role == ASSISTANT || role == TOOL`, and because contiguous same-type messages merge, a reasoning-bearing assistant turn now joins a neighbouring tool sequence group instead of standing alone. `GroupType.BYPASS` already encodes exactly what is wanted here ("always start a new group, mapped individually"), emits the same items, and drops the hidden coupling to the tool-branch filter. Purely a readability/robustness point.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/ResponsesMultiAgentFormatter.java
```diff
@@ -0,0 +1,355 @@
...
+                msg.getMetadata().get(MessageMetadataKeys.BYPASS_MULTIAGENT_HISTORY_MERGE);
+        return Boolean.TRUE.equals(bypassFlag);
+    }
+
+    /**
+     * Checks whether a message carries encrypted reasoning content that must be
+     * preserved as a Responses reasoning replay item (not merged into history text).
+     */
+    private static boolean hasEncryptedReasoning(Msg msg) {
+        ThinkingBlock thinkingBlock = msg.getFirstContentBlock(ThinkingBlock.class);
```
oss-maintainer, line 204: **[Info]** `getFirstContentBlock(ThinkingBlock.class)` inspects only the first block, and `ResponsesRequestMapper.mapAssistantMessage` makes the same assumption. Today the parser emits at most one `ThinkingBlock` per reply, so this holds. It stops holding the moment a reply can carry more than one — for example the metadata-only empty-text terminal block plus a summary block, or history assembled from another provider and replayed here — and then the predicate and the payload read from the wrong block, so the reasoning item is silently folded into history instead of replayed. Iterating the `ThinkingBlock`s and taking the first with a non-empty `MD_REASONING_ENCRYPTED_CONTENT` makes both call sites independent of block order.

oss-maintainer (commented):
### Summary

Re-review of `bb20ab81` (delta against `b51a889d`; the earlier reviewed commit `24f3b42b` is gone from the branch). The one defect I had gated on is fixed, and the reasoning-metadata move is coherent end to end.

**StreamResponse leak — fixed.** `streamResponse` is hoisted out of the `try` in `buildStreamingFlux` and closed in the `catch`, so the window where `assemble` threw before its `doFinally` guard was attached is covered, and `streamThrowsInAssembleClosesSdkStream` asserts it by making `stream()` throw and waiting on the close latch. `closeQuietly` was widened to package-private for exactly this, which is the narrower of the two options I suggested.

**Reasoning metadata onto `ThinkingBlock` — verified against core, and it holds.** `ThinkingAccumulator` does `metadata.putAll(block.getMetadata())` and rebuilds the block with that metadata, so `encrypted_content` now survives aggregation into the stored assistant message and is read back by `mapAssistantMessage` — the previous `ChatResponse.metadata` location was the one that got dropped on aggregation. I also checked the empty-text case: `ReActAgent`'s block emitter skips deltas when `getThinking()` is empty and `AguiStreamContext.closeReasoningMessage` no-ops for a message id that was never started, so a metadata-only `ThinkingBlock` does not surface as an empty reasoning message to AG-UI clients. Good.

The schema-level `strict` precedence (`schema.getStrict()` overriding the builder default) and its two tests look correct.

**Still open from the previous round (unchanged by this commit):** the seven `gpt-5.x` context-window entries plus `glm-5.3` are still added to `agentscope-core`'s `ModelContextWindows`, still with no citable source per value and no resolution test. I am not re-flagging it inline because it is a scoping call for the maintainers, but it was my last substantive `Warning` and it has not moved.

### Findings
- **[Info]** `ResponsesMultiAgentFormatter.java:183` — reusing `GroupType.TOOL_SEQUENCE` where `GroupType.BYPASS` is the type that means "mapped individually, never merged"; current behaviour is correct but relies on the tool-branch role filter.
- **[Info]** `ResponsesMultiAgentFormatter.java:204` — first-`ThinkingBlock`-only assumption shared with `ResponsesRequestMapper.mapAssistantMessage`.

### Compatibility note
Dropping `MD_REASONING_SUMMARY` removes `openai.reasoning.summary` from parsed-response metadata. Nothing in this repo reads it, and the module is unreleased, so this is fine — but it is a silent output-shape change for anyone already coding against the Responses module, so it is worth one line in the module's docs/changelog.

### CI / CLA / why not approving
All checks pass on `bb20ab81` (`build ubuntu-latest`, `build windows-latest`, `Check License`, `Check Module Sync`, `codecov/patch`, `validate`) and CLA is signed. I am deliberately **not** approving: `Aias00`'s `CHANGES_REQUESTED` is still the most recent review state on the PR, and the `ServicesResourceTransformer` addition to the `agentscope-all` shade config is the response to it — that resolution needs the requesting maintainer's re-review, not mine. Nothing here blocks them; both inline notes are non-blocking.

---

Aias00 (approved):
Re-reviewed current head bb20ab81dc519c4a667673fb0b52279ab694d615. The previous blocking issues are fixed: agentscope-all now merges service descriptors with ServicesResourceTransformer, encrypted reasoning replay is preserved through ThinkingBlock metadata/ReasoningContext aggregation and covered by cross-turn tests, the multi-agent formatter now passes reasoning-bearing assistant messages through to the mapper, and schema-level JsonSchema.strict takes precedence over the builder fallback with tests. CI and CLA are green. Approving.

zouyx (approved):

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/ResponsesRequestMapper.java
```diff
@@ -0,0 +1,790 @@
...
+                        JsonSchema schema = format.getJsonSchema();
+                        if (schema == null || schema.getName() == null) {
+                            throw new OpenAIOfficialModelException(
+                                    "json_schema response format requires a non-null"
+                                            + " schema or schema name.");
+                        }
+                        ResponseFormatTextJsonSchemaConfig.Builder schemaBuilder =
+                                ResponseFormatTextJsonSchemaConfig.builder()
+                                        .name(schema.getName())
+                                        .schema(buildSchema(schema.getSchema()));
```
oss-maintainer, line 181: The guard above only rejects a null `schema.getName()`, but `buildSchema(schema.getSchema())` iterates the map directly (`for (Map.Entry<String, Object> entry : schema.entrySet())`). A `ResponseFormatJsonSchema` carrying a name but a null schema map therefore throws a raw `NullPointerException` from deep inside request mapping instead of the descriptive `OpenAIOfficialModelException` used everywhere else in this class. Please also reject a null/empty `schema.getSchema()` next to the existing name check.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/OpenAIOfficialModelProvider.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        ProxyConfig proxyConfig = context.component(ProxyConfig.class);
+        if (proxyConfig != null) {
+            builder.proxy(proxyConfig);
+        }
+
+        Object raw = context.option("additionalHeaders");
+        if (raw instanceof Map<?, ?> map) {
+            builder.additionalHeaders((Map<String, String>) map);
```
oss-maintainer, line 114: This is an unchecked cast from `Object` to `Map<String, String>`. Because of type erasure the cast always succeeds, so a caller that passes `Map.of(1, 2)` or a nested map through `context.option("additionalHeaders")` gets no failure here — the `ClassCastException` surfaces later inside the SDK when the headers are actually written, which is a long way from the misconfigured option. Copy into a new `HashMap<String, String>` with explicit `instanceof String` checks on both key and value (or throw `OpenAIOfficialModelException` naming the offending entry) so the error points at the config.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/ResponsesStreamingAssembler.java
```diff
@@ -0,0 +1,322 @@
...
+            String delta = event.delta();
+            if (!delta.isEmpty()) {
+                reasoningTextAccumulator.append(delta);
+            }
+        }
+
+        private ChatResponse handleFunctionCallArgumentsDelta(
+                ResponseFunctionCallArgumentsDeltaEvent event) {
+            ToolCallInfo info = itemRegistry.get(event.itemId());
+            String callId = info != null ? info.callId() : "";
```
oss-maintainer, line 213: When a `function_call_arguments.delta` arrives with no matching `output_item.added` entry, this branch still emits a `ToolUseBlock` with an empty id and the `__fragment__` placeholder name. Downstream, tool dispatch and `tool_use`/`tool_result` pairing key off the id, so an empty id can silently drop the call or mismatch it against the wrong result, and `__fragment__` is not a registered tool name. Since the terminal `response.output()` pass already re-extracts the authoritative tool calls, consider skipping unregistered fragments (and logging at debug) rather than forwarding a malformed block into the stream.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/ResponsesHelper.java
```diff
@@ -0,0 +1,113 @@
...
+        long inputTokens = respUsage.inputTokens();
+        long outputTokens = respUsage.outputTokens();
+        long cachedTokens = respUsage.inputTokensDetails().cachedTokens();
+        long cacheCreationTokens = respUsage.inputTokensDetails().cacheWriteTokens();
+        long reasoningTokens = respUsage.outputTokensDetails().reasoningTokens();
+
+        double time = Duration.between(startTime, Instant.now()).toMillis() / 1000.0;
+
+        return ChatUsage.builder()
+                .inputTokens((int) inputTokens)
```
oss-maintainer, line 105: The SDK reports usage as `long` and `ChatUsage` stores `int`, so these casts silently wrap on overflow rather than failing. Realistic token counts will not hit it today, but with a 1M-token context window plus long reasoning output a corrupted (possibly negative) usage figure is cheap to rule out: `Math.toIntExact(inputTokens)` gives an `ArithmeticException` instead of wrong numbers in tracing/cost accounting.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-openai-official/src/main/java/io/agentscope/extensions/model/openaiofficial/credential/OpenAIOfficialCredential.java
```diff
@@ -0,0 +1,127 @@
...
+        return TYPE;
+    }
+
+    @JsonProperty("api_key")
+    public String getApiKey() {
+        return apiKey;
+    }
+
+    @JsonProperty("organization")
+    public String getOrganization() {
```
oss-maintainer, line 66: `organization` is parsed, stored and serialized, but nothing in this module consumes it — `OpenAISdkClientFactory` and `OpenAIResponsesChatModel.Builder` never call `getOrganization()`, unlike the official SDK's `OpenAIOkHttpClient.Builder#organization`. Either wire it through the client factory or drop the field, so a user who sets it does not silently get a request against the wrong organization.

oss-maintainer (commented):
### Summary
Re-review after the new commits. This is a well-structured module: the SPI registration, the `OpenAIOfficialModelException` translation layer, the request-mapper/streaming-assembler split, and the test suite (roughly 5.5k lines against ~2.5k of production code, including `CrossTurnTest` for server-side-state continuation and proxy configuration) all line up with how the existing `agentscope-extensions-model-openai` module is organised, and CI is green on both Linux and Windows.

The findings are concentrated in the paths where an invalid or out-of-order input degrades quietly instead of failing loudly: a null JSON schema map surfaces as a `NullPointerException` rather than the module's own exception type, an unchecked `additionalHeaders` cast defers a `ClassCastException` deep into the SDK, and an unregistered `function_call_arguments.delta` is forwarded as a `ToolUseBlock` with an empty id that downstream tool matching cannot key on. The two `info` items are consistency gaps against the official SDK surface rather than bugs.

None of these look like merge blockers, though the empty-tool-call-id case is worth a decision before this ships as the reference Responses API implementation.

---

chickenlj (approved):

PR #3087, opened 2026-09-10
fix(core): derive Version.VERSION from Maven project version instead of hardcoding

### LINKED ISSUE #3086: [Bug]: Version.VERSION is hardcoded and reports a stale version (1.0.13-SNAPSHOT vs 2.0.3-SNAPSHOT)
**Describe the bug**
`io.agentscope.core.Version#VERSION` is a hardcoded string literal that has drifted
out of sync with the actual Maven project version. As a result, the `User-Agent`
header sent to all model providers reports the stale version
`agentscope-java/1.0.13-SNAPSHOT` instead of the current `2.0.3-SNAPSHOT`.

**To Reproduce**
```java
import io.agentscope.core.Version;

System.out.println(Version.VERSION);          // prints 1.0.13-SNAPSHOT
System.out.println(Version.getUserAgent());   // agentscope-java/1.0.13-SNAPSHOT; ...
```
Compare with the version declared in `pom.xml` (`<revision>2.0.3-SNAPSHOT</revision>`).

**Expected behavior**
`Version.VERSION` should match the Maven project version, so `getUserAgent()` reports
the version of the artifact actually being built/released.

**Environment**
- Java Version: 21
- OS: Windows 11

**Additional context**
Root cause: the version is duplicated in three places that must be manually kept in
sync — `pom.xml` `<revision>`, `Version.java` `VERSION`, and
`VersionTest#testVersionConstant`. Proposed fix: make `${project.version}` the single
source of truth and inject it at build time via Maven resource filtering (add a
`version.properties`, read it lazily from `Version`, and cross-check it in `VersionTest`).

amyaxy:
Fixes #3086

## AgentScope-Java Version

2.0.3

### Background

`Version.VERSION` was a hardcoded constant (`1.0.13-SNAPSHOT`) that had
drifted from the actual Maven project version (`2.0.3-SNAPSHOT` in `pom.xml`
via the CI-friendly `<revision>` property), so callers of the `Version` API
could receive a stale version string at runtime.

### Purpose

Make Maven `<revision>` the single source of truth for the runtime version,
eliminating manual synchronization between `pom.xml` and `Version.java`.

### Changes

- `agentscope-core/src/main/resources/META-INF/agentscope/version.properties`
  (new): filtered with `${project.version}` during the build
- `agentscope-core/src/main/java/io/agentscope/core/Version.java`: load
  `VERSION` from the classpath resource at runtime; fall back to `"unknown"`
  if an unfiltered build leaks the literal `${` placeholder
- `agentscope-core/pom.xml`: enable resource filtering for the version
  properties file only; keep all other resources unfiltered
- `agentscope-core/src/test/resources/agentscope-test-version.properties`
  (new): test-only filtered resource for cross-verification
- `agentscope-core/src/test/java/io/agentscope/core/VersionTest.java`: assert
  `Version.VERSION` equals the Maven-resolved version and matches
  semantic-versioning; gracefully skip under non-Maven (IDEA) builds

### How to test

```
mvn -pl agentscope-core test -Dtest=VersionTest
mvn -pl agentscope-core clean verify
```

## Checklist

- [x] All tests are passing (`mvn test`) — pending CI confirmation
- [x] Related documentation has been updated (N/A — no doc changes in this PR)

### DIFF
--- a/agentscope-core/pom.xml
+++ b/agentscope-core/pom.xml
@@ -206,7 +206,77 @@
     </dependencies>
 
     <build>
+        <resources>
+            <!-- Non-filtered default, excluding the version file so this block never writes the
+                 same target path as the filtered block below. If both blocks copied
+                 META-INF/agentscope/version.properties, the unfiltered copy (which runs first and
+                 rewrites the target with a newer timestamp than the source) would make the
+                 filtered copy be skipped by maven-resources-plugin (default overwrite=false),
+                 silently packaging the literal ${project.version} placeholder. -->
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>false</filtering>
+                <excludes>
+                    <exclude>META-INF/agentscope/version.properties</exclude>
+                </excludes>
+            </resource>
+            <!-- Filter only the version file so ${project.version} is resolved. This block is the
+                 sole writer of the target version.properties. -->
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+                <includes>
+                    <include>META-INF/agentscope/version.properties</include>
+                </includes>
+            </resource>
+        </resources>
+        <testResources>
+            <testResource>
+                <directory>src/test/resources</directory>
+                <filtering>false</filtering>
+                <excludes>
+                    <exclude>agentscope-test-version.properties</exclude>
+                </excludes>
+            </testResource>
+            <testResource>
+                <directory>src/test/resources</directory>
+                <filtering>true</filtering>
+                <includes>
+                    <include>agentscope-test-version.properties</include>
+                </includes>
+            </testResource>
+        </testResources>
         <plugins>
+            <!-- Fail the build loudly if the filtered version resource still contains the
+                 ${project.version} placeholder (i.e. Maven resource filtering did not apply).
+                 Without this check the packaged jar would silently carry an unfiltered literal,
+                 and Version.resolveVersionFrom would mask it by returning "unknown". -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <version>3.2.0</version>
+                <executions>
+                    <execution>
+                        <id>verify-version-resource-filtered</id>
+                        <phase>process-classes</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <fail message="META-INF/agentscope/version.properties still contains the unfiltered 'dollar-brace' placeholder (the Maven project.version expression was not resolved): Maven resource filtering did not apply. Check the resource filtering configuration in agentscope-core/pom.xml.">
+                                    <condition>
+                                        <!-- A single '$' is enough to detect the unmapped
+                                             ${project.version} placeholder: the filtered file is
+                                             just 'version=2.0.3-SNAPSHOT' and contains none. -->
+                                        <resourcecontains resource="${project.build.outputDirectory}/META-INF/agentscope/version.properties" substring="$" />
+                                    </condition>
+                                </fail>
+                            </target>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
             <!-- Publish a test-jar so downstream modules (agentscope-harness) can reuse the
                  shared test fixtures (MockModel, MockToolkit, TestConstants, TestUtils). -->
             <plugin>
--- a/agentscope-core/src/main/java/io/agentscope/core/Version.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/Version.java
@@ -15,19 +15,147 @@
  */
 package io.agentscope.core;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /**
  * AgentScope version and User-Agent information.
  *
  * <p>Provides a unified User-Agent string for all model requests to identify AgentScope Java
  * clients and collect usage statistics.
+ *
+ * <p>The version is resolved at class-load time through a three-level fallback chain:
+ *
+ * <ol>
+ *   <li>the {@code META-INF/agentscope/version.properties} classpath resource, whose
+ *       {@code ${project.version}} placeholder is resolved by Maven resource filtering;
+ *   <li>the {@code Implementation-Version} manifest entry of the containing jar, which is
+ *       populated from {@code ${project.version}} by the Maven jar plugin;
+ *   <li>the literal {@code "unknown"} as a last resort (e.g. running straight from IDE
+ *       {@code target/classes} with no manifest), accompanied by a one-time warning log.
+ * </ol>
  */
 public final class Version {
 
-    /** AgentScope Java version */
-    public static final String VERSION = "1.0.13-SNAPSHOT";
+    private static final Logger log = LoggerFactory.getLogger(Version.class);
+
+    private static final String VERSION_RESOURCE = "/META-INF/agentscope/version.properties";
+
+    /** Sentinel returned when no build-resolved version can be discovered. */
+    static final String UNKNOWN = "unknown";
+
+    /**
+     * AgentScope Java version.
+     *
+     * <p>Injected at build time by Maven resource filtering from {@code ${project.version}};
+     * see the class Javadoc for the fallback chain.
+     */
+    public static final String VERSION = resolveVersion();
 
     private Version() {
         // Utility class - prevent instantiation
+    }
+
+    private static String resolveVersion() {
+        // 1) Maven-filtered classpath resource (normal Maven builds, tests).
+        String fromResource;
+        try (InputStream in = Version.class.getResourceAsStream(VERSION_RESOURCE)) {
+            fromResource = resolveVersionFromResource(in);
+        } catch (IOException e) {
+            log.debug("Failed to read {} from the classpath", VERSION_RESOURCE, e);
+            fromResource = UNKNOWN;
+        }
+        if (!UNKNOWN.equals(fromResource)) {
+            return fromResource;
+        }
+
+        // 2) jar manifest Implementation-Version (packaged jars whose META-INF/agentscope
+        //    directory may have been dropped, e.g. by shading or custom packaging).
+        String fromManifest = resolveVersionFromManifest(getImplementationVersion());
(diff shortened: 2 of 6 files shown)


agentscope-core/src/main/java/io/agentscope/core/Version.java
```diff
@@ -23,13 +27,54 @@
...
 
+    private static String resolveVersion() {
+        try (InputStream in = Version.class.getResourceAsStream(VERSION_RESOURCE)) {
+            if (in != null) {
+                Properties props = new Properties();
+                props.load(in);
+                return resolveVersionFrom(props);
+            }
+        } catch (IOException e) {
+            // Fall through to the fallback below.
```
tanzzj: and some log or catch (IOException ignore) ?

agentscope-core/pom.xml
```diff
@@ -206,6 +206,34 @@
     </dependencies>
 
     <build>
+        <resources>
+            <!-- Non-filtered default: avoids mangling future ${...} in other resources. -->
+            <resource>
```
oss-maintainer, line 211: Relying on declaration order to make the filtered copy win is fragile here: both resource blocks read the same `src/main/resources` tree and the same target path, and `maven-resources-plugin` skips a copy when the destination is not older than the source (default `overwrite=false`). Since the unfiltered copy runs first and rewrites `target/classes/META-INF/agentscope/version.properties` with a *newer* timestamp than the source, the filtered copy can be skipped and the packaged jar ends up containing the literal `${project.version}`. `Version.resolveVersionFrom` then masks that by returning `"unknown"`, so the failure is silent rather than loud. Safer: drop the catch-all unfiltered block's duplicate and exclude the version file from it, e.g. keep one `<resource>` for `src/main/resources` with `<excludes><exclude>META-INF/agentscope/version.properties</exclude></excludes>` plus the filtered resource — then add an assertion (or a `maven-enforcer`/`groovy` check) that the packaged resource no longer contains `${`.

agentscope-core/src/main/resources/META-INF/agentscope/version.properties
```diff
@@ -0,0 +1 @@
+version=${project.version}
```
oss-maintainer, line 1: Missing trailing newline (note the `\ No newline at end of file` marker). Harmless for `Properties.load`, but spotless/POSIX hygiene tools flag it.

agentscope-core/src/main/java/io/agentscope/core/Version.java
```diff
@@ -23,13 +27,54 @@
...
+        }
+        return "unknown";
```
oss-maintainer, line 53: `getUserAgent()` is used as a real User-Agent for model-provider traffic (and `Version.VERSION` feeds `McpClientBuilder` client info and `TelemetryTracer` instrumentation version). Falling back to `"unknown"` means every run that resolves outside a Maven-filtered classpath — IDE run configs, `java -cp target/classes`, shaded/relocated jars that drop `META-INF/agentscope`, custom packaging — now reports `agentscope-java/unknown`, which is worse telemetry than the previous hardcoded constant. Consider keeping a build-time fallback that is still meaningful (e.g. a `generateGitPropertiesFile`/`impl-version` from the manifest's `Implementation-Version`, then the old hardcoded value) and logging once at debug/warn when the resource is unresolved.

agentscope-core/src/test/java/io/agentscope/core/VersionTest.java
```diff
@@ -26,12 +30,64 @@
...
+            props.load(in);
+        }
+        String expectedVersion = props.getProperty("version");
+        Assertions.assertNotNull(expectedVersion, "version property should be present");
+
+        // When Maven resource filtering was not run (e.g. running directly from an IDE), the
+        // value is an unfiltered ${project.version} literal and there is no real version to
+        // cross-check, so skip the strict assertions. CI runs Maven, so filtering always applies.
+        boolean filtered = !expectedVersion.contains("${");
+        Assumptions.assumeTrue(
```
oss-maintainer, line 48: `Assumptions.assumeTrue(filtered, ...)` makes the strict cross-check self-skipping, so the scenario described above (Maven *did* run but the filtered resource was not applied) is reported as a pass/skip instead of a failure. If the resource is present on the test classpath at all, an unfiltered `${project.version}` is arguably a genuine failure rather than an assumption violation — maybe gate the skip on the resource being absent, not on the value being unfiltered.

oss-maintainer (commented):
### Summary
Deriving `Version.VERSION` from the Maven project version removes a hardcoded string that had already gone stale — good direction, and the unfiltered-placeholder guard plus the extra unit tests show real care. The blocker is that the two overlapping `<resource>` blocks may leave the **unfiltered** file in the jar, and the `"unknown"` fallback would hide that.

### Findings
- **[Critical]** `pom.xml:211` — catch-all unfiltered resource + filtered resource writing the same target path; without `<excludes>` (or `overwrite`) the filtered copy can be skipped, shipping the literal `${project.version}`.
- **[Warning]** `Version.java:53` — `"unknown"` in the User-Agent / MCP client info / telemetry version is a regression vs. the previous constant for any non-Maven-filtered runtime.
- **[Info]** `version.properties:1` — no trailing newline.
- **[Info]** `VersionTest.java:48` — the `Assumptions` skip can swallow exactly the packaging bug above.

### Suggestions
```xml
<resource>
    <directory>src/main/resources</directory>
    <filtering>false</filtering>
    <excludes>
        <exclude>META-INF/agentscope/version.properties</exclude>
    </excludes>
</resource>
<resource>
    <directory>src/main/resources</directory>
    <filtering>true</filtering>
    <includes>
        <include>META-INF/agentscope/version.properties</include>
    </includes>
</resource>
```
Then verify the artifact directly in CI, e.g. `unzip -p agentscope-core/target/*.jar META-INF/agentscope/version.properties | grep -q '${' && exit 1`.

---

agentscope-core/pom.xml
```diff
@@ -206,7 +206,77 @@
...
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <fail message="META-INF/agentscope/version.properties still contains the unfiltered 'dollar-brace' placeholder (the Maven project.version expression was not resolved): Maven resource filtering did not apply. Check the resource filtering configuration in agentscope-core/pom.xml.">
+                                    <condition>
+                                        <!-- A single '$' is enough to detect the unmapped
+                                             ${project.version} placeholder: the filtered file is
+                                             just 'version=2.0.3-SNAPSHOT' and contains none. -->
+                                        <resourcecontains resource="${project.build.outputDirectory}/META-INF/agentscope/version.properties" substring="$" />
```
oss-maintainer, line 272: Good addition - this is the loud failure the previous round asked for. One nit: `substring="$"` is a blunt instrument, because any future `$` in the file (a comment line, a build-metadata entry) would break the build for an unrelated reason. The thing you actually want to reject is the literal `${project.version}` placeholder, and since Maven interpolates `${...}` inside the POM, matching on `project.version` gets you that precision without any escaping trick:

```xml
<resourcecontains resource="${project.build.outputDirectory}/META-INF/agentscope/version.properties" substring="project.version" />
```

A correctly filtered file (`version=2.0.3-SNAPSHOT`) can never contain that token.

agentscope-core/pom.xml
```diff
@@ -206,7 +206,77 @@
...
+                 Without this check the packaged jar would silently carry an unfiltered literal,
+                 and Version.resolveVersionFrom would mask it by returning "unknown". -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <version>3.2.0</version>
+                <executions>
+                    <execution>
+                        <id>verify-version-resource-filtered</id>
+                        <phase>process-classes</phase>
```
oss-maintainer, line 261: `process-classes` runs on every `mvn compile`, so this execution also fires in builds that never produced `target/classes/META-INF/agentscope/version.properties` (resources phase skipped, stale/partial `target/`). Ant's `resourcecontains` generally reports a missing resource as its own error, which would replace the curated `fail message` with something less actionable. If you want the diagnostic to survive those cases, gate it first, e.g. fail when `<not><available file="${project.build.outputDirectory}/META-INF/agentscope/version.properties"/></not>`, or combine the two conditions with `<and>`. Not blocking - CI and normal builds are covered.

pom.xml
```diff
@@ -332,6 +333,22 @@
...
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>${maven-jar-plugin.version}</version>
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <!-- Record the resolved ${project.version} in the jar manifest so
+                                 Version can fall back to Implementation-Version when the filtered
+                                 version.properties resource is absent from the classpath (e.g.
+                                 shaded/relocated jars or custom packaging). -->
+                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
```
oss-maintainer, line 347: Heads-up on blast radius rather than a defect: this sits in the parent's `<build><plugins>`, so `Implementation-Title` / `Implementation-Version` / `Implementation-Vendor` now land in **every** module jar's manifest, not just `agentscope-core`. That is harmless (and useful for support tooling), but if you want to keep the change scoped to the module that reads it, move the plugin block into `agentscope-core/pom.xml`. Worth noting the shaded `agentscope-all` artifact already carries core's filtered `META-INF/agentscope/version.properties`, so level 1 covers it and the manifest entry is mostly a safety net for custom packaging / Spring Boot fat jars where `Package#getImplementationVersion()` can come back null.

oss-maintainer (approved):
### Summary
Re-review after `c4eb8bef`: all four findings from the previous round are properly resolved — the unfiltered resource block now excludes `META-INF/agentscope/version.properties`, so the filtered block is the sole writer and the declaration-order race is gone; the `"unknown"` fallback is now a three-level chain with a jar-manifest `Implementation-Version` rung plus a one-time warning; the antrun check fails the build loudly instead of silently shipping the placeholder; and the test no longer self-skips the exact packaging bug it was written to catch (it only aborts when the resource is genuinely absent). New tests cover the manifest / resource / IO branches.

CI at review time: `Check License`, `Check Module Sync`, `build (ubuntu-latest)`, `codecov/patch` and `license/cla` are green, `build (windows-latest)` was still queued.

The three remaining notes are `Info`-level only (match `project.version` instead of a bare `$`, tolerate a missing file in the antrun condition, and the parent-pom manifest scope) — none of them block merging.

LGTM, and thanks for the thorough follow-up.

---

tanzzj:
LGTM

PR #3093, opened 2026-09-10
fix(ollama): serialize tool result name as tool_name

### LINKED ISSUE #3092: [Bug]: Ollama tool result messages use name instead of tool_name
## Describe the bug

When AgentScope-Java serializes a tool result for an Ollama chat request, the tool name is emitted as the JSON field `name`.

According to Ollama's official tool-calling protocol, a tool result message should use the `tool_name` field:

https://docs.ollama.com/capabilities/tool-calling

For example:

```json
{
  "role": "tool",
  "tool_name": "get_temperature",
  "content": "22°C"
}
```

## To Reproduce

Using the current `main` branch:

```java
ToolResultBlock toolResult =
        new ToolResultBlock(
                "call123",
                "test_tool",
                List.of(TextBlock.builder().text("Tool result output").build()),
                null);

Msg msg = Msg.builder()
        .role(MsgRole.TOOL)
        .content(toolResult)
        .build();

OllamaMessage converted = new OllamaMessageConverter().convertMessage(msg);
String json = new JacksonJsonCodec().toJson(converted);
```

## Actual behavior

The tool result name is serialized using the `name` field:

```json
{
  "role": "tool",
  "content": "Tool result output",
  "tool_call_id": "call123",
  "name": "test_tool"
}
```

## Expected behavior

The serialized message should use Ollama's `tool_name` field:

```json
{
  "role": "tool",
  "content": "Tool result output",
  "tool_call_id": "call123",
  "tool_name": "test_tool"
}
```

## Impact

The Java object retains the correct tool name, so object-level conversion tests can pass. However, the final wire-format JSON does not use the field name expected by Ollama's tool-calling protocol.

This may prevent Ollama from correctly interpreting tool results during subsequent tool-calling turns.

## Evidence

- The existing conversion test verifies `OllamaMessage.getName()`, but does not verify the serialized JSON field name.
- A wire-format regression test reproduces the mismatch on the current `main` branch.
- The mismatch is isolated to the JSON property mapping of the Ollama tool-result message.

## Scope

This issue concerns the serialization of tool result messages only. Tool definitions and `tool_calls.function.name` should remain unchanged.

liugy789:
## Summary

- Serialize Ollama tool result messages with the `tool_name` JSON field.
- Add a wire-format regression test to prevent the incorrect `name` field from returning.

Fixes #3092

## Root Cause

`OllamaMessage` stores the tool name in the internal Java field `name`, but the field was serialized with:

```java
@JsonProperty("name")
```

Ollama's official tool-calling protocol expects tool result messages to use `tool_name`:

https://docs.ollama.com/capabilities/tool-calling

As a result, AgentScope-Java produced tool result payloads like:

```json
{
  "role": "tool",
  "content": "Tool result output",
  "tool_call_id": "call123",
  "name": "test_tool"
}
```

The expected payload is:

```json
{
  "role": "tool",
  "content": "Tool result output",
  "tool_call_id": "call123",
  "tool_name": "test_tool"
}
```

## Implementation

- Changed the Jackson mapping in `OllamaMessage` from `name` to `tool_name`.
- Kept the existing Java getter and setter unchanged.
- Added a regression test covering:

  `ToolResultBlock → OllamaMessageConverter → JSON serialization`

The fix applies to the shared Ollama message DTO used by the regular converter, chat formatter, and multi-agent formatter.

## Compatibility and Scope

- No public Java accessor changes.
- No changes to `tool_calls.function.name`.
- No changes to `tool_call_id`.
- No changes to Python code or other model providers.
- The change is limited to Ollama tool result message serialization.

## Tests

- [x] Added a regression test that verifies `tool_name` is present in the serialized JSON.
- [x] Verified that the incorrect `name` field is not emitted.
- [x] Ollama formatter and converter tests pass.
- [x] Full Ollama module tests pass.
- [x] Spotless check passes.
- [x] `git diff --check` passes.

No live Ollama server E2E test was run locally. The regression test validates the wire-format serialization boundary directly.


amyaxy:
LGTM

amyaxy (commented):
LGTM

oss-maintainer (approved):
LGTM — verified the fix at the DTO level rather than taking the description on faith:

- `OllamaMessage` is annotated `@JsonInclude(NON_NULL)` at class level, so the renamed `tool_name` never leaks onto `system`/`user`/`assistant` messages (they leave `name` null).
- `setName(...)` is only reached on the `ToolResultBlock` paths (`OllamaMessageConverter:92`, `OllamaChatFormatter:174`, `OllamaMultiAgentFormatter:404`), and the multi-agent formatter sets `role = "tool"` before setting the name — so no site produces a non-tool message carrying `tool_name`.
- `tool_calls.function.name` and `tool_call_id` are untouched, so the request shape stays protocol-correct in both directions.
- I grepped the remaining `@JsonProperty("name")` occurrences: `OllamaFunction` / `OllamaToolFunction` are tool *declaration* DTOs where `name` is correct per the Ollama docs — no sibling fix needed.

Two optional follow-ups (not blocking, no need to force-push for these):
1. `assertTrue(json.contains("\"tool_name\":\"test_tool\""))` couples the assertion to Jackson's field ordering and to the current no-pretty-print output. Deserialising the JSON back to a `Map` and asserting on the keys (`containsKey("tool_name")`, `!containsKey("name")`) would keep the same regression covered without that coupling.
2. Consider `@JsonAlias("name")` on the field if any tests/fixtures or replayed conversations in the wild carry the old `name` key on tool messages, so deserialization stays tolerant of the previous wire shape.

Nice catch and a clean, minimal fix. 🎉

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/dto/OllamaMessage.java
```diff
@@ -48,7 +48,7 @@ public class OllamaMessage {
     private String toolCallId;
 
     /** The name of the tool/function (optional/required for tool messages). */
-    @JsonProperty("name")
+    @JsonProperty("tool_name")
```
oss-maintainer, line 51: Renaming the property on the DTO is a wire-format change for every `/api/chat` request carrying a tool result, not just the `OllamaMessageConverter` path covered by the new test. Three call sites set this field: `OllamaMessageConverter:92`, `OllamaChatFormatter:174`, `OllamaMultiAgentFormatter:404`. Please state the minimum supported Ollama server version in the description and, ideally, a smoke result for the other two formatters: a server that only understands `name` now receives neither key and loses the tool association silently. If a version range must keep working, emitting both keys would be needed — `@JsonAlias` only helps the read direction — so that trade-off is worth deciding explicitly.

Same DTO, read direction: `OllamaResponse.message` deserializes into this class with `ignoreUnknown = true`, so a server echoing `"name"` now leaves `getName()` null. Nothing on main reads it off a parsed response today, but `@JsonAlias("name")` is free and keeps the read side tolerant of both spellings.

oss-maintainer (commented):
### Summary
The fix is sound and verified at the DTO level: `OllamaMessage` is `@JsonInclude(NON_NULL)`, so `tool_name` cannot leak onto non-tool messages, and the new test asserts both the presence of `tool_name` and the absence of `name`. Re-reviewed because new commits landed after the earlier approval. Filing COMMENT (not a second APPROVE) so the server-version question gets answered first.

### Findings
- **[Warning]** `OllamaMessage.java:51` — the rename affects all three formatters, but only the converter path is tested; document the minimum supported Ollama version and consider `@JsonAlias("name")` for the response direction.

### Suggestions
One serialization assertion per formatter (`OllamaChatFormatter`, `OllamaMultiAgentFormatter`) for a tool result would close the coverage gap.

---

jujn (approved):

PR #3094, opened 2026-09-10
fix(harness): defer self-managed sandbox shutdown until session mirro…

### LINKED ISSUE #3036: [Bug]: SessionTree mirror NPEs when it outlives the sandbox — PinnedSandboxFilesystem cannot know the pinned sandbox is closed (regression from #2946, v2.0.3)

xy-ygz:
## Summary
Fix NPE caused by synchronous call shutdown clearing sandbox connections before async sandbox write completes
Fixes #3036

## Behavior notes (ops)
Repair Ideas：
When a call completes, if there are still outstanding asynchronous mirror tasks relying on the sandbox connection, do not destroy the connection immediately. Defer the release ownership until the reference count drops to zero.

Key changes:
1. Add a new deferred‑release coordinator
‑ Enqueue mirror task: call `retain()` (increment reference count by 1)
‑ Task completion: call `releaseMirror()` (decrement count by 1; perform real shutdown only when count reaches zero)
‑ Call teardown: call `requestRelease()` — if pending tasks exist, defer destruction; otherwise shut down immediately
‑ Self‑managed sandbox: keep existing behavior; do not physically destroy the sandbox instance
2. In `SessionTree`, prepare data before enqueuing asynchronous mirror tasks and then invoke `retain()`. Roll back the reference count upon submission failure. Decrement the count inside the `finally` block.
3. Destruction must not block mirror threads. Deferred `stop()/shutdown()` runs on a dedicated cleanup thread, preventing slow‑teardown operations from blocking the global mirror queue.
4. Defensive safeguards
‑ When the sandbox is already stopped (`isRunning() == false`), treat upload/download as soft‑fail to avoid crashing on dead connections.
‑ Prevent unintended destruction when reference‑count underflow occurs (count goes negative).
5. Test coverage
‑ Deferred‑release logic
‑ Sandbox shutdown only after completion of dual tasks (segment‑level + full‑file mirroring)
‑ Avoid premature teardown on reference‑count underflow

## Test plan

- [x] `mvn -pl agentscope-harness -am -DskipITs -Dtest=SandboxMirrorReleaseCoordinatorTest,SessionTreeMirrorTest -Dsurefire.failIfNoSpecifiedTests=false test`
- [x] `mvn -pl agentscope-harness -am -DskipITs -Dtest=SandboxBackedFilesystemTest -Dsurefire.failIfNoSpecifiedTests=false test`
- [x] Manual: self-managed K8s/Docker call with session flush — mirror completes, then pod/claim terminates; no NPE on `session-tree-mirror`
- [x] Manual: user-managed sandbox path still reuses sandbox across calls (no shutdown on release)

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
@@ -61,6 +61,14 @@ public class SandboxBackedFilesystem extends BaseSandboxFilesystem implements Sa
 
     private static final Logger log = LoggerFactory.getLogger(SandboxBackedFilesystem.class);
 
+    /**
+     * Shared message for a missing or already-stopped sandbox. Matches {@link
+     * #requireSandbox(RuntimeContext)} so async mirrors degrade to the same warn text instead of
+     * NPE after {@code stop}/{@code shutdown}.
+     */
+    static final String NO_ACTIVE_SANDBOX_MESSAGE =
+            "No active sandbox — sandbox filesystem used outside of a call context";
+
     private final String fsId;
     private volatile Sandbox sandbox;
 
@@ -122,7 +130,14 @@ public class SandboxBackedFilesystem extends BaseSandboxFilesystem implements Sa
     @Override
     public List<FileUploadResponse> uploadFiles(
             RuntimeContext runtimeContext, List<Map.Entry<String, byte[]>> files) {
-        Sandbox active = requireSandbox(runtimeContext);
+        Sandbox active = resolveSandboxForTransfer(runtimeContext);
+        if (active == null) {
+            List<FileUploadResponse> failed = new ArrayList<>(files.size());
+            for (Map.Entry<String, byte[]> file : files) {
+                failed.add(FileUploadResponse.fail(file.getKey(), NO_ACTIVE_SANDBOX_MESSAGE));
+            }
+            return failed;
+        }
         List<FileUploadResponse> results = new ArrayList<>(files.size());
 
         for (Map.Entry<String, byte[]> file : files) {
@@ -177,7 +192,14 @@ public class SandboxBackedFilesystem extends BaseSandboxFilesystem implements Sa
     @Override
     public List<FileDownloadResponse> downloadFiles(
             RuntimeContext runtimeContext, List<String> paths) {
-        Sandbox active = requireSandbox(runtimeContext);
+        Sandbox active = resolveSandboxForTransfer(runtimeContext);
+        if (active == null) {
+            List<FileDownloadResponse> failed = new ArrayList<>(paths.size());
+            for (String path : paths) {
+                failed.add(FileDownloadResponse.fail(path, NO_ACTIVE_SANDBOX_MESSAGE));
+            }
+            return failed;
+        }
         List<FileDownloadResponse> results = new ArrayList<>(paths.size());
 
         for (String path : paths) {
@@ -248,10 +270,35 @@ public class SandboxBackedFilesystem extends BaseSandboxFilesystem implements Sa
             s = sandbox;
         }
         if (s == null) {
-            throw new SandboxException.SandboxConfigurationException(
-                    "No active sandbox — sandbox filesystem used outside of a call context");
+            throw new SandboxException.SandboxConfigurationException(NO_ACTIVE_SANDBOX_MESSAGE);
         }
         return s;
+    }
+
+    /**
+     * Like {@link #requireSandbox}, but when the bound sandbox has already stopped: soft-fail
+     * callers ({@link #softFailWhenStopped()} {@code true}, i.e. pinned async mirrors) get {@code
+     * null} after a warn; hard-fail callers (default sync tool path) still throw.
+     */
+    private Sandbox resolveSandboxForTransfer(RuntimeContext runtimeContext) {
+        Sandbox s = requireSandbox(runtimeContext);
+        if (!s.isRunning()) {
+            if (softFailWhenStopped()) {
+                log.warn("[sandbox-fs] {}", NO_ACTIVE_SANDBOX_MESSAGE);
+                return null;
+            }
+            throw new SandboxException.SandboxConfigurationException(NO_ACTIVE_SANDBOX_MESSAGE);
+        }
+        return s;
+    }
+
+    /**
+     * Whether upload/download should soft-fail when the bound sandbox is present but already
+     * stopped. Default {@code false} keeps synchronous tool calls loud; {@link
+     * PinnedSandboxFilesystem} overrides to {@code true} for async session mirrors.
+     */
+    protected boolean softFailWhenStopped() {
+        return false;
     }
 
     private String shellSingleQuote(String s) {
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTree.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTree.java
@@ -21,7 +21,9 @@ import io.agentscope.harness.agent.filesystem.AbstractFilesystem;
 import io.agentscope.harness.agent.filesystem.model.ReadResult;
 import io.agentscope.harness.agent.filesystem.sandbox.PinnedSandboxFilesystem;
 import io.agentscope.harness.agent.sandbox.Sandbox;
+import io.agentscope.harness.agent.sandbox.SandboxAcquireResult;
 import io.agentscope.harness.agent.sandbox.SandboxAware;
+import io.agentscope.harness.agent.sandbox.SandboxMirrorReleaseCoordinator;
 import io.agentscope.harness.agent.transcript.ObjectStoreTranscriptStore;
 import io.agentscope.harness.agent.transcript.TranscriptRef;
 import io.agentscope.harness.agent.transcript.TranscriptStore;
@@ -112,9 +114,15 @@ public class SessionTree {
      * @return {@code true} if the mirrors quiesced within the timeout
      */
     public static boolean awaitMirrorQuiescence(long timeout, TimeUnit unit) {
+        long deadlineNanos = System.nanoTime() + unit.toNanos(timeout);
         try {
-            MIRROR_EXECUTOR.submit(() -> {}).get(timeout, unit);
-            return true;
+            long mirrorWaitNanos = Math.max(0L, deadlineNanos - System.nanoTime());
+            MIRROR_EXECUTOR.submit(() -> {}).get(mirrorWaitNanos, TimeUnit.NANOSECONDS);
+            // Deferred stop/shutdown runs on a separate executor; drain it too so graceful close
+            // does not race sandbox teardown that was scheduled from mirror finally blocks.
+            long releaseWaitNanos = Math.max(0L, deadlineNanos - System.nanoTime());
+            return SandboxMirrorReleaseCoordinator.awaitReleaseQuiescence(
+                    releaseWaitNanos, TimeUnit.NANOSECONDS);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             return false;
@@ -550,7 +558,7 @@ public class SessionTree {
         if (transcriptStore == null || transcriptRef == null || entries.isEmpty()) {
             return;
         }
-        // Same pin as scheduleMirror: async segment upload must survive call unbind.
+        // Prepare work before retain so a failure here cannot leak a deferred sandbox release.
         final AbstractFilesystem mirrorFs = pinIfSandbox(filesystem);
         final TranscriptStore store = transcriptStoreForMirror(mirrorFs);
         final TranscriptRef ref = transcriptRef;
@@ -558,9 +566,10 @@ public class SessionTree {
         for (SessionEntry entry : entries) {
             sb.append(JsonUtils.getJsonCodec().toJson(entry)).append('\n');
         }
-        byte[] payload = sb.toString().getBytes(StandardCharsets.UTF_8);
-        String wid = writerId;
-        MIRROR_EXECUTOR.execute(
+        final byte[] payload = sb.toString().getBytes(StandardCharsets.UTF_8);
+        final String wid = writerId;
+        submitPinnedMirror(
+                mirrorFs,
                 () -> {
                     try {
                         store.appendSegment(ref, seqStart, seqEnd, wid, payload);
@@ -584,10 +593,12 @@ public class SessionTree {
         if (filesystem == null || workspaceRoot == null) {
             return;
         }
+        // Resolve paths before retain so retain↔execute has no prepare-side leak window.
         final AbstractFilesystem mirrorFs = pinIfSandbox(filesystem);
         final String contextRel = resolveRelativePath(contextFile);
         final String logRel = resolveRelativePath(logFile);
-        MIRROR_EXECUTOR.execute(
+        submitPinnedMirror(
+                mirrorFs,
                 () -> {
                     mirrorToFilesystem(mirrorFs, contextFile, contextRel);
                     mirrorToFilesystem(mirrorFs, logFile, logRel);
@@ -595,17 +606,75 @@ public class SessionTree {
     }
 
     /**
-     * When {@code fs} is a call-scoped sandbox proxy with an active binding, return a pinned
-     * filesystem that keeps that sandbox for async uploads. Otherwise return {@code fs} as-is.
+     * Retains a pinned sandbox (if any), submits {@code task} to the mirror executor, and always
+     * pairs {@code releaseMirror} — either in the task {@code finally} or immediately when
+     * submission itself fails.
      */
-    private static AbstractFilesystem pinIfSandbox(AbstractFilesystem fs) {
-        if (fs instanceof SandboxAware aware) {
-            Sandbox sb = aware.getSandbox();
-            if (sb != null) {
-                return new PinnedSandboxFilesystem(sb);
+    private static void submitPinnedMirror(AbstractFilesystem mirrorFs, Runnable task) {
+        final Sandbox retainedSandbox = sandboxForMirrorRetain(mirrorFs);
+        if (retainedSandbox != null) {
+            SandboxMirrorReleaseCoordinator.retain(retainedSandbox);
+        }
+        try {
+            MIRROR_EXECUTOR.execute(
+                    () -> {
+                        try {
+                            task.run();
+                        } finally {
+                            if (retainedSandbox != null) {
+                                SandboxMirrorReleaseCoordinator.releaseMirror(retainedSandbox);
+                            }
+                        }
+                    });
+        } catch (RuntimeException e) {
+            if (retainedSandbox != null) {
+                SandboxMirrorReleaseCoordinator.releaseMirror(retainedSandbox);
             }
+            throw e;
+        }
+    }
+
+    /**
+     * When {@code fs} is a sandbox-backed filesystem, return a pinned filesystem that keeps the
+     * <em>per-call</em> sandbox for async uploads. Prefers {@link SandboxAcquireResult} on {@link
+     * #fsRc} (issue #2490) over the legacy shared {@link SandboxAware#getSandbox()} field, which
+     * is last-writer-wins under concurrent distinct sessions.
+     */
+    private AbstractFilesystem pinIfSandbox(AbstractFilesystem fs) {
+        Sandbox sb = resolveSandboxForPin(fs);
+        if (sb != null) {
+            return new PinnedSandboxFilesystem(sb);
         }
         return fs;
+    }
+
+    /**
+     * Resolves the sandbox to pin for an async mirror: per-call {@link RuntimeContext} binding
+     * first, then the shared {@link SandboxAware} fallback for context-free callers.
+     */
+    private Sandbox resolveSandboxForPin(AbstractFilesystem fs) {
+        RuntimeContext rc = fsRc;
+        if (rc != null) {
+            SandboxAcquireResult bound = rc.get(SandboxAcquireResult.class);
+            if (bound != null && bound.getSandbox() != null) {
+                return bound.getSandbox();
+            }
+        }
+        if (fs instanceof SandboxAware aware) {
+            return aware.getSandbox();
+        }
+        return null;
+    }
+
+    /**
+     * Returns the sandbox pinned for an async mirror task, or {@code null} when the mirror does
+     * not hold a sandbox connection that must defer self-managed release.
+     */
+    private static Sandbox sandboxForMirrorRetain(AbstractFilesystem mirrorFs) {
+        if (mirrorFs instanceof PinnedSandboxFilesystem pinned) {
+            return pinned.getSandbox();
+        }
+        return null;
     }
 
     private TranscriptStore transcriptStoreForMirror(AbstractFilesystem mirrorFs) {
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
@@ -21,6 +21,7 @@ import io.agentscope.harness.agent.sandbox.Sandbox;
 import io.agentscope.harness.agent.sandbox.SandboxAcquireResult;
 import io.agentscope.harness.agent.sandbox.SandboxContext;
 import io.agentscope.harness.agent.sandbox.SandboxManager;
+import io.agentscope.harness.agent.sandbox.SandboxMirrorReleaseCoordinator;
 import java.util.function.Consumer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -39,10 +40,12 @@ import org.slf4j.LoggerFactory;
  *
  * <h2>doFinally</h2>
  * <ol>
+ *   <li>Clear this call's session binding from the {@link RuntimeContext} (and the filesystem
+ *       proxy fallback) so concurrent calls cannot observe a stale binding</li>
  *   <li>Persist sandbox session state via {@link SandboxManager} and
  *       {@link io.agentscope.harness.agent.sandbox.SessionSandboxStateStore}</li>
- *   <li>Release the session via {@link SandboxManager} (stop + optional shutdown)</li>
- *   <li>Clear this call's session binding from the {@link RuntimeContext}</li>
+ *   <li>Request release via {@link SandboxMirrorReleaseCoordinator} (defers stop/shutdown while
+ *       session mirrors still need the connection; otherwise releases immediately)</li>
  * </ol>
  *
  * <p>Post-call failures (persist, release) are logged but do not propagate — this ensures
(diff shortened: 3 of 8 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTree.java
```diff
@@ -608,6 +647,17 @@ private static AbstractFilesystem pinIfSandbox(AbstractFilesystem fs) {
         return fs;
     }
 
+    /**
+     * Returns the sandbox pinned for an async mirror task, or {@code null} when the mirror does
+     * not hold a sandbox connection that must defer self-managed release.
+     */
+    private static Sandbox sandboxForMirrorRetain(AbstractFilesystem mirrorFs) {
```
oss-maintainer, line 673: `sandboxForMirrorRetain(mirrorFs)` derives the sandbox from `pinIfSandbox`, which reads `SandboxAware.getSandbox()` — the legacy **shared** `sandbox` field of `SandboxBackedFilesystem`. Issue #2490 established that the authoritative per-call binding is `SandboxAcquireResult` on the `RuntimeContext`: `requireSandbox(runtimeContext)` prefers it, `releaseForCall` nulls it, and `filesystemProxy.setSandbox`/`clearSandboxIfCurrent` keep the field only as a best-effort fallback. So with two concurrent distinct-session calls on one agent bean, `retain()` can be applied to the sibling's sandbox, which means the call that actually owns the pending mirror sees `pendingMirrors == 0` in `doRequestRelease` and stops/shuts down the sandbox mid-upload — the NPE this PR fixes stays reachable under exactly the concurrency the deferral was added for, and the unrelated sandbox is held open instead. (Mirroring bytes into the wrong sandbox follows from the same fallback field and predates this PR, but the reference count now inherits the ambiguity.) Please resolve the sandbox for retention from `fsRc` first, with `getSandbox()` only as the context-free fallback, and cover it with a two-session concurrency test.

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxMirrorReleaseCoordinator.java
```diff
@@ -0,0 +1,280 @@
...
+    private static final ExecutorService RELEASE_EXECUTOR =
+            Executors.newSingleThreadExecutor(
+                    r -> {
+                        Thread t = new Thread(r, "sandbox-mirror-release");
+                        t.setDaemon(true);
+                        return t;
+                    });
+
+    private final Object lock = new Object();
+    private final IdentityHashMap<Sandbox, State> states = new IdentityHashMap<>();
```
oss-maintainer, line 69: The process-wide `IdentityHashMap<Sandbox, State>` only shrinks in `doReleaseMirror` (count hits 0 or goes negative) and in `doRequestRelease`. A `retain()` whose mirror task never runs — `MIRROR_EXECUTOR` rejecting/shutting down while a task is queued, or a task killed before its `finally` — pins both the entry and the deferred `SandboxAcquireResult` (lease, pod/claim) forever, and `awaitReleaseQuiescence` does not observe it because nothing was ever submitted to `RELEASE_EXECUTOR`. Consider stamping each registration and force-releasing (with an `error` log) any deferred release older than a bounded grace period, so a lost mirror task cannot leak sandboxes.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
```diff
@@ -126,14 +129,14 @@ public void acquireForCall(RuntimeContext ctx) {
                 ctx.put(SandboxAcquireResult.class, null);
                 filesystemProxy.clearSandboxIfCurrent(sandbox);
                 try {
-                    sandboxManager.release(result);
+                    // No mirrors can be pending before start succeeds; release immediately.
+                    SandboxMirrorReleaseCoordinator.requestRelease(sandboxManager, result);
```
oss-maintainer, line 135: Moving the destructive release plus lease close behind the coordinator looks correct for the two call sites here, and the duplicate-request branch closes the extra lease explicitly. Could you confirm `getLease()` is non-null on every path that reaches `requestRelease` — in particular `SandboxAcquireResult.userManaged(external)`, where `SandboxManager.release` is a no-op but the lease is still closed, which previously happened in the middleware too, so behaviour should be unchanged; a one-line note in the class javadoc would save the next reader that check.

oss-maintainer (changes requested):
### Summary
Defers self-managed sandbox `stop`/`shutdown` (and the lease close) until in-flight session mirrors drain, via a new process-wide `SandboxMirrorReleaseCoordinator`, plus soft-fail when the bound sandbox has already stopped. The root cause reads correctly and the coordinator's pairing/underflow handling is careful.

### Verdict
One blocking concern: the sandbox that gets retained is resolved from the shared `SandboxAware.getSandbox()` field instead of the per-call `RuntimeContext` binding, which is precisely the race #2490 fixed — under concurrent distinct sessions the deferral protects the wrong sandbox and the fix is a no-op for the reported scenario.

---

xy-ygz:
> ### Summary
> Defers self-managed sandbox `stop`/`shutdown` (and the lease close) until in-flight session mirrors drain, via a new process-wide `SandboxMirrorReleaseCoordinator`, plus soft-fail when the bound sandbox has already stopped. The root cause reads correctly and the coordinator's pairing/underflow handling is careful.通过一个新的进程级机制，将自我管理的沙箱 `stop` / `shutdown` 的相关操作（包括租约的终止）推迟到当前正在进行的会话处理完成后再执行。如果被管理的沙箱已经停止了运行，系统会采取软故障处理方式。该方案的根源处理方式是正确的，协调器在处理配对或流量不足的情况时也相当谨慎。
> 
> ### Verdict
> One blocking concern: the sandbox that gets retained is resolved from the shared `SandboxAware.getSandbox()` field instead of the per-call `RuntimeContext` binding, which is precisely the race #2490 fixed — under concurrent distinct sessions the deferral protects the wrong sandbox and the fix is a no-op for the reported scenario.一个需要解决的阻塞问题在于：被保留的“沙箱”实际上是从共享的 `SandboxAware.getSandbox()` 字段中获取的，而不是从每次调用时单独绑定的 `RuntimeContext` 字段中获取的。这正是之前所解决的“竞态条件 #2490 ”问题。在多个并发会话的情况下，这种处理方式会错误地保护了错误的“沙箱”；而对于所描述的场景来说，这种修复措施实际上没有任何效果。
> 

Thanks for the review — agreed on the blocking #2490 concern.

**Per-call pin/retain (blocking)**  
Fixed: `pinIfSandbox` now resolves the sandbox from `SessionTree.fsRc`'s `SandboxAcquireResult` first, and only falls back to the shared `SandboxAware.getSandbox()` field for context-free callers. Added `flush_retainsPerCallSandboxFromFsRc_notSharedProxyField` to cover the concurrent distinct-session case (shared proxy field points at B while `fsRc` binds A → retain/defer A, immediate release B).

**Lease non-null**  
Confirmed: `SandboxAcquireResult` always supplies a non-null lease (`SandboxLease.noop()` when no guard), including `userManaged`. Documented on `SandboxMirrorReleaseCoordinator` and `SandboxLifecycleMiddleware`.

**Stale retain / TTL force-release**  
`submitPinnedMirror` already rolls back `retain` if `MIRROR_EXECUTOR.execute` fails. A task that never reaches `finally` (executor shutdown / thread kill) is not covered in this PR — the mirror executor is a process-lifetime single-thread daemon in normal operation. Happy to track a bounded grace-period sweeper as a follow-up if you want it in-tree before merge; otherwise we'll keep it as a known residual.

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxMirrorReleaseCoordinator.java
```diff
@@ -0,0 +1,285 @@
...
+            } else {
+                if (state != null) {
+                    states.remove(sandbox);
+                }
+                releaseNow = true;
+            }
+        }
+        if (releaseNow) {
+            // No outstanding mirrors: keep historical sync teardown on the call thread.
+            performRelease(manager, result);
```
oss-maintainer, line 245: This is the one branch I would still like pinned down. If a *late* `retain()` (line ~169) lands after call teardown already took the `releaseNow` path here, the map is left with `pendingMirrors == 1, deferred == null`. The next `releaseMirror()` then hits the `pendingMirrors == 0` branch, removes the entry and schedules nothing — so the self-managed sandbox is never stopped and its lease is never closed. `submitPinnedMirror` looks like it prevents this today (retain is the first statement inside a synchronous method), but the coordinator itself has no guard. Consider recording a `released` marker (or dropping the deferred result on the floor with a warn) so a future caller that retains after teardown cannot leak a running sandbox, and/or assert-and-log when `releaseMirror` finds `deferred == null` while a release for that sandbox has already been requested.

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxMirrorReleaseCoordinator.java
```diff
@@ -0,0 +1,285 @@
...
+                    try {
+                        result.getLease().close();
+                    } catch (Exception e) {
+                        log.warn(
+                                "[sandbox-mirror-release] Extra lease close failed: {}",
+                                e.getMessage(),
+                                e);
+                    }
+                    return;
+                }
```
oss-maintainer, line 232: For the "duplicate release request while mirrors are pending" case the comment says the *first* deferred result keeps ownership of the destructive release. That is right for a single-call lease, but if a sandbox is shared by two concurrent calls that both reached teardown (same instance, two `SandboxAcquireResult`s), the second call's `SandboxManager.release(result)` never runs at all — only the first result is released. If that shape cannot happen (one self-managed sandbox per call), it would help to say so in the class javadoc; if it can, the deferred slot should hold a list of pending results so each call's release/lease-close is executed.

agentscope-harness/src/main/java/io/agentscope/harness/agent/filesystem/sandbox/SandboxBackedFilesystem.java
```diff
@@ -122,7 +130,14 @@ public ExecuteResponse execute(
     @Override
     public List<FileUploadResponse> uploadFiles(
             RuntimeContext runtimeContext, List<Map.Entry<String, byte[]>> files) {
-        Sandbox active = requireSandbox(runtimeContext);
+        Sandbox active = requireRunningSandbox(runtimeContext);
```
oss-maintainer, line 133: `requireRunningSandbox` turns "sandbox present but not running" into a warn + per-file `fail(...)` response instead of the exception `requireSandbox` throws. That is exactly what the async mirrors need, but it also changes behaviour for *synchronous* tool calls that happen to hit a stopped sandbox: they now silently receive failed upload/download responses instead of an exception. Could you confirm that mirrors are the only consumers relying on this, or (safer) route the soft-fail through an explicit "mirror/best-effort" flag on the pinned filesystem so a sync-path regression still surfaces loudly?

oss-maintainer (commented):
### Summary
Re-review after the new commits (`75abefb1`): the blocking concern from my previous review is **resolved**. `SandboxMirrorReleaseCoordinator.requestRelease(manager, result)` now takes the per-call `SandboxAcquireResult` from `RuntimeContext` instead of resolving the sandbox from the shared `SandboxAware.getSandbox()` field, so the deferral protects the correct sandbox and the #2490 per-call-binding invariant is preserved (the middleware javadoc also documents this now). `submitPinnedMirror` closes the retain/release pairing and moves prepare-side failures before `retain`, and `awaitMirrorQuiescence` drains both the mirror executor and the new release executor under one shared deadline. This looks good to me — the remaining notes are hardening/documentation, not blockers.

**Verdict:** COMMENT (approve-equivalent). CLA is signed ✅; I am leaving the previous `CHANGES_REQUESTED` state for a maintainer to lift, since the follow-up commits addressed it.

Findings:
- **[Warning]** `SandboxMirrorReleaseCoordinator.java:245` — a `retain()` landing after teardown took the `releaseNow` path leaves `pendingMirrors=1, deferred=null`, and the next `releaseMirror()` drops the entry without releasing (potential leaked running sandbox). No guard inside the coordinator today.
- **[Warning]** `SandboxMirrorReleaseCoordinator.java:232` — duplicate release requests keep only the *first* deferred result; please confirm two concurrent calls can never share one self-managed sandbox instance.
- **[Info]** `SandboxBackedFilesystem.java:133` — `requireRunningSandbox` soft-fails for synchronous callers too, not just async mirrors.

---

xy-ygz:
> ### Summary
> Re-review after the new commits (`75abefb1`): the blocking concern from my previous review is **resolved**. `SandboxMirrorReleaseCoordinator.requestRelease(manager, result)` now takes the per-call `SandboxAcquireResult` from `RuntimeContext` instead of resolving the sandbox from the shared `SandboxAware.getSandbox()` field, so the deferral protects the correct sandbox and the #2490 per-call-binding invariant is preserved (the middleware javadoc also documents this now). `submitPinnedMirror` closes the retain/release pairing and moves prepare-side failures before `retain`, and `awaitMirrorQuiescence` drains both the mirror executor and the new release executor under one shared deadline. This looks good to me — the remaining notes are hardening/documentation, not blockers.在新提交之后再次进行审查：我之前指出的那个问题已经得到解决。现在， `SandboxMirrorReleaseCoordinator.requestRelease(manager, result)` 会从 `RuntimeContext` 那里获取每次调用所需的 `SandboxAcquireResult` 数据，而不是从共享的 `SandboxAware.getSandbox()` 字段中获取这些数据。这样一来，延迟处理机制就能确保正确的沙箱环境得到维护，同时 #2490 的每次调用绑定规则也得以保持不变（中间件的文档中也对此进行了说明）。 `submitPinnedMirror` 负责处理保留/释放的操作顺序，将准备阶段的错误处理放在 `retain` 之前处理。 `awaitMirrorQuiescence` 则确保镜像执行器和新版本执行器都在同一个截止时间内完成工作。我觉得这样很好——剩下的问题都只是与代码的完善或文档记录相关，不会影响系统的正常运行。
> 
> **Verdict:** COMMENT (approve-equivalent). CLA is signed ✅; I am leaving the previous `CHANGES_REQUESTED` state for a maintainer to lift, since the follow-up commits addressed it.结论：评论已通过。CLA 协议已签署✅；我会把之前的 `CHANGES_REQUESTED` 状态保留下来，等待维护人员来处理，因为后续的提交已经解决了该问题。
> 
> Findings:  研究结果：
> 
> * **[Warning]** `SandboxMirrorReleaseCoordinator.java:245` — a `retain()` landing after teardown took the `releaseNow` path leaves `pendingMirrors=1, deferred=null`, and the next `releaseMirror()` drops the entry without releasing (potential leaked running sandbox). No guard inside the coordinator today.[警告] `SandboxMirrorReleaseCoordinator.java:245` ——在解构操作之后， `retain()` 通过 `releaseNow` 路径进行着陆，但最终离开了 `pendingMirrors=1, deferred=null` 。接下来的 `releaseMirror()` 操作没有正确释放该入口点（可能导致运行中的沙箱被泄露）。今天，协调器内部没有任何保护机制。
> * **[Warning]** `SandboxMirrorReleaseCoordinator.java:232` — duplicate release requests keep only the _first_ deferred result; please confirm two concurrent calls can never share one self-managed sandbox instance.[警告] `SandboxMirrorReleaseCoordinator.java:232` —— 对于重复的发布请求，系统只保留第一个延迟处理的结果。请注意，两个同时进行的请求绝不能共享同一个自我管理的沙箱实例。
> * **[Info]** `SandboxBackedFilesystem.java:133` — `requireRunningSandbox` soft-fails for synchronous callers too, not just async mirrors.[信息] `SandboxBackedFilesystem.java:133` — `requireRunningSandbox` 的软故障也适用于同步调用者，而不仅仅是异步调用者。
> 

addressed the three hardening notes:

1. Late retain after releaseNow
Added a released tombstone. If a late retain shows up after teardown already released, we log it, keep retain/releaseMirror pairing clean, and do not stop the sandbox a second time (it was already released).

2. Duplicate requestRelease on a shared sandbox
Deferred releases are now a list: each call’s release + lease close runs when mirrors drain. Usual case remains one-call-one-sandbox; shared-sandbox is covered.

3. Soft-fail scope
Soft-fail only on PinnedSandboxFilesystem (async mirrors). Sync SandboxBackedFilesystem tool calls again throw when the sandbox is already stopped.

Tests added/updated for all three. Happy to adjust further if anything still looks off.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/SandboxLifecycleMiddleware.java
```diff
@@ -169,10 +174,11 @@ public void releaseForCall(RuntimeContext ctx) {
             log.warn("[sandbox-mw] Failed to persist sandbox state: {}", e.getMessage(), e);
         }
         try {
-            sandboxManager.release(result);
+            // Hand destructive stop/shutdown (and lease close) to outstanding session mirrors
+            // when present; otherwise release immediately. Call binding is already cleared above.
+            SandboxMirrorReleaseCoordinator.requestRelease(sandboxManager, result);
```
oss-maintainer, line 179: **[Warning]** Deferred-release race worth confirming: when call A's release is deferred behind outstanding mirrors, a call B for the same `(userId, sessionId)` that starts before the deferred `stop()`/`shutdown()` executes can re-acquire the very same self-managed sandbox instance (via persisted session state) — and then A's pending `performRelease` shuts it down under B, reproducing the original NPE symptom from the other direction. The window is small (requires B to acquire between `requestRelease` deferral and the executor actually running the release), but a targeted regression test for "new call re-acquires while release is deferred" would close the doubt. Alternatively, have `performRelease` skip destructive stop/shutdown when a live acquire for the same sandbox/session exists.

agentscope-harness/src/main/java/io/agentscope/harness/agent/sandbox/SandboxMirrorReleaseCoordinator.java
```diff
@@ -0,0 +1,350 @@
...
+            if (state == null) {
+                throw new IllegalStateException("expected state for " + sandbox);
+            }
+            state.pendingMirrors = 0;
+        }
+    }
+
+    private void doRetain(Sandbox sandbox) {
+        synchronized (lock) {
+            State state = states.computeIfAbsent(sandbox, ignored -> new State());
```
oss-maintainer, line 180: **[Info]** `doRetain` unconditionally creates a `State` entry (`computeIfAbsent`), and a "late retain after released" only logs an error — the entry is removed again when the paired `releaseMirror` brings the count to zero, so the tombstone path looks GC-safe. One residual case: if a mirror task is discarded without ever running its `finally` (e.g. executor shutdown drops queued tasks during JVM teardown), the `states` entry and its `Sandbox` key leak. Since `Sandbox` keys are strong references in an `IdentityHashMap`, consider a periodic sweep or a max-age on non-pending registrations if long-lived JVMs with repeated sandbox churn are a real deployment shape here.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTree.java
```diff
@@ -112,9 +114,15 @@ public class SessionTree {
      * @return {@code true} if the mirrors quiesced within the timeout
      */
     public static boolean awaitMirrorQuiescence(long timeout, TimeUnit unit) {
```
oss-maintainer, line 116: **[Info]** Nice fix draining the release executor within a shared deadline. Two small notes: (1) `awaitReleaseQuiescence` submits an empty task, so it only waits for releases *enqueued before* the barrier call — releases scheduled by mirror tasks that begin executing after the submit are not covered; that matches the documented "submitted before this call" contract, just make sure graceful-shutdown callers invoke it after `awaitMirrorQuiescence` returns true (which the current call ordering does). (2) The process-wide singleton coordinator means multi-agent JVMs share one release thread — fine for teardown throughput, but a slow `stop()` head-of-line blocks other sessions' deferred releases; acceptable given it is off the mirror path.

oss-maintainer (approved):
### Summary
Re-review after the force-push (new head `e243a67f`). The deferred-release design is sound: refcount + tombstone pairing, per-call `SandboxAcquireResult` pinning (fixes the last-writer-wins shared-slot hazard from #2490), lease always closed, soft-fail scoped to pinned async mirrors while sync tool paths keep hard-fail, and both executors drained within one shared deadline in `awaitMirrorQuiescence`. Test coverage is thorough (deferral ordering, underflow clamping, per-call pinning, hard-fail vs soft-fail matrix). CLA signed.

**Verdict: APPROVED** — one low-probability race worth a follow-up test (see inline on `releaseForCall`), plus two non-blocking notes.

---

oss-maintainer:

This PR currently conflicts with `main` and cannot be merged (head `feature_fix_sandbox_async_npe`). Please rebase (or merge `main` in) and resolve the conflicts:

```bash
git fetch upstream
git checkout feature_fix_sandbox_async_npe
git rebase upstream/main
# resolve conflicts, then:
git push --force-with-lease
```

---

PR #3099, opened 2026-09-11
fix: emit permission-denied tool result events

### LINKED ISSUE #3097: Permission HITL: user-denied tool calls emit no tool-result events (rule-denied ones do)[Bug]:
**Version:** agentscope-java 2.0.3 (agentscope-core)

### What happens

When the user denies a tool call in permission-mode HITL, the resumed run emits no
`ToolResultStartEvent` / `ToolResultEndEvent` for it. A client driving the AG-UI protocol therefore
never learns the outcome of that tool call and has to synthesise the state from its own local record
of the decision.

Rule-based denial behaves the opposite way and does emit the events, so the two denial paths are
inconsistent.

### Where

- `ReActAgent.PermissionFlow#applyConfirmResults` — for a denied `ConfirmResult` it only appends a
  `ToolResultBlock.text("Permission denied by user").withState(DENIED)` to the message context.
  No `publishEvent(...)`.
- `ReActAgent#runToolBatch` — rule-denied calls *do* publish `ToolResultStartEvent` with
  `"Permission denied by rules"` and a matching end event.

### Why it matters

The AG-UI mapping is event-driven. Without an event, `AguiStreamContext` produces no
`TOOL_CALL_RESULT`, so a conforming client cannot distinguish "denied" from "result never arrived"
without keeping its own side-channel state. It also means the denial is invisible to anything else
that observes the agent event stream (tracing, audit, replay).

### Suggested fix

Publish the same `ToolResultStartEvent` → delta → `ToolResultEndEvent(state = DENIED)` sequence from
the user-denial path that `runToolBatch` already publishes for rule-denied calls.

Possibly the same root cause as #2492.

guslegend0510:
Fixes #3097

User-denied permission confirmations previously updated the message context without emitting tool-result events, leaving event-driven clients without a tool outcome.

This change:
- Emits the complete start → text delta → end (DENIED) sequence for user denials.
- Emits rule-denied results in batches that also contain tools awaiting confirmation.
- Restores validated tool-call tracking on AG-UI resume without replaying call start/end events.
- Supports older sessions without persisted confirmation reply IDs.

Validation:
- Core: 2,338 tests, 9 skipped.
- Harness: 1,030 tests, 14 skipped.
- AG-UI: 413 tests.
- Zero failures or errors; formatting and diff checks passed.
-

### DIFF
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -52,10 +52,8 @@ jobs:
             repo-cache-path: '~/.m2/repository'
           - os: windows-latest
             repo-cache-path: '~\.m2\repository'
-            mvnd-cache-path: '~\.mvnd'
 
     env:
-      MVND_VERSION: '1.0.3'
       JAVA_VERSION: '17'
 
     steps:
@@ -75,32 +73,6 @@ jobs:
           restore-keys: |
             maven-${{ runner.os }}-
 
-      - name: Cache mvnd [${{ runner.os }}]
-        if: runner.os == 'Windows'
-        id: cache-mvnd
-        uses: actions/cache@v4
-        with:
-          path: ${{ matrix.mvnd-cache-path }}
-          key: mvnd-${{ env.MVND_VERSION }}-${{ runner.os }}
-
-      - name: Install mvnd [Windows]
-        if: steps.cache-mvnd.outputs.cache-hit != 'true' && runner.os == 'Windows'
-        shell: pwsh
-        run: |
-          Invoke-WebRequest `
-            -Uri "https://github.com/apache/maven-mvnd/releases/download/${{ env.MVND_VERSION }}/maven-mvnd-${{ env.MVND_VERSION }}-windows-amd64.zip" `
-            -OutFile "mvnd.zip"
-          mkdir -Path "~\.mvnd" -Force
-          Expand-Archive -Path "mvnd.zip" -DestinationPath "mvnd" -FORCE
-          Copy-Item -Path "mvnd\maven-mvnd-${{ env.MVND_VERSION }}-windows-amd64\*" -Destination "~\.mvnd\" -Recurse
-          Set-Content -Path "~\.mvnd\conf\mvnd.properties" -Value @"
-          # mvnd client timeout settings
-          mvnd.connectTimeout=120000
-          mvnd.socketConnectTimeout=30000
-          mvnd.cancelConnectTimeout=30000
-          "@ -Encoding Utf8 -Force
-          Remove-Item -Path "mvnd", "mvnd.zip" -Recurse -Force
-
       - name: Build and Test with Coverage [Linux]
         if: runner.os == 'Linux'
         shell: bash
@@ -114,8 +86,8 @@ jobs:
         if: runner.os == 'Windows'
         run: |
           # -T1: single-threaded build so reactor order is respected (see Linux step).
-          # -Dmvnd.maxLostKeepAlive=120: avoid StaleAddressException on long javadoc phases (mvnd#161).
-          & "~\.mvnd\bin\mvnd.cmd" -B -T1 "-Dmvnd.maxLostKeepAlive=120" clean verify
+          # Use standard Maven to avoid mvnd daemon startup/communication failures.
+          mvn -B -T1 clean verify
 
       - name: Upload coverage reports to Codecov
         if: runner.os == 'Linux'
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -1908,12 +1908,14 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
             }
 
             String replyId = resolvePendingRequestReplyId(Msg.METADATA_CONFIRM_REQUEST_REPLY_ID);
-            if (!replyId.isEmpty()) {
-                publishEvent(new UserConfirmResultEvent(replyId, normalized));
-                clearPendingRequestReplyId(Msg.METADATA_CONFIRM_REQUEST_REPLY_ID);
+            if (replyId.isEmpty()) {
+                // Older persisted ASKING calls may not carry request correlation metadata.
+                replyId = UUID.randomUUID().toString().replace("-", "");
             }
+            publishEvent(new UserConfirmResultEvent(replyId, normalized));
+            clearPendingRequestReplyId(Msg.METADATA_CONFIRM_REQUEST_REPLY_ID);
 
-            applyConfirmResults(normalized);
+            applyConfirmResults(normalized, replyId);
         }
 
         /** Resolve the reply id for the pending HITL request stored on the last assistant message. */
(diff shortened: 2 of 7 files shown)


oss-maintainer:

@guslegend0510 This PR currently conflicts with `main` and cannot be merged. Please rebase or merge `main` into your branch:

```bash
git fetch origin
git checkout codex/fix-3097-denied-tool-events
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

This is a one-time reminder. Feel free to @mention me for a re-review once conflicts are resolved.

---

.github/workflows/maven-ci.yml
```diff
@@ -114,8 +86,8 @@ jobs:
         if: runner.os == 'Windows'
         run: |
           # -T1: single-threaded build so reactor order is respected (see Linux step).
-          # -Dmvnd.maxLostKeepAlive=120: avoid StaleAddressException on long javadoc phases (mvnd#161).
-          & "~\.mvnd\bin\mvnd.cmd" -B -T1 "-Dmvnd.maxLostKeepAlive=120" clean verify
+          # Use standard Maven to avoid mvnd daemon startup/communication failures.
```
oss-maintainer, line 89: Two things here: (1) this CI change is unrelated to the PR's stated purpose (permission-denied tool result events) — it belongs in its own PR so it can be reverted independently; (2) dropping mvnd on Windows also drops the workaround the removed comment described (`-Dmvnd.maxLostKeepAlive=120` for `StaleAddressException` during long javadoc phases, apache/maven-mvnd#161). Falling back to plain `mvn -B -T1 clean verify` is a legitimate choice, but please confirm Windows CI is green with it and say so in the description, otherwise the next long javadoc build may flake and the cause will be hard to trace back to this PR.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1908,12 +1908,14 @@ private void validateAndAcceptConfirmResults(List<Msg> msgs, List<ToolUseBlock>
             }
 
             String replyId = resolvePendingRequestReplyId(Msg.METADATA_CONFIRM_REQUEST_REPLY_ID);
-            if (!replyId.isEmpty()) {
-                publishEvent(new UserConfirmResultEvent(replyId, normalized));
-                clearPendingRequestReplyId(Msg.METADATA_CONFIRM_REQUEST_REPLY_ID);
+            if (replyId.isEmpty()) {
+                // Older persisted ASKING calls may not carry request correlation metadata.
+                replyId = UUID.randomUUID().toString().replace("-", "");
```
oss-maintainer, line 1913: Inventing a fresh `replyId` when the persisted `METADATA_CONFIRM_REQUEST_REPLY_ID` is absent breaks the correlation this PR is trying to restore: the client saw `RequireUserConfirmEvent` under the *original* reply id, so the `ToolResult*` events (and the `ToolCallResult` messageId keyed as `replyId + ":" + toolCallId` in the AG-UI adapter) will land under an id the client has never seen — an orphan result rather than a resolved pending call. It also makes `UserConfirmResultEvent` publish unconditionally now, where before it was suppressed for the blank case. Could the blank-metadata case instead recover the id from the last assistant message (the same place `persistPendingRequestReplyId` writes it), or fall back to the previous no-op behaviour plus a `log.warn`? A test for the metadata-missing resume path would pin the intent either way.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2919,11 +2923,21 @@ Flux<AgentEvent> actingStream(
                                 resultHolder.set(List.of());
                                 persistPendingRequestReplyId(
                                         Msg.METADATA_CONFIRM_REQUEST_REPLY_ID, replyId);
-                                return Flux.<AgentEvent>just(
-                                        new RequireUserConfirmEvent(replyId, pending),
-                                        new RequestStopEvent(
-                                                "permission asking",
-                                                GenerateReason.PERMISSION_ASKING));
+                                return Flux.fromIterable(toolCalls)
+                                        .filter(tc -> autoDenied.contains(tc.getId()))
```
oss-maintainer, line 2927: Emission ordering detail: the auto-denied `ToolResult*` triples are concatenated *before* `RequireUserConfirmEvent` / `RequestStopEvent`. A consumer that opens a tool-call card on `RequireUserConfirmEvent` will therefore receive results for tool calls it has not been told are pending yet. Emitting the denied triples after the confirm/stop pair (or right after `writeAutoDeniedResults`) reads more naturally; either is fine as long as it is deliberate.

agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/strategy/AguiStreamContext.java
```diff
@@ -232,6 +232,14 @@ public void endToolCall(String toolCallId) {
         }
     }
 
+    /** Recognize a validated permission resume without replaying the previous run's call events. */
```
oss-maintainer, line 235: `resumeToolCall` overlaps with what is already on `main`: `AguiStreamContext.adoptToolCall(toolCallId)` plus the `hasKnownToolCall(...)` guard on `beginToolResult` / `appendToolResultText` / `endToolResult`, and `PermissionConfirmEventConverter` already calls `adoptToolCall(...)` for every validated `ConfirmResult`. This is very likely the source of the current merge conflict. Please rebase onto `main` and drop the `AguiStreamContext` / `PermissionConfirmEventConverter` hunks — the core-side fix plus the existing adoption path should be sufficient. Note the difference in semantics too: `resumeToolCall` writes both `startedToolCalls` and `endedToolCalls` (which suppresses a later `ToolCallEnd` for a call that is re-executed in this run), whereas `adoptToolCall` keeps those sets separate.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentHitlTest.java
```diff
@@ -408,6 +416,186 @@ void askingToolResumeWithDeniedConfirmResultProducesDeniedToolResult() {
         assertTrue(foundDenied, "expected a DENIED ToolResultBlock for the rejected tool");
     }
 
+    @ParameterizedTest
```
oss-maintainer, line 419: The parameterized user-denied / rule-denied coverage and `assertDeniedEvents` are a good fit for this change, and promoting `AskingTool` to assert execution counts is a nice way to prove the denied path never invokes the tool. The one missing case is the blank-`replyId` resume discussed above, which is exactly the branch whose behaviour changed.

oss-maintainer (commented):
### Summary
Adds a shared `deniedToolResultEvents(...)` helper and routes every permission-denial path (user confirm, auto-deny during acting, rule deny in `runToolBatch`) through it, with parameterized tests — a genuinely more complete treatment of the denied-event gap than the current state on `main`. COMMENT because three things need resolution first: this PR currently has merge conflicts, the AG-UI half duplicates machinery that already landed, and the synthesized `replyId` changes correlation semantics.

### Findings
- **[Critical]** `ReActAgent.java:1913` — a random `replyId` for the metadata-missing case produces results keyed to an id the client never saw; also now publishes `UserConfirmResultEvent` unconditionally.
- **[Warning]** `AguiStreamContext.java:235` — `resumeToolCall` overlaps with `adoptToolCall` + `hasKnownToolCall` + the existing `PermissionConfirmEventConverter` adoption on `main`; likely the conflict source.
- **[Warning]** `maven-ci.yml:89` — unrelated Windows mvnd→mvn switch that also removes the documented `mvnd#161` workaround.
- **[Info]** `ReActAgent.java:2927` — denied triples are emitted before `RequireUserConfirmEvent`.
- **[Info]** test — blank-`replyId` branch not covered.

### Cross-PR Note
PR #3104 covers the same user-denied gap with a narrower change. Suggest the maintainers pick one home for the shared helper: this PR's `deniedToolResultEvents(...)` factoring (which also fixes the auto-denied path) looks like the better base once the conflict is resolved and the AG-UI hunks are dropped.

---

PR #3100, opened 2026-09-11
fix(agui): preserve tool results after permission resume

### LINKED ISSUE #3096: [Bug]:AG-UI: tool results from a resumed run are dropped because the stream context is per-run
**Version:** agentscope-java 2.0.3 (agentscope-extensions-agui)

### What happens

Permission-mode HITL spans two runs. Run 1 emits `TOOL_CALL_START/ARGS/END` and finishes with an
interrupt. After the user approves, run 2 executes the tool and produces `ToolResultStartEvent` →
delta events → `ToolResultEndEvent`.

Every AG-UI event from run 2 for that tool call is silently discarded, so the client never receives
`TOOL_CALL_RESULT` and the tool card stays unresolved forever.

### Why

1. `AguiAgentAdapter#streamWithRuntimeContext` creates a **new** `AguiStreamContext` per run (both the
   ReActAgent and HarnessAgent branches). Nothing carries over from the previous run.
2. `ReActAgent#resumeAgent()` is `return acting(0)` — it does not replay `ToolCallStart/Delta/End` for
   the tool call that was paused, so the new context never records the `toolCallId`.
3. `AguiStreamContext#beginToolResult` / `appendToolResultText` / `appendToolResultData` /
   `endToolResult` all guard on the private `startedToolCalls` set via `hasStartedToolCall(...)`.
   The id is absent, so the whole result is dropped.

There is no public API to query or seed `startedToolCalls`, so an application-side converter cannot
tell "this is a cross-run result" from "this is a bogus id" either.

### What the AG-UI spec expects

A resumed run should emit `TOOL_CALL_RESULT` for the original `toolCallId` without replaying
`TOOL_CALL_START/ARGS/END`. The client correlates by `toolCallId`. (`@ag-ui/client`'s `verifyEvents`
does not require a preceding `TOOL_CALL_START` for `TOOL_CALL_RESULT`, so the standard shape is
accepted by the official client.)

### Current workaround

A custom `AgentEventConverter` overriding `ToolResultStartEvent` that makes the context adopt the id
silently: call `startToolCall(id, name)` + `endToolCall(id)`, then `beginEvent()` to discard the
`TOOL_CALL_START`/`TOOL_CALL_END` those two calls queued (the registry calls `beginEvent()` before and
`drainEvents()` after each event, so only the current event's output is cleared), then
`beginToolResult(id)`. This produces the spec-compliant single `TOOL_CALL_RESULT`, but it leans on
internal ordering guarantees and is fragile across upgrades.

### Suggested fix

Either

- let a resumed run map tool results for tool calls belonging to a known resume interrupt (the
  interrupt id already carries `replyId:toolCallId`), or
- expose a public entry point on `AguiStreamContext` to adopt a tool call id without emitting events,
  e.g. `adoptToolCall(String toolCallId, String toolCallName)`.

Related: #2437 (interrupt not propagated, fixed by #2495), #2492, #2956. PR #1231 shows the previous
adapter could back-fill a missing tool-call; that capability was lost in the converter-based design.

guslegend0510:
Fixes #3096

After permission approval, a resumed run executes the pending tool but drops its result because the new AguiStreamContext does not recognize the previous run's toolCallId.

Adopt validated tool-call IDs when handling UserConfirmResultEvent, allowing the resumed run to emit TOOL_CALL_RESULT without replaying TOOL_CALL_START/ARGS/END. Unknown IDs remain ignored, and adopted IDs stay local to the current stream context.

Validation:
- Added 7 regression cases covering result aggregation, suspension, context isolation, and real ReActAgent/HarnessAgent approval round trips.
- Full Core, AG-UI, and Harness suites: 3,875 tests, 0 failures, 0 errors, 23 skipped.

### DIFF
--- a/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/strategy/AguiStreamContext.java
+++ b/agentscope-extensions/agentscope-extensions-protocol/agentscope-extensions-agui/src/main/java/io/agentscope/core/agui/adapter/strategy/AguiStreamContext.java
@@ -53,6 +53,7 @@ public class AguiStreamContext {
     private final Set<String> endedReasoningMessages = new LinkedHashSet<>();
     private final Set<String> startedToolCalls = new LinkedHashSet<>();
     private final Set<String> endedToolCalls = new LinkedHashSet<>();
+    private final Set<String> adoptedToolCalls = new LinkedHashSet<>();
     private String currentTextMessageId;
     private String currentReasoningMessageId;
     private final Map<String, StringBuilder> toolResultContent = new LinkedHashMap<>();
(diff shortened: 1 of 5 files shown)


Buktal (approved):
LGTM

PR #3104, opened 2026-09-11
fix(core): emit tool result events for user-denied HITL calls

CryoThrust:
### What problem does this PR solve?

When a user denies a permission-mode HITL tool call, the resumed run writes a DENIED ToolResultBlock to context but emits no ToolResultStartEvent/ToolResultTextDeltaEvent/ToolResultEndEvent. This leaves AG-UI and other event-driven consumers unable to observe the outcome. Rule-denied calls already emit the corresponding sequence.

### What changed

- Carry the persisted HITL reply id into confirmation-result application.
- Emit the same start → text delta → end sequence for user-denied calls, with `ToolResultState.DENIED`.
- Add an end-to-end regression test covering correlation, ordering, tool-call id, and denied state.

### Verification

- `mvn -pl agentscope-core -DskipTests -Dspotless.check.skip=false spotless:check`
- `mvn -pl agentscope-core -Dtest=ReActAgentHitlTest test` (14 tests passed)

Signed-off-by: Yohanes <CryoThrust@users.noreply.github.com>

### DIFF
--- a/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
+++ b/agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
@@ -1913,7 +1913,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                 clearPendingRequestReplyId(Msg.METADATA_CONFIRM_REQUEST_REPLY_ID);
             }
 
-            applyConfirmResults(normalized);
+            applyConfirmResults(normalized, replyId);
         }
 
         /** Resolve the reply id for the pending HITL request stored on the last assistant message. */
@@ -1978,7 +1978,7 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
          *       the tool will no longer be pending on resume.</li>
          * </ul>
          */
-        private void applyConfirmResults(List<ConfirmResult> results) {
+        private void applyConfirmResults(List<ConfirmResult> results, String replyId) {
             // Replace ASKING ToolUseBlocks with possibly-modified ones from the user, and
             // promote them to ALLOWED. Collect denied ones for separate handling.
             List<ToolUseBlock> deniedToolCalls = new ArrayList<>();
@@ -2011,6 +2011,22 @@ public class ReActAgent extends AgentBase implements AutoCloseable {
                         ToolResultMessageBuilder.buildToolResultMsg(
                                 deniedResult, denied, getName());
                 state.contextMutable().add(deniedMsg);
+                if (replyId != null && !replyId.isEmpty()) {
+                    publishEvent(
+                            new ToolResultStartEvent(replyId, denied.getId(), denied.getName()));
+                    publishEvent(
+                            new ToolResultTextDeltaEvent(
+                                    replyId,
+                                    denied.getId(),
+                                    denied.getName(),
+                                    "Permission denied by user"));
+                    publishEvent(
+                            new ToolResultEndEvent(
+                                    replyId,
+                                    denied.getId(),
+                                    denied.getName(),
+                                    ToolResultState.DENIED));
+                }
             }
         }
 
--- a/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentHitlTest.java
+++ b/agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentHitlTest.java
@@ -26,6 +26,8 @@ import io.agentscope.core.event.ConfirmResult;
 import io.agentscope.core.event.RequestStopEvent;
 import io.agentscope.core.event.RequireUserConfirmEvent;
 import io.agentscope.core.event.ToolResultEndEvent;
+import io.agentscope.core.event.ToolResultStartEvent;
+import io.agentscope.core.event.ToolResultTextDeltaEvent;
 import io.agentscope.core.event.UserConfirmResultEvent;
 import io.agentscope.core.message.ContentBlock;
 import io.agentscope.core.message.GenerateReason;
@@ -328,6 +330,39 @@ class ReActAgentHitlTest {
         assertEquals(req.getReplyId(), confirm.getReplyId());
         assertEquals(1, confirm.getConfirmResults().size());
         assertEquals("tc1", confirm.getConfirmResults().get(0).getToolCall().getId());
+    }
+
+    @Test
+    void deniedResumeEmitsToolResultEventsCorrelatedToRequireEvent() {
+        ChatModelBase model =
+                new ScriptedModel(
+                        List.of(
+                                () -> Flux.just(toolUseResponse("tc1", "ask", "x")),
+                                () -> Flux.just(textResponse("done"))));
+        ReActAgent agent = buildAgent(model, toolkitWith(new AskingTool("ask")));
+
+        List<AgentEvent> pauseEvents = agent.streamEvents(List.of()).collectList().block();
+        assertNotNull(pauseEvents);
+        RequireUserConfirmEvent req =
+                (RequireUserConfirmEvent)
+                        pauseEvents.get(indexOf(pauseEvents, RequireUserConfirmEvent.class));
+
+        List<AgentEvent> resumeEvents =
+                agent.streamEvents(List.of(confirmMsg(false, req.getToolCalls().get(0))))
+                        .collectList()
+                        .block();
+        assertNotNull(resumeEvents);
+
+        int iStart = indexOf(resumeEvents, ToolResultStartEvent.class);
+        int iDelta = indexOf(resumeEvents, ToolResultTextDeltaEvent.class);
+        int iEnd = indexOf(resumeEvents, ToolResultEndEvent.class);
+        assertTrue(iStart >= 0, "denied resume must emit ToolResultStartEvent");
+        assertTrue(iDelta > iStart, "denied resume must emit a result delta");
+        assertTrue(iEnd > iDelta, "denied resume must emit ToolResultEndEvent");
+        ToolResultEndEvent end = (ToolResultEndEvent) resumeEvents.get(iEnd);
+        assertEquals(req.getReplyId(), end.getReplyId());
+        assertEquals("tc1", end.getToolCallId());
+        assertEquals(ToolResultState.DENIED, end.getState());
     }
 
     @Test


oss-maintainer (commented):
### Summary
Emitting the missing `ToolResultStartEvent` / `TextDeltaEvent` / `ToolResultEndEvent` triple for user-denied HITL tool calls is the right fix, it mirrors the existing rule-denied emission in `runToolBatch`, and it ships with a focused regression test. Left as COMMENT because two questions are open (blank-`replyId` no-op, and publishing events from a synchronous helper rather than the stream).

### Findings
- **[Warning]** `ReActAgent.java:2014` — the blank-`replyId` guard silently emits nothing, so the stuck-pending symptom survives for older persisted sessions.
- **[Warning]** `ReActAgent.java:1916` — events are published outside the reactive chain; please confirm subscriber ordering.
- **[Info]** `ReActAgent.java:2022` — duplicate literal for the denial reason.
- **[Info]** test — no coverage for the blank-replyId branch or the auto-denied path.

### Cross-PR Note
PR #3099 (also "emit permission-denied tool result events") overlaps heavily with this change and takes a different approach: it extracts a shared `deniedToolResultEvents(...)` helper, covers the auto-denied path, and mints a reply id when the correlation metadata is missing. Maintainers may want to land one of the two and fold the useful parts of the other in, rather than both.

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2011,6 +2011,22 @@ private void applyConfirmResults(List<ConfirmResult> results) {
                         ToolResultMessageBuilder.buildToolResultMsg(
                                 deniedResult, denied, getName());
                 state.contextMutable().add(deniedMsg);
+                if (replyId != null && !replyId.isEmpty()) {
```
oss-maintainer, line 2014: Skipping emission whenever `replyId` is blank leaves exactly the symptom this PR fixes: for an `ASKING` tool call persisted before the correlation metadata existed (or after `clearPendingRequestReplyId`), no `ToolResult*` events are published, so an AG-UI client still renders a permanently pending tool call. Could this path emit under a freshly minted reply id (as PR #3099 does) or at least log at warn level so the silent no-op is diagnosable?

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2011,6 +2011,22 @@ private void applyConfirmResults(List<ConfirmResult> results) {
...
+                    publishEvent(
+                            new ToolResultStartEvent(replyId, denied.getId(), denied.getName()));
+                    publishEvent(
+                            new ToolResultTextDeltaEvent(
+                                    replyId,
+                                    denied.getId(),
+                                    denied.getName(),
+                                    "Permission denied by user"));
```
oss-maintainer, line 2022: The reason string is now duplicated in two places (`ToolResultBlock.text(...)` above and this delta). If they ever drift, the message written to context and the one streamed to the client disagree. Worth extracting a `private static final String DENIED_BY_USER = "Permission denied by user";` and reusing it in both spots.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1913,7 +1913,7 @@ private void validateAndAcceptConfirmResults(List<Msg> msgs, List<ToolUseBlock>
                 clearPendingRequestReplyId(Msg.METADATA_CONFIRM_REQUEST_REPLY_ID);
             }
 
-            applyConfirmResults(normalized);
+            applyConfirmResults(normalized, replyId);
```
oss-maintainer, line 1916: `applyConfirmResults` is a synchronous state-mutation helper, so these events are pushed through `publishEvent` while the *resume* call is still validating input, i.e. outside the reactive chain that emits `RequireUserConfirmEvent` / `RequestStopEvent`. Please confirm the ordering guarantees: a subscriber that starts collecting after `validateAndAcceptConfirmResults` returns (or a replay from persisted state) may never observe the denied triple. Returning `Flux<AgentEvent>` from the helper — as `runToolBatch` does — would keep emission inside the stream.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentHitlTest.java
```diff
@@ -330,6 +332,39 @@ void confirmedResumeEmitsUserConfirmResultEventCorrelatedToRequireEvent() {
         assertEquals("tc1", confirm.getConfirmResults().get(0).getToolCall().getId());
     }
 
+    @Test
```
oss-maintainer, line 335: Nice regression test. Two gaps worth covering: (1) resume without `METADATA_CONFIRM_REQUEST_REPLY_ID` (the blank-replyId branch, which today emits nothing), and (2) the rule-denied path — `writeAutoDeniedResults` still writes DENIED blocks without publishing events, so an auto-denied tool has the same stuck-pending symptom this PR addresses for user denials.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2037,6 +2037,22 @@ private void applyConfirmResults(List<ConfirmResult> results) {
...
+                                    denied.getId(),
+                                    denied.getName(),
+                                    "Permission denied by user"));
+                    publishEvent(
+                            new ToolResultEndEvent(
+                                    replyId,
+                                    denied.getId(),
+                                    denied.getName(),
+                                    ToolResultState.DENIED));
+                }
```
oss-maintainer, line 2055: **[Performance / maintainability]** This is the third copy of the "synthesise a DENIED tool-result event triple" shape — `runToolBatch()` already builds the identical `ToolResultStartEvent` / `ToolResultTextDeltaEvent` / `ToolResultEndEvent` sequence for rule-denied calls (around `ReActAgent.java:3006-3021` on main), and the literal `"Permission denied by user"` is now duplicated with line 2033 of this file. Please extract one private helper (e.g. `Flux<AgentEvent> deniedToolResultEvents(String replyId, ToolUseBlock call, String reason)` or a small `publishDeniedToolResult(...)`), so the user-denied, rule-denied and any future deny path cannot drift apart in payload text or event ordering.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2037,6 +2037,22 @@ private void applyConfirmResults(List<ConfirmResult> results) {
                         ToolResultMessageBuilder.buildToolResultMsg(
                                 deniedResult, denied, getName());
                 state.contextMutable().add(deniedMsg);
+                if (replyId != null && !replyId.isEmpty()) {
```
oss-maintainer, line 2040: **[Correctness]** The `replyId != null && !replyId.isEmpty()` guard is the only place in this file where event emission is skipped on a missing reply id, and the no-event branch has no test. Consequence of falling into it: the DENIED `ToolResultBlock` is written to context but the UI/AG-UI stream never sees a `ToolResultEndEvent`, which is exactly the desynchronisation this PR sets out to fix. Worth either (a) a test that pins the silent-resume behaviour, or (b) an explicit contract assertion that `replyId` is always non-null on this path (it comes from `resolvePendingRequestReplyId()` at the caller) so the guard can be dropped rather than left untested.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2037,6 +2037,22 @@ private void applyConfirmResults(List<ConfirmResult> results) {
...
+                            new ToolResultStartEvent(replyId, denied.getId(), denied.getName()));
+                    publishEvent(
+                            new ToolResultTextDeltaEvent(
+                                    replyId,
+                                    denied.getId(),
+                                    denied.getName(),
+                                    "Permission denied by user"));
+                    publishEvent(
+                            new ToolResultEndEvent(
+                                    replyId,
```
oss-maintainer, line 2051: **[Compatibility]** Overloading `applyConfirmResults(List<ConfirmResult>)` rather than changing its signature would silently leave a future caller on the single-arg version and drop the events again; the javadoc block above ("write a DENIED ToolResultBlock to context so the tool will no longer be pending on resume") also does not mention that the method now publishes tool-result events as a side effect. Since there is exactly one call site (`validateAndAcceptConfirmResults`), consider changing the signature outright and extending that javadoc to state the event contract.

oss-maintainer (commented):
### Summary
Right fix for the right symptom: a user-denied HITL tool call previously wrote a DENIED `ToolResultBlock` into context without ever telling the event stream, so AG-UI style consumers were left with a tool call that started and never ended. Sourcing `replyId` from the pending confirm request and correlating start/delta/end with the same `toolCallId` + `ToolResultState.DENIED` is exactly the contract the rule-denied path already uses, and the new test pins the ordering and the correlation. CI green on `8fa232e`, CLA signed.

Caveat for the maintainers: this overlaps with PR #3099 ("fix: emit permission-denied tool result events"), which is currently in conflict with main — please land one of the two rather than both.

### Findings
- **[Warning]** `agentscope-core/.../ReActAgent.java:2040-2055` — third copy of the DENIED start/delta/end triple and of the `"Permission denied by user"` literal (`runToolBatch()` already does this for rule-denied calls); see inline comment on extracting one helper.
- **[Info]** `ReActAgent.java:2040` — the blank-`replyId` guard silently skips emission and no test covers it.
- **[Info]** `ReActAgent.java:2051` — overloaded signature + unchanged javadoc leave a hook for a future caller to drop the events.

Not approving — this is review feedback only; the merge decision stays with the maintainers.

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2038,6 +2038,22 @@ private void applyConfirmResults(List<ConfirmResult> results) {
...
                 state.contextMutable().add(deniedMsg);
+                if (replyId != null && !replyId.isEmpty()) {
+                    publishEvent(
+                            new ToolResultStartEvent(replyId, denied.getId(), denied.getName()));
+                    publishEvent(
+                            new ToolResultTextDeltaEvent(
+                                    replyId,
+                                    denied.getId(),
+                                    denied.getName(),
+                                    "Permission denied by user"));
```
oss-maintainer: The denial payload is now written twice: `ToolResultBlock.text("Permission denied by user")` for the context message above and this literal string for the emitted delta. `runToolBatch()` already repeats the same literal for the rule-denied path. Suggest a single source of truth, e.g. build the `ToolResultBlock denied` once and derive both the message and the delta text from `denied.getText()`, or extract a `DENIED_MESSAGE` constant — otherwise a future wording/localisation change will silently diverge between what the model sees in context and what the UI renders.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2038,6 +2038,22 @@ private void applyConfirmResults(List<ConfirmResult> results) {
                         ToolResultMessageBuilder.buildToolResultMsg(
                                 deniedResult, denied, getName());
                 state.contextMutable().add(deniedMsg);
+                if (replyId != null && !replyId.isEmpty()) {
```
oss-maintainer: Silent degradation when `replyId` is blank: the DENIED result is still written to context, but no `ToolResultStart/Delta/End` events reach the client, so the UI keeps showing the tool call as pending forever. `replyId` comes from `resolvePendingRequestReplyId()`-style metadata lookup, which can legitimately return empty when the pending marker was already cleared. Suggest at least a `log.warn` in the `else` branch (and consider reusing `state.getReplyId()` as a fallback) so an operator can tell "denied but not rendered" apart from "never denied".

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2038,6 +2038,22 @@ private void applyConfirmResults(List<ConfirmResult> results) {
...
+                                    replyId,
+                                    denied.getId(),
+                                    denied.getName(),
+                                    "Permission denied by user"));
+                    publishEvent(
+                            new ToolResultEndEvent(
+                                    replyId,
+                                    denied.getId(),
+                                    denied.getName(),
+                                    ToolResultState.DENIED));
```
oss-maintainer: These events are pushed through `publishEvent()` (i.e. directly into the `FluxSink`) rather than being returned as part of the pipeline the way `runToolBatch()` returns its denied-event `Flux`. That works for `streamEvents`, but on a plain `call()` resume `eventSink`/`externalEventEmitter` are both null, so the three new calls are no-ops with no signal that the emission path was skipped. Returning the events from the resume pipeline (as `runToolBatch` does) would keep both call styles consistent and testable.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentHitlTest.java
```diff
@@ -330,6 +332,39 @@ void confirmedResumeEmitsUserConfirmResultEventCorrelatedToRequireEvent() {
...
+
+        List<AgentEvent> resumeEvents =
+                agent.streamEvents(List.of(confirmMsg(false, req.getToolCalls().get(0))))
+                        .collectList()
+                        .block();
+        assertNotNull(resumeEvents);
+
+        int iStart = indexOf(resumeEvents, ToolResultStartEvent.class);
+        int iDelta = indexOf(resumeEvents, ToolResultTextDeltaEvent.class);
+        int iEnd = indexOf(resumeEvents, ToolResultEndEvent.class);
```
oss-maintainer, line 402: `indexOf(resumeEvents, ToolResultEndEvent.class)` resolves the *first* `ToolResultEndEvent` of the resume stream. That happens to be the denied one here because the script fails the call immediately, but the assertion would silently pass on a wrong event if a resumed turn ever emits an earlier tool result. Asserting `resumeEvents.subList(iStart, iEnd + 1)` contains exactly one `ToolResultEndEvent` with `state == DENIED` would pin the ordering invariant this PR is about.

oss-maintainer (commented):
### Summary
Re-review after the new commits since the previous pass (`8fa232e4` → `4989dfdc`). The fix is on the right track: a user-denied HITL tool call now emits the same `ToolResultStart → TextDelta → End(DENIED)` triple that the rule-denied path in `runToolBatch` already produces, so AG-UI/frontends stop rendering the call as pending. The added test pins reply-id correlation and event ordering. Four follow-ups below, none blocking.

### Findings
- **[Warning]** `ReActAgent.java:2049` — denial text duplicated between context message and emitted delta
- **[Warning]** `ReActAgent.java:2041` — blank `replyId` silently drops all three events
- **[Info]** `ReActAgent.java:2055` — pushed via `publishEvent()` instead of returned in the pipeline, unlike `runToolBatch`
- **[Info]** `ReActAgentHitlTest.java:358` — ordering assertion resolves the *first* end event, not necessarily the denied one

### CLA
`license/cla` status was not found on head `4989dfdc`, so CLA state is unknown here — please confirm the bot check is green before merging; I am not approving while it is unresolved.

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1935,12 +1938,14 @@ private void validateAndAcceptConfirmResults(List<Msg> msgs, List<ToolUseBlock>
             }
 
             String replyId = resolvePendingRequestReplyId(Msg.METADATA_CONFIRM_REQUEST_REPLY_ID);
-            if (!replyId.isEmpty()) {
-                publishEvent(new UserConfirmResultEvent(replyId, normalized));
-                clearPendingRequestReplyId(Msg.METADATA_CONFIRM_REQUEST_REPLY_ID);
+            if (replyId.isEmpty()) {
+                replyId = UUID.randomUUID().toString().replace("-", "");
+                log.warn("Missing confirmation reply id; generated fallback {}", replyId);
```
oss-maintainer, line 1945: The fallback reply id is now generated for **every** confirm resume whose metadata is missing, not just denied ones. That is a visible behaviour change for streaming consumers: `UserConfirmResultEvent.replyId` will no longer match the `RequireUserConfirmEvent.replyId` they correlated on, so a strict correlation check fails (silently, with only a WARN in the agent log). Consider (a) gating the fallback so it only kicks in when there is something to emit (i.e. the denial path that needs an anchor), and (b) emitting a trace/metric counter here so we can see in production how often persisted HITL metadata actually goes missing — if it is never, the fallback is dead weight; if it is common, that points at a state-persistence bug worth its own issue.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2031,16 +2037,29 @@ private void applyConfirmResults(List<ConfirmResult> results) {
...
                 Msg deniedMsg =
                         ToolResultMessageBuilder.buildToolResultMsg(
                                 deniedResult, denied, getName());
                 state.contextMutable().add(deniedMsg);
+                deniedToolResultEvents(denied, replyId, PERMISSION_DENIED_BY_USER)
+                        .forEach(this::publishEvent);
             }
         }
 
+        /** Build the complete DENIED tool-result lifecycle for any permission-denial path. */
```
oss-maintainer, line 2050: Nice consolidation of the previously triplicated literals — this removes the drift risk I flagged on the previous revision. One nit: the javadoc says "any permission-denial path" while `reason` is still passed in from two call sites with two different constants. That is fine; just make the contract explicit ("`reason` is the same text that is written into the DENIED `ToolResultBlock`"), so a future caller cannot pass a reason string that diverges from what the model sees in the context.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentHitlTest.java
```diff
@@ -330,6 +376,125 @@ void confirmedResumeEmitsUserConfirmResultEventCorrelatedToRequireEvent() {
...
+        assertTrue(iConfirm >= 0, "legacy resume must still emit UserConfirmResultEvent");
+        assertTrue(iStart > iConfirm, "denied resume must emit ToolResultStartEvent");
+        assertTrue(iDelta > iStart, "denied resume must emit a result delta");
+        assertTrue(iEnd > iDelta, "denied resume must emit ToolResultEndEvent");
+
+        UserConfirmResultEvent confirm = (UserConfirmResultEvent) resumeEvents.get(iConfirm);
+        ToolResultEndEvent end = (ToolResultEndEvent) resumeEvents.get(iEnd);
+        assertFalse(confirm.getReplyId().isBlank(), "fallback reply id must not be blank");
+        assertNotEquals(req.getReplyId(), confirm.getReplyId());
```
oss-maintainer, line 446: `assertNotEquals(req.getReplyId(), confirm.getReplyId())` locks in the fallback behaviour as intended, but nothing in this test asserts what a *consumer* should do with two different ids. If the intent is "a fallback id is uncorrelated but self-consistent across the confirm + tool-result triple", it is worth asserting that the paired `RequireUserConfirmEvent`/`ToolResultEndEvent` stream from a *first* call still uses a single id — i.e. that this only regresses the legacy/metadata-lost case and never the normal case.

agentscope-core/src/test/java/io/agentscope/core/agent/ReActAgentHitlTest.java
```diff
@@ -173,6 +177,35 @@ public Mono<ToolResultBlock> callAsync(ToolCallParam param) {
         }
     }
 
+    private static final class DenyingTool extends ToolBase {
+        DenyingTool(String name) {
```
oss-maintainer, line 181: `DenyingTool` passes eight positional constructor args (`false, true, false, null, false, false`) to `ToolBase`. If `ToolBase`'s constructor is a boolean-heavy internal API, a small static factory or a builder in the test would keep this readable when flags are added. Also consider naming it `RuleDeniedTool` — "denying" is ambiguous against the user-denied case this PR is about, since this fixture exercises the *rules* path.

oss-maintainer (commented):
### Summary
Re-review after the new commits (`f3ab33c7` → `0ec18615`). All four points from the previous review were addressed: the denial literals are now single-sourced through `PERMISSION_DENIED_BY_USER` / `PERMISSION_DENIED_BY_RULES`, a blank `replyId` no longer silently drops the event triple, the rule-denied path in `actingStream` emits the DENIED lifecycle before `RequireUserConfirmEvent`, and the tests assert the ordered `Start → TextDelta → End(DENIED)` triple instead of "the first `ToolResultEndEvent`". The shared `deniedToolResultEvents(...)` helper is the right shape — it is now the single place the rules path, the HITL-denial path and the mixed-batch path converge on.

Not approving only because two non-blocking items are worth a decision before merge (details inline):
- **The missing-`replyId` fallback now fires for approved resumes too**, so `UserConfirmResultEvent.replyId` can differ from the `RequireUserConfirmEvent.replyId` a client correlated on. Worth an explicit call on whether that is intended contract or a safety net that should be narrowed to the emit path that needs it.
- **The sibling `METADATA_EXTERNAL_EXECUTION_REQUEST_REPLY_ID` path is untouched** and still drops `ExternalExecutionResultEvent` silently on a blank reply id — the same class of bug this PR fixes.

CLA is signed (`license/cla=success` on `0ec18615`), so these are advisory, not blockers.

---

- **[WARNING]** `agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java:2306` — Asymmetry left behind: the sibling external-execution path still silently swallows `ExternalExecutionResultEvent` when `replyId` is empty, exactly the failure mode this PR fixes for the confirm path. If a resume with missing metadata is reachable here too, the caller's UI keeps the pending external call in limbo with no signal at all. Either reuse the same fallback (extract `resolvePendingRequestReplyId` + fallback into one helper used by both call sites) or state in the PR description why the confirm path is special. _(line outside diff)_

jujn (approved):

PR #3106, opened 2026-09-11
fix(harness): avoid blocking stream completion on transcript persistence

### LINKED ISSUE #3059: [Bug]:Memory flush / consolidation / transcript block completion of the agent event stream (concatWith instead of fire-and-forget), contradicting the documentation
Summary

MemoryFlushMiddleware, MemoryMaintenanceMiddleware and TranscriptMiddleware attach their post-turn work to the agent event stream with concatWith(...). Because concatWith is sequential composition, the downstream subscriber does not receive onComplete until that work finishes. subscribeOn(Schedulers.boundedElastic()) only moves the work to another thread — it does not make it fire-and-forget.

The documentation states the opposite, and for flush this is a regression from the previous implementation.

For streaming deployments (SSE / WebSocket), where stream completion is what closes the client connection, this turns background memory maintenance into user-visible latency. We measured up to 85 seconds of dead time after the assistant's final text had already been delivered.

Environment

- agentscope-harness 2.0.2, and current main (verified — see code references below)
- Java 17, Spring Boot 3.3.5, HarnessAgent + RemoteFilesystemSpec (Redis-backed workspace), IsolationScope.USER
- Streaming via agent.streamEvents(...) piped to an SSE endpoint

Expected behavior (per the docs)

docs/v2/zh/docs/harness/memory.md (line 61 on main):

▎ Flush 和 offload 都是异步执行的：它们在响应流结束后通过 doOnComplete 以 fire-and-forget 方式启动，不会阻塞当前 call() 的返回。

("Flush and offload are executed asynchronously: they are started fire-and-forget via doOnComplete after the response stream ends, and do not block the return of the current call().")

Actual behavior

All three middlewares use concatWith, so the caller waits:

// MemoryFlushMiddleware#onAgent (main, L137-L152)
return next.apply(input)
        .concatWith(
                        .onErrorResume(e -> { log.warn("Memory flush failed: {}", e.getMessage()); return Mono.empty(); })
                        .then(Mono.<AgentEvent>empty()));

MemoryMaintenanceMiddleware#onAgent (L137-L152) is identical in shape, and the work it wraps blocks on an LLM call:

// MemoryMaintenanceMiddleware#consolidateMemory (main, L246)

TranscriptMiddleware#onAgent follows the same concatWith pattern for transcript append.

The class javadoc of MemoryMaintenanceMiddleware (L44) documents the mechanism explicitly — "Fires on the agent invocation completion (via onAgent concatWith, after MemoryFlushMiddleware)" — so the implementation is self-consistent; it is the docs and the streaming use case that are at odds with it.

How this regressed

Flush used to be genuinely fire-and-forget. Commit 7ab520db (feat(harness): introduce async tool and notification support. #1802, 2026-06-17) changed it:

- return next.apply(input).doOnComplete(() -> doFlush(agent, rc).subscribe());
+ return next.apply(input).concatWith(Mono.defer(...).subscribeOn(Schedulers.boundedElastic())...);

(Consolidation appears to have used concatWith from the start; only flush is a regression. If the change was intentional — e.g. to order async-tool notifications correctly — that rationale isn't captured anywhere, and the docs were not updated.)

Impact

The severity depends entirely on how the caller consumes the stream:

- CLI / batch agents: invisible. The output has already been printed; nobody notices the process taking a few more seconds before exiting. Guaranteeing "memory is persisted before the turn is considered over" is arguably the right trade-off here.
- Long-lived streaming connections (SSE / WebSocket): directly user-visible. onComplete is what ends the response; every millisecond spent in these hooks is a user staring at a spinner after the assistant's answer is already fully rendered.

The cost also grows with how much the agent remembers: MemoryConsolidator rewrites MEMORY.md in full on every run, and its prompt budget is maxChars = maxMemoryTokens * 4 (MemoryConsolidator L145), i.e. 16,000 characters by default (consolidationMaxTokens = 4000). A richer memory file means a longer blocking LLM generation.

Measurements from production

A. Same session, same user, consecutive turns — time from the final AGENT_END event to stream completion:

┌──────┬────────────────────────────────────────────┬──────────────┐
│ Turn │          Consolidation triggered           │ Tail latency │
├──────┼────────────────────────────────────────────┼──────────────┤
│ n    │ no                                         │ 1.76 s       │
├──────┼────────────────────────────────────────────┼──────────────┤
│ n+1  │ yes — MEMORY.md consolidated (16295 chars) │ 85.4 s       │
└──────┴────────────────────────────────────────────┴──────────────┘

B. Consolidation time is proportional to the rewritten output (same deployment, after we lowered consolidationMaxTokens):

┌───────────────────────────────┬────────────────────┐
│       MEMORY.md output        │ Consolidation time │
├───────────────────────────────┼────────────────────┤
│ 16,295 chars (default budget) │ 76 s               │
├───────────────────────────────┼────────────────────┤
│ 5,170 chars                   │ 24 s               │
├───────────────────────────────┼────────────────────┤
│ 1,218 chars                   │ 5.3 s              │
└───────────────────────────────┴────────────────────┘

C. Breakdown of one tail (from the middlewares' own debug logs), memory hooks alone:

AGENT_END                          +0
Running memory maintenance...      +1 ms
MEMORY.md consolidated             +5.3 s     <- consolidation LLM call
Memory maintenance completed       +16 ms
Memory flush completed             +3.9 s     <- flush LLM call
Message offload completed          +164 ms
                                   = 9.4 s of blocked stream completion

All of this happens after the last text token has been emitted.

Why it can't be worked around via configuration

MemoryConfig exposes only: model, flushPrompt, consolidationPrompt, consolidationMaxTokens, consolidationMinGap, dailyFileRetentionDays, sessionRetentionDays, flushTrigger. There is no switch for the execution mode.

- flushTrigger(NEVER) disables only the per-call flush; consolidation and transcript append still run — and still block.
- disableMemoryHooks() is all-or-nothing: it removes automatic extraction entirely.

Tuning consolidationMaxTokens / consolidationMinGap only reduces the magnitude and frequency of the stall, not the fact that it sits on the critical path.

Suggested fix

Either of:

1. Restore fire-and-forget for these hooks, e.g. doOnComplete(() -> mono.subscribe()), so that onComplete reaches the subscriber immediately (this is also what the docs describe today); or
2. Make it configurable, e.g. MemoryConfig.blockOnCompletion(boolean) (defaulting to whatever preserves current behavior), so callers on long-lived streaming connections can opt out while batch/CLI callers keep the "memory durably written before the turn ends" guarantee.

Either way, docs/v2/zh/docs/harness/memory.md should be updated to match the implementation.

Related, smaller observation

While profiling the above we also found WorkspaceIndex.open(...) creating a bare SQLite connection with no journal_mode=WAL and no busy_timeout, in autocommit mode, shared across threads. MemoryFlushManager#offloadMessages → WorkspaceManager#updateSessionIndex (on the stream-completion path) and the single-threaded SessionTree mirror executor both write through it, so they serialize on the JDBC connection. We captured a thread dump showing the stream-completion thread BLOCKED in WorkspaceIndex.upsert while the mirror thread held it. On slower/networked volumes this adds seconds to the same critical path. Happy to file that separately if you'd prefer.

ningmao-hlyz:
## Summary

Fixes #3059.

- Keep transcript persistence fire-and-forget for regular agent calls.
- Track pending transcript writes with `MemoryBackgroundTasks` so shutdown can wait for them.
- Keep sandbox-backed calls in the stream until transcript persistence finishes.
- Serialize concurrent transcript appends for the same session.
- Add regression coverage for stream completion, shutdown quiescence, and sandbox release ordering.

## Problem

`TranscriptMiddleware` used `concatWith(...)` for transcript persistence. Because the persistence task was part of the sequential stream, downstream subscribers did not receive `onComplete` until transcript writing finished.

This added user-visible tail latency for streaming callers such as SSE and WebSocket deployments, even after the final assistant event had already been emitted.

## Changes

- Regular calls now schedule transcript persistence from `doOnComplete(...)` without delaying downstream completion.
- Pending asynchronous writes are counted by `MemoryBackgroundTasks`, preserving lifecycle shutdown/quiescence guarantees.
- The conversation context is copied before the asynchronous task is scheduled.
- Same-session transcript read/merge/write cycles are serialized across writer instances.
- Sandbox-backed calls intentionally keep persistence inside the stream so the sandbox is not released before the writer finishes.

`MemoryFlushMiddleware` and `MemoryMaintenanceMiddleware` already use tracked fire-and-forget execution and are unchanged by this PR.

## Testing

- `mvn -pl agentscope-harness -Dtest=SessionTranscriptWriterTest,TranscriptMiddlewareAsyncTest test`
- `mvn -pl agentscope-harness spotless:check`


agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +73,40 @@ public Flux<AgentEvent> onAgent(
...
+    }
+
+    private void scheduleTranscript(Agent agent, RuntimeContext rc) {
+        AgentState state = RuntimeContext.resolveAgentState(rc, agent);
+        if (state == null) {
+            return;
+        }
+        // Capture the defensive copy before detaching so a later turn cannot change this task's
+        // transcript input.
+        List<Msg> messages = state.getContext();
```
oss-maintainer, line 101: `state.getContext()` returns `Collections.unmodifiableList(new ArrayList<>(context))` (see `AgentState.getContext()`), so the whole conversation is copied **synchronously on the stream's terminal thread** inside `doOnComplete`. For long conversations that is exactly the tail latency this PR removes for the last event. Suggest doing the snapshot inside the deferred work, e.g. keep `begin()` on the caller thread but move the copy into `Mono.fromRunnable(...)`, or dispatch via `Mono.defer(...)` + `subscribeOn(boundedElastic())` after `begin()` — `begin()` must still run synchronously so `HarnessAgent.close()` cannot observe `inFlight == 0` while a write is being scheduled.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +73,40 @@ public Flux<AgentEvent> onAgent(
...
+                    Mono.defer(() -> appendTranscript(agent, rc))
+                            .subscribeOn(Schedulers.boundedElastic())
+                            .onErrorResume(
+                                    e -> {
+                                        log.warn("Transcript append failed: {}", e.getMessage());
+                                        return Mono.empty();
+                                    })
+                            .then(Mono.<AgentEvent>empty()));
+        }
+        return stream.doOnComplete(() -> scheduleTranscript(agent, rc));
```
oss-maintainer, line 91: `doOnComplete` has no error guard here: anything thrown by `scheduleTranscript` (state resolution, `getContext()`, an executor rejection inside `subscribe`) surfaces as an `onError` emitted **after** `onComplete` has already been signalled downstream, which breaks the reactive contract and typically ends up as a dropped error rather than a log line. The previous `concatWith(...).onErrorResume(...)` chain could not fail this way. A `try/catch (Exception e) { log.warn(...) }` around the body (or `Mono.fromRunnable(...).subscribeOn(...).subscribe(null, ...)`) would keep transcript persistence strictly best-effort.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +73,40 @@ public Flux<AgentEvent> onAgent(
...
-                                            log.warn(
-                                                    "Transcript append failed: {}", e.getMessage());
-                                            return Mono.empty();
-                                        })
-                                .then(Mono.<AgentEvent>empty()));
+        Flux<AgentEvent> stream = next.apply(input);
+        // The per-call sandbox is released by the outer Flux.using scope as soon as the stream
+        // completes. Keep transcript persistence in the stream for sandbox calls so the writer
+        // cannot start after the sandbox has been released.
+        if (rc.get(SandboxAcquireResult.class) != null) {
```
oss-maintainer, line 80: The sandbox branch depends on `TranscriptMiddleware.onAgent` being composed *inside* `HarnessAgent.wrappedStreamEvents`' `Flux.using`, so that the `SandboxAcquireResult` binding is still set when this hook runs and `doOnComplete` fires before sandbox release. `TranscriptMiddlewareAsyncTest` locks the behaviour for a `userManaged` binding; it would be worth an assertion (or a comment next to the middleware registration) that pins the ordering assumption, otherwise a future middleware-order change silently reintroduces the race this branch exists to avoid.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
```diff
@@ -51,13 +53,20 @@
...
 
     private static final Logger log = LoggerFactory.getLogger(SessionTranscriptWriter.class);
 
     private static final String SESSION_CONTEXT_TAG = "<session_context>";
 
+    /**
+     * Active per-session locks for the transcript read/merge/write cycle; idle entries are evicted.
+     */
+    private static final ConcurrentHashMap<String, WriteLock> WRITE_LOCKS =
+            new ConcurrentHashMap<>();
```
oss-maintainer, line 68: The new `WRITE_LOCKS` map serialises read/merge/write per session path **within one JVM**. Under `agentscope-distribution` (stateless horizontal scaling, several replicas sharing the same workspace/transcript store) two pods can still interleave `tree.load()`/`syncFromRemote()` and `tree.flush()` and lose entries. Could the class javadoc say explicitly that this is process-local only (and that cross-instance safety still relies on `TranscriptStore`/`SessionTree` merge semantics)? The PR description's "serialized across writer instances" reads a bit stronger than the code guarantees.

oss-maintainer (commented):
### Summary
Fixes #3059 by scheduling transcript persistence from `doOnComplete` instead of `concatWith`, keeping sandbox-backed calls in-stream, and serialising same-session read/merge/write behind a ref-counted per-path lock. The direction is right, the fire-and-forget task is correctly tracked by `MemoryBackgroundTasks`, and the regression tests cover completion ordering, sandbox release ordering and the pre-detach snapshot.

### Verdict
No blocking issue found — inline comments are latency/robustness polish plus one doc-accuracy note.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
```diff
@@ -90,66 +99,30 @@ public void appendMessages(
...
-                    lastId = entry.getId();
-                    appended++;
-                }
+            String lockKey = contextFile.toAbsolutePath().normalize().toString();
+            // Count waiters before blocking so the last releaser cannot evict a lock still in use.
+            WriteLock writeLock =
+                    WRITE_LOCKS.compute(
+                            lockKey,
+                            (k, existing) -> {
+                                WriteLock next = existing != null ? existing : new WriteLock();
```
oss-maintainer, line 107: **[Warning]** The refcount is released before `lock.unlock()`, so mutual exclusion can be lost. `computeIfPresent` here only decrements `users` (it returns a non-null value when the count is still > 0), but the monitor for that key is the map bin lock, not `writeLock.lock`. If a new caller runs its `compute` increment while we are still between this line and the `unlock()` below, the counter is pinned at >= 1 and the eviction branch can never be taken, so a third caller legitimately creates a *different* `WriteLock` and enters the read/merge/write cycle concurrently with us — exactly the interleaving this PR is meant to prevent, just much rarer.

Suggested fix: release the reference *after* unlocking, and let the unlock itself happen inside the map mutation so no new user can be attached to an about-to-be-discarded lock:

```java
try {
    appendMessagesLocked(rc, messages, agentId, sessionId, contextFile);
} finally {
    writeLock.lock.lock(); // never contended, just re-enters the monitor we own
    try {
        WRITE_LOCKS.compute(lockKey, (k, current) -> {
            if (current != writeLock) {
                return current; // defensive: another lock replaced us
            }
            return --current.users == 0 ? null : current;
        });
    } finally {
        writeLock.lock.unlock();
    }
}
```

With this ordering the eviction and the increment are serialized by the same lock we hold, so a fresh `WriteLock` can never be published while an append is in flight.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
```diff
@@ -51,13 +53,20 @@
...
  */
 public class SessionTranscriptWriter {
 
     private static final Logger log = LoggerFactory.getLogger(SessionTranscriptWriter.class);
 
     private static final String SESSION_CONTEXT_TAG = "<session_context>";
 
+    /**
+     * Active per-session locks for the transcript read/merge/write cycle; idle entries are evicted.
+     */
```
oss-maintainer: **[Warning]** `WRITE_LOCKS` is process-local, but the guarded cycle is `tree.load()` → `tree.syncFromRemote()` → `tree.flush()` (see `appendMessagesLocked`). Two JVMs sharing the same workspace/`TranscriptStore` (agentscope-distribution is explicitly the multi-instance path) will both pass through this lock and can still interleave a remote read-modify-write, losing entries. The Javadoc wording ("serialized across writer instances") could be read as covering that case — could we either state explicitly that this only serializes writers inside one JVM, or push the guard down to the session-index/`TranscriptStore` layer where a cross-process lease/version check is possible? A follow-up issue is fine if the intended scope is single-process.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +73,40 @@ public Flux<AgentEvent> onAgent(
...
+            return stream.concatWith(
+                    Mono.defer(() -> appendTranscript(agent, rc))
+                            .subscribeOn(Schedulers.boundedElastic())
+                            .onErrorResume(
+                                    e -> {
+                                        log.warn("Transcript append failed: {}", e.getMessage());
+                                        return Mono.empty();
+                                    })
```
oss-maintainer, line 96: **[Warning]** Inconsistency with the old behavior: the sandbox path keeps `.onErrorResume(e -> { log.warn(...); return Mono.empty(); })`, while the new background path swallows the error only via the `subscribe(null, consumer)` overload at line 105, and the `Mono.fromRunnable` in `appendTranscript(agent, rc, messages)` no longer has the previous `onErrorResume`/`doOnSuccess` wrappers. That works today because `SessionTranscriptWriter.appendMessages` has a catch-all `catch (Exception e)` — but `Mono.fromRunnable` also turns a *cancellation/interrupt* of the boundedElastic worker, or any `Error` (e.g. `NoClassDefFoundError` on a lazily loaded class), into an `onError` signal that will now reach the subscriber's error consumer, and a throwing `log.warn` handler there would surface as `ErrorCallbackNotImplemented`. Since the two branches of this method now have different error-discipline, could we apply `.onErrorResume(e -> { log.warn("Transcript append failed: {}", e.getMessage()); return Mono.empty(); })` to the background `append` as well, so both paths are identical and the failure policy is explicit at the call site?

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +73,40 @@ public Flux<AgentEvent> onAgent(
...
+                            .then(Mono.<AgentEvent>empty()));
+        }
+        return stream.doOnComplete(() -> scheduleTranscript(agent, rc));
+    }
+
+    private void scheduleTranscript(Agent agent, RuntimeContext rc) {
+        AgentState state = RuntimeContext.resolveAgentState(rc, agent);
```
oss-maintainer: **[Info]** `scheduleTranscript` is invoked from `doOnComplete`, so it runs on whatever thread emitted the stream terminal signal, and `RuntimeContext.resolveAgentState(rc, agent)` plus the defensive `state.getContext()` copy happen inline there. For the goal of this PR (not delaying completion) it is worth checking whether `resolveAgentState` can touch the state store: if it can, that work should move onto `boundedElastic` together with the `append`, e.g. by doing the capture inside the `Mono.fromRunnable(...)` that is then `subscribeOn`'d, keeping only the already-obtained `state` reference on the stream thread.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +73,40 @@ public Flux<AgentEvent> onAgent(
...
+        // transcript input.
+        List<Msg> messages = state.getContext();
+        if (messages.isEmpty()) {
+            return;
+        }
+        Mono<Void> append = appendTranscript(agent, rc, messages);
+        MemoryBackgroundTasks.begin();
+        append.subscribeOn(Schedulers.boundedElastic())
+                .doFinally(signal -> MemoryBackgroundTasks.end())
+                .subscribe(null, e -> log.warn("Transcript append failed: {}", e.getMessage()));
```
oss-maintainer: **[Info]** `MemoryBackgroundTasks.begin()` is called synchronously before `subscribeOn`, and `end()` is paired via `doFinally`, which is the right shape (no window where the task is uncounted). One residual gap: `end()` runs on `doFinally`, and `awaitQuiescence` is only consulted in `HarnessAgent.close()` per the class docs of `MemoryBackgroundTasks`. If a session is evicted / a non-sandbox agent call completes and the harness drops the agent before the write lands, the transcript for that turn is still lost silently (background failures are only logged at WARN). It would help operators if the failure log carried `agent=`/`session=` context here so a dropped transcript is traceable — the previous `concatWith` path had the same weakness, so this is just an observability suggestion, not a blocker.

oss-maintainer (commented):
### Summary
Moves non-sandbox transcript persistence off the critical path: `TranscriptMiddleware` now dispatches the append in `doOnComplete` as a tracked fire-and-forget `MemoryBackgroundTasks` task (sandbox calls keep it in-stream so the per-call sandbox stays alive), and `SessionTranscriptWriter` guards the `load → syncFromRemote → flush` cycle with a reference-counted per-path lock so concurrent appends for one session can't interleave. The direction is right and the test coverage is genuinely good (`onComplete` fires before the write finishes, sandbox writes stay in-stream, message snapshot taken before detaching, failure paths don't break the stream). One ordering bug in the new lock protocol and a scope question on the guard are worth fixing before this lands.

### Findings
- **[Warning]** `SessionTranscriptWriter.java:108` — the refcount is dropped *before* `unlock()`, so a new caller can publish a second `WriteLock` for the same path while an append is still in flight; the eviction check runs under the map bin lock, not the write lock.
- **[Warning]** `SessionTranscriptWriter.java:66` — the lock is process-local while the guarded cycle reaches the remote `TranscriptStore`; two JVMs on a shared workspace are still unprotected. Please confirm the intended scope and document it.
- **[Warning]** `TranscriptMiddleware.java:88` — the sandbox branch and the background branch now have different error-discipline; the previous `onErrorResume` was dropped from `appendTranscript`.
- **[Info]** `TranscriptMiddleware.java:84` / `:91` — state resolution runs inline on the completing thread; failure logging lost its `agent=`/`session=` context.
- **[Info]** `TranscriptMiddlewareAsyncTest.java:242` — no test pins down the cancel semantics of the `doOnComplete`-only dispatch.

### Suggestions
The lock fix is small — take the eviction decision while still holding the monitor (see the inline comment for a snippet), so `users == 0 → remove` and `users++ → reuse` can never overlap with an in-flight append. For the error-discipline item, hoisting the existing `onErrorResume` lambda into a `private static Mono<Void> logFailure(Throwable)` and applying it in both branches would keep the two paths from drifting again.

Neither item blocks the design: the change is a clear improvement over blocking stream completion on a disk/remote write, and the added tests exercise exactly the races that matter.

---

agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/TranscriptMiddlewareAsyncTest.java
```diff
@@ -0,0 +1,326 @@
...
+                                agent,
+                                rc,
+                                new AgentInput(List.of(userMessage)),
+                                input -> Flux.just(new AgentEndEvent("reply-1")))
+                        .count()
+                        .block(Duration.ofSeconds(2)));
+    }
+
+    @Test
+    void sandboxTranscriptFailureDoesNotFailAgentStream() {
```
oss-maintainer, line 243: **[Info]** `sandboxTranscriptWriteStaysInStreamUntilPersistenceFinishes` asserts the sandbox path via `Flux.using` + a blocking `collectList()`, which is good, but there is no equivalent test that a *cancelled* non-sandbox stream still persists what it had (the PR title's "completion" signal is `doOnComplete` only). If cancellation without persistence is the intended contract, a small test asserting `doOnCancel` does not schedule a write would lock that decision down; if it is not intended, `doOnCancel`/`doFinally` would need the same dispatch.

oss-maintainer:
Follow-up on CI for the current head (`105959a6`) — the Linux build is failing, and the failure shape looks related to this change rather than flaky:

```
[ERROR] AgentSpawnToolPlanModeTest.agentSend_reappliesParentPlanModeBeforeInvokingExistingChild » JUnit Failed to close extension context
Caused by: java.io.IOException: Failed to delete temp directory /tmp/junit-…
  The following paths could not be deleted: <root>, user, user/agents, user/agents/worker/sessions
[ERROR] Tests run: 1048, Failures: 0, Errors: 1, Skipped: 9
```

https://github.com/agentscope-ai/agentscope-java/actions/runs/34626731278

The leftover path is a *session transcript* directory under a `@TempDir` workspace, and the error is `DirectoryNotEmptyException` during JUnit's extension teardown — i.e. something wrote into the temp workspace after the test body finished. With this PR the non-sandbox transcript append moves to `doOnComplete` + a detached `boundedElastic` task, so a `@TempDir`-based test whose agent stream completes can now have an in-flight write outliving the directory. `MemoryBackgroundTasks.awaitQuiescence(...)` is already the right tool: awaiting it in the affected test's teardown (or in a shared `@AfterEach` for harness tests that use `@TempDir` workspaces) should make the ordering explicit instead of timing-dependent.

Not blocking on this, and if you can show the same test failing on an unrelated head please treat this as a red herring — but it is worth ruling the ordering out before the build is re-run a few times and assumed flaky.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +73,79 @@ public Flux<AgentEvent> onAgent(
...
+                    agentNameForLog(agent),
+                    rc.getSessionId(),
+                    e.getMessage());
+        }
+    }
+
+    private List<Msg> snapshotMessages(Agent agent, RuntimeContext rc) {
+        try {
+            AgentState state = RuntimeContext.resolveAgentState(rc, agent);
+            return state == null ? List.of() : state.getContext();
```
oss-maintainer, line 137: **[Warning]** Still open from the previous round: `state.getContext()` is `Collections.unmodifiableList(new ArrayList<>(context))` (see `AgentState.getContext()`), so the full conversation is deep-copied **synchronously on the thread that emits the terminal signal**, because `scheduleTranscript` is invoked straight from `doOnComplete`. For a long-running session that copy is proportional to the context size and lands exactly where this PR is trying to remove tail latency. `resolveAgentState` is also resolved inline here. Consider handing the `Agent`/`RuntimeContext` to the deferred `Mono.fromRunnable` and taking the snapshot inside it (keeping `MemoryBackgroundTasks.begin()` on the calling thread so the count is published before the subscribe), or copying only the messages appended by this turn.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +73,79 @@ public Flux<AgentEvent> onAgent(
...
+                                                "Transcript append failed for agent={}, session={}:"
+                                                        + " {}",
+                                                agentNameForLog(agent),
+                                                rc.getSessionId(),
+                                                e.getMessage());
+                                        return Mono.empty();
+                                    })
+                            .then(Mono.<AgentEvent>empty()));
+        }
+        return stream.doOnComplete(() -> scheduleTranscript(agent, rc));
```
oss-maintainer, line 99: **[Info]** `doOnComplete` fires only on normal completion, so a cancelled client (SSE disconnect / WebSocket close — the primary deployments for this framework) now drops the transcript write entirely, whereas the old `concatWith(...)` also skipped it but did so via a different path. If "cancel means discard the turn" is the intended contract, please say so in the class javadoc and add the small cancel-path test that was missing last round; if not, a `takeUntilOther`-style guard or handling `SignalType.CANCEL` explicitly would keep partial turns durable.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
```diff
@@ -90,66 +98,39 @@ public void appendMessages(
...
+                    WRITE_LOCKS.computeIfPresent(
+                            lockKey,
+                            (k, current) -> {
+                                if (current != writeLock) {
+                                    return current;
+                                }
+                                boolean idle = --current.users == 0;
+                                // Keep the map mutation and unlock ordered together: a new caller
+                                // cannot publish a replacement lock until this append is released.
+                                writeLock.lock.unlock();
```
oss-maintainer, line 127: **[Info]** Lock-protocol fix confirmed: unlocking inside `computeIfPresent` closes the window where a new caller could increment `users` on a lock that was logically already handed back, and the `isHeldByCurrentThread()` fallback keeps the counter and the monitor consistent if the entry was already replaced. One residual nit: if `computeIfPresent` sees `current != writeLock`, `users` is never decremented for this caller — that is only reachable when the entry was evicted and recreated, so the orphaned count belongs to a dead `WriteLock` and is harmless, but a one-line comment saying so would save the next reader the analysis.

oss-maintainer (commented):
### Summary

Re-reviewed against the updated head `eb67eb22`. The lock protocol I flagged last round is now correct: the unlock happens *inside* `WRITE_LOCKS.computeIfPresent`, so a new caller can no longer publish a replacement lock for a key whose append has not actually been released, and the waiters are counted before blocking. Class docs now say explicitly that the serialisation is process-local and that cross-process coordination belongs to the `TranscriptStore`, which is the other thing I asked for. Both earlier warnings are resolved — thanks.

Downgraded to COMMENT (not APPROVE) because the latency point on the snapshot is still open on the exact path this PR exists to optimise, and the cancellation contract is still undocumented. CLA status is `unknown` for this repo/head (`license/cla` context not reported), so per my gating rule I'm not approving either way.

### Findings
- **[Warning]** `TranscriptMiddleware.java:132` — `state.getContext()` deep-copies the whole conversation synchronously on the thread that emits `onComplete`; this is the tail latency the PR is meant to remove.
- **[Info]** `TranscriptMiddleware.java:96` — `doOnComplete` means a cancelled stream persists nothing; either document + test that contract or cover `CANCEL`.
- **[Info]** `SessionTranscriptWriter.java:125` — lock fix confirmed; the `current != writeLock` branch leaves the (already-dead) `WriteLock` counter undecremented, worth one clarifying comment.

### Verification
- No new imports of `Schedulers` misuse; `MemoryBackgroundTasks.begin()` stays on the calling thread so the task is counted before `subscribeOn` hands off — the pairing with `doFinally(end())` plus the `catch (RuntimeException)` `end()` is now symmetric, which closes the "uncounted window" note from last round.
- Sandbox branch still keeps persistence in-stream, and `TranscriptMiddlewareAsyncTest` still covers that ordering.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
ningmao-hlyz: Regarding the snapshot warning: I intentionally take the defensive context snapshot before detaching the write. AgentState is mutable, and a new call for the same session may start immediately once the current stream completes. Moving getContext() into the background task could read messages from the next turn or lose messages after compaction. getContext() only copies the list without cloning Msg objects, and all persistence I/O stays off the completion path. A true per-turn delta snapshot requires exposing call boundaries from the core lifecycle, so I plan to address this in a follow-up instead of changing consistency semantics in this PR.

I will document the discard-on-cancel behavior and add the clarifying comment for the lock branch.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +76,79 @@ public Flux<AgentEvent> onAgent(
...
+                    agentNameForLog(agent),
+                    rc.getSessionId(),
+                    e.getMessage());
+        }
+    }
+
+    private List<Msg> snapshotMessages(Agent agent, RuntimeContext rc) {
+        try {
+            AgentState state = RuntimeContext.resolveAgentState(rc, agent);
+            return state == null ? List.of() : state.getContext();
```
oss-maintainer, line 137: **[Warning]** Still open from the previous round: `state.getContext()` is `Collections.unmodifiableList(new ArrayList<>(context))` (see `AgentState.getContext()`), so the whole conversation is copied **synchronously on the thread that emits the terminal signal**, before `onComplete` reaches the downstream subscriber. That is the same tail latency this PR removes from the `concatWith` path, just relocated: an O(context) copy on the stream thread instead of an O(context) write. Since `scheduleTranscript` already detaches the write, could the snapshot move inside `Mono.defer(...)` (boundedElastic) instead? If the synchronous copy is deliberate — to freeze the turn before a later `contextMutable()` write mutates the backing list in place — then please say so here; the comment above currently reads as if the copy were already off-thread.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +76,79 @@ public Flux<AgentEvent> onAgent(
...
+    }
+
+    private void scheduleTranscript(Agent agent, RuntimeContext rc) {
+        // AgentState is mutable and may be reused by a later turn, so capture this turn's
+        // defensive snapshot before detaching the transcript write from the stream.
+        List<Msg> messages = snapshotMessages(agent, rc);
+        if (messages.isEmpty()) {
+            return;
+        }
+        MemoryBackgroundTasks.begin();
```
oss-maintainer, line 111: **[Info]** `MemoryBackgroundTasks` is a process-wide in-flight counter shared with memory flush/maintenance, so transcript writes now participate in every `HarnessAgent.close()` quiescence wait: a busy unrelated session can make an idle agent block up to the 5s timeout. Pre-existing shape rather than something this PR introduces, but this PR widens who contributes to the counter — worth a follow-up if per-agent (or per-workspace) accounting is cheap enough.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/session/SessionTranscriptWriter.java
```diff
@@ -90,66 +98,41 @@ public void appendMessages(
...
+                                WriteLock next = existing != null ? existing : new WriteLock();
+                                next.users++;
+                                return next;
+                            });
+            writeLock.lock.lock();
+            try {
+                appendMessagesLocked(rc, messages, agentId, sessionId, contextFile);
+            } finally {
+                try {
+                    WRITE_LOCKS.computeIfPresent(
```
oss-maintainer, line 116: **[Info]** Confirmed fixed on this head and unchanged by the rebase: the unlock happens inside `computeIfPresent`, so a new caller cannot increment `users` on a lock whose append has logically already been handed back, and the `isHeldByCurrentThread()` fallback keeps the counter and the monitor consistent if the entry was already replaced. The class javadoc now also states the process-local scope and that cross-process coordination belongs to the `TranscriptStore` — that closes both items from the earlier round.

oss-maintainer (commented):
### Summary
Re-reviewed against the force-pushed head `e36a99c0`. The lock protocol and the sandbox-vs-background split both read correctly, the process-local scope is now stated in the class javadoc, and the new `TranscriptMiddlewareAsyncTest` covers the cases raised last round (in-stream sandbox write, shutdown quiescence, snapshot-before-write, failure isolation on both paths). CI is green on ubuntu and windows, and `license/cla` now reports signed.

Both warnings from the previous round are resolved; downgraded to COMMENT rather than APPROVE only because the snapshot latency point is still open on the exact path this PR exists to optimise.

### Findings
- **[Warning]** `TranscriptMiddleware.java:135` — the full-conversation `state.getContext()` copy still happens synchronously on the terminal-signal thread inside `doOnComplete`.
- **[Info]** `TranscriptMiddleware.java:109` — transcript writes now join the process-wide `MemoryBackgroundTasks` counter that every `HarnessAgent.close()` waits on.
- **[Info]** `SessionTranscriptWriter.java:116` — previous lock-protocol warning confirmed fixed and unchanged by the rebase.

### Checklist for approval
1. Either move the snapshot inside the `Mono.defer(...)` body, or document why the copy must stay on the stream thread.
2. Confirm the cancellation contract (persist-or-discard) is the intended product decision for SSE/WebSocket disconnects — the javadoc now states it, which is enough for me.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
ningmao-hlyz: AgentState is mutable and may be reused by the next turn as soon as the current stream completes. Moving the snapshot into the deferred background task could therefore capture messages from the next turn or miss messages from the current one. The comment in TranscriptMiddleware now documents this consistency trade-off explicitly. Only the transcript persistence I/O is offloaded to the background scheduler.

agentscope-harness/src/main/java/io/agentscope/harness/agent/HarnessAgent.java
```diff
@@ -460,14 +460,14 @@ public io.agentscope.core.permission.PermissionMode getPermissionMode(
...
-            // race with resource cleanup (e.g., temp workspace deletion in tests).
-            io.agentscope.harness.agent.memory.session.SessionTree.awaitMirrorQuiescence(
-                    5, java.util.concurrent.TimeUnit.SECONDS);
             // Drain fire-and-forget memory flush/maintenance so async memory/*.md writes do not
             // race with resource cleanup (e.g., temp workspace deletion in tests).
             io.agentscope.harness.agent.memory.MemoryBackgroundTasks.awaitQuiescence(
                     5, java.util.concurrent.TimeUnit.SECONDS);
+            // Drain session/transcript mirrors after background tasks have finished submitting
+            // them, so async workspace writes do not race with resource cleanup.
+            io.agentscope.harness.agent.memory.session.SessionTree.awaitMirrorQuiescence(
```
oss-maintainer, line 469: **[Info]** Non-blocking, and the reorder itself is correct — draining `MemoryBackgroundTasks` before `SessionTree.awaitMirrorQuiescence` is what makes the guarantee real, since a transcript append can submit mirror work. Two consequences of the drain chain worth one javadoc sentence on `close()`:

1. Worst case is now ~10s of blocking in `close()` (5s + 5s), and because both counters are process-wide, an agent closed on a request/HTTP thread can wait on writes started by *unrelated* agents. Callers that close agents inline will feel that as tail latency.
2. The drain stays best-effort: `doOnComplete` can fire after `close()` has entered the first await, so a task started during the drain is not covered by it. That is fine for the `@TempDir` teardown case this fixes (the new `@HarnessQuiescence` extension makes the ordering explicit), but it is not a general promise that no write outlives `close()`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/TranscriptMiddleware.java
```diff
@@ -65,17 +76,81 @@ public Flux<AgentEvent> onAgent(
...
+        }
+        return stream.doOnComplete(() -> scheduleTranscript(agent, rc));
+    }
+
+    private void scheduleTranscript(Agent agent, RuntimeContext rc) {
+        // AgentState is mutable and may be reused by a later turn, so capture this turn's
+        // defensive snapshot synchronously before detaching the transcript write from the
+        // stream. Moving this list copy into the background task could capture the next turn or
+        // miss messages from this one; only the transcript I/O is intentionally offloaded.
+        List<Msg> messages = snapshotMessages(agent, rc);
```
oss-maintainer, line 107: **[Info]** Resolved as far as I am concerned — I asked to either move the snapshot into `Mono.defer(...)` or document why the copy must stay on the stream thread, and the comment here plus the class javadoc do exactly that. `AgentState.getContext()` is `Collections.unmodifiableList(new ArrayList<>(context))`, i.e. a shallow reference copy, so the synchronous cost is O(n) on pointer widths rather than a deep clone of every `Msg`; deferring it really would capture the next turn.

The one thing this makes permanent is that the two paths now have different per-call cost: sandbox calls still pay transcript IO inside the stream (`concatWith`, line 84), non-sandbox calls no longer do. Worth a line in the class javadoc so nobody reads the PR title as "transcript IO is off the critical path" for sandbox deployments.

oss-maintainer (approved):
### Summary
Re-reviewed the force-pushed head `bf1acbb` (this work now lands as a single squashed commit). Both items from my previous approval checklist are resolved: the synchronous snapshot is justified in place inside `scheduleTranscript` rather than left implicit, and the cancellation contract is now stated in the class javadoc ("Cancellation intentionally discards the in-progress turn"). The `HarnessAgent.close()` reorder — draining `MemoryBackgroundTasks` before `SessionTree.awaitMirrorQuiescence` — plus the shared `@HarnessQuiescence` extension on `AgentSpawnToolPlanModeTest` is the ordering fix I pointed at on 2026-09-11 for the `DirectoryNotEmptyException` teardown failure, and it is the right shape: one explicit shared teardown contract instead of per-test timing workarounds.

### Findings
No blocking issues. Two non-blocking notes are inline (the process-wide drain cost now paid in `close()`, and the sandbox vs non-sandbox per-call cost asymmetry this PR makes permanent).

### Verification
- `TranscriptMiddlewareAsyncTest` and `SessionTranscriptWriterTest` cover completion ordering, snapshot-before-detach, in-stream sandbox writes, shutdown quiescence and failure isolation on both paths.
- `AgentState.getContext()` is `Collections.unmodifiableList(new ArrayList<>(context))` — a shallow defensive copy — which confirms keeping the snapshot on the terminal-signal thread is a correctness requirement, not an oversight.
- The `WRITE_LOCKS` refcount protocol is unchanged from the previous head and still correct: waiters are counted before blocking, and the unlock happens inside the map mutation, so no caller can publish a replacement lock for a path whose append is still in flight.
- CI is green on `bf1acbb` (ubuntu + windows builds, License, Module Sync, codecov) and `license/cla` reports signed; the PR is mergeable and blocked only on review.

LGTM — approving. Thanks for the patience across the force-pushes, and for turning the teardown flake into a shared test contract.

---

ningmao-hlyz:
@jujn 打扰了，可以看看这个pr嘛，目前处于approve状态

PR #3111, opened 2026-09-12
fix(harness): use zone-aware timestamp in memory flush header (#3088)

### LINKED ISSUE #3088: [Bug]:HarnessAgent memory timezone error

**Describe the bug**
HarnessAgent多次运行，memory目录下md文件内容，“Memory Flush”都是UTC时区

**To Reproduce**

1. You code：https://java.agentscope.io/v2/zh/docs/quickstart “第一个智能体”代码
2.  How to execute：运行几次，查看工程memory目录下的文件
3. See error：## Memory Flush — 2026-09-10T08:05:32.309657400Z

尝试加vm参数，memory文件记录时间还是UTC时区，不管用：-Duser.timezone=Asia/Shanghai

**Expected behavior**

我当前在中国东八区，期望的时间是2026-09-10T16:05:32.309657400Z

Coder-xiaosuo:
## AgentScope-Java Version

2.0.3-SNAPSHOT

### Background

Fixes #3088.

`MemoryFlushManager` appends a timestamped section to today's daily memory ledger
(`memory/YYYY-MM-DD.md`) on every flush. The section header was rendered with
`java.time.Instant.now().toString()`, and the file name date with `LocalDate.now()`.

### Root Cause

`Instant` is an absolute UTC point in time, so `Instant.toString()` *always* emits a
`...Z` suffix. As a result `-Duser.timezone=Asia/Shanghai` (or `TZ=...`) had no effect
on the header: a UTC+8 user expected `2026-09-10T16:05:32+08:00` but saw
`2026-09-10T08:05:32Z`. Additionally, the file name date used `LocalDate.now()` (the
default zone) while the header used `Instant`, so the two could disagree across a day
boundary.

### Changes

- **`MemoryFlushManager`**
  - Added an injectable `java.time.Clock` (new 4-arg constructor). Production default is
    `Clock.systemDefaultZone()`, so `-Duser.timezone` / `TZ` are honored; the existing
    2-arg and 3-arg constructors are preserved and delegate to it.
  - The section header is now rendered via
    `ZonedDateTime.now(clock).format(DateTimeFormatter.ISO_OFFSET_DATE_TIME)`, which
    preserves the local UTC offset.
  - `flushMemories` captures a single `ZonedDateTime now` and passes it down to
    `writeMemoryFiles`, so the daily file name date and the header timestamp always come
    from the same instant (no cross-day mismatch).
  - Extracted a package-private `formatTimestamp(Clock)` helper so the behavior is
    directly unit-testable.
- **Tests** (`MemoryFlushManagerTest`) — existing tests kept untouched; added:
  - `formatTimestamp_rendersClockZoneOffset` — parameterized over UTC / Asia/Shanghai /
    America/New_York / a cross-day case, asserting the rendered string and that
    `OffsetDateTime.parse(actual).toInstant()` equals the original `Instant`.
  - `formatTimestamp_honorsSystemDefaultZone` — verifies the production default
    (`Clock.systemDefaultZone()`) picks up the JVM default zone.
  - `flushMemories_writesZoneAwareHeaderAndMatchingFileName` — end-to-end: with a fixed
    clock at `2026-09-10T20:00:00Z` in `Asia/Shanghai`, asserts the ledger file is named
    `memory/2026-09-11.md` and the header is `## Memory Flush — 2026-09-11T04:00:00+08:00`.
- **Docs** — `docs/v2/en/docs/harness/memory.md` and `docs/v2/zh/docs/harness/memory.md`
  now document that the header timestamp is a zone-aware ISO-8601 offset date-time
  (previously always `...Z`) and that it still round-trips via `OffsetDateTime.parse`.

### Note on the exact format

`DateTimeFormatter.ISO_OFFSET_DATE_TIME` strips trailing zeros from the fractional
second, so `.309657400` is rendered as `.3096574`. The offset and the instant it parses
back to are unaffected. We kept this standard formatter (rather than a custom 9-digit
pattern) so the value remains a plain ISO-8601 offset date-time.

### Compatibility

- On-disk format changes from `...Z` to a local offset (`...+08:00` / `...-04:00`).
- Consumers parsing with `OffsetDateTime.parse(actual).toInstant()` (or any ISO-8601
  offset parser) are unaffected — the instant is identical, no data ambiguity.
- Nothing else on disk changed; no public API is broken (constructors are additive).

### How to test

```bash
# Multi-zone matrix (all pass)
TZ=UTC               mvn -pl agentscope-harness -am test -Dtest='MemoryFlushManager*Test'
TZ=Asia/Shanghai     mvn -pl agentscope-harness -am test -Dtest='MemoryFlushManager*Test'
TZ=America/New_York  mvn -pl agentscope-harness -am test -Dtest='MemoryFlushManager*Test'

# JVM property path
mvn -pl agentscope-harness -am test -Dtest='MemoryFlushManager*Test' -Duser.timezone=Asia/Shanghai
```

A/B verification matrix:

| Group | Zone | Input Instant | Rendered output | Parses back to |
|---|---|---|---|---|
| Old | UTC | `2026-09-10T08:05:32.309657400Z` | `2026-09-10T08:05:32.309657400Z` | equal |
| New | UTC | same | `2026-09-10T08:05:32.3096574Z` | equal |
| New | Asia/Shanghai | same | `2026-09-10T16:05:32.3096574+08:00` | equal |
| New | America/New_York | same | `2026-09-10T04:05:32.3096574-04:00` | equal |
| New | Asia/Shanghai (cross-day) | `2026-09-10T20:00:00Z` | `2026-09-11T04:00:00+08:00` | equal |

`Fixes #3088`

## Checklist

````

### CODE REVIEW

File: agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java [hunk 1]
```diff
@@ -101,9 +103,37 @@ public MemoryFlushManager(WorkspaceManager workspaceManager, Model model) {
...
+    /**
+     * @param flushPrompt SYSTEM prompt for the extraction LLM call. {@code null} falls back to
+     *     {@link #DEFAULT_FLUSH_PROMPT}.
+     * @param clock time source used for the daily file name and the flushed section header. The
+     *     zone of this clock determines the rendered UTC offset, so {@code -Duser.timezone} is
+     *     honored by the production default ({@link Clock#systemDefaultZone()}). Tests can inject
+     *     {@link Clock#fixed} to pin an instant.
+     */
+    public MemoryFlushManager(
+            WorkspaceManager workspaceManager, Model model, String flushPrompt, Clock clock) {
```
oss-maintainer, line 117: **[Warning]** The new `Clock` overload is not reachable from production wiring: the three construction sites still use the 2-/3-arg constructors — `HarnessAgent.java:1078`, `CompactionMiddleware.java:107`, `MemoryFlushMiddleware.java:269`. So in a real run the zone can only ever be the JVM default, and `Clock.systemDefaultZone()` is effectively fixed at construction time.

Fixing the header rendering is the right first step, but for #3088 to be *configurable* (rather than "whatever `-Duser.timezone` happens to be") it would help to plumb a `ZoneId`/`Clock` through `MemoryFlushMiddleware`/`HarnessAgent.Builder` and pass it down here. Happy to see that as a follow-up issue rather than in this PR — just want to make sure the injectable clock does not stay test-only.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
```diff
@@ -101,9 +103,37 @@ public MemoryFlushManager(WorkspaceManager workspaceManager, Model model) {
...
+     * {@code 2026-09-10T16:05:32.3096574+08:00}. Unlike {@code Instant.toString()} (which always
+     * renders {@code ...Z}), the offset is preserved so the value round-trips through
+     * {@link java.time.OffsetDateTime#parse(CharSequence)} back to the original instant.
+     */
+    static String formatTimestamp(Clock clock) {
+        return formatTimestamp(ZonedDateTime.now(clock));
+    }
+
+    private static String formatTimestamp(ZonedDateTime now) {
+        return now.format(DateTimeFormatter.ISO_OFFSET_DATE_TIME);
```
oss-maintainer: **[Warning]** Two things worth pinning down before this format ships in the on-disk ledger:

1. The rendered value is no longer `Instant.parse`-compatible. `Instant.parse("2026-09-10T16:05:32.3096574+08:00")` throws, so any consumer (Studio/export tooling/user scripts) that parses the `## Memory Flush — ` header as an `Instant` will break on the new output. Nothing in this repo parses the header today, so the risk is downstream only, but it is a user-visible storage-format change.
2. `ISO_OFFSET_DATE_TIME` drops trailing zeros in the fractional part, so the timestamp width is variable (`.3096574`, `.3`, or none at all for a whole second). Anything doing fixed-width regex/lexicographic sorting on the header will be surprised.

If stable output matters more than formatter brevity, an explicit pattern such as `DateTimeFormatter.ofPattern("uuuu-MM-dd'T'HH:mm:ss.SSSXXX")` gives fixed 3-digit millis and a preserved offset. Otherwise please add a line to the changelog / migration notes calling out the header format change.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
```diff
@@ -222,15 +252,11 @@ public void offloadMessages(
...
-
+    private void writeMemoryFiles(RuntimeContext rc, String content, ZonedDateTime now) {
         String dailyEntry =
-                String.format(
-                        "\n## Memory Flush — %s\n%s\n",
-                        java.time.Instant.now().toString(), content);
+                String.format("\n## Memory Flush — %s\n%s\n", formatTimestamp(now), content);
 
-        String dailyRelPath = WorkspaceConstants.MEMORY_DIR + "/" + today + ".md";
+        String dailyRelPath = WorkspaceConstants.MEMORY_DIR + "/" + now.toLocalDate() + ".md";
```
oss-maintainer: **[Warning]** This is only half of the daily-ledger write path. `MemorySaveTool.memorySave` (`agentscope-harness/.../tool/MemorySaveTool.java:76-79`) appends into the *same* `memory/YYYY-MM-DD.md` file and still does:

```java
String today = LocalDate.now().format(DateTimeFormatter.ISO_LOCAL_DATE);
...
"\n## Memory Save — %s\n%s\n", Instant.now().toString(), ...
```

That path still has exactly the bug #3088 describes (file name from the default-zone date, header from a UTC `Instant` — they disagree across a local midnight), and after this PR a single ledger file will contain two different timestamp conventions side by side.

Consider extracting the zone-aware helper (e.g. a small `MemoryTimestamps.now(clock)` used by both) or filing a follow-up so the two writers stay consistent.

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerTest.java
```diff
@@ -76,6 +84,59 @@ void flushMemories_summaryOnlyInputDoesNotCallModel() {
...
+
+        assertEquals(expected, actual);
+        assertEquals(instant, OffsetDateTime.parse(actual).toInstant());
+    }
+
+    @Test
+    void formatTimestamp_honorsSystemDefaultZone() {
+        TimeZone original = TimeZone.getDefault();
+        try {
+            TimeZone.setDefault(TimeZone.getTimeZone("Asia/Shanghai"));
```
oss-maintainer: **[Warning]** Mutating the JVM-global default time zone from a unit test is a flake source: `TimeZone.setDefault(...)` leaks to every other test in the same fork, and Surefire/JUnit concurrent execution can run them while it is set. The `finally` restore only helps if nothing else runs concurrently.

Suggested options, cheapest first:
- annotate the method with `@ResourceLock("user.timezone")` (and `@Execution(SAME_THREAD)`), or
- replace the mutation with an assertion that does not need it, e.g. `assertEquals(ZoneId.systemDefault(), Clock.systemDefaultZone().getZone())` plus a fixed-instant check through `ZonedDateTime.now(Clock.system(ZoneId.of("Asia/Shanghai")))`.

Also note the CI signal here is still missing — the only check reported on this PR is `license/cla`, so the test suite has not actually run yet (workflows for first contributions normally need maintainer approval). Could you confirm `mvn -pl agentscope-harness test -Dtest=MemoryFlushManager*Test` is green locally under the different `TZ` values you listed?

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerTest.java
```diff
@@ -76,6 +84,59 @@ void flushMemories_summaryOnlyInputDoesNotCallModel() {
...
+    // ISO_OFFSET_DATE_TIME strips trailing zeros from the fractional second, so .309657400 is
+    // rendered as .3096574; the offset and the instant it parses back to are unaffected.
+    @ParameterizedTest
+    @CsvSource({
+        "UTC,              2026-09-10T08:05:32.309657400Z, 2026-09-10T08:05:32.3096574Z",
+        "Asia/Shanghai,    2026-09-10T08:05:32.309657400Z, 2026-09-10T16:05:32.3096574+08:00",
+        "America/New_York, 2026-09-10T08:05:32.309657400Z, 2026-09-10T04:05:32.3096574-04:00",
+        "Asia/Shanghai,    2026-09-10T20:00:00Z,           2026-09-11T04:00:00+08:00"
+    })
+    void formatTimestamp_rendersClockZoneOffset(String zoneId, String instantStr, String expected) {
```
oss-maintainer: **[Info]** Nice coverage on the helper and the end-to-end file-name/header pairing. One gap: every new assertion goes through the package-private `formatTimestamp(Clock)` or the explicit 4-arg constructor. There is no regression test that pins the *reported* symptom — the 2-arg production constructor honouring the configured zone. That is the path a future refactor is most likely to silently break. Something like "default constructor + fixed default zone ⇒ header offset matches `ZoneId.systemDefault()`" would close it (the `honorsSystemDefaultZone` case above covers the helper, not the constructor wiring).

oss-maintainer (commented):
### Summary

Zone-aware flush header + injectable `Clock` is a correct fix for the `Instant.now()`/`LocalDate.now()` mismatch in #3088, and the multi-zone test matrix is welcome. Not approving yet for two reasons: (1) no CI signal — the only status on this head is `license/cla`, so the suite has not actually run; (2) the change covers only half of the ledger write path and changes the on-disk timestamp format, which deserves a closer look.

### Findings
- **[Warning]** `MemoryFlushManager.java:118` — the new `Clock` overload is never used by production construction sites (`HarnessAgent.java:1078`, `CompactionMiddleware.java:107`, `MemoryFlushMiddleware.java:269`), so the zone is still only whatever the JVM default happens to be.
- **[Warning]** `MemoryFlushManager.java:136` — `ISO_OFFSET_DATE_TIME` output is not `Instant.parse`-compatible and has variable fractional-second width; this is a user-visible storage format change.
- **[Warning]** `MemoryFlushManager.java:259` — `MemorySaveTool` writes into the same `memory/YYYY-MM-DD.md` ledger and still has the original `LocalDate.now()` + `Instant.now()` mismatch, so one file will end up with two timestamp conventions.
- **[Warning]** `MemoryFlushManagerTest.java:109` — mutating the JVM-global default time zone in a unit test is a flake risk under concurrent execution; please confirm the suite is green locally.
- **[Info]** `MemoryFlushManagerTest.java:96` — no regression test pins the 2-arg production constructor honoring the configured zone.

### Suggestions
- Use an explicit pattern such as `uuuu-MM-ddTHH:mm:ss.SSSXXX` if a stable width is preferable, or document the header format change.
- Consider a shared `MemoryTimestamps`-style helper so `MemoryFlushManager` and `MemorySaveTool` stay consistent.
- `@ResourceLock("user.timezone")` instead of an unprotected `TimeZone.setDefault`.

---

Coder-xiaosuo:
Thanks for this fix — the zone-aware header and the multi-zone test matrix are
exactly the right direction for #3088.

I've also been working on the remaining half of this ledger write path, and I
wanted to check with you before opening a separate PR, to avoid a duplicate or
competing fix.

What I have locally:
- [one-line description of your change, e.g. makes MemorySaveTool.memorySave use
  the same zone-aware timestamp convention as MemoryFlushManager, so a single
  memory/YYYY-MM-DD.md no longer contains two different timestamp formats]
- [if applicable: extracts a shared MemoryTimestamps-style helper used by both
  writers]
- [if applicable: plumbs a ZoneId/Clock through MemoryFlushMiddleware /
  HarnessAgent.Builder so the zone is actually configurable rather than just
  the JVM default]

Two questions:

1. Is this something you'd prefer as a follow-up PR based on main once #3111
   lands, or would you rather keep it inside #3111 to avoid a second file
   touching the same write path?
2. If a separate PR is fine, should I wait for #3111 to merge first so I'm not
   basing on unmerged changes?

Happy to hold off or adjust the scope — just want to make sure I'm not
duplicating what you already have planned.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryFlushManager.java
```diff
@@ -119,8 +134,8 @@ public Mono<Void> flushMemories(RuntimeContext rc, List<Msg> messages) {
         }
 
         String existingMemory = readExistingContent(rc, WorkspaceConstants.MEMORY_MD);
-        String today = LocalDate.now().format(DateTimeFormatter.ISO_LOCAL_DATE);
-        String dailyRelPath = WorkspaceConstants.MEMORY_DIR + "/" + today + ".md";
+        ZonedDateTime now = MemoryTimestamps.now(clock);
```
oss-maintainer, line 137: `now` is sampled before the extraction model call, so the ledger entry (and its file name) is stamped with the flush *start*, which can be minutes earlier than the write. That is a deliberate improvement over the old code — one `ZonedDateTime` feeds both the file name and the header, so they can no longer disagree across a day boundary — but the semantic is now "flush started at" rather than "written at". One sentence in the `flushMemories` javadoc would make that explicit for anyone debugging a midnight-boundary ledger.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryTimestamps.java
```diff
@@ -0,0 +1,63 @@
...
+ * {@code Instant#toString()}, which always renders {@code ...Z} and ignores the JVM zone, so
+ * {@code -Duser.timezone} / {@code TZ} are honored.
+ *
+ * <p>{@link DateTimeFormatter#ISO_OFFSET_DATE_TIME} trims trailing zeros from the fractional
+ * second ({@code .309657400} renders as {@code .3096574}). The value is lossless and round-trips
+ * through {@link java.time.OffsetDateTime#parse(CharSequence)}; note it is an offset date-time, not
+ * an {@code Instant}-only string, so parse it with {@code OffsetDateTime} (or {@code ZonedDateTime})
+ * rather than {@link java.time.Instant#parse(CharSequence)}.
+ */
+public final class MemoryTimestamps {
```
oss-maintainer, line 40: This is a new `public` type in `agentscope-harness`, so it becomes part of the module's API surface (the `agentscope-core` interface cascade rule applies to harness too). Today both callers are inside this module — consider package-private, or state in the javadoc that it is a supported extension point so custom writers reuse the same convention instead of re-implementing it.

oss-maintainer (commented):
### Summary
Re-reviewed the new head. The previous gap is closed: `MemorySaveTool` now shares `MemoryTimestamps`, so one `memory/YYYY-MM-DD.md` can no longer mix `...Z` and offset timestamps, and the file name and the section header come from the same `ZonedDateTime` — that is exactly the #3088 defect. Docs updated in both `en` and `zh`, and `MemoryTimestampsTest` pins the format across three zones plus the fractional-second trimming quirk. Holding off on approval for one reason only: the sole status on this head is `license/cla`, so there is still no CI signal for the harness module.

On the coordination question in the thread: the ledger write path is already fully covered by this PR, so no second PR is needed for it. If you want per-agent zone configuration (a `ZoneId`/`Clock` plumbed through `HarnessAgent.Builder` and the tool registration instead of the JVM default), send that as a follow-up based on `main` once this lands — keeping #3111 a focused bug fix is the better split.

---

agentscope-harness/src/test/java/io/agentscope/harness/agent/memory/MemoryFlushManagerTest.java
```diff
@@ -76,6 +82,61 @@ void flushMemories_summaryOnlyInputDoesNotCallModel() {
...
+                content.contains("## Memory Flush — 2026-09-11T04:00:00+08:00"),
+                "header should carry the clock's offset: " + content);
+        assertFalse(content.contains("2026-09-10T20:00:00Z"), content);
+    }
+
+    // Regression for #3088: the production 2-arg constructor must honor the JVM default zone
+    // instead of silently falling back to UTC. The format contract itself is covered by
+    // MemoryTimestampsTest.
+    @Test
+    @ResourceLock("user.timezone")
```
oss-maintainer, line 112: `TimeZone.setDefault(...)` is still JVM-global, and `@ResourceLock` only serialises tests that declare the same lock — any other test in the module that reads the default zone (memory/session naming, transcript timestamps) can still observe `Asia/Shanghai` while this one runs if the suite is ever switched to parallel execution. Safer variants: assert against a `Clock.fixed(instant, ZoneId.of("Asia/Shanghai"))` for the offset behaviour and keep only one narrow test that verifies `Clock.systemDefaultZone()` is what the 2-arg constructor selects (e.g. by asserting the rendered offset equals `ZoneId.systemDefault()` computed inside the test, without mutating it).

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/MemorySaveTool.java
```diff
@@ -40,9 +40,19 @@
...
 
     public MemorySaveTool(WorkspaceManager workspaceManager) {
+        this(workspaceManager, Clock.systemDefaultZone());
+    }
+
+    /**
+     * @param clock time source for the daily ledger file name and section header. Defaults to
+     *     {@link Clock#systemDefaultZone()} so {@code -Duser.timezone} / {@code TZ} are honored.
+     */
+    public MemorySaveTool(WorkspaceManager workspaceManager, Clock clock) {
```
oss-maintainer, line 53: Good — sharing `MemoryTimestamps` with the flush path removes the mixed-format ledger I flagged on the previous head. Note the production wiring still calls the 1-arg constructor, so the zone remains the JVM default rather than a configured value; that is consistent with the scope of #3088, but if a per-agent/per-tenant zone is ever wanted it needs to be plumbed through the tool registration as well.

PR #3112, opened 2026-09-12
fix: fail early for incompatible Redisson state-store clients

### LINKED ISSUE #3063: [Bug]:Class org.redisson.api.RScript$ReturnType does not have member field 'org.redisson.api.RScript$ReturnType LONG'
## 问题描述

在使用 `RedisAgentStateStore` 进行 Agent 状态持久化测试时，运行抛出 `NoSuchFieldError`。
异常发生在框架内部 `RedissonClientAdapter.evalScript()` 方法，代码硬编码使用 `RScript.ReturnType.LONG`。
`agentscope 2.0.3` 的 jar 包通过 shade 内嵌了一份旧版 Redisson 类，内嵌版本的`RScript.ReturnType`枚举**不存在 LONG 常量**。
虽然项目 POM 已经引入 Redisson 3.52.0（该版本包含 LONG 枚举），但是 JVM 类加载优先加载 AgentScope jar 内自带的旧 Redisson 类，编译正常，运行时报字段缺失。业务代码无法修改框架源码。

## 复现步骤

1. 创建 SpringBoot 3.5.11 项目，pom 引入依赖：`io.agentscope:agentscope:2.0.3`、`redisson-spring-boot-starter:3.52.0`
2. 使用框架提供的 `RedisAgentStateStore` 实现 Agent 状态持久化
3. 启动项目，触发状态持久化逻辑，底层会调用 `RedissonClientAdapter.evalScript` 执行 Lua 脚本
4. 抛出 `NoSuchFieldError` 异常

## 预期行为

`RedisAgentStateStore` 正常执行 Lua 脚本完成 Agent 状态持久化；框架内嵌 Redisson 不应该和项目引入的 Redisson 发生类冲突。

## 报错信息

```
java.lang.NoSuchFieldError: No static field LONG of type Lorg/redisson/api/RScript$ReturnType;
	at io.agentscope.adapter.RedissonClientAdapter.evalScript(RedissonClientAdapter.java:228)
	at io.agentscope.storage.redis.RedisAgentStateStore.[相关持久化方法]
```

异常发生在虚拟线程，调用链路来自`RedisAgentStateStore`持久化流程。

## 运行环境

- AgentScope-Java Version：2.0.3
- Java Version：21
- OS：Windows

## 补充说明

1. Maven 依赖树显示项目依赖 Redisson 3.52.0，该版本`RScript.ReturnType`包含 LONG 枚举，版本本身没问题。
2. 冲突根本原因：`agentscope-2.0.3.jar`使用 shade 打包内嵌旧 Redisson，**未做包名 relocation 隔离**；Maven 依赖树无法识别这种内嵌在 jar 里的 class。
3. JVM 优先加载 agentscope 包内旧的`org.redisson.api.RScript$ReturnType`，该枚举缺少 LONG 常量。
4. 触发场景：`RedisAgentStateStore`持久化 Agent 状态，底层调用`RedissonClientAdapter.evalScript`执行 Lua 脚本触发异常。
5. 建议修复方案二选一：
   - 移除 shade 内嵌 Redisson，改用 Maven 正常依赖管理；
   - 对内置 Redisson 做包名重定向（relocation），避免与业务项目引入的 Redisson 产生类冲突。

mvanhorn:
## AgentScope-Java Version

In each existing Redisson constructor, resolve `RScript.ReturnType.valueOf("LONG")` into a final instance field and translate a missing constant into an actionable `IllegalStateException` explaining that this state-store integration requires the Redisson 4.x API, that 3.x is incompatible, and that the project currently manages 4.2.0. Preserve existing argument validation in the deprecated store, resolve the constant before any Redis operations, retain the original exception as the cause, and include the loaded Redisson API package implementation version when available with an explicit unknown fallback; capability detection, rather than parsing a nullable version string, decides compatibility. Use that stored return type in `RedissonClientAdapter.evalScript` and `RedissonAgentStateStore.evalSave`, covering both production script callers without a new helper, injectable resolver, static initializer, or INTEGER fallback.

The reporter encounters `NoSuchFieldError: RScript$ReturnType.LONG` when persisting agent state with AgentScope 2.0.3 and Redisson starter 3.52.0. The September 11 COLLABORATOR evaluation confirms the failure but corrects its cause: Redisson 3.x exposes INTEGER, while AgentScope's code targets the LONG constant introduced in 4.x; the evaluation reports no embedded `org/redisson/**` classes in the published aggregate. The local clone corroborates both LONG references and the BOM's Redisson 4.2.0 baseline, and its Redis extension already declares Redisson as a dependency. The thread proposes a construction-time guard and version documentation separately from an aggregate packaging decision, which remains pending.

Fixes #3063

## Description

In each existing Redisson constructor, resolve `RScript.ReturnType.valueOf("LONG")` into a final instance field and translate a missing constant into an actionable `IllegalStateException` explaining that this state-store integration requires the Redisson 4.x API, that 3.x is incompatible, and that the project currently manages 4.2.0. Preserve existing argument validation in the deprecated store, resolve the constant before any Redis operations, retain the original exception as the cause, and include the loaded Redisson API package implementation version when available with an explicit unknown fallback; capability detection, rather than parsing a nullable version string, decides compatibility. Use that stored return type in `RedissonClientAdapter.evalScript` and `RedissonAgentStateStore.evalSave`, covering both production script callers without a new helper, injectable resolver, static initializer, or INTEGER fallback.

## Checklist

- [ ] Code has been formatted with `mvn spotless:apply`
  Not run: no test command resolved in this workspace, so nothing was executed to pass.
  Not run: no test command resolved in this workspace, so nothing was executed to pass.

### DIFF
--- a/agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonAgentStateStore.java
+++ b/agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonAgentStateStore.java
@@ -66,6 +66,7 @@ public class RedissonAgentStateStore implements AgentStateStore {
     private static final String LIST_SUFFIX = ":list";
 
     private final RedissonClient redissonClient;
+    private final RScript.ReturnType scriptReturnType;
     private final String keyPrefix;
 
     private RedissonAgentStateStore(Builder builder) {
@@ -74,6 +75,19 @@ public class RedissonAgentStateStore implements AgentStateStore {
         }
         if (builder.redissonClient == null) {
             throw new IllegalArgumentException("RedissonClient cannot be null");
+        }
+        try {
+            this.scriptReturnType = RScript.ReturnType.valueOf("LONG");
+        } catch (IllegalArgumentException e) {
+            String version = RScript.class.getPackage().getImplementationVersion();
+            throw new IllegalStateException(
+                    "Redisson state-store integration requires the Redisson 4.x API"
+                            + " (RScript.ReturnType.LONG); Redisson 3.x is incompatible."
+                            + " Align your Redisson dependencies, including any starter, with"
+                            + " the project's currently managed version 4.2.0."
+                            + " Loaded Redisson API implementation version: "
+                            + (version == null ? "unknown" : version),
+                    e);
         }
         this.keyPrefix = builder.keyPrefix;
         this.redissonClient = builder.redissonClient;
@@ -150,7 +164,7 @@ public class RedissonAgentStateStore implements AgentStateStore {
                             .eval(
                                     RScript.Mode.READ_WRITE,
                                     RedisStateVersionSupport.SAVE_SCRIPT,
-                                    RScript.ReturnType.LONG,
+                                    scriptReturnType,
                                     new ArrayList<>(scriptKeys),
                                     scriptArgs.toArray());
             long newVersion = ((Number) result).longValue();
--- a/agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonClientAdapter.java
+++ b/agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonClientAdapter.java
@@ -119,8 +119,22 @@ import org.redisson.client.codec.StringCodec;
 public class RedissonClientAdapter implements RedisClientAdapter {
 
     private final RedissonClient redissonClient;
+    private final RScript.ReturnType scriptReturnType;
 
     private RedissonClientAdapter(RedissonClient redissonClient) {
+        try {
+            this.scriptReturnType = RScript.ReturnType.valueOf("LONG");
+        } catch (IllegalArgumentException e) {
+            String version = RScript.class.getPackage().getImplementationVersion();
+            throw new IllegalStateException(
+                    "Redisson state-store integration requires the Redisson 4.x API"
+                            + " (RScript.ReturnType.LONG); Redisson 3.x is incompatible."
+                            + " Align your Redisson dependencies, including any starter, with"
+                            + " the project's currently managed version 4.2.0."
+                            + " Loaded Redisson API implementation version: "
+                            + (version == null ? "unknown" : version),
+                    e);
+        }
         this.redissonClient = redissonClient;
     }
 
@@ -225,7 +239,7 @@ public class RedissonClientAdapter implements RedisClientAdapter {
                         .eval(
                                 RScript.Mode.READ_WRITE,
                                 script,
-                                RScript.ReturnType.LONG,
+                                scriptReturnType,
                                 keyObjects,
                                 args.toArray());
         if (result instanceof Number number) {
--- a/agentscope-extensions/agentscope-extensions-redis/src/test/java/io/agentscope/extensions/redis/state/redisson/RedissonClientAdapterTest.java
+++ b/agentscope-extensions/agentscope-extensions-redis/src/test/java/io/agentscope/extensions/redis/state/redisson/RedissonClientAdapterTest.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.extensions.redis.state.redisson;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.mockStatic;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.when;
+
+import io.agentscope.extensions.redis.state.RedisAgentStateStore;
+import io.agentscope.extensions.redis.state.RedisClientAdapter;
+import java.math.BigInteger;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.mockito.MockedStatic;
+import org.redisson.api.RScript;
+import org.redisson.api.RedissonClient;
+import org.redisson.client.codec.StringCodec;
+
+class RedissonClientAdapterTest {
+
+    private RedissonClient redissonClient;
+
+    @BeforeEach
+    void setUp() {
+        redissonClient = mock(RedissonClient.class);
+    }
+
+    @Test
+    void constructsAdapterAndUnifiedStoreWithoutContactingRedis() {
+        assertDoesNotThrow(() -> RedissonClientAdapter.of(redissonClient));
+        assertDoesNotThrow(
+                () -> RedisAgentStateStore.builder().redissonClient(redissonClient).build());
+        verifyNoInteractions(redissonClient);
+    }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {false, true})
+    void rejectsIncompatibleApiAtPublicConstructionBoundary(boolean throughBuilder) {
+        IllegalArgumentException cause = new IllegalArgumentException("No enum constant LONG");
+        try (MockedStatic<RScript.ReturnType> returnType = mockStatic(RScript.ReturnType.class)) {
+            returnType.when(() -> RScript.ReturnType.valueOf("LONG")).thenThrow(cause);
+
+            IllegalStateException error =
+                    assertThrows(
+                            IllegalStateException.class,
+                            () -> {
+                                if (throughBuilder) {
+                                    RedisAgentStateStore.builder().redissonClient(redissonClient);
+                                } else {
+                                    RedissonClientAdapter.of(redissonClient);
+                                }
+                            });
+
+            assertSame(cause, error.getCause());
+            assertTrue(error.getMessage().contains("requires the Redisson 4.x API"));
+            assertTrue(error.getMessage().contains("Redisson 3.x is incompatible"));
+            assertTrue(
+                    error.getMessage()
+                            .contains("Align your Redisson dependencies, including any starter"));
+            assertTrue(error.getMessage().contains("4.2.0"));
+            String version = RScript.class.getPackage().getImplementationVersion();
+            assertTrue(
+                    error.getMessage()
+                            .contains(
+                                    "Loaded Redisson API implementation version: "
+                                            + (version == null ? "unknown" : version)));
+            verifyNoInteractions(redissonClient);
+        }
+    }
+
+    @Test
+    void customAdapterDoesNotResolveRedissonReturnType() {
+        RedisClientAdapter client = mock(RedisClientAdapter.class);
+        try (MockedStatic<RScript.ReturnType> returnType = mockStatic(RScript.ReturnType.class)) {
+            assertDoesNotThrow(() -> RedisAgentStateStore.builder().clientAdapter(client).build());
+            returnType.verifyNoInteractions();
+            verifyNoInteractions(client);
+        }
+    }
+
+    @Test
+    void evalScriptDelegatesWithResolvedReturnTypeAndConvertsNumbers() {
+        RedissonClientAdapter adapter = RedissonClientAdapter.of(redissonClient);
+        RScript script = mock(RScript.class);
+        RScript.ReturnType longReturnType = RScript.ReturnType.valueOf("LONG");
+        when(redissonClient.getScript(StringCodec.INSTANCE)).thenReturn(script);
+        when(script.eval(
+                        RScript.Mode.READ_WRITE,
+                        "return ARGV[1]",
+                        longReturnType,
+                        List.of("key1", "key2"),
+                        "arg1",
+                        "arg2"))
+                .thenReturn(42L, 7, BigInteger.valueOf(9));
+
+        try (MockedStatic<RScript.ReturnType> returnType = mockStatic(RScript.ReturnType.class)) {
+            assertEquals(
+                    42L,
+                    adapter.evalScript(
+                            "return ARGV[1]", List.of("key1", "key2"), List.of("arg1", "arg2")));
+            assertEquals(
+                    7L,
+                    adapter.evalScript(
+                            "return ARGV[1]", List.of("key1", "key2"), List.of("arg1", "arg2")));
+            assertEquals(
+                    9L,
+                    adapter.evalScript(
+                            "return ARGV[1]", List.of("key1", "key2"), List.of("arg1", "arg2")));
+            returnType.verifyNoInteractions();
+        }
+        verify(redissonClient, times(3)).getScript(StringCodec.INSTANCE);
+        verify(script, times(3))
+                .eval(
+                        RScript.Mode.READ_WRITE,
+                        "return ARGV[1]",
+                        longReturnType,
+                        List.of("key1", "key2"),
+                        "arg1",
+                        "arg2");
+    }
+
+    @Test
+    void evalScriptRejectsNonnumericAndNullResults() {
+        RedissonClientAdapter adapter = RedissonClientAdapter.of(redissonClient);
+        RScript script = mock(RScript.class);
+        when(redissonClient.getScript(StringCodec.INSTANCE)).thenReturn(script);
+        when(script.eval(
+                        RScript.Mode.READ_WRITE,
+                        "return ARGV[1]",
+                        RScript.ReturnType.LONG,
+                        List.of("key"),
+                        "arg"))
+                .thenReturn("42", (Object) null);
+
+        assertEquals(
+                "Unexpected Lua script result: 42",
+                assertThrows(
+                                IllegalStateException.class,
+                                () ->
+                                        adapter.evalScript(
+                                                "return ARGV[1]", List.of("key"), List.of("arg")))
+                        .getMessage());
+        assertEquals(
+                "Unexpected Lua script result: null",
+                assertThrows(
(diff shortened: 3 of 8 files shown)


oss-maintainer (commented):
### Summary

Converts the Redisson 3.x/4.x `RScript.ReturnType.LONG` linkage failure into an actionable `IllegalStateException` raised at construction time, resolved once into a final field and reused by both production `eval` callers, plus regression tests and EN/ZH docs. The approach is sound — capability detection via `valueOf("LONG")` instead of version-string parsing is the right call, and the "no Redis interaction before the check" assertions are valuable coverage. No blocking correctness issue found; my comments are about keeping the diagnostic and the guidance from drifting.

### Findings
- **[Warning]** `RedissonClientAdapter.java:128` — `getPackage()` may be `null` and `getImplementationVersion()` is usually `null` for library jars, so the failure path can either NPE or print `unknown`; suggested a code-source fallback.
- **[Warning]** `RedissonClientAdapter.java:133` — `4.2.0` is hard-coded in a runtime message with nothing tying it to `agentscope-dependencies-bom`, so the guidance goes stale silently on the next bump.
- **[Warning]** `RedissonAgentStateStore.java:79-91` — the whole block including the message text is duplicated verbatim from the adapter; hoisting just the string constant respects the "no new helper" constraint while removing the drift risk.
- **[Info]** `RedissonClientAdapterTest.java:72` — the `IllegalStateException` surfaces from `Builder.redissonClient(...)`, not `build()`; the new failure mode should be documented with `@throws` on the two public entry points.
- **[Info]** `docs/.../redis.md:27` — the same two paragraphs now live in four files, one of which is marked superseded.

### Strengths
- `constructsWithoutContactingRedis` / `constructsAdapterAndUnifiedStoreWithoutContactingRedis` pin the "fail before touching the network" contract, which is the part users actually care about.
- `validatesPrefixBeforeResolvingReturnType` and `validatesClientBeforeResolvingReturnType` correctly lock the validation-before-capability-detection ordering, so argument errors stay `IllegalArgumentException`.
- `customAdapterDoesNotResolveRedissonReturnType` guards against the check leaking into non-Redisson adapters, and `saveIfVersionPreservesVersionAndConflictResults` verifies the constant is resolved once at construction rather than per call.
- Both call sites are covered and no remaining `RScript.ReturnType.LONG` reference exists in main sources, so the fix is complete for the reported stack trace.

### Questions
1. Is a `String`-code-source fallback acceptable for the "loaded Redisson API implementation version" field when the manifest omits `Implementation-Version`, or would you prefer it stay `unknown` and drop the code-source detail?
2. Do you want the `@throws IllegalStateException` Javadoc in this PR, or tracked separately?

---

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonClientAdapter.java
```diff
@@ -119,8 +119,22 @@
 public class RedissonClientAdapter implements RedisClientAdapter {
 
     private final RedissonClient redissonClient;
+    private final RScript.ReturnType scriptReturnType;
 
     private RedissonClientAdapter(RedissonClient redissonClient) {
+        try {
+            this.scriptReturnType = RScript.ReturnType.valueOf("LONG");
+        } catch (IllegalArgumentException e) {
+            String version = RScript.class.getPackage().getImplementationVersion();
```
oss-maintainer: `Class.getPackage()` is documented to return `null` for some class-loader/shaded layouts, and `getImplementationVersion()` returns `null` unless the Redisson manifest actually declares `Implementation-Version`. Both risks land on the *failure* path only, which is exactly where a good diagnostic matters most: an NPE here would replace the actionable `IllegalStateException` with a confusing stack trace, and in the common case the message just prints `unknown`.

Suggested hardening (keeps the spec's "no new helper / no static initializer" constraint):

```java
Package apiPackage = RScript.class.getPackage();
String version = apiPackage == null ? null : apiPackage.getImplementationVersion();
if (version == null) {
  // fall back to the code source so users can see which jar won
  version = String.valueOf(RScript.class.getProtectionDomain().getCodeSource());
}
```

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonClientAdapter.java
```diff
@@ -119,8 +119,22 @@
...
+            throw new IllegalStateException(
+                    "Redisson state-store integration requires the Redisson 4.x API"
+                            + " (RScript.ReturnType.LONG); Redisson 3.x is incompatible."
+                            + " Align your Redisson dependencies, including any starter, with"
+                            + " the project's currently managed version 4.2.0."
```
oss-maintainer: Hard-coding `4.2.0` inside the runtime message couples a user-facing diagnostic to `agentscope-dependencies-bom/pom.xml` (`<redisson.version>4.2.0</redisson.version>`) with nothing enforcing the link. When the BOM is bumped, the exception starts telling users to align to a version the project no longer manages, and the failure is silent: the only guard is `assertTrue(error.getMessage().contains("4.2.0"))` in two test classes, so a maintainer has to remember to edit three places plus four docs files.

Consider wording that doesn't pin the number, e.g. `"... with the Redisson version managed by agentscope-dependencies-bom."`, or keep the literal but source it from a single package-private `static final String` shared by both constructors. Either way the guidance stays accurate without a helper class.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonAgentStateStore.java
```diff
@@ -75,6 +76,19 @@ private RedissonAgentStateStore(Builder builder) {
...
+            String version = RScript.class.getPackage().getImplementationVersion();
+            throw new IllegalStateException(
+                    "Redisson state-store integration requires the Redisson 4.x API"
+                            + " (RScript.ReturnType.LONG); Redisson 3.x is incompatible."
+                            + " Align your Redisson dependencies, including any starter, with"
+                            + " the project's currently managed version 4.2.0."
+                            + " Loaded Redisson API implementation version: "
+                            + (version == null ? "unknown" : version),
+                    e);
+        }
```
oss-maintainer, line 90: This 13-line block (including the full message text) is a verbatim copy of the one added to `RedissonClientAdapter`. I understand the spec deliberately ruled out a new helper / injectable resolver / static initializer, so I'm not asking to restructure the control flow — but the *message string* is not a helper. Two independent copies asserted by two independent test classes will drift on the next wording or version change, and the drift will only surface in a user-facing stack trace.

Minimal option that respects the constraint: hoist just the message into one package-private `static final String INCOMPATIBLE_REDISSON_API = "..."` (or a `RedisStateVersionSupport` constant, which both classes already can see) and reference it from both constructors. `RedissonAgentStateStore` could also simply delegate to the adapter it already wraps if that's not blocked by the deprecation rules.

agentscope-extensions/agentscope-extensions-redis/src/test/java/io/agentscope/extensions/redis/state/redisson/RedissonClientAdapterTest.java
```diff
@@ -0,0 +1,176 @@
...
+        IllegalArgumentException cause = new IllegalArgumentException("No enum constant LONG");
+        try (MockedStatic<RScript.ReturnType> returnType = mockStatic(RScript.ReturnType.class)) {
+            returnType.when(() -> RScript.ReturnType.valueOf("LONG")).thenThrow(cause);
+
+            IllegalStateException error =
+                    assertThrows(
+                            IllegalStateException.class,
+                            () -> {
+                                if (throughBuilder) {
+                                    RedisAgentStateStore.builder().redissonClient(redissonClient);
```
oss-maintainer, line 72: This assertion documents an important behaviour change worth making visible in the public API docs: because `RedisAgentStateStore.Builder.redissonClient(...)` eagerly calls `RedissonClientAdapter.of(...)`, the new `IllegalStateException` surfaces from the *builder setter*, not from `build()`. That is fine for `@Bean`-style wiring, but it is a new failure mode on a public method that previously could not throw for this reason.

Could you add `@throws IllegalStateException` (plus one sentence about the Redisson 4.x baseline) to the Javadoc of `RedissonClientAdapter.of(...)` and `RedisAgentStateStore.Builder.redissonClient(...)`? `of(...)` currently has no `@throws` tag at all.

docs/v2/en/integration/distributed/redis.md
```diff
@@ -14,7 +14,17 @@ title: Redis
...
+
+```xml
+<dependency>
+    <groupId>org.redisson</groupId>
+    <artifactId>redisson</artifactId>
+    <version>4.2.0</version>
+</dependency>
+```
+
+If you manage your own Redisson Spring Boot starter, a 3.x starter (for example, 3.52.0) is incompatible with this state-store integration. Align the starter and its Redisson dependencies with the current 4.2.0 baseline and check the resolved runtime dependencies. Both `RedisAgentStateStore.builder().redissonClient(...)` and the deprecated `RedissonAgentStateStore` fail during construction when the loaded API lacks `LONG`. This check reports the incompatibility; it does not repair dependency resolution or make Redisson 3.x persistence supported.
```
oss-maintainer, line 27: Nice that the Chinese page was updated in lockstep. One maintainability note: these two paragraphs now exist in four files (`distributed` + `session`, EN + ZH), and `session/redis.md` is explicitly marked *"This page has been superseded by Distributed Storage — Redis"*. Duplicating guidance onto a superseded page means the next compatibility change has four copies to remember (and the `4.2.0` literal in the snippet adds two more).

Since the session page already redirects, consider replacing the block there with a one-line note plus a link to `/v2/en/integration/distributed/redis`, keeping the full text in exactly one EN and one ZH file.

mvanhorn:
Thanks, that is a much better diagnostic. All five addressed:

`RedissonClientAdapter.java:128` - `getPackage()` is now null-checked, and when the manifest carries no `Implementation-Version` the message falls back to `RScript.class.getProtectionDomain().getCodeSource()` so the user can at least see which jar won. Same treatment in the `RedissonAgentStateStore` copy.

`RedissonClientAdapter.java:133` - dropped the `4.2.0` literal. The message now points at "the Redisson version managed by agentscope-dependencies-bom", so a BOM bump cannot leave the runtime text stale.

`RedissonAgentStateStore.java:91` - you are right that the message string is not a helper. It is now a single package-private `RedissonClientAdapter.INCOMPATIBLE_REDISSON_API` constant used by both constructors, and both tests assert equality against that constant rather than against re-typed prose. The control flow is untouched, as the spec asked.

`RedissonClientAdapterTest.java:72` - added `@throws IllegalStateException` to both `RedissonClientAdapter.of(...)` and `RedisAgentStateStore.Builder.redissonClient(...)`, with a sentence on the latter noting the check fires from the setter rather than from `build()`.

`docs/.../distributed/redis.md:27` - the full Redisson 4.x guidance now lives only on the EN/ZH distributed pages. The superseded session pages carry a one-line pointer instead, so there are two copies to maintain rather than four, and no version literal in either.

`mvn -pl agentscope-extensions/agentscope-extensions-redis -am test` is green: 18 tests, 0 failures, spotless clean.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonClientAdapter.java
```diff
@@ -118,9 +118,28 @@
...
 
     private RedissonClientAdapter(RedissonClient redissonClient) {
+        try {
+            this.scriptReturnType = RScript.ReturnType.valueOf("LONG");
+        } catch (IllegalArgumentException e) {
+            Package apiPackage = RScript.class.getPackage();
+            String version = apiPackage == null ? null : apiPackage.getImplementationVersion();
+            if (version == null) {
+                // fall back to the code source so users can see which jar won
+                version = String.valueOf(RScript.class.getProtectionDomain().getCodeSource());
```
oss-maintainer, line 139: **[Info]** `String.valueOf(codeSource)` renders as the literal `"null"` when the class has no code source, and `CodeSource#toString()` is a `"(file:/...jar <no signers>)"` blob rather than a version — so the sentence reads "implementation version: (file:...)". Consider an explicit `unknown` fallback and a separate label, e.g. `"... implementation version: " + version + ", code source: " + location`, so users grepping for a version still get one. Non-blocking.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonAgentStateStore.java
```diff
@@ -75,6 +76,18 @@ private RedissonAgentStateStore(Builder builder) {
...
             throw new IllegalArgumentException("RedissonClient cannot be null");
         }
+        try {
+            this.scriptReturnType = RScript.ReturnType.valueOf("LONG");
+        } catch (IllegalArgumentException e) {
+            Package apiPackage = RScript.class.getPackage();
+            String version = apiPackage == null ? null : apiPackage.getImplementationVersion();
+            if (version == null) {
+                // fall back to the code source so users can see which jar won
+                version = String.valueOf(RScript.class.getProtectionDomain().getCodeSource());
```
oss-maintainer, line 86: **[Info]** Hoisting the message string fixed the drift risk, but the 4-line version-resolution fallback (package → `Implementation-Version` → code source) is now duplicated verbatim here and in `RedissonClientAdapter`. Both test classes also duplicate the same `loadedRedissonApiVersion()` helper. Since this class is deprecated, fine to leave as is — flagging only so a future edit doesn't diverge the two paths.

oss-maintainer (approved):
### Summary

Re-review of `8dcda477 → 7ac7282`. All five points from the previous round are addressed: `getPackage()` is null-guarded with a code-source fallback, the hard-coded `4.2.0` is replaced by a BOM reference, the duplicated message is hoisted into `RedissonClientAdapter.INCOMPATIBLE_REDISSON_API`, `@throws IllegalStateException` documents the early-failure boundary on `redissonClient(...)`, and the redundant Redisson paragraphs are collapsed into a link from the superseded session pages to `distributed/redis.md`. The tests now assert the exact composed message instead of a substring, which is the stronger form of the same guarantee. No blocking issue found — approving; the three items below are optional cleanups.

### Findings
- **[Info]** `RedissonClientAdapter.java:139` — code-source fallback prints `null` / a `CodeSource` blob where a version is expected.
- **[Info]** `RedissonAgentStateStore.java:86` — the version-resolution fallback is duplicated verbatim in the adapter and in both test helpers.

### Strengths
- Replacing the version-string literal with a BOM pointer removes the silent-staleness failure mode, and `validatesPrefixBeforeResolvingReturnType` / `customAdapterDoesNotResolveRedissonReturnType` keep the ordering and containment guarantees pinned.
- The docs collapse is the right direction: one canonical compatibility page instead of four copies that drift independently.

---

- **[INFO]** `agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonAgentStateStore.java:416` — **[Info]** `@throws IllegalStateException` was added to `RedisAgentStateStore.Builder.redissonClient(...)` and `RedissonClientAdapter.of(...)`, which is the right pair of entry points. `RedissonAgentStateStore.Builder.build()` throws the same way and its Javadoc is still silent — one line there would close the loop. _(line outside diff)_

mvanhorn:
Both Info items done in c19b7bc, since they were quick.

The code-source fallback no longer interpolates a CodeSource blob or the literal "null" into a sentence promising a version; it drops the version clause when it cannot determine one. And the package to Implementation-Version to code-source fallback now lives in one place instead of being duplicated across the adapter, the store and both test classes.

No change to the compatibility guard itself, so your approval should still hold.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/redisson/RedissonClientAdapter.java
```diff
@@ -132,11 +147,58 @@ private RedissonClientAdapter(RedissonClient redissonClient) {
...
+        return INCOMPATIBLE_REDISSON_API + LOADED_API_VERSION_CLAUSE + version;
+    }
+
+    /**
+     * Package implementation version of the loaded {@code org.redisson.api} classes.
+     *
+     * @return the implementation version, or {@code null} when none can be determined. A
+     *     code-source location is not a version and is never returned.
+     */
+    static String loadedRedissonApiVersion() {
```
oss-maintainer, line 190: **[Info]** Both earlier Info items are resolved here — dropping the `CodeSource` blob from the version slot and de-duplicating the resolution into one `static` helper is the right call, and `Implementation-Version` does resolve from the released jars (`redisson-4.2.0.jar` ships `Implementation-Version: 4.2.0`, and Boot/`BOOT-INF/lib` nested jars define the package from that manifest), so the omitted-clause path really is the rare case.

One non-blocking thought: the case where the version is blank is also the case where the classpath is most ambiguous (shaded/uber jar, duplicate `org.redisson.api` on the classpath) — exactly when users would still want to know which artifact won. If you want that signal back without re-introducing the `toString()` blob, keep it as its own labelled field rather than in the version slot, e.g. `" Loaded Redisson API code source: " + location` from `RScript.class.getProtectionDomain().getCodeSource()` guarded for `null` (that one can be `null` too), appended only when `loadedRedissonApiVersion()` returns `null`. Happy to see it left as-is — the message is unambiguous either way, and please don't let this block the PR.

oss-maintainer (approved):
### Summary

Re-reviewed the delta since my previous approval (`7ac72824` → `c19b7bc0`, "refactor: hoist Redisson version resolution and fix its fallback message"). It is exactly the two follow-ups I asked for, done well: the 4-line package/`Implementation-Version`/code-source fallback is hoisted into `RedissonClientAdapter.loadedRedissonApiVersion()` + `incompatibleRedissonApiMessage(String)` so the deprecated `RedissonAgentStateStore` no longer keeps a verbatim copy, and the diagnostic no longer interpolates a `CodeSource` blob or the literal `"null"` into a sentence — the version clause is simply omitted when the version is unknown. `INCOMPATIBLE_REDISSON_API` no longer ends in a dangling `" ... version: "`, which also removes the coupling of the user-facing text to a hard-coded `4.2.0`. The stale BOM sentence in the docs is gone and the "this check reports the incompatibility, it does not repair dependency resolution" caveat is present in both the EN and ZH pages.

Tests: the duplicated private `loadedRedissonApiVersion()` helpers were removed from both test classes and replaced with direct coverage of the new contract — blank/whitespace/null all collapse to the base message, a known version is appended verbatim, and `loadedApiVersionIsNeverACodeSourceLocation` pins the regression (no `<no signers>`, no `version: null`, and a `file:` location can never be returned) while staying tolerant of whatever version the test classpath actually carries. `assertFalse(...)` on the construction-boundary tests guards against the old formatting silently coming back.

One optional, non-blocking thought left as an inline note at `RedissonClientAdapter.java:190`: the blank-version case is also the case where the classpath is most ambiguous, so a separately labelled code-source field could restore the "which jar won" signal without putting a location in the version slot. Fine to skip.

Verified on this head: `license/cla` = signed, `Check License` / `Check Module Sync` / `build (ubuntu-latest)` / `build (windows-latest)` / `validate` / `codecov/patch` all passing, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`. Approving on `c19b7bc0`; merging is left to a maintainer.

---

PR #3113, opened 2026-09-12
fix(harness): keep memory flush/consolidation IO off the per-call sandbox

### LINKED ISSUE #3110: [Bug]:现在的记忆是异步写入，会有问题，写入的时候沙箱已经关闭，就会警告memory flush failed: no active sandbox, sandbox filesystem used outside of call context

**Describe the bug**

**To Reproduce**

1. You code
2. How to execute
3. See error

**Expected behavior**

**Error messages**
Detailed error messages.

- AgentScope-Java Version: [e.g. 1.0.12, check your pom.xml dependency version or run `mvn dependency:tree | grep agentscope-parent:pom`(only mac/linux)]
- Java Version: [e.g. 17]
- OS: [e.g. macos, windows]

**Additional context**

birdie7761:
Fixes #3110

## Description

The fire-and-forget memory flush (and the periodic consolidation) run **after** the agent event stream terminates — by then the sandbox lease may already be released, and every write through `SandboxBackedFilesystem` fails with `No active sandbox — sandbox filesystem used outside of a call context`. Per-call memory extraction was silently lost in sandbox-backed deployments.

Memory files (`MEMORY.md`, `memory/*.md`, the memory state) are cross-call metadata: like task records (#2803), their IO must not depend on a per-call sandbox instance. Following option (b) recorded on the issue (sandbox-independent storage), applied to the **whole memory family** so listing, reads and writes can never split between two stores:

- `WorkspaceManager` gains **memory-scoped IO** (`readMemoryFileUtf8` / `appendMemoryFileUtf8` / `writeMemoryFileUtf8`) that routes to the host workspace whenever the filesystem serving the path is the per-call sandbox proxy; `RoutedSandboxFilesystem` prefix routes still take over. `readMemoryMd` delegates to the routed read so a host-side write is never shadowed by a stale sandbox copy.
- **User isolation is preserved**: the host-side location is the context's runtime-data namespace (`resolveRuntimeDataPath`) — the same namespaced layout `RemoteFilesystem` applies to memory paths in distributed deployments. With no namespace factory the paths resolve to the workspace root, the location the no-filesystem mode has always used.
- **`listMemoryFileInfos`** is the single home for memory listing: under a sandbox-backed filesystem it enumerates the namespaced host memory directory (real sizes and mtimes); otherwise it globs the filesystem layer. The unified probe resolves `RoutedSandboxFilesystem` configurations (a bare `instanceof` in the consolidator would misroute them).
- `MemoryFlushManager` reads existing content and appends the daily ledger; `MemoryConsolidator` lists ledgers via the unified listing and rewrites `MEMORY.md` and the state watermark — **read/write symmetric on the same store** (an asymmetric watermark pair would silently re-merge all historical ledgers every run).
- The **memory tools** join the same store: `memory_save` appends and `memory_get` / `memory_search` read through the memory-scoped IO, and the search listing observes the same namespaced scope. Without this, tools would write into the per-lease sandbox copy while flush appends host-side — splitting one day's ledger across two stores and keeping tool-written memories out of consolidation forever.

**Behavior-change disclosure** (sandbox-backed deployments only; local and remote filesystem modes are untouched): memory files previously lived inside per-lease sandbox snapshots, where the post-stream flush could not write at all (the bug) and in-call tool writes died with the lease. The authoritative copies now live host-side under the runtime-data namespace; memory accumulated inside old sandbox snapshots becomes invisible (it was already unwritable post-stream, so the loss surface is the failed writes themselves).

**No file overlap** with #3094 (sandbox-mirror release path) or #3106 (transcript quiescence) — this fix composes with both.

## Test evidence (TDD)

New `MemoryFlushSandboxIndependenceTest` (7 cases), red on main for the three original cases:

| Case | Evidence |
|---|---|
| flush after stream ends writes the daily ledger host-side | red on main: no file lands (write swallowed); green: host file with content |
| consolidation reads ledgers and rewrites MEMORY.md host-side | red on main: throws through the dead proxy; green: host rewrite |
| MEMORY.md readable under a dead sandbox proxy | red: unreachable; green: host copy read |
| `memory_save` lands host-side, immediately readable by the family | no split storage |
| `RoutedSandboxFilesystem`-wrapping-dead-sandbox consolidates | the bare-instanceof misroute would be silent |
| namespaced deployment keeps real isolation | file lands under `workspace/<ns>/memory`, **no bare-root copy** |
| watermark persists — second consolidation is a no-op | counting model called exactly once across two runs |

Full harness suite: only pre-existing failures remain (`AguiPermissionResumeTest` 4/4 fails identically on pristine main — stash-verified; `honorsParentMemoryConfig` pair and the `DynamicHookBuilder` teardown are timing flakes that pass in isolation/re-run).

## Checklist

- [x] Related documentation has been updated (class/method javadocs document the routing, the isolation contract, and the deprecated-free `getMemoryDir` contract)

### DIFF
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryConsolidator.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryConsolidator.java
@@ -22,7 +22,6 @@ import io.agentscope.core.message.TextBlock;
 import io.agentscope.core.model.Model;
 import io.agentscope.harness.agent.filesystem.AbstractFilesystem;
 import io.agentscope.harness.agent.filesystem.model.FileInfo;
-import io.agentscope.harness.agent.filesystem.model.GlobResult;
 import io.agentscope.harness.agent.filesystem.remote.store.BaseStore;
 import io.agentscope.harness.agent.filesystem.remote.store.StoreItem;
 import io.agentscope.harness.agent.workspace.WorkspaceManager;
@@ -232,13 +231,17 @@ public class MemoryConsolidator {
             return "";
         }
 
-        GlobResult glob = fs.glob(rc, "*.md", "memory");
-        if (!glob.isSuccess() || glob.matches() == null || glob.matches().isEmpty()) {
+        // Listing delegates to the WorkspaceManager's single-home memory listing: under a
+        // sandbox-backed filesystem it enumerates the host-side authoritative copies (the
+        // sandbox glob would throw once the lease is released; routed configurations are
+        // resolved by the unified probe there, not a bare instanceof here).
+        List<FileInfo> matches = workspaceManager.listMemoryFileInfos(rc);
+        if (matches.isEmpty()) {
             return "";
         }
 
         List<FileInfo> eligible = new ArrayList<>();
-        for (FileInfo fi : glob.matches()) {
+        for (FileInfo fi : matches) {
             if (fi.isDirectory()) {
                 continue;
             }
@@ -255,7 +258,7 @@ public class MemoryConsolidator {
         StringBuilder sb = new StringBuilder();
         for (FileInfo fi : eligible) {
             String rel = toRelative(fi.path());
-            String content = workspaceManager.readManagedWorkspaceFileUtf8(rc, rel);
+            String content = workspaceManager.readMemoryFileUtf8(rc, rel);
             if (content != null && !content.isBlank()) {
                 sb.append("### ").append(fileName(fi.path())).append("\n");
                 sb.append(content.strip()).append("\n\n");
@@ -289,7 +292,7 @@ public class MemoryConsolidator {
     /**
      * Converts an absolute filesystem path (e.g. {@code /memory/2025-01-01.md}) to a
      * workspace-relative path ({@code memory/2025-01-01.md}) for use with
-     * {@link WorkspaceManager#readManagedWorkspaceFileUtf8}.
+     * {@link WorkspaceManager#readMemoryFileUtf8}.
      */
     private static String toRelative(String path) {
         if (path == null) {
@@ -299,7 +302,7 @@ public class MemoryConsolidator {
     }
 
     private void writeConsolidatedMemory(RuntimeContext rc, String content) {
-        workspaceManager.writeUtf8WorkspaceRelative(rc, "MEMORY.md", content);
+        workspaceManager.writeMemoryFileUtf8(rc, "MEMORY.md", content);
     }
 
     static final String STATE_REL_PATH = "memory/" + STATE_FILE;
@@ -328,7 +331,7 @@ public class MemoryConsolidator {
 
     private Instant readWatermarkFromFile(RuntimeContext rc) {
         try {
-            String value = workspaceManager.readManagedWorkspaceFileUtf8(rc, STATE_REL_PATH);
+            String value = workspaceManager.readMemoryFileUtf8(rc, STATE_REL_PATH);
             if (value == null || value.isBlank()) {
                 return Instant.EPOCH;
             }
@@ -347,7 +350,7 @@ public class MemoryConsolidator {
             writeWatermarkToStore(ts);
         }
         try {
-            workspaceManager.writeUtf8WorkspaceRelative(rc, STATE_REL_PATH, ts.toString());
+            workspaceManager.writeMemoryFileUtf8(rc, STATE_REL_PATH, ts.toString());
         } catch (Exception e) {
             log.warn(
                     "Failed to write consolidation watermark at {}: {}",
--- a/agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
+++ b/agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
@@ -257,9 +257,122 @@ public class WorkspaceManager implements AutoCloseable {
         return readWithOverride(rc, KNOWLEDGE_DIR + "/" + KNOWLEDGE_MD);
     }
 
-    /** Reads MEMORY.md content (two-layer: filesystem override, local fallback). */
+    /**
+     * Reads MEMORY.md content. Under a sandbox-backed filesystem this reads the host copy —
+     * memory files are cross-call metadata written by the fire-and-forget flush after the
+     * sandbox may be gone (#3110), so the host copy is the authoritative one; a stale sandbox
+     * copy must never shadow it.
+     */
     public String readMemoryMd(RuntimeContext rc) {
-        return readWithOverride(rc, MEMORY_MD);
+        return readMemoryFileUtf8(rc, MEMORY_MD);
+    }
+
+    /**
+     * Reads a memory-family UTF-8 file (e.g. {@code MEMORY.md}, {@code memory/&lt;date&gt;.md},
+     * memory state). Each family member follows the backend serving its own path — a routed
+     * persistent store when one is mounted for it (e.g. on the {@code memory/} prefix), the
+     * host runtime-data namespace when that backend is the per-call sandbox proxy (memory
+     * files are cross-call metadata written by the fire-and-forget flush after the sandbox
+     * may be gone, #3110), and the standard two-layer read otherwise.
+     *
+     * <p><b>Durability assumption:</b> host-side routing presumes the host workspace outlives
+     * the process or is shared across replicas. A stateless multi-replica deployment with
+     * pod-local workspaces needs the whole family on a shared persistent backend — either a
+     * non-sandbox persistent primary filesystem, or explicit routes covering both {@code
+     * memory/} and root-level {@code MEMORY.md} — otherwise a session whose next call is
+     * served by another replica sees no {@code MEMORY.md} and no ledgers.
+     */
+    public String readMemoryFileUtf8(RuntimeContext rc, String relativePath) {
+        String normalized = requireSafeRelativePath(relativePath);
+        if (normalized.isEmpty()) {
+            return "";
+        }
+        if (memoryFilesRouteToLiveSandbox(normalized)) {
+            Path hostPath = resolveRuntimeDataPath(rc, normalized);
+            if (!Files.isRegularFile(hostPath)) {
+                warnWhenSandboxLayerStillHolds(rc, normalized, hostPath);
+            }
+            return readFileQuietly(hostPath);
+        }
+        return readWithOverride(rc, normalized);
+    }
+
+    /**
+     * Best-effort upgrade visibility (#3110): on the first calls after upgrading a sandbox
+     * deployment the host copy does not exist yet while the pre-fix sandbox snapshot may
+     * still hold one. The sandbox copy is deliberately not adopted — it is stale by design,
+     * every post-stream write was already lost from it — so this only surfaces the situation
+     * for an operator who wants to copy the file host-side. When the sandbox layer itself is
+     * unreachable (post-lease) the probe is skipped silently: there is nothing left to adopt.
+     */
+    private void warnWhenSandboxLayerStillHolds(RuntimeContext rc, String relPath, Path hostPath) {
+        try {
+            if (filesystem.exists(rc, relPath)) {
+                log.warn(
+                        "Memory file {} exists in the sandbox layer but not host-side ({}); the"
+                                + " host copy is authoritative (#3110) and sandbox content is not"
+                                + " adopted — copy it host-side to keep it",
+                        relPath,
+                        hostPath);
+            }
+        } catch (Exception e) {
+            log.debug("Sandbox presence probe failed for {}: {}", relPath, e.getMessage());
+        }
+    }
+
+    /**
+     * Appends to a memory-family UTF-8 file. Host-side append when the filesystem serving the
+     * path is the per-call sandbox proxy (the fire-and-forget flush runs after the stream —
+     * and possibly the sandbox — is gone, #3110); otherwise the standard filesystem
+     * read-merge-write append.
+     */
+    public void appendMemoryFileUtf8(RuntimeContext rc, String relativePath, String content) {
+        if (relativePath == null || content == null) {
+            return;
+        }
+        String normalized = requireSafeRelativePath(relativePath);
+        if (normalized.isEmpty()) {
+            return;
+        }
+        if (memoryFilesRouteToLiveSandbox(normalized)) {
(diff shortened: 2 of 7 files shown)


agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
```diff
@@ -257,9 +257,85 @@ public String readKnowledgeMd(RuntimeContext rc) {
...
+     * Reads a memory-family UTF-8 file (e.g. {@code MEMORY.md}, {@code memory/&lt;date&gt;.md},
+     * memory state). Host-side only when the filesystem serving the path is the per-call
+     * sandbox proxy; otherwise the standard two-layer read (filesystem first, local fallback).
+     */
+    public String readMemoryFileUtf8(RuntimeContext rc, String relativePath) {
+        String normalized = requireSafeRelativePath(relativePath);
+        if (normalized.isEmpty()) {
+            return "";
+        }
+        if (memoryFilesRouteToLiveSandbox(normalized)) {
```
oss-maintainer, line 290: **[Warning]** Routing memory IO host-side trades "the post-stream write fails" for "the write lands on whatever disk this process owns". In a stateless multi-replica deployment the host workspace is pod-local, and the store-backed filesystems exist precisely because it is not shared — so a session whose next call is served by another replica sees no `MEMORY.md` and no ledgers, rather than seeing them late. Before this change the authoritative copy travelled with the sandbox snapshot / the configured store. Could you state the durability assumption this rests on? If a shared or persistent host workspace is a precondition, it belongs in this javadoc and in the distributed docs (next to #3094's mirror deferral and #3106's quiescence); if it is not, the memory family needs the same store/mirror path transcripts and session trees already use, and host-side is only correct for single-node setups.

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
```diff
@@ -931,15 +1037,67 @@ private static String requireSafeRelativePath(String relativePath) {
...
+     * eligibility filtering. Under a sandbox-backed filesystem the authoritative copies live
+     * host-side in the context's runtime-data namespaced memory directory — the same scope
+     * the memory-scoped IO reads and writes, mirroring the namespaced layout of remote
+     * filesystems — so the listing is host-side too (the sandbox glob would throw once the
+     * lease is released, and its copies are stale by design). Otherwise the filesystem
+     * layer is globbed. This is the single home for memory listing so callers (consolidator,
+     * tools) can never drift from the routing decision.
+     */
+    public List<FileInfo> listMemoryFileInfos(RuntimeContext rc) {
+        if (memoryFilesRouteToLiveSandbox(MEMORY_DIR)) {
```
oss-maintainer, line 1091: **[Warning]** The routing probe here is the **directory** (`memory`), but every other member of the family probes its own path — and `MEMORY.md` is a root-level path, not one under `memory/`. So with a route configured on the `memory/` prefix (which is exactly what `RoutedSandboxFilesystem` exists for), `memoryFilesRouteToLiveSandbox("memory")` follows the route while `readMemoryFileUtf8(rc, "MEMORY.md")` / `writeMemoryFileUtf8(rc, "MEMORY.md", ...)` resolve to the default sandbox backend, and the family splits across two stores: ledgers on the routed backend, `MEMORY.md` host-side. That is the split this PR removes, and each call is individually self-consistent, so it fails silently. Please resolve the decision once for the family — one canonical probe whose result is passed to the ledger and `MEMORY.md` calls — and add a test with a `memory/` prefix route to pin the intended behaviour. (Bare `memory` versus `memory/x.md` do agree with each other, via `routeForPath`'s equals branch; it is the `MEMORY.md` leg that diverges.)

oss-maintainer (commented):
### Summary

Fixes #3110 by taking the memory family off the per-call sandbox proxy: `WorkspaceManager` gains memory-scoped read / append / write that route to the host runtime-data namespace when the serving filesystem is the sandbox, `listMemoryFileInfos` becomes the single listing home, and flush, consolidation, the watermark and the three memory tools all go through it. The diagnosis is right and the design is the correct one — reads and writes are symmetric, the routed configuration is resolved by the unified probe instead of a bare `instanceof`, and the 7 new cases in `MemoryFlushSandboxIndependenceTest` (three of which you show red on `main`) cover exactly the paths that regressed. Holding at `COMMENT` because `build (ubuntu-latest)` and `build (windows-latest)` are still pending on `1f11efe7` and `mergeStateStatus` is `BLOCKED`; the three inline notes are about storage assumptions rather than defects in the routing code.

### Findings
- **[Warning]** `WorkspaceManager.java:280` — host-side is only equivalent to the old store if the host workspace is shared across replicas; on a per-pod workspace this trades a failed post-stream write for memory that the next replica cannot see. Needs a stated durability assumption (or the mirror path used for transcripts/session trees).
- **[Warning]** `WorkspaceManager.java:1051` — listing probes the `memory` directory while `MEMORY.md`, the ledgers and the watermark each probe their own path, so a `memory/` prefix route splits the family across two stores (ledgers on the routed backend, `MEMORY.md` host-side) and `listMemoryFilePaths` still reads `MEMORY.md` through the sandbox proxy — the same post-lease path #3110 is about. Resolve the decision once per family and add a routed-prefix test.
- **[Warning]** `WorkspaceManager.java:272` — on upgrade the host copy does not exist yet while the old sandbox copy does, so the first call after deployment starts from empty memory and the flush cements that. Consider a one-shot adoption, or at least a log line when a memory path is absent host-side but present in the sandbox.
- **[Info]** `WorkspaceManager.java:547` — `memoryFilesRouteToLiveSandbox` and `taskRecordsRouteToLiveSandbox` are now identical probes; either one method for both families, or per-family resolution as asked above.

### Tests
Coverage is strong for the sandbox case. The two gaps map to the findings: no case where `memory/` and `MEMORY.md` route to different backends, and no case for upgrading from a deployment whose memory only exists in an old sandbox snapshot.

### Cross-repo Note
The durability assumption above is the same one #3094 (sandbox-mirror release path) and #3106 (transcript quiescence) have to answer; if memory now writes host-side while those still mirror through the sandbox, the three paths need one documented rule about what lives where.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
```diff
@@ -257,9 +257,85 @@ public String readKnowledgeMd(RuntimeContext rc) {
...
+     * copy must never shadow it.
+     */
     public String readMemoryMd(RuntimeContext rc) {
-        return readWithOverride(rc, MEMORY_MD);
+        return readMemoryFileUtf8(rc, MEMORY_MD);
+    }
+
+    /**
+     * Reads a memory-family UTF-8 file (e.g. {@code MEMORY.md}, {@code memory/&lt;date&gt;.md},
+     * memory state). Host-side only when the filesystem serving the path is the per-call
```
oss-maintainer: **[Warning]** The javadoc says the host copy is "the authoritative one; a stale sandbox copy must never shadow it", which is right going forward — but on upgrade the host copy does not exist yet while the old one lives in a sandbox snapshot, so the first call after deployment starts from empty memory. Your behaviour-change disclosure covers this, and for post-stream writes the old copy was already unreachable, so the loss surface is limited to in-call tool writes that died with the lease. A cheap improvement: log once when a memory path is absent host-side while a store/sandbox copy exists, so a user reporting "the agent forgot everything after the upgrade" has something in the logs to find; a one-shot adoption (copy the store/sandbox content into the namespace when the host file is missing) would remove the surprise entirely if you want to go that far.

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
```diff
@@ -449,6 +530,25 @@ public void updateSessionIndex(
...
+     * proxy. Memory files ({@code MEMORY.md}, {@code memory/*.md}, the memory state) are
+     * cross-call metadata: the fire-and-forget flush and the periodic consolidation run AFTER
+     * the agent event stream terminates — by then the sandbox lease may already be released,
+     * and every write through {@code SandboxBackedFilesystem} would fail with "No active
+     * sandbox" (#3110). When this holds, memory-file IO is routed to the host workspace
+     * directly — the same location used when no filesystem layer is configured — so reads and
+     * writes stay symmetric (a host-side write is never shadowed by a stale sandbox copy on
+     * read). Explicit prefix routes are respected and still take over.
+     */
+    private boolean memoryFilesRouteToLiveSandbox(String relPath) {
```
oss-maintainer, line 587: **[Info]** `memoryFilesRouteToLiveSandbox` and `taskRecordsRouteToLiveSandbox` are now identical one-line delegations to `isLiveSandboxServingPath`, differing only in javadoc. That is fine as documentation, but the memory family still needs the per-family resolution described in the note on `listMemoryFileInfos` — with the two identical probes it is easy for a caller to pass a directory here and a file there and get different answers.

agentscope-harness/src/main/java/io/agentscope/harness/agent/workspace/WorkspaceManager.java
```diff
@@ -257,9 +257,122 @@ public String readKnowledgeMd(RuntimeContext rc) {
...
+     * memory/} and root-level {@code MEMORY.md} — otherwise a session whose next call is
+     * served by another replica sees no {@code MEMORY.md} and no ledgers.
+     */
+    public String readMemoryFileUtf8(RuntimeContext rc, String relativePath) {
+        String normalized = requireSafeRelativePath(relativePath);
+        if (normalized.isEmpty()) {
+            return "";
+        }
+        if (memoryFilesRouteToLiveSandbox(normalized)) {
+            Path hostPath = resolveRuntimeDataPath(rc, normalized);
```
oss-maintainer, line 291: **[Warning]** This probe makes an `exists` call through the filesystem layer on **every** host-side miss, and `readMemoryMd` is reached from `WorkspaceContextMiddleware.onSystemPrompt`, i.e. once per agent call. A sandbox deployment that has not flushed its first memory yet — and any deployment with memory hooks disabled but a caller still reading the family — pays one sandbox round-trip per call forever, not just during the upgrade window. `upgradeShape_sandboxCopyDoesNotShadowAbsentHostCopy` pins the cost at one interaction *per read*, which is the thing I am asking about rather than one per process. Consider a one-shot memo per relative path (a small `Set<String>` of already-probed paths guarded by the same lock, or an `AtomicBoolean` per family) so the operator gets the warning once and the steady-state read stays host-only. Non-blocking — the probe is correctly wrapped so a released lease degrades to `log.debug`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryConsolidator.java
```diff
@@ -232,13 +231,17 @@ private String readDailyEntries(RuntimeContext rc, Instant watermark) {
             return "";
         }
 
-        GlobResult glob = fs.glob(rc, "*.md", "memory");
-        if (!glob.isSuccess() || glob.matches() == null || glob.matches().isEmpty()) {
+        // Listing delegates to the WorkspaceManager's single-home memory listing: under a
+        // sandbox-backed filesystem it enumerates the host-side authoritative copies (the
+        // sandbox glob would throw once the lease is released; routed configurations are
+        // resolved by the unified probe there, not a bare instanceof here).
+        List<FileInfo> matches = workspaceManager.listMemoryFileInfos(rc);
```
oss-maintainer, line 238: **[Info]** With the listing delegated here, the `fs == null` early return just above is a leftover from when this method globbed the filesystem itself. It now reads as if consolidation requires a filesystem layer, while `listMemoryFileInfos` already handles `filesystem == null` — and handles it differently from its sibling: `listMemoryFilePaths` unions the host `getMemoryDir(rc)`, `listMemoryFileInfos` does not. So the javadoc's "single home for memory listing so callers can never drift from the routing decision" is true for the *routing* decision but the two entry points still disagree on the no-filesystem case, and consolidator callers cannot reach that disagreement because the guard short-circuits first. Either drop the guard here or have `listMemoryFileInfos` fall back to the host directory listing; one place, one answer.

oss-maintainer (approved):
### Summary

Re-review of `df3e8e1e` (branch was rebased/force-pushed since `1f11efe7`). All three previous findings are addressed, and the fixes are the right shape rather than band-aids:

- **Durability assumption** — now stated explicitly in the `readMemoryFileUtf8` javadoc, including the concrete failure mode (a stateless multi-replica deployment with pod-local workspaces needs the whole family on a shared persistent backend, or a session resumed on another replica sees neither `MEMORY.md` nor the ledgers).
- **Split family under a `memory/` prefix route** — each family member is now anchored on its own path, `listMemoryFilePaths` no longer probes `MEMORY.md` through the per-call sandbox proxy, and `routedMemoryDirPrefix_eachFamilyMemberFollowsItsOwnBackend` covers exactly the configuration that was missing.
- **Upgrade path** — `warnWhenSandboxLayerStillHolds` surfaces a sandbox-only copy instead of silently starting from empty, with a test that asserts the stale copy is *not* adopted. Adoption remains declined, and the comment now says why (every post-stream write was already lost from it), which is the answer I was looking for.
- The duplicated probe is collapsed into `isLiveSandboxServingPath`, with the per-family javadoc kept where the semantics differ.

I also checked the pieces the delta depends on: `listMemoryFileInfos` returns `FileInfo.ofFile(...)` with a real mtime so `isModifiedAfter` still filters against the watermark, the `.md` filter keeps `memory_state.ts` out of the eligible set (and the consolidator still filters by name), and the host paths go through `resolveRuntimeDataPath(rc, ...)` so the read side, the write side and `listMemoryFilePaths` all observe the same namespaced scope — which is what `namespacedDeployment_memorySaveThenSearchReadsBack` pins.

### Findings
- **[Warning]** `WorkspaceManager.java:291` — the existence probe runs on every host-side miss, and `readMemoryMd` is per agent call, so the cost is per call rather than per upgrade. Suggest a one-shot memo per path. Non-blocking.
- **[Info]** `MemoryConsolidator.java:238` — the `fs == null` guard is now vestigial and hides that `listMemoryFileInfos` and `listMemoryFilePaths` still disagree on the no-filesystem case.

### CI / CLA
`build (ubuntu-latest)` and `build (windows-latest)` both pass on `df3e8e1e`, `Check License` / `Check Module Sync` pass, CLA is signed. `mergeStateStatus` is `BLOCKED` only because the branch still needs an approving review.

Previous head held at `COMMENT` pending CI; CI is green, the routing code is symmetric and the regression coverage is strong, so approving. The two notes above are follow-up polish, not gates.

---

Issue #3085, opened 2026-09-10
[Bug]: After upgrading to version 2.0.3, there was an error when writing to the Redis cluster.


1065609785:
After upgrading to version 2.0.3, there was an error when writing to the Redis cluster.
**Describe the bug**
The AgentScope 2.0.2 version is operating normally. Only upgrading AgentScope to 2.0.3 without modifying any business or Redis cluster-related configurations resulted in errors when writing to the Redis cluster after the upgrade.

**To Reproduce**
The store is using this: DistributedStore store = RedisDistributedStore.fromJedis(getJedisCluster(), agentScopeProperties.getRedis().getKeyPrefix());

**Expected behavior**

**Error messages**
系统异常

java.lang.RuntimeException: Failed to save state if version: agent_state
	at io.agentscope.extensions.redis.state.RedisAgentStateStore.saveIfVersion(RedisAgentStateStore.java:289)
	at io.agentscope.core.ReActAgent.persistAgentStateCas(ReActAgent.java:557)
	at io.agentscope.core.ReActAgent.lambda$saveStateToSession$1(ReActAgent.java:485)
	at reactor.core.publisher.MonoRunnable.subscribe(MonoRunnable.java:49)
	at reactor.core.publisher.Mono.subscribe(Mono.java:4569)
	at reactor.core.publisher.MonoSubscribeOn$SubscribeOnSubscriber.run(MonoSubscribeOn.java:127)
	at io.opentelemetry.instrumentation.reactor.v3_1.ContextPropagationOperator$RunnableWrapper.run(ContextPropagationOperator.java:374)
	at reactor.core.scheduler.WorkerTask.call(WorkerTask.java:90)
	at reactor.core.scheduler.WorkerTask.call(WorkerTask.java:37)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:309)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
	at java.base/java.lang.Thread.run(Thread.java:1474)
Caused by: redis.clients.jedis.exceptions.JedisClusterOperationException: Keys must belong to same hashslot.
[...]

PR #3129, opened 2026-09-13
fix(agent): agent state redis key without same hashslot

rudy2steiner:
## Background

`RedisAgentStateStore` uses Lua scripts for versioned state writes and CAS updates. In Redis Cluster, every key passed to one Lua script must belong to the same hash slot. The old AgentState key layout built related keys independently, for example:

- `agentscope:session:userid/sessionId:agent_state`
- `agentscope:session:userid/sessionId:agent_state:ver`
- `agentscope:session:userid/sessionId:_keys`

These keys are logically part of the same session, but without a Redis hash tag they may be assigned to different cluster slots. That can make versioned saves fail in Redis Cluster when the Lua script receives keys from multiple slots. original bug issue report: https://github.com/agentscope-ai/agentscope-java/issues/3085
## How This Fix Works

This change introduces a centralized `RedisAgentStateKeyLayout` helper and uses Redis hash tags for new versioned state sessions.

New versioned state keys use a hash-tagged session segment:

- v0 legacy layout: `agentscope:session:userid/sessionId:agent_state`
- v1 cluster-safe layout: `agentscope:session:{userid/sessionId}:agent_state`

The hash tag keeps the state key, version key, and marker key in the same Redis Cluster hash slot while preserving the same logical `userid/sessionId` segment.

Compatibility behavior:

- Session layout resolution keeps existing v0 sessions on v0 and uses v1 for new sessions.
- List state uses the same resolved layout as the session instead of always using v0.
- `exists` checks the v1 marker first and only falls back to the v0 marker when the v1 marker is missing.
- `listSessionIds` scans and merges both v0 and v1 marker layouts.

## Changes

- Added `RedisAgentStateKeyLayout` to centralize Redis key construction, marker patterns, layout resolution, and session id parsing.
- Updated versioned state save/read/CAS paths to resolve the correct layout before building Redis keys.
- Updated session delete logic to use the same layout resolution as versioned state operations.
- Updated list save/read operations to use the same resolved layout as other session state.
- Updated session existence checks to handle both v1 and v0 marker keys.
- Updated session listing to scan both v1 hash-tagged marker keys and v0 legacy marker keys.
- Added focused tests for:
    - cluster-safe Lua keys for new sessions
    - legacy v0 session compatibility
    - list state preserving v0 for existing v0 sessions
    - list writes not moving an existing v1 session back to v0
    - delete behavior for v0 and v1 layouts
    - `exists` v1-first behavior and v0 fallback
    - listing mixed v0/v1 sessions without duplicate or brace leakage
    - key layout helper behavior


agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateStore.java
```diff
@@ -294,10 +303,10 @@ public long saveIfVersion(
 
     @Override
     public void save(String userId, String sessionId, String key, List<? extends State> values) {
-        String slotId = slotId(userId, sessionId);
-        String listKey = getListKey(slotId, key);
+        RedisAgentStateKeyLayout keyLayout = v0KeyLayout(userId, sessionId);
```
oss-maintainer: **[Critical] A list write flips an existing v1 session back to the v0 layout, making already-written v1 state invisible.**

`save(list)` always uses `v0KeyLayout(...)`, and the marker it writes (`addToSet(keysKey, ...)` at line 339) is therefore the **v0** `:_keys` key. But `RedisAgentStateKeyLayout.resolve()` decides the layout solely from `keyExists(v0._keys)`.

Sequence for a session created after this change:
1. `save(userId, sessionId, "state", value)` -> no v0 marker -> v1 layout; state + `:ver` + v1 `:_keys` written under `prefix{{user}/{sess}}:...`.
2. `save(userId, sessionId, "items", list)` -> writes v0 list keys **and creates the v0 `:_keys` marker**.
3. Any later `get(...)`/`getVersioned(...)`/`saveIfVersion(...)` now resolves to **v0** and reads `prefix{user}/{sess}:state`, which does not exist -> the state written in step 1 silently disappears (empty Optional / version reset), and the following `save` overwrites under a different key.
4. `delete(...)` also resolves to v0 afterwards, so the v1 `state`/`ver`/`_keys` keys are never removed -> permanent orphan keys.

This is data loss in a normal usage pattern (versioned state + message list for the same session), and it also means the layout a session ends up in depends on write order. Suggested fix: keep the layout decision per key-family rather than per session — i.e. list keys always v0 and *only* presence of a v0 versioned state/`_keys` marker drives versioned reads — or put list state under the same resolved layout as versioned state so the v0 marker can never be created by a list write for a v1 session. Please also add a regression test for the interleaved sequence `save(single) -> save(list) -> getVersioned(single)`.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateKeyLayout.java
```diff
@@ -0,0 +1,250 @@
...
+     * @param client Redis client used to test whether the v0 marker exists
+     * @param keyPrefix Redis key prefix configured for the state store
+     * @param userId original user id, nullable
+     * @param sessionId session id, required
+     * @return resolved layout for the session
+     */
+    static RedisAgentStateKeyLayout resolve(
+            RedisClientAdapter client, String keyPrefix, String userId, String sessionId) {
+        RedisAgentStateKeyLayout v0 = v0(keyPrefix, userId, sessionId);
+        if (client.keyExists(v0.getKeysKey())) {
```
oss-maintainer, line 65: **[Warning] Layout probing adds one `EXISTS` round trip to every versioned read/write, and is not concurrency-safe.**

`resolve()` is called from `saveVersioned`, `getVersioned`, `saveIfVersion`, `get`, `delete` — so every state operation now costs an extra round trip before the real work, on a path that is typically hit once per agent turn. Two consequences worth documenting/optimising:

- **Cost:** for standalone/sentinel deployments that never had this key, a per-session layout cache (bounded, or a one-shot flag stored on the session) avoids paying `EXISTS` on every call.
- **Race:** two threads writing the same brand-new session can both observe `keyExists == false` and pick v1 (fine), but a thread that runs *after* the v0 marker first appears will switch layouts mid-flight — see the interleaving problem flagged on `RedisAgentStateStore.save(list)`.

A deterministic choice (config-selected layout version, or layout pinned by the first write and never inferred from key existence) would remove both issues.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateStore.java
```diff
@@ -372,19 +381,29 @@ public <T extends State> List<T> getList(
...
+                            + sessionId,
+                    e);
         }
     }
 
     @Override
     public void delete(String userId, String sessionId) {
-        String slotId = slotId(userId, sessionId);
-        String keysKey = getKeysKey(slotId);
+        RedisAgentStateKeyLayout keyLayout = resolveKeyLayout(userId, sessionId);
```
oss-maintainer, line 405: **[Warning] `delete()` only ever removes one layout's keys, so v0+v1 sessions leak keys.**

`delete()` resolves a single layout and deletes only the keys derived from it. Because list state is intentionally written under v0 while versioned state may be under v1 (and vice versa, depending on write order), a session can legitimately own keys in both layouts; the current implementation cannot clean up both. `clearAllSessions()` (pattern `keyPrefix + "*"`) does cover both, which highlights the asymmetry.

Consider collecting keys from both v0 and v1 markers in `delete()` (two `SMEMBERS` + one `deleteKeys` per layout, or a merged set) so that deletion is complete regardless of which layout the session ended up in.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateKeyLayout.java
```diff
@@ -0,0 +1,250 @@
...
+     *
+     * <p>Example: {@code agentscope:session:{user/session}:state}.
+     *
+     * @param keyPrefix Redis key prefix configured for the state store
+     * @param userId original user id, nullable
+     * @param sessionId session id, required
+     * @return v1 cluster-safe key layout
+     */
+    static RedisAgentStateKeyLayout v1(String keyPrefix, String userId, String sessionId) {
+        return new RedisAgentStateKeyLayout(
```
oss-maintainer, line 82: **[Warning] `userId`/`sessionId` are interpolated raw into a hash tag and into `KEYS`/glob patterns.**

`v1()` builds `"{" + normalizeUser(userId) + "/" + sessionId + "}"`, and `v1KeysPattern`/`v0KeysPattern` are consumed by `findKeysByPattern`. That means:

- a `userId`/`sessionId` containing `{`, `}` or `*`/`?`/`[` moves or breaks the hash tag and the glob match (keys could land in the wrong slot, or `listSessionIds` silently returns nothing);
- a `sessionId` containing `/` is mis-parsed by `parseV0SessionIdKeysKey`/`parseV1SessionIdFromKeysKey`, which split on the first `/` boundary implicitly by prefix length.

Since these ids come from external callers (tenant/user identifiers), please validate or encode them (e.g. reject or escape `{}`, `*`, `?`, `[`, `/`) in `sessionSegment(...)` and add a test for hostile ids. Note #3112 (`fail early for incompatible Redisson state-store clients`) shows this area is sensitive to silently-wrong behaviour.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateStore.java
```diff
@@ -394,60 +413,58 @@ public void delete(String userId, String sessionId) {
...
-            String pattern = keyPrefix + userSegment + "/*" + KEYS_SUFFIX;
-            Set<String> keysKeys = client.findKeysByPattern(pattern);
             Set<String> sessionIds = new HashSet<>();
-            String userPrefix = keyPrefix + userSegment + "/";
-            for (String keysKey : keysKeys) {
-                String withoutPrefix = keysKey.substring(userPrefix.length());
-                String sessionId =
-                        withoutPrefix.substring(0, withoutPrefix.length() - KEYS_SUFFIX.length());
-                sessionIds.add(sessionId);
+            String v1KeysPattern = RedisAgentStateKeyLayout.v1KeysPattern(keyPrefix, userSegment);
```
oss-maintainer, line 441: **[Warning] The Redis Cluster bug is fixed for keys inside one session, but the multi-key `DEL` in `delete()` still spans slots.**

The hash tag makes `state`, `ver` and `_keys` share a slot, which is what the Lua CAS needs — good. But `delete()` finishes with `client.deleteKeys(keysToDelete.toArray(...))`, and `clearAllSessions()` deletes a whole `keyPrefix + "*"` scan set in one call. Under Redis Cluster those calls pass keys from many sessions (many slots); depending on adapter that is a `CROSSSLOT` error or a silent partial delete, so the cluster-safety story is incomplete. Please confirm the adapters' behaviour (Redisson batch vs `JedisCluster.del(String...)` vs Lettuce) and, if needed, group keys per slot / delete one-by-one. Worth calling out explicitly in the class javadoc either way.

agentscope-extensions/agentscope-extensions-redis/src/test/java/io/agentscope/extensions/redis/state/RedisAgentStateStoreTest.java
```diff
@@ -81,4 +94,216 @@ void saveIfVersionCasDelegatesToLua() {
         assertEquals(5L, version);
         verify(client, times(1)).evalScript(any(), anyList(), anyList());
```
oss-maintainer, line 97: **[Info] Test coverage is good for the happy paths; the risky transitions are untested.** Mock-based tests assert which key strings reach the client, which is the right level, but please add: (a) the interleaved `save(single) -> save(list) -> getVersioned(single)` sequence from the finding above; (b) `delete()` for a session that owns keys in both layouts; (c) `listSessionIds` when a `userId`/`sessionId` contains glob or hash-tag characters. A testcontainers-backed cluster smoke test (testcontainers is already a project dependency) would be the only way to prove the slot assumption end-to-end.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateKeyLayout.java
```diff
@@ -0,0 +1,250 @@
...
+    static final String KEYS_SUFFIX = ":_keys";
+
+    private static final String LIST_SUFFIX = ":list";
+
+    private static final String ANON_USER = "__anon__";
+
+    private final String keyPrefix;
+
+    private final String slotId;
+
```
oss-maintainer, line 43: **[Info] Class-level javadoc of `RedisAgentStateKeyLayout`: the v0/v1 split has no operator-facing escape hatch.** Deployments that would rather start clean (or roll back) cannot choose a layout explicitly, and there is no migration path documented for v0 -> v1 (v0 sessions stay non-cluster-safe forever, which is exactly the reported failure mode). Consider a `RedisAgentStateStore` builder option (e.g. `keyLayoutVersion(v1)` / `forceClusterSafeLayout()`) plus a short migration note in the docs.

oss-maintainer (commented):
### Summary
Introduces `RedisAgentStateKeyLayout` so the versioned state key, its `:ver` key and the session `:_keys` marker share one Redis Cluster hash slot (fixing the CROSSSLOT failure of the Lua CAS path), keeps legacy v0 sessions readable, and adds a solid mock-based test suite. The layout extraction itself is clean and well documented — but the v0/v1 decision is inferred from the *existence* of the v0 `:_keys` marker while list state is always written under v0, which lets a list write flip an existing v1 session back to v0. That is a data-loss path and needs to be addressed before merge.

### Findings
- **[Critical]** `RedisAgentStateStore.java:306` — `save(list)` creates the v0 `:_keys` marker, so `resolve()` afterwards returns v0 for a session whose versioned state lives under v1: reads silently return empty, the next write duplicates the key, and `delete()` orphans the v1 keys. Layout depends on write order.
- **[Warning]** `RedisAgentStateKeyLayout.java:65` — `EXISTS` probe on every versioned op (extra round trip per state access) and non-deterministic under concurrency; prefer an explicit/pinned layout choice.
- **[Warning]** `RedisAgentStateStore.java:405` — `delete()` removes only one layout while a session can legitimately own keys in both.
- **[Warning]** `RedisAgentStateKeyLayout.java:82` — raw `userId`/`sessionId` interpolated into a hash tag and `KEYS` glob patterns (`{}`, `*`, `?`, `[`, `/`) corrupt slot assignment and parsing.
- **[Warning]** `RedisAgentStateStore.java:441` — cluster-safety is incomplete for the multi-key `deleteKeys(...)` in `delete()`/`clearAllSessions()`, which still spans slots across sessions.
- **[Info]** tests — interleaved sequence, dual-layout delete and hostile ids are untested; a testcontainers cluster test would prove the fix end-to-end.
- **[Info]** `RedisAgentStateKeyLayout.java:43` — no operator escape hatch (explicit layout version) and no documented v0 -> v1 migration path.

### Suggestions
The smallest safe fix for the Critical item: make `resolve()` decide the versioned layout from a marker that only versioned writes create (e.g. existence of the v0 state/`:ver` key or a dedicated `:_layout` key written once), or track list keys under the same resolved layout as versioned keys — then the v0 marker can never be produced by a list write.

### CI / Merge state
- `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED` (required review/approval outstanding).
- Codecov reports patch coverage `89.02%` with 9 lines uncovered.
- No `license/cla` status is reported on this head, so CLA state is unknown to me — a maintainer should confirm before merging.

Nice work on the key-layout extraction and the documentation comments; the mixed-layout interaction is the one thing I would not merge without resolving.

---

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateStore.java
```diff
@@ -47,15 +47,28 @@
  *   <li>Redisson - Standalone, Cluster, Sentinel, Master/Slave</li>
  * </ul>
  *
- * <p>The session state is stored in Redis with following key structure:
+ * <p>The session state is stored in Redis with following key structure. The session segment is
+ * {@code {userId}/{sessionId}} (or {@code __anon__/{sessionId}} when user id is absent). Existing
+ * versioned state entries keep their v0 non-hash-tagged session segment. New versioned state
+ * entries use a Redis Cluster-safe hash-tagged session segment, where {@code {userId}/{sessionId}}
+ * is wrapped in braces. List state keeps the v0 non-hash-tagged layout.
```
oss-maintainer, line 54: **[Warning]** The new commit `330ead5f` moved list state onto the resolved layout, but this class javadoc still states the opposite: `"List state keeps the v0 non-hash-tagged layout"`, and the key-structure bullets at lines 60-63 document list keys as `{prefix}{userId}/{sessionId}:{stateKey}:list`. Since `save(list)`/`getList(...)` now call `resolveKeyLayout(...)`, a brand-new session's list keys are in fact hash-tagged (`{prefix}{user/session}:messages:list`). Please update this block (and the matching sentence in `RedisAgentStateKeyLayout` lines 29-30) — the layout is what operators use to size, scan and migrate keys, so a doc that contradicts the code is a real hazard here.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateStore.java
```diff
@@ -372,19 +381,29 @@ public <T extends State> List<T> getList(
 
     @Override
     public boolean exists(String userId, String sessionId) {
-        String slotId = slotId(userId, sessionId);
-        String keysKey = getKeysKey(slotId);
+        RedisAgentStateKeyLayout v1KeyLayout = v1KeyLayout(userId, sessionId);
         try {
-            return client.keyExists(keysKey) && client.getSetSize(keysKey) > 0;
+            String v1KeysKey = v1KeyLayout.getKeysKey();
+            if (client.keyExists(v1KeysKey)) {
```
oss-maintainer, line 387: **[Warning]** `exists()` and `resolve()` use opposite layout precedence. `exists()` probes **v1 first** and only falls back to v0 when the v1 marker is missing, while `RedisAgentStateKeyLayout.resolve()` (used by `get`/`getVersioned`/`save`/`saveIfVersion`/`delete`) probes **v0 first** and only uses v1 when no v0 marker exists. For a session that owns both markers — a v0 session created before this change that later got versioned state, or any partially migrated session — `exists()` can report the v1 view while every read/write/delete acts on the v0 view, so a caller sees `exists == false` for data it can still read (or vice versa). Worth picking one documented precedence rule and sharing a single helper between `exists()` and `resolve()`; the `existsSelectsClusterSafeLayoutBeforeLegacyFallback` test currently locks in the v1-first behaviour, so it would need updating with whichever direction you choose.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateStore.java
oss-maintainer, line 405: **[Warning]** Carried over from the previous round and still open: `delete()` resolves exactly one layout, so a session that legitimately owns keys under both v0 and v1 is only half removed and the other layout's `state`/`ver`/`_keys` entries stay as permanent orphans. `clearAllSessions()` (`keyPrefix + "*"`) does cover both, which shows the asymmetry. Collecting tracked members from both markers (two `SMEMBERS` + one merged `deleteKeys`) would make deletion complete regardless of how a session ended up mixed.

oss-maintainer (commented):
### Summary

`330ead5f` fixes the blocking issue from my previous round: `save(list)` / `getList(...)` now go through `resolveKeyLayout(...)` instead of hard-coding the v0 layout, so a list write can no longer create a v0 `:_keys` marker for a session whose versioned state lives under v1 — and `listSaveDoesNotMoveExistingClusterSafeSessionBackToLegacyLayout` pins exactly the interleaving that used to lose data. CI is green (build on ubuntu + windows) and the CLA is signed. Thanks for the quick and precise turnaround.

Downgrading to COMMENT rather than APPROVE because two layout-resolution inconsistencies are still open (one of them a direct consequence of the fix): `exists()` and `resolve()` now disagree about which marker wins, `delete()` still cleans only one layout, and the class javadoc still documents the behaviour the new commit removed. None of these look hard to close.

### Findings
- **[Warning]** `RedisAgentStateStore.java:54` — class javadoc still says list state keeps the v0 layout, which `330ead5f` made untrue.
- **[Warning]** `RedisAgentStateStore.java:387` — `exists()` probes v1 first, `resolve()` probes v0 first: contradictory answers for a session with both markers.
- **[Warning]** `RedisAgentStateStore.java:405` — `delete()` removes only one layout's keys (carried over, still open).
- **[Info]** `RedisAgentStateStoreTest.java:217` — regression test for the fixed sequence confirmed; delete-mixed-layout and hostile-id cases still untested.

Carried over and unchanged from the previous round, so I won't re-annotate them: the per-operation `EXISTS` probe cost in `resolve()`, raw `userId`/`sessionId` interpolation into hash tags and `KEYS` glob patterns, the cross-slot `deleteKeys(...)` in `delete()` / `clearAllSessions()`, and the missing operator-facing layout-version switch.

---

agentscope-extensions/agentscope-extensions-redis/src/test/java/io/agentscope/extensions/redis/state/RedisAgentStateStoreTest.java
```diff
@@ -81,4 +96,250 @@ void saveIfVersionCasDelegatesToLua() {
...
+                        invocation -> {
+                            String keysKey = invocation.getArgument(0);
+                            if ("agentscope:session:user/s1:_keys".equals(keysKey)) {
+                                v0MarkerCreated.set(true);
+                            }
+                            return null;
+                        })
+                .when(client)
+                .addToSet(anyString(), anyString());
+
```
oss-maintainer, line 217: **[Info]** `listSaveDoesNotMoveExistingClusterSafeSessionBackToLegacyLayout` is exactly the regression test the previous round asked for — thanks, the interleaved `save(single) -> save(list) -> getVersioned(single)` sequence is now pinned and the atomic `AtomicBoolean` marker simulation makes the failure mode impossible to reintroduce silently. Two gaps still open from last time: `delete()` for a session that owns keys in *both* layouts, and `listSessionIds`/key building when a `userId`/`sessionId` contains `{`, `}`, `*`, `?` or `/`.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateKeyLayout.java
oss-maintainer, line 66: **[Warning]** `RedisAgentStateKeyLayout.resolve()` performs an extra `EXISTS` round trip on every versioned read/write and the v0/v1 decision is not atomic. Between the probe at line 65 and the subsequent write, another client can create the v0 marker, causing the same logical session to be split across both layouts. Consider caching the resolved layout per `(userId, sessionId)` in the store or providing an explicit, atomic migration path.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateStore.java
oss-maintainer: **[Warning]** `delete()` resolves the layout once (line 405) and only deletes keys belonging to that layout. If a session ever ends up with both v0 and v1 marker sets (e.g., due to the `resolve()` race or a partial migration), keys from the other layout leak forever. Delete both layouts, or ensure `resolve()` can never produce a mixed-layout session.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateKeyLayout.java
```diff
@@ -0,0 +1,250 @@
...
+     *
+     * @param userId original user id, nullable
+     * @param sessionId session id, required
+     * @return normalized session segment
+     */
+    private static String sessionSegment(String userId, String sessionId) {
+        if (sessionId == null || sessionId.isBlank()) {
+            throw new IllegalArgumentException("sessionId must not be blank");
+        }
+        return normalizeUser(userId) + "/" + sessionId;
```
oss-maintainer, line 249: **[Warning]** `sessionSegment()` (line 248) interpolates `userId` and `sessionId` literally into Redis keys, hash tags, and SCAN patterns without escaping. A `sessionId` such as `s*1` turns `v1KeysPattern()` into `prefix{user/s*1}:_keys` and matches unrelated sessions; `}` or `{` in `userId`/`sessionId` breaks Redis hash-tag parsing and can make v1 keys hash to the wrong cluster slot or collide with other sessions. Sanitize or reject characters that have Redis pattern/key semantics.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateStore.java
oss-maintainer: **[Warning]** `exists()` checks the v1 marker first and falls back to v0 (lines 387-392), while `resolveKeyLayout()` (used by every read/write) prefers v0 whenever the v0 marker exists. For a session with both markers, `exists()` and the actual storage layout can disagree on which keys matter. Use the same precedence in both paths.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateStore.java
oss-maintainer: **[Info]** The class-level Javadoc still states that "List state keeps the v0 non-hash-tagged layout" (line 54), but `save(userId, sessionId, key, List)` now calls `resolveKeyLayout()` (line 306), so new sessions write list keys with the hash-tagged v1 layout. Update the Javadoc to reflect that list layout follows the resolved session layout.

agentscope-extensions/agentscope-extensions-redis/src/test/java/io/agentscope/extensions/redis/state/RedisAgentStateStoreTest.java
```diff
@@ -81,4 +96,250 @@ void saveIfVersionCasDelegatesToLua() {
         assertEquals(5L, version);
         verify(client, times(1)).evalScript(any(), anyList(), anyList());
     }
+
+    @Test
+    @DisplayName("new session save passes only cluster-safe keys to Lua")
+    void newSessionSaveUsesOneHashSlotForLuaKeys() {
```
oss-maintainer, line 102: **[Info]** `newSessionSaveUsesOneHashSlotForLuaKeys` asserts that Lua keys contain the string `"{user/s1}"` (via `assertClusterSafeSessionKeys`), but string containment does not prove the keys share the same Redis Cluster hash slot. Add a hash-slot assertion (e.g., using `JedisClusterCRC16.getSlot`) over the captured Lua keys, or an integration test against a real cluster.

agentscope-extensions/agentscope-extensions-redis/src/test/java/io/agentscope/extensions/redis/state/RedisAgentStateStoreTest.java
```diff
@@ -81,4 +96,250 @@ void saveIfVersionCasDelegatesToLua() {
...
+        verify(client, never()).getSetMembers("agentscope:session:user/s1:_keys");
+        ArgumentCaptor<String[]> deletedKeys = ArgumentCaptor.forClass(String[].class);
+        verify(client).deleteKeys(deletedKeys.capture());
+        List<String> deleted = Arrays.asList(deletedKeys.getValue());
+        assertTrue(deleted.contains("agentscope:session:{user/s1}:_keys"));
+        assertTrue(deleted.contains("agentscope:session:{user/s1}:agent_state"));
+        assertTrue(deleted.contains("agentscope:session:{user/s1}:agent_state:ver"));
+    }
+
+    @Test
```
oss-maintainer, line 256: **[Info]** `deleteSessionUsesLegacyLayoutWhenOnlyLegacyMarkerExists` covers deletion of a purely v0 session, but there is no test for a mixed-layout session where both v0 and v1 marker sets exist. Such a test would expose that `delete()` currently leaves the unresolved layout behind.

oss-maintainer (changes requested):
### Summary

This PR introduces a Redis Cluster-safe v1 key layout for `RedisAgentStateStore` by wrapping the session segment in a Redis hash tag (`{userId/sessionId}`), while keeping existing v0 sessions on the legacy non-tagged layout so previously persisted state remains readable and writable. Key construction is extracted into `RedisAgentStateKeyLayout`, and several unit tests are added for layout selection, backward compatibility, and cluster-safe key strings. The critical regression where `save(userId, sessionId, key, List)` forced v0 layout for an existing v1 session is fixed in the current head.

However, the layout resolution is still racy and performed on every call, `delete()` only cleans one of the two possible layouts, user/session identifiers are interpolated into keys and patterns unsafely, and the precedence between `exists()` and `resolveKeyLayout()` is inconsistent. These issues affect correctness, safety in Redis Cluster, and key hygiene, so the PR needs another revision before approval.

### Previous findings status

- **[Critical] `save(list)` used `v0KeyLayout`, flipping an existing v1 session back to v0.** Fixed in the current head; `save(userId, sessionId, key, List)` now calls `resolveKeyLayout()` and the new test `listSaveDoesNotMoveExistingClusterSafeSessionBackToLegacyLayout` guards it.
- **[Warning] `RedisAgentStateKeyLayout.resolve()` adds an extra `EXISTS` round trip on every versioned read/write and is not concurrency-safe.** Still open.
- **[Warning] `delete()` removes only one resolved layout, leaking keys of the other layout.** Still open.
- **[Warning] `userId`/`sessionId` interpolated raw into the hash tag and into `KEYS`/glob patterns (wildcard injection / cross-session key collision risk).** Still open.
- **[Warning] `exists()` probes v1-first while `resolve()` may pick v0 — inconsistent precedence.** Still open.
- **[Info] Class javadoc of `RedisAgentStateStore` still claims list state keeps the v0 layout.** Still open; the statement is now actively misleading because new sessions write list keys with the v1 layout.
- **[Info] No operator-facing way to pin/roll back the layout (v0 vs v1) for a deployment.** Still open.
- **[Info] Test gaps: interleaved `save(single)`/`save(list)` ordering, delete of mixed-layout session, `exists()` precedence.** Partially addressed — `exists()` precedence and list-v1-layout are now tested, but mixed-layout deletion and interleaved ordering are not.

### Verdict

**REQUEST_CHANGES** — The racy per-call layout resolution, single-layout deletion, and unsanitized `userId`/`sessionId` interpolation are material correctness and cluster-safety issues that should be fixed before this PR is merged.
### Findings
- **[WARNING]** `RedisAgentStateKeyLayout.java:65` — `RedisAgentStateKeyLayout.resolve()` performs an extra `EXISTS` round trip on every versioned read/write and the v0/v1 decision is not atomic. Between the probe at line 65 and the subsequent write, another client can create the v0 marker, causing the same logical session to be split across both layouts. Consider caching the resolved layout per `(userId, sessionId)` in the store or providing an explicit, atomic migration path.
- **[WARNING]** `RedisAgentStateStore.java:405` — `delete()` resolves the layout once (line 405) and only deletes keys belonging to that layout. If a session ever ends up with both v0 and v1 marker sets (e.g., due to the `resolve()` race or a partial migration), keys from the other layout leak forever. Delete both layouts, or ensure `resolve()` can never produce a mixed-layout session.
- **[WARNING]** `RedisAgentStateKeyLayout.java:248` — `sessionSegment()` (line 248) interpolates `userId` and `sessionId` literally into Redis keys, hash tags, and SCAN patterns without escaping. A `sessionId` such as `s*1` turns `v1KeysPattern()` into `prefix{user/s*1}:_keys` and matches unrelated sessions; `}` or `{` in `userId`/`sessionId` breaks Redis hash-tag parsing and can make v1 keys hash to the wrong cluster slot or collide with other sessions. Sanitize or reject characters that have Redis pattern/key semantics.
- **[WARNING]** `RedisAgentStateStore.java:387` — `exists()` checks the v1 marker first and falls back to v0 (lines 387-392), while `resolveKeyLayout()` (used by every read/write) prefers v0 whenever the v0 marker exists. For a session with both markers, `exists()` and the actual storage layout can disagree on which keys matter. Use the same precedence in both paths.
- **[INFO]** `RedisAgentStateStore.java:54` — The class-level Javadoc still states that "List state keeps the v0 non-hash-tagged layout" (line 54), but `save(userId, sessionId, key, List)` now calls `resolveKeyLayout()` (line 306), so new sessions write list keys with the hash-tagged v1 layout. Update the Javadoc to reflect that list layout follows the resolved session layout.
- **[INFO]** `RedisAgentStateStoreTest.java:102` — `newSessionSaveUsesOneHashSlotForLuaKeys` asserts that Lua keys contain the string `"{user/s1}"` (via `assertClusterSafeSessionKeys`), but string containment does not prove the keys share the same Redis Cluster hash slot. Add a hash-slot assertion (e.g., using `JedisClusterCRC16.getSlot`) over the captured Lua keys, or an integration test against a real cluster.
- **[INFO]** `RedisAgentStateStoreTest.java:257` — `deleteSessionUsesLegacyLayoutWhenOnlyLegacyMarkerExists` covers deletion of a purely v0 session, but there is no test for a mixed-layout session where both v0 and v1 marker sets exist. Such a test would expose that `delete()` currently leaves the unresolved layout behind.

---
*Automated review by github-manager-bot — round 3 (re-review after the merge of main). All findings above are inline comments on the current head `f4e6e167`; the previous `save(list)` critical regression is confirmed fixed.*

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateKeyLayout.java
```diff
@@ -0,0 +1,251 @@
...
+     * @param client Redis client used to test whether the v0 marker exists
+     * @param keyPrefix Redis key prefix configured for the state store
+     * @param userId original user id, nullable
+     * @param sessionId session id, required
+     * @return resolved layout for the session
+     */
+    static RedisAgentStateKeyLayout resolve(
+            RedisClientAdapter client, String keyPrefix, String userId, String sessionId) {
+        RedisAgentStateKeyLayout v0 = v0(keyPrefix, userId, sessionId);
+        if (client.keyExists(v0.getKeysKey())) {
```
oss-maintainer, line 66: **[Warning]** This is the item still open from the previous round: `resolve()` issues an `EXISTS` round trip on **every** read/write (`get`, `save`, `saveVersioned`, `getList`, `exists`, …), and the result is never cached, so two concurrent first-writes for the same new session can both observe "no v0 marker" and diverge if the marker appears in between. Consider resolving once per logical operation and, if acceptable for the deployment model, caching the settled layout per session (v0 marker presence is immutable once a session stops being written to the legacy layout). At minimum, documenting the read-your-writes assumption here would help operators reason about it.

agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateKeyLayout.java
```diff
@@ -0,0 +1,251 @@
...
+
+    /**
+     * Build the key pattern used to scan v1 session marker keys for a user segment.
+     *
+     * @param keyPrefix Redis key prefix configured for the state store
+     * @param userSegment normalized user segment
+     * @return Redis scan pattern for v1 session marker keys
+     */
+    static String v1KeysPattern(String keyPrefix, String userSegment) {
+        return keyPrefix + "{" + userSegment + "/*}" + KEYS_SUFFIX;
```
oss-maintainer, line 119: **[Warning]** Unsanitized interpolation, still open from the previous round. `normalizeUser(userId)` is inserted raw into the hash tag here and in `v1()` (`"{" + sessionSegment + "}"`). A user id (or session id) containing `}`, `{`, `*` or `?` breaks the glob/hash-tag matching in `listSessionIds`, and worse, can split the hash tag so the state/version/marker keys of one "session" land in **different cluster slots**, which makes the multi-key Lua CAS path fail with CROSSSLOT. v0 keys had the same shape, but v1 makes slot grouping a functional invariant, so it is worth hardening now: reject or escape `{}`/glob metacharacters in `normalizeUser`/`sessionSegment` (or hash them) rather than relying on callers to pass clean ids.

oss-maintainer (commented):
### Summary

Delta re-review at `65e52f5c` (previous pass at `f4e6e167`, CHANGES_REQUESTED). The new commit directly targets the blocking findings from the last round, and three of them are now fixed cleanly:

### Previous findings status

- **[Warning] `delete()` removed only one resolved layout, leaking keys of the other layout.** Fixed — `delete()` now unconditionally sweeps both `deleteUserSession(v0)` and `deleteUserSession(v1)`, and `deleteRemovesBothLayoutsWithoutCombiningRedisSlots` guards it (importantly, keys of the two layouts are never combined into one multi-key command, which would CROSSSLOT on a cluster).
- **[Warning] `exists()` probed v1-first while `resolve()` may pick v0 — inconsistent precedence.** Fixed — `exists()` now goes through `resolveKeyLayout()` and reports `getSetSize(keysKey) > 0`, with both precedence branches tested.
- **[Info] Class javadoc claimed list state keeps the v0 layout.** Fixed — docs in both classes now state that scalar and list state share the resolved layout, which matches the code path (all of `save(list)`/`getList()` resolve via `resolveKeyLayout`).
- **[Warning] `resolve()` extra `EXISTS` round trip per call, not concurrency-safe.** Still open — see inline comment.
- **[Warning] `userId`/`sessionId` interpolated raw into hash tags and scan patterns.** Still open — the new `v1KeysPattern` makes this more consequential, not less. See inline comment.
- **[Info] No operator-facing way to pin/roll back the layout.** Still open (non-blocking).

### Assessment

The layout unification is coherent now: single resolution precedence everywhere, symmetric dual-layout delete, and 118 lines of new tests covering exactly the fixed behaviors. CI (ubuntu + windows builds, module sync, license, codecov) is green on this head, and CLA is signed.

What keeps this from an approval is the same two hardening items as last round: the per-call `EXISTS` (hot-path cost + TOCTOU on first concurrent writes) and unsanitized ids inside the v1 hash tag (CROSSSLOT risk for adversarial/incidental `{}`/`*` in user or session ids — e.g. an id like `a}b` gets wrapped as `{user/a}b/...`, silently breaking slot grouping). Both are small, contained changes in `RedisAgentStateKeyLayout`; I'd like to see them addressed in this PR since v1's slot grouping is the feature's whole point.

Overall: strong progress, two focused fixes away from done.

---

jujn:
@xzxiaoshan PTAL Thanks.

xzxiaoshan:
下面整理了表格，统一说明：

### 待修复问题

| # | 大概的代码位置 | 说明 |
|---|---------|------------|
| 1 | JedisClientAdapter.java:155-169 | Cluster 下无 key 的 SCAN 只会落到一个节点上，listSessionIds 会漏掉其他节点上的会话，而且没有任何报错，比直接失败更隐蔽。Jedis 有现成的全节点迭代 API scanIteration()，建议换掉这个循环。Lettuce 和 Redisson 的实现已经覆盖全部 master，不用动。这也是 v1-only 下唯一跨 slot 作用域的方法，修好它跨 slot 面就补齐了。 |
| 2 | RedisAgentStateStore.java:479-493 | 这个方法建议直接删掉。全库清除没有用户维度，隔离只有 keyPrefix 一层，多个应用共用一个 Redis 时会互相误伤；Jedis Cluster 上还会漏分片、批量 DEL 跨 slot 报错。它不在 AgentStateStore 接口上，框架内部也没有任何调用，全库清理是一个很危险的行为，如果业务真需要全库清理的业务自己实现更合适。 |
| 3 | RedisAgentStateStore.java:382-394 | 这里把 sessionId 的校验异常包进了 RuntimeException，delete() 也是同样写法。但 save/get 路径的布局构造在 try 外面，blank sessionId 抛的是原始 IllegalArgumentException，两边行为不一致。建议把校验挪到 try 之前，和其他方法对齐。 |
| 4 | RedisAgentStateKeyLayout.java:118-120 | userId 是直接拼进 SCAN pattern 的，含 * ? [ 这类字符时会误匹配别的用户的会话。#2579 里有现成的 glob 转义，建议一并移植，顺带对 id 和 prefix 里的花括号加个校验。 |
| 5 | RedisAgentStateStoreTest.java | 单测都是 mock 的，验证 key 拼接没问题，但验证不了 Cluster 上 CROSSSLOT 真的消失。#2579 里有现成的 Testcontainers 集成测试，可以考虑移植或者看什么方式比较合适。如果已经在真实 Cluster 上手动验证过，也可以在 PR 描述里补一句结论。 |
| 6 | RedisAgentStateKeyLayout.java:63-70 | PR 描述里写的是 exists 先查 v1 marker、缺失才回退 v0，代码实际是先查 v0，测试 existsUsesResolvedLegacyLayoutWhenBothMarkersExist 断言的也是 v0 优先。应该是最后一个 commit 改了优先级但描述没跟上，两边需要取齐；如果按 v1-only 方向改，兼容行为这节描述直接重写。 |

@rudy2steiner 处理一下

oss-maintainer:
### Summary
Re-check at the same head `65e52f5c`: no new commits since my previous pass, and the thread has since picked up a detailed list from @xzxiaoshan, so this comment only records what still stands on my side and where it overlaps. I am not re-listing items @xzxiaoshan already located.

### What still blocks the approval (both pinned inline)
- **Unsanitized `userId`/`sessionId` interpolated into the hash tag and into the SCAN glob patterns** (`RedisAgentStateKeyLayout.java:118`, also 84/130/245). Overlaps @xzxiaoshan item #4 on the glob side; the hash-tag side (`{`, `}`, `/` in an id) additionally breaks the cross-slot guarantee that is the whole point of the v1 layout.
- **Layout resolution costs an extra `EXISTS` round trip on every state operation and the v0/v1 choice is not atomic** (`RedisAgentStateKeyLayout.java:66`, called from 7 places in `RedisAgentStateStore`).

### Note on @xzxiaoshan item #6 (description vs code precedence)
From my earlier pass the code side is consistent: `exists()` goes through `resolveKeyLayout()` and the both-markers test asserts v0-first. So #6 looks like a PR-description fix rather than a behaviour fix - the "Compatibility behavior" section should be rewritten to say v0-first (or v1-only if that is the chosen direction), not the other way round.

### Not verified by me
Items #1 (Jedis cluster SCAN visiting a single shard), #2 (whole-DB clear helper), #3 (validation exception wrapping inconsistency), #5 (mock-only tests vs real-cluster proof) are outside what I re-checked this round; I have no counter-evidence and they read as legitimate - I would treat #1 and #2 as blocking on the maintainer's judgement.

### Path forward
Push the fixes for the two inline items together with the maintainer list and ping me - I will review only the delta against `65e52f5c`. CI (Check License, Check Module Sync, ubuntu + windows builds, codecov) is green on this head and CLA is signed; `reviewDecision` stays `CHANGES_REQUESTED` until the items above are addressed.


### Inline notes (anchored to this head)
- **[WARNING]** `agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateKeyLayout.java:66` — Re-verified at this head: still open. `resolve(...)` calls `client.keyExists(v0.getKeysKey())` on every invocation, and `RedisAgentStateStore` resolves on every `save`/`get`/`saveIfVersion`/`getVersioned`/`getList`/`exists` (RedisAgentStateStore.java:234, 262, 286, 305, 347, 363, 384) - one extra round trip per state operation, and the check-then-choose sequence is not atomic: between the EXISTS returning false and the following v1 read/write, another caller can create a v0 marker, so a session can end up split across layouts. Smallest fix: cache the resolved layout per (userId, sessionId) after the first resolution, which removes both the round trip and the TOCTOU.

- **[CRITICAL]** `agentscope-extensions/agentscope-extensions-redis/src/main/java/io/agentscope/extensions/redis/state/RedisAgentStateKeyLayout.java:118` — Re-verified at this head: still open, and this is the same root cause as @xzxiaoshan's item #4 plus the hash-tag half that #4 does not cover. `sessionSegment` (line 245) and `v1` (line 84) concatenate raw `userId`/`sessionId` into the Redis hash tag, while `v1KeysPattern` (this line, `v1KeysPattern`) and `v0KeysPattern` (line 130) concatenate the raw user segment into SCAN glob patterns; the Jedis/Lettuce/Redisson client adapters pass the pattern through unchanged. An id containing `*`, `?`, `[` or `]` lets `listSessionIds` match other users' marker keys (tenant-isolation leak in listing), and an id containing `{`, `}` or `/` silently breaks the slot grouping that v1 exists to provide. Smallest fix: escape/encode `userId` and `sessionId` before building the segment and the patterns, and decode them in `parseV1SessionIdFromKeysKey`/`parseV0SessionIdKeysKey`.

---

rudy2steiner:
> 下面整理了表格，统一说明：
> 
> ### 待修复问题
> #	大概的代码位置	说明
> 1	JedisClientAdapter.java:155-169	Cluster 下无 key 的 SCAN 只会落到一个节点上，listSessionIds 会漏掉其他节点上的会话，而且没有任何报错，比直接失败更隐蔽。Jedis 有现成的全节点迭代 API scanIteration()，建议换掉这个循环。Lettuce 和 Redisson 的实现已经覆盖全部 master，不用动。这也是 v1-only 下唯一跨 slot 作用域的方法，修好它跨 slot 面就补齐了。
> 2	RedisAgentStateStore.java:479-493	这个方法建议直接删掉。全库清除没有用户维度，隔离只有 keyPrefix 一层，多个应用共用一个 Redis 时会互相误伤；Jedis Cluster 上还会漏分片、批量 DEL 跨 slot 报错。它不在 AgentStateStore 接口上，框架内部也没有任何调用，全库清理是一个很危险的行为，如果业务真需要全库清理的业务自己实现更合适。
> 3	RedisAgentStateStore.java:382-394	这里把 sessionId 的校验异常包进了 RuntimeException，delete() 也是同样写法。但 save/get 路径的布局构造在 try 外面，blank sessionId 抛的是原始 IllegalArgumentException，两边行为不一致。建议把校验挪到 try 之前，和其他方法对齐。
> 4	RedisAgentStateKeyLayout.java:118-120	userId 是直接拼进 SCAN pattern 的，含 * ? [ 这类字符时会误匹配别的用户的会话。#2579 里有现成的 glob 转义，建议一并移植，顺带对 id 和 prefix 里的花括号加个校验。
> 5	RedisAgentStateStoreTest.java	单测都是 mock 的，验证 key 拼接没问题，但验证不了 Cluster 上 CROSSSLOT 真的消失。#2579 里有现成的 Testcontainers 集成测试，可以考虑移植或者看什么方式比较合适。如果已经在真实 Cluster 上手动验证过，也可以在 PR 描述里补一句结论。
> 6	RedisAgentStateKeyLayout.java:63-70	PR 描述里写的是 exists 先查 v1 marker、缺失才回退 v0，代码实际是先查 v0，测试 existsUsesResolvedLegacyLayoutWhenBothMarkersExist 断言的也是 v0 优先。应该是最后一个 commit 改了优先级但描述没跟上，两边需要取齐；如果按 v1-only 方向改，兼容行为这节描述直接重写。
> @rudy2steiner 处理一下



> 下面整理了表格，统一说明：
> 
> ### 待修复问题
> #	大概的代码位置	说明
> 1	JedisClientAdapter.java:155-169	Cluster 下无 key 的 SCAN 只会落到一个节点上，listSessionIds 会漏掉其他节点上的会话，而且没有任何报错，比直接失败更隐蔽。Jedis 有现成的全节点迭代 API scanIteration()，建议换掉这个循环。Lettuce 和 Redisson 的实现已经覆盖全部 master，不用动。这也是 v1-only 下唯一跨 slot 作用域的方法，修好它跨 slot 面就补齐了。
> 2	RedisAgentStateStore.java:479-493	这个方法建议直接删掉。全库清除没有用户维度，隔离只有 keyPrefix 一层，多个应用共用一个 Redis 时会互相误伤；Jedis Cluster 上还会漏分片、批量 DEL 跨 slot 报错。它不在 AgentStateStore 接口上，框架内部也没有任何调用，全库清理是一个很危险的行为，如果业务真需要全库清理的业务自己实现更合适。
> 3	RedisAgentStateStore.java:382-394	这里把 sessionId 的校验异常包进了 RuntimeException，delete() 也是同样写法。但 save/get 路径的布局构造在 try 外面，blank sessionId 抛的是原始 IllegalArgumentException，两边行为不一致。建议把校验挪到 try 之前，和其他方法对齐。
> 4	RedisAgentStateKeyLayout.java:118-120	userId 是直接拼进 SCAN pattern 的，含 * ? [ 这类字符时会误匹配别的用户的会话。#2579 里有现成的 glob 转义，建议一并移植，顺带对 id 和 prefix 里的花括号加个校验。
> 5	RedisAgentStateStoreTest.java	单测都是 mock 的，验证 key 拼接没问题，但验证不了 Cluster 上 CROSSSLOT 真的消失。#2579 里有现成的 Testcontainers 集成测试，可以考虑移植或者看什么方式比较合适。如果已经在真实 Cluster 上手动验证过，也可以在 PR 描述里补一句结论。
> 6	RedisAgentStateKeyLayout.java:63-70	PR 描述里写的是 exists 先查 v1 marker、缺失才回退 v0，代码实际是先查 v0，测试 existsUsesResolvedLegacyLayoutWhenBothMarkersExist 断言的也是 v0 优先。应该是最后一个 commit 改了优先级但描述没跟上，两边需要取齐；如果按 v1-only 方向改，兼容行为这节描述直接重写。
> @rudy2steiner 处理一下

我晚点看看

PR #3130, opened 2026-09-13
fix(Toolkit): route callTool through executeWithInfrastructure

### LINKED ISSUE #3114: [Bug]: Toolkit.callTool returns a ToolResultBlock with null id and name (bypasses executeWithInfrastructure)
**Describe the bug**

`Toolkit.callTool(ToolCallParam)` returns a `ToolResultBlock` whose `id` and `name` are `null`, while `Toolkit.callTools(...)` stamps both correctly for the same tool and the same call.

Root cause: the two public entry points do not share the stage that stamps them.

- `Toolkit.callTools(...)` → `ToolExecutor.executeAll(...)` → `executeWithInfrastructure(...)`, which applies `.map(result -> result.withIdAndName(toolCall.getId(), toolCall.getName()))`.
- `Toolkit.callTool(param)` → `ToolExecutor.execute(param)` directly, which bypasses `executeWithInfrastructure(...)` entirely.

`withIdAndName(...)` is called in exactly two places in `ToolExecutor`, both inside `executeWithInfrastructure(...)`, so nothing on the `callTool` path ever stamps the result.

This matters because the id and name are what pair a result back to its call: `ToolResultMessageBuilder.buildToolResultMsg(...)` and the tool-call pairing model both key off them. A caller following the usage example in `Toolkit.callTool`'s own javadoc gets back a block it cannot correlate.

**To Reproduce**

1. Add the following test to `agentscope-core/src/test/java/io/agentscope/core/tool/CallToolIdNameReproTest.java`:

```java
package io.agentscope.core.tool;

import io.agentscope.core.message.ToolResultBlock;
import io.agentscope.core.message.ToolUseBlock;
import java.util.List;
import java.util.Map;

/** Repro: Toolkit.callTool vs Toolkit.callTools — id/name stamping on the returned block. */
class CallToolIdNameReproTest {

    static class EchoTools {
        @Tool(name = "echo", description = "Echoes back")
        public String echo() {
            return "ok";
        }
    }

    private static ToolUseBlock call(String id) {
        return ToolUseBlock.builder()
                .id(id).name("echo").input(Map.of()).content("{}").build();
    }

    @Test
    void compare() {
        toolkit.registerTool(new EchoTools());

        ToolResultBlock viaCallTool = toolkit
                .callTool(ToolCallParam.builder().toolUseBlock(call("call-single")).build())
                .block();

        List<ToolResultBlock> viaCallTools =
                toolkit.callTools(List.of(call("call-batch")), null, null, null).block();
        ToolResultBlock batch = viaCallTools.get(0);

        System.out.println("REPRO callTool  id=" + viaCallTool.getId()
                + " name=" + viaCallTool.getName() + " state=" + viaCallTool.getState());
        System.out.println("REPRO callTools id=" + batch.getId()
                + " name=" + batch.getName() + " state=" + batch.getState());
    }
}
```

2. Run it from the repository root:

```
mvn -pl agentscope-core test -Dtest=CallToolIdNameReproTest -Dsurefire.failIfNoSpecifiedTests=false
```

**Expected behavior**

Both entry points stamp the result identically:

```
REPRO callTool  id=call-single name=echo state=RUNNING
REPRO callTools id=call-batch  name=echo state=RUNNING
```

**Error messages**

No exception is thrown. Actual output on `main` at `c5db8f72`:

```
REPRO callTool  id=null name=null state=RUNNING
REPRO callTools id=call-batch name=echo state=RUNNING
```

- AgentScope-Java Version: 2.0.3-SNAPSHOT (reproduced on `main` at commit `c5db8f72`)
- Java Version: 25 (the issue is version-independent)
- OS: macOS (the issue is OS-independent)

**Additional context**

The stale javadoc on `ToolExecutor.execute(...)` suggests how the paths drifted apart — it still reads:

```java
/**
 * Execute a single tool call with full infrastructure support.
 */
Mono<ToolResultBlock> execute(ToolCallParam param) {
```

but `execute(...)` is the method that applies none of it; `executeWithInfrastructure(...)` is where the infrastructure lives, and it is private. This looks like an incomplete extraction: `executeAll(...)` was repointed at the new wrapper and `Toolkit.callTool(...)` was left on the inner method.

Beyond the missing id/name, the same divergence means the `callTool` path also skips `applyScheduling`, `applyTimeout`, `applyRetry` and `applyShutdownGuard`. The practical consequences are that a configured `ExecutionConfig` timeout and retry policy are silently inert for direct calls, and a tool invoked this way is not covered by the graceful-shutdown guard. I have kept this report scoped to the id/name defect because it is unambiguous and independently verifiable; the retry behaviour overlaps with #2829, whose author notes they are preparing a fix in this area.

Two notes for whoever picks this up:

- If the intended contract is that `Toolkit.callTool(...)` is a deliberately thin, infrastructure-free entry point, then the id/name stamping and the `execute(...)` javadoc should still be corrected, and the difference documented on `Toolkit.callTool(...)` so callers are not surprised.
- Repointing `Toolkit.callTool(...)` at `executeWithInfrastructure(...)` would fix all of the above at once, but it is an observable behaviour change for external callers currently relying on the no-timeout semantics, so it likely belongs in a minor release rather than a patch.

Nothing in the repository calls `Toolkit.callTool(...)` today — the only occurrences are the usage examples in its own javadoc — so the exposure is entirely to external users following the documented example.

KIM406-CMD:
## Summary

Fix `Toolkit.callTool` (single-call) so it applies the full infrastructure
pipeline that `callTools` (batch) already uses — aligning the two public
entry points and fixing `ToolResultBlock.id` / `ToolResultBlock.name` being
`null` on single calls.

**Closes #3114**

---

## Background

`Toolkit` exposes two public methods for invoking tools:

| Method | What it did (before this PR) |
|---|---|
| `callTool(ToolCallParam)` | Delegated straight to `ToolExecutor.execute(param)` — **pure execution only** |
| `callTools(...)` | Routed each call through `ToolExecutor.executeWithInfrastructure(...)` — **full pipeline** |

`executeWithInfrastructure(...)` wraps the raw execution with 4 infrastructure
layers plus id/name stamping:

```java
// ToolExecutor.java (batch path, before this PR)
execution = applyScheduling(execution);
execution = applyTimeout(execution, config, toolCall);
execution = applyRetry(execution, config, toolCall);
execution = applyShutdownGuard(execution);
return execution.map(result -> result.withIdAndName(toolCall.getId(), toolCall.getName()));
```

Since `callTool` bypassed this wrapper entirely, single calls got **none** of that.

### The divergence matrix (before fix)

| Feature | `callTool` (single) | `callTools` (batch) |
|---|---|---|
| `ToolResultBlock.id` | **`null` ❌** | ✅ populated |
| `ToolResultBlock.name` | **`null` ❌** | ✅ populated |
| `ExecutionConfig` timeout | **silently ignored ❌** | ✅ applied |
| `ExecutionConfig` retry | **silently ignored ❌** | ✅ applied |
| `ShutdownGuard` (graceful shutdown) | **bypassed ❌** | ✅ applied |
| Scheduling (`boundedElastic`) | **not applied ❌** | ✅ applied |

### Root Cause

`withIdAndName(...)`, `applyScheduling`, `applyTimeout`, `applyRetry`, and
`applyShutdownGuard` all live inside `executeWithInfrastructure(...)`, which
was `private` and only reachable from the batch path (`executeAll`). The
single-call path called `execute(...)` directly, which applies zero
infrastructure.

Evidence this was unintentional:

- The stale javadoc on `ToolExecutor.execute(...)` still reads *"Execute a single
  tool call with full infrastructure support"* — the method does nothing of the
  sort.
- ReActAgent (the project's own core agent) uses `callTools` exclusively,
  suggesting the in-tree developers naturally gravitated toward the path that
  actually works.
- Nothing in-tree hits the broken path today; only external users following
  the javadoc example on `Toolkit.callTool(...)` get a `ToolResultBlock` they
  cannot correlate back to the tool call (both `ToolResultMessageBuilder` and
  the tool-call pairing model key off `id` / `name`).

---
t
## Fix Strategy: Full Infrastructure Alignment  Option One

Issue #3114 proposed two directions. I chose **Option One**:

| | Option One |  Option Two |
|---|---|---|
| **Way** | Route `callTool` through `executeWithInfrastructure` full pipeline | Only patch `withIdAndName` onto the pure-execution path, leave other infrastructure skipped |
| **Fixes** | id/name, timeout, retry, ShutdownGuard, scheduling — **all five** | Only id/name |
| **Long-term cost** | One behavioural change now, then single and batch stay aligned forever | Freezes "callTool is intentionally thin" as a public API contract; future users will hit silently-missing timeout/retry, and adding them later would be a breaking change |

### Why Option One is the right call

1. **This was an extraction oversight, not a design choice.** If `callTool`
   was intentionally thin, ReActAgent would use it. It doesn't. The stale
   `execute(...)` javadoc is the smoking gun — someone forgot to repoint the
   single-call path when they introduced `executeWithInfrastructure`.

2. **The change is tiny and safe.** `executeCore` (pure execution) is never
   touched. `executeAll` (batch path) is never touched. We are simply making
   the single-call path join the same validated pipeline.

3. **The "behaviour change" is aligned with the batch path.**
   `TOOL_DEFAULTS.timeout = 5min` — no normal tool exceeds that.
   `TOOL_DEFAULTS.maxAttempts = 1` — retry stays effectively off, exactly as
   before. ShutdownGuard is a protection layer, not a functional change.
   Scheduling matches the batch path. There is essentially no scenario where
   the new behaviour surprises a user, because it is the same behaviour they
   already get from the batch API.

4. ** pushes the problem forward.** Hardcoding "no timeout, no retry"
   into a public API with no naming signal creates a silent footgun. Every
   new user who reads the javadoc and picks `callTool` will get the broken
   semantics. When someone eventually fixes it properly, it becomes a
   breaking change for whoever started relying on the thin semantics.

---

### 1. `agentscope-core/.../tool/ToolExecutor.java` (+22 / -6)

**Before — a single `private` 4-param method that silently dropped fields:**

```java
private Mono<ToolResultBlock> executeWithInfrastructure(
        ToolUseBlock toolCall,
        ExecutionConfig executionConfig,
        Agent agent,
        RuntimeContext agentRuntimeContext) {
    ToolCallParam param = ToolCallParam.builder()
            .toolUseBlock(toolCall)
            .agent(agent)
            .runtimeContext(agentRuntimeContext)
            .build();
    // ⚠️ input, emitter, and any other ToolCallParam fields were dropped here.
    Mono<ToolResultBlock> execution = execute(param);
    execution = applyScheduling(execution);
    execution = applyTimeout(execution, executionConfig, toolCall);
    execution = applyRetry(execution, executionConfig, toolCall);
    execution = applyShutdownGuard(execution);
    return execution.map(result -> result.withIdAndName(toolCall.getId(), toolCall.getName()))
            .onErrorResume(...);
}
```

**After — refactored into two package-private overloads:**

```java
// Overload A (4 params — used by batch path executeAll, behaviour unchanged):
Mono<ToolResultBlock> executeWithInfrastructure(
        ToolUseBlock toolCall,
        ExecutionConfig executionConfig,
        Agent agent,
        RuntimeContext agentRuntimeContext) {
    ToolCallParam param = ToolCallParam.builder()
            .toolUseBlock(toolCall)
            .agent(agent)
            .runtimeContext(agentRuntimeContext)
            .build();
    return executeWithInfrastructure(param, executionConfig);   // delegate ↓
}

// Overload B (2 params — NEW core implementation, used by single-call path):
Mono<ToolResultBlock> executeWithInfrastructure(
        ToolCallParam param, ExecutionConfig executionConfig) {
    ToolUseBlock toolCall = param.getToolUseBlock();
    Mono<ToolResultBlock> execution = execute(param);   // still goes through Tracer wrapper

    execution = applyScheduling(execution);
    execution = applyTimeout(execution, executionConfig, toolCall);
    execution = applyRetry(execution, executionConfig, toolCall);
    execution = applyShutdownGuard(execution);

    return execution
            .map(result -> result.withIdAndName(toolCall.getId(), toolCall.getName()))
            .onErrorResume(e -> {
                logger.warn("Tool call failed: {}", toolCall.getName(), e);
                String errorMsg = ExceptionUtils.getErrorMessage(e);
                return Mono.just(
                        ToolResultBlock.error("Tool execution failed: " + errorMsg)
                                .withIdAndName(toolCall.getId(), toolCall.getName()));
            });
}
```

Key points:
- **Visibility widened** from `private` to package-private. Safe because the
  `ToolExecutor` class itself is package-private — we are just matching
  method visibility to class visibility.
- **Overload B preserves the full `ToolCallParam`** — notably `input`, which
  `executeCore` prioritises over `toolCall.getInput()`. The old 4-param
  version silently dropped this field; the new 2-param version does not.
- **Overload A delegates** to Overload B, so the batch path behaves
  identically to before. `executeAll` has zero code changes.

### 2. `agentscope-core/.../tool/Toolkit.java` (+6 / -1)

**Before:**
```java
public Mono<ToolResultBlock> callTool(ToolCallParam param) {
    return executor.execute(param);   // pure execution, no infrastructure
}
```

**After:**
```java
public Mono<ToolResultBlock> callTool(ToolCallParam param) {
    ExecutionConfig effectiveConfig =
            ExecutionConfig.mergeConfigs(
                    config.getExecutionConfig(), ExecutionConfig.TOOL_DEFAULTS);

    return executor.executeWithInfrastructure(param, effectiveConfig);
}
```

Config merge precedence is identical to `callTools`:
toolkit-level config → `TOOL_DEFAULTS` fallback (5min timeout, maxAttempts=1).

### 3. `agentscope-core/.../tool/ToolkitTest.java` (+41 / 0)

Two new tests covering both success and error paths:

```java
@Test
@DisplayName("callTool single should populate id and name on ToolResultBlock")
void testCallToolSinglePopulatesIdAndName() {
    toolkit.registerTool(sampleTools);
    ToolUseBlock toolCall = ToolUseBlock.builder()
            .id("call-single-001").name("add")
            .input(Map.of("a", 2, "b", 3)).build();

    ToolResultBlock result = toolkit
            .callTool(ToolCallParam.builder().toolUseBlock(toolCall).build())
            .block();

    assertNotNull(result);
    assertEquals("call-single-001", result.getId());   // was null
    assertEquals("add", result.getName());              // was null
}

@Test
@DisplayName("callTool single should populate id and name on error results too")
void testCallToolSingleErrorResultAlsoHasIdAndName() {
    toolkit.registerTool(sampleTools);
    ToolUseBlock toolCall = ToolUseBlock.builder()
            .id("call-single-err").name("error_tool")
            .input(Map.of("message", "boom")).build();

    ToolResultBlock result = toolkit
            .callTool(ToolCallParam.builder().toolUseBlock(toolCall).build())
            .block();

    assertNotNull(result);
    assertEquals("call-single-err", result.getId());    // was null
    assertEquals("error_tool", result.getName());       // was null
}
```

No existing tests were modified or weakened.

---

## Behaviour Impact Summary

| Area | Change | Risk |
|---|---|---|
| `callTools` (batch) | **ZERO change** — 4-param overload still delegates identically | None |
| `callTool` id/name | Now populated ✅ | Pure bug fix |
| `callTool` timeout | Now 5min default (`TOOL_DEFAULTS`) ⚠️ | Minimal — matches batch; 5min is generous for any normal tool |
| `callTool` retry | Still effectively **off** (`maxAttempts=1`) | None — identical to batch default |
| `callTool` ShutdownGuard | Now covers single calls ✅ | Improvement — matches batch |
| `callTool` scheduling | Now runs on `boundedElastic` ⚠️ | Minimal — matches batch; ThreadLocal-dependent tool code is rare |
| `executeCore` | **NOT TOUCHED** | None — pure-execution layer untouched, zero regression risk |

---

## Test Results

Ran on `agentscope-core` module:

```
[INFO] Tests run: 12, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.573 s
    -- in io.agentscope.core.tool.ToolExecutorTest

[INFO] Tests run: 51, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.088 s
    -- in io.agentscope.core.tool.ToolkitTest
        (includes 2 new tests:
         testCallToolSinglePopulatesIdAndName
         testCallToolSingleErrorResultAlsoHasIdAndName)

[INFO] BUILD SUCCESS
```

**63 tests, 0 failures, 0 errors.**

---

## How to Verify

1. Run just the two new tests:
   ```bash
   mvn test -pl agentscope-core \
       "-Dtest=ToolkitTest#testCallToolSinglePopulatesIdAndName+testCallToolSingleErrorResultAlsoHasIdAndName"
   ```

2. Run the full Toolkit + ToolExecutor suites:
   ```bash
   mvn test -pl agentscope-core "-Dtest=ToolkitTest,ToolExecutorTest"
   ```

3. Reproduce the original issue: build `CallToolIdNameReproTest` from #3114
   against this branch. Before fix it printed:
   ```
   REPRO callTool id=null name=null
   REPRO callTools id=call-batch name=echo
   ```
   After fix both lines show proper ids and names.

---

### Why not patch `withIdAndName` directly in `execute` or `executeCore`?

That would only stamp id/name. The same divergence also skips scheduling,
timeout, retry, and ShutdownGuard — four more bugs left unfixed. And it would
require duplicating the infrastructure layers somewhere else instead of
reusing the single wrapper that already exists.

### Why keep the 4-param overload?

`executeAll` currently calls it, and changing the batch path adds unnecessary
risk. The 4-param overload is now a thin shim that just delegates to the
2-param one. We could inline it later in a follow-up PR but keeping both
minimises the diff and guarantees zero behavioural change for `callTools`.

### The `input` field got fixed as a bonus

While refactoring, I noticed the original 4-param `executeWithInfrastructure`
rebuilt a `ToolCallParam` from only `toolUseBlock`, `agent`, and
`runtimeContext` — silently dropping `input`, which `executeCore` prioritises
over `toolCall.getInput()`. That was a latent data-loss bug that never
surfaced because the batch path always sets input on `ToolUseBlock`. The
new 2-param overload receives the **full original `ToolCallParam`**, so all
user-supplied fields survive untouched.


agentscope-core/src/main/java/io/agentscope/core/tool/Toolkit.java
```diff
@@ -488,7 +488,11 @@ public void setInternalChunkCallback(BiConsumer<ToolUseBlock, ToolResultBlock> c
      * @return Mono containing execution result
      */
     public Mono<ToolResultBlock> callTool(ToolCallParam param) {
-        return executor.execute(param);
+        ExecutionConfig effectiveConfig =
+                ExecutionConfig.mergeConfigs(
+                        config.getExecutionConfig(), ExecutionConfig.TOOL_DEFAULTS);
+
+        return executor.executeWithInfrastructure(param, effectiveConfig);
```
oss-maintainer, line 495: **[Warning] Compatibility — `callTool` semantics change silently for a public API.**

Routing `callTool` through `executeWithInfrastructure` does more than populate `id`/`name`: it also (a) converts failures into a *successful* `Mono` carrying `ToolResultBlock.error(...)` (`onErrorResume` in `executeWithInfrastructure`), so caller-side `onErrorResume` / `retryWhen` / `try-catch` around `callTool(...).block()` no longer sees exceptions; (b) applies the 5-minute `TOOL_DEFAULTS` timeout and `applyShutdownGuard`; (c) moves execution onto the toolkit scheduler via `applyScheduling`; (d) enables retry if `config.getExecutionConfig()` carries `maxAttempts > 1` — that can **double-execute non-idempotent tools** for callers that never had retries on this path before.

The `id`/`name` regression itself is real and worth fixing. Could you make the blast radius explicit: split "attach id/name" from "apply full infrastructure", or at minimum document the new behavior in the javadoc + changelog so downstream (harness / extensions / studio) callers are not caught by surprise?

agentscope-core/src/main/java/io/agentscope/core/tool/Toolkit.java
```diff
@@ -488,7 +488,11 @@ public void setInternalChunkCallback(BiConsumer<ToolUseBlock, ToolResultBlock> c
      * @return Mono containing execution result
      */
     public Mono<ToolResultBlock> callTool(ToolCallParam param) {
```
oss-maintainer, line 490: **[Info] Please extend the javadoc of this public method.** It currently only says "Execute multiple tools..." style docs; after this change `callTool` applies scheduling, timeout, retry and shutdown guard and returns errors as results rather than `Mono.error`. Worth one paragraph, plus a note that the effective config is `mergeConfigs(toolkitConfig, TOOL_DEFAULTS)` (i.e. no agent-level config is honoured here, unlike `callTools`).

agentscope-core/src/main/java/io/agentscope/core/tool/ToolExecutor.java
```diff
@@ -370,30 +370,42 @@ private boolean isConcurrencySafe(ToolUseBlock toolCall) {
...
+    /**
+     * Execute a single tool call with full infrastructure, preserving all fields from the
+     * original {@link ToolCallParam} (including input).
+     *
+     * <p>This overload is used by {@code Toolkit.callTool} so that user-supplied fields on the
+     * param object are not silently discarded before reaching {@link #executeCore}.
+     */
+    Mono<ToolResultBlock> executeWithInfrastructure(
+            ToolCallParam param, ExecutionConfig executionConfig) {
+        ToolUseBlock toolCall = param.getToolUseBlock();
```
oss-maintainer, line 400: **[Info] The two overloads are now asymmetric in field preservation.** The 4-arg overload still rebuilds a fresh `ToolCallParam` from `toolCall`/`agent`/`runtimeContext` (so any `input`/`emitter` attached to an existing param is not carried through), while the new overload preserves the param verbatim. That is fine for today's call sites (`executeAll` only has the `ToolUseBlock`), but the doc comment here ("preserving all fields ... not silently discarded") reads as if the batch path had the same guarantee. Consider adding a note that the batch path intentionally constructs the param, or unify both to accept a `ToolCallParam`.

agentscope-core/src/test/java/io/agentscope/core/tool/ToolkitTest.java
```diff
@@ -1290,4 +1290,45 @@ void testGetToolSchemasIncludesMcpOutputSchema() {
...
+
+        ToolUseBlock toolCall =
+                ToolUseBlock.builder()
+                        .id("call-single-001")
+                        .name("add")
+                        .input(Map.of("a", 2, "b", 3))
+                        .build();
+
+        ToolResultBlock result =
+                toolkit.callTool(ToolCallParam.builder().toolUseBlock(toolCall).build()).block();
```
oss-maintainer, line 1308: **[Warning] Tests cover the symptom but not the stated motivation.** Both new tests assert `id`/`name` only. The PR description motivates the second overload by "preserving all fields from the original `ToolCallParam` (including `input`)" — there is no test that a `ToolCallParam` carrying `input` (or preset parameters / `emitter`) still reaches the tool unchanged, which is exactly the regression the overload guards against. Two more cases worth adding:

1. `input`/preset-parameter preservation through `callTool` (guards the overload split).
2. Timeout/retry actually taking effect on the `callTool` path (e.g. toolkit config with a small `timeout` and a slow tool, asserting the error result still carries `id`/`name`).

Also `error_tool` in `SampleTools` throws `RuntimeException`, so `executeCore` returns `ToolResultBlock.error(...)` for the *caught* case — the second test therefore does not really exercise the `onErrorResume` branch; a tool that propagates an error `Mono` would.

agentscope-core/src/main/java/io/agentscope/core/tool/Toolkit.java
```diff
@@ -488,7 +488,11 @@ public void setInternalChunkCallback(BiConsumer<ToolUseBlock, ToolResultBlock> c
...
-        return executor.execute(param);
+        ExecutionConfig effectiveConfig =
+                ExecutionConfig.mergeConfigs(
```
oss-maintainer, line 492: **[Info] Nit:** `effectiveConfig` is recomputed on every call. `config.getExecutionConfig()` looks immutable here, so this could be resolved once (field or lazily-cached) and reused; the same merge is repeated in `callTools`. A small shared private helper would keep the two paths from drifting.

oss-maintainer (commented):
### Summary
Routes the single-call path `Toolkit.callTool` through `executeWithInfrastructure` (new `ToolCallParam`-taking overload) so results keep `id`/`name`, plus two regression tests. The bug is real and the direction is right, but the change also swaps the error contract, threading, timeout and retry behaviour of a public API — please confirm the blast radius and cover the param-preservation claim in tests.

### Findings
- **[Warning]** `Toolkit.java:495` — public `callTool` semantics change: errors become results, 5-min timeout + shutdown guard + scheduler switch, and retry enabled from toolkit config (double-execution risk for non-idempotent tools).
- **[Warning]** `ToolkitTest.java:1308` — no test for `input`/param preservation (the overload's stated purpose); the error-path test does not exercise `onErrorResume`; timeout/retry on this path untested.
- **[Info]** `Toolkit.java:490` — document the new behaviour on this public method.
- **[Info]** `ToolExecutor.java:400` — the two overloads are asymmetric in field preservation; clarify or unify.
- **[Info]** `Toolkit.java:492` — the merged config could be resolved once and shared with `callTools`.

### Suggestions
- Keep the good part minimal if you prefer a low-risk change: apply `.map(r -> r.withIdAndName(...))` (and the matching error mapping) inside `Toolkit.callTool` / `ToolExecutor.execute(param)` without pulling in scheduling/timeout/retry, then open a separate PR (with changelog + migration notes) for the full-infrastructure routing.
- Note the merge order here differs from `callTools`: no agent-level `ExecutionConfig` participates — worth stating explicitly so per-request config is not expected to apply.
- CI is currently `BLOCKED` (missing required review); CLA status is not reported by any commit status check on this head, so a maintainer may want to confirm the contributor has signed before merge.

---

oss-maintainer:
**CLA Not Signed**

@KIM406-CMD please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.

---

oss-maintainer:
Thanks for the update — a quick correction on the CLA gate and the CI signal on this head.

**CLA:** `license/cla` now reports **success** on `bb72ae0` ("All contributors have signed the Contributor License Agreement"), so you are all clear here — please disregard my earlier CLA reminder comment on this PR, which was raised against a stale pending status.

**CI:** the two failing jobs are `build (ubuntu-latest)` / `build (windows-latest)`, and the failure is in `agentscope-extensions-aistio` surefire tests, not in `agentscope-core` where this change lives:

```
[ERROR] Failed to execute goal ...maven-surefire-plugin:3.5.5:test (default-test) on project agentscope-extensions-aistio: There are test failures.
```

Worth one re-run before concluding it is yours; if it reproduces on a clean `origin/main` merge, it is pre-existing and unrelated to this PR.

**Still open from my review** (unchanged, `bb72ae0` is the commit I reviewed):
1. `Toolkit.callTool` now inherits the infrastructure path's error contract (errors become tool-result values), the default timeout / shutdown guard and toolkit-level retry — that is a behavioural change on a public API, so a note on intended blast radius (or a narrower `.map(r -> r.withIdAndName(...))` variant) would make this easier to land.
2. The new tests cover id/name preservation on success, but not the `input`/param pass-through that motivates the overload, nor the error/timeout/retry path.

No action needed on the CLA item. Feel free to @mention me for a re-review once the above is addressed.

---

oss-maintainer:

This PR now conflicts with `main` and cannot be merged. Please rebase or merge `main` into your branch and resolve the conflicts:

```bash
git fetch origin
git checkout fix
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

CI on this head also shows a failing/cancelled check — please re-run it after the rebase.

---

KIM406-CMD:
Thanks for the detailed review and the follow-up — all points noted.
CLA: confirmed green, no action from my side. Thanks for correcting the stale reminder.
Merge conflict: will rebase onto origin/main and resolve the ToolkitTest.java conflict, then push with --force-with-lease and re-run CI. If agentscope-extensions-aistio still fails on a clean origin/main merge, I'll treat it as pre-existing and unrelated to this PR (the change is in agentscope-core) and link the evidence.
On the two Warnings — I'll take the low-risk split:
1. This PR will keep only the minimal fix: attach id/name (plus the matching error mapping) inside Toolkit.callTool / ToolExecutor.execute(param), without pulling in scheduling, timeout, shutdown guard, or retry. That keeps the public callTool error contract intact.
2. The full-infrastructure routing through executeWithInfrastructure will go into a separate follow-up PR, with javadoc + changelog + migration notes, and an explicit note that no agent-level ExecutionConfig participates here (unlike callTools).
Tests I'll add:
· input/preset-parameter pass-through through callTool (guards the overload split).
· Timeout/retry actually taking effect on this path, asserting the error result still carries id/name.
· A tool that propagates Mono.error(...) so onErrorResume is genuinely exercised — you're right that error_tool's caught RuntimeException never reaches it.
I'll also address the Info/Nits: javadoc on callTool, the overload-asymmetry note, and the shared resolveEffectiveConfig() helper.
Will @mention you once pushed. Thanks!

KIM406-CMD:
@oss-maintainer rebased onto upstream/main, ToolkitTest conflict resolved (kept both upstream's MCP propagateMeta tests and our callTool id/name tests). CI is fully green — ready for re-review!

agentscope-core/src/main/java/io/agentscope/core/tool/Toolkit.java
```diff
@@ -489,7 +489,11 @@ public void setInternalChunkCallback(BiConsumer<ToolUseBlock, ToolResultBlock> c
...
+                        config.getExecutionConfig(), ExecutionConfig.TOOL_DEFAULTS);
+
+        return executor.executeWithInfrastructure(param, effectiveConfig);
```
oss-maintainer: **[Warning]** Routing `callTool` through `executeWithInfrastructure` also hands it a timeout and a retry policy it never had before. Worth being explicit about the real blast radius, because it is narrower but stranger than "errors become results":

- Tool *exceptions* are unaffected. `executeCore` already converts them into `ToolResultBlock.error(...)` values (and empty completions via `switchIfEmpty`) before any of these operators wrap it, so the new `onErrorResume` here only catches what escapes outside `executeCore`.
- `TOOL_DEFAULTS` is `maxAttempts(1)`, and `applyRetry` returns the source unchanged for `maxAttempts <= 1`, so nothing changes for the default toolkit.
- The risk is for a toolkit built with `ToolkitConfig.executionConfig(...)` carrying `maxAttempts > 1` (previously honoured only by the batch path). There, a tool that is merely *slow* now times out at 5 minutes and gets **re-subscribed**, so a non-idempotent tool can execute twice where `callTool` used to attempt it exactly once. Timeouts are the one signal that does reach `applyRetry`, since `applyTimeout` is applied before it.

That is probably the behaviour you want, but it is a semantic change on a public API that a caller cannot see from the signature. Could you say so in the method javadoc (and the release notes) in one line, and add a test that pins `maxAttempts > 1` + a slow tool on this path?

agentscope-core/src/test/java/io/agentscope/core/tool/ToolkitTest.java
```diff
@@ -1371,4 +1370,45 @@ void testRegistrationPerToolPropagateMetaRejectsNullName() {
...
+    @DisplayName(
+            "callTool single should populate id and name on ToolResultBlock (was null before fix)")
+    void testCallToolSinglePopulatesIdAndName() {
+        toolkit.registerTool(sampleTools);
+
+        ToolUseBlock toolCall =
+                ToolUseBlock.builder()
+                        .id("call-single-001")
+                        .name("add")
+                        .input(Map.of("a", 2, "b", 3))
```
oss-maintainer: **[Warning]** Neither new test exercises the thing this overload's javadoc says it protects. Both build `ToolCallParam.builder().toolUseBlock(toolCall).build()`, so `param.getInput()` is empty and `executeCore` takes the `else if (!toolCall.getInput().isEmpty())` branch at ToolExecutor.java:249 — the `param.getInput()` precedence branch at line 247 is never run.

So the tests correctly pin the `id`/`name` fix (the user-visible bug), but "preserving all fields from the original `ToolCallParam` (including input)" is currently guaranteed only by the fact that the new overload forwards `param` instead of rebuilding it from the `ToolUseBlock`. A future caller that reaches the 4-arg overload instead would silently drop a param-level `input` and every test here would still pass. A third case — same `toolUseBlock`, plus `param.input(...)` overriding one key — would make the claim actual rather than aspirational.

agentscope-core/src/test/java/io/agentscope/core/tool/ToolkitTest.java
```diff
@@ -1299,7 +1299,6 @@ void testRegistrationPropagateMetaOverridesWrapper() {
         McpClientWrapper mcpClientWrapper =
                 McpClientWrapperTestSupport.mockWrapper("external-mcp-client", true);
         when(mcpClientWrapper.initialize()).thenReturn(Mono.empty());
-        // Wrapper explicitly allows propagation; registration-level setting must win
```
oss-maintainer: **[Info]** This hunk is unrelated to the fix and looks like fallout from resolving the `ToolkitTest` conflict: it deletes a comment line belonging to upstream's `propagateMeta` test (`// Wrapper explicitly allows propagation; registration-level setting must win`) with nothing replacing it. The code below still depends on that intent, so the explanation is just lost.

Please restore it so the rebase doesn't quietly remove coverage documentation from someone else's test — easy to re-add by accident on the next conflict resolution.

oss-maintainer (commented):
### Summary

Re-review after the rebase onto `main` (head `f15d7894`). The rebase is clean and CI is green, and the fix itself is correct: `callTool` was bypassing `executeWithInfrastructure`, so single-call results lost the `withIdAndName(...)` decoration the batch path applies, and the two new tests pin exactly that. CLA signed, no conflicts, and the `ToolkitTest` merge kept both upstream's `propagateMeta` cases and the new `callTool` ones.

I owe the author a correction from my previous review on this PR: I claimed routing through the infrastructure layer changes the error contract so that "errors become results". That was wrong — `executeCore` already ends in `onErrorResume(...) → ToolResultBlock.error(...)` plus a `switchIfEmpty`, both before this PR, so tool exceptions were already materialised as error values here. What the wrapping genuinely adds is the timeout / retry / shutdown-guard / scheduler layers, which is what the first finding below now describes.

- **[Warning]** `Toolkit.java:496` — `callTool` now inherits a timeout and, for toolkits configured with `maxAttempts > 1`, a retry policy it previously never had. Because `applyTimeout` is applied before `applyRetry`, a merely *slow* non-idempotent tool can be re-subscribed and run twice. Default `TOOL_DEFAULTS` is `maxAttempts(1)`, so nothing changes out of the box, but it is an invisible semantic change on a public API.
- **[Warning]** `ToolkitTest.java:1384` — both new tests set `input` on the `ToolUseBlock`, so `param.getInput()` is empty and `executeCore` takes the fallback branch; the `param.input(...)` precedence branch that the new overload's javadoc says it is protecting is untested.
- **[Info]** `ToolkitTest.java:1301` — the rebase drops an upstream comment line in `testRegistrationPropagateMetaOverridesWrapper` with nothing replacing it.

None of these is a defect in the change itself, so this stays a COMMENT rather than a block: the first is a docs/test request, the second coverage hygiene, the third a one-line restore. Happy to approve once the timeout/retry semantics are stated on the method.

---

PR #3131, opened 2026-09-13
feat(agent): support recovery mechanism for streaming tool call errors

### LINKED ISSUE #3102: [Bug]:ReActAgent 调用不存在的工具时直接退出 ReAct 循环，未给模型重试机会
# ReActAgent 调用不存在的工具时直接退出 ReAct 循环，未给模型重试机会

## 问题描述

当 LLM 在流式响应中尝试调用一个不在 tools schema 里的工具（如把 Skill 名 `metric_condition_generation` 当工具调用），LLM 服务端会返回 error。agentscope 的 `OpenAIClient` 把这个 error 包装为 `OpenAIException` 抛出，`ReActAgent.reasoning()` 只捕获 `InterruptedException`，其他异常直接传播终止整个 `streamEvents` 流。

模型没有机会纠正错误、换一个工具或改用纯文本回复——整个对话直接以 error 结束。

## 复现场景

1. 配置了一个 Agent，有 tools 但没有 `metric_condition_generation` 这个工具
2. 模型在 reasoning 阶段幻觉式调用了 `metric_condition_generation`（这是一个 Skill 名，不是工具名）
3. LLM 服务端在流式响应中返回 error：`LLM is trying to invoke a non-exist tool: "metric_condition_generation"`
4. `OpenAIClient` 把 error 包装为 `OpenAIException` 并抛出
5. `ReActAgent.reasoning()` 的 `onErrorResume` 只捕获 `InterruptedException`，`OpenAIException` 直接传播
6. 整个 `streamEvents` 流以 error 终止，对话直接结束

## 实际结果

```
OpenAI API error in streaming response: LLM is trying to invoke a non-exist tool: "metric_condition_generation", you can add some few shots examples or adjust the prompt.
```

对话直接终止，用户看到错误，没有结果。

## 期望结果

当 LLM 尝试调用不存在的工具时，框架应该：
1. 捕获这个 error，不终止整个 ReAct 循环
2. 构造一个 `ToolResultBlock.error("工具 X 不存在，请使用已注册的工具或直接回复用户")` 作为该 tool_call 的结果
3. 把这个 error 结果加入上下文，继续下一轮 reasoning——让模型看到错误后自行纠正
4. 如果模型多次（如 2~3 次）仍然调用不存在的工具，才终止循环

### 错误产生路径

```
LLM 返回 tool_call(name="metric_condition_generation")
  → LLM 服务端发现该工具不在 tools schema 里
    → 流式响应返回 error
      → OpenAIClient/OpenAIResponseParser 解析 error
        → 抛出 OpenAIException("OpenAI API error in streaming response: LLM is trying to invoke a non-exist tool: ...")
          → ReActAgent.reasoning() 的 model.stream() 报错
            → onErrorResume 只捕获 InterruptedException
              → OpenAIException 直接传播
                → streamEvents 流终止
```

### 涉及的代码

**OpenAIClient**（agentscope-extensions-model-openai）：
- `OpenAIClient.stream()` 解析流式响应时，发现 `OpenAIResponse.isError()` 为 true
- 构造 `OpenAIException` 并通过 `SynchronousSink.error()` 抛出

**OpenAIResponseParser**（agentscope-extensions-model-openai）：
- 同样有 "OpenAI API error in streaming response" 字符串

**ReActAgent.CallExecution.reasoning()**（agentscope-core）：
- `model.stream().doOnNext(this::publishEvent).then(...)` 的 stream 因 error 终止
- `onErrorResume(InterruptedException.class, ...)` 只捕获 `InterruptedException`
- 其他异常（如 `OpenAIException`）直接传播，终止整个 reasoning 循环

### 对比 ToolExecutor 的处理

`ToolExecutor.invokeTool()` 在工具不存在时（`toolRegistry.getTool(name)` 返回 null），会构造 `ToolResultBlock.error("Tool not found: " + name)` 并正常返回，不抛异常。这给了模型重试的机会。

但 LLM 服务端在流式阶段就拦截了——工具名不在 tools schema 里，服务端直接返回 error，不走 agentscope 的 `ToolExecutor`。所以 `ToolExecutor` 的兜底逻辑在这个场景下不生效。

### 方案 A：在 reasoning 的 onErrorResume 中捕获 OpenAIException

```java
// ReActAgent.CallExecution.reasoning() 里
.onErrorResume(OpenAIException.class, error -> {
    // 构造一个 error ToolResultBlock，让模型在下一轮看到错误后自行纠正
    Msg errorMsg = Msg.builder()
        .role(MsgRole.ASSISTANT)
        .content(TextBlock.builder().text("工具调用失败: " + error.getMessage()).build())
        .build();
    state.contextMutable().add(errorMsg);
    // 继续下一轮 reasoning，给模型纠正的机会
    return reasoning(iter + 1, true);  // ignoreMaxIters=true 保证至少重试一次
})
```

### 方案 B：在 OpenAIClient 里不抛异常，改为返回 error ChatResponse

```java
// OpenAIClient.stream() 里解析到 error response 时
if (response.isError()) {
    // 不抛异常，改为构造一个携带错误信息的 ChatResponse
    return Flux.just(ChatResponse.error(response.getEffectiveErrorMessage()));
}
```

这样 `ReActAgent.modelCallStream` 的 `concatMap` 会正常处理这个 error ChatResponse，可以转换为 error event 发给下游，不会终止流。

### 推荐方案 B

方案 A 在 `reasoning` 层兜底，但 error 的语义不明确（是工具不存在还是网络错误？）；方案 B 在源头处理，让 error 成为流中的一个正常事件，框架可以在 `concatMap` 里把它转成 error `ToolResultBlock`，和 `ToolExecutor` 的 "Tool not found" 处理对齐。

## 环境信息

- agentscope-core: 2.0.3
- agentscope-extensions-model-openai: 2.0.3
- JDK: 17
- 模型服务: 百炼 qwen-plus / antchat Qwen3.5-35B-A3B

thinkw:
## AgentScope-Java Version

`2.0.3-SNAPSHOT`, based on main `39cd304a`.

## Description

Fixes #3102.

### Problem

When the model hallucinates a tool name during streaming, the provider rejects
the call mid-stream and `OpenAIClient` surfaces it as an `OpenAIException`
(`OpenAI API error in streaming response: LLM is trying to invoke a non-exist
tool: ...`). `ReActAgent.reasoning()` only resumes on `InterruptedException`, so
the exception terminates the whole `streamEvents` flow — the turn dies with an
error and the model never gets a chance to self-correct. The local
`ToolResultBlock.error("Tool not found: ...")` handling in `ToolExecutor` is
unreachable for this path because the LLM service rejects the name before the
agent-side executor is ever invoked.

### Solution

Recovery now happens in `ReActAgent.modelCallStream()`, at the one layer that
has both the error and the accumulated conversation state:

```mermaid
flowchart TD
    A["Model streams tool_call chunk"] --> B["ReasoningContext accumulates ToolUseBlock"]
    B --> C["Provider error: unknown tool"]
    C --> D{"isRecoverableStreamError?<br/>(ModelHttpException + non-retryable<br/>+ tool/function keyword)"}
    D -- "yes, budget left" --> E["Stream completes normally<br/>(recovery budget +1)"]
    E --> F["acting executes the hallucinated call"]
    F --> G["ToolExecutor returns ToolResultBlock.error<br/>'Tool not found: X' tied to call id/name"]
    G --> H["Next reasoning round:<br/>model sees the error and self-corrects"]
    D -- "yes, budget exhausted" --> I["Original error propagates, turn fails"]
    D -- no --> I
    H -. "a later stream completes without recovery" .-> J["Recovery budget resets to 0"]
```

Key properties:

1. **Original assistant/tool-call intent is preserved.** Chunks accumulated
   before the error stay in the `ReasoningContext`; `buildFinalMessage()` emits
   the assistant message with the tool call, and the acting phase executes it.
   The fed-back error is a structured `ToolResultBlock` tied to the attempted
   tool-call id/name (via `ToolResultMessageBuilder`), not a free-form assistant
   text — mirroring `ToolExecutor`'s existing "Tool not found" handling instead
   of duplicating it.
2. **Narrow error classification (`isRecoverableStreamError`).** All three must
   hold:
   - the exception implements `ModelHttpException` (so every provider benefits:
     OpenAI, DashScope, Anthropic, Gemini, Ollama, ...);
   - the HTTP status is non-retryable — 429 rate limits and 5xx server errors
     still fail fast and remain the transport-level retry mechanism's job;
   - the message contains `tool` / `function` / `invalid_tool` keywords
     suggesting the provider rejected an unknown tool name.
3. **Configurable consecutive-recovery budget.** New
   `ReactConfig.max_tool_error_recoveries` (default 3, JSON key
   `max_tool_error_recoveries`, builder `maxToolErrorRecoveries(int)`). Each
   recovered streaming error increments a per-call counter; once exhausted, the
   original provider error propagates and fails the turn with a clear warning —
   a confused model cannot loop forever on hallucinated names.
4. **"Consecutive" semantics.** A model stream that completes *without*
   recovery resets the budget (`doOnComplete` on the model-call stream). The
   normal `maxIters` iteration budget still applies throughout — no
   `ignoreMaxIters` bypass.

### Why not inside `OpenAIClient` (the issue's option B)?

The issue author suggested converting the error chunk into an error
`ChatResponse` inside the OpenAI client. This PR deliberately keeps provider
error semantics intact and recovers at the agent layer instead:

- the client layer has no visibility into the `ReasoningContext`, so it cannot
  know whether tool calls were already accumulated (the precondition for a
  meaningful recovery);
- unconditional error-chunk-to-response conversion there would silently lose
  fail-fast semantics for auth/rate-limit/schema errors, which the issue
  discussion explicitly wants kept;
- classifying on the `ModelHttpException` interface keeps the fix
  provider-agnostic — no new typed capability on `ChatResponse` is needed, and
  no OpenAI-extension-only behavior is introduced.

The end state in the conversation is exactly what option B aimed for (assistant
tool-call intent + structured tool error + continued reasoning), achieved
through the existing acting/executor path.

### Changes

| File | Change |
|------|--------|
| `agentscope-core/.../ReActAgent.java` | Recovery (`onErrorResume`) in `modelCallStream()`; `isRecoverableStreamError()` classification; per-call `streamToolErrorRecoveries` budget with reset-on-clean-completion; `maxToolErrorRecoveries` agent field + builder method |
| `agentscope-core/.../agent/config/ReactConfig.java` | New `max_tool_error_recoveries` component (default 3, validated > 0, JSON-creator back-compat with persisted states, 2-arg convenience constructor kept for source compatibility) |
| `agentscope-core/.../agent/config/ReactConfigTest.java` | Coverage for the new config field |
| `agentscope-core/.../agent/ReActAgentStreamErrorRecoveryTest.java` | New test class (see below) |

### Testing

All scenarios called out in the issue discussion are covered:

- **Recoverable**: one unknown-tool error followed by a valid response (model
  called again, turn completes); `function`-keyword variant.
- **Budget**: 3 consecutive unknown-tool errors recover, the 4th propagates
  (exactly 4 model calls); budget resets after an intervening successful
  stream (scripted err/err/err → valid-tool → err → text completes all 6
  calls and returns the final text); custom cap via the builder (cap=1 → 2nd
  error propagates).
- **Non-recoverable still propagates**: plain `RuntimeException`, 500, 429,
  no-accumulated-tool-calls, and non-tool-related messages.

Full `agentscope-core` suite passes: **2350 tests, 0 failures** (8 pre-existing
skips).

### Also fixed (unrelated to #3102, found by this PR's CI)

- #3132 — `HttpSelfRegistration.close()` could skip the control-plane
  DELETE when a heartbeat re-register cycle was in flight (flaky test on a
  loaded Windows runner; real instance leak).

### Known follow-ups (out of scope here)

- Surface registered tool names in the fed-back error hint (discussed as
  optional / "only if safe to expose" in the issue).
- `HarnessAgent.Builder.fromAgent()` currently inherits `stopOnReject` only;
  migrated agents fall back to the default recovery cap of 3.

## Checklist


oss-maintainer (commented):
### Summary
Well-scoped fix for #3102 with an unusually thorough test suite (551 new lines covering the recoverable, budget-exhausted, budget-reset and non-recoverable paths), and the choice to recover in `ReActAgent.modelCallStream()` rather than inside `OpenAIClient` is the right one — it is the only layer that can see both the provider error and the accumulated `ReasoningContext`. I traced the happy path through `buildFinalMessage()` → `runPostReasoningPipeline()` (`state.contextMutable().add(eventMsg)`) → `acting()` and the assistant message with the hallucinated call does reach persisted state, so the model genuinely gets `Tool not found` feedback.

The reservations are all about the classification being looser than the design assumes: the `isRetryableHttpStatus()` guard is a no-op whenever a provider exception carries no status code, and `getAllAccumulatedToolCalls()` cannot distinguish a complete tool call from one truncated mid-arguments. Recommend tightening both before merge; the rest are nits.

Not approving as review feedback only (and the CLA is still unsigned) — merge decision stays with the maintainers.

### Findings
- **[Warning]** `ReActAgent.java:2649` — keyword heuristic is the only real guard when `getStatusCode()` is null; tighten with provider error codes.
- **[Warning]** `ReActAgent.java:2650` — recovery can fire on a tool call truncated mid-arguments and invoke a real tool with corrupt input.
- **[Warning]** `ReActAgent.java:2661` — unsanitized provider error text logged at `warn`; add truncation/redaction.
- **[Warning]** `ReActAgent.java:4856` — `ReactConfig.maxToolErrorRecoveries` is not read back by the builder, so config/persisted-state routes ignore it.
- **[Info]** `ReActAgent.java:2687` — `doOnComplete` misses cancellation, so "consecutive" reset is best-effort.
- **[Info]** `ReActAgent.java:2864` — `invalid_tool` branch is subsumed by the `tool` branch.
- **[Info]** `ReactConfig.java:36` — wire compatibility OK; flag the default-on behaviour change in release notes and consider a disabled (0) escape hatch.

### Suggestions
```java
// illustrative only — tune the alternatives against the real provider error
// catalogue ("non-exist tool", "no such function", "tool_not_found", ...)
// instead of relying on an open-ended substring match.
private static final Pattern UNKNOWN_TOOL = Pattern.compile(
        "(non-?exist|not found|does not exist|unknown|invalid|unavailable)"
      + "[a-z_ ]{0,20}(tool|function)"
      + "|(tool|function)[a-z_ ]{0,20}"
      + "(non-?exist|not found|does not exist|unknown|unavailable)",
        Pattern.CASE_INSENSITIVE);

private boolean isRecoverableStreamError(Throwable error) {
    if (!(error instanceof ModelHttpException he)) {
        return false;
    }
    Integer status = he.getStatusCode();
    // no status => cannot prove fail-fast is safe; require an exact error code instead
    if (he.isRetryableHttpStatus()) {
        return false;                       // 429 / 5xx stay with transport retry
    }
    String message = error.getMessage();
    return message != null
            && (UNKNOWN_TOOL.matcher(message).find()
                || RECOVERABLE_ERROR_CODES.contains(providerErrorCode(error)));
}
```
and for the completeness gate:
```java
List<ToolUseBlock> pending = context.getAllAccumulatedToolCalls();
boolean streamComplete = pending.stream().allMatch(t ->
        toolkit.getToolSchema(t.getName()) != null
                && (t.getContent() == null || t.getContent().isBlank()
                    || JsonUtils.isValidJsonObject(t.getContent())));
if (isRecoverableStreamError(error) && !pending.isEmpty() && streamComplete) { ... }
```

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2611,7 +2635,40 @@ private Flux<AgentEvent> modelCallStream(
...
+                            // hallucinated tool names). When the model accumulated tool calls
+                            // before the error, let the stream complete normally so the acting
+                            // phase can execute the tools and return "Tool not found" results,
+                            // giving the model a chance to self-correct. Consecutive recoveries
+                            // are capped by maxToolErrorRecoveries so a confused model cannot
+                            // loop forever on hallucinated tool names.
+                            .onErrorResume(
+                                    error -> {
+                                        if (isRecoverableStreamError(error)
+                                                && !context.getAllAccumulatedToolCalls()
```
oss-maintainer, line 2657: **Recoverability hinges on a keyword heuristic that `isRetryableHttpStatus()` rarely guards.** `OpenAIException.create(...)` only sets a status code when it is non-null, and several constructors (including `(message, errorCode, responseBody)`, which is the shape a provider *error chunk inside a 200 SSE stream* produces) leave `getStatusCode() == null`. For those, `ModelHttpException.isRetryableHttpStatus()` returns `false`, the "429/5xx still fail fast" guarantee in the PR description silently does not apply, and the only thing preventing a spurious recovery is `message.contains("tool"|"function")`. A 401/403 or content-policy error whose text happens to contain "function"/"tool" would then be swallowed and re-reported to the model as `Tool not found`, masking the real failure. Suggest tightening the predicate: require an explicit provider error code (`OpenAIException` already exposes `errorCode` — e.g. `invalid_parameter_error`, or a status in 400..499 that is not 401/403/422) *and* a tool-specific phrase (`not found` / `does not exist` / `unavailable` next to `tool|function`) instead of the bare substring match.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2611,7 +2635,40 @@ private Flux<AgentEvent> modelCallStream(
...
+                                                        .isEmpty()) {
```
oss-maintainer, line 2658: `getAllAccumulatedToolCalls().isEmpty()` proves *something* started streaming, not that a tool call is *complete*. Providers emit `tool_calls` incrementally, so the very common failure is an error arriving mid-arguments while `name` is already set. On that path the recovery still fires, `buildFinalMessage()` persists an assistant message whose `ToolUseBlock` carries truncated JSON, and the acting phase then runs the **real** tool with half an argument payload — producing an `Invalid arguments`-style error instead of the provider error, and (worse) actually invoking a registered tool with corrupt input. Please gate recovery on per-call completeness (e.g. name resolves to a registered tool *and* the accumulated arguments parse as JSON, or explicitly mark "truncated" calls and feed a synthetic `ToolResultBlock.error` for them without executing).

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2611,7 +2635,40 @@ private Flux<AgentEvent> modelCallStream(
...
+                                                    >= maxToolErrorRecoveries) {
+                                                log.warn(
+                                                        "Tool-error streaming recovery budget"
+                                                                + " exhausted after {} consecutive"
+                                                                + " recoveries, propagating: {}",
+                                                        streamToolErrorRecoveries.get(),
+                                                        error.getMessage());
+                                                return Mono.error(error);
+                                            }
+                                            streamToolErrorRecoveries.incrementAndGet();
```
oss-maintainer, line 2669: Two raw `log.warn` calls print `error.getMessage()` verbatim on every recovery. Provider error messages frequently embed the request payload (and, for 401/403-shaped errors, echoed credential fragments), which is exactly what `agentscope-core` review guidance asks us not to leak into logs; `log.warn` is also on by default, so a mis-classified error stream will spam it. Suggest one shared helper that truncates (e.g. 200 chars), redacts `sk-…`-like tokens, and downgrades the happy-recovery line to `debug` while keeping the exhausted-budget line at `warn`.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2622,7 +2679,17 @@ private Flux<AgentEvent> modelCallStream(
                                 return Flux.fromIterable(events);
                             });
 
-            return Flux.concat(Flux.just(new ModelCallStartEvent(replyId)), modelEvents, endEvents);
+            return Flux.concat(Flux.just(new ModelCallStartEvent(replyId)), modelEvents, endEvents)
+                    .doOnComplete(
+                            () -> {
+                                // A model stream that completes WITHOUT recovery resets the
+                                // consecutive-recovery budget ("consecutive" semantics). A
+                                // recovered call leaves the budget untouched so repeated
```
oss-maintainer, line 2702: `doOnComplete` fires only on natural completion. If a middleware short-circuits with `RequestStopEvent`, a downstream timeout cancels the subscription, or the consumer drops the flux, the `!recoveredThisCall` reset never runs and the counter keeps stale credit for the rest of the reply — so "consecutive" is not quite what the field doc claims. Since the counter is per-`CallExecution` anyway, resetting at reasoning-loop entry (or `doFinally` with `SignalType.ON_COMPLETE`) would make the budget deterministic.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -4745,6 +4848,20 @@ public Builder maxIters(int maxIters) {
             return this;
         }
 
+        /**
+         * Sets the maximum number of consecutive recoveries from provider-level "unknown
+         * tool" streaming errors within a single reply. Each recovery feeds a "Tool not
+         * found" result back to the model so it can self-correct; once the budget is
+         * exhausted, further tool errors propagate and fail the turn.
+         *
```
oss-maintainer, line 4947: The value is only reachable through `Builder.maxToolErrorRecoveries(...)`; `assembleReactConfig()` reads `b.maxToolErrorRecoveries`, so a `ReactConfig` carrying `max_tool_error_recoveries` (a persisted agent state, a config-driven agent, or `HarnessAgent.Builder.fromAgent()`) never feeds back into it — which is why the PR notes migrated agents silently get 3. Please add the symmetric `Builder.reactConfig(ReactConfig)` (or read the value off the assembled config) and cover the round-trip in a test, otherwise `ReactConfig.maxToolErrorRecoveries()` is a decorative accessor that changes behaviour for nobody.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2763,6 +2830,41 @@ private String resolveToolCallId(ToolUseBlock tub, ReasoningContext context) {
...
+            }
+            // Check error message for tool-related keywords that suggest the provider
+            // rejected a hallucinated tool name
+            String message = error.getMessage();
+            if (message == null) {
+                return false;
+            }
+            String lowerMessage = message.toLowerCase();
+            return lowerMessage.contains("tool")
+                    || lowerMessage.contains("function")
```
oss-maintainer: Nit: `contains("invalid_tool")` is a strict subset of the `contains("tool")` branch above it, so it is dead as an alternative. `message.toLowerCase()` also allocates even when the cheaper checks already failed — acceptable on an error path, but folding it into a single compiled `Pattern` (which would also express the tightened rule suggested above) keeps the intent readable.

agentscope-core/src/main/java/io/agentscope/core/agent/config/ReactConfig.java
```diff
@@ -23,33 +23,57 @@
...
+ * consecutive recoveries from provider-level "unknown tool" streaming errors within a single
+ * reply (each recovery lets the acting phase return "Tool not found" so the model can
+ * self-correct).
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
 public record ReactConfig(
         @JsonProperty("max_iters") int maxIters,
-        @JsonProperty("stop_on_reject") boolean stopOnReject) {
+        @JsonProperty("stop_on_reject") boolean stopOnReject,
+        @JsonProperty("max_tool_error_recoveries") int maxToolErrorRecoveries) {
```
oss-maintainer, line 37: Wire compatibility looks handled (`@JsonCreator` defaults the missing key, `@JsonIgnoreProperties(ignoreUnknown = true)` covers the forward case, and the 2-arg convenience constructor preserves source compatibility). The behavioural-compatibility question is separate: recovery is on by default (`DEFAULT_MAX_TOOL_ERROR_RECOVERIES = 3`), so from this release a provider tool error no longer fails a turn out of the box. Worth calling out explicitly in the description/release notes and adding a `0`/disabled escape hatch for users who rely on fail-fast. Note also that `ReactConfig(int, boolean)` makes the new field un-settable for every existing caller — fine, but confirm that is intended for persisted states.

thinkw:
Thanks for the thorough review — traced findings like the happy-path
persistence check are exactly what we wanted validated. Status of each
finding (all fixes in the follow-up commit):

**[Warning] classification when `getStatusCode()` is null — accepted.**
Recovery now requires an explicit, non-retryable 4xx status: 401/403
(auth/permission) and 422 (schema validation) are excluded, 429/5xx stay
with transport retry, and **a missing status code never recovers** (if
fail-fast safety cannot be proven, we fail fast). The bare keyword match
is replaced by a compiled pattern requiring an explicit unknown-tool
phrase — a negation ("non-existent", "not found", "does not exist",
"unknown", "unavailable", "invalid") adjacent to `tool|function` — so an
auth error whose text happens to contain "function" no longer matches.

**[Warning] truncated tool calls — accepted, and the issue goes one
level deeper than described.** The suggested completeness gate on the
aggregated blocks can never fire: `ToolCallsAccumulator.ToolCallBuilder#build()`
already normalizes a truncated raw argument payload to an empty object
(`content = "{}"`), so the corruption is invisible at
`getAllAccumulatedToolCalls()` — recovery would proceed with a
**seemingly-valid empty argument object**, which is arguably worse than a
corrupt one. The fix therefore reads the raw stream state:
`ToolCallsAccumulator#hasIncompleteArguments()` (per-builder
`hasCompleteArguments()`: empty payload = argument-less call, otherwise
the accumulated raw JSON must parse), surfaced through
`ReasoningContext#hasIncompleteToolCallArguments()`, and the recovery
condition now requires `!context.hasIncompleteToolCallArguments()`.
Tests cover the truncated-mid-arguments shape propagating, and an
argument-less hallucinated call still recovering.

One part of the suggestion we did not take: gating on "name resolves to a
registered tool". A provider rejects a hallucinated name precisely because
it is not in the tool schema list — which the framework derives from the
agent's own toolkit — so requiring a registered name would disable the
core scenario of #3102 by construction. The JSON-completeness gate is the
safe boundary: truncated calls fail fast, argument-less hallucinated calls
recover into `Tool not found`.

**[Warning] unsanitized error text in logs — accepted.** Both call sites
now go through a `sanitizeForLog` helper (200-char truncation +
redaction of `sk-…`-style tokens); the happy-recovery line is `debug`,
the exhausted-budget line stays `warn`.

**[Warning] `ReactConfig` value not read back — accepted.** Added
`Builder.reactConfig(ReactConfig)` which initializes
maxIters/stopOnReject/maxToolErrorRecoveries from an existing config
(explicit setters afterwards take precedence), with a round-trip test.
This also gives `HarnessAgent.Builder.fromAgent()` a clean path to
inherit the cap later.

**[Info] `doOnComplete` misses cancellation — explained, javadoc
clarified, no logic change.** Cancellation of a model-call stream ends
this per-call execution anyway: the counter lives on `CallExecution`
(one per call), so there is no cross-turn leak; and within one call,
keeping the credit across a middleware re-subscription matches the
per-call budget semantics. Resetting at reasoning-loop entry (the
suggested alternative) would make the budget meaningless — it would no
longer count across iterations. The field doc now states this
explicitly.

**[Info] dead `invalid_tool` branch — accepted.** Subsumed by the
compiled pattern.

**[Info] default-on behavior change — accepted.** Validation is now
`>= 0`, and `maxToolErrorRecoveries(0)` / `max_tool_error_recoveries: 0`
disable recovery entirely for users who rely on fail-fast. We'll call
the default-on change out in the release notes.

**Codecov:** the follow-up commit adds tests for 401/422/null-status
propagation, truncated-argument propagation, cap=0, and the
`reactConfig` round-trip, which should close the reported gaps —
please re-check after CI.

The issue author's reported message ("LLM is trying to invoke a
non-exist tool: ...", non-numeric error code → status 400) still
recovers under the tightened rules, verified locally.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
thinkw: Done in the follow-up commit — see the main reply ("classification when getStatusCode() is null").
Recovery now requires an explicit, non-retryable 4xx status (401/403/422 excluded; a missing
status never recovers) plus a compiled pattern matching an explicit unknown-tool phrase
(negation adjacent to tool/function) instead of a bare substring match.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
thinkw: Done — with one correction to the suggested gate: the aggregated blocks can never reveal
truncation, because ToolCallBuilder#build() already normalizes truncated raw JSON to "{}".
The gate therefore reads the raw stream state: ToolCallsAccumulator#hasIncompleteArguments()
(surfaced as ReasoningContext#hasIncompleteToolCallArguments()); truncated calls propagate,
argument-less calls still recover. We did not add the "registered tool" check — a hallucinated
name is by definition absent from the toolkit (the provider validated against the schema list
derived from it), so that check would disable the core #3102 scenario. See the main reply.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
thinkw: Done — both call sites go through a sanitizeForLog helper (200-char truncation + sk-…-style
redaction); the happy-recovery line is now debug, the exhausted-budget line stays warn.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
thinkw: Explained in the main reply; javadoc clarified, no logic change. Cancellation ends this
per-call execution anyway (the counter lives on CallExecution), so there is no cross-turn
leak, and keeping the credit across a middleware re-subscription matches the per-call budget
semantics. Resetting at reasoning-loop entry — the suggested alternative — would break the
budget: it must count across iterations to cap "consecutive" recoveries.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
thinkw: Done — Builder.reactConfig(ReactConfig) initializes maxIters/stopOnReject/maxToolErrorRecoveries
from an existing config (explicit setters afterwards take precedence), with a round-trip test.
This also gives HarnessAgent.Builder.fromAgent() a clean path to inherit the cap later.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
thinkw: Done — folded into the compiled pattern (UNKNOWN_TOOL_NAME_ERROR); no separate branch.

agentscope-core/src/main/java/io/agentscope/core/agent/config/ReactConfig.java
thinkw: Done — validation is now >= 0, and maxToolErrorRecoveries(0) / max_tool_error_recoveries: 0
disables recovery entirely for users who rely on fail-fast; the default-on behavior change
will be called out in the release notes. The 2-arg constructor keeping the default cap is
intended: source compatibility for existing callers, and persisted states round-trip via the
@JsonCreator defaults.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2763,6 +2837,88 @@ private String resolveToolCallId(ToolUseBlock tub, ReasoningContext context) {
...
         }
 
+        /**
+         * Compiled heuristic matching provider errors that explicitly indicate an unknown
+         * tool/function name: a negation phrase ("non-existent", "not found", "does not
+         * exist", "unknown", "unavailable", "invalid") appearing next to "tool" or
+         * "function". A bare keyword match is deliberately not enough — an auth error whose
+         * text happens to contain "function" must still fail fast.
+         */
+        private static final Pattern UNKNOWN_TOOL_NAME_ERROR =
```
oss-maintainer, line 2862: **[Warning]** The tightened pattern is narrower than the messages it needs to catch, so recovery can silently stop firing for real provider output. The separator class `[\w\s-]{0,24}` does not include backticks/quotes/dots/commas, so an error such as ``The tool `my_tool` does not exist`` matches neither alternative (the reverse branch needs `tool` first, and after `tool` the next char `` ` `` is not in `\w`, `\s` or `-`). It also cannot express *both* readings of the ambiguous `non-?exist` alternative ("non-existent **tool**" vs "tool **does not exist**").

Suggested fix: allow punctuation in the gap and cover both orders explicitly, e.g. `[\w\s.,:;"'`()\[\]-]{0,24}`, and add positive regression tests built from real provider payloads — including the exact `OpenAIClient` message quoted in #3102 — so a future edit to the pattern cannot silently disable the feature. The new tests only cover the negative (propagate) side today.

agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
```diff
@@ -149,6 +149,18 @@ private boolean isPlaceholder(String name) {
...
         }
+
+        /**
+         * Whether the accumulated argument payload is complete: either no argument
+         * content was streamed (an argument-less call) or the accumulated raw JSON
+         * parses as a JSON object. Arguments truncated mid-stream fail this check —
+         * note that {@link #build()} normalizes such payloads to an empty object, so
+         * this raw-state flag is the only reliable truncation signal.
+         */
+        boolean hasCompleteArguments() {
```
oss-maintainer, line 162: **[Warning]** `raw.isEmpty()` is the only "argument-less call" signal here, so a provider that echoes a non-JSON payload for a zero-argument call (`"\n"`, `" "`, or a JSON `null` — `JsonUtils.isValidJsonObject` rejects both `""` and `"null"` because `fromJson` returns `null` or throws) is classified as truncated and blocks recovery for a call that is actually complete.

Could you `String.trim()` before the emptiness check and treat a payload that parses to JSON `null` as "no arguments"? Recovery stays fail-safe in the current form, but it will not kick in for exactly the argument-less hallucinated-name calls that issue #3102 describes.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2611,7 +2636,42 @@ private Flux<AgentEvent> modelCallStream(
...
+                            // accumulated tool call is complete (no call truncated
+                            // mid-arguments), so the stream can complete normally and the
+                            // acting phase returns "Tool not found" results for the model
+                            // to self-correct. Consecutive recoveries are capped by
+                            // maxToolErrorRecoveries (0 disables recovery) so a confused
+                            // model cannot loop forever.
+                            .onErrorResume(
+                                    error -> {
+                                        if (isRecoverableStreamError(error)
+                                                && !context.getAllAccumulatedToolCalls().isEmpty()
```
oss-maintainer: **[Info]** The gate rebuilds every accumulated call twice in the error path: `getAllAccumulatedToolCalls()` is `buildAllToolCalls()` (allocating a `ToolUseBlock` per builder, with a JSON parse inside `build()`), and `hasIncompleteToolCallArguments()` then walks the same builders again. Roughly O(calls × argument size) of temporary allocation per recovery.

Minor cleanup: flip the order (the `hasIncompleteArguments()` scan is the cheap one), or expose a single accessor that returns the accumulated state once and reuse it for both checks.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -4745,6 +4902,40 @@ public Builder maxIters(int maxIters) {
...
+        /**
+         * Sets the maximum number of consecutive recoveries from provider-level "unknown
+         * tool" streaming errors within a single reply. Each recovery feeds a "Tool not
+         * found" result back to the model so it can self-correct; once the budget is
+         * exhausted, further tool errors propagate and fail the turn.
+         *
+         * @param maxToolErrorRecoveries Maximum consecutive recoveries, must be positive
+         * @return This builder instance for method chaining
+         */
+        public Builder maxToolErrorRecoveries(int maxToolErrorRecoveries) {
```
oss-maintainer, line 4952: **[Info]** Two small follow-ups on this setter now that the contract changed: the javadoc `@param` above still says *"must be positive"* (0 is valid, negatives are not), and the setter performs no validation — a negative value only blows up later in `assembleReactConfig()` during `build()`, with the record's message rather than the caller's context. Validating here (`if (maxToolErrorRecoveries < 0) throw new IllegalArgumentException(...)`) gives a much clearer failure for users coming from the previous positive-only API.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -2611,7 +2636,42 @@ private Flux<AgentEvent> modelCallStream(
...
+                                                log.warn(
+                                                        "Tool-error streaming recovery budget"
+                                                                + " exhausted after {} consecutive"
+                                                                + " recoveries, propagating: {}",
+                                                        streamToolErrorRecoveries.get(),
+                                                        sanitizeForLog(error.getMessage()));
+                                                return Mono.error(error);
+                                            }
+                                            streamToolErrorRecoveries.incrementAndGet();
+                                            recoveredThisCall.set(true);
```
oss-maintainer, line 2670: **[Info]** Dropping a provider error and continuing the turn is now completely invisible above `debug`: neither the log line nor any hook/event carries it, so operators cannot tell why a turn burned an extra reasoning iteration, and middlewares/metrics cannot count recoveries. Consider a `warn`-level line with the counter and cap (`"recovered streaming tool error {n}/{max}"`, sanitized), or a notification counter, so the swallowed error stays diagnosable in production where `debug` is off.

oss-maintainer (commented):
### Summary

Re-review after `91873d89` (new commits since the previous review). All six points from the first round are addressed, and addressed in the right places:

- **Classification** now requires `ModelHttpException` + an *explicit* non-retryable 4xx (401/403/422 excluded, missing status never recovers) + a compiled unknown-tool phrase — auth and schema errors can no longer be swallowed by a bare `tool`/`function` keyword hit.
- **Truncation gate** reads raw accumulator state (`ToolCallsAccumulator#hasIncompleteArguments`) instead of the aggregated blocks, which is indeed the only signal that survives `build()` normalizing a broken payload to `{}`.
- **Log hygiene** (`sanitizeForLog`: 200-char cap + `sk-…` redaction), **`maxToolErrorRecoveries(0)` = disabled**, and **`Builder.reactConfig(...)`** round-trip all look correct, with targeted tests for the 401/422/missing-status/truncated-call paths and CI green (`build` ubuntu + windows, license, codecov) on an already-signed CLA.

The remaining items are non-blocking: two precision gaps that can make recovery silently under-fire (pattern coverage, argument-less payloads) and three maintainability notes. Nice fix on the original bug — routing the recovery through the existing `ToolExecutor` "Tool not found" path instead of duplicating it keeps the assistant/tool-call intent intact.

### Findings
- **[Warning]** `ReActAgent.java:2847` — unknown-tool pattern is narrower than real provider messages (backtick/quoted names, ambiguous `non-?exist` order); needs positive regression tests from real payloads.
- **[Warning]** `ToolCallsAccumulator.java:160` — whitespace / JSON-`null` raw payloads count as "truncated", blocking recovery for argument-less calls.
- **[Info]** `ReActAgent.java:2652` — accumulated tool calls are rebuilt twice per error-path check.
- **[Info]** `ReActAgent.java:4914` — stale "must be positive" javadoc and no validation in the setter.
- **[Info]** `ReActAgent.java:2665` — a swallowed provider error is invisible above `debug`.

### Test coverage note
`@730` new lines in `ReActAgentStreamErrorRecoveryTest` is a strong suite for a first contribution; only the "must not fire" side is encoded today — please add the "must fire" cases with real provider error text so the heuristic cannot rot.

---

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
thinkw: Done — separator class widened to include quoting/punctuation, and `no such` added to the
negation set. One note: the suggested class `[\w\s.,:;"'()\[\]-]` still wouldn't have matched
your own example (`The tool \`my_tool\` does not exist`) — backticks are missing from it — so
we added a backtick to the separator class as well. Added positive "must fire" regression
tests built from the exact #3102 payload, the quoted variant, and "No such function: ..."
so the pattern cannot silently rot.

agentscope-core/src/main/java/io/agentscope/core/agent/accumulator/ToolCallsAccumulator.java
thinkw: Done — hasCompleteArguments() now trims the payload and treats a literal JSON null as
"no arguments", so blank/null argument-less calls no longer block recovery. Tests cover
both shapes recovering; they are exactly the argument-less hallucinated-name shape of #3102.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
thinkw: Done — flipped the order: the cheap raw-state scan (hasIncompleteToolCallArguments()) now
runs first and short-circuits truncated-argument calls before getAllAccumulatedToolCalls()
allocates the aggregated blocks.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
thinkw: Done — the setter now validates (< 0 throws immediately with the caller's context) and the
javadoc says "must not be negative" with 0 documented as the disable switch. A test pins the
fail-fast behavior at the call site.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
thinkw: Done — with a nod to your first-round note ("downgrade to debug to avoid spam"): the recovery
line is now info with the counter ("Recovered streaming tool error {n}/{max} (consecutive)",
sanitized text), so it stays visible by default in production without the warn-level noise a
mis-classified stream would cause; the exhausted-budget line remains warn. That sits between
the two rounds' asks.

thinkw:
> ### Summary
> Re-review after `91873d89` (new commits since the previous review). All six points from the first round are addressed, and addressed in the right places:
> 
> * **Classification** now requires `ModelHttpException` + an _explicit_ non-retryable 4xx (401/403/422 excluded, missing status never recovers) + a compiled unknown-tool phrase — auth and schema errors can no longer be swallowed by a bare `tool`/`function` keyword hit.
> * **Truncation gate** reads raw accumulator state (`ToolCallsAccumulator#hasIncompleteArguments`) instead of the aggregated blocks, which is indeed the only signal that survives `build()` normalizing a broken payload to `{}`.
> * **Log hygiene** (`sanitizeForLog`: 200-char cap + `sk-…` redaction), **`maxToolErrorRecoveries(0)` = disabled**, and **`Builder.reactConfig(...)`** round-trip all look correct, with targeted tests for the 401/422/missing-status/truncated-call paths and CI green (`build` ubuntu + windows, license, codecov) on an already-signed CLA.
> 
> The remaining items are non-blocking: two precision gaps that can make recovery silently under-fire (pattern coverage, argument-less payloads) and three maintainability notes. Nice fix on the original bug — routing the recovery through the existing `ToolExecutor` "Tool not found" path instead of duplicating it keeps the assistant/tool-call intent intact.
> 
> ### Findings
> * **[Warning]** `ReActAgent.java:2847` — unknown-tool pattern is narrower than real provider messages (backtick/quoted names, ambiguous `non-?exist` order); needs positive regression tests from real payloads.
> * **[Warning]** `ToolCallsAccumulator.java:160` — whitespace / JSON-`null` raw payloads count as "truncated", blocking recovery for argument-less calls.
> * **[Info]** `ReActAgent.java:2652` — accumulated tool calls are rebuilt twice per error-path check.
> * **[Info]** `ReActAgent.java:4914` — stale "must be positive" javadoc and no validation in the setter.
> * **[Info]** `ReActAgent.java:2665` — a swallowed provider error is invisible above `debug`.
> 
> ### Test coverage note
> `@730` new lines in `ReActAgentStreamErrorRecoveryTest` is a strong suite for a first contribution; only the "must not fire" side is encoded today — please add the "must fire" cases with real provider error text so the heuristic cannot rot.
> 
> _Automated review by github-manager-bot_

All five follow-ups addressed in the new commit — see inline replies. Worth flagging: the
suggested separator class didn't cover the backticks in your own example, so we extended it
and pinned both rounds' asks (positive "must fire" + negative "must not fire") with tests.
Local CI: 2364 tests, 0 failures.

oss-maintainer (approved):
### Summary

Re-review after `49a55e57`. All five items from the previous round are fixed, and the fixes are better than what was suggested:

- **Pattern** (`UNKNOWN_TOOL_NAME_ERROR`) — separator class widened to `[\w\s.,:;"'()\[\]`-]{0,24}` (backtick included) and `no such` added to the negation set on both alternatives, so ``The tool `my_tool` does not exist``, `No such function: ...` and the exact #3102 payload now all match. Positive "must fire" regression tests pin each shape, which is what protects the heuristic from rotting later.
- **Argument-less payloads** — `hasCompleteArguments()` trims and accepts a literal `null`, so zero-arg calls (the exact #3102 shape) are no longer misread as truncated.
- **Gate ordering** — the cheap raw-state scan now runs first and short-circuits before `getAllAccumulatedToolCalls()` allocates/aggregates; the double rebuild on the error path is gone.
- **Setter contract** — `maxToolErrorRecoveries(-1)` fails fast at the call site, javadoc/`@throws` updated, behaviour pinned by a test.
- **Observability** — the recovery line is `info` with `n/max` counters and sanitized text; the exhausted-budget line stays `warn`. Reasonable middle ground between the two rounds' asks.

### Findings
- **[Info]** `ReactConfig.java:40` — `DEFAULT_MAX_TOOL_ERROR_RECOVERIES = 3` means every existing agent silently gains "swallow a provider 4xx and re-run the reasoning loop" behaviour on upgrade. The default-on choice is defensible (bounded, fail-safe at the cap, now logged at info), but since it changes turn behaviour for users who relied on fail-fast, please make sure the release note / migration section calls it out explicitly, as proposed in the thread.
- **[Info]** CI — `build (ubuntu-latest)` and `build (windows-latest)` are both **failing** on `49a55e57` ([run 34763894558](https://github.com/agentscope-ai/agentscope-java/actions/runs/34763894558)). Please rebase on latest `main` and push the failing log excerpt; if the breakage is unrelated to this PR, a maintainer re-run would make that visible in the review queue.

### Verdict
Code quality, test coverage (`+162` lines of targeted regression tests on top of the existing suite) and responsiveness to review feedback are all well above the bar for a first contribution. Approving on the code; merging should wait until the build checks are green.

---

agentscope-extensions/agentscope-extensions-aistio/src/main/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistration.java
```diff
@@ -140,11 +140,16 @@ public void close() {
...
             scheduler = null;
         }
-        if (!registered.get()) {
+        String id = registeredInstanceId.get();
+        // Best-effort unregister: a concurrent heartbeat re-register cycle may have
+        // cleared `registered` while an identity is still set (observed on loaded CI
+        // runners as a flaky HttpSelfRegistrationTest). Skipping the DELETE in that
+        // window would leak the instance on the control plane, so unregister whenever
+        // an identity exists; never-registered calls (no identity) remain a no-op.
+        if (!registered.get() && (id == null || id.isBlank())) {
```
oss-maintainer: **[Warning]** Widening the guard to "unregister whenever an identity exists" is the right call, and it does fix the window described in the comment. What it cannot fix by itself is that `close()` is not atomic against the heartbeat thread: `shutdownNow()` only *interrupts*, it does not join.

If a heartbeat was already inside `tryRegister()` when `close()` ran, that call can still return after `close()` has read `id` and after the `finally` has cleared both fields, so lines 239-252 republish `registeredInstanceId` and set `registered=true` on an instance whose scheduler is gone. The resulting entry is heartbeated by nobody and DELETEd by nobody — which is the same control-plane leak this commit sets out to remove, just moved to a different interleaving.

Closing it needs a claim on the identity rather than a read, plus a flag the re-register path respects:

```java
private final AtomicBoolean closed = new AtomicBoolean(false);

public void close() {
    closed.set(true);
    if (scheduler != null) {
        scheduler.shutdownNow();
        scheduler = null;
    }
    String id = registeredInstanceId.getAndSet(null);   // claim: concurrent close() gets null
    registered.set(false);
    if (id == null || id.isBlank()) {
        return;
    }
    try {
        request("DELETE", "/api/v1/dataplanes/" + id, Map.of("generation", generation.get()));
        LOG.info(() -> "aistio: unregistered instance " + instanceKey);
    } catch (Exception e) {
        LOG.log(Level.INFO, "aistio: unregister failed for " + instanceKey, e);
    }
}

private void tryRegister() {
    if (closed.get()) {
        return;   // never resurrect a registration on a closed instance
    }
    ...
}
```

Two side benefits: the DELETE becomes exactly-once instead of once-per-calling-thread, and `instanceKey` is no longer logged as unregistered when `id` was in fact blank. Note also that a failed DELETE is currently logged at `Level.FINE`, so with `registered` no longer gating the path, a dropped unregistration is invisible at default log levels.

agentscope-extensions/agentscope-extensions-aistio/src/main/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistration.java
```diff
@@ -140,11 +140,16 @@ public void close() {
...
             return;
         }
         try {
-            String id = registeredInstanceId.get();
             if (id != null && !id.isBlank()) {
```
oss-maintainer: **[Info]** `close()` resets `registered` and `registeredInstanceId` but leaves `identity` and `identityListener` in place, so after a close `identity()` still returns a `RegisteredIdentity` for an instance the control plane has forgotten, and any listener can still be handed it. It also matters for the `start()`-after-`close()` path: `start()` is re-entrant here (it only early-returns on `scheduler != null`), and `tryRegister()` picks its claim credential from `identity.get().registrationCredential()` when non-null, so a restart would replay the credential of a registration that has already been DELETEd rather than claiming a fresh one.

Clearing `identity` in `close()` (or refusing `start()` once closed, which the `closed` flag above gives you for free) keeps the lifecycle contract single-valued.

agentscope-extensions/agentscope-extensions-aistio/src/test/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistrationTest.java
```diff
@@ -93,7 +93,9 @@ void registrationUsesTenantAndDurableIdentityForHeartbeatAndDelete() throws Exce
                         List.of("sessions"),
                         20)) {
             registration.start();
-            assertTrue(heartbeat.await(Duration.ofSeconds(2).toMillis(), TimeUnit.MILLISECONDS));
+            // 10s latches: await returns immediately when the callback fires; the
+            // headroom only absorbs slow/loaded CI runners, not the happy path
+            assertTrue(heartbeat.await(Duration.ofSeconds(10).toMillis(), TimeUnit.MILLISECONDS));
```
oss-maintainer, line 106: **[Info]** The 2s → 10s headroom is acceptable — `CountDownLatch.await` returns as soon as the countdown happens, so this only widens the tolerance for a loaded runner and cannot mask a real failure. But the commit's actual behaviour change is not pinned by a test: nothing here drives the case the new branch exists for, i.e. `registered == false` while `registeredInstanceId` is still set.

A direct assertion would be cheaper and more durable than the timing window — construct the instance, force `registered` down the same way `heartbeatSafe()` does on a 404 (or fail one heartbeat), then `close()` and assert the DELETE still arrives. That is also what would have caught the `closed`-flag interleaving above.

oss-maintainer (commented):
### Summary
Re-review of the commit pushed after my approve on `49a55e57`: `HttpSelfRegistration.close()` now unregisters whenever a durable identity exists rather than only while the `registered` flag is still set, plus wider latch timeouts in the aistio test. The direction is correct — skipping the DELETE while an identity is live leaks the instance on the control plane, and clearing `registered` mid re-register is exactly how that happens on a loaded runner. The two new findings are about closing the rest of the window (the heartbeat thread is not joined) and about the lifecycle state left behind after `close()`.

### Findings
- **[Warning]** `HttpSelfRegistration.java:149` — `shutdownNow()` interrupts but does not join, so a heartbeat already inside `tryRegister()` can republish `registeredInstanceId` / `registered=true` after `close()` has cleared them; claim the identity with `getAndSet(null)` and gate `tryRegister()` on a `closed` flag.
- **[Info]** `HttpSelfRegistration.java:153` — `identity` and `identityListener` survive `close()`, so `identity()` keeps advertising a deregistered instance and a `start()`-after-`close()` replays a credential that has already been DELETEd.
- **[Info]** `HttpSelfRegistrationTest.java:98` — the 10s headroom is harmless, but the branch this commit adds (`registered == false` with an identity still set) is not pinned by any assertion.

### Verdict
Not a regression — this commit is strictly better than the guard it replaces, and everything here is non-blocking polish on a path that is already best-effort. I have left my earlier approval in place rather than re-approving the new head, since the interleaving above is worth one look before this is considered finished. CI is green on `22bd6ac` on both ubuntu and windows, `license/cla` is signed, and there are no unresolved review threads from the earlier rounds.

---

agentscope-extensions/agentscope-extensions-aistio/src/main/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistration.java
```diff
@@ -231,6 +263,25 @@ private void tryRegister() {
...
+                        LOG.fine(() -> "aistio: discarded registration issued after close");
+                    } catch (Exception deleteFailure) {
+                        LOG.log(
+                                Level.WARNING,
+                                "aistio: failed to delete registration issued after close",
+                                deleteFailure);
+                    }
+                    return;
+                }
                 registeredInstanceId.set(durableId);
```
oss-maintainer: **[Warning]** The `closed` double-check at line 270 and the publication at line 285 are not one atomic step, so `close()` can still land between them and the registration is resurrected after `close()` has already returned.

Interleaving:
1. heartbeat thread enters `tryRegister()`, passes the `closed.get()` guard, the POST is in flight;
2. `close()` runs: sets `closed`, `shutdownNow()`, `registeredInstanceId.getAndSet(null)` returns **`null`** (nothing published yet) so it early-returns — no DELETE, `identity` untouched;
3. heartbeat thread resumes at line 285: `registeredInstanceId.set(durableId)`, `identity.set(...)`, `registered.set(true)`.

Result is exactly the state the new guard is documented to make impossible — an instance live on the control plane, heartbeated by nobody and deleted by nobody — plus `identity()` still reports it after `close()`, which is the invariant `startAfterCloseIsRejectedAndIdentityIsCleared` asserts on the non-concurrent path only. The window is much smaller than the one the previous commit closed, but it is the same failure mode.

Non-blocking fix that also removes the need to join the thread: put the claim and the publication under one monitor, e.g. `private final Object lifecycleLock = new Object();` held by `close()` around `getAndSet` + DELETE, and by `tryRegister()` around the `closed.get()` check through `registered.set(true)`. The DELETE would then have to be issued off-lock (or accept holding the lock across one short request), which is why I am flagging rather than insisting on the exact shape — the requirement is just that "read `closed` + publish" and "set `closed` + claim" cannot interleave. Retract-after-publish (`if (closed.get()) { if (registeredInstanceId.getAndSet(null) != null) { delete; } identity.set(null); registered.set(false); return; }` immediately after line 298) is an alternative that keeps it lock-free.

agentscope-extensions/agentscope-extensions-aistio/src/main/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistration.java
```diff
@@ -106,6 +114,13 @@ public HttpSelfRegistration(
     }
 
     public void start() {
+        if (closed.get()) {
```
oss-maintainer, line 131: **[Info]** Two small things now that `start()` has a hard lifecycle contract:

1. The rejection is a behaviour change for a public extension type — a `start()`-after-`close()` used to be silently accepted (and broken). `SessionBridge` always allocates a fresh instance per `start()`, so nothing in-repo is affected, but the class javadoc has no `@throws IllegalStateException` for `start()`; worth one line so callers outside this repo get the contract from the docs, not from a stack trace.
2. `start()`'s `closed.get()` / `scheduler != null` / `scheduler = ...` sequence is check-then-act on a **non-volatile** field. The only in-repo caller is `synchronized` on `SessionBridge`, so it is fine today, but the class itself publishes no such requirement — two concurrent `start()` calls leak a scheduler (and a duplicate registration), and a `close()` on another thread may not observe the write at all and skip `shutdownNow()`. Making `start()`/`close()` `synchronized` would fix this together with the race above; otherwise make `scheduler` `volatile` and document that callers must serialize lifecycle calls.

agentscope-extensions/agentscope-extensions-aistio/src/test/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistrationTest.java
```diff
@@ -117,6 +124,152 @@ private static JsonNode readJson(HttpExchange exchange) throws IOException {
...
+    /**
+     * Locks the close()-during-re-register-window branch: after a heartbeat failure
+     * clears {@code registered} (same as the 404 / exception paths inside
+     * heartbeatSafe), {@code close()} must still send the control-plane DELETE
+     * instead of skipping it and leaking the instance. The window is forced
+     * deterministically via reflection so the test does not depend on runner
+     * timing (which is what made it flaky on loaded CI runners).
+     */
+    @Test
+    void closeStillDeletesWhenHeartbeatCycleClearedRegisteredFlag() throws Exception {
```
oss-maintainer, line 141: **[Info]** Nice move driving the re-register window deterministically through reflection instead of relying on runner timing — that is what the flakiness needed.

The branch that has no coverage is the one this commit actually adds: `closed` observed *after* the registration round trip (line 270 in `tryRegister()`). It is straightforward to pin here, and it would have caught the remaining gap in my other comment: make the `/api/v1/agent-registrations` handler block on a latch that `close()` is allowed to satisfy (i.e. answer 201 only after `close()` has fully returned), then assert that the freshly issued `durableId` still receives a `DELETE` and that `identity()` is `null` afterwards.

Nit on the existing case: it asserts the `generation` in the DELETE body but not that a *second* DELETE was not issued — `try (HttpSelfRegistration ...)` closes again on leaving the block, so an exactly-once assertion (`assertEquals(1, deleteCount.get())`) would lock down the claim semantics the commit message is about.

oss-maintainer (commented):
### Summary

Re-review of the single commit pushed after my round-2 comment (`22bd6ac6..ebe75db6`). It takes the `close()` atomicity work seriously and closes most of the gap: a `closed` flag set before anything else, `registeredInstanceId.getAndSet(null)` so exactly one caller claims the DELETE, `tryRegister()` refusing to resurrect a closed instance both on entry and after the control-plane round trip, `identity` cleared in the `finally`, deregistration failures promoted from `FINE` to `WARNING`, `start()` after `close()` refused, and two deterministic tests that drive the re-register window through reflection instead of relying on runner timing. All three items from the previous round are genuinely addressed.

### Findings

- **[Warning]** `HttpSelfRegistration.java:285` — the `closed` check (line 270) and the publication (line 285) are still two separate steps, so `close()` can claim `null` in between and a heartbeat that finishes the round trip afterwards republishes `registeredInstanceId` / `identity` / `registered=true`. That is the same "live on the control plane, heartbeated by nobody, deleted by nobody" state the commit exists to prevent, just through a much narrower window. Suggested shape: one lifecycle monitor around "read `closed` → publish" and "`closed.set(true)` → claim", or a lock-free retract immediately after line 298.
- **[Info]** `HttpSelfRegistration.java:117` — `start()` now has a hard lifecycle contract but no `@throws` in the javadoc, and its `closed` / `scheduler != null` / `scheduler = ...` sequence is check-then-act on a non-volatile field. Safe for the only in-repo caller (`SessionBridge`, `synchronized`), undeclared for external ones; making `start()`/`close()` `synchronized` fixes this together with the finding above.
- **[Info]** `HttpSelfRegistrationTest.java:136` — the branch this commit actually adds (`closed` observed *after* the registration round trip) has no test; blocking the registration handler until `close()` has returned pins it, and would also have surfaced the race above. Worth adding `assertEquals(1, deleteCount.get())` in the existing case to pin exactly-once deregistration, since try-with-resources closes a second time.

### Verdict

**COMMENT, nothing blocking.** None of the above is a regression — `ebe75db6` is strictly better than the guard it replaces and every earlier item is fixed. CI is now fully green on `ebe75db6` (build ubuntu + windows, license, module sync, codecov patch), `license/cla` is signed, and `mergeable_state` is `clean` (it was `unstable` on the previous head only because the windows job was still running). The remaining window is a one-line-of-defence gap in code that is already best-effort with a `WARNING` log, so the merge call is maintainers'.

---

agentscope-extensions/agentscope-extensions-aistio/src/main/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistration.java
thinkw: Done — implemented the claim-and-flag design: a `closed` AtomicBoolean is set first in
close(); the identity is claimed via registeredInstanceId.getAndSet(null) so exactly one
closer DELETEs; tryRegister() refuses to run once closed, and the registration success path
double-checks `closed` after the round trip — a registration issued post-close is deleted
immediately instead of being published. The failed-DELETE log was also raised from FINE to
WARNING (a lost deregistration must be visible at default log levels). Two deterministic
tests lock the re-register-window branch and the lifecycle contract.

agentscope-extensions/agentscope-extensions-aistio/src/main/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistration.java
thinkw: Done — close() now clears the cached identity, and start() is refused once closed (the
closed flag gives that for free), so identity() can no longer report a forgotten instance
and a restart cannot replay a deleted registration credential. Pinned by
startAfterCloseIsRejectedAndIdentityIsCleared.

agentscope-extensions/agentscope-extensions-aistio/src/test/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistrationTest.java
thinkw: Done — closeStillDeletesWhenHeartbeatCycleClearedRegisteredFlag forces the exact window
state (`registered == false` while the identity is still set, via reflection to keep it
deterministic — the same state heartbeatSafe() leaves on a 404/exception) and asserts the
DELETE still arrives with the right generation.
startAfterCloseIsRejectedAndIdentityIsCleared pins the lifecycle contract.

agentscope-extensions/agentscope-extensions-aistio/src/main/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistration.java
thinkw: Done — went with the lifecycle-monitor option: a `lifecycleLock` now serializes the publish
path ("read `closed` → publish") against the claim path ("`closed.set(true)` →
`registeredInstanceId.getAndSet(null)`"), with `closed` re-checked inside the lock. Whichever
side takes the lock second deletes the freshly created or just-claimed instance, so the
interleaving you described (claim returns null → the heartbeat republishes after close) is
closed: a post-close registration self-deletes, exactly once. `start()`/`close()` are also
`synchronized` now, which fixes the check-then-act on the non-volatile `scheduler` field you
flagged in the Info below.

agentscope-extensions/agentscope-extensions-aistio/src/main/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistration.java
thinkw: Done — close() clears the cached identity in all exit paths (the finally and the
discarded-registration branch), so identity() can no longer report an instance the control
plane has forgotten, and a restart cannot replay a deleted registration credential because
start() after close() now throws IllegalStateException (javadoc @throws added). Pinned by
startAfterCloseIsRejectedAndIdentityIsCleared.

agentscope-extensions/agentscope-extensions-aistio/src/test/java/io/agentscope/extensions/aistio/transport/HttpSelfRegistrationTest.java
thinkw: Done — registrationCompletedAfterCloseIsDeletedInsteadOfPublished drives the branch
deterministically: the second registration response is held on a server-side gate, `closed`
is set via reflection only after the POST is in flight (avoiding the shutdownNow interrupt
that would cancel the in-flight request), and the released registration must arrive as a
DELETE with registeredInstanceId/identity never published. The three pre-existing tests also
assert deleteCount == 1 (exactly-once, including the double close from try-with-resources).

oss-maintainer (approved):
### Summary

Re-review of `ebe75db6..ff5b86a6`. Both items from the previous round are resolved, and the race is closed in the way it needed to be closed rather than papered over: `close()` sets `closed=true` before it claims `registeredInstanceId` via `getAndSet(null)`, and `tryRegister()` re-checks `closed` **inside the same lifecycle lock** before publishing — so whichever side loses the race deterministically deletes the surviving registration instead of leaving an instance that is live on the control plane, heartbeated by nobody and deleted by nobody. `identity` is now cleared on every `close()` exit path and `start()` after `close()` is refused and documented.

The two new tests are the part I most wanted to see: `closeStillDeletesWhenHeartbeatCycleClearedRegisteredFlag` and `registrationCompletedAfterCloseIsDeletedInsteadOfPublished` drive the exact window states through the latch/reflection instead of hoping a runner timing works out, which is the only way a fix like this stays fixed.

### Findings
No blocking or non-blocking findings on this increment.

### Notes (non-blocking, take or leave)
- The early `closed.get()` in `tryRegister()` is intentionally outside the lock to avoid a wasted HTTP round trip; correctness comes from the in-lock re-check, which is fine — worth a one-line comment saying exactly that so a later refactor doesn't "simplify" the second check away.
- The DELETE is issued while holding the lifecycle lock, so a concurrent `close()` can be delayed by one client timeout. Bounded and acceptable for a registration component; only reconsider if aistio ever lowers that timeout aggressively.

Verdict: re-approved on the current head. Merge decision stays with the maintainers.

---

PR #3134, opened 2026-09-14
fix(harness): cancel leaked memory background model calls on agent shutdown

cheny813:
## Problem

`HarnessAgent.close()` cancelled memory work by passing `this` as the task owner, but `MemoryFlushManager` and `MemoryConsolidator` register their work with their own owner tokens. The owner never matched, so flush and consolidation kept running against a disposed agent. When a background task is blocked on a streaming model call, the underlying connection is never closed and the task never completes.

## Root cause

```java
// HarnessAgent.close() - the owner passed here is never the one used at registration time
MemoryBackgroundTasks.cancelAll(this);
MemoryBackgroundTasks.shutdown(this);
```

The callers of `MemoryBackgroundTasks.register(...)` sit inside the memory managers, which use their own tokens. Cancellation therefore silently no-ops and in-flight work is never disposed.

## Fix

Give each `HarnessAgent` a private owner token, thread it into `MemoryFlushMiddleware` and `MemoryMaintenanceMiddleware`, and cancel with that token in `close()`. `MemoryBackgroundTasks` gains owner-aware APIs (`register(owner, d)`, `unregister(d)`, `cancelAll(owner)`, `shutdown(owner)`) next to the existing global `cancelAll()` / `awaitQuiescence()`. `register` re-checks the shutdown marker after publishing a task and disposes it if the marker was set in between, so work dispatched concurrently with `close()` is dropped instead of starting a model call.

When the owner is already shut down, `MemoryFlushMiddleware.runFlush` drops the flush, releases its in-flight slot, and drains the rest of the queue. That drain is a loop, not a call into `drainFlushQueue`: every queued flush takes the same branch while the owner is shut down, so draining by running the next task would recurse once per queued flush and could overflow the stack inside `close()`.

A note on timeouts: cancellation now propagates from the agent-level `cancelAll(owner)`/`shutdown(owner)` through `MemoryBackgroundTasks.register`. No new timeout behaviour is introduced by this change.

## Tests

`HarnessAgentMemoryCancelTest` drives a real `HarnessAgent` built through `HarnessAgent.builder()` (never the middlewares directly, which is what makes the owner mismatch visible), with four cases:

- `closeCancelsThisAgentsHungMemoryFlush`: the memory model never emits, standing in for a stuck SSE connection; asserts the task is still in flight before close, that `close()` cancels it, and that `awaitQuiescence()` returns.
- `closeLeavesOtherOwnersTasksUntouched`: a second owner's task is registered before close; asserts the other owner's task is still running while the agent's own task is cancelled.
- `closeOfOneAgentLeavesAnotherLiveAgentsTaskRunning`: two real agents run concurrently, each with a hung memory model; closing the first cancels only its own task, the second agent's task stays in flight, and it is cancelled only when that agent is closed. The two agents use distinct user ids because `MemoryFlushMiddleware` serialises flushes per isolation key and the default `IsolationScope.USER` keys on user id alone.
- `closeReleasesEveryQueuedFlushWhenTheQueueIsDeep`: the first flush hangs and 400 more queue behind it under one user id; asserts that `close()` still brings the registry to quiescence, i.e. the queue is drained to the bottom and every slot is released.

`MemoryBackgroundTasksCancelTest` covers owner-aware cancel/shutdown/awaitQuiescence semantics.

## Compatibility

The legacy `register(d)` keeps the previous global behaviour (`taskOwner == null`), so existing callers and constructors are unaffected. `begin()` / `end()` semantics for tests are unchanged.


agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryBackgroundTasks.java
```diff
@@ -58,6 +97,99 @@ public static void end() {
...
+     */
+    public static void register(Object owner, Disposable d) {
+        if (d == null || d.isDisposed()) {
+            return;
+        }
+        if (owner != null && isShutdown(owner)) {
+            d.dispose();
+            return;
+        }
+        IN_FLIGHT_TASKS.put(d, owner != null ? owner : UNOWNED);
```
oss-maintainer: `register(owner, d)` is not atomic with `shutdown(owner)` + `cancelAll(owner)`: the `isShutdown` guard is evaluated at line 123 and the entry is only published here. If `HarnessAgent.close()` runs `shutdown(this)`/`cancelAll(this)` between the check and this `put`, the task is registered against an owner that is already shut down and nothing will ever dispose it — which is exactly the leak this PR fixes (the model call and its HTTP connection outlive the closed agent).

Suggested fix: re-check after publishing and undo, e.g.

```java
IN_FLIGHT_TASKS.put(d, owner != null ? owner : UNOWNED);
if (owner != null && isShutdown(owner) && IN_FLIGHT_TASKS.remove(d) != null) {
    d.dispose();
    return;
}
```

Note `cancelAll(Object)` snapshots with `List.copyOf(...)`, so making `cancelAll(owner)` re-scan once after the snapshot does not close the window by itself — the double-check inside `register` does.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryConsolidator.java
```diff
@@ -190,6 +191,7 @@ public Mono<Void> consolidate(RuntimeContext rc) {
                         .build());
 
         return model.stream(messages, null, null)
+                .timeout(Duration.ofMinutes(2))
```
oss-maintainer: This hardcodes a 2-minute inter-signal timeout on the consolidation model stream (the same line is added in `MemoryFlushManager.flushMemories`). Two things worth confirming:

1. `main` has no `.timeout(` in either file, so this is a **new** behaviour change, not a pre-existing guard — the PR description's "the memory managers already apply `Duration.ofMinutes(2)`" does not match what the diff introduces. On a large ledger a legitimate consolidation can exceed 2 min, and both call sites swallow the failure into a `log.warn` (`onErrorResume` in `runMaintenance`, the `subscribe(null, ...)` in `runFlush`), so memory silently stops being maintained with no metric and no retry.
2. `Flux#timeout(Duration)` resets on every `onNext`, so it bounds the gap between chunks, not the call — which is fine for the leak, but it does not make a slow-drip response abortable, contrary to the motivation stated in the description.

Please expose the value through `MemoryConfig` (with the current 2 min as default) and mention the new limit in the PR description, so operators can raise it instead of losing flushes.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryBackgroundTasks.java
```diff
@@ -37,6 +55,27 @@ public final class MemoryBackgroundTasks {
...
+     */
+    private static final ConcurrentHashMap<Disposable, Object> IN_FLIGHT_TASKS =
+            new ConcurrentHashMap<>();
+
+    /**
+     * Owners (agent instances) that have been shut down. Weak references so a closed agent can
+     * be garbage-collected once nothing else references it.
+     */
+    private static final Set<Object> SHUTDOWN_OWNERS =
+            Collections.synchronizedSet(Collections.newSetFromMap(new WeakHashMap<>()));
```
oss-maintainer, line 77: `SHUTDOWN_OWNERS` is a process-wide weak set keyed by the agent instance, while `MemoryFlushMiddleware`/`MemoryMaintenanceMiddleware` keep a strong `taskOwner` field. That works, but it means the shutdown flag can only ever be observed through a static registry, and one entry is retained per closed agent until the middleware referencing it is collected too.

An alternative that keeps the same semantics without the weak-set indirection: let `HarnessAgent` own a `volatile boolean memoryShutdown` (or an `AtomicBoolean` passed into both middlewares at build time) and have `register`/dispatch consult that flag. The weak set also relies on `HarnessAgent` keeping identity `equals`/`hashCode` — if that is ever overridden for value comparison, `cancelAll(owner)` and `isShutdown(owner)` start matching the wrong agents silently.

agentscope-harness/src/test/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddlewareCancelTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+            assertTrue(
+                    flushStarted.await(5, TimeUnit.SECONDS),
+                    "flush model.stream must be invoked (the real leaked path)");
+            assertFalse(
+                    MemoryBackgroundTasks.awaitQuiescence(1, TimeUnit.SECONDS),
+                    "without cancel, the hung flush must keep in-flight > 0 (proves the leak"
+                            + " existed)");
+
+            MemoryBackgroundTasks.cancelAll();
```
oss-maintainer, line 94: Both new middleware tests release the hung call through the **global** `MemoryBackgroundTasks.cancelAll()`, while the production path is `HarnessAgent.close()` → `shutdown(this)` + `cancelAll(this)`. `MemoryBackgroundTasksCancelTest` does cover the owner-scoped registry, so the gap is specifically the wiring: nothing asserts that closing one `HarnessAgent` cancels only its own in-flight flush/consolidation, leaves a second live agent's task running, and that a flush dispatched after `close()` drops out (releases the in-flight slot and drains the queue) instead of starting a model call.

Could you add that end-to-end case over the `HarnessAgent` lifecycle? It is the part a future refactor is most likely to break, and it would also cover the `register`/`shutdown` interleaving above.

oss-maintainer (commented):
### Summary
Nice catch on the leak, and the design is the right shape: agent-scoped `Disposable` tracking with `shutdown(owner)` + `cancelAll(owner)`, and dropping the blocking `consolidate(...).block()` so the registered disposable actually cancels the model call. The legacy 7-arg constructors keep delegating with a `null` owner, so this stays source/binary compatible for existing users.

Four things I would like addressed before this is ready:

1. **[Warning]** `MemoryBackgroundTasks.register` has a check-then-put window against `shutdown(owner)` — a task dispatched concurrently with `close()` can be registered after the cancel pass and never disposed, which reproduces the exact leak being fixed.
2. **[Warning]** the new hardcoded `.timeout(Duration.ofMinutes(2))` on the consolidation/flush model streams is a behaviour change (there is no such timeout on `main`), it is not configurable, and both call sites swallow the error into a `log.warn`, so memory maintenance can silently stop. The PR description also reads as if the timeout already existed.
3. **[Info]** the process-wide weak `SHUTDOWN_OWNERS` set relies on identity `equals`/`hashCode`; an owner-owned flag would be easier to reason about.
4. **[Warning]** the two new middleware tests only exercise the global `cancelAll()`; nothing covers `HarnessAgent.close()` -> `cancelAll(owner)` end to end, i.e. that one agent closing does not disturb another live agent's task.

Note on CI: `build (ubuntu-latest)` / `build (windows-latest)` / `Check License` / `codecov` are all green on head `75318c9f`, and the `@Test` annotations in the three new test classes are intact, so the tests do run — my comments above are about coverage of the production path, not about compilation.

CLA: no `license/cla` status context was found on this head, so I could not confirm the signature state; if your CLA bot has not run yet, please make sure it is signed before merge. I am only commenting here — merging stays with the maintainers.

---
*Automated review by github-manager-bot*"

cheny813:
Thanks for the review — it was against `75318c9f`; the branch has since been force-pushed and the fix commits squashed, so the head is now `6abb812a`. Point by point:

**1. check-then-put window in `register` (Warning) — fixed.** `register(owner, d)` now re-checks the shutdown marker *after* publishing into `IN_FLIGHT_TASKS`, and disposes the task if the marker was set in between:

```java
IN_FLIGHT_TASKS.put(d, effectiveOwner);
if (isShutdown(effectiveOwner) && IN_FLIGHT_TASKS.remove(d) != null) {
    disposeQuietly(d);
}
```

Together with the pre-put check this closes the window: a shutdown marked before the put, or between the put and the re-check, both end up disposed. A shutdown marked after the re-check is still seen by the subsequent `cancelAll(owner)` pass, which removes each entry before disposing it. `HarnessAgent.close()` also calls `shutdown(owner)` *before* `cancelAll(owner)`, so tasks dispatched concurrently with close are dropped instead of starting a model call.

**2. Hardcoded `Duration.ofMinutes(2)` (Warning) — resolved by removing the timeouts.** They are gone from `MemoryConsolidator`, `MemoryFlushManager`, `MemoryFlushMiddleware` and `MemoryMaintenanceMiddleware`; there is no `Duration.ofMinutes(2)` left anywhere in the memory pipeline, so no behaviour is changed relative to `main` and nothing is swallowed into `log.warn`. Cancellation now propagates from the agent-scoped `cancelAll(owner)` / `shutdown(owner)` through `MemoryBackgroundTasks.register`. The PR description has been corrected to say this, since it previously read as if the timeout already existed.

**3. Identity semantics of `SHUTDOWN_OWNERS` (Info) — kept by design, documented.** The owner is a private sentinel object created once per agent inside `Builder.build()` and shared with that agent's memory middlewares; `HarnessAgent` does not override `equals`/`hashCode`, so identity is the intended and only meaningful semantics here. The weak set keeps a closed agent collectable (the alternative — an owned flag on the agent — would widen the public surface for the same behaviour). The class javadoc now states the owner contract and the shutdown-marker behaviour explicitly.

**4. `close()` -> `cancelAll(owner)` end-to-end (Warning) — covered by `HarnessAgentMemoryCancelTest`.** Both cases drive a real `HarnessAgent` built through `HarnessAgent.builder()`, never the middlewares directly, which is what makes the owner/token mismatch visible:

- `closeCancelsThisAgentsHungMemoryFlush`: the memory model never emits (stands in for the stuck SSE connection), asserts the task is still in flight before close, and that `close()` cancels it and `awaitQuiescence()` returns.
- `closeLeavesOtherOwnersTasksUntouched`: registers a second owner's task, closes the agent, and asserts the other owner's task is still running while the agent's own task is cancelled.

On "another *live agent*": owners are per-agent tokens minted in the builder and never exposed, so a foreign owner token is what another agent's registrations look like to the registry — the scoping assertion is the same code path. If you would still prefer a second real `HarnessAgent` on the other side of that assertion, say so and I will add it as a third case.

Other status: `license/cla` is now green on this head, and `build (ubuntu-latest)`, `build (windows-latest)`, `Check License`, `Check Module Sync` and `codecov/patch` all pass on `6abb812a`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
```diff
@@ -201,14 +231,26 @@ private void scheduleFlush(Agent agent, RuntimeContext rc) {
...
-                            MemoryBackgroundTasks.end();
-                            drainFlushQueue(key);
-                        })
-                .subscribe(null, e -> log.warn("Memory flush failed: {}", e.getMessage()));
+        if (MemoryBackgroundTasks.isShutdown(taskOwner)) {
+            // The owning agent is already closed: drop this flush, release its in-flight slot,
+            // and keep draining so every remaining queued task releases its slot too, instead
+            // of starting a model call that would outlive the closed agent.
+            MemoryBackgroundTasks.end();
+            drainFlushQueue(key);
```
oss-maintainer: The shutdown early-return drains synchronously, so the queue walks itself as `drainFlushQueue → next.run() → runFlush → drainFlushQueue` on the closing thread — one stack frame per queued task. In the normal path the recursion is broken by `subscribeOn(boundedElastic)`; here it is not, so a deep `FLUSH_QUEUES` deque for this key (many flushes queued while one model call was in flight, then `close()`) recurses until the stack gives up. An iterative drain (loop while the queue yields a task, invoking each outside the recursion) would keep the same slot-releasing semantics without the latent `StackOverflowError`.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryBackgroundTasks.java
```diff
@@ -58,6 +97,107 @@ public static void end() {
...
+     * Registers an in-flight background task owned by {@code owner} so
+     * {@link #cancelAll(Object)} can cancel it when that agent shuts down. The task's
+     * {@link Disposable} must be passed here instead of being discarded, otherwise its
+     * underlying model call (and the HTTP connection behind it) can outlive the agent.
+     *
+     * <p>If {@code owner} has already been {@link #shutdown(Object)}-marked, the disposable is
+     * disposed immediately: the task was dispatched concurrently with shutdown and must not
+     * start real work.
+     */
+    public static void register(Object owner, Disposable d) {
```
oss-maintainer, line 119: The early-return disposal here calls `d.dispose()` directly, while every other disposal site in this class goes through `disposeQuietly` (line 125 vs. the re-check at line 134). Reactor's `dispose()` rarely throws, but if a custom `Disposable` does throw, this propagates out of `register(...)` into the middleware/flush-drain path and aborts the rest of the cleanup loop. Wrapping it in `disposeQuietly(d)` keeps the two paths consistent.

oss-maintainer (commented):
### Summary
Re-reviewed at `6abb812a`. This closes every gap from the previous round: the check-then-put window is now covered by the re-check after publishing, the unconfigurable 2-minute `timeout(...)` is gone, `SHUTDOWN_OWNERS` + `shutdown(owner)` gate registrations that race `close()`, and there is an end-to-end test that one agent's `close()` leaves another live agent's task alone. The per-owner scoping (`register(owner, d)` / `cancelAll(owner)` / `unregister(holder)` from `doFinally`) is the right shape for the leak in #3134, and the legacy constructors still delegate with a `null` owner, so the public surface stays compatible.

Two small follow-ups are inline below; neither is a design problem, and the first only bites on a deep flush queue, so this is ready for a maintainer to weigh once they are handled (or explicitly waved off).

### Reviewed as correct
- `cancelAll(owner)` copies the entry set and removes each task *before* disposing, so a task registered concurrently with the cancel pass stays tracked instead of being silently dropped.
- `MemoryBackgroundTasks.end()` is still called on the shutdown early-return, so the in-flight slot is released and `awaitIdle` cannot hang on a closed agent.
- The `holder[0]` indirection registers the `Disposable` after `subscribe(...)` returns while `doFinally` unregister runs only on a terminal signal; the one theoretical window (a synchronous terminal signal before `register`) ends up with a disposed-and-removed entry, i.e. no leak.
- New tests cover both `shutdown(owner)` gating and `cancelAll(owner)` isolation, not just the global `cancelAll()`.

### CI
`build (ubuntu-latest)`, `build (windows-latest)`, `Check License`, `Check Module Sync` and `codecov/patch` are all green on this head. CLA is now reported signed (`license/cla` = success), which the previous review could not confirm.

---

cheny813:
Added the two-real-agent case as a follow-up (`be8a74b3`), so point 4 no longer rests on the "a foreign owner token is equivalent to another agent" argument.

`closeOfOneAgentLeavesAnotherLiveAgentsTaskRunning` builds two `HarnessAgent`s through the public builder, each with its own hung memory model, and asserts:

1. both background model calls start,
2. `close()` on the first agent cancels *its* call (observed through `doOnCancel`, not inferred from an in-flight count),
3. the second agent's call is still not cancelled afterwards,
4. it is cancelled only once the second agent is closed, after which `awaitQuiescence()` returns.

One thing worth flagging for reviewers, since it is a property of the framework rather than of this change: the two agents must use **different user ids**. `MemoryFlushMiddleware` serialises flushes per isolation key and the default `IsolationScope.USER` keys on `userId` alone, so two agents sharing a user id queue their flushes behind each other — the second flush would never start while the first one hangs. That is pre-existing behaviour, not something this PR introduces, but it is easy to trip over in a test like this one, so it is called out in a comment at the call site.

Also updated the PR description's Tests section to list all three cases. Full module run: `HarnessAgentMemoryCancelTest` 3 passed, `MemoryBackgroundTasksCancelTest` 10 passed, repeated twice with no flakes.

agentscope-harness/src/main/java/io/agentscope/harness/agent/middleware/MemoryFlushMiddleware.java
cheny813: Good catch — fixed in `282a2cf8`.

The shutdown branch now drains in a loop instead of calling `drainFlushQueue`:

```java
if (MemoryBackgroundTasks.isShutdown(taskOwner)) {
    while (true) {
        MemoryBackgroundTasks.end();
        if (takeNextQueuedFlush(key) == null) {
            return;
        }
    }
}
```

`takeNextQueuedFlush` is the dequeue half of the old `drainFlushQueue` pulled out on its own: it removes and returns the next task (or evicts the key and returns `null` when the key has gone idle) without running it. `drainFlushQueue` is now a thin wrapper over it, so the non-shutdown path is unchanged — it still runs exactly one next task, and that task's own `doFinally` continues the chain asynchronously.

Each iteration releases the slot of the task it skips, which is what the recursive version did too, just with a stack frame per queued flush.

Added `closeReleasesEveryQueuedFlushWhenTheQueueIsDeep` to `HarnessAgentMemoryCancelTest`: hang the first flush, queue 400 more behind it under one user id (distinct session ids so they are not coalesced), close the agent, assert `awaitQuiescence()` returns — 5.1s. To be precise about what that test does and does not prove: 400 frames is a functional-regression depth, not enough to overflow a default JVM stack on its own, so the test proves the queue is drained to the bottom with no slot left behind; the stack-safety argument rests on the loop shape, not on that number.

agentscope-harness/src/main/java/io/agentscope/harness/agent/memory/MemoryBackgroundTasks.java
cheny813: Fixed in `282a2cf8` — the early return now goes through `disposeQuietly(d)` like the other two paths in the file, so a throwing `Disposable` cannot propagate out of `register`.

oss-maintainer (approved):
### Summary
Re-reviewed the increment since my last pass (`6abb812a` -> `282a2cf8`). Both remaining inline points are now fixed and the production shutdown path is finally covered end to end, so this is an **approve**. CI (`ubuntu`, `windows`, Check License, Check Module Sync, codecov/patch) is green on this head and the CLA is signed.

### What the increment does
- `MemoryFlushMiddleware.runFlush` now drains a shutdown owner's queue with an iterative `while` loop over the new `takeNextQueuedFlush(key)` instead of recursing through `drainFlushQueue -> run -> runFlush`. That removes the per-queued-flush stack frame I flagged on `6abb812a`, and the `begin()`/`end()` pairing is still exact: the first `end()` releases the current task's slot, and every skipped queued task gets exactly one `end()` on the iteration that drops it. The key is evicted once the queue drains, so no stale `FlushQueue` entry is left behind.
- `MemoryBackgroundTasks.register(owner, d)`'s already-shutdown early return now goes through `disposeQuietly`, consistent with the post-publish re-check and with both `cancelAll` paths, so a throwing custom `Disposable` can no longer break the drain chain.
- `HarnessAgentMemoryCancelTest` adds the two coverage gaps from the earlier round. `closeOfOneAgentLeavesAnotherLiveAgentsTaskRunning` drives the real `HarnessAgent.close()` -> `shutdown(this)` + `cancelAll(this)` path with two distinct users and asserts that the other agent's task keeps running; using `doOnCancel` on a never-emitting `Flux` to tell "still running" apart from "cancelled" is the right way to make that assertion non-vacuous. `closeReleasesEveryQueuedFlushWhenTheQueueIsDeep` pins the counter bookkeeping at 400 queued flushes, which is the case that used to recurse.

### Note (not a blocker)
`closeReleasesEveryQueuedFlushWhenTheQueueIsDeep` costs CI wall time — 401 blocking `agent.call(...)` round trips under a 180 s timeout. It passes on both runners here, so keep it if it is the only way to reproduce the deep-queue behaviour; if it later turns slow, a smaller depth still stays above the recursion limit it guards against.

---

PR #3137, opened 2026-09-14
fix(model): decouple DashScope streaming from thinking mode

### LINKED ISSUE #3067: [Feature]: DashScopeChatModel 在开启思考时强制 stream=true，与 Python SDK 行为不一致，建议解耦
### 标题
`DashScopeChatModel` 在开启思考时强制 `stream=true`，与 Python SDK 行为不一致，建议解耦

---

**环境**
- agentscope-java: `2.0.3-SNAPSHOT` (commit `d20ebbe`)

**问题描述**
Java 版 `DashScopeChatModel` 构造时，只要 `enableThinking=true` 就无条件把 `stream` 改写为 `true`，用户显式设置的 `stream=false` 被忽略。而 Python SDK 中 `stream` 与 `thinking` 是完全独立的，两者行为不一致。

**证据**

1）Java 强制改写 `stream` —— `DashScopeChatModel.java` L154-161：
```java
// Thinking mode requires streaming; override stream setting if needed
if (enableThinking != null && enableThinking && !stream) {
    log.info("...Forcing stream=true as thinking mode requires streaming.");
}
this.stream = enableThinking != null && enableThinking ? true : stream;
```

2）Python 版无此约束 —— Python 的 `DashScopeChatModel` 是 OpenAI 兼容实现（用 `openai` SDK 打 `compatible-mode/v1`），`stream` 忠实透传、`enable_thinking` 独立设置，二者解耦。

3）Java 自己的 `OpenAIChatModel` 也无此约束 —— 同仓库 `OpenAIChatModel.java` 完全没有 thinking→强制 stream 的逻辑，`stream` 由 `GenerateOptions.getStream()` 独立决定（默认 `false`）。可见强制流式是 `DashScopeChatModel` 独有的，并非框架统一约定。

**服务端真实约束（以阿里云百炼文档为例）**
"非流式 + 思考"的约束是**按模型区分**的：
- 商业版（`qwen-plus` / `qwen3-max` / `qwen-flash` / `qwen3.7-plus`）：**支持**非流式 + 思考；
- 开源版思考模型（`qwen3-235b-a22b` / `qwen3-32b`）：非流式 + 思考才报 `400 parameter.enable_thinking only support stream call`。

即使"thinking 必须 stream"只对部分模型成立，各家也有不同的报错兜底，不应无差别套用到所有模型。

**影响**
1. 需要"非流式 + 思考"的场景（流式返回不稳定）被无差别剥夺；
2. 与 Python SDK 行为不一致，跨语言用户困惑。

**建议方案（择一）**
- **方案 A（推荐，opt-out）**：`Builder.stream` 改用 `Boolean`，区分"用户显式设置"与"未设置"。仅当未显式设置时才默认强制 `true`；用户显式 `stream(false)` 则尊重。改动小、向后兼容。
- **方案 B（模型感知）**：仅对已知"非流式+思考会报错"的开源思考模型强制流式，商业版不强制。

**疑问**
Java 版是否有计划与 Python 版对齐，将 `stream` 与 `thinking` 解耦？

CryoThrust:
## Summary

Implements option A from the evaluation: `stream` and `enableThinking` are decoupled, and the caller's explicit choice wins.

- `DashScopeChatModel` no longer rewrites the flag from `enableThinking`; an explicit `stream(false)` is passed through.
- Streaming and thinking are independent at the public API. Whether a model supports non-streaming + thinking is left to the DashScope endpoint, which reports it as a provider error.

## Tests

- `testExplicitNonStreamingHonoredWithThinking`: `stream(false)` + `enableThinking(true)` produces `"incremental_output":false` on the wire.
- `testStreamingDefaultsTrueWhenUnset`: when `stream` is not set, the request still uses `"incremental_output":true` even with thinking enabled.
- Both fail on `main` (the first asserts the exact reported behavior) and pass with the change.
- Full `agentscope-extensions-model-dashscope` suite: 482 passed, 2 skipped. `spotless:check` clean.

Closes #3067.


oss-maintainer (commented):
### Summary
Decouples `stream` from `enableThinking` in `DashScopeChatModel` so an explicit `stream(false)` is no longer silently rewritten, with the null-default resolved in `build()`. The core change looks correct and the two new wire-level assertions cover exactly the reported behavior; my comments are about the docs/javadoc that still describe the old coupling and about diagnosability, not about blocking logic.

### Findings
- **[Warning]** `modelName` is available at the point of the new warning but not included, and the level moved from `info` to `warn` for a combination that is valid for most models.
- **[Warning]** the repo still documents the previous "thinking forces streaming" contract in `DashScopeModelConfig` (scheduler) and `docs/v1/{en,zh}/docs/task/model.md`; the Spring Boot starter `stream: false` + `enable-thinking: true` combination changes behavior.
- **[Info]** the "unset means streaming" default is now expressed in both `Builder.build()` and `DashScopeModelProvider`.
- **[Info]** the new tests shadow the fixture MockWebServer and only shut it down on the success path.

### Suggestions
1. Add the model name to the warning text and mention the MultiModal/`qvq*` path where non-streaming thinking is actually rejected.
2. Update the stale javadoc/doc snippets listed above (en + zh), and call out the starter behavior change.
3. Reuse the existing MockWebServer fixture or wrap shutdown in try/finally.

CI at the reviewed commit: `build (ubuntu-latest)` passing, `build (windows-latest)` pending, `license/cla` signed. No conflicts.

---
*Automated review by github-manager-bot*"

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeChatModel.java
```diff
@@ -151,13 +151,18 @@ public DashScopeChatModel(
...
-            log.info(
-                    "Thinking mode is enabled but stream=false was specified. "
-                            + "Forcing stream=true as thinking mode requires streaming.");
+        // Streaming and thinking are independent options. DashScope only rejects the
+        // non-streaming + thinking combination for some open-source thinking models, and
+        // it reports that as a provider error, so the caller's explicit choice is kept
+        // as-is here rather than silently rewritten.
+        if (!stream && enableThinking != null && enableThinking) {
+            log.warn(
+                    "Thinking mode is enabled together with a non-streaming request. Some "
```
oss-maintainer: The message does not name the model, but `modelName` is already assigned two lines above — and the model is exactly what determines whether the endpoint rejects the combination. Worth including (e.g. `model={}`) plus a hint for the realistic case, vision/`qvq*` models that route to the MultiModal API in `streamWithHttpClient()` and only accept thinking with streaming. Also consider keeping this at `info`: models are sometimes built per call (Spring prototype beans, `ModelRegistry` cache misses), so a new WARN level here can add noise for a configuration that is valid for many models.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeChatModel.java
```diff
@@ -401,7 +406,10 @@ public boolean supportsNativeStructuredOutput() {
     public static class Builder {
         private String apiKey;
         private String modelName;
-        private boolean stream = true;
+
+        /** Null means "not set by the caller"; resolved to {@code true} in {@link #build()}. */
+        private Boolean stream;
```
oss-maintainer: Minor duplication of the "unset means streaming" decision: `DashScopeModelProvider` already resolves it explicitly (`context.getStream() != null ? context.getStream() : true`, DashScopeModelProvider.java:81-82), so the provider path never exercises the new null default. Not wrong, but if the builder is now the single place that owns the default, the provider could pass `context.getStream()` through unchanged.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeChatModel.java
```diff
@@ -443,8 +451,10 @@ public Builder modelName(String modelName) {
         /**
          * Sets whether streaming should be enabled.
          *
-         * <p>This setting is ignored if enableThinking is set to true, as thinking mode
-         * automatically enables streaming.
+         * <p>Streaming and thinking mode are independent. If this is not called, streaming
+         * defaults to {@code true}. An explicit {@code false} is honored even when thinking
```
oss-maintainer: Please align the places in this repo that still document the old coupling, otherwise the decoupling is only half-documented:
- `agentscope-extensions-scheduler-common/.../DashScopeModelConfig.java:260-261` — "This setting is ignored if enableThinking is set to true, as thinking mode automatically enables streaming" and `:274` — "this automatically enables streaming and may override the stream setting".
- `docs/v1/en/docs/task/model.md:166` and `docs/v1/zh/docs/task/model.md:166` — "Enables thinking mode and automatically enables streaming" / "并自动启用流式输出".
This also has a user-visible effect on the Spring Boot starter: `DashScopeProperties.stream` is a primitive defaulting to `true`, so an existing `agentscope.dashscope.stream: false` + `enable-thinking: true` config silently switches from forced-streaming to a non-streaming thinking request after this change. A short note in the docs (or the release notes) for that combination would prevent surprise reports.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/test/java/io/agentscope/extensions/model/dashscope/DashScopeChatModelTest.java
```diff
@@ -138,6 +138,76 @@ void testNonStreamingModelCreation() {
         assertNotNull(nonStreamingModel, "Non-streaming model should be created");
     }
 
+    @Test
+    @DisplayName("Explicit stream=false is honored when thinking mode is enabled")
+    void testExplicitNonStreamingHonoredWithThinking() throws Exception {
+        MockWebServer mockServer = new MockWebServer();
```
oss-maintainer, line 144: Both new tests declare a local `MockWebServer` that shadows the fixture field from `setUp()`, and `shutdown()` is only called on the success path — a failing `assertTrue` leaks the running server. The `@BeforeEach`/`@AfterEach` pair already manages a server for you, and `testModelWithEncryption` in this file shows the try/finally pattern; either works and keeps the new cases consistent with the rest of the class.

jujn (approved):

PR #3144, opened 2026-09-14
docs: CompactionConfig.keepTokens documented default value does not match source code

xingfudeshi:
**Problem**
The documentation states keepTokens defaults to 0, but the actual source code (CompactionConfig.Builder) sets it to -1.

**Impact**
These two values produce completely different behavior:

0: determineCutoffIndex takes the message-based branch, keepMessages works as configured.
-1: Triggers dynamic mode. keepTokens is auto-computed from contextWindow * keepTokensRatio, which overrides keepMessages entirely — the user's keepMessages setting is silently ignored.
When users configure triggerMessages + keepMessages without explicitly setting keepTokens (relying on the documented default), keepMessages has no effect. Compaction barely reduces message count and re-triggers in a loop.

**Fix**
Correct the documented default value from 0 to -1, and clarify that -1 enables dynamic mode (auto-computed from model context window). To use keepMessages verbatim, set keepTokens=0 explicitly.

image


docs/v2/zh/docs/harness/memory.md
```diff
@@ -79,9 +79,9 @@ HarnessAgent agent = HarnessAgent.builder()
 | 参数 | 默认 | 含义 |
 |------|------|------|
 | `triggerMessages` | `50` | 按条数触发（`0` 表示关闭） |
-| `triggerTokens` | `80_000` | 按 token 估算触发（`0` 表示关闭） |
+| `triggerTokens` | `0` | 按 token 估算触发（`0` 表示动态计算，基于模型上下文窗口减去 `reserved`） |
```
oss-maintainer, line 82: The zh row for `triggerTokens` was corrected to `0` but its meaning still reads "按 token 估算触发（`0` 表示关闭）" — i.e. *0 = off*, which is exactly the bug this PR fixes. In `CompactionConfig.Builder`, `triggerTokens` defaults to `0` and the Javadoc states `0` means **dynamic mode** (computed from the model context window minus `reserved`), so `0` is not "off" at all. The en row now says `0` = dynamic; the zh row must be aligned, e.g. "按 token 估算触发（`0` = 动态：由模型上下文窗口减去 `reserved` 推算）".

docs/v2/en/docs/harness/memory.md
```diff
@@ -80,9 +80,9 @@ Common options:
 | Field | Default | Meaning |
 |-------|---------|---------|
 | `triggerMessages` | `50` | Trigger by message count (`0` = off) |
-| `triggerTokens` | `80_000` | Trigger by estimated tokens (`0` = off) |
+| `triggerTokens` | `0` | Trigger by estimated tokens; `0` = dynamic (model context window minus `reserved`) |
 | `keepMessages` | `20` | Number of tail messages to keep |
-| `keepTokens` | `0` | When non-zero, walk back by token budget; overrides `keepMessages` |
+| `keepTokens` | `-1` | When non-zero, walk back by token budget; overrides `keepMessages` |
```
oss-maintainer: With the default now correctly documented as `-1`, the sentence "When non-zero, walk back by token budget; overrides `keepMessages`" describes the *default* configuration without explaining what `-1` actually does. In `CompactionConfig` the dynamic keep budget is `min(keepTokensMax, max(keepTokensMin, usable * keepTokensRatio))` — i.e. `-1` auto-computes the value and silently overrides `keepMessages`, which is the surprising behaviour the issue report is about. Worth stating here (and in the zh row at line 84) that `-1` = dynamic and that `keepTokens=0` is what makes `keepMessages` apply verbatim, as the PR body promises.

oss-maintainer (commented):
### Summary

Documentation-only fix in the v2 harness memory tables: `CompactionConfig` defaults are corrected to match `CompactionConfig.Builder` (`triggerTokens = 0`, `keepTokens = -1`), verified against the source at this head. The reported mismatch is real and the `keepTokens` correction is right.

### Findings
- **[Warning]** `docs/v2/zh/docs/harness/memory.md:82` — the zh row now shows the correct default (`0`) but still describes `0` as "关闭" (off), which is the exact misconception this PR removes; the en row in the same PR says `0` = dynamic mode.
- **[Info]** `docs/v2/en/docs/harness/memory.md:85` — with `-1` documented as the default, "when non-zero ... overrides `keepMessages`" now describes the default case without saying that `-1` means "auto-compute from `keepTokensRatio`", i.e. that a configured `keepMessages` is ignored unless `keepTokens=0` is set explicitly.

### Suggestions
Align the zh `triggerTokens` description with en, and spell out the `-1` = dynamic semantics in both languages, since that is the behaviour the issue is about. Non-blocking for a docs change — happy to re-review after the zh row is updated.

CI: `build (ubuntu-latest)` / `build (windows-latest)` still pending; `Check License`, `Check Module Sync`, `validate`, `license/cla` green.

---

oss-maintainer (approved):
### Summary
Re-reviewed the delta since my previous pass (`f710bc84` → `23437775`): only the two doc rows changed, and both non-blocking items from my earlier review are addressed — the zh `triggerTokens` row no longer says `0` = "关闭", and both languages now spell out the three-way `keepTokens` semantics (`-1` dynamic / `0` = `keepMessages` / `>0` fixed budget). Verified again against `CompactionConfig.Builder` at this head: `triggerTokens = 0`, `keepMessages = 20`, `keepTokens = -1`, and the `-1` branch is documented in the javadoc as `min(keepTokensMax, max(keepTokensMin, usable * keepTokensRatio))`. Docs now match source in both languages.

One remaining nit is informational only: `keepTokens` dynamic mode is clamped by `keepTokensMin`/`keepTokensMax` (`2_000`/`8_000`) with `keepTokensRatio = 0.25`, so "auto-computed from the model context window" is an approximation. If a reader ever sets `keepTokens(-1)` expecting the whole window, the clamp is the surprise — a one-line mention would close that gap, but it does not block a docs fix.

### Findings
- **[Info]** `docs/v2/en/docs/harness/memory.md:85` — dynamic mode is `usable * keepTokensRatio` clamped to `[keepTokensMin, keepTokensMax]`; the table says only "auto-computed from the model context window".
- **[Info]** `docs/v2/zh/docs/harness/memory.md:84` — same wording gap in the zh row.

### Verdict
Approving: the reported mismatch is fixed, the change is docs-only, and `Check License` / `Check Module Sync` / `build (ubuntu-latest)` / `build (windows-latest)` / `validate` / `codecov/patch` / `license/cla` are all green on `23437775`.

---

PR #3146, opened 2026-09-14
fix(ollama): surface thinking output as a ThinkingBlock

### LINKED ISSUE #3140: [Bug]: OllamaChatModel drops thinking output — OllamaMessage DTO has no "thinking" field, so reasoning is silently discarded
`Ollama 思考模型的 reasoning 输出被静默丢弃`

### Description

Ollama's `/api/chat` API returns the model's chain-of-thought in a dedicated `message.thinking` field (both in the final response and as per-chunk deltas in streaming mode) when thinking is enabled via the `think` request option. The request side is already supported by `ThinkOption` (`OllamaOptions`), but the response side is not parsed:

- `OllamaMessage` (dto) has no `thinking` field, so Jackson silently drops it during deserialization.
- `OllamaResponseParser` only produces `TextBlock` and `ToolUseBlock` — no `ThinkingBlock` is ever created for the Ollama provider.

As a result, even with `think: true`, the reasoning content never reaches the framework: no `ThinkingBlock` in the final `Msg`, no `ThinkingBlockDeltaEvent` in the agent event stream, nothing to persist or stream to the frontend.

### Reproduction

```java
OllamaChatModel model = OllamaChatModel.builder()
        .baseUrl("http://localhost:11434")
        .modelName("deepseek-r1:8b")
        .defaultOptions(OllamaOptions.builder()
                .thinkOption(ThinkOption.ThinkBoolean.ENABLED)
                .build())
        .build();

model.stream(List.of(new UserMessage("What is 2+2?")), null, null)
        .doOnNext(r -> System.out.println(r.getContent()))
        .blockLast();
```

Actual Ollama response (abridged):

```json
{"message":{"role":"assistant","thinking":"Okay, the user wants 2+2. That's basic arithmetic...","content":"2 + 2 = 4"},"done":true}
```

Observed: only `TextBlock("2 + 2 = 4")` is emitted. The `thinking` content is silently discarded — for every chunk in streaming mode too (`"message":{"thinking":"Okay, ..."}` chunks arrive before content chunks).

Expected: a `ThinkingBlock` should be produced (non-streaming), and `ThinkingBlockDeltaEvent`s should flow through the agent event stream (streaming), consistent with the OpenAI module (`reasoning_content` + `@JsonAlias("reasoning")`) and the Anthropic module (`thinking` / `thinking_delta`).

### Environment

- AgentScope Java 2.0.3 (also verified against current `main` — `OllamaMessage` still has no `thinking` field there)
- Ollama 0.9+ with a thinking model (deepseek-r1, qwen3, gpt-oss, ...)

### Suggested fix

Mirror the Anthropic module's approach:

1. `OllamaMessage`: add `@JsonProperty("thinking") private String thinking;` (+ getter/setter)
2. `OllamaResponseParser`: emit `ThinkingBlock.builder().thinking(msg.getThinking()).build()` when non-empty (before the text block, matching the Anthropic/OpenAI ordering). Both streaming and non-streaming paths go through `formatter.parseResponse`, so a single change covers both.
3. (Optional, for multi-turn replay) `OllamaMessageConverter`: convert assistant `ThinkingBlock` back to `message.thinking` on the request side, so reasoning survives round-trips for thinking models.

Happy to submit a PR with tests if the approach sounds good.

CryoThrust:
## Summary
- bind `message.thinking` on `OllamaMessage` (response-only; Ollama's request schema has no such field)
- emit a `ThinkingBlock` from `OllamaResponseParser` before the text block, matching the OpenAI and Anthropic modules

Both `stream(...)` and the blocking call go through the same `parseResponse`, so this covers streaming and non-streaming, and `ReActAgent` already turns a `ThinkingBlock` into the `ThinkingBlock*Event`s. No core changes.

## Tests
Added to `OllamaResponseParserTest`:
- a raw Ollama JSON payload with `message.thinking` deserializes and produces `ThinkingBlock` then `TextBlock` (this is the end-to-end shape of the reported case)
- a thinking-only chunk produces just the `ThinkingBlock`
- a response without thinking is unchanged

Without the DTO/parser change these do not even compile against the new getter and the JSON case fails, so they are a real regression guard.

- `agentscope-extensions-model-ollama`: 169 run, 0 failures, 1 skipped
- `spotless:check` clean

Closes #3140.


agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/dto/OllamaMessage.java
```diff
@@ -35,6 +35,13 @@ public class OllamaMessage {
     @JsonProperty("content")
     private String content;
 
+    /**
+     * The model's reasoning content, returned by thinking models when the request enables the
+     * {@code think} option. Response-only: Ollama's request message schema does not accept it.
+     */
+    @JsonProperty("thinking")
+    private String thinking;
```
oss-maintainer, line 43: **[Info]** `OllamaMessage` is shared by the request and response paths, so a response-only field is easy to leak back outbound. `@JsonInclude(NON_NULL)` on the class hides it while it stays `null`, which is why this is safe today. Could you add a short note (or a guard) in the request formatter so a future change that copies a whole `OllamaMessage` into an outbound turn cannot start sending `thinking` back to `/api/chat`? That would make the "response-only" claim in the javadoc enforced rather than conventional.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/dto/OllamaMessage.java
```diff
@@ -35,6 +35,13 @@ public class OllamaMessage {
     @JsonProperty("content")
     private String content;
 
+    /**
+     * The model's reasoning content, returned by thinking models when the request enables the
+     * {@code think} option. Response-only: Ollama's request message schema does not accept it.
+     */
+    @JsonProperty("thinking")
```
oss-maintainer, line 42: **[Warning]** Ollama exposes reasoning text under two field names depending on the surface: `message.thinking` for `/api/chat` with `think: true`, and `message.reasoning` for some reasoning models (and in the OpenAI-compatible layer). Only `thinking` is bound here, so for those models the reasoning text is still silently dropped — the same symptom #3140 reports. Consider falling back to `reasoning` when `thinking` is absent, or state the supported model/API set explicitly in the PR description so the issue can be closed with a documented scope.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/formatter/OllamaResponseParser.java
```diff
@@ -48,12 +49,17 @@ public ChatResponse parseResponse(OllamaResponse response) {
 
         List<ContentBlock> contentBlocks = new ArrayList<>();
 
-        // 1. Handle Text Content
+        // 1. Handle Thinking Content (thinking models with the `think` option enabled)
+        if (msg != null && msg.getThinking() != null && !msg.getThinking().isEmpty()) {
+            contentBlocks.add(ThinkingBlock.builder().thinking(msg.getThinking()).build());
```
oss-maintainer, line 54: **[Info]** In streaming mode every chunk goes through `parseResponse`, so N thinking deltas become N separate `ThinkingBlock`s in N `ChatResponse`s. Please confirm the downstream `ReActAgent` → `ThinkingBlock*Event` path concatenates them (the OpenAI/Anthropic modules have the same shape, so this is likely already handled) and consider adding one streaming-level test that feeds a multi-chunk thinking sequence — the three new tests are all single-response, so the fragmentation case is currently unguarded.

oss-maintainer (commented):
### Summary
`OllamaMessage.thinking` + a `ThinkingBlock` emitted ahead of the text block, mirroring the OpenAI/Anthropic modules. The change is correctly scoped to the Ollama extension, the block ordering matches the other providers, and `stream(...)`/blocking both go through `parseResponse`, so the fix covers both paths. No API breakage: the new DTO field is additive and `@JsonInclude(NON_NULL)` keeps it out of outbound payloads. CI is green on `9e1cd877` (build ubuntu/windows, Check License, Check Module Sync, codecov/patch). Verdict: good to go apart from the field-name coverage question below.

### Findings
- **[Warning]** `OllamaMessage.java:42` — only `thinking` is bound; Ollama uses `message.reasoning` on some models/paths, which would keep the reported symptom alive for those users.
- **[Info]** `OllamaMessage.java:43` — response-only field on a DTO shared with the request path; safe today, worth a guard/note so it cannot leak outbound later.
- **[Info]** `OllamaResponseParser.java:54` — streaming yields one `ThinkingBlock` per chunk; the new tests only cover single-response parsing, so the delta-concatenation path is unguarded.

### Suggestions
If `reasoning` fallback is out of scope, mentioning it in the PR description (or in a follow-up issue) is enough — the `Closes #3140` claim should not be read as "all thinking-capable Ollama models now surface thinking".

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/dto/OllamaMessage.java
CryoThrust: Good catch. In `a63902d2`, I added an explicit sanitization pass in `OllamaChatFormatter.buildRequest` and `OllamaMultiAgentFormatter.buildRequest` that clears `thinking` (sets it to null) across all messages before the request payload is constructed. So even if an `OllamaMessage` is copied or reused from a prior response, `thinking` is never leaked outbound to `/api/chat`. Also added `testBuildRequestClearsThinkingFromOutboundMessages` in `OllamaChatFormatterTest` to verify this behavior.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/dto/OllamaMessage.java
CryoThrust: Added `@JsonAlias({"reasoning", "reasoning_content"})` to `OllamaMessage.thinking` in `a63902d2`. This handles native `/api/chat` `thinking` as well as OpenAI-compatible proxies or models exposing `reasoning` or `reasoning_content`, surfacing all of them as a `ThinkingBlock`. Verified with `testReasoningAliasesSurviveJsonDeserialization` in `OllamaResponseParserTest`.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/formatter/OllamaResponseParser.java
CryoThrust: Traced the downstream path: in `ReActAgent`, streaming chunks go through `ReasoningContext`, which uses `ThinkingAccumulator` to concatenate incoming `ThinkingBlock.getThinking()` deltas into the full reasoning content (while emitting each block immediately for real-time events).

To guard against fragmentation regressions, I added `testStreamingThinkingAccumulation` in `OllamaResponseParserTest` (`a63902d2`), which simulates an Ollama multi-chunk delta sequence and verifies `ThinkingAccumulator` stitches the deltas into the complete string.

oss-maintainer (approved):
### Summary

Re-reviewed at `a63902d2`. All three items from the previous round are addressed with tests, and CI (build ubuntu/windows, Check License, Check Module Sync, codecov/patch) is green on this commit. Approving.

- `@JsonAlias({"reasoning", "reasoning_content"})` on `OllamaMessage.thinking` — closes the field-name coverage gap; serialization still emits `thinking` only.
- Outbound sanitization pass in both `OllamaChatFormatter.buildRequest` and `OllamaMultiAgentFormatter.buildRequest` clears `thinking` before the request goes out, with a pinning test.
- Streaming delta path is now covered by `testStreamingThinkingAccumulation` against `ThinkingAccumulator`.

Non-blocking [Info] OllamaChatFormatter.java (buildRequest) — the guard mutates the caller-supplied `OllamaMessage` list in place (`setThinking(null)`). If a caller ever reuses the same message objects across providers or retries, their thinking content would be silently dropped downstream. A copy-on-format approach would remove that side effect; fine to leave as-is given per-request formatting.

---

jujn:
> bind message.thinking on OllamaMessage (response-only; Ollama's request schema has no such field)

请求测 OllamaMessage 不支持 thinking，这个结论是怎么得出的？查了下貌似 Ollama 的文档和源码不一致

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/dto/OllamaMessage.java
```diff
@@ -35,6 +36,16 @@ public class OllamaMessage {
     @JsonProperty("content")
     private String content;
 
+    /**
+     * The model's reasoning content, returned by thinking models when the request enables the
+     * {@code think} option. In native Ollama {@code /api/chat} this is returned in {@code thinking};
+     * some reasoning models or proxies expose it under {@code reasoning} or {@code reasoning_content}.
+     * Response-only: Ollama's request message schema does not accept it.
```
oss-maintainer, line 43: **[Warning]** `Response-only: Ollama's request message schema does not accept it` is not accurate, and it is the premise behind the new outbound strip in `a63902d2` — it is also exactly the question @jujn left on this PR, which is still open.

Upstream accepts `thinking` inbound:

- `api/types.go` (ollama/ollama@main): `Thinking string `json:"thinking,omitempty"`` sits on the same `Message` struct used for `ChatRequest.Messages`, i.e. request history — `omitempty` means optional, not rejected.
- `docs/api.md` → *Generate a chat completion* → **Parameters** → "The `message` object has the following fields": `thinking`: (for thinking models) the model's thinking process.
- `docs/capabilities/tool-calling.md` builds the follow-up request as `messages.append({'role': 'assistant', 'thinking': thinking, 'content': content, 'tool_calls': tool_calls})` — the accumulated reasoning is deliberately echoed back.

Suggested wording: the Java request path does not populate it today, rather than the wire schema rejecting it.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/formatter/OllamaChatFormatter.java
```diff
@@ -447,6 +447,16 @@ public OllamaRequest buildRequest(
             List<ToolSchema> tools,
             ToolChoice toolChoice) {
 
+        // Ensure response-only fields like 'thinking' are not sent outbound to /api/chat,
+        // even if an OllamaMessage was copied or reused from a prior response.
+        if (messages != null) {
+            for (OllamaMessage msg : messages) {
+                if (msg != null && msg.getThinking() != null) {
+                    msg.setThinking(null);
```
oss-maintainer, line 455: **[Warning]** This is a guard against a case that cannot happen today, and it blocks the case that upstream recommends.

Today it is a no-op on the real path: request messages are built by `OllamaMessageConverter`, which maps only `TextBlock` / `ImageBlock` / `ToolUseBlock` — a `ThinkingBlock` from a prior turn is already dropped during conversion, so an outbound `OllamaMessage` never carries `thinking` (which is also why `@JsonInclude(NON_NULL)` alone was enough). So the guard adds a rule, not a fix.

The rule it encodes is the wrong direction: per `docs/capabilities/tool-calling.md`, for `think: true` + tools the assistant `thinking` should travel back with `tool_calls` in the follow-up request. If someone later wires `ThinkingBlock` into the history (the natural next step for thinking models), this loop will silently discard it and the model loses its own reasoning across the tool round-trip.

Options, any is fine:
1. Remove the loop and instead make the intent explicit where it belongs — keep `ThinkingBlock` out of `OllamaMessageConverter`, or map it and let it go outbound — plus reword the `OllamaMessage` note (see the comment on that line);
2. Keep it but scope it (skip assistant messages that carry `tool_calls`) and reference the upstream doc so the next reader does not widen it.

Non-blocking either way: `setThinking(null)` mutates the caller's `OllamaMessage` objects in place, so a list reused across a retry or another provider is silently rewritten. Building sanitized copies keeps `buildRequest` pure.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/formatter/OllamaMultiAgentFormatter.java
```diff
@@ -295,6 +295,16 @@ public OllamaRequest buildRequest(
             List<ToolSchema> tools,
             ToolChoice toolChoice) {
 
+        // Ensure response-only fields like 'thinking' are not sent outbound to /api/chat,
+        // even if an OllamaMessage was copied or reused from a prior response.
+        if (messages != null) {
+            for (OllamaMessage msg : messages) {
+                if (msg != null && msg.getThinking() != null) {
+                    msg.setThinking(null);
```
oss-maintainer, line 303: **[Info]** Same loop, second copy. If the guard stays at all, one shared helper (or a single call in `OllamaChatModel` before `buildRequest`) is enough — the two formatters otherwise drift, and a future fix (e.g. scoping it for tool-call continuations) has to be applied twice.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/test/java/io/agentscope/extensions/model/ollama/formatter/OllamaChatFormatterTest.java
```diff
@@ -526,6 +527,22 @@ void testBuildRequest() {
...
+        OllamaMessage assistantMsg = new OllamaMessage("assistant", "Hi there");
+        assistantMsg.setThinking("Internal chain-of-thought");
+
+        List<OllamaMessage> messages = Arrays.asList(userMsg, assistantMsg);
+        OllamaRequest request =
+                formatter.buildRequest("test-model", messages, false, null, null, null, null);
+
+        assertNotNull(request);
+        assertEquals(2, request.getMessages().size());
+        assertNull(request.getMessages().get(1).getThinking());
```
oss-maintainer, line 543: **[Info]** `request.getMessages()` is the very list that was handed to `buildRequest`, so `assertNull(...)` pins the in-place mutation instead of the wire format — it would still pass if the serializer emitted `thinking`. For the intent stated in the comment, assert on the payload: serialise the built request (the same codec the transport uses) and assert the JSON has no `"thinking"` key. If the tool-call continuation case is meant to keep thinking, that positive case needs an assertion too — right now no test distinguishes "never sent" from "dropped by mistake".

oss-maintainer (commented):
### Summary

Delta re-review of `a63902d2` (previous round was approved at `9e1cd877`). The three items from that round are genuinely addressed — `@JsonAlias({"reasoning", "reasoning_content"})` closes the field-name gap, and `testStreamingThinkingAccumulation` covers the delta-concatenation path. CI is green on this head (build ubuntu/windows, Check License, Check Module Sync, codecov/patch) and `license/cla` is `success`.

What I want to slow down is the new **outbound sanitization** added in this commit. Its stated premise — "`thinking` is response-only; Ollama's request schema does not accept it" — does not hold against upstream, and that is the same question @jujn raised on the PR at 04:09 which is still unanswered:

- `api/types.go` (ollama/ollama@main) declares `Thinking string \`json:"thinking,omitempty"\`` on the `Message` struct used for `ChatRequest.Messages` — inbound history accepts it, `omitempty` makes it optional rather than illegal.
- `docs/api.md` → *Generate a chat completion* → Parameters → "The `message` object has the following fields" lists `thinking`.
- `docs/capabilities/tool-calling.md` composes the follow-up request as `messages.append({'role': 'assistant', 'thinking': thinking, 'content': content, 'tool_calls': tool_calls})` — thinking is deliberately echoed back for `think: true` + tools.

On top of that, the guard cannot bite today: request messages come from `OllamaMessageConverter`, which maps only `TextBlock`/`ImageBlock`/`ToolUseBlock` and never `ThinkingBlock`, so an outbound `OllamaMessage.thinking` is always `null` — which is exactly why `@JsonInclude(NON_NULL)` was already sufficient. Net effect: two new copies of a loop that changes nothing now, and encodes a rule that will discard reasoning content the moment thinking is threaded through history.

### Findings
- **[Warning]** `OllamaMessage.java:43` — "request schema does not accept it" is contradicted by `api/types.go` and `docs/api.md`; reword, since the guard's rationale depends on it.
- **[Warning]** `OllamaChatFormatter.java:455` — no-op today, wrong direction against the documented tool-call continuation flow; also mutates the caller's message list in place.
- **[Info]** `OllamaMultiAgentFormatter.java:303` — second copy of the same loop; one helper if it stays.
- **[Info]** `OllamaChatFormatterTest.java:543` — asserts the in-place mutation, not the serialized payload.

### Suggestions

Either drop the guard and reword the DTO note, or keep it scoped (skip assistant messages carrying `tool_calls`) with a comment citing the upstream doc. Then pin the behaviour in the test that actually matters — serialize the built request and assert no `"thinking"` key appears — so the intent survives future edits. Nothing here touches the parser change itself, which looks right: `ThinkingBlock` before `TextBlock`, shared by streaming and blocking paths, no core impact.

Holding this as a comment rather than a re-approval until @jujn's question is resolved — the reasoning-block emission is fine, but the commit's stated invariant is not, and it is the part a future reader will trust.

---

CryoThrust:
@jujn 确实是我之前看走眼了！我一开始只看了几个 response 的 json 例子和最初提 issue 时的描述，误以为入参只认 content 和 tool_calls，把 thinking 当成纯返回字段了。

后来回头翻了下 Ollama 官方的 `api/types.go` 和 `docs/capabilities/tool-calling.md`，发现请求和响应共用的同一个 `Message` 结构体，而且多轮工具调用时官方就是把 assistant 的 thinking 带在上下文里往回传的。

上一版我在 formatter 里多加的那层强行置空拦截纯属帮倒忙，最新的提交（`144049a8`）已经把那段置空代码全部删掉了，注释也修正了。多谢指出！

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/dto/OllamaMessage.java
CryoThrust: Agreed completely. Updated the javadoc on `thinking` in `144049a8`: removed the inaccurate "response-only" note and explained that Ollama's schema accepts it optionally (e.g. for tool calling reasoning echo), while `OllamaResponseParser` extracts it inbound into a `ThinkingBlock`.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/formatter/OllamaChatFormatter.java
CryoThrust: Went with Option 1 in `144049a8`: removed the stripping loop entirely from `OllamaChatFormatter`. That keeps `buildRequest` pure without in-place mutation, and leaves the door open for tool-calling continuations to echo reasoning back as upstream recommends.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/main/java/io/agentscope/extensions/model/ollama/formatter/OllamaMultiAgentFormatter.java
CryoThrust: Removed the loop here as well in `144049a8`.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/test/java/io/agentscope/extensions/model/ollama/formatter/OllamaChatFormatterTest.java
CryoThrust: Updated the tests in `144049a8` to assert directly on the serialized payload:
- `testRequestSerializationOmitsThinkingWhenNull`: when `thinking` is null (the usual converted case), `@JsonInclude(NON_NULL)` omits the `"thinking"` key entirely;
- `testRequestSerializationIncludesThinkingWhenPresent`: when `thinking` is present, it serializes `"thinking"` into the JSON payload properly.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-ollama/src/test/java/io/agentscope/extensions/model/ollama/formatter/OllamaChatFormatterTest.java
```diff
@@ -526,6 +527,42 @@ void testBuildRequest() {
...
+                        Collections.singletonList(assistantMsg),
+                        false,
+                        null,
+                        null,
+                        null,
+                        null);
+
+        assertNotNull(request);
+        String json = io.agentscope.core.util.JsonUtils.getJsonCodec().toJson(request);
+        assertTrue(json.contains("\"thinking\":\"Planning tool call\""));
```
oss-maintainer, line 564: **[Info]** Asserting on the literal `"thinking":"Planning tool call"` couples the test to Jackson's exact output formatting. Today `JsonUtils.getJsonCodec()` writes without spaces after the colon, so this passes, but any future change to the codec configuration (pretty printing, a shared mapper tweak) breaks it for reasons unrelated to the intent. `assertTrue(json.contains("\"thinking\""))` plus a tree-parse check on the value keeps the same pinning without the formatting coupling. Non-blocking — the null-omission half of this pair is the assertion that actually guards the PR.

oss-maintainer (approved):
### Summary

Delta re-review at `144049a8` (previous round was `COMMENTED` on `a63902d2`). All four open items from that round are genuinely resolved and the fix itself is unchanged:

- The javadoc no longer claims the request schema rejects `thinking`; it now states it is an optional message field and points at `OllamaResponseParser` for the inbound extraction. That removes the premise the previous commit was built on.
- Both outbound stripping loops are gone (`OllamaChatFormatter.buildRequest`, `OllamaMultiAgentFormatter.buildRequest`), so `buildRequest` no longer mutates the caller's message list and tool-call continuations can still echo reasoning back as upstream documents. The duplicated loop is moot rather than deduplicated — fine.
- The wire-format test is now the one that actually pins intent: `testRequestSerializationOmitsThinkingWhenNull` asserts the built request has no `"thinking"` key via the shared `JsonUtils` codec, and `testRequestSerializationIncludesThinkingWhenPresent` pins the opposite direction so a future `@JsonInclude` change cannot silently drop reasoning.

Net effect on behaviour: `OllamaMessageConverter` still never populates `thinking`, so outbound payloads are byte-identical to before this PR for every current code path, while inbound `thinking` / `reasoning` / `reasoning_content` now surfaces as a `ThinkingBlock` ahead of the text block on both the streaming and blocking paths. No core API impact.

CI is green on this head (build ubuntu-latest / windows-latest, Check License, Check Module Sync, codecov/patch) and `license/cla` is `success`.

### Findings

Nothing blocking. One non-blocking note is inline.

### Suggestions

The three oldest threads on this PR (field-name coverage, the shared-DTO leak note, streaming delta fragmentation) were addressed in `a63902d2` but are still showing as unresolved on GitHub — worth resolving them so the PR reads clean for a maintainer.

Approving — thanks for turning the guard around instead of patching it.

---

oss-maintainer:
⚠️ **Review anchoring note — CI needs one more green run on the current head**

My `APPROVE` landed at 03:44 UTC and GitHub anchored it to `25a2151a` (`Merge branch 'main' into fix/ollama-thinking-block`, pushed 03:41:51 UTC), while the review text was written against `144049a8`. To be explicit about what was actually reviewed:

- The reviewed delta is `144049a8` — the javadoc correction plus removing the two outbound stripping loops and rewriting the tests to assert on the serialized payload. The verdict below applies to that change.
- I re-checked the merge commit: `OllamaResponseParser.parseResponse` at `25a2151a` still emits `ThinkingBlock` before the text block, and main's changes to `OllamaResponse` / `OllamaResponseParser` / `OllamaResponseParserTest` merged without altering this PR's behaviour. So the approval still holds on the code.

What does not hold is the CI claim in my review body — green was `144049a8`, not the current head:

- `build (windows-latest)` — **failed** on `LocalFilesystemPersonalAssistantExampleTest.localFilesystem_filesPersistAcrossCalls`, which does not touch the Ollama module. The same test fails on an unrelated branch (`codex/issue-3166-force-sync-cap`, run 35497408703), so this reads as a pre-existing Windows-only flake, not something #3146 introduced.
- `build (ubuntu-latest)` — **cancelled** before finishing, so it says nothing either way.

No code change needed from this. Please just push a no-op or re-run the workflow so the merge head has a clean `build (ubuntu-latest)`, and confirm the Windows failure is tracked as a known flake rather than attributed to this PR. Keeping this comment short and not re-reviewing — the content review stands.

---

PR #3149, opened 2026-09-15
fix(core): handle synchronous failures in model fallback

### LINKED ISSUE #3148: [Bug]: Synchronous model stream failures bypass the configured fallback
**Describe the bug**

With `fallbackModel` configured, a failure before the primary emits its first response takes different paths depending on how `Model.stream()` reports it:

- `return Flux.error(failure)` switches to the fallback and notifies `FailoverListener`.
- `throw failure` before returning the Flux aborts the agent call; neither the fallback nor the listener runs.

`ReActAgent.modelForCall()` invokes `model.stream(...)` before installing `switchOnFirst`. A synchronous exception therefore escapes that fallback boundary. This matters for custom model adapters and provider-specific setup/formatting failures. `ChatModelBase.stream()` also delegates through `Tracer.callModel()` directly to the supplied implementation, so it does not generally normalize synchronous exceptions.

**To Reproduce**

In the repository's `agentscope-core` JUnit tests, the following uses the existing `MockModel` and `TestUtils` fixtures; no credentials or model endpoint are needed:

```java
import static org.junit.jupiter.api.Assertions.assertEquals;
import io.agentscope.core.ReActAgent;
import io.agentscope.core.agent.test.MockModel;
import io.agentscope.core.agent.test.TestUtils;
import io.agentscope.core.model.ChatResponse;
import io.agentscope.core.model.GenerateOptions;
import io.agentscope.core.model.ToolSchema;
import java.time.Duration;
import java.util.List;
import reactor.core.publisher.Flux;

class SynchronousFallbackTest {
    @Test
    void shouldUseConfiguredFallback() {
        Model primary = new Model() {
            @Override
            public Flux<ChatResponse> stream(
                throw new IllegalStateException("primary unavailable");
            }
            @Override
            public String getModelName() { return "primary"; }
        };
        MockModel fallback = new MockModel("fallback response");
                .name("demo").model(primary).fallbackModel(fallback).build();
        Msg response = agent.call(TestUtils.createUserMessage("user", "hello"))
                .block(Duration.ofSeconds(5));
        assertEquals("fallback response", TestUtils.extractTextContent(response));
        assertEquals(1, fallback.getCallCount());
    }
}
```

Run `mvn -pl agentscope-core -Dtest=SynchronousFallbackTest -Dspotless.skip=true test`.
Replacing the `throw` with `return Flux.error(new IllegalStateException("primary unavailable"));` makes the same case succeed.

**Expected behavior**

A synchronous model invocation failure before any response should use the same configured fallback path as an initial reactive error, with the original error delivered to the listener. Errors after a response has already been emitted should continue to propagate without switching models.

**Error messages**

`IllegalStateException: primary unavailable` escapes the agent call. A parameterized local regression covering both error forms plus a mid-stream-error control reports `Tests run: 3, Failures: 0, Errors: 1, Skipped: 0` on unmodified main; the synchronous case is the sole error.

**Environment**

- AgentScope Java: `2.0.3-SNAPSHOT`, main `60934595a7e9f6805bd0649a68d26766b158042e`
- Java: Temurin 17.0.15; Maven 3.9.12
- OS: Windows

**Additional context**

The minimal fix I tested defers the primary invocation with `Flux.defer(() -> model.stream(...))` before `switchOnFirst`. All three boundary cases then pass, as do the selected agent, builder, configuration and timeout/retry tests (57 tests total). It does not add retry or cooldown policy.

I checked #1851, #3145, #2864 and #2968: the listener addition and original single-fallback wiring retain the eager invocation; the optional multi-model chain leaves the legacy single-fallback path unchanged; empty-completion handling is a different case. I would like to contribute the small fix and regression tests.

AI assistance was used for source investigation and the patch; the reported local tests were executed and the diff reviewed. No live provider outage or HTTP response is claimed.

pengmoubuaixuexi:
## AgentScope-Java Version

`2.0.3-SNAPSHOT`, based on main `60934595a7e9f6805bd0649a68d26766b158042e`.

## Description

Fixes #3148.

A configured fallback currently runs when the primary returns `Flux.error(error)`, but not when `Model.stream()` throws the same error before returning its publisher. The eager invocation happens before `switchOnFirst`, so the agent fails without trying the fallback or notifying the failover listener.

Defer primary invocation so synchronous setup failures enter the existing first-signal fallback path. A primary that has already emitted a response still propagates a subsequent error without switching, avoiding mixed output from two models. Retry budgets and cooldown policy are unchanged.

The new test covers both initial error forms, the original error/model delivered to the listener, one fallback invocation, and no switch after partial output. On unmodified main: 3 tests, 2 pass and the synchronous case errors. With this fix: 57 selected core tests pass (0 failures/errors/skips):

```text
mvn -B -pl agentscope-core -Dtest=ReActAgentFailoverBoundaryTest,ReActAgentTest,ReActAgentNewLoopBuilderTest,ReActAgentGenerateOptionsTest,ModelConfigTest,ModelTimeoutRetryTest -Dspotless.skip=true test
```

Scoped Spotless apply/check and `git diff --cached --check` pass. Tests use local model fixtures and make no provider requests. The full repository suite was not run locally.

AI assistance was used for investigation, implementation and test preparation; the stated tests were executed and the diff reviewed.

## Checklist

- [x] Code has been formatted with scoped `mvn spotless:apply` and verified with `spotless:check`
- [ ] All repository tests are passing (`mvn test`) — 57 relevant tests passed locally; full-suite CI pending
- [x] No public API or Javadoc contract changes
- [x] No user-facing configuration changes; the inline comment explains the deferred invocation


oss-maintainer (approved):
### Summary
Wrapping the primary `model.stream(...)` call in `Flux.defer(...)` so a synchronous setup failure (thrown before a publisher exists) enters the existing `switchOnFirst` error path instead of escaping `call()`. This closes the gap where a configured fallback and the failover listener were only honoured for the `Flux.error(...)` form — the fix is minimal, correctly targeted, and does not change retry budgets, cooldown policy, or any public API.

`ReActAgentFailoverBoundaryTest` parameterises both failure forms (throw / `Flux.error`) and additionally pins the two invariants that matter here: the fallback is invoked exactly once with the original error delivered to the listener, and a primary that has already emitted content is not switched mid-stream (so no mixed output from two models). Tests use local fixtures only, CI is green on ubuntu + windows, and the change is bounded to the fallback decorator.

No findings. LGTM.

---

gmjneko:
LGTM

jujn (approved):

Issue #3154, opened 2026-09-15
[Bug]:deepseekv4.1使用dashscope调用百炼平台报错


ligyff:
errorType=io.agentscope.extensions.model.dashscope.DashScopeHttpClient$DashScopeHttpException | message=DashScope API error: url error, please check url！ For details,

v4.1是一个多模态的模型，是不是内部没有添加判断条件

jujn:
是个问题，DashScope 多模态模型检测是框架写死的，得找个方案让用户自行扩展

oss-maintainer:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed** (code-level root cause found; one detail needed from the reporter)

Verified against `main` (`97696a3a`).

**Root cause:** `DashScopeHttpClient` picks the native API endpoint purely from a hardcoded model-name whitelist (`isMultimodalModel`), which only matches `qvq` / `-vl` / `-asr` / `qwen3.x` / `kimi-k` patterns. `deepseek-*` is absent, so a multimodal DeepSeek call is routed to `/api/v1/services/aigc/text-generation/generation`, and Bailian answers with its documented model/endpoint-mismatch error — surfaced verbatim as the reported `DashScope API error: url error, please check url`. The hypothesis in the report (a missing multimodal branch) is correct.

**Evidence:**
- `DashScopeHttpClient.java:343-362` — AUTO routing selects text vs multimodal endpoint by model name only
- `DashScopeHttpClient.java:392-413` — `isMultimodalModel` whitelist has no `deepseek` pattern
- `DashScopeChatModel.java:159` — `endpointType` defaults to `AUTO`, so the whitelist governs unless overridden
- `DashScopeHttpClient.java:214-219` — upstream error body wrapped as `DashScopeHttpException("DashScope API error: ...")`
- `DashScopeHttpClient.java:70-75,549-550` — client is native-protocol only; `buildUrl` is a plain `baseUrl + endpoint` concat

**Impact:** `agentscope-extensions-model-dashscope` (`DashScopeChatModel` / `DashScopeHttpClient`) and the Spring Boot DashScope starter that builds the same model. Deterministic for every multimodal model name missing from the whitelist.

**Severity:** medium — a workaround exists today:
```java
DashScopeChatModel.builder()... .endpointType(EndpointType.MULTIMODAL).build();
```

[...]

PR #3163, opened 2026-09-15
fix(dashscope): make multimodal endpoint detection user-extensible

xiechimon:
fix(dashscope): make multimodal endpoint detection user-extensible

**Problem**

`DashScopeHttpClient` picks the native API endpoint purely from a hardcoded model-name
whitelist (`isMultimodalModel`), which only matches `qvq` / `-vl` / `-asr` / `qwen3.x` /
`kimi-k` patterns. Multimodal models not in the list — e.g. `deepseek-v4.1-flash` from
#3154 — are routed to `/api/v1/services/aigc/text-generation/generation`, and Bailian
answers with its documented model/endpoint-mismatch error surfaced as
`DashScope API error: url error, please check url`.

The only workaround today is setting `endpointType(EndpointType.MULTIMODAL)` per model, and
every new multimodal release re-triggers the same class of failure (maintenance treadmill
noted in the issue evaluation).

**Solution**

Make the AUTO detection user-extensible instead of extending the hardcoded whitelist:

- Add `multimodalModelPatterns(Collection<String>)` to `DashScopeHttpClient.Builder`
  and `DashScopeChatModel.Builder`.
- Patterns are normalized (trimmed, lowercased, blanks skipped) and matched as
  case-insensitive substrings of the model name, mirroring the existing `contains`-style
  rules of `isMultimodalModel`.
- In AUTO mode, a model is routed to the multimodal API if it matches the built-in
  whitelist **or** any user pattern. The whitelist stays as the default; `TEXT` /
  `MULTIMODAL` explicit overrides are unchanged.
- No new dependencies; the existing public constructors keep working (new constructor
  overload delegates with `null`).

Example:
```java
DashScopeChatModel model = DashScopeChatModel.builder()
    .apiKey("sk-xxx")
    .modelName("deepseek-v4.1-flash")
    .multimodalModelPatterns(List.of("deepseek-v4"))
    .build();
```

**Verification (live, against Bailian)**

Same model (`deepseek-v4.1-flash`), same request body, only the endpoint differs.

| | endpoint | result |
|---|---|---|
| AUTO today (no patterns) | `text-generation` | **HTTP 400** `url error, please check url` |
| AUTO with `multimodalModelPatterns("deepseek-v4")` | `multimodal-generation` | **HTTP 200** `Hi! How can I help you today?` |

Reproduced both with plain `curl` (server-side) and through the patched client end-to-end
(`VerifyFix.java`, screenshot below):

verify-3154

<details><summary>VerifyFix.java (40 lines, classpath from <code>dependency:build-classpath</code>)</summary>

```java
import io.agentscope.extensions.model.dashscope.DashScopeChatModel;
import io.agentscope.extensions.model.dashscope.DashScopeHttpClient;
import io.agentscope.core.message.MsgRole;
import java.util.List;

public class VerifyFix {
    static final String MODEL = "deepseek-v4.1-flash";

    public static void main(String[] args) {
        String key = args.length > 0 ? args[0] : System.getenv("DASHSCOPE_API_KEY");
        Msg hi = Msg.builder().role(MsgRole.USER).textContent("hi").build();

        System.out.println("model: " + MODEL);
        System.out.println("AUTO routing says multimodal? "
                + DashScopeHttpClient.isMultimodalModel(MODEL));
        System.out.println();

        System.out.println("[1/2] WITHOUT multimodalModelPatterns  -> text endpoint");
        run(key, hi, null);

        System.out.println();
        System.out.println("[2/2] WITH multimodalModelPatterns(\"deepseek-v4\")  -> multimodal endpoint");
        run(key, hi, List.of("deepseek-v4"));
    }

    static void run(String key, Msg hi, List<String> patterns) {
        try {
            var b = DashScopeChatModel.builder()
                    .apiKey(key).modelName(MODEL).stream(false);
            if (patterns != null) b.multimodalModelPatterns(patterns);
            var r = b.build().stream(List.of(hi), null, null).blockFirst();
            System.out.println("      HTTP 200 OK");
            System.out.println("      reply: " + ((io.agentscope.core.model.ChatResponse) r)
                    .getContent());
        } catch (Exception e) {
            Throwable t = e;
            while (t.getCause() != null) t = t.getCause();
            System.out.println("      HTTP 400 " + t.getMessage());
        }
    }
}
```

</details>

**Changes**

- `DashScopeHttpClient.java` — new `userMultimodalPatterns` field, 6-arg constructor
  overload, `normalizePatterns`, `matchesUserPattern`, AUTO branches of `selectEndpoint`
  and `requiresMultimodalApi` now include user patterns, builder option.
- `DashScopeChatModel.java` — new 14-arg constructor overload (13-arg delegates with
  `null`), builder option forwarded to the HTTP client.
- `DashScopeHttpClientTest.java` — 3 new tests: extension for `deepseek-v4.1`, case
  insensitivity / trimming / substring semantics, and null/empty patterns having no
  effect on built-in detection.

**Unit tests**

- `mvn -pl agentscope-extensions-model-dashscope -am test -Dtest=DashScopeHttpClientTest`
  — **75/75 pass**, including 3 new tests:
  - `testUserMultimodalPatternsExtendDetection` — `deepseek-v4.1` routes to TEXT
    without patterns (asserts current broken-by-design behavior), to MULTIMODAL with
    pattern `deepseek-v4`, and explicit `TEXT` still overrides the pattern.
  - `testUserMultimodalPatternsAreCaseInsensitiveAndTrimmed` — `" DeepSeek-V4 "`, `""`,
    `null` normalize correctly; substring semantics documented; non-matching model
    unaffected.
  - `testUserMultimodalPatternsNullAndEmptyNoEffect` — `null` / empty keeps existing
    routing, built-in detection intact.
- The new tests do not compile against the base (builder API absent) — they exercise
  exactly the new code path.

**Notes for Reviewer**

- Scope deliberately limited to AUTO-mode detection; the compatible-mode `baseUrl`
  fail-fast from the issue evaluation is intentionally NOT included because
  `DashScopeCredential` legitimately defaults to `.../compatible-mode/v1` and maps to this
  client — failing there would be a breaking change for that path.
- `isMultimodalModel(String)` static API is untouched for backward compatibility.

Refs #3154


oss-maintainer:
**CLA Not Signed**

@xiechimon please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/agentscope-ai/agentscope-java. Once signed, the `license/cla` status will turn green.

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeHttpClient.java
```diff
@@ -438,8 +478,32 @@ public boolean requiresMultimodalApi(String modelName, EndpointType endpointType
...
+     *
+     * <p>Patterns are compared as case-insensitive substrings against the (trimmed) model
+     * name, mirroring the {@code contains}-style rules used by
+     * {@link #isMultimodalModel(String)}. Returns {@code false} when no patterns are
+     * configured or the model name is {@code null}.
+     *
+     * @param modelName the model name
+     * @return true if the model name matches a user-supplied multimodal pattern
+     */
+    private boolean matchesUserPattern(String modelName) {
```
oss-maintainer, line 496: `requiresMultimodalApi` reaches this line with a `null` `modelName` when `endpointType` is `AUTO` (the `MULTIMODAL`/`TEXT` early-returns don't fire), and `userMultimodalPatterns.isEmpty()` short-circuits before the null check — so `matchesUserPattern(null)` dereferences `modelName.trim()` and throws NPE only for users who configured patterns. Guard with `modelName == null ||` first (or flip the condition to check the model name before the pattern set), and add a `null` case to the pattern tests.

oss-maintainer (commented):
### Summary
Extending AUTO endpoint detection with user-supplied substring patterns is the right escape hatch for models outside the built-in whitelist, and the overload-chaining keeps the public API backward compatible; the test set covers case-insensitivity, trimming, null/blank entries and the explicit-`TEXT`-wins case. Two things block an approve: (1) the NPE path below, and (2) the CLA is still pending (see the reminder comment) — a PR cannot be approved or merged unsigned.

### Also
- **Heads-up:** this PR and #3164 are the same author pushing the identical branch state (both HEAD at `e782aec2753d`, same 3 files, +251/-6) — they are exact duplicates. Please keep one (this one, the earlier) and close the other, otherwise the fix will land twice or half-land from the stale copy.

---

jujn (approved):

oss-maintainer (commented):
### Summary
Re-review after `b4f0c118`. The NPE from my previous pass is fixed correctly (`modelName == null` is now checked before the pattern-set short-circuit), the duplicated branch PR #3164 has been closed, and `license/cla` is now `success` on the current head — so all three earlier blockers are cleared. The new commit also extends the option beyond the client into `DashScopeModelProvider` and the Spring Boot starter, which is the right shape for a user-facing escape hatch. Remaining items are non-blocking: both new tests assert things that stay green even if the wiring they are named after is removed, `stringCollectionOption` is stricter than its sibling readers in the same class, and the new user-facing property is missing from the model docs that document every sibling DashScope option.

### Findings
- **[Warning]** `DashScopeModelProvider.java:170` — `multimodalModelPatterns` rejects any non-`Collection` value with `IllegalArgumentException`, while `endpointTypeOption` in the same class coerces strings; string-shaped config will hard-fail model creation.
- **[Warning]** `DashScopeModelProviderTest.java:100` / **[Warning]** `DashScopeAutoConfigurationTest.java:203` — neither new test can detect a regression in the wiring it covers; both pass unchanged if `builder.multimodalModelPatterns(...)` is deleted. Needs a pattern observable on the built model.
- **[Warning]** `DashScopeProperties.java:80` — `docs/v1/{en,zh}/docs/task/model.md` document every sibling DashScope option (`endpointType`, `enableThinking`, `nativeStructuredOutput`) plus an `endpointType` precedence section; the new option is absent from both, even though its purpose is discoverability.
- **[Info]** `DashScopeHttpClient.java:141` — default-locale `toLowerCase()` in `normalizePatterns`/`matchesUserPattern`; `Locale.ROOT` is already the repo convention (~30 usages).
- **[Info]** `DashScopeModelProvider.java:163` — redundant `@SuppressWarnings("unchecked")`, no unchecked cast occurs.
- **[Resolved]** `DashScopeHttpClient.java:497` — the `matchesUserPattern(null)` NPE flagged in my previous review is fixed on this head, and the reordering is exactly the right fix. `EndpointType.TEXT` still wins over user patterns (covered by `testUserMultimodalPatternsAreCaseInsensitiveAndTrimmed` / the `TEXT` assertion in `testUserMultimodalPatternsExtendDetection`).

### Suggestions
The one change I would ask for before merge is making the propagation observable so the two wiring tests are real:

```java
// DashScopeChatModel (package-private or @VisibleForTesting)
boolean requiresMultimodalApi(String name) {
    return httpClient.requiresMultimodalApi(name, endpointType);
}
```

then assert `assertTrue(model.requiresMultimodalApi("deepseek-v4.1"))` in the provider test and the autoconfiguration test, plus a negative case for a plain-text model. A docs row in the two `model.md` files closes it out; after that this looks good to land.

### Cross-repo Note
`isMultimodalModel` is `public static`, so any downstream caller that used it to predict routing will still not see user-supplied patterns. Worth calling out in the docs entry above so users do not treat the static helper as the source of truth for endpoint choice.

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeModelProvider.java
```diff
@@ -151,4 +159,27 @@ private static EndpointType endpointTypeOption(ModelCreationContext context, Str
...
     }
+
+    @SuppressWarnings("unchecked")
+    private static Collection<String> stringCollectionOption(
+            ModelCreationContext context, String key) {
+        Object value = context.option(key);
+        if (value == null) {
+            return null;
+        }
+        if (value instanceof Collection<?> raw) {
```
oss-maintainer, line 170: **[Warning]** Type coercion here is stricter than the sibling option readers in this same class: `endpointTypeOption` (lines 148-161) accepts either the typed value or a `String`, while `multimodalModelPatterns` only accepts a `Collection` and hard-fails with `IllegalArgumentException` on anything else.

`ModelCreationContext.option` values are frequently produced from string-ish config (YAML/env → `Map<String,Object>`, JSON, or a `String[]` from a property splitter), so a user writing `option("multimodalModelPatterns", "deepseek-v4")` gets a model-creation exception instead of a working model. Suggest coercing `String` (comma-split) and `String[]` the same way `endpointTypeOption` coerces, e.g.:

```java
if (value instanceof String text) {
    return Arrays.stream(text.split(",")).map(String::trim).filter(s -> !s.isEmpty()).toList();
}
if (value instanceof String[] arr) {
    return List.of(arr);
}
```

If the strict contract is intentional, please say so in the javadoc/exception message so the failure is self-explanatory.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeModelProvider.java
```diff
@@ -151,4 +159,27 @@ private static EndpointType endpointTypeOption(ModelCreationContext context, Str
         throw new IllegalArgumentException(
                 "ModelCreationContext option " + key + " must be an EndpointType or string");
     }
+
+    @SuppressWarnings("unchecked")
```
oss-maintainer, line 163: **[Info]** `@SuppressWarnings("unchecked")` is not needed here: the method never performs an unchecked cast (`value instanceof Collection<?> raw` is fully checked and elements are converted via `toString()`). Leaving a redundant suppression in place tends to mask real unchecked warnings later on, so it can just be dropped.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/test/java/io/agentscope/extensions/model/dashscope/DashScopeModelProviderTest.java
```diff
@@ -84,4 +85,19 @@ void modelRegistryFindsDashScopeProviderFromServiceLoader() {
...
+        DashScopeModelProvider provider = new DashScopeModelProvider();
+        ModelCreationContext context =
+                ModelCreationContext.builder()
+                        .apiKey("test-dashscope-key")
+                        .option("multimodalModelPatterns", List.of("deepseek-v4.1"))
+                        .build();
+
+        Model model = provider.create("dashscope:deepseek-v4.1", context);
+
+        assertTrue(model instanceof DashScopeChatModel);
```
oss-maintainer, line 100: **[Warning]** `createPassesMultimodalModelPatterns` is named after the wiring, but nothing it asserts depends on the wiring: `model instanceof DashScopeChatModel` and the model name would both still hold if `builder.multimodalModelPatterns(...)` (or the whole `stringCollectionOption` call) were deleted. So this test cannot regress.

To make it assert the actual behavior, the patterns need to be observable on the created model — e.g. a package-private/`@VisibleForTesting` accessor on `DashScopeChatModel` (`Set<String> multimodalModelPatterns()` or `boolean requiresMultimodalApi(String)`) and then:

```java
assertTrue(model.requiresMultimodalApi("deepseek-v4.1", EndpointType.AUTO));
assertFalse(model.requiresMultimodalApi("qwen-plus", EndpointType.AUTO));
```

Also worth pinning the negative case: a non-Collection value (e.g. a plain `String`) currently throws `IllegalArgumentException` — assert whichever behavior you decide in the finding above so the contract is locked in.

Nit: prefer `assertEquals("deepseek-v4.1", model.getModelName())` over `assertTrue(a.equals(b))` — it prints both values on failure.

agentscope-extensions/agentscope-spring-boot-starters/agentscope-dashscope-spring-boot-starter/src/test/java/io/agentscope/spring/boot/dashscope/DashScopeAutoConfigurationTest.java
```diff
@@ -188,6 +188,26 @@ void shouldDelegateAcceptToCustomizeOnDashScopeChatModelBuilderCustomizer() {
...
+                .withPropertyValues(
+                        "agentscope.model.provider=dashscope",
+                        "agentscope.dashscope.api-key=test-dashscope-key",
+                        "agentscope.dashscope.model-name=deepseek-v4.1",
+                        "agentscope.dashscope.multimodal-model-patterns[0]=deepseek-v4",
+                        "agentscope.dashscope.multimodal-model-patterns[1]=deepseek-v5")
+                .run(
+                        context -> {
+                            DashScopeChatModel model = context.getBean(DashScopeChatModel.class);
+                            assertThat(model.getModelName()).isEqualTo("deepseek-v4.1");
```
oss-maintainer, line 203: **[Warning]** Same gap as the provider test: this verifies relaxed-binding of `agentscope.dashscope.multimodal-model-patterns[N]` into `DashScopeProperties`, but the code under review in `DashScopeAutoConfiguration` (the new `builder.multimodalModelPatterns(...)` call) is not covered — deleting those 4 lines keeps every assertion here green, because the assertions are on `getModelName()` and on the properties bean, not on the model's routing.

If you expose the accessor suggested in `DashScopeModelProviderTest`, assert through it here too (`context.getBean(DashScopeChatModel.class)` → pattern reaches endpoint selection), and keep the properties-level assertion as the binding check.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-dashscope/src/main/java/io/agentscope/extensions/model/dashscope/DashScopeHttpClient.java
```diff
@@ -98,11 +102,46 @@ public DashScopeHttpClient(
...
+                userMultimodalPatterns == null || userMultimodalPatterns.isEmpty()
+                        ? Set.of()
+                        : normalizePatterns(userMultimodalPatterns);
+    }
+
+    private static Set<String> normalizePatterns(Collection<String> patterns) {
+        Set<String> normalized = new LinkedHashSet<>();
+        for (String pattern : patterns) {
+            if (pattern != null && !pattern.isBlank()) {
+                normalized.add(pattern.trim().toLowerCase());
```
oss-maintainer, line 141: **[Info]** Case folding uses the default locale (`String.toLowerCase()`), and the same applies to `matchesUserPattern` at line 500. Model names are ASCII, but that only holds as long as every JVM agrees on the mapping: under a Turkish locale, `"I".toLowerCase()` yields `ı` (dotless), so a pattern like `"vision"` from a `tr_TR` process would not match an ASCII-folded model name. The repo already uses `Locale.ROOT` in ~30 places, so `pattern.trim().toLowerCase(Locale.ROOT)` (and the matching call site) keeps the semantics locale-independent. `isMultimodalModel` has the same pre-existing issue, so fixing both here (or in a follow-up) would be consistent.

agentscope-extensions/agentscope-spring-boot-starters/agentscope-dashscope-spring-boot-starter/src/main/java/io/agentscope/spring/boot/dashscope/DashScopeProperties.java
```diff
@@ -66,6 +68,17 @@ public class DashScopeProperties {
...
+    /**
+     * Additional case-insensitive substring patterns for multimodal model detection.
+     *
+     * <p>When {@code endpointType} is {@code AUTO} (the default), a model name matching
+     * any of these patterns is routed to the multimodal generation API, in addition to
+     * the built-in model-name rules. This allows using multimodal models not yet covered
+     * by the built-in whitelist (e.g. {@code deepseek-v4.1}) without waiting for a
+     * framework update.
+     */
+    private List<String> multimodalModelPatterns;
```
oss-maintainer, line 80: **[Warning]** New user-facing knob, but `docs/v1/en/docs/task/model.md` and `docs/v1/zh/docs/task/model.md` are not updated. Those files document the sibling DashScope options (`endpointType`, `enableThinking`, `nativeStructuredOutput`, ...) in the "DashScope → Configuration" table and the `endpointType` section right below it, and `endpointType` is exactly the thing this option interacts with.

This option's whole purpose is discoverability for a user who just hit a wrong-endpoint error on a model missing from the whitelist, so a doc row is arguably the most important part of the change:

```
| `multimodalModelPatterns` | Case-insensitive substring patterns that extend AUTO multimodal endpoint detection (default: none) |
```

Please add it to the builder-option table (and the zh mirror), plus a line in the `endpointType` section clarifying that explicit `TEXT`/`MULTIMODAL` always wins over user patterns — that precedence is currently only expressed by a test comment.

PR #3165, opened 2026-09-15
fix(state): resolve phantom CAS conflicts on version-0 migrated rows …

### LINKED ISSUE #3162: [Bug]: ensureVersionColumn() backfills version=0, the same sentinel getVersioned() uses for "row absent" — every pre-existing session hits a phantom CAS conflict after upgrading
**Describe the bug**

`ensureVersionColumn()` backfills the new optimistic-locking `version` column with `DEFAULT 0`, but `0` is the same value the store uses to mean *"the row does not exist"*. Every session row that existed before the upgrade therefore looks absent to `saveIfVersion()`, which takes the INSERT branch, hits a duplicate key, and reports a **phantom CAS conflict** on the first turn after upgrading.

Confirmed on `main` for `MysqlAgentStateStore` and `PostgresAgentStateStore`. Three places have to agree on the meaning of `0` and they do not:

1. `ensureVersionColumn()` — runs unconditionally from the constructor, independent of `createIfNotExist`:

```java
// MysqlAgentStateStore.java (main)
String alterSql = "ALTER TABLE " + getFullTableName()
        + " ADD COLUMN version BIGINT NOT NULL DEFAULT 0";
```

`PostgresAgentStateStore.ensureVersionColumn()` is identical (`ADD COLUMN IF NOT EXISTS version BIGINT NOT NULL DEFAULT 0`).

2. `getVersioned()` — returns `0` to mean *absent*:

```java
if (!rs.next()) {
    return new VersionedState<>(null, 0L);
}
```

3. `saveIfVersion()` — reads `expectedVersion == 0` as *"the row should not exist yet"*:

```java
result[0] = (expectedVersion == 0L)
        ? insertIfAbsent(conn, slotId, key, value)   // duplicate key -> UNVERSIONED
        : updateIfVersion(conn, slotId, key, value, expectedVersion);
```

Note that both write paths *create* rows at version `1` (`INSERT ... VALUES (?, ?, ?, ?, 1)`), so `0` is never a legitimate stored version. It only ever appears on rows the `ALTER TABLE` backfilled.

**To Reproduce**

1. Run any agent with a SQL `AgentStateStore` on 2.0.2 (or any build predating the `version` column) and have at least one session persisted.
2. Upgrade to 2.0.3 / current `main` and restart. `ensureVersionColumn()` adds `version BIGINT NOT NULL DEFAULT 0`; all pre-existing rows are now at `0`.
3. Send one more message to any of those sessions.
4. `ReActAgent.persistAgentStateCas()` loads `expectedVersion = 0`, calls `saveIfVersion(..., 0)`, gets `UNVERSIONED` back, increments `stateConflictCount` and enters the conflict branch — even though nothing else touched the row.

Equivalent SQL, no application needed:

```sql
-- simulates a row the ALTER TABLE backfilled
UPDATE agentscope_sessions SET version = 0 WHERE session_id = ? AND state_key = 'agent_state';
-- the next saveIfVersion(..., 0) will attempt an INSERT and fail on the primary key
```

**Expected behavior**

Upgrading should not produce conflicts on sessions that only one writer is touching. Either the migration should not leave rows at the sentinel value, or the sentinel should not be a value the column can hold.

Suggested fix, in `ensureVersionColumn()` right after the `ALTER`:

```sql
UPDATE <table> SET version = 1 WHERE version = 0;
```

This is safe because both insert paths already start at `1`. Making the added column `DEFAULT 1` is not sufficient on its own — MySQL applies the default to existing rows at `ALTER` time, but Postgres users who added the column by hand, or anyone whose table was created by an older `CREATE TABLE`, would still need the backfill. An alternative, more invasive option is to use `-1`/`UNVERSIONED` rather than `0` as the "absent" sentinel in `getVersioned()`.

Worth checking `JdbcAgentStateStore` too: it shares the `0`-means-absent logic (`new VersionedState<>(null, 0L)`, `expectedVersion == 0L` -> insert) but I could not find an `ensureVersionColumn()` equivalent, so an upgraded JDBC table may fail differently (unknown column).

**Error messages**

How the phantom conflict surfaces depends on `ConflictPolicy`, and two of the three are worse than a log line:

- `OVERWRITE` (the default) on current `main`: self-heals. The fallback `saveIfVersion(..., UNVERSIONED)` writes and bumps `0 -> 1`, so only the first turn is affected, plus a misleading warning:
  `agent_state CAS conflict — OVERWRITE applied (userId=..., sessionId=..., expectedVersion=0)`
- `OVERWRITE` on the released **2.0.3** jar: the turn fails outright, because that fallback is #3047 (fixed on `main` by #3048, not yet in a release):
  ```
  java.lang.RuntimeException: Failed to get versioned state: agent_state
    Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException:
      Cannot construct instance of `io.agentscope.core.state.State`
  ```
- `FAIL` and `APPEND_MERGE`, including on current `main`: **the session is permanently stuck.** `APPEND_MERGE` re-reads `latest.version()`, gets `0` again, calls `saveIfVersion(..., 0)`, hits the same duplicate key and throws `ConcurrentSessionModificationException`. Because nothing ever writes, the row stays at `0` and every subsequent turn throws the same way.

That last case is the reason this seems worth fixing separately from #3047: #3047 hides it behind a crash today, and once #3047 ships the remaining symptom is silent for the default policy but fatal for the other two.

**Environment**

- AgentScope-Java Version: 2.0.3 (root cause re-verified against `main`)
- Java Version: 21
- OS: macOS / Linux
- Store: `MysqlAgentStateStore` (MySQL 8), table created by hand rather than by `createIfNotExist` — though this is irrelevant: a table auto-created by 2.0.2 has no `version` column either and gets the same `ALTER`.

**Additional context**

Discovered on a production upgrade from 2.0.2 to 2.0.3: every pre-existing chat session threw once on its next turn and then recovered. Worked around with `UPDATE agentscope_sessions SET version = 1 WHERE state_key = 'agent_state' AND version = 0`. I scoped that update to a single `state_key` on purpose — list-valued states go through `insertItems`, which does not write the `version` column at all (so those rows are legitimately `0`), and they never take the CAS path.

Related: #3047 / #3048 (the `State.class` deserialization crash in the `UNVERSIONED` branch) is what makes this fatal on the 2.0.3 release; #3076 and #2146 also touch the same unconditional-write path.

hexly666:
…(#3162)

## Description

Close #3162.

`ensureVersionColumn()` backfilled the version column with `DEFAULT 0` — the same sentinel `getVersioned()` uses for "row absent" — so every pre-existing session took the INSERT branch in `saveIfVersion(..., 0)`, hit a duplicate key, and reported a phantom CAS conflict after upgrading. FAIL/APPEND_MERGE policies left the session permanently stuck.

`saveIfVersion(expectedVersion=0)` now falls back to `UPDATE ... WHERE version = 0` when the INSERT finds an existing row, so backfilled rows satisfy the CAS and self-heal to version 1 (MySQL, PostgreSQL, JDBC; the JDBC store guards the INSERT with a savepoint for vendors that abort the transaction on statement failure). The migration ALTER now uses `DEFAULT 1` so new upgrades no longer produce version-0 rows. Concurrency is unchanged: a row already past version 0 still correctly reports UNVERSIONED.

## AgentScope-Java Version

## Description

## Checklist


oss-maintainer (commented):
### Summary
Closes a real migration hole: `expectedVersion == 0` doubles as the "row absent" sentinel, so rows backfilled at version 0 by the old `ALTER TABLE ... DEFAULT 0` produced phantom CAS conflicts. The insert-conflict → `UPDATE ... WHERE version = 0` fallback fixes already-migrated deployments without a data migration, while new migrations use `DEFAULT 1`; the PostgreSQL savepoint handling is the right detail to get the fallback to survive at all, and the three-backend test set covers both the success and the genuine-conflict path. Left one non-blocking note inline; looks good to merge from a review standpoint.

### Non-blocking
- **[Info]** The same `getVersioned()` contract line ("an absent key reports `version == 0`") is read by every store implementation. Only the JDBC/MySQL/PostgreSQL trio gets the insert-conflict fallback here — worth confirming the Redis (and any other) `AgentStateStore` can't reach the same version-0 ambiguity, otherwise #3162 will resurface on another backend.

---

agentscope-extensions/agentscope-extensions-jdbc/src/main/java/io/agentscope/extensions/jdbc/state/JdbcAgentStateStore.java
```diff
@@ -265,29 +266,35 @@ public long saveIfVersion(
...
-                        } catch (SQLException e) {
-                            if (expectedVersion == 0L && isDuplicateKey(e)) {
+                        if (expectedVersion == 0L) {
+                            BoundSql insertSql =
+                                    dialect.sessionStateInsertIfAbsent(
+                                            slotId, key, SINGLE_STATE_INDEX, json);
+                            // Guard the INSERT with a savepoint: on vendors like Postgres a
+                            // failed statement aborts the whole transaction, which would
+                            // poison the fallback UPDATE below.
+                            Savepoint savepoint = conn.setSavepoint("cas_insert_if_absent");
```
oss-maintainer, line 276: Nice catch on the savepoint — without it the fallback UPDATE would abort on PostgreSQL. One residual edge: `rollback(savepoint)` only unwinds to *this* savepoint, so any savepoint set by an outer caller inside the same transaction after this one would also be released. Not reachable today (`executeInWriteTransaction` owns the transaction), but worth a one-line comment so a future nesting caller doesn't trip on it.

xiaoyuuuuuupeng (approved):
LGTM — I think this PR correctly fixes the phantom CAS conflict described in #3162.

The main fix looks sound:

* `expectedVersion == 0` first attempts the insert, then falls back to `UPDATE ... WHERE version = 0`.
* A row that has already advanced to version >= 1 is still protected by CAS and returns `UNVERSIONED`, so the fallback cannot overwrite a real concurrent update.
* The JDBC savepoint is important for databases such as PostgreSQL, where a failed INSERT would otherwise abort the transaction and prevent the fallback UPDATE.
* The dedicated PostgreSQL implementation avoids that issue with `ON CONFLICT DO NOTHING`.
* Changing the migration `ALTER` to `DEFAULT 1` prevents new upgrades from creating version-0 rows.

I don't see a correctness blocker here.

A few non-blocking follow-ups:

1. The updated `AgentStateStore` contract now allows an existing stored version `0` to CAS to version `1`. `MongoAgentStateStore`, however, currently handles `expectedVersion == 0` by matching only when the version field does not exist. If an explicit `version = 0` state exists, it would return `UNVERSIONED` instead. This does not appear reachable through the normal Mongo write path, so I don't think it blocks #3165, but the Mongo implementation or the interface wording could be aligned in a follow-up.

2. PostgreSQL has a regression test for the successful version-0 fallback. It would also be useful to add the symmetric case where the row has already advanced past version 0 and must still return `UNVERSIONED`.

3. `CREATE TABLE` still uses `DEFAULT 0` while the migration `ALTER` now uses `DEFAULT 1`. Normal single-state writes explicitly write version `1`, so this does not look blocking, but aligning the DDLs would make the schema semantics clearer.

jujn (approved):

PR #3169, opened 2026-09-16
fix(rag): escape WordReader Markdown table cells

### LINKED ISSUE #3168: [Bug]: WordReader produces invalid Markdown tables when cells contain pipes or line breaks
**Describe the bug**

WordReader with TableFormat.MARKDOWN inserts Word table cell text into Markdown rows without escaping it. A literal pipe adds a column; when it occurs in the header, the header and delimiter row have different column counts, so a GFM renderer does not recognize the table. A line break inside a cell also splits a table row. A backslash before a pipe changes the rendered cell text.

**To Reproduce**

1. Create a .docx table with two columns and two rows:
   - Header: A|B ; Path \| label
   - Data: 1|2 ; Line 1 followed by a line break and Line 2
2. Read it using WordReader(4096, SplitStrategy.CHARACTER, 0, false, true, TableFormat.MARKDOWN).
3. Inspect the returned Document content.

Actual Markdown:

    | A|B | Path \| label |
    | --- | --- |
    | 1|2 | Line 1
    Line 2 |

**Expected behavior**

The original two-column structure and the literal cell text should survive Markdown rendering:

    | A\|B | Path \\\| label |
    | --- | --- |
    | 1\|2 | Line 1<br>Line 2 |

**Error messages**

No exception is thrown; the returned Markdown is malformed.

**Environment**

- AgentScope-Java version: 2.0.3-SNAPSHOT (main commit 6c8ccc5c)
- Java version: 21.0.8
- OS: macOS 26.6.2

**Additional context**

The TableFormat.MARKDOWN Javadoc already notes that newlines may render incorrectly and suggests JSON. It does not document the pipe or backslash behavior. JSON preserves the original cell values but is a different requested output format. A focused fix can escape backslashes and pipes, render cell line breaks as <br>, and add a regression test through WordReader.read().

This is distinct from #2964 / #2965, which address boundaries between Word document paragraphs rather than text inside Markdown table cells.

tengjiaozhai:
## AgentScope-Java Version

2.0.3-SNAPSHOT (based on upstream/main at 6c8ccc5c)

## Description

WordReader inserted table cell text directly into Markdown rows. A literal pipe in a header cell could make the header and delimiter row have different column counts, preventing GFM from recognizing the table. Cell line breaks could split a row, and a backslash before a pipe could change the rendered text.

This change:

- Escapes backslashes and pipes in each header and data cell, then represents cell line breaks with <br>.
- Uses the same row formatter for the header and data rows.
- Updates the TableFormat.MARKDOWN Javadoc to describe the new line-break behavior.
- Adds a regression test that creates a .docx table and checks the public WordReader.read() output.

Closes #3168

## Verification

- Passed: targeted WordReaderTest, 12 tests.
- Passed: module and dependency reactor test command (mvn -pl agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple -am test).
- Passed: module Spotless apply and check; git diff --check.
- Passed: full repository test run with this machine's NO_PROXY and no_proxy unset (env -u NO_PROXY -u no_proxy mvn -B -T1 test).
- The initial run with the machine's default NO_PROXY stopped in KubernetesSerializationCompatTest because Fabric8 rejected a wildcard entry in that variable. The failing test and the full repository run passed after removing those variables for the command.

## Checklist

- [x] Code has been formatted with mvn spotless:apply for the changed module.
- [x] All tests are passing (mvn test, with the local NO_PROXY environment adjustment described above).
- [x] Javadoc comments are complete and follow project conventions.
- [x] Related documentation has been updated (TableFormat.MARKDOWN Javadoc).


agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/reader/WordReader.java
```diff
@@ -383,14 +381,25 @@ private String tableToMarkdown(List<List<String>> tableData) {
...
-            md.append(String.join(" | ", tableData.get(i)));
-            md.append(" |\n");
+            md.append(formatMarkdownTableRow(tableData.get(i)));
         }
 
         return md.toString();
     }
 
+    /** Formats a table row after escaping each cell's Markdown table delimiters. */
+    private String formatMarkdownTableRow(List<String> cells) {
```
oss-maintainer: **[Info]** Escape order is right — backslashes before pipes, so a literal `\|` in a cell cannot smuggle a column break, and the shared `formatMarkdownTableRow` removes the duplicated header/data formatting. One residual gap in the same area (pre-existing, not introduced here): `numCols` comes from `tableData.get(0).size()`, while `extractTableData` collects `row.getTableCells()` per row, so tables with merged/irregular rows still emit a ragged table that GFM will not render. Padding each row to `numCols` inside `tableToMarkdown` would close it — fine as a separate change.

agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/reader/WordReader.java
```diff
@@ -383,14 +381,25 @@ private String tableToMarkdown(List<List<String>> tableData) {
...
+        return "| "
+                + String.join(" | ", cells.stream().map(this::escapeMarkdownTableCell).toList())
+                + " |\n";
+    }
+
+    /** Keeps cell text on one Markdown table row while preserving literal pipes and backslashes. */
+    private String escapeMarkdownTableCell(String cell) {
+        String normalized = cell.replace("\r\n", "\n").replace('\r', '\n');
+        return normalized.replace("\\", "\\\\").replace("|", "\\|").replace("\n", "<br>");
```
oss-maintainer, line 407: **[Info]** `<br>` is the correct choice for rendered Markdown. Worth one thought before this ships widely: `WordReader` output is mostly consumed by the RAG chunking/embedding path rather than by a renderer, so every document with multi-paragraph cells now carries literal `<br>` tokens in the indexed text, and a `<br>` that was already present in the source text becomes indistinguishable from an injected one. If that turns out to matter for retrieval quality, making the substitution configurable (defaulting to `<br>`) keeps both consumers happy. Not a request for this PR.

oss-maintainer (approved):
### Summary
Escapes backslashes and literal pipes per cell and renders in-cell line breaks as `<br>`, so a `|` in a Word table cell no longer changes the column count and silently destroys the GFM table, and header and data rows now go through one formatter. The fix is correctly ordered, the `TableFormat.MARKDOWN` Javadoc is updated in step, and the regression test asserts the real `WordReader.read()` output on a generated `.docx` instead of the private helper — LGTM.

### Findings
- **[Info]** WordReader.java:391 — merged/irregular rows can still produce a ragged table (`numCols` is taken from row 0 only); pre-existing, fine to follow up separately
- **[Info]** WordReader.java:400 — `<br>` now lands in text that is usually indexed for RAG; only worth revisiting if it proves noisy for retrieval

### Notes
- Both points are informational; nothing here blocks merging.
- CI is green (`build`, `Check License`, `Check Module Sync`, `codecov/patch`) and the escaping behaviour is pinned by the new test, so future formatter changes cannot silently regress it.

---

agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/reader/WordReader.java
oss-maintainer: **[Info]** Escape order is right — backslashes before pipes, so a literal `\|` in a cell cannot smuggle a column break, and routing header and data rows through one formatter removes the duplicated builder code. One residual gap in the same area (pre-existing, not introduced here): `numCols` comes from `tableData.get(0).size()`, while `extractTableData` collects `row.getTableCells()` per row, so a table with merged or irregular rows still emits a ragged table that GFM will not render. Padding each row to `numCols` inside `tableToMarkdown` would close it — fine as a follow-up.

agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/reader/WordReader.java
oss-maintainer, line 407: **[Info]** `<br>` is the correct choice for rendered Markdown. Worth one thought before this ships widely: `WordReader` output is mostly consumed by the RAG chunking/embedding path rather than by a renderer, so every document with multi-paragraph cells now carries literal `<br>` tokens in the indexed text, and a `<br>` that was already present in the source text becomes indistinguishable from an injected one. If that turns out to matter for retrieval quality, making the substitution configurable (defaulting to `<br>`) serves both consumers. Not a request for this PR.

oss-maintainer (approved):
### Summary
Escapes backslashes and literal pipes per cell and renders in-cell line breaks as `<br>`, so a `|` inside a Word table cell no longer changes the column count and silently destroys the GFM table, and header and data rows now share one formatter. The escape order is correct, the `TableFormat.MARKDOWN` Javadoc is updated in step with the behaviour change, and the regression test asserts the public `WordReader.read()` output on a generated `.docx` rather than the private helper — LGTM.

### Findings
- **[Info]** WordReader.java:391 — merged/irregular rows can still produce a ragged table (`numCols` is taken from row 0 only); pre-existing, fine as a follow-up
- **[Info]** WordReader.java:400 — `<br>` now lands in text that is usually indexed for RAG rather than rendered; only worth revisiting if it proves noisy for retrieval

### Notes
- Both points are informational; nothing here blocks merging.
- CI is green (`build` ubuntu/windows, `Check License`, `Check Module Sync`, `codecov/patch`, `license/cla`) and the new test pins the escaping behaviour, so a future formatter change cannot silently regress it.

---

agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/reader/WordReader.java
tengjiaozhai: Addressed in 39d4470: `formatMarkdownTableRow` now pads each row to the header `numCols` with empty strings that still go through `escapeMarkdownTableCell`, so merged/irregular rows no longer emit a ragged GFM table. Added `testMarkdownTablePadsIrregularRows` to assert header/delimiter/data column counts match.

agentscope-extensions/agentscope-extensions-rag/agentscope-extensions-rag-simple/src/main/java/io/agentscope/core/rag/reader/WordReader.java
tengjiaozhai: Agreed on the tradeoff — keeping `<br>` as the default for rendered Markdown is right for this PR. `WordReader` currently has no builder/options surface (only fixed constructors), so making the line-break substitution configurable would mean a new public constructor parameter or introducing a Builder. Skipping that API redesign here; happy to follow up if retrieval quality shows the literal `<br>` tokens matter.

PR #3173, opened 2026-09-16
fix(wecom): propagate send rejections as stream errors

### LINKED ISSUE #3172: [Bug]: WeCom channel silently swallows outbound send failures
**Describe the bug**

Found while integrating the WeCom channel extension in a production agent deployment.

`WeComOutboundClient.checkSendResponse` (`agentscope-extensions-channel-wecom`, lines 125-141) treats a non-zero `errcode` in the response body as a log line only; token-expiry codes (42001/40014) additionally invalidate the cached token, but no error is propagated. Since WeCom returns HTTP 200 with `errcode` in the body, the `Mono<Void>` returned by `send(...)` completes as if the message had been delivered.

Rejected sends (invalid `touser`, content over limit, agent disabled, …) are therefore invisible to the caller: retry, alerting, and delivery-receipt semantics built on `onError` never fire — the failure only exists in server logs. By contrast, `FeishuOutboundClient.checkSendResponse` (`FeishuOutboundClient.java`, lines 133-146) throws `IllegalStateException` on `code != 0` after invalidating token codes, so the two channels expose inconsistent error semantics.

**To Reproduce**

Configure a WeCom channel and send a reply targeting an invalid/deleted `touser` (any body the API rejects with a non-zero `errcode`, still HTTP 200). `send()`'s `Mono` completes normally; only a `log.warn` appears in server logs.

**Expected behavior**

A rejected send propagates as `onError` (e.g. `IllegalStateException` carrying errcode/errmsg), consistent with the Feishu channel.

**Error messages**

None — the bug is exactly that no error surfaces. The only trace is `WeCom send returned errcode=..., errmsg=...` at WARN level.

- AgentScope-Java Version: 2.0.3-SNAPSHOT (source build from main @ 6c8ccc5; affected code unchanged at current main)
- Java Version: 17
- OS: Windows 11 (dev), Linux (production)

**Additional context**

A PR aligning `checkSendResponse` with the Feishu semantics is submitted and linked below.

Buktal:
## AgentScope-Java Version

2.0.3-SNAPSHOT (built from main)

## Description

Fixes #3172

WeCom reports delivery failures as HTTP 200 with a non-zero `errcode` in the body. `WeComOutboundClient.checkSendResponse` only logged such rejections, so the `Mono<Void>` returned by `send(...)` completed as if the message had been delivered — rejected sends (invalid `touser`, content over limit, …) were invisible to callers and any retry / alerting / receipt logic built on `onError` never fired.

This PR aligns the WeCom client with the existing `FeishuOutboundClient.checkSendResponse` semantics:

- an empty 200 response body, an unparseable body, a missing `errcode`, or a non-zero `errcode` → `IllegalStateException`, propagating as `onError` through the `send` stream
- token-expiry codes (42001/40014) still invalidate the cached token before failing, so the next send refetches a fresh token
- error messages embed at most a 200-char response-body snippet: the unparseable path sees gateway HTML pages or echoed request payloads rather than the expected small JSON, and the message travels to callers' `onError` handlers

No public API changes; everything else in the client is untouched.

Tests (`WeComOutboundClientTest`, new — the module had none — backed by a local `MockWebServer`):

- accepted send completes
- rejected send / unparseable body / missing `errcode` / empty body all error with `IllegalStateException`
- a 42001 rejection invalidates the cached token: the next send refetches (asserted via the request sequence and the fresh token in the outgoing request)
- long unparseable bodies are truncated in the error message

## Checklist

- [x]  All tests are passing (`mvn test` in `agentscope-extensions-channel-wecom`)
- [x]  Related documentation has been updated (no documentation describes the previous swallow-and-log behavior, so nothing to update)


agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComOutboundClient.java
```diff
@@ -122,21 +119,32 @@ private Mono<Void> post(String path, String token, Map<String, Object> body) {
                 .then();
```
oss-maintainer, line 124: **[Warning]** An empty 200 response still reports the message as delivered, which is the same class of bug this PR fixes.

`bodyToMono(String.class)` completes **empty** when WeCom (or an intermediate gateway) answers 200 with no body, so `doOnNext(this::checkSendResponse)` never runs and `.then()` completes normally — `send(...)` is `onComplete` for a response that carried no `errcode` at all. `sendErrorsWhenErrcodeMissing` covers `{}`, and `sendErrorsOnUnparseableResponseBody` covers HTML, but not the zero-length case.

Make the empty body flow into the new strict check instead:

```java
.bodyToMono(String.class)
.defaultIfEmpty("")
.timeout(Duration.ofSeconds(10))
.doOnNext(this::checkSendResponse)
.then();
```

`MAPPER.readTree("")` returns a `MissingNode`, so `!node.path("errcode").isIntegralNumber()` fires and the rejection propagates as intended. (Feishu has the same shape via `.map(...)`, so it likely wants the same treatment — out of scope here, just noting the parity argument cuts both ways.)

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComOutboundClient.java
```diff
@@ -122,21 +119,32 @@ private Mono<Void> post(String path, String token, Map<String, Object> body) {
...
-            if (errcode != 0) {
-                log.warn(
-                        "WeCom send returned errcode={}, errmsg={}",
-                        errcode,
-                        node.path("errmsg").asText());
-            }
+            node = MAPPER.readTree(body);
         } catch (Exception e) {
-            log.warn("Failed to parse WeCom send response: {}", e.getMessage());
+            throw new IllegalStateException("Invalid WeCom send response: " + body, e);
```
oss-maintainer: **[Warning]** The raw response body is interpolated into the exception message. This is the unparseable path, i.e. precisely the case where the payload is *not* the small `{"errcode":..,"errmsg":..}` JSON the code expects — it can be an HTML error page, a gateway trace, or a body echoing part of the request, and it now travels through `onError` into whatever the caller logs or surfaces (`WeComChannel.java:186` does `.doOnError(...).subscribe()`, and the AG-UI path can render stream errors to the end user).

Truncate to keep diagnostics without shipping arbitrary payloads downstream:

```java
        "Invalid WeCom send response: " + Truncate.truncate(body, 200), e);
```

The `errcode missing` branch at line 136 has the same exposure. Note also that the token-expiry invalidation at line 140 happens *before* the throw — correct order, and the test at `WeComOutboundClientTest.rejectedTokenExpirySendInvalidatesCachedToken` pins it well.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComOutboundClient.java
```diff
@@ -122,21 +119,32 @@ private Mono<Void> post(String path, String token, Map<String, Object> body) {
                 .then();
     }
 
+    /**
+     * Validates a WeCom send response body. WeCom reports delivery failures as HTTP 200 with a
+     * non-zero {@code errcode}; such rejections propagate as {@link IllegalStateException} so the
+     * caller's {@code send} stream terminates with {@code onError}. Token-expiry codes invalidate
+     * the cached token before failing.
+     */
     private void checkSendResponse(String body) {
```
oss-maintainer, line 133: **[Info]** Two consistency notes, both non-blocking.

1. Feishu's equivalent (`FeishuOutboundClient.java:133`) validates inside `.map(this::checkSendResponse)` and returns the body; this one keeps `void` + `.doOnNext(...)`. A `void` validator is arguably the better shape (nothing to return), so the nit is really on Feishu's side — but since the PR explicitly aims to align the two clients, aligning the operator as well saves the next reader a diff-vs-diff comparison.
2. The unused slf4j `log` field and imports are removed here. That is a net win (the failure is no longer swallowed), but it does mean the module now logs *nothing* about a WeCom rejection at the client layer; everything depends on the caller subscribing with an error handler. `WeComChannel.java:186` and `:225` do log, so the observable behaviour is fine — flagging only so it is a deliberate choice rather than a side effect.

oss-maintainer (commented):
### Summary

Aligns `WeComOutboundClient.checkSendResponse` with the Feishu client: a non-zero `errcode`, a missing `errcode`, or an unparseable body now terminates the `send` stream with `IllegalStateException` instead of a `log.warn`, and token-expiry codes still invalidate the cached token before failing. The diagnosis is right, the fix is minimal, and the new `MockWebServer`-backed test class is a real addition for a module that had none. Commented rather than approved for two things: an empty 200 body still looks like a successful delivery (the same invisibility this PR removes), and the raw response body is interpolated into the exception message that now travels to callers.

### Findings

- **[Warning]** `WeComOutboundClient.java:119` — `bodyToMono(String.class)` completes empty on a zero-length 200, so `doOnNext` never runs and `send(...)` still reports success; `defaultIfEmpty("")` routes it into the new strict check.
- **[Warning]** `WeComOutboundClient.java:133` (and `:136`) — raw response body interpolated into the exception message on the unparseable path, where the payload is by definition not the small JSON expected; it propagates through `onError` to `WeComChannel.java:186`/`:225` and onward to stream consumers.
- **[Info]** `WeComOutboundClient.java:128` — operator differs from `FeishuOutboundClient.java:133` (`.doOnNext` vs `.map`); the removal of the client-layer logger means rejections are only visible if the caller installs an error handler.

### Suggestions

```java
.bodyToMono(String.class)
.defaultIfEmpty("")
.timeout(Duration.ofSeconds(10))
.doOnNext(this::checkSendResponse)
.then();
```

and cap the body in the two message strings (e.g. first 200 chars). Both are one-line changes with a matching test case (`sendErrorsOnEmptyResponseBody`).

What is good here is worth keeping: the invalidation-before-throw ordering, the request-sequence assertion that proves the token is actually refetched, and the PR body's explicit "no public API changes" scoping.

---

oss-maintainer (approved):
### Summary

Delta re-review at `4c0a8712` (previous pass was `171c45ad`). Both open findings from the last round are addressed in this commit, and each comes with the test case that was asked for. Approving.

- `bodyToMono(String.class).defaultIfEmpty("")` now routes a zero-length `200` into `checkSendResponse`, so an unvalidated send no longer reports as delivered — covered by `sendErrorsOnEmptyResponseBody`.
- The raw-body interpolation is capped by `truncate` at `MAX_SNIPPET_LENGTH = 200` on both the unparseable and missing-`errcode` paths, and `sendErrorMessageTruncatesLongUnparseableBody` asserts the cap holds end to end.
- Dropping the client-layer `Logger` is the right call now that every rejection terminates the stream; `WeComChannel.java:186`/`:225` already keep a `doOnError` log, so rejections stay visible without a second warning per send.

`Check License`, `Check Module Sync`, `build (ubuntu-latest)`, `build (windows-latest)`, `codecov/patch` and `license/cla` are all green. `mergeStateStatus` is `BLOCKED` only for the missing approval, which this review supplies.

### Findings

None blocking. One nit, no change needed: `truncate` counts characters of the response body while the exception message also carries a fixed prefix, so the `length() < 300` test bound is tighter than `MAX_SNIPPET_LENGTH` alone; if the cap is ever raised the test bound has to move with it.

---

PR #3175, opened 2026-09-16
feat(channel): pluggable inbound-event deduplication and bot-loop-guard idle eviction

### LINKED ISSUE #3174: [Bug]: Channel webhook deduplication is JVM-local and non-pluggable; BotLoopGuard never evicts peer state
**Describe the bug**

Two related issues in `agentscope-extensions-channel-common`, found while integrating Feishu/WeCom/DingTalk channels in a multi-instance production deployment.

**1. `IdempotencyStore` is JVM-local and cannot be replaced**

`IdempotencyStore` is a `final` class backed by a `ConcurrentHashMap` in the JVM heap (`IdempotencyStore.java` L30, L34). Webhook platforms retry deliveries — WeCom and DingTalk redeliver callbacks when no timely response is sent, GitHub/GitLab redeliver webhook events — and four channels deduplicate through this store:

- `WeComCallbackController.java` L144 — `channel.idempotency().firstSeen(channelId + "|" + msgId)`
- `DingTalkChannel.java` L188
- `GitHubWebhookController.java` L108
- `GitLabWebhookController.java` L101

The seen-set lives in one process, with two consequences:

- With 2+ instances behind a load balancer, a platform retry that lands on a different instance is not recognized as a duplicate — the same message executes twice (duplicate agent runs, duplicate replies to the user).
- Even single-instance deployments lose protection across a restart: the seen-set is memory-only while the platform keeps retrying.

There is no application-level way out: the class is `final`, exposes no interface, and every channel's `fromProperties` hardcodes `new IdempotencyStore()` (`FeishuChannel.java` L109, `WeComChannel.java` L117, `DingTalkChannel.java` L105, `GitHubChannel.java` L104, `GitLabChannel.java` L98). Backing dedup with shared storage (Redis, DB) currently requires forking the framework.

The framework itself already treats dedup as shared-state territory on one channel: the Feishu callback controller never consults the in-process store and instead carries the comment "Durable intake owns deduplication. Marking an event seen before persistence would lose work when the control plane is temporarily unavailable." (`FeishuCallbackController.java` L145-146). The other four channels still rely on the process-local map.

**2. `BotLoopGuard` never evicts peer state**

`BotLoopGuard.states` (`ConcurrentHashMap<String, PeerState>`, `BotLoopGuard.java` L35) grows via `computeIfAbsent` in `allow()` (L60) and the class has no removal path. The per-peer event deque is trimmed to the sliding window (L66-68), but the map entry and its object live forever. On public deployments the peer set is unbounded (one key per DM user / group chat), so a long-running process grows monotonically.

Notably `IdempotencyStore` itself is bounded (TTL + max entries with eviction) — `BotLoopGuard` has neither bound.

**To Reproduce**

1. Multi-instance duplicates: run two instances of an app using the WeCom (or DingTalk/GitHub/GitLab) channel behind one callback URL; a platform retry that lands on the other instance is processed a second time. Restart variant on a single instance: restart the process while the platform is still retrying — redeliveries execute again.
2. Unbounded guard state: drive events from N distinct peer keys through `BotLoopGuard.allow()`; after the window and cooldown have fully elapsed, `states` still holds all N entries — no API or background path ever removes them.

**Expected behavior**

- Deduplication for channels that rely on it should hold across instances (and restarts) — or, at minimum, the framework should provide an official extension point so an application can back the store with shared storage.
- Idle `BotLoopGuard` peer state should be evicted, consistent with `IdempotencyStore`'s bounded design.

**Error messages**

None — both issues are silent. Duplicate executions and slow memory growth produce no errors.

- Java Version: 17
- OS: Windows (production deployment: Linux, multiple instances)

**Additional context**

Suggested direction — happy to contribute:

- Extract the dedup role into an interface (a single `boolean firstSeen(String key)` method suffices), keep the current in-memory implementation as the default so existing behavior is unchanged, and let channel `fromProperties` accept an injected store. The full channel constructors already take the store as a constructor parameter; only the `fromProperties` path hardcodes it.
- Add idle eviction to `BotLoopGuard`: drop peers with no events for longer than the window once any cooldown has expired — such a peer's state is informationally empty, so eviction is behavior-preserving.
- For what it is worth, we run a Redis-backed implementation in production (SETNX, key = `channel:botId:platformMessageId`) and would be glad to contribute it as a reference implementation if there is interest.

Buktal:
## AgentScope-Java Version

2.0.3-SNAPSHOT (main @ 6c8ccc5c)

## Description

Fixes #3174.

Two related changes in the channel extensions. Both are additive — no existing public signature or default behavior changes.

**1. Pluggable inbound-event deduplication**

- New SPI `InboundEventDeduplicator` in `agentscope-extensions-channel-common`: a single `boolean firstSeen(String key)` contract for dropping platform redeliveries (WeCom/DingTalk redeliver callbacks when no timely acknowledgment is sent; GitHub/GitLab redeliver webhook events). The Javadoc states the contract, the thread-safety requirement for implementations, and the multi-instance guidance (back it with shared storage, e.g. a Redis `SET ... NX EX` keyed by channel and platform event id).
- `IdempotencyStore` now implements the interface and remains the default process-local implementation. Its Javadoc claim of insertion-order eviction is corrected to match the actual arbitrary-victim behavior, and its expiry path now restarts the key's retention window so only the first redelivery observed after the TTL lapses proceeds — previously every post-expiry redelivery was treated as first-seen (review finding).
- All five channels (`feishu`/`wecom`/`dingtalk`/`github`/`gitlab`) gain a `fromProperties(channelId, routing, rawProperties, InboundEventDeduplicator)` overload so an application can inject a shared-storage deduplicator. The existing 3-arg factories delegate with the default store, so existing wiring is untouched. Note: the Feishu callback controller currently delegates deduplication to the durable intake, so the injected store is held but not consulted there — this is documented on the overload.

**2. `BotLoopGuard` idle-peer eviction**

- Peer state was retained forever (`states` had no removal path), growing monotonically on public deployments with an unbounded peer set (one key per DM user / group).
- Peers with no recorded event for longer than `windowMillis + cooldownMillis` are now evicted — by that point the sliding window is empty and any cooldown has expired, so the state is informationally empty.
- Sweeping is opportunistic with no background thread: the first `allow()` call after each `windowMillis + cooldownMillis` period sweeps; `evictIdlePeers()` exposes the same sweep explicitly and `trackedPeers()` is a small observability helper.
- The change to `BotLoopGuard` is additive except for one initializer line: a newly created `PeerState` is stamped with its creation time, so a fresh entry can never be mistaken for idle and evicted mid-acquisition (review finding). An evicted peer simply restarts with a fresh window on its next event; by eviction time its window is already empty and any cooldown has expired.

**How to test**

- New `agentscope-extensions-channel-common` test suite (11 tests): dedup first-observation/redelivery/null-key/TTL-expiry (including the window restart: the first post-expiry redelivery is accepted, an immediate second is dropped)/size-bound; guard budget/cooldown/null-peer; eviction timing (idle peer evicted after window+cooldown, active and in-cooldown peers kept, sweep fires on the first call past the interval).
- `mvn test` across all six touched modules passes.

## Checklist

- [x]  Related documentation has been updated (Javadoc on the new SPI and overloads; these classes have no separate docs pages)


agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/IdempotencyStore.java
```diff
@@ -53,6 +57,7 @@ public IdempotencyStore(long ttlMillis, int maxEntries) {
      * Records {@code key} as seen. Returns {@code true} when this is the first time {@code key} is
      * seen (the caller should proceed), {@code false} when it has already been seen within the TTL.
      */
+    @Override
     public boolean firstSeen(String key) {
```
oss-maintainer, line 63: **[Warning]** The post-expiry branch returns `true` without writing the new timestamp, so a key that has aged past `ttlMillis` is reported as first-seen **every** time it is redelivered, not just the first time.

Trace (default TTL 5m): `firstSeen(k)` at t=0 stores `k -> 0`. Two redeliveries at t=301s and t=302s both hit `prior = 0`, both compute `now - 0 > 300_000`, both return `true` — so both are processed, which contradicts the contract the new `InboundEventDeduplicator` Javadoc states ("each redelivery must execute at most once"). The stale entry also keeps the map from ever reclaiming it, since `sweep()` only runs when the map is already at `maxEntries`.

Refresh on the expiry path so only the first post-expiry delivery proceeds:

```java
Long prior = seen.putIfAbsent(key, now);
if (prior == null) {
    return true;
}
if (now - prior > ttlMillis) {
    seen.put(key, now);   // restart the retention window for this key
    return true;
}
return false;
```

This is pre-existing code, but the PR turns it into a public SPI contract and adds the first tests for it, so it seems like the right moment to pin the semantics down. `IdempotencyStoreTest` has a slot ready: assert that after the TTL expires, the *first* redelivery is accepted and an immediate second one is dropped.

oss-maintainer (commented):
### Summary

Two additive changes in the channel extensions: a new `InboundEventDeduplicator` SPI with a 4-arg `fromProperties` overload on all five channels so a shared-storage deduplicator can replace the process-local `IdempotencyStore`, plus opportunistic idle-peer eviction in `BotLoopGuard` so per-peer state stops accumulating forever. The design intent is sound — the existing 3-arg factories delegate to the default store, `IdempotencyStore` keeps its behaviour while implementing the new interface, and the Javadoc is honest about the multi-instance limits. Commented rather than approved: the SPI's headline use case (cross-instance dedup) is not reachable from the `ChannelFactory` seam the harness actually uses, one eviction race has a one-line fix, and `IdempotencyStore`'s post-TTL branch does not honour the "at most once" contract the new interface now states.

### Findings

- **[Warning]** `IdempotencyStore.java:61` — the `now - prior > ttlMillis` branch returns `true` without writing the new timestamp, so every subsequent redelivery of the same key is also treated as first-seen. Pre-existing, but this PR promotes that method into a documented SPI contract ("each redelivery must execute at most once") and adds its first tests.
- **[Warning]** `BotLoopGuard.java:71` — a freshly created `PeerState` has `lastEventMs == 0`, so `evictIdlePeers` sees it as idle and `states.remove(key, state)` succeeds against the exact object another thread is about to record into; the Javadoc's "rare interleaving" is reachable whenever a new peer appears during a sweep.
- **[Warning]** `InboundEventDeduplicator.java:28` — the "inject it via the channel factories" guidance does not match `ChannelFactory#create(channelId, routing, properties)`, which has no deduplicator slot; config-driven deployments still get the process-local store.
- **[Info]** `InboundEventDeduplicator.java:22` — caller-composed keys (`channelId|platformEventId`) need a deployment-scoped prefix once backed by shared storage, or a collision drops live traffic silently.
- **[Info]** `FeishuChannel.java:108` — the 4-arg overload is a documented no-op for Feishu, with an identical signature to the four channels where it works.
- **[Info]** `BotLoopGuard.java:98` — the sweep only runs while traffic arrives, and the eviction count is discarded, so there is no operator-visible signal for the leak this fixes.
- **[Info]** `BotLoopGuardTest.java:63` — eviction tests sleep 60–200 ms against thresholds of the same order; an injectable clock would make them deterministic.

### Suggestions

The `BotLoopGuard` race collapses to a constructor change:

```java
PeerState state = states.computeIfAbsent(peerKey, k -> new PeerState(now));
```

With `lastEventMs` stamped at creation, `now - state.lastEventMs > idleThreshold` is false for a brand-new entry, the conditional remove can no longer drop an in-flight acquisition, and the documented caveat can go away rather than be accepted.

For the SPI gap, either a `default Channel create(String, ChannelConfig, Map, InboundEventDeduplicator)` on `ChannelFactory` that existing implementations ignore, or an explicit note that injection is programmatic today plus a follow-up issue. Either is fine — what should not happen is a Javadoc that reads as available configuration.

Good parts worth keeping: the additive overload strategy (no signature churn for existing callers), the corrected `IdempotencyStore` eviction Javadoc, and the 11 tests covering both the store and the guard's new paths.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/BotLoopGuard.java
```diff
@@ -57,6 +67,7 @@ public boolean allow(String peerKey) {
             return true;
         }
         long now = System.currentTimeMillis();
+        maybeSweep(now);
         PeerState state = states.computeIfAbsent(peerKey, k -> new PeerState());
```
oss-maintainer: **[Warning]** A brand-new `PeerState` has `lastEventMs == 0`, so the very first event of a peer is always "idle" as far as `evictIdlePeers` is concerned. The window is not really "rare": the creator thread is between `computeIfAbsent` (line 71) and `synchronized (state)` (line 72) precisely when a concurrent `allow()` on another thread runs its opportunistic sweep (line 70), and the sweep's `states.remove(key, state)` succeeds because the map still holds that exact object.

Result: the fresh entry is dropped, the creator records into an orphaned object, and the peer's budget restarts on the next event — i.e. the throttle can be skipped by a peer that happens to appear during a sweep, rather than only being "momentarily relaxed".

One-line fix that makes the conditional-remove guard actually hold — stamp the state at creation so it can never look idle:

```java
PeerState state = states.computeIfAbsent(peerKey, k -> new PeerState(now));
...
private static final class PeerState {
    final Deque<Long> events = new ArrayDeque<>();
    final long lastEventMs0;      // or: long lastEventMs set in the constructor
    long cooldownUntilMs;
    PeerState(long now) { this.lastEventMs = now; }
}
```

With `lastEventMs` initialised to `now`, the `now - state.lastEventMs > idleThreshold` test is false for an entry that was just created, the eviction/creation interleaving collapses, and the Javadoc caveat about the "rare concurrent interleaving" can be dropped.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/BotLoopGuard.java
```diff
@@ -69,13 +80,27 @@ public boolean allow(String peerKey) {
...
         }
     }
 
+    /**
+     * Sweeps idle peers piggybacked on {@link #allow(String)} calls: the first call after each
+     * {@code windowMillis + cooldownMillis} period sweeps. No background thread is started;
+     * concurrent callers may sweep redundantly, which is harmless.
+     */
+    private void maybeSweep(long now) {
+        if (now - lastSweepMs >= windowMillis + cooldownMillis) {
```
oss-maintainer, line 98: **[Info]** Two small observations on the opportunistic sweep:

1. **It never runs without traffic.** `allow()` is the only trigger, so a channel that goes quiet keeps its whole `states` map until the next inbound event arrives. For the "long-running public deployment" motivation in the Javadoc that is usually fine (traffic resumes), but it means the reclamation guarantee is "within one period *of observed traffic*", not one wall-clock period. A one-line Javadoc clarification would be enough.
2. **No trace of what it reclaimed.** The return value of `evictIdlePeers()` is dropped in `maybeSweep`. For a memory-growth fix, a `log.debug("evicted {} idle peers ({} tracked)", ...)` (or exposing the count via the existing observability helpers) makes a leak regression diagnosable instead of invisible. `BotLoopGuard` currently has no logger at all, so this is optional — just flagging that the operator-facing signal for this change is otherwise absent.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/InboundEventDeduplicator.java
```diff
@@ -0,0 +1,48 @@
...
+ * Deduplicates inbound platform events across redeliveries.
+ *
+ * <p>Channel adapters consult this before processing an inbound event: delivery platforms (WeCom,
+ * DingTalk, GitHub, GitLab, ...) redeliver the same event when no timely acknowledgment is
+ * received, and each redelivery must execute at most once.
+ *
+ * <p>Keys are opaque and caller-composed — typically {@code channelId + "|" + platformEventId}.
+ *
+ * <p>The default {@link IdempotencyStore} keeps state in the JVM heap, which only deduplicates
+ * within a single process: a redelivery that lands on another instance, or after a restart, is not
```
oss-maintainer, line 31: **[Warning]** "...inject it via the channel factories" is not reachable from the SPI that the harness actually uses. `io.agentscope.harness.agent.gateway.channel.ChannelFactory#create(channelId, routing, properties)` has no deduplicator slot, and the factory-method references in the Javadoc of the 3-arg overloads point at that interface. So a deployment configured through `agentscope.json` (the normal path — `ChannelFactory` resolves `channelClass` and calls the 3-arg overload) still gets a process-local `IdempotencyStore`, and only hand-wired bootstraps like `CodingBootstrap` can inject a shared one.

Not a blocker for merging the additive API, but it would be good to either close the loop here or say so explicitly:

- add a deduplicator-provider seam to `ChannelFactory` (e.g. a default `create(channelId, routing, properties, InboundEventDeduplicator)` overload that the existing implementations ignore), or
- state in this Javadoc that the injection point today is the programmatic `fromProperties(...)` overload and file a follow-up for the config-driven path.

Otherwise the multi-instance guidance reads as available configuration when it is currently only available to code that builds channels by hand.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/InboundEventDeduplicator.java
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.agentscope.extensions.channel.common;
+
+/**
+ * Deduplicates inbound platform events across redeliveries.
+ *
+ * <p>Channel adapters consult this before processing an inbound event: delivery platforms (WeCom,
+ * DingTalk, GitHub, GitLab, ...) redeliver the same event when no timely acknowledgment is
```
oss-maintainer, line 22: **[Info]** Worth one more sentence about key namespace, since the point of the SPI is shared storage. Keys are caller-composed `channelId + "|" + platformEventId`; platform event ids are not globally unique across *deployments* that share one Redis (staging + production on the same instance, or several environments), and a collision silently drops live traffic rather than duplicating it. Suggest the Javadoc recommend a deployment-scoped prefix (e.g. `env + "|" + channelId + "|" + eventId`) or note that the implementation owns namespacing. Cheap to state, expensive to debug later.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuChannel.java
```diff
@@ -89,9 +90,30 @@ private FeishuChannel(
...
             String channelId, ChannelConfig routing, Map<String, Object> rawProperties) {
+        return fromProperties(channelId, routing, rawProperties, new IdempotencyStore());
+    }
+
+    /**
+     * Factory variant that lets the application supply the {@link InboundEventDeduplicator} used
+     * to drop platform redeliveries — for example a shared-storage implementation so duplicates
+     * are recognized across instances. The process-local {@link IdempotencyStore} is used
+     * otherwise. Note the Feishu callback controller currently delegates deduplication to the
+     * durable intake, so the supplied store is held for channels that consult it.
```
oss-maintainer, line 108: **[Info]** The 4-arg overload is disclosed as a no-op for Feishu, which is honest, but the parameter is easy to pass by mistake: the signature is identical to the other four channels and `Objects.requireNonNull(idempotency, ...)` in the constructor accepts a Redis-backed store that is then never consulted. Options that keep the API shape but remove the foot-gun:

- drop the overload for Feishu until the controller consults the store (callers then get a compile error instead of a silent no-op), or
- keep it and emit a `log.warn` at construction when the supplied deduplicator is not the default `IdempotencyStore`.

Fine to leave as is if the team prefers API symmetry across the five channels — the Javadoc does say it out loud.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/test/java/io/agentscope/extensions/channel/common/BotLoopGuardTest.java
```diff
@@ -0,0 +1,98 @@
...
+    @Test
+    void cooldownStateIsNotEvictedWhileActive() {
+        BotLoopGuard guard = new BotLoopGuard(1, 60_000L, 60_000L);
+        assertTrue(guard.allow("peer-1"));
+        assertFalse(guard.allow("peer-1")); // trips the cooldown
+        assertEquals(0, guard.evictIdlePeers()); // idle threshold is 120s away
+        assertTrue(guard.isCoolingDown("peer-1"));
+    }
+
+    @Test
```
oss-maintainer, line 63: **[Info]** `idlePeersAreEvictedAfterWindowPlusCooldown` (160ms threshold) and `sweepRunsOnTheFirstCallAfterTheSweepInterval` sleep 60–200ms against the same order of magnitude. The assertions are directional so they are not flaky in the obvious way, but on a loaded `windows-latest` runner a >160ms scheduling hiccup between `allow("stale-peer")` and `allow("busy-peer")` changes *which* peers exist, and a 20ms stall can flip `assertEquals(0, guard.evictIdlePeers())` in `cooldownStateIsNotEvictedWhileActive`-style timing checks.

Since the guard already takes its bounds through the constructor, injecting the clock the same way (a `LongSupplier`/`Supplier<Long>` `now` seam with a `System::currentTimeMillis` default) would make both eviction tests deterministic and let the fresh-entry interleaving above be tested without threads. Reasonable as a follow-up rather than a requirement for this PR.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/IdempotencyStore.java
Buktal: Confirmed — good catch, and agreed this is the right moment to pin the semantics. Fixed in b58f2637: the expiry path now writes `seen.put(key, now)` before returning `true`, so the retention window restarts and only the first post-expiry redelivery proceeds. `entriesExpireAfterTtl` now asserts exactly this: after the TTL lapses the first redelivery is accepted and an immediate second is dropped. Noted in the PR description that this tightens pre-existing behavior, as the SPI contract requires.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/BotLoopGuard.java
Buktal: Confirmed for the freshly-created-entry variant — `lastEventMs == 0` made a brand-new state look epoch-idle, so the conditional remove could drop an in-flight acquisition. Fixed in b58f2637 with the constructor stamp you suggested (`new PeerState(now)`), so a fresh entry can never satisfy the idle check.

One nuance on "the Javadoc caveat can go away": it is narrowed, not removed. A *genuinely idle* entry can still be evicted between another caller's acquisition and its recording, and closing that would require re-acquiring/re-validating under the map, which this heuristic throttle deliberately avoids. The effect is bounded either way — the lost state was already informationally empty, so the budget relaxes by at most one event. The Javadoc now states both facts precisely.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/BotLoopGuard.java
Buktal: Agreed on both points; kept the code as-is for now. On (1), the class Javadoc already ties sweeping to the first `allow()` after each period, so reclamation is traffic-driven rather than wall-clock — happy to make that more explicit if maintainers want. On (2), this module is currently dependency-free (JDK-only imports, no slf4j in the pom), and adding a logger for a single debug line did not seem worth expanding the pom for. `evictIdlePeers()`'s return value and `trackedPeers()` are the observability seams; logging can ride along when the module grows a logger dependency for a real need.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/InboundEventDeduplicator.java
Buktal: You're right — reworded in b58f2637: the Javadoc now says injection today is programmatic (the four-argument `fromProperties` overloads for hand-wired setups), and that `agentscope.json` deployments get the default process-local store. Deliberately did not widen `ChannelFactory` in this PR: it is a `@FunctionalInterface` today, and adding a deduplicator slot is an SPI-shape decision we would rather leave to maintainers. Happy to follow up with the overload if config-driven injection is wanted.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/InboundEventDeduplicator.java
Buktal: Added in b58f2637: the keys paragraph now tells shared-storage implementations to namespace keys per deployment (environment or application prefix), noting that a cross-deployment collision silently drops live traffic.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuChannel.java
Buktal: Keeping the overload for signature symmetry across the five channels, with the non-use documented on it — a divergent factory surface across sibling channels seemed worse for discoverability than the documented no-op. Agreed a compile error would fail louder; happy to drop it if maintainers prefer that trade-off.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/test/java/io/agentscope/extensions/channel/common/BotLoopGuardTest.java
Buktal: Deferred as you suggest: an injectable clock means touching the public constructors of both classes, and the current assertions are directional. Will fold in a `LongSupplier` seam if these tests ever show flakiness on CI.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/IdempotencyStore.java
```diff
@@ -63,7 +70,12 @@ public boolean firstSeen(String key) {
         if (prior == null) {
             return true;
         }
-        return now - prior > ttlMillis;
+        if (now - prior > ttlMillis) {
+            // Restart the retention window so only the first post-expiry redelivery proceeds.
+            seen.put(key, now);
```
oss-maintainer: Restarting the retention window here is the right fix for the post-TTL leak, but `get`-style check plus `put` is not atomic: two threads that both read an expired `prior` both take this branch and both return `true`, so a redelivery can execute twice — exactly what the new `InboundEventDeduplicator` contract ("each redelivery must execute at most once") rules out. Make the restart a CAS so only one winner proceeds:



A single-threaded test cannot catch the interleaving, so it is fine to leave the test as is and just tighten the primitive.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/BotLoopGuard.java
```diff
@@ -87,8 +112,50 @@ public boolean isCoolingDown(String peerKey) {
...
+        }
+        return removed;
+    }
+
     private static final class PeerState {
         final Deque<Long> events = new ArrayDeque<>();
         long cooldownUntilMs;
+        long lastEventMs;
+
+        PeerState(long now) {
```
oss-maintainer, line 155: Stamping `lastEventMs` at creation closes the eviction race properly — a brand-new entry can no longer look idle. With that in place the "rare concurrent interleaving" caveat in the `evictIdlePeers` Javadoc above now describes an unreachable state; it could be dropped (or narrowed to "an entry evicted after its idle threshold expires simply restarts its window") rather than documenting a relaxation that no longer exists.

oss-maintainer (commented):
### Summary

Re-review of the fix round on top of my previous findings. All three Warnings are addressed and the fixes are the right shape, not band-aids: `IdempotencyStore` now restarts the retention window on an accepted post-TTL observation, `BotLoopGuard` stamps `lastEventMs` at creation so a brand-new peer can never be mistaken for idle, and the `InboundEventDeduplicator` Javadoc now states plainly that injection is programmatic today while the config-driven path keeps the process-local store — which is the honest version of the claim I objected to. The deployment-namespacing `[Info]` is documented too. One Warning remains, in the new restart path itself; it is a two-line change, so I am commenting rather than approving.

### Findings

- **[Warning]** `IdempotencyStore.java:75` — the expired-key restart is a read-then-write, so two concurrent redeliveries of the same key can both take the branch and both return `true`, executing a delivery twice under exactly the contract this PR just wrote down. `seen.replace(key, prior, now)` gives the same semantics with a single winner.
- **[Info]** `BotLoopGuard.java:156` — with the creation timestamp stamped, the "rare concurrent interleaving" caveat in the `evictIdlePeers` Javadoc now documents an unreachable case; worth trimming so the doc does not describe a relaxation the code no longer permits.

### Suggestions

```java
if (now - prior > ttlMillis) {
    // Only the CAS winner restarts the window; the loser sees a redelivery.
    return seen.replace(key, prior, now);
}
return false;
```

Good parts worth keeping: the `PeerState(now)` constructor is the minimal correct fix and the added assertion in `entriesExpireAfterTtl` pins the new restart behaviour; the pom cleanups folded in from main are unrelated and fine. Once the CAS restart lands, this is good to go.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/IdempotencyStore.java
oss-maintainer: Code snippet dropped by my posting script above — the intended change is:

```java
if (now - prior > ttlMillis) {
    // Only the thread that wins the CAS restarts the window; the loser sees a redelivery.
    return seen.replace(key, prior, now);
}
return false;
```

Same behaviour for the single-threaded path, but concurrent redeliveries of an expired key can no longer both proceed.

Buktal: Done in 1db70687 — the restart is now `return seen.replace(key, prior, now)`, so only the CAS winner restarts the window and concurrent redeliveries of an expired key can no longer both proceed. Test left as-is per your note.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/BotLoopGuard.java
Buktal: Trimmed in 1db70687.

One precision note, so the history is accurate: the interleaving the caveat described is still reachable — it was the *other* variant (a genuinely idle entry acquired by one thread and evicted by another), which the creation stamp does not close; only the fresh-entry variant is gone. Its effect is exactly "the entry restarts its window" though — by eviction time the entry's window is already empty, so at most one in-flight event escapes counting — so the Javadoc now states that as plain behavior in your wording instead of framing it as a tolerated race, which reads better anyway.

oss-maintainer (approved):
### Summary

Re-reviewed the delta since my last pass (`032f55da` → `1db70687` → `efc8ee25`); the later commits are merge-from-main only. The new non-merge commit addresses both correctness issues I raised previously, and the fix is the stronger of the two options. LGTM.

### Previously raised — now resolved

- **`IdempotencyStore.firstSeen` post-expiry window** — the expiry branch now returns `seen.replace(key, prior, now)` instead of an unconditional `put` + `true`, so only the CAS winner restarts the retention window and concurrent post-expiry redeliveries cannot both proceed. `IdempotencyStoreTest.entriesExpireAfterTtl` pins the contract (accept after TTL, drop the immediate second call).
- **`BotLoopGuard` fresh-entry eviction** — `PeerState` is now stamped at creation (`computeIfAbsent(peerKey, k -> new PeerState(now))`) and `evictIdlePeers` keeps the conditional `states.remove(key, state)`, so a newly created entry can no longer look idle. The Javadoc caveat about the "rare concurrent interleaving" was dropped accordingly, which matches the code.

### Residual notes (non-blocking)

- **[Info]** `IdempotencyStore.firstSeen` — if `sweep()` evicts `key` between this thread's `putIfAbsent` and its `replace`, the CAS fails and the event is reported as a redelivery (`false`) even though the key is no longer in the map. It only happens at the `maxEntries` boundary, where dedup is already approximate, so a comment acknowledging "CAS failure under cap pressure is treated as redelivery" would be enough — no code change needed.
- **[Info]** `BotLoopGuard.maybeSweep` — the check-then-act on the `volatile lastSweepMs` still lets two callers both sweep (documented as harmless), and the losing writer can roll the timestamp back to a slightly older `now`. Both are benign for a heuristic throttle; noting it in case this class ever grows a stricter budget guarantee.
- **[Info]** From the earlier pass, still open and optional: `evictIdlePeers()` returns the reclaimed count but `maybeSweep` discards it, so there is no operator-visible signal if reclamation stops working.

### Verification

- CLA: `license/cla` = success on `efc8ee25`.
- Diff scope: 3 files in `channel-common` (2 main, 1 Javadoc) plus per-channel wiring; the DingTalk callback additions in this PR's range are merge-from-main content, not part of this delta.

---

PR #3177, opened 2026-09-16
feat(dingtalk): add HTTP callback receive mode alongside Stream

### LINKED ISSUE #3176: [Feature]: Add DingTalk HTTP callback receive mode alongside Stream

We are integrating DingTalk into a multi-tenant product where each tenant configures their own enterprise-internal robot: credentials are stored server-side, and robots are added, removed, and re-configured at runtime. Verified against `main` @ `6c8ccc5c`, the dingtalk channel module can only receive messages through the Stream protocol:

- `DingTalkStreamClient` opens one long-lived WebSocket per `DingTalkChannelProperties` — registration, heartbeat, and reconnect with exponential backoff are all self-managed inside the client, and the credentials are bound for the lifetime of the connection (`DingTalkStreamClient.java`, class javadoc lines 40–57, `properties` field and constructor).
- `DingTalkChannel` creates and owns that connection at construction time (`new DingTalkStreamClient(properties, this::onInboundPayload)` in the constructor; `fromProperties` wires exactly one fixed credential set).

This fits a single-tenant deployment with static credentials. For N tenant-configured robots it forces the application to manage a pool of WebSocket lifecycles itself: connect on tenant onboarding, tear down on offboarding, re-register on credential rotation, and monitor liveness for all of them — a connection-management subsystem per application, just to receive messages.

DingTalk offers two reception modes for robots: **Stream** and **HTTP callback** ([official overview](https://open.dingtalk.com/document/orgapp/robot-receive-message)). In HTTP mode DingTalk POSTs each message to a public endpoint with `timestamp`/`sign` headers, optionally AES-encrypted with an app-configured `aes_key` — operationally the same shape the framework already handles in the other channels: `FeishuCallbackController`, `WeComCallbackController`, `GitHubWebhookController`, and `GitLabWebhookController` all exist. DingTalk is the only one of the five channels with no HTTP receive path.

An opt-in HTTP callback receive mode for the dingtalk channel, mirroring the shape of the WeCom callback controller:

1. A `DingTalkCallbackController` that dispatches by `channelId` from the URL path via a channel registry (the same pattern the WeCom controller uses).
2. Signature verification before any processing: DingTalk signs each request as `Base64(HmacSHA256(appSecret, timestamp + "\n" + appSecret))` and sends `timestamp`/`sign` as request headers; reject on signature mismatch or a stale timestamp (the [verification doc](https://open.dingtalk.com/document/robots/verify-valid-requests-for-intra-enterprise-robot-group-chat) specifies a 1-hour freshness window and ships a Java sample).
3. Optional decryption: when the robot is configured with an `aes_key` (43-char Base64), the POST body is a `{"encrypt": "..."}` envelope using DingTalk's callback encryption protocol (AES-CBC, PKCS7-padded — the same envelope family as WeCom's EncodingAESKey scheme); decrypt before parsing ([protocol doc](https://open.dingtalk.com/document/development/callback-event-message-body-encryption-and-decryption), [official crypto library](https://github.com/open-dingtalk/DingTalk-Callback-Crypto)).
4. Reuse everything downstream: the HTTP-mode POST body is the same bot-message object DingTalk documents for robot reception (`msgtype`, `conversationId`, `conversationType`, `senderStaffId`, `msgId`, ...), so the existing `DingTalkInboundMapper` covers the mapping — at most a small normalization is needed for field-placement differences between the two modes. The existing msgId deduplication, bot-loop guard, and dispatch pipeline apply unchanged, and replies keep going through the existing `DingTalkOutboundClient`, which is reception-mode independent.
5. Configuration through `DingTalkChannelProperties` (e.g. a `mode=stream|http` switch plus `aesKey`), with Stream remaining the default so existing setups are unaffected.

- **Stream-only, plus a dynamic credential registration API on `DingTalkStreamClient`** (register/unregister connections at runtime): keeps the receive path centralized, but moves the whole connection-pool lifecycle problem — per-bot connect/teardown, reconnect on rotation, liveness monitoring — into every multi-tenant application, each re-implementing it against a bespoke API.
- **Application-level HTTP controller outside the framework**: feasible today because `DingTalkInboundMapper` and `DingTalkOutboundClient` are reusable classes, but each application re-wires signature verification, decryption, deduplication, and dispatch itself, and the channel integration forks from the framework.

**Additional context**

- DingTalk positions Stream as the recommended mode for new apps (the official callback crypto repository itself notes the older callback scheme is more involved), so Stream should remain the default. The HTTP mode targets deployments where it genuinely fits: multi-tenant dynamic credentials, or operations that already expose public callback endpoints for the other four channels and prefer a uniform ingress.
- Reception modes overview: https://open.dingtalk.com/document/orgapp/robot-receive-message
- HTTP-mode message protocol (headers incl. `timestamp`/`sign`): https://open.dingtalk.com/document/robots/receive-message
- Request signature verification: https://open.dingtalk.com/document/robots/verify-valid-requests-for-intra-enterprise-robot-group-chat
- Callback body encryption protocol: https://open.dingtalk.com/document/development/callback-event-message-body-encryption-and-decryption
- Official reference crypto library (Java/Go/Python/C#): https://github.com/open-dingtalk/DingTalk-Callback-Crypto

Buktal:
## AgentScope-Java Version

2.0.3-SNAPSHOT (main @ 6c8ccc5c)

## Description

Adds an opt-in HTTP callback reception mode to the DingTalk channel, alongside the existing Stream (WebSocket) mode, mirroring the shape of the WeCom/Feishu callback controllers. Fixes #3176.

**What was added**

- `DingTalkCallbackController` — `POST /api/channels/dingtalk/{channelId}/callback`, dispatching by `channelId` via `DingTalkChannelRegistry` (the WeCom controller pattern). Unknown channel or unverified request → 401 (indistinguishable, so the endpoint is not a channelId oracle; the unknown-id path logs at debug), unusable body → 400 (including plain bodies on a channel configured with an `aesKey`, and bodies over 64 KB, both rejected before parsing/dispatch); benign cases (duplicates, non-text payloads, loop guard, dispatch failure) ack with 200 so DingTalk does not retry. Error snippets carried into logs are truncated to 200 characters.
- `DingTalkCallbackCrypto` — verifies the `timestamp`/`sign` request headers (`Base64(HmacSHA256(appSecret, timestamp + "\n" + appSecret))`, constant-time compare; the header is trimmed once and the same value feeds both the freshness window and the HMAC input) before any body processing: timestamps may be up to 1 hour old (per DingTalk docs) but only up to 5 minutes ahead of local time (clock-skew allowance). When the robot is configured with an `aesKey`, decrypts the `{"encrypt": ...}` envelope (AES-CBC/PKCS7 with a 32-byte block — the WeCom EncodingAESKey family — strict unpadding, `16 random bytes | 4-byte length | message` layout, trailing bytes tolerated).
- `DingTalkChannelRegistry` — registration fails fast (`IllegalStateException`) when a different instance already holds the channel id, and unregistration is instance-bound (`ConcurrentHashMap.remove(key, value)`), so a stale channel can neither silently hijack nor detach another's callbacks.
- `DingTalkChannel` — `mode`-selected lifecycle: `http` registers with the registry and never opens the WebSocket; `stream` (default) is unchanged. Both modes share the same intake pipeline (`onInboundPayload`: msgId dedup → mapper → bot-loop guard → dispatch), and replies keep flowing through the existing `DingTalkOutboundClient`.
- `DingTalkChannelProperties` — new `mode` (`stream`|`http`, default `stream`) and optional `aesKey`, both fail-fast validated; a `@Deprecated` bridge constructor preserves the previous 6-argument signature.
- Docs updated (en + zh): mode comparison, HTTP-mode quickstart, config table, message flow, and a note recommending `aesKey` encryption for internet-facing callbacks (the signature authenticates only the timestamp header).

**How to test**

- `mvn -pl agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk test` — 34 tests: signature verification (valid / padded header / wrong secret / wrong timestamp / stale / future / future beyond clock skew / non-numeric), envelope decryption (round-trip, trailing-bytes tolerance, tampered framing and tampered padding rejected, missing aesKey), controller dispatch (unknown channel and bad signature both 401, 400 semantics, plain and encrypted delivery, plain body rejected on an encrypted channel, oversized body rejected, msgId dedup drops the duplicate, non-text acks without dispatch), registry semantics (duplicate channel id fails fast, restart idempotent, instance-bound unregistration), and properties validation.
- Manual: configure a robot's message reception to HTTP mode with the callback URL `https://your-host/api/channels/dingtalk/{channelId}/callback`, set `mode=http` (+ `aesKey` if body encryption is enabled), and message the bot.

## Checklist


agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackController.java
```diff
@@ -0,0 +1,128 @@
...
+        if (body == null || body.isBlank()) {
+            throw new IllegalStateException("Request body is empty");
+        }
+        JsonNode root;
+        try {
+            root = MAPPER.readTree(body);
+        } catch (Exception e) {
+            throw new IllegalStateException("Request body is not valid JSON: " + e.getMessage(), e);
+        }
+        JsonNode encrypt = root.get("encrypt");
```
oss-maintainer, line 124: When `aesKey` is configured the encrypted envelope is treated as optional: a body without an `encrypt` field falls through and is consumed as a plain payload. Combined with the header signature (which covers only `timestamp`, never the body), that means an attacker who captures one valid `timestamp`/`sign` pair can post an arbitrary **plaintext** body to a channel that is supposed to be encryption-only, for the whole skew window. Suggest failing closed here: if `channel.crypto()` has an `aesKey`, reject bodies that do not carry an `encrypt` envelope (400), and only accept plain JSON when encryption is not configured. Worth a matching case in `DingTalkCallbackControllerTest`.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackController.java
```diff
@@ -0,0 +1,128 @@
...
+     */
+    @PostMapping(
+            value = "/{channelId}/callback",
+            consumes = MediaType.APPLICATION_JSON_VALUE,
+            produces = MediaType.TEXT_PLAIN_VALUE)
+    public Mono<ResponseEntity<String>> dispatch(
+            @PathVariable String channelId,
+            @RequestHeader(value = "timestamp", required = false) String timestamp,
+            @RequestHeader(value = "sign", required = false) String sign,
+            @RequestBody String body) {
```
oss-maintainer, line 81: `@RequestBody String body` is unbounded, and the parse-failure path at line 90 logs `e.getMessage()` verbatim. Jackson's messages embed a snippet of the offending source, so an unauthenticated probe (signature is checked first here, but see the `extractPayload` failure path for a *valid* signature) can push attacker-controlled content of arbitrary size into the warn log and into heap. Recommend a request-size guard (reject over, say, 64 KB before parsing) and capping the logged exception message the way `WeComOutboundClient.truncate` does in #3173.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackController.java
```diff
@@ -0,0 +1,128 @@
...
+        DingTalkChannel channel = registry.get(channelId);
+        if (channel == null) {
+            log.warn("DingTalk dispatch: no channel registered for id='{}'", channelId);
```
oss-maintainer: Unknown-channel probes are answered with `404` plus a `warn` log before any signature check, so this endpoint doubles as a free `channelId` oracle and a log-flooding target. Consider logging at `debug` and returning the same response shape as the `401` path so probes are not distinguishable.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackCrypto.java
```diff
@@ -0,0 +1,167 @@
...
+        if (timestamp == null || sign == null) {
+            return false;
+        }
+        long ts;
+        try {
+            ts = Long.parseLong(timestamp.trim());
+        } catch (NumberFormatException e) {
+            return false;
+        }
+        if (Math.abs(System.currentTimeMillis() - ts) > MAX_TIMESTAMP_SKEW_MS) {
```
oss-maintainer: `sign = Base64(HmacSHA256(appSecret, timestamp + "\n" + appSecret))` is DingTalk's documented scheme, but note what it does *not* do: it authenticates a timestamp only, not the request body, and `Math.abs` accepts timestamps up to an hour in the past **and** future. A captured header pair is therefore replayable for up to two hours against any channel sharing that appSecret, with any body (the msgId dedup only stops identical `msgId`s, which the attacker controls). Two cheap hardenings: bound the future direction to a small clock-skew allowance (e.g. 5 min) and keep a short-lived seen-cache of `(timestamp, sign)` pairs sized to the accepted window.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackCrypto.java
```diff
@@ -0,0 +1,167 @@
...
+            }
+            return new String(unpad, 20, msgLen, StandardCharsets.UTF_8);
+        } catch (RuntimeException re) {
+            throw re;
+        } catch (Exception e) {
+            throw new IllegalStateException("DingTalk decrypt failed: " + e.getMessage(), e);
+        }
+    }
+
+    private static byte[] pkcs7Unpad(byte[] in) {
```
oss-maintainer, line 173: Lenient unpadding: `pad > 32` accepts a pad byte larger than the 16-byte AES block, and an out-of-range pad returns the input unchanged instead of failing. The downstream `msg_len` framing check catches most damage, but a strict check (`1 <= pad <= 16`, all trailing bytes equal to `pad`) makes the failure explicit. Related: the comment in `decryptRejectsTamperedFraming` says body integrity is "the header signature's job", which is not the case given the signature does not cover the body (see the finding above) — worth correcting before it misleads the next reader.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkChannelRegistry.java
```diff
@@ -0,0 +1,59 @@
...
+    private DingTalkChannelRegistry() {}
+
+    /** Returns the process-wide singleton instance. */
+    public static DingTalkChannelRegistry instance() {
+        return INSTANCE;
+    }
+
+    /** Registers a channel under its channel id; replaces any prior registration for that id. */
+    public void register(DingTalkChannel channel) {
+        channels.put(channel.channelId(), channel);
```
oss-maintainer: `put` silently replaces an existing registration. Since http mode is the mode advertised for multi-tenant deployments with robots added at runtime, a `channelId` collision would hijack an already-running channel's traffic with no trace. Suggest `putIfAbsent` and, on collision, either refuse to start the new channel or at least log the replaced instance at `warn`.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkChannel.java
```diff
@@ -130,17 +149,26 @@ public void init(Gateway gateway) {
...
+                properties.mode(),
                 properties.appKey(),
                 properties.robotCode());
     }
 
     @Override
     public void stop() {
-        streamClient.stop();
+        if (crypto != null) {
+            registry.unregister(channelId);
```
oss-maintainer: `unregister(channelId)` removes by key, so if two instances ever share an id (see the registry finding) stopping the old one also detaches the new one's registration and its callbacks start answering `404`. Bind the removal to the instance: expose `unregister(String channelId, DingTalkChannel expected)` backed by `ConcurrentHashMap.remove(key, value)`, and have `stop()` pass `this`.

oss-maintainer (commented):
### Summary

Adds an HTTP-callback reception mode for the DingTalk channel: `DingTalkCallbackController` (`POST /api/channels/dingtalk/{channelId}/callback`) verifies the `timestamp`/`sign` headers, optionally decrypts the `encrypt` envelope, and feeds the result into the existing `DingTalkChannel` intake pipeline. Selection is by `mode` on `DingTalkChannelProperties`, with a process-wide `DingTalkChannelRegistry` bridging Spring's component-scanned controller and the factory-created channel.

The structure is right — reuse of the shared intake path (dedup, mapping, bot-loop guard) instead of a second pipeline, `MessageDigest.isEqual` for the signature comparison, a back-compatible deprecated constructor on the record, and genuinely thorough tests (404/401/stale-signature/duplicate-msgId/encrypted/plain/malformed) with matching EN+ZH docs. Commented rather than approved because the authentication layer has one structural gap that weak mode selection would expose, plus a couple of state-lifecycle nits.

### Findings

- **[Warning]** `DingTalkCallbackController.java:112` — the `encrypt` envelope is optional even when `aesKey` is configured, so a plain body is accepted on an "encrypted" channel. Since the header signature never covers the body (next finding), that is the difference between encryption being enforced and being advisory.
- **[Warning]** `DingTalkCallbackCrypto.java:99` — `sign` authenticates `timestamp` only, and `Math.abs` accepts a one-hour window in *both* directions: a captured header pair is replayable for up to two hours with any attacker-chosen body and any `msgId`. Bound the future direction and add a short-lived `(timestamp, sign)` seen-cache.
- **[Warning]** `DingTalkChannelRegistry.java:47` / `DingTalkChannel.java:168` — `register` replaces an existing id silently while `stop` unregisters by key, so on a `channelId` collision the old instance can detach the new one's registration. Use `putIfAbsent` and the two-arg `remove(key, value)`.
- **[Warning]** `DingTalkCallbackController.java:75` — unbounded `@RequestBody String` plus verbatim `e.getMessage()` in the warn log; cap the body size and truncate the logged snippet.
- **[Info]** `DingTalkCallbackController.java:78` — the unknown-channel `404` is returned before any signature check, making the endpoint a `channelId` oracle and a log-flooding target.
- **[Info]** `DingTalkCallbackCrypto.java:157` — `pad > 32` admits a pad byte larger than the AES block, and invalid padding returns the input unchanged rather than failing; the "header signature's job" comment in `decryptRejectsTamperedFraming` overstates what the signature protects.

### Suggestions

```java
// controller: fail closed when the channel is configured for encryption
if (channel.crypto().isEncrypted()) {
    JsonNode encrypt = root.get("encrypt");
    if (encrypt == null || !encrypt.isTextual() || encrypt.asText().isBlank()) {
        throw new IllegalStateException("Channel expects an encrypted callback body");
    }
    return MAPPER.readTree(channel.crypto().decrypt(encrypt.asText()));
}
```

```java
// crypto: asymmetric skew
long delta = System.currentTimeMillis() - ts;
if (delta > MAX_TIMESTAMP_SKEW_MS || -delta > FUTURE_SKEW_MS) {
    return false; // FUTURE_SKEW_MS ~ 5 minutes
}
```

The registry could also drop the global-singleton shape entirely if the controller were constructed through the same wiring that creates channels, but the current bridge is documented and the two-arg `remove` fix is enough to make the collision case safe.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackController.java
Buktal: Addressed in 5d9e6ac4: channels configured with an `aesKey` now fail closed — a body without an `encrypt` envelope is rejected with 400 before any dispatch, symmetric with the existing envelope-without-aesKey rejection inside `decrypt()`. Covered by `returns400ForPlainBodyOnEncryptedChannel`.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackController.java
Buktal: Addressed in 5d9e6ac4: bodies over 64 KB are rejected with 400 before parsing (`MAX_BODY_LENGTH`), and the warn log truncates the exception message to 200 characters (`MAX_SNIPPET_LENGTH` + `truncate`, the same treatment #3173 applies to WeCom response snippets). Covered by `returns400ForOversizedBody`.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackController.java
Buktal: Kept as-is, for two reasons. Operationally, the 404 is the misconfiguration signal for the operator wiring the callback URL (the channelId is part of the URL they registered with DingTalk); hiding it behind a 401-shaped response makes a wrong path indistinguishable from a wrong secret. Security-wise, a probe only learns id-existence — dispatch is still gated on the signature. The in-repo WeCom/GitHub/GitLab callback controllers answer with the same shape (e.g. `WeComCallbackController` returns 404 + warn before any signature check), so changing only DingTalk would fragment the pattern; happy to align all of them in a separate change if the hardened shape is preferred.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackCrypto.java
Buktal: Partially addressed in 5d9e6ac4. The future direction is now bounded to a 5-minute clock-skew allowance (`MAX_FUTURE_TIMESTAMP_MS`); the past direction keeps the documented one-hour window, so a captured pair is no longer replayable far into the future. Not adding a seen-cache: capturing a valid header pair requires observing legitimate traffic (with TLS in front, that presumes endpoint compromise); a cache is per-node while http mode targets exactly multi-instance deployments, so it would deliver partial protection at the cost of re-introducing time-bounded unbounded state; and since the signature in this scheme does not cover the body, per-node caching cannot provide full replay defense anyway.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackCrypto.java
Buktal: Partially addressed in 5d9e6ac4. The lenient passthrough is fixed: out-of-range or inconsistent padding now throws `IllegalStateException`, and all trailing pad bytes are verified equal (strict PKCS#7 validity) — `decryptRejectsTamperedPadding` corrupts the final ciphertext block. The `1..16` bound itself is not adopted: this envelope family (WeCom EncodingAESKey → DingTalk-Callback-Crypto) pads to a 32-byte block, so valid pad values span 1..32 rather than the 16-byte AES block size; a comment now records why. The overstated test comment is also reworded — the header signature covers only the timestamp, so body tampering is caught structurally (framing/padding checks), not cryptographically.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkChannelRegistry.java
Buktal: Addressed in 5d9e6ac4: `register` is now `putIfAbsent` — a different instance already holding the id throws `IllegalStateException` at `start()` (fail fast instead of silent hijack), and re-registering the same instance is a no-op (`duplicateChannelIdFailsFastAndKeepsOriginalRegistration`, `restartingSameChannelIsIdempotent`). Note the in-repo WeCom registry shares the replace-on-collision shape; aligning it would be a small separate change if wanted.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkChannel.java
Buktal: Addressed in 5d9e6ac4: `unregister(channelId, expected)` is backed by `ConcurrentHashMap.remove(key, value)` and `stop()` passes `this`, so a stale instance can never detach a successor registration (`unregisterOnlyRemovesTheMatchingInstance`).

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackController.java
```diff
@@ -0,0 +1,155 @@
...
```
oss-maintainer: The unknown-channel `404` is still returned before any signature check, so an unauthenticated caller can enumerate valid `channelId` values and emit one warn log per probe. Verifying `timestamp`/`sign` first (or returning the same 404 shape for a bad signature) removes the oracle; a rate limit on this warn would also help. Not blocking — it leaks existence only, never message content.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackCrypto.java
```diff
@@ -0,0 +1,189 @@
...
+            return false;
+        }
+        long ts;
+        try {
+            ts = Long.parseLong(timestamp.trim());
+        } catch (NumberFormatException e) {
+            return false;
+        }
+        long age = System.currentTimeMillis() - ts;
+        if (age > MAX_TIMESTAMP_AGE_MS || -age > MAX_FUTURE_TIMESTAMP_MS) {
```
oss-maintainer, line 110: Residual, by design of DingTalk's scheme: the signature covers only `timestamp
 + appSecret`, so a captured header pair stays replayable for the past-direction one-hour window with an arbitrary body and an arbitrary `msgId` (the shared `IdempotencyStore` cannot dedup a fresh id). The asymmetric skew already cuts the worst case from ~2h to ~1h; the only further mitigation is a short-lived `(timestamp, sign)` seen-cache. If that stays out of scope, a one-line note in the docs table under `mode: http` would set the right expectation for operators.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackCrypto.java
```diff
@@ -0,0 +1,189 @@
...
+     * local time than {@link #MAX_FUTURE_TIMESTAMP_MS}, and {@code sign} must equal {@code
+     * Base64(HmacSHA256(appSecret, timestamp + "\n" + appSecret))}.
+     */
+    public boolean verify(String timestamp, String sign) {
+        if (timestamp == null || sign == null) {
+            return false;
+        }
+        long ts;
+        try {
+            ts = Long.parseLong(timestamp.trim());
```
oss-maintainer: Minor consistency nit: the freshness window parses `timestamp.trim()` while the HMAC below is computed over the untrimmed header, so a padded header passes the window check and then fails the signature. Harmless (fails closed) — trimming once into a local and using it for both would make the intent clearer.

oss-maintainer (approved):
### Summary

Re-review of the new head (`5d9e6ac4`, squashed single commit). Four of the six previous findings are fully fixed and the fixes are correct, not cosmetic; the two remaining items are non-blocking. Approving.

Verified fixes:
- **Fail-closed envelope handling** — `extractPayload` now rejects a plain body when `aesKey` is configured, so encryption is enforced rather than advisory; the reverse mismatch still fails inside `decrypt()`.
- **Asymmetric timestamp skew** — `MAX_FUTURE_TIMESTAMP_MS` (5 min) bounds the future direction while keeping the documented one-hour freshness window, which is exactly the shape suggested.
- **Registry collision safety** — `register` uses `putIfAbsent` and fails fast when a *different* instance already holds the id, and `unregister(channelId, expected)` uses the two-arg `remove`, so a stale `stop()` can no longer detach a successor's registration.
- **Body/log bounds** — `MAX_BODY_LENGTH` caps accepted bodies and `truncate()` caps exception snippets, with the rationale (Jackson messages embed parsed source) documented at the constant.
- **Padding** — `pkcs7Unpad` now validates every pad byte and throws instead of returning the input unchanged, and the 1..32 range is justified by the WeCom/DingTalk envelope family rather than the 16-byte AES block.

### Findings

- **[Info]** `DingTalkCallbackController.java:88` — the unknown-channel `404` is still returned before any signature check, so the path remains a cheap `channelId` oracle and an unauthenticated warn-log flood target. Consider verifying the signature first (the channel lookup can happen inside the authenticated branch) or rate-limiting these warnings. Non-blocking: it leaks only existence, not message content.
- **[Info]** `DingTalkCallbackCrypto.java:113` — `ts` is parsed from `timestamp.trim()` but the HMAC is computed over the untrimmed header, so a padded header verifies the window and then fails the signature. Harmless in practice (DingTalk sends unpadded values) and fails closed; a comment or trimming once would make the intent explicit.
- **[Info]** Replay of a captured `timestamp`/`sign` pair is still possible for up to the one-hour window with an attacker-chosen body on a channel without `aesKey`, since the signature covers the timestamp only. The shared `IdempotencyStore` keyed on `msgId` limits duplicate *processing* but not arbitrary payloads. This is inherent to DingTalk's robot scheme, so documenting "enable message encryption for internet-facing callbacks" in the docs table would be enough.

### Suggestions

```java
// verify(): one canonical value for both the window check and the HMAC
String normalized = timestamp.trim();
long ts = Long.parseLong(normalized);
...
mac.doFinal((normalized + "\n" + appSecret).getBytes(StandardCharsets.UTF_8));
```

Test coverage is strong — stale/future/valid/invalid signatures, encrypted and plain paths, duplicate `msgId`, oversized body, malformed JSON, tampered ciphertext framing, registry ownership, and the properties validation matrix, with EN and ZH docs updated in lockstep. Thanks for the thorough round, and for the clear `@deprecated` bridge constructor that keeps existing stream-mode wiring source-compatible.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackController.java
Buktal: Addressed in 1d61237a: unknown channel ids and bad signatures now answer the same 401, and the unknown-id path logs at debug so probes no longer reach the warn stream. Verifying the signature first is not feasible here — the HMAC key is the per-channel appSecret, so an unknown id has nothing to verify against. Unified 401 keeps cannot-authenticate semantics (the same do-not-reveal-existence idiom as unified login failures); if the 404 shape is preferred instead it is a two-line flip. Covered by `returns401ForUnknownChannel`.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackCrypto.java
Buktal: Agreed that the seen-cache stays out of scope. Added the operator note in 1d61237a to both en and zh docs, directly under the config table: internet-facing `http` callbacks should enable message encryption (`aesKey`), because the signature authenticates only the timestamp header and a captured pair remains replayable within the freshness window.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackCrypto.java
Buktal: Addressed in 1d61237a: the header is trimmed once into a local that feeds both the freshness window and the HMAC input, so a padded header now verifies consistently instead of passing the window check and then failing the signature. `verifiesPaddedTimestampHeader` locks the behavior in.

Issue #3179, opened 2026-09-17
[Feature]: Surface inbound media messages (image/voice/video/file) in IM channel mappers


Buktal:
**Is your feature request related to a problem? Please describe.**

We run a production assistant over the Feishu, WeCom and DingTalk channels (`agentscope-extensions-channel-{feishu,wecom,dingtalk}`). Whenever a user sends an image, voice, video or file message to the bot, the event is acknowledged and dropped inside the inbound mapper — no `InboundMessage` is created, so the agent never sees the turn at all:

- `DingTalkInboundMapper` returns empty for every `msgtype` other than `text` (`DingTalkInboundMapper.java:64-67`; the javadoc at :31-33 states "For MVP only `msgtype=text` is mapped. Other inbound types (image, file, picture, ...) are returned as `Optional#empty()` so the caller can ack the event without dispatching to the agent")
- `FeishuInboundMapper` returns empty for every `message_type` other than `text` (`FeishuInboundMapper.java:123-126`; javadoc :50-51)
- `WeComInboundMapper` returns empty for every `MsgType` other than `text` (`WeComInboundMapper.java:62-65`; javadoc :37-39 — "Other inbound types (event, image, voice, ...)")

Two consequences for applications built on these channels:

1. **Total silence towards the user.** Someone sends a picture and gets no reply — not even a canned "media is not supported here". The application has no hook to produce such a reply, because nothing reaches the router or the agent.
2. **No opt-in path.** The mappers are final classes instantiated inside the channel implementations (`DingTalkChannel.java:97`, `FeishuChannel.java:100`, `WeComChannel.java:108`), with no constructor or properties-based injection point. An application cannot supply its own mapper either — the only workaround today is copying the whole channel class.

Meanwhile the provider payloads carry everything needed to act on the media later. Verified against the official receiving-message references:

- **WeCom** (callback message-format doc, decrypted XML): `MsgType` ∈ {`image`, `voice`, `video`} besides text/location/link. `image` carries `PicUrl` + `MediaId`; `voice` carries `MediaId` + `Format`; `video` carries `MediaId` + `ThumbMediaId`. A `MediaId` is exchanged at `GET /cgi-bin/media/get` with the app `access_token` and is valid for three days.
[...]

PR #3180, opened 2026-09-17
feat(channel): surface inbound media messages as neutral metadata

Buktal:
## AgentScope-Java Version

2.0.3-SNAPSHOT (main @ `ca524056`)

## Description

Implements #3179. `Fixes #3179`

The Feishu, WeCom and DingTalk inbound mappers currently return `Optional.empty()` for every non-text message, so media events are acked and dropped — the agent never sees the turn and applications cannot even reply "media is not supported". This PR makes the mappers surface media events instead, while leaving every decoding decision to the application.

**Approach — mapping is mechanism, support is policy:**

- Each media event maps to a user `Msg` whose text content is a neutral marker (`[image]`, `[audio]`, `[video]`, `[file: report.pdf]`) and whose metadata carries the new `ChannelMediaMetadata` contract (channel-common): normalized kind (`image`/`audio`/`video`/`file`), the original provider type token, the primary material identifier verbatim (`MediaId` / `image_key` / `file_key` / `downloadCode`), and the platform metadata that exists per kind (`PicUrl`, `Format`/`videoType`, `ThumbMediaId`/cover `image_key`, `file_name`, duration in ms, DingTalk's `recognition` ASR text). All values are strings; absent keys are omitted (sparse map). The duration key states its millisecond unit on the contract itself — Feishu and DingTalk both document the field in milliseconds. The file-name copy embedded in the marker is sanitized (brackets, control characters and invisible formatting characters — zero-width spaces, bidirectional overrides — removed, whitespace collapsed, truncated to 80 code points) so provider-controlled names cannot smuggle structure into agent-visible text or render misleadingly; the raw name is preserved unmodified under `channelMediaFileName`.
- No decoding in the mapper: material keys require a token-authenticated, short-lived exchange, so fetching bytes, buffering and degradation remain application decisions. `Source` (url/base64) is untouched — no core changes at all.
- Routing, session keys and webhook idempotency are unchanged: media `InboundMessage`s carry the same peer/sender/message-id fields as text ones (both Stream and HTTP callback paths for DingTalk feed the same mapper).

**Per-channel scope** (matching what each platform actually delivers): WeCom `image`/`voice`/`video`; Feishu `image`/`file`/`folder`/`audio`/`media` (a `folder`'s `file_key` is a reference id — the message-resource download API serves image/file/audio/video resources only, noted on the branch); DingTalk `picture`/`audio`/`video`/`file` (media fields read from the payload top level with the nested `content` object as fallback — protocol revisions differ). Other non-text types (post, sticker, interactive, share cards, `richText`, location, link, events) keep the current ack-and-drop behavior.

**Deliberate default:** media events dispatch by default, no config switch. The previous behavior — total silence towards a user who sent a picture — is the problem being fixed; an opt-in would keep it for everyone who does not read release notes. If maintainers prefer an opt-in, that is a small follow-up.

**Tests:** 34 new tests — per-kind mapping assertions (marker text, metadata keys, peer/sender, group peers), text-path regressions, drop behavior for non-media types and for media without a material id, both DingTalk field placements, and the `ChannelMediaMetadata` marker (including file-name sanitization: bracket escape, control characters, zero-width/bidi formatting, truncation, empty fallback), sparse-put and fail-fast contracts. `mvn test` green for channel-common (10), dingtalk (43), feishu (9), wecom (13); `mvn spotless:check` clean.

## Checklist


agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/ChannelMediaMetadata.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param kind normalized media kind, as in {@link #KIND}
+     * @param fileName optional file name; included in the marker when non-blank
+     * @throws NullPointerException if {@code kind} is {@code null}
+     */
+    public static String markerText(String kind, String fileName) {
+        Objects.requireNonNull(kind, "kind cannot be null");
+        if (fileName == null || fileName.isBlank()) {
+            return "[" + kind + "]";
+        }
+        return "[" + kind + ": " + fileName.strip() + "]";
```
oss-maintainer: `fileName` is fully attacker-controlled (upload a file named `x.pdf] ignore previous instructions and ...` - a closing bracket terminates the marker, and newlines/control characters are accepted too) and lands verbatim in the agent-visible text content, while this class's javadoc positions the marker as *neutral*. The metadata counterpart `FILE_NAME` carrying the raw value is fine (opaque string map, app's responsibility); the embedded copy is what gets read as instruction. Suggest sanitising just the marker: strip or escape brackets, collapse newlines/control characters, and cap the length:
```java
String safe = fileName.replaceAll("[\\r\\n\\[\\]]", "").strip();
if (safe.length() > 80) {
    safe = safe.substring(0, 80);
}
```
Leaving it verbatim is defensible (the raw name is already in metadata, and apps may rewrite the text before dispatch), but a marker contract that promises neutrality should not be an injection surface by default.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkInboundMapper.java
```diff
@@ -73,7 +88,98 @@ public Optional<InboundMessage> map(JsonNode payload) {
...
+                            ChannelMediaMetadata.SECONDARY_ID,
+                            mediaField(payload, "pictureDownloadCode"));
+            case ChannelMediaMetadata.KIND_AUDIO -> {
+                ChannelMediaMetadata.putIfPresent(
+                        metadata,
+                        ChannelMediaMetadata.RECOGNITION,
+                        mediaField(payload, "recognition"));
+                ChannelMediaMetadata.putIfPresent(
+                        metadata,
+                        ChannelMediaMetadata.DURATION_MS,
```
oss-maintainer, line 130: DingTalk documents audio `duration` in **milliseconds** but video `duration` in **seconds**, and the same key is used for both verbatim here. That also quietly contradicts the "All values are strings" sentence in the contract javadoc on the JSON paths - `"duration":5000` is a JSON number, so `mediaField`/`asText(null)` yields its text. Both behaviours are acceptable, but worth one line: either normalize per kind before putting, or state that `DURATION_MS` is "platform-reported value as text - unit follows the provider payload" so downstream code does not divide by 1000 twice.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuInboundMapper.java
```diff
@@ -183,4 +200,78 @@ public Optional<InboundMessage> map(JsonNode envelope) {
...
+            ChannelMediaMetadata.putIfPresent(
+                    metadata,
+                    ChannelMediaMetadata.SECONDARY_ID,
+                    content.path("image_key").asText(null));
+        }
+        if (ChannelMediaMetadata.KIND_AUDIO.equals(kind)
+                || ChannelMediaMetadata.KIND_VIDEO.equals(kind)) {
+            ChannelMediaMetadata.putIfPresent(
+                    metadata,
+                    ChannelMediaMetadata.DURATION_MS,
```
oss-maintainer, line 241: Same theme as the DingTalk note: `content.path("duration").asText(null)` on a JSON number yields the text verbatim (and `12.0` if the SDK serialises it as a double), so `DURATION_MS` is opaque and unit-unverified across the three channels. The name implies milliseconds the mapper does not enforce - either coerce numbers (`isNumber()` -> `asLong()` -> `Long.toString`), or document the passthrough semantics on `ChannelMediaMetadata.DURATION_MS` itself.

oss-maintainer (commented):
### Summary

Makes Feishu / WeCom / DingTalk inbound mappers surface media messages (previously `Optional.empty()` — acked and silently dropped) as a user `Msg` with a neutral marker text plus a `ChannelMediaMetadata` key contract, deliberately leaving byte fetching and degradation to the application. The mechanism/policy split is the right call, routing, session keys and webhook idempotency are untouched, and the per-provider test coverage is thorough (nested vs top-level DingTalk fields, resource-required `extractMediaId`, kind normalisation, empty-string guards). CI is green on this head. One thing I would fix before merge, then three minor notes; diff is above the 1000-line threshold, so I focused on the contract and the three mappers rather than line-reviewing every test.

### Findings
- **[Warning]** `ChannelMediaMetadata.java:100` — the "neutral" marker embeds a fully attacker-controlled file name verbatim into agent-visible text; a `]` closes the marker and newlines are accepted, which makes it a prompt-injection surface by default. The raw name in metadata is fine; suggest sanitising and truncating only the marker copy.
- **[Info]** `DingTalkInboundMapper.java:129` — audio `duration` is ms, video `duration` is seconds on DingTalk, both stored under `DURATION_MS`, and JSON numbers are stringified; worth normalising or documenting the passthrough.
- **[Info]** `FeishuInboundMapper.java:241` — same point on the Feishu path: the key name promises milliseconds the mapper does not enforce.
- **[Info]** `FeishuInboundMapper.java:259` — `folder` is surfaced as `KIND_FILE`, but its `file_key` is not fetchable through the normal message-resource download; a one-line comment would save integrators a debugging cycle.

### Suggestions
The filename note is the only one I'd consider blocking: every consumer that trusts "the marker text is neutral" inherits the injection, and it is a three-line fix in one helper. For the duration notes, a short sentence on `DURATION_MS` ("platform-reported value as text; unit follows the provider payload") resolves the ambiguity without forcing a normalisation the platforms make awkward.

Otherwise this reads as a clean widening of the inbound contract: unchanged dispatch path, marker text that keeps text-only agents aware, and ids preserved verbatim so the app can decide what to fetch.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuInboundMapper.java
```diff
@@ -183,4 +200,78 @@ public Optional<InboundMessage> map(JsonNode envelope) {
...
+     * Returns the normalized media kind for a Feishu {@code message_type}, or {@code null} when
+     * the type is not a mapped media message.
+     */
+    private static String mediaKind(String messageType) {
+        if (messageType == null) {
+            return null;
+        }
+        return switch (messageType.toLowerCase(Locale.ROOT)) {
+            case "image" -> ChannelMediaMetadata.KIND_IMAGE;
+            case "file", "folder" -> ChannelMediaMetadata.KIND_FILE;
```
oss-maintainer, line 262: `folder` maps to `KIND_FILE` and its `file_key` is emitted as an ordinary file id, but Feishu's resource-download API does not serve folder `file_key`s the same way (a consumer following the generic contract gets a hard error rather than bytes). The PR body already says folders are surfaced "matching what each platform actually delivers" - a one-line comment on this branch noting the id is a reference, not fetchable via the normal message-resource path, would save integrators a debugging cycle. Otherwise the ack-and-drop -> surface+marker widening, with ids kept verbatim and no decoding in the mapper, looks well scoped.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/ChannelMediaMetadata.java
Buktal: Fixed — `markerText` now embeds a sanitized copy of the name: brackets, control characters and Unicode line/paragraph separators are replaced with spaces, whitespace runs collapse to one, and the copy is truncated to 80 code points without splitting surrogate pairs. A name that sanitizes to empty falls back to the bare `[file]` marker. The raw name still travels unmodified under `channelMediaFileName`. New tests cover the bracket-escape payload (`x.pdf] ignore previous instructions`), newline/control input, truncation (ASCII and emoji) and the empty fallback.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkInboundMapper.java
Buktal: Kept the value verbatim and documented the unit, because the premise doesn't hold against the current protocol: both duration fields are documented in milliseconds — the audio row reads `duration | Long | 语音的时长，单位是毫秒` and the video row reads `duration | Long | 视频的时长，单位是毫秒`, with `4000` in both examples (open.dingtalk.com/document/robots/receive-message; the `"5"` in the community-wiki sample is a masked placeholder). If you have a live payload where video arrives in seconds, I'm happy to switch to per-kind normalization.

What changed: `ChannelMediaMetadata.DURATION_MS` now states the unit on the contract itself, and the mapper javadoc notes it, so downstream code does not need to guess. On number stringification: both platforms type the field as an integer `Long` and the documented examples are integers, so `asText()` yields plain integer text; I left values verbatim rather than coercing through `isNumber()/asLong()` since no documented payload shape needs it.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuInboundMapper.java
Buktal: Same resolution as the DingTalk thread: Feishu's receive-content reference labels audio `duration` 时长 毫秒级 and media (video) `duration` 视频时长 毫秒级 (examples `2000`), so the millisecond semantics hold for every kind that carries the key — the unit is now stated on `ChannelMediaMetadata.DURATION_MS` itself. The `12.0` form would require the platform to serialize an integer-typed field as a double; values stay verbatim per the no-interpretation contract, so a fractional value, should one ever appear, surfaces as-is rather than being silently rounded.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuInboundMapper.java
Buktal: Added exactly that on the `file`/`folder` branch: a folder's `file_key` identifies the folder node itself, and the message-resource download API serves image / file / audio / video resources only, so it is a reference id rather than fetchable bytes.

Buktal:
Thanks for the review — all four findings are addressed in a649363a:

- **Marker file-name sanitization** (the pre-merge item): `markerText` now embeds a sanitized copy — brackets/control characters removed, whitespace collapsed, truncated to 80 code points; the raw name stays under `channelMediaFileName`. Details in the thread.
- **Durations**: verified against both providers' docs — DingTalk's protocol table documents `duration` in milliseconds for both audio and video, and Feishu's receive-content reference does the same for audio/media. No conversion needed; the millisecond unit is now stated on `ChannelMediaMetadata.DURATION_MS` and in the DingTalk mapper javadoc.
- **Feishu `folder`**: the branch now notes its `file_key` is a reference id — the message-resource download API serves image/file/audio/video only.

Tests are now 33 new (channel-common 9); `spotless:check` clean.

References for the doc-verified claims:

- DingTalk receive-message protocol — audio row `duration | Long | 语音的时长，单位是毫秒`, video row `duration | Long | 视频的时长，单位是毫秒`: https://open.dingtalk.com/document/robots/receive-message
- Feishu receive-message content reference — audio `duration` marked 时长 毫秒级, media (video) `duration` marked 视频时长 毫秒级: https://open.feishu.cn/document/server-docs/im-v1/message-content-description/message_content?lang=zh-CN
- Feishu message-resource download, whose supported types are audio / video / image / file: https://open.feishu.cn/document/server-docs/im-v1/message/get-2?lang=zh-CN

oss-maintainer (approved):
### Summary
Re-reviewed the new head (`a649363a`) — the pre-merge blocker is fixed and the three documentation items are resolved with evidence. Approving, with one non-blocking nit below.

### What changed since my review
- **[resolved, was blocking] Marker injection via `fileName`** — `sanitizeMarkerFileName` closes the escape route: `]`/`[` and control characters are stripped so a crafted name cannot terminate the marker early and continue as agent-visible instruction text, whitespace collapses, and the 80-code-point truncation via `offsetByCodePoints` avoids splitting a surrogate pair. Keeping the raw name under `channelMediaFileName` is the right call — the marker stays a display convenience while consumers that need fidelity still have it.
- **[resolved] Duration units** — thanks for doing the doc verification rather than arguing from my note; both DingTalk (audio *and* video `duration` in ms) and Feishu (audio/media `duration` 毫秒级) contradict my original assumption that DingTalk video was in seconds. Stating the unit on `DURATION_MS` and in the mapper javadoc is the fix I wanted.
- **[resolved] Feishu `folder`** — the comment now records that a folder `file_key` is a reference id, not something the message-resource API will serve.

### Suggestion (non-blocking)
`ChannelMediaMetadata.sanitizeMarkerFileName` — `replaceAll("[\\[\\]\\p{Cntrl}\\u2028\\u2029]", " ")` still lets through the zero-width and bidi-override class (`U+200B`–`U+200F`, `U+202A`–`U+202E`, `U+FEFF`). Those cannot break out of the marker, so the security property holds, but a name like `report\u202Epdf` renders misleadingly in a terminal/IDE and an RTL override can hide what follows it. `\\p{Cf}` would cover the whole format category in one go if you touch this file again. Not worth another push on its own.

Nit only: `DURATION_MS`'s `asText()` path still returns a float-serialised `12.0` verbatim if a platform ever sends a JSON number with a fraction — harmless for the marker use case, mentioned for completeness.

CI green (6/6 SUCCESS), CLA signed on this head.

---

Buktal:
Thanks for the re-review. Adopted the Cf suggestion, with one adjustment to the recipe: the pushed class is `[\p{Cc}\p{Cf}\p{Zl}\p{Zp}]`, because a literal `\p{Cf}` swap would have narrowed coverage rather than widened it — `\p{Cf}` excludes the line/paragraph separators (U+2028/U+2029 are Zl/Zp) and the controls (`\p{Cf}` and `\p{Cc}` are disjoint; Java's `\p{Cntrl}` is ASCII-only and never matched the C1 range either). The union class covers the full invisible set — zero-width and bidi formatting, BOM, soft hyphen, C1 — and the new test asserts one code point per family, checked against the JDK's own regex engine. The `asText()` float nit stays as-is, per the documented integer typing. CI re-running.

PR #3182, opened 2026-09-17
feat(channel): pluggable access-token storage for channel providers

### LINKED ISSUE #3181: [Feature]: Pluggable access-token storage for channel providers

The Feishu, WeCom, and DingTalk channel extensions each fetch and cache a platform access token through a dedicated provider (`FeishuAccessTokenProvider`, `WeComAccessTokenProvider`, `DingTalkAccessTokenProvider`). In all three, the cache is a private in-process field with no way to replace it:

- Each provider is a `final` class holding `private final AtomicReference<TokenSlot> slot` (`FeishuAccessTokenProvider.java:41`, `WeComAccessTokenProvider.java:37`, `DingTalkAccessTokenProvider.java:41`), and the `TokenSlot` record is duplicated verbatim in all three.
- The channel factories construct the providers internally, and the channel constructors are private, so applications cannot assemble the graph with different token storage.

This is fine for a single instance. When channels run on several replicas (behind a load balancer, or one process per bot), the token lifecycle is duplicated per instance:

1. **N independent fetch cycles.** Every instance fetches its own token for the same credential. Platforms rate-limit token issuance, so fetch traffic multiplies with instance count for no benefit.
2. **Invalidation does not propagate.** The outbound clients invalidate their provider when the platform reports an expired/invalid token (e.g. `FeishuOutboundClient` on codes 99991663/99991664, `WeComOutboundClient` on 42001/40014). With process-local caches, only the instance that hit the error drops the token; other replicas keep sending a token the platform just rejected until their own copy reaches its refresh deadline — showing up as confusing partial failures.
3. **Cold starts stampede.** After a rolling restart every instance starts with an empty cache and fetches immediately; a shared cache would warm each instance instantly.

Extract the "where is the token cached" decision into a small SPI in `channel-common`, keeping all policy in the providers (refresh timing at ~80% of TTL, response parsing, error-code invalidation):

```java
public interface AccessTokenStore {
    /** Returns the cached token, or {@code null} when none is cached. */
    CachedAccessToken get();

    void put(CachedAccessToken token);

    /** Drops the cached token; the next token() call refreshes. */
    void clear();
}

public record CachedAccessToken(String value, long refreshAtMs) {}
```

- The provider remains the expiry authority: it computes `refreshAtMs` and applies the staleness check; the store is an opaque cache slot, one store instance per provider (i.e. per credential). A Redis-backed deployment wires one store per channel, keyed however the application prefers.
- A default `InMemoryAccessTokenStore` in `channel-common` reproduces today's behavior exactly.
- Each provider gains a constructor overload taking the store; existing constructors delegate with the in-memory default, so current callers are unaffected.
- A channel-factory variant passes the store through, alongside the existing `InboundEventDeduplicator` injection in `fromProperties` (the factories are the only construction path, since channel constructors are private). With a shared store, one instance's refresh or invalidation serves the whole deployment.

Our deployment runs channel replicas and would plug in a Redis-backed implementation of this seam; happy to contribute the PR.

- **Subclass the providers** — impossible (`final`) and undesirable; the useful variance is the storage, not the fetch logic.
- **Generalize the `OutboundClient` constructors** (currently bound to the concrete provider classes, e.g. `FeishuOutboundClient.java:54`, `WeComOutboundClient.java:52-53`) to a token-supplier interface — that would additionally allow swapping the whole provider, but it is a wider API change and is not required to share storage.
- **Accept per-instance caches** — workable at small scale, but the points above grow with replica count, and non-propagating invalidation in particular shows up as hard-to-diagnose partial outages.

**Additional context**

Non-goals, to keep the change small:

- **No single-flight refresh.** Concurrent `token()` calls on a stale cache may each fetch (already the case today); with a shared store the redundant fetches are benign last-writer-wins. Coordinated refresh can be layered later if wanted.
- **No change to refresh timing, response parsing, or error-code invalidation** — those stay in the providers.
- **No keyed multi-credential store.** One store instance per provider; applications sharing one connection compose facades.

Exact shapes (naming, `Optional` vs nullable `get()`, record vs plain parameters) are open to maintainer preference — the capability is the point.

Buktal:
## AgentScope-Java Version

2.0.3-SNAPSHOT (main, fe50c3c7)

## Description

Adds a small SPI for where channel access tokens are cached, so multi-instance deployments can share them. Fixes #3181.

- New `AccessTokenStore` interface, `CachedAccessToken` record, and default `InMemoryAccessTokenStore` in `agentscope-extensions-channel-common`. The store is an opaque cache slot — one store instance per credential; refresh timing (~80% of TTL), response parsing, and error-code invalidation stay in the providers. Fetched tokens are stored via `putIfNewer` (freshest-wins) so racing refreshes cannot regress the cache to an earlier deadline; `put` replaces unconditionally, and shared-storage implementations can override `putIfNewer` with an atomic conditional write.
- `FeishuAccessTokenProvider` / `WeComAccessTokenProvider` / `DingTalkAccessTokenProvider` gain a constructor overload taking the store; existing constructors delegate to the in-memory default, and the three duplicated private `TokenSlot` records are removed.
- The three channel factories gain a `fromProperties` overload passing the store through next to the existing `InboundEventDeduplicator` injection; the shorter overloads delegate with in-memory defaults, so current callers are unaffected.
- With a shared store, one instance's refresh or invalidation (e.g. after a 42001/40014 rejection) serves the whole deployment, and freshly started replicas warm from the shared cache instead of stampeding the token endpoint.
- Docs: "Multi-instance deployments" section in `docs/v2/{en,zh}/integration/channel/index.md`.

How to test: run the four touched channel modules with `mvn test` — 65 tests pass, including 13 new ones (store contract incl. `putIfNewer` merge semantics + per-provider injection seams; the seam tests point the API base at an unreachable address so any regression that bypasses the store or serves a stale entry fails fast). The existing WeCom and DingTalk suites pass unchanged.

## Checklist


agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/AccessTokenStore.java
```diff
@@ -0,0 +1,55 @@
...
+ * serves one credential.
+ *
+ * <p>The default {@link InMemoryAccessTokenStore} keeps the token in the JVM heap, so it is not
+ * shared across processes. Deployments running multiple channel instances should provide an
+ * implementation backed by shared storage (for example Redis keyed by channel id) and inject it
+ * through the channel factories: one instance's refresh or invalidation then serves the whole
+ * deployment.
+ *
+ * <p>Implementations must be thread-safe: providers invoke these methods from concurrent reactive
+ * pipelines, and concurrent {@link #put} calls are permitted (last write wins).
```
oss-maintainer: LWW + an un-keyed slot has one sharp edge for the *shared* implementations this SPI exists for. Concurrent cold-start refreshes are expected (there is no cross-instance lock here): replica A's request starts first but returns last, and its `put` overwrites B's strictly-fresher token. The surviving `refreshAtMs` is then the earlier deadline, and with provider-side rotation the stale value may already be dead - the whole deployment pins the older token until the next error-triggered `invalidate()`.

Suggest making the contract freshest-wins instead of last-wins. A provider-side guard keeps the SPI unchanged:
```java
CachedAccessToken cur = store.get();
if (cur == null || refreshAt > cur.refreshAtMs()) {
    store.put(new CachedAccessToken(token, refreshAt));
}
```
- or add `default boolean putIfNewer(...)` here so shared backends can implement it atomically (Redis `SET ... GT`-style) instead of read-then-write.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/AccessTokenStore.java
```diff
@@ -0,0 +1,55 @@
...
+     *     deadline carried in {@link CachedAccessToken#refreshAtMs()}
+     */
+    CachedAccessToken get();
+
+    /**
+     * Caches {@code token}, replacing any previous entry.
+     *
+     * @param token the token to cache
+     */
+    void put(CachedAccessToken token);
```
oss-maintainer, line 60: `get()`/`put()` carry no key and no credential identity, so injecting one store instance into providers built from *different* credentials (two bots, same extension) silently mixes their tokens. The class javadoc's "one store instance serves one credential" is the right rule - worth restating it on the `put`/`get` contract (or on the factory `tokenStore` `@param`) as an explicit "do not share one instance across credentials" obligation, since implementers of the SPI read the method docs first.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/test/java/io/agentscope/extensions/channel/wecom/WeComAccessTokenProviderTest.java
```diff
@@ -0,0 +1,51 @@
...
+class WeComAccessTokenProviderTest {
+
+    private static final String UNREACHABLE_API_BASE = "http://localhost:1";
+
+    private final InMemoryAccessTokenStore store = new InMemoryAccessTokenStore();
+    private final WeComAccessTokenProvider provider =
+            new WeComAccessTokenProvider(UNREACHABLE_API_BASE, "corpId", "secret", store);
+
+    @Test
+    void servesCachedTokenFromInjectedStore() {
```
oss-maintainer, line 40: The three provider seam tests cover the cache-hit and `invalidate()` paths, but not the branch this refactor touches most: an entry whose `refreshAtMs` has passed must *not* be served. `token()` still returns `Mono.just(cached.value())` as long as the store hands an entry back, so only the provider applies the deadline - exactly the invariant a regression could break. A cheap addition next to `http://localhost:1`:
```java
@Test
void expiredEntryTriggersRefreshNotStaleToken() {
    store.put(new CachedAccessToken("stale", System.currentTimeMillis() - 1_000));
    assertThrows(RuntimeException.class, () -> provider.token().block(Duration.ofSeconds(2)));
}
```
(Fails if the deadline check is removed - the refresh attempt against the unreachable base throws, so the stale token never escapes.)

oss-maintainer (commented):
### Summary

Introduces an `AccessTokenStore` SPI so the three channel providers' access tokens can live in shared storage instead of a private `AtomicReference`, wired through new constructor/`fromProperties` overloads. The seam is well judged: the provider keeps refresh timing, parsing and error-code invalidation, the old constructors delegate to an in-memory default so no existing caller changes, and the duplicated `TokenSlot` records go away. CI is green on this head (build ubuntu/windows, validate, license, module sync, codecov). One design nit worth resolving before this becomes a public SPI, plus two smaller notes.

### Findings
- **[Warning]** `AccessTokenStore.java:33` — last-write-wins is the wrong merge rule for the shared backends this SPI exists for; concurrent cold-start refreshes can pin the older/shorter-lived token deployment-wide. Suggest a freshest-wins guard (or `putIfNewer`) so the contract is safe by construction.
- **[Info]** `AccessTokenStore.java:51` — the store is un-keyed, so "one instance per credential" is a real hazard; worth stating it on the method/`@param` contract, not only in class javadoc.
- **[Info]** `WeComAccessTokenProviderTest.java:40` — the new expiry-deadline branch (stale entry must not be served) is the one path the seam tests don't pin; cheap to add given the unreachable API base already fails fast.

### Suggestions
The freshest-wins change is three lines per provider and keeps the interface as-is; if you would rather push it into the store, a `default boolean putIfNewer(CachedAccessToken t)` that falls back to unconditional `put` preserves today's behaviour for in-memory implementations while letting a Redis-backed store do it atomically.

Nice touch keeping the en/zh docs in sync with the `fromProperties` example.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/AccessTokenStore.java
Buktal: Agreed — the merge rule belongs in the store contract so shared backends can make it atomic. Went with the `putIfNewer` option: added `default boolean putIfNewer(CachedAccessToken)` to the SPI (default is a non-atomic read-compare-write; the javadoc directs shared-storage implementations to override it with an atomic conditional write), `InMemoryAccessTokenStore` overrides it with a CAS loop, and the three providers now store fetches via `putIfNewer`. `put` stays an unconditional replace, so an older-completing refresh can no longer pin an earlier deadline.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-common/src/main/java/io/agentscope/extensions/channel/common/AccessTokenStore.java
Buktal: Restated on the method contracts: `get()` and `put()` now both say the instance holds a single credential's token and must not be shared across providers built from different credentials, and the `fromProperties` `@param tokenStore` docs carry the same rule.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/test/java/io/agentscope/extensions/channel/wecom/WeComAccessTokenProviderTest.java
Buktal: Added `expiredEntryTriggersRefreshNotStaleToken` to all three provider tests (stale entry → refresh against the unreachable base throws, so removing the deadline check fails the test), plus `putIfNewer` semantics coverage in `InMemoryAccessTokenStoreTest` (older entry refused, fresher replaces).

oss-maintainer (approved):
### Summary
Re-reviewed the new head (`fef8d7bd`) after your follow-up commit — all three findings from the previous round are addressed, and the fix is the one I would have picked. No remaining blockers; LGTM.

### What changed since my review
- **[resolved] Un-keyed LWW slot** — `putIfNewer` now makes racing cold-start refreshes monotonic in `refreshAtMs`, so an older token can no longer pin the cache ahead of a fresher one. The `InMemoryAccessTokenStore` CAS loop keeps the guarantee genuinely atomic in-process, and the interface javadoc explicitly tells shared-storage implementations to override it with a conditional write rather than inheriting the non-atomic default — that is the right division of labour for an SPI.
- **[resolved] One credential per store instance** — the constraint is now stated on the type and repeated at both `get()` and `put()`, which is where a reader actually looks before injecting a shared store into two credentials. Documenting rather than keying is a defensible choice for v1 given every provider constructs its own store.
- **[resolved] Missing expiry-branch coverage** — `expiredEntryTriggersRefreshNotStaleToken` on all three providers exercises exactly the branch the refactor touched (`refreshAtMs` in the past → refresh attempted, not the stale token returned), and `putIfNewerStoresOnlyFresherEntries` pins the new contract at the store level.

### Cross-repo note
Nothing blocking, but worth a follow-up outside this PR: the v2 docs (`docs/v2/{en,zh}/integration/channel/index.md`) describe the store as pluggable for multi-instance deployments without stating the CAS-script requirement for custom implementations. Whoever writes the first Redis store should be pointed at the `putIfNewer` javadoc, since the default implementation is silently wrong under concurrency there.

CI green (7/7 SUCCESS) and CLA signed on this head.

---

Issue #3187, opened 2026-09-17
[Feature]: Per-tenant channel credential resolution at callback dispatch


Buktal:
**Is your feature request related to a problem? Please describe.**

The channel extensions bind credentials at construction time. `*Channel.fromProperties(...)` builds the callback crypto and the outbound API client (with its access-token cache) as immutable fields, and the callback controllers resolve channels from a process-wide registry keyed by channel id. That fits deployments whose credentials are known at startup. It does not fit multi-tenant bot platforms: we run a service where each customer connects their own WeCom corp, Feishu app or DingTalk enterprise app, so credentials are runtime data — created when a tenant onboards, rotated while the process serves traffic, and removed when the tenant leaves.

Today every one of those operations has to be expressed as "rebuild the channel and replace it":

- A rebuilt channel starts with a fresh deduplication store and bot-loop guard, so platform redeliveries that arrive around the swap window can be processed twice.
- The swap races in-flight callbacks: a request that already resolved the old channel object verifies and decrypts with credentials that may have just been rotated out.
- The process-wide registry accumulates one channel per tenant for the lifetime of the process, and the application has to drive construction, registration and teardown itself.

**Describe the solution you'd like**

An opt-in per-tenant wiring for the callback controllers, with the same shape in the wecom, feishu and dingtalk extensions:

1. A `*CredentialResolver` SPI — `Optional<*ChannelProperties> resolve(String tenantKey)`. The application implements it against its own tenant store; storage, caching and lookup policy stay application-owned.
2. A `*TenantChannelManager` wired into the callback controller (optional constructor; the default registry wiring is unchanged):
   - materializes a tenant's channel on first use, atomically per tenant key;
   - consults the resolver on every callback and refreshes the channel's credentials **in place** when they changed — the channel object is never replaced, so deduplication state and the bot-loop guard survive every rotation;
   - refresh is grouped by what the credentials feed: rotating a callback-only secret keeps the cached access token, rotating the app secret discards it;
   - an explicit `evict(tenantKey)` releases a departed tenant's channel and token cache; a resolver that transiently returns empty must not tear down a live tenant.
3. Materialized channels are owned by the manager — they are registered neither in the process-wide channel registry nor in a harness `ChannelManager`.
4. The tenant key doubles as the channel id, so sessions, deduplication keys and bot-loop guards are namespaced per tenant by construction.
[...]

PR #3188, opened 2026-09-17
feat(wecom): per-tenant credential resolution at callback dispatch

Buktal:
## AgentScope-Java Version

## Description

Part of #3187

WeCom channel credentials are currently build-time data: `WeComChannel.fromProperties(...)` constructs the callback crypto and the outbound client (with its token cache) as immutable fields, and `WeComCallbackController` resolves channels from the process-wide registry. In a multi-tenant deployment — one process serving many WeCom corps, with credentials added, rotated and removed at runtime — that forces a rebuild-and-replace per change, which resets deduplication state, races in-flight callbacks and grows the registry forever.

This PR adds an opt-in per-tenant wiring:

- **`WeComCredentialResolver`** — SPI, `Optional<WeComChannelProperties> resolve(String tenantKey)`. Storage and lookup policy stay application-owned.
- **`WeComTenantChannelManager`** — materializes a tenant's channel on first use (atomic per tenant key), consults the resolver on every callback, refreshes credentials in place when they changed, and exposes `evict(tenantKey)` for offboarding. A resolver that transiently returns empty does not tear down a live tenant's channel. Resolve-and-refresh is serialized per tenant, so when callbacks race a rotation the snapshot the tenant keeps is the newest resolver answer rather than the one whose thread happened to arrive last.
- **`WeComCallbackController`** — optional constructor taking the manager (`/api/channels/wecom/{tenantKey}/callback`); the default registry wiring is unchanged. The GET verification handshake resolves tenant credentials too. The manager constructor carries `@Autowired(required = false)`, so component scanning selects it whenever a `WeComTenantChannelManager` bean exists and falls back to the no-argument constructor when none does — neither mode needs a component-scan exclusion or a second controller bean.
- **`WeComChannel`** — credentials live in an immutable snapshot; `refreshCredentials` swaps it under a lock, rebuilding only what the changed fields feed:

| Rotated | Rebuilt | Kept |
|---|---|---|
| callback `token` / `encodingAesKey` | callback crypto + inbound mapper | outbound client and cached access token |
| `secret` / `agentId` / `apiBase` | outbound client | — (access token re-minted on next send) |
| `corpId` (feeds both groups: crypto, mapper and outbound) | callback crypto + inbound mapper + outbound client | — (access token re-minted on next send) |

Each outbound generation mints into its own access-token store, so the one-store-per-credential contract of `AccessTokenStore` holds across a rotation: a request still holding the previous generation's client cannot seed the cache the new generation reads.

`WeComChannelProperties.toString()` masks `secret`, `token` and `encodingAesKey` while the routing fields stay visible, so a resolver result that reaches a log line does not leak credentials. This changes the rendering of a public record and is called out here deliberately.

Materialized channels are owned by the manager: they are registered neither in `WeComChannelRegistry` nor in a harness `ChannelManager`. The tenant key doubles as the channel id, so sessions, deduplication keys and bot-loop guards are namespaced per tenant by construction; two tenants whose user ids collide never share state.

**Backward compatibility:** purely additive — the default controller constructor and all existing `fromProperties` overloads behave exactly as before (a static-wiring regression test is included). The `fromProperties` overload that accepts an `AccessTokenStore` instance builds a channel that does not rotate in place, since one store may serve one credential only; its javadoc states this.

**How to test:** `mvn -pl agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom -am test` — 25 new tests:

- `WeComTenantChannelManagerTest` (12): materialization, resolver consulted per callback, in-place rotation (same object, new crypto takes effect), callback-credential rotation keeps the outbound runtime, secret rotation rebuilds it, empty resolution keeps a live tenant, resolver failure propagates, eviction drops the cached channel, resolve-and-refresh serialized per tenant (the newest answer is the one kept, and the resolver is never entered concurrently for one tenant), concurrent lookups during rotations keep serving the same channel.
- `WeComCallbackControllerTest` (9): end-to-end dispatch with resolved credentials (token request carries the tenant's corp id; session namespaced by tenant key), cross-tenant signature rejected, unknown tenant 404, rotation takes effect on the next callback without losing the token cache (3 API requests, not 4), secret rotation mints a fresh token, GET verify handshake resolves tenant credentials, the same msg id from two tenants is dispatched twice, a reply keeps the credential generation its request entered with, static registry wiring regression.
- `WeComChannelCredentialRotationTest` (2): a callback-only rotation keeps minting into the same token store; a rotated app credential never reads the previous generation's token store.
- `WeComCallbackControllerSpringWiringTest` (2): constructor selection with and without a `WeComTenantChannelManager` bean.

Documentation: "Multi-tenant deployments" section added to `docs/v2/{en,zh}/integration/channel/wecom.md`.

## Checklist


agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComChannel.java
```diff
@@ -249,17 +271,114 @@ public void deliver(OutboundAddress address, List<Msg> messages) {
...
+     */
+    void refreshCredentials(WeComChannelProperties properties) {
+        Objects.requireNonNull(properties, "properties");
+        synchronized (credentialLock) {
+            Credentials current = credentials;
+            if (callbackCredentialsEqual(current.properties(), properties)
+                    && appCredentialsEqual(current.properties(), properties)) {
+                return;
+            }
+            if (!appCredentialsEqual(current.properties(), properties)) {
```
oss-maintainer: `refreshCredentials` reuses the same `AccessTokenStore` across app-credential generations. It clears the store on rotation and then builds a new `WeComAccessTokenProvider` with that same store, but an in-flight request holding the old `WeComOutboundClient` can refill the cleared slot with a token minted for the old `corpId/secret`. The next request then reads that stale token through the new provider and sends messages with an access token belonging to the rotated-out credentials. `AccessTokenStore` is documented as serving one credential only; each credential generation needs its own store instance (or the provider must key/cache by credential identity).

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComTenantChannelManager.java
```diff
@@ -0,0 +1,165 @@
...
+    public Optional<WeComChannel> channelFor(String tenantKey) {
+        Objects.requireNonNull(tenantKey, "tenantKey");
+        Optional<WeComChannelProperties> resolved = resolver.resolve(tenantKey);
+        if (resolved.isEmpty()) {
+            return Optional.empty();
+        }
+        WeComChannelProperties properties = resolved.get();
+        WeComChannel channel =
+                channels.computeIfAbsent(tenantKey, key -> materialize(key, properties));
+        channel.refreshCredentials(properties);
```
oss-maintainer: `channelFor` resolves credentials before `computeIfAbsent` and then refreshes the channel with that captured snapshot. Under concurrency, two callbacks for the same tenant can resolve different credential generations; the one whose `refreshCredentials` call runs last wins, even if its resolved properties are older than the other thread's resolver result. This can make the credential snapshot flap between generations and transiently reject valid callbacks signed with the fresher credentials. Re-resolving after materialization (or serializing resolution with the refresh) would guarantee the latest resolver result wins.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComCallbackController.java
```diff
@@ -53,15 +63,52 @@ public class WeComCallbackController {
...
+     * <p>All handlers are safe to invoke concurrently; resolution and credential refresh are
+     * serialized per tenant inside the manager, so this controller adds no shared mutable state.
+     *
+     * <p>Spring wiring: this constructor replaces the default one, so an application must expose
+     * exactly one {@link WeComCallbackController} bean. If component scanning also picks up the
+     * no-argument bean, both instances register the same request mappings and startup fails.
+     *
+     * @param manager the tenant channel manager, typically a singleton bean
+     */
+    public WeComCallbackController(WeComTenantChannelManager manager) {
```
oss-maintainer, line 93: Exposing both a public no-arg constructor and a public `WeComTenantChannelManager` constructor makes Spring wiring fragile. If component scanning covers this package, it will instantiate the no-arg version in addition to any user-defined multi-tenant bean, producing duplicate `@RequestMapping` registrations and startup failure. Consider making the no-arg constructor package-private for test use, or annotate the multi-tenant constructor with `@Autowired`/`@Primary` and document the required component-scan exclusion more prominently.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/test/java/io/agentscope/extensions/channel/wecom/WeComTenantChannelManagerTest.java
```diff
@@ -0,0 +1,225 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
+
+/**
+ * Tests for {@link WeComTenantChannelManager}: lazy materialization, in-place credential refresh and
+ * eviction.
+ */
+class WeComTenantChannelManagerTest {
```
oss-maintainer, line 61: The tests cover rotation, eviction, and reuse, but do not exercise the concurrent path where two callbacks for the same tenant resolve to different credential generations while one is materializing the channel. A stress test that drives `channelFor` from multiple threads during a rotation would verify the documented last-write-wins behavior and catch races like the stale-properties overwrite described above.

oss-maintainer (changes requested):
This PR adds multi-tenant WeCom callback support: `WeComCredentialResolver` resolves tenant credentials per request, `WeComTenantChannelManager` lazily materializes one `WeComChannel` per tenant, and `WeComChannel` refreshes its credential-derived collaborators in place so sessions, deduplication state, and the access-token cache survive rotations. The static wiring mode and public factory APIs remain backward compatible, and the new tests cover tenant isolation, rotation semantics, and static-mode regression.

**Verdict: REQUEST_CHANGES.** The design is sound and the tests are thorough, but the critical access-token-store sharing bug in `WeComChannel.refreshCredentials` can cause messages to be sent with an access token belonging to rotated-out credentials. That must be fixed before merge—each app-credential generation needs its own `AccessTokenStore` instance.

**Notable strengths:** clean per-request credential snapshots for inbound verify/decrypt/mapping; good separation of callback vs. app credential groups; clear documentation of Spring wiring constraints; per-tenant session namespacing via the tenant key as channel id.

### Additional note
- **[Info]** `WeComChannelProperties.java` — `WeComChannelProperties` is a Java record, so its auto-generated `toString()` includes `secret`, `token`, and `encodingAesKey`. In the multi-tenant flow the resolver returns these records and they may be passed through application/framework logging (e.g., `Optional`, exception messages, `@ControllerAdvice` loggers). Consider overriding `toString()` to mask the credential fields, or document that callers must never log resolver results directly.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComChannel.java
Buktal: Fixed — each credential generation now has its own store.

`WeComChannel` takes a `Supplier<AccessTokenStore>` instead of an instance and calls it on every rebuild of the outbound group, so a rotated secret produces a fresh store rather than clearing a shared one; the `clear()` call is gone. A request still holding the previous generation's client writes the token it minted into *that* generation's store, which nothing reads afterwards. The manager passes `InMemoryAccessTokenStore::new`, so the store implementation stays a caller decision. The public `fromProperties(..., AccessTokenStore)` overload wraps the given instance in a constant supplier, and its javadoc states that channels built that way do not rotate in place.

Regression tests in `WeComChannelCredentialRotationTest`:

- `aRotatedAppCredentialNeverReadsThePreviousGenerationsTokenStore` — rotates the secret, writes a token minted from the rotated-out secret into generation 1's store, then asserts generation 2 starts empty, mints its own token and sends with it. Against the previous code the send carries the stale token, so the test fails.
- `aCallbackOnlyRotationKeepsMintingIntoTheSameTokenStore` pins the other direction: a callback-only rotation must not rebuild the outbound runtime, so no second `gettoken` is issued.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComTenantChannelManager.java
Buktal: Fixed by serializing resolution with the refresh, as you suggested.

Resolve-and-refresh now runs under the tenant's credential lock, so a callback cannot ask the resolver while another callback for the same tenant is between its resolve and its refresh — the answer applied last is therefore the newest one, and the snapshot can no longer flap back to an older generation. The lock is the channel's own monitor: it exists only for tenants that resolved successfully, so tenant keys arriving on the URL path cannot grow manager-side lock state. The existing behaviour is preserved: a lookup that resolves empty serves no callback but leaves the live channel in place.

One detail worth stating: first use resolves twice — once to seed `computeIfAbsent`, once under the lock. The second result is the one the channel keeps, which is what makes a rotation racing materialization still land on the newest answer. The code comments say so, and the class javadoc now states that the resolver is consulted on every callback and that a lookup for one tenant waits out another lookup's resolver call for the same tenant.

Regression test: `WeComTenantChannelManagerTest.resolutionIsSerializedPerTenantSoTheNewestAnswerWins` — thread A stalls inside the resolver on the older generation, thread B looks the tenant up meanwhile, and the assertions are `maxConcurrent == 1` (the resolver is never entered concurrently for one tenant) plus the retained snapshot being B's newer answer. Against the previous code it reports max 2.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComCallbackController.java
Buktal: Fixed with `@Autowired(required = false)` on the manager constructor.

Spring treats an unsatisfiable non-required candidate as "not a candidate" and falls back to the no-argument constructor, so component scanning yields exactly one controller bean in both modes: multi-tenant when a `WeComTenantChannelManager` bean exists, static registry wiring otherwise. No component-scan exclusion is needed, and no second controller bean (which would register the same `@RequestMapping` twice and fail startup). Making the no-argument constructor package-private was the alternative, but it is the documented entry point for applications that register channels directly, so it stays public. Both constructors' javadoc and the class javadoc now state that an application exposes the manager bean and declares no other controller bean.

Test: `WeComCallbackControllerSpringWiringTest` runs the real constructor-resolution path (`DefaultListableBeanFactory` + `AutowiredAnnotationBeanPostProcessor`; spring-beans is already a dependency) in both directions — with the manager bean registered, the path segment resolves through the manager; with none registered, the controller is constructed and falls back to the registry.

The docs snippet was updated to a `@Bean WeComTenantChannelManager` method accordingly.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/test/java/io/agentscope/extensions/channel/wecom/WeComTenantChannelManagerTest.java
Buktal: Added, as two tests rather than one.

- `resolutionIsSerializedPerTenantSoTheNewestAnswerWins` is the deterministic one: the resolver counts overlapping entries and asserts it is never entered concurrently for a single tenant, then asserts the retained snapshot is the newer answer. It fails against the previous code (max concurrency 2).
- `concurrentLookupsDuringRotationsKeepServingTheSameChannel` is the stress test: 4 threads × 100 `channelFor` calls while one of them rotates the credentials every 10 iterations, asserting every lookup serves the same channel instance and nothing fails.

Buktal:
Disposition of the **[Info]** item from the review body, which has no inline thread to reply on:

`WeComChannelProperties.toString()` is now overridden to mask `secret`, `token` and `encodingAesKey`; `corpId`, `agentId`, `callbackPath` and `apiBase` stay visible, since those are what make a log line useful for locating the tenant. The record's component accessors are unchanged. Because this changes the rendering of a public record, it is also called out in the PR description rather than left for a reader to notice.

oss-maintainer (approved):
### Summary
Re-reviewed the force-pushed head (`092277e4`). All four items from my previous review are resolved — upgrading to approve.

- **AccessTokenStore generation isolation (was the blocker)** — `WeComChannel` now takes a `tokenStoreFactory` that is invoked once per credential generation (`newCredentials(...)`), so an in-flight request holding a rotated-out `WeComOutboundClient` can no longer repopulate or read the token slot of the current generation.
- **Rotation race in `channelFor`** — resolve-and-refresh now happens inside `synchronized (channel.credentialLock())`, and `refreshCredentials` short-circuits when both credential groups are unchanged, so the last writer can no longer apply a stale snapshot over a newer one while the first materialization seeds outside the lock by design.
- **Spring wiring** — the tenant-aware constructor is `@Autowired(required = false)` (`WeComCallbackController.java:92`), removing the ambiguous-constructor / duplicate `@RequestMapping` failure mode; `WeComCallbackControllerSpringWiringTest` pins both the static and multi-tenant shapes.
- **Concurrency coverage** — `WeComTenantChannelManagerTest` now drives `channelFor` from multiple threads across credential generations (latched resolver + `maxConcurrent == 1` assertion), which is the stress case I asked for.

Credential handling is otherwise good: per-request snapshots for verify/decrypt, tenant key as channel id for session/dedup namespacing, and no secrets in log output.

LGTM.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComChannel.java
```diff
@@ -146,34 +159,53 @@ public static WeComChannel fromProperties(
...
-        WeComOutboundClient outbound =
-                new WeComOutboundClient(props.apiBase(), tokenProvider, props.agentId());
-        WeComInboundMapper mapper = new WeComInboundMapper(channelId, props.corpId());
+        Objects.requireNonNull(tokenStore, "tokenStore");
+        return fromProperties(
+                channelId,
+                routing,
+                WeComChannelProperties.from(channelId, rawProperties),
+                idempotency,
+                properties -> tokenStore);
```
oss-maintainer, line 179: **[Warning] The fixed-store overload satisfies the per-generation contract only by never rotating — and nothing keeps that promise if a caller ever does.**

```java
}
```

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComChannel.java
```diff
@@ -249,17 +301,125 @@ public void deliver(OutboundAddress address, List<Msg> messages) {
...
+        WeComOutboundClient outbound =
+                appUnchanged
+                        ? previous.outboundClient()
+                        : new WeComOutboundClient(
+                                properties.apiBase(),
+                                new WeComAccessTokenProvider(
+                                        properties.apiBase(),
+                                        properties.corpId(),
+                                        properties.secret(),
+                                        tokenStoreFactory.apply(properties)),
```
oss-maintainer, line 398: **[Warning] `tokenStoreFactory.apply(...)` now runs under `credentialLock`, so an application-supplied shared-storage store can block the callback path.**

Either is sufficient:

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-wecom/src/main/java/io/agentscope/extensions/channel/wecom/WeComTenantChannelManager.java
```diff
@@ -0,0 +1,229 @@
...
+            WeComCredentialResolver resolver,
+            ChannelConfig routing,
+            Gateway gateway,
+            InboundEventDeduplicator idempotency) {
+        this(
+                resolver,
+                routing,
+                gateway,
+                idempotency,
+                (tenantKey, properties) -> new InMemoryAccessTokenStore());
```
oss-maintainer, line 128: **[Info] The default factory ignores both of its arguments — correct here, but the signature invites the implementation that is not.**

Discarding `tenantKey` and `properties` to return a fresh `WeComInMemoryAccessTokenStore`-style instance is exactly right for the default. The thing worth noting is that a reader implementing the shared-storage case is handed a `tenantKey` and a properties object, and the natural-looking memoisation — `stores.computeIfAbsent(tenantKey, k -> sharedSlot(k))` — satisfies the signature, is thread-safe, and quietly reintroduces the cross-generation slot sharing that this refactor exists to prevent. The prose javadoc does warn against it; the type does not.

oss-maintainer (commented):
### Summary
Re-reviewed the new head `2e8ffa65` (one commit since my approval on `092277e4`). Widening the seam from `Supplier<AccessTokenStore>` to `Function<DingTalkChannelProperties, AccessTokenStore>`, and threading the tenant key through the manager, is the right call: without the generation's properties an application-side shared store cannot tell one credential from the next, so the previous shape could not express the contract at all. The overload is source-compatible for `fromProperties` callers (`() -> tokenStore` → `properties -> tokenStore`) and the two adapted tests plus the new `tokenStoreFactorySeesTenantAndGenerationAndRunsOncePerGeneration` pin the intended once-per-generation behaviour, including the AES-key-only rotation arm that must *not* rebuild the store. `Check License`, `Check Module Sync` and both builds are green and the CLA is signed.

- **[Warning]** `WeComChannel.java:179` — the fixed-store overload now satisfies "one store per credential generation" only because that construction never rotates; `refreshCredentials` is package-private on the same class, so the invariant is a convention rather than something enforced.
- **[Warning]** `WeComChannel.java:398` — `tokenStoreFactory.apply(...)` runs inside `synchronized (credentialLock)` on the callback thread (and inside the `computeIfAbsent` bin lock on first materialisation), which is a new place for application-supplied store construction to block.
- **[Info]** `WeComTenantChannelManager.java:128` — the default factory ignores both new arguments, which is correct; the shared-storage signature it advertises is where a memoised `tenantKey`-only implementation would silently reintroduce the cross-generation sharing this refactor exists to prevent.

---

PR #3189, opened 2026-09-17
feat(feishu): per-tenant credential resolution at callback dispatch

Buktal:
## AgentScope-Java Version

## Description

Part of #3187

Feishu channel credentials are currently build-time data: `FeishuChannel.fromProperties(...)` constructs the callback crypto and the outbound client (with its token cache) as immutable fields, and `FeishuCallbackController` resolves channels from the process-wide registry. In a multi-tenant deployment — one process serving many Feishu apps, with credentials added, rotated and removed at runtime — that forces a rebuild-and-replace per change, which resets per-channel state, races in-flight callbacks and grows the registry forever.

This PR adds an opt-in per-tenant wiring:

- **`FeishuCredentialResolver`** — SPI, `Optional<FeishuChannelProperties> resolve(String tenantKey)`. Storage and lookup policy stay application-owned.
- **`FeishuTenantChannelManager`** — materializes a tenant's channel on first use (atomic per tenant key), consults the resolver on every callback, refreshes credentials in place when they changed, and exposes `evict(tenantKey)` for offboarding. A resolver that transiently returns empty does not tear down a live tenant's channel. Resolution and refresh run as one critical section on the channel's own credential lock, so the snapshot a tenant keeps is the newest resolver answer rather than the last thread to arrive; the lock lives on the channel, which exists only for a tenant that resolved, so unresolved tenant keys from the callback URL accumulate no state.
- **`FeishuCallbackController`** — optional constructor taking the manager (`/api/channels/feishu/{tenantKey}/callback`); the default registry wiring is unchanged. The constructor is annotated `@Autowired(required = false)`: an unsatisfiable non-required candidate is not a candidate, so component scanning falls back to the no-argument constructor when no manager bean exists, while an application that exposes one gets the tenant route without declaring a controller bean. The `url_verification` challenge handshake resolves tenant credentials too.
- **`FeishuChannel`** — credentials live in an immutable snapshot; `refreshCredentials` swaps it under a lock, rebuilding only what the changed fields feed:

| Rotated | Rebuilt | Kept |
|---|---|---|
| callback `encryptKey` / `verificationToken` | callback crypto (nullable for plaintext tenants) | outbound client and cached access token |
| `appSecret` / `appId` / `apiBase` | outbound client, with an access-token store of its own | — (access token re-minted on next send) |

A token store serves one credential, so each credential generation gets a store of its own: `FeishuChannel` takes a `Supplier<AccessTokenStore>` and calls it once per rebuild of the outbound group. An in-flight request holding the previous snapshot therefore cannot leave the token it minted from the rotated-out secret in the slot the new generation reads.

Each callback captures one credential snapshot and threads it through signature verification, decryption, token matching and dispatch (`dispatch(InboundMessage, Credentials)`), so a rotation landing mid-request cannot mix generations between the steps that verify a callback and the reply it produces.

The inbound mapper is credential-independent, so it stays channel-level and is never rebuilt. Materialized channels are owned by the manager: they are registered neither in `FeishuChannelRegistry` nor in a harness `ChannelManager`. The tenant key doubles as the channel id, so sessions and bot-loop guards are namespaced per tenant by construction; deduplication is delegated to the durable intake, as in the static wiring.

`FeishuChannelProperties.toString()` masks `appSecret`, `encryptKey` and `verificationToken`, keeping `appId` and the routing fields visible — a resolver result reaches application and framework logging (inside an `Optional`, an exception message, a controller-advice logger) where the generated rendering would print the secrets in full.

**Backward compatibility:** purely additive — the default controller constructor and all existing `fromProperties` overloads behave exactly as before (a static-wiring regression test is included). Tenants without an `encryptKey` (plaintext callbacks) are served through the same path. The public `fromProperties` overload still accepts a single `AccessTokenStore`: a channel built there serves that one credential and does not rotate in place.

**How to test:** `mvn -pl agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu -am test` — 25 new tests, 37 in the module, all green:

- `FeishuTenantChannelManagerTest` (12): materialization, resolver consulted per callback, in-place rotation (same object, new crypto takes effect), callback-credential rotation keeps the outbound runtime, secret rotation rebuilds it, empty resolution keeps a live tenant, resolver failure propagates, eviction drops the cached channel, resolve-and-refresh serialized per tenant so the newest answer wins, concurrent lookups during rotations keep serving the same channel.
- `FeishuCallbackControllerTest` (8): end-to-end dispatch with resolved credentials (token request carries the tenant's app id; session namespaced by tenant key), plaintext callback for an unencrypted tenant, cross-tenant signature rejected, unknown tenant 404, rotation takes effect on the next callback without losing the token cache (3 API requests, not 4), secret rotation mints a fresh token, the `url_verification` handshake resolves tenant credentials, static registry wiring regression.
- `FeishuChannelCredentialRotationTest` (3): a callback-only rotation keeps minting into the same token store, a rotated app credential never reads the previous generation's token store, a reply keeps the credential generation its request entered with.
- `FeishuCallbackControllerSpringWiringTest` (2): without a manager bean the static wiring serves the request, with one the multi-tenant wiring does.

Documentation: "Multi-tenant deployments" section added to `docs/v2/{en,zh}/integration/channel/feishu.md`.

## Checklist


agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuChannel.java
```diff
@@ -245,13 +279,105 @@ public void deliver(OutboundAddress address, List<Msg> messages) {
...
+     * </ul>
+     *
+     * <p>The channel object itself never changes identity, so a tenant keeps its sessions and
+     * bot-loop guard across a rotation, and a request in flight keeps verifying against the
+     * credential generation it started with. Safe to call concurrently; concurrent rotations follow
+     * last-write-wins.
+     *
+     * @param properties the tenant's current credentials
+     */
+    void refreshCredentials(FeishuChannelProperties properties) {
```
oss-maintainer, line 360: `refreshCredentials()` clears the shared `AccessTokenStore` and then rebuilds the outbound client with the same store instance (line 414). Because the store is not keyed by `appId`, an in-flight request holding the old `Credentials` snapshot can still fetch a token with the rotated-out `appId`/`appSecret` and write it into the same cleared slot, causing the new outbound client to use a stale or compromised token.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuTenantChannelManager.java
```diff
@@ -0,0 +1,164 @@
...
+    public Optional<FeishuChannel> channelFor(String tenantKey) {
+        Objects.requireNonNull(tenantKey, "tenantKey");
+        Optional<FeishuChannelProperties> resolved = resolver.resolve(tenantKey);
+        if (resolved.isEmpty()) {
+            return Optional.empty();
+        }
+        FeishuChannelProperties properties = resolved.get();
+        FeishuChannel channel =
+                channels.computeIfAbsent(tenantKey, key -> materialize(key, properties));
+        channel.refreshCredentials(properties);
```
oss-maintainer: `channelFor()` resolves credentials before `computeIfAbsent` and then applies that captured snapshot via `refreshCredentials(...)`. Under concurrency a thread that resolved an older credential generation can overwrite a newer one already applied — ordering is last-writer-wins by arrival, not by freshness. Re-resolve after materialization (or serialize resolve+refresh) so the freshest resolver result always wins.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuCallbackController.java
```diff
@@ -55,15 +65,52 @@ public class FeishuCallbackController {
     private static final Logger log = LoggerFactory.getLogger(FeishuCallbackController.class);
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
-    private final FeishuChannelRegistry registry;
+    /** Resolves the channel serving a request's path segment; {@code null} when none exists. */
+    private final ChannelSource channelSource;
 
     public FeishuCallbackController() {
```
oss-maintainer, line 74: The class is annotated with `@RestController` and exposes both a public no-arg constructor (line 30) and a public `FeishuTenantChannelManager` constructor (line 49). If Spring component scanning is enabled alongside an explicit bean definition using the multi-tenant constructor, both instances will be registered and startup will fail with duplicate `@RequestMapping` mappings.

oss-maintainer (changes requested):
### Summary
Mirror of #3188/#3190 for Feishu: per-tenant credential resolution via `FeishuCredentialResolver` + lazily materialized `FeishuTenantChannelManager`, in-place `refreshCredentials` on an immutable snapshot, additive controller route. Design is sound and backward compatible, and the static-wiring regression test is a plus.

**Verdict: REQUEST_CHANGES** — the same three defects confirmed on the WeCom and DingTalk siblings are present here.

**Notable strengths:** per-request credential snapshots for verify/decrypt, tenant key as channel id (session/bot-loop namespacing), plaintext-callback tenants served through the same path, inbound mapper kept credential-independent.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuChannel.java
Buktal: Fixed — the store is no longer shared between credential generations. `FeishuChannel` now takes a `Supplier<AccessTokenStore>` and calls it once per rebuild of the outbound group; the `tokenStore.clear()` on rotation is gone, and each generation's token provider is built with a store of its own. An in-flight request holding the previous `Credentials` snapshot can still fetch a token with the rotated-out credentials, but it writes into the store that snapshot owns, which nothing reads any more.

The public `fromProperties` overload still accepts a single `AccessTokenStore` — a channel built there serves the one credential it is given and does not rotate in place, which its javadoc now states. Multi-tenant wiring goes through the manager, which passes `InMemoryAccessTokenStore::new` and so never shares a store across generations or across tenants.

Pinned by `FeishuChannelCredentialRotationTest.aRotatedAppCredentialNeverReadsThePreviousGenerationsTokenStore`: it rotates the app secret while a request on the previous generation is still in flight, has that request write the token it minted from the rotated-out secret into the store it holds, and asserts the new generation's client authenticates with a token it minted itself. Reverting the factory to a shared instance makes the test fail — the new generation then reads the stale (non-expired) entry and issues no token request at all.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuTenantChannelManager.java
Buktal: Fixed by serializing resolve-and-refresh. `channelFor` now runs the resolve and the refresh as one critical section on the channel's own credential lock, so the snapshot a tenant keeps is the newest resolver answer, not the one whose thread happened to arrive last.

The resolve before `computeIfAbsent` is still there, but it is no longer authoritative: it only seeds the materialization, because the channel that carries the lock does not exist until a resolve has succeeded. That is documented in the javadoc, together with the consequence for implementors — a resolver must tolerate being called more than once for a single callback.

The lock lives on the channel, so it exists only for a tenant that resolved, and unresolved tenant keys taken from the callback URL cannot grow a lock map.

Pinned by `FeishuTenantChannelManagerTest.resolutionIsSerializedPerTenantSoTheNewestAnswerWins` (the resolver records concurrent entries; `maxConcurrent == 1` is non-vacuous because the terminal assertion requires the second resolve to have run and been applied) and by `concurrentLookupsDuringRotationsKeepServingTheSameChannel` (4 threads × 100 lookups with rotations every 10).

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuCallbackController.java
Buktal: Fixed — the multi-tenant constructor is now annotated `@Autowired(required = false)`. An unsatisfiable non-required candidate is not a candidate, so Spring falls back to the no-argument constructor when no `FeishuTenantChannelManager` bean exists, and prefers the manager constructor when one does.

Both directions are covered by `FeishuCallbackControllerSpringWiringTest`, which drives a real `DefaultListableBeanFactory` with an `AutowiredAnnotationBeanPostProcessor`: with no manager bean the static registry serves the request, with one the injected manager does.

The duplicate-mapping failure you describe is real, but it is a separate mistake, and the javadoc and the docs now say so explicitly: exposing a second `FeishuCallbackController` bean (a `@Bean` method building one alongside component scanning) registers the same request mappings twice and fails startup. A multi-tenant application exposes the manager bean and nothing else.

Buktal:
Two further items from the same review round are applied here, and one coverage boundary is worth stating explicitly.

**One credential snapshot per request.** The controller captures `channel.credentials()` once and threads it through signature verification, decryption, verification-token matching and dispatch, through a new `dispatch(InboundMessage, Credentials)` overload. A rotation landing mid-request can no longer mix generations between the callback that was verified and the reply it produces. The single-argument `dispatch` still reads the live snapshot and delegates to the new overload, so existing callers are unaffected.

**`toString()` masking.** `FeishuChannelProperties.toString()` renders `appSecret`, `encryptKey` and `verificationToken` as `***`, keeping `appId` and the routing fields visible. A resolver result reaches application and framework logging — inside an `Optional`, an exception message, a controller-advice logger — where the generated record rendering would print the secrets in full.

**Coverage boundary.** The per-request snapshot is pinned at the channel level (`FeishuChannelCredentialRotationTest.aReplyKeepsTheCredentialGenerationItsRequestEnteredWith`) but not at the controller level: that test calls the two-argument overload directly, and no test would fail if the controller reverted to `dispatch(in)`. The controller consults no injectable collaborator between `channel.credentials()` and the dispatch — the mapper and the bot-loop guard are channel internals, and deduplication is delegated to the durable intake — while `FeishuChannel` and `FeishuInboundMapper` are `final` with construction that is not open to tests, so there is no seam through which a test could force a rotation between the capture and the dispatch. Adding one would be production surface for a single test, so we are disclosing the gap rather than closing it.

Test counts: 25 new tests in this PR, 37 in the module, all green.

oss-maintainer (approved):
### Summary
Re-reviewed the force-pushed head (`702621b3`). The three defects behind my previous REQUEST_CHANGES are fixed — upgrading to approve.

- **Shared token store across generations** — `FeishuChannel` now receives a `tokenStoreFactory` that builds a fresh `AccessTokenStore` per credential snapshot (`newCredentials(properties, previous, tokenStoreFactory)`), so an old `Credentials` snapshot can no longer mint or serve tokens through the current generation's store.
- **Ordering in `channelFor`** — resolve + `refreshCredentials` are serialized on `credentialLock`, so a thread that resolved an older generation can no longer overwrite a newer applied snapshot.
- **Spring wiring** — the tenant-aware constructor is `@Autowired(required = false)` (`FeishuCallbackController.java:94`) and `FeishuCallbackControllerSpringWiringTest` covers both wiring modes, so component scanning plus an explicit multi-tenant bean no longer produces two controllers with colliding routes.

`FeishuTenantChannelManagerTest` also gained the multi-threaded rotation stress test, matching #3188. Approving.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuChannel.java
```diff
@@ -145,25 +173,43 @@ public static FeishuChannel fromProperties(
...
-                        props.apiBase(), props.appId(), props.appSecret(), tokenStore);
-        FeishuOutboundClient outbound = new FeishuOutboundClient(props.apiBase(), tokenProvider);
-        FeishuInboundMapper mapper = new FeishuInboundMapper(channelId);
+        Objects.requireNonNull(tokenStore, "tokenStore");
+        return fromProperties(
+                channelId,
+                routing,
+                FeishuChannelProperties.from(channelId, rawProperties),
+                idempotency,
+                properties -> tokenStore);
```
oss-maintainer, line 182: **[Warning] The fixed-store overload satisfies the per-generation contract only by never rotating — and nothing keeps that promise if a caller ever does.**

```java
}
```

oss-maintainer (commented):
### Summary
Re-reviewed the new head `8e47900d` (one commit since my approval on `702621b3`). Widening the seam from `Supplier<AccessTokenStore>` to `Function<DingTalkChannelProperties, AccessTokenStore>`, and threading the tenant key through the manager, is the right call: without the generation's properties an application-side shared store cannot tell one credential from the next, so the previous shape could not express the contract at all. The overload is source-compatible for `fromProperties` callers (`() -> tokenStore` → `properties -> tokenStore`) and the two adapted tests plus the new `tokenStoreFactorySeesTenantAndGenerationAndRunsOncePerGeneration` pin the intended once-per-generation behaviour, including the AES-key-only rotation arm that must *not* rebuild the store. `Check License`, `Check Module Sync` and both builds are green and the CLA is signed.

- **[Warning]** `FeishuChannel.java:182` — the fixed-store overload now satisfies "one store per credential generation" only because that construction never rotates; `refreshCredentials` is package-private on the same class, so the invariant is a convention rather than something enforced.
- **[Warning]** `FeishuChannel.java:399` — `tokenStoreFactory.apply(...)` runs inside `synchronized (credentialLock)` on the callback thread (and inside the `computeIfAbsent` bin lock on first materialisation), which is a new place for application-supplied store construction to block.
- **[Info]** `FeishuTenantChannelManager.java:132` — the default factory ignores both new arguments, which is correct; the shared-storage signature it advertises is where a memoised `tenantKey`-only implementation would silently reintroduce the cross-generation sharing this refactor exists to prevent.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuChannel.java
```diff
@@ -245,13 +309,116 @@ public void deliver(OutboundAddress address, List<Msg> messages) {
...
+        FeishuOutboundClient outbound =
+                appUnchanged
+                        ? previous.outboundClient()
+                        : new FeishuOutboundClient(
+                                properties.apiBase(),
+                                new FeishuAccessTokenProvider(
+                                        properties.apiBase(),
+                                        properties.appId(),
+                                        properties.appSecret(),
+                                        tokenStoreFactory.apply(properties)));
```
oss-maintainer, line 399: **[Warning] `tokenStoreFactory.apply(...)` now runs under `credentialLock`, so an application-supplied shared-storage store can block the callback path.**

Either is sufficient:

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-feishu/src/main/java/io/agentscope/extensions/channel/feishu/FeishuTenantChannelManager.java
```diff
@@ -0,0 +1,241 @@
...
+            FeishuCredentialResolver resolver,
+            ChannelConfig routing,
+            Gateway gateway,
+            InboundEventDeduplicator idempotency) {
+        this(
+                resolver,
+                routing,
+                gateway,
+                idempotency,
+                (tenantKey, properties) -> new InMemoryAccessTokenStore());
```
oss-maintainer, line 132: **[Info] The default factory ignores both of its arguments — correct here, but the signature invites the implementation that is not.**

Discarding `tenantKey` and `properties` to return a fresh `FeishuInMemoryAccessTokenStore`-style instance is exactly right for the default. The thing worth noting is that a reader implementing the shared-storage case is handed a `tenantKey` and a properties object, and the natural-looking memoisation — `stores.computeIfAbsent(tenantKey, k -> sharedSlot(k))` — satisfies the signature, is thread-safe, and quietly reintroduces the cross-generation slot sharing that this refactor exists to prevent. The prose javadoc does warn against it; the type does not.

PR #3190, opened 2026-09-17
feat(dingtalk): per-tenant credential resolution for http callbacks

Buktal:
## AgentScope-Java Version

## Description

Part of #3187

DingTalk channel credentials are currently build-time data: `DingTalkChannel.fromProperties(...)` constructs the callback crypto and the outbound client (with its token cache) as immutable fields, and `DingTalkCallbackController` resolves channels from the process-wide registry. In a multi-tenant deployment — one process serving many DingTalk enterprise apps, with credentials added, rotated and removed at runtime — that forces a rebuild-and-replace per change, which resets deduplication state, races in-flight callbacks and grows the registry forever.

This PR adds an opt-in per-tenant wiring for the **http callback mode** (a stream-mode WebSocket binds its credentials at connect time, so it cannot serve per-request resolution):

- **`DingTalkCredentialResolver`** — SPI, `Optional<DingTalkChannelProperties> resolve(String tenantKey)`. Storage and lookup policy stay application-owned; resolved properties must configure `mode=http` — anything else fails fast with `IllegalArgumentException`.
- **`DingTalkTenantChannelManager`** — materializes a tenant's channel on first use (atomic per tenant key), consults the resolver on every callback, refreshes credentials in place when they changed, and exposes `evict(tenantKey)` for offboarding. A resolver that transiently returns empty does not tear down a live tenant's channel. Resolution and refresh are serialized per tenant on the channel's own credential lock, so the snapshot a tenant keeps is the newest resolver answer rather than the one whose thread arrived last; the lock lives on the channel, so it exists only for tenants that resolved and callback URL segments cannot grow a lock map. The single call that materializes a tenant resolves twice — the channel carrying the lock does not exist until a resolve has succeeded — so a resolver must tolerate being called more than once per callback.
- **`DingTalkCallbackController`** — optional constructor taking the manager (`/api/channels/dingtalk/{tenantKey}/callback`); the default registry wiring is unchanged. The multi-tenant constructor is `@Autowired(required = false)`, so component scanning prefers it when a manager bean exists and falls back to the no-argument constructor when none does. An unknown tenant answers 401, indistinguishable from a bad signature, so the endpoint is not a tenant-key oracle (same posture as the static controller).
- **`DingTalkChannel`** — credentials live in an immutable snapshot; `refreshCredentials` swaps it under a lock, rebuilding only what the changed fields feed. Each credential generation gets an `AccessTokenStore` of its own (the channel holds a `Supplier<AccessTokenStore>`), so a request still in flight on the previous generation cannot leave its token behind for the new one. The controller captures the snapshot once per request and threads it through signature verification, decryption, the intake and the reply, so a rotation landing mid-request cannot attribute the callback to another app key or send its reply through another client. Calling `refreshCredentials` on a stream-mode channel throws `IllegalStateException`, since its WebSocket is bound to its construction-time credentials. The `start()`/`stop()` mode discriminant moves from "has a callback crypto" to "has a stream client" — behavior-identical for existing stream and http channels.
- **`DingTalkChannelProperties.toString()`** — renders `appSecret` and `aesKey` as `***`, keeping `appKey` and the routing fields visible. A resolver result reaches application and framework logging (inside an `Optional`, an exception message, a controller-advice logger) where the generated record rendering would print the secrets in full.

The app secret feeds both groups, which makes the two rotation cases observable:

| Rotated | Rebuilt | Kept |
|---|---|---|
| `aesKey` | callback crypto | outbound client, and the cached access token with it |
| `appSecret` | callback crypto (its HMAC) **and** outbound client | — (the new generation mints into a token store of its own) |
| `appKey` / `robotCode` / `apiBase` | outbound client + inbound mapper | — (the new generation mints into a token store of its own) |

Materialized channels are owned by the manager: they are registered neither in `DingTalkChannelRegistry` nor in a harness `ChannelManager`. The tenant key doubles as the channel id, so sessions, deduplication keys and bot-loop guards are namespaced per tenant by construction; two tenants whose user ids collide never share state.

**Backward compatibility:** purely additive — the default controller constructor and all existing `fromProperties` overloads behave exactly as before; a channel built from an `AccessTokenStore` instance serves the one credential it is given and does not rotate in place (its javadoc says so), and the static registry wiring keeps its regression coverage in `DingTalkCallbackControllerTest`.

**How to test:** `mvn -pl agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk -am test` — 26 new tests, 63 in the module:

- `DingTalkTenantChannelManagerTest` (14): materialization, resolver consulted per callback, in-place rotation (same object, new crypto takes effect), aesKey rotation keeps the outbound runtime, secret rotation rebuilds both, `mode=stream` resolution rejected, stream channel cannot refresh in place, empty resolution keeps a live tenant, resolver failure propagates, eviction drops the cached channel, resolution serialized per tenant so the newest answer wins, concurrent lookups during rotations keep serving the same channel.
- `DingTalkTenantCallbackControllerTest` (7): end-to-end dispatch with resolved credentials (session namespaced by tenant key; account id is the tenant's app key), plaintext callback for a tenant without `aesKey`, cross-tenant signature rejected, unknown tenant 401, aesKey rotation takes effect on the next callback without losing the token cache (the old envelope fails decryption → 400; 3 API requests, not 4), secret rotation mints a fresh token (4 requests), the same msg id from two tenants is dispatched twice.
- `DingTalkChannelCredentialRotationTest` (3): a callback-only rotation keeps minting into the same token store, a rotated app credential never reads the previous generation's store, a callback keeps the credential generation its request entered with.
- `DingTalkCallbackControllerSpringWiringTest` (2): with a manager bean the component-scanned controller is built by the multi-tenant constructor, without one it falls back to the static registry.

Documentation: "Multi-tenant deployments" section added to `docs/v2/{en,zh}/integration/channel/dingtalk.md`.

## Checklist


agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkChannel.java
```diff
@@ -302,39 +337,149 @@ void onInboundPayload(JsonNode payload) {
...
+        }
+        synchronized (credentialLock) {
+            Credentials current = credentials;
+            if (callbackCredentialsEqual(current.properties(), properties)
+                    && appCredentialsEqual(current.properties(), properties)) {
+                return;
+            }
+            if (!appCredentialsEqual(current.properties(), properties)) {
+                // The token minted from the old app credentials is invalid for the rotated ones.
+                tokenStore.clear();
```
oss-maintainer: `tokenStore.clear()` does not isolate credential generations because an old snapshot's `DingTalkAccessTokenProvider` retains the same store. If an old outbound token request finishes after this clear, it can repopulate the slot; the new provider then accepts that old token and sends as the newly configured bot. Use a generation-specific (or credential-keyed) token store, or guard writes by generation, and add a concurrent rotation/refresh test.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkChannel.java
```diff
@@ -279,7 +314,7 @@ void onInboundPayload(JsonNode payload) {
                     channelId);
             return;
         }
-        Optional<InboundMessage> inbound = mapper.map(payload);
+        Optional<InboundMessage> inbound = credentials.mapper().map(payload);
```
oss-maintainer: `credentials` is read again here even though the controller captured it before signature verification and decryption. A rotation between those operations can map an authenticated old-generation callback with the new `appKey` and make the reply use the new outbound client, breaking the claimed per-request snapshot and potentially producing the wrong account, session, or reply identity. Pass the captured `Credentials` through the HTTP intake and dispatch/reply path, and test this interleaving.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackController.java
```diff
@@ -51,15 +61,53 @@ public class DingTalkCallbackController {
...
+     * <p>All handlers are safe to invoke concurrently; resolution and credential refresh are
+     * serialized per tenant inside the manager, so this controller adds no shared mutable state.
+     *
+     * <p>Spring wiring: this constructor replaces the default one, so an application must expose
+     * exactly one {@link DingTalkCallbackController} bean. If component scanning also picks up the
+     * no-argument bean, both instances register the same request mappings and startup fails.
+     *
+     * @param manager the tenant channel manager, typically a singleton bean
+     */
+    public DingTalkCallbackController(DingTalkTenantChannelManager manager) {
```
oss-maintainer, line 92: With both unannotated constructors, Spring component scanning uses the no-argument constructor rather than injecting a `DingTalkTenantChannelManager`. Registering a second controller bean then creates the duplicate request mappings described above, so conventional multi-tenant Spring wiring either remains static or fails to start. Expose mutually exclusive or conditional controller beans (or otherwise make constructor selection explicit) and add an `ApplicationContext` test.

oss-maintainer (commented):
PR #3190 adds a `DingTalkCredentialResolver` and tenant channel manager that lazily materialize HTTP-mode channels and refresh their credential-derived collaborators. **COMMENT**: concurrent app-credential rotation can restore and consume an old token, a callback can span multiple credential snapshots, and the advertised Spring wiring cannot select the tenant manager through ordinary component scanning. Three findings are warnings worth addressing; none is proven to be a merge blocker on its own, so posting as COMMENT. Strengths include tenant-key namespacing for channels and deduplication, HTTP-mode enforcement, and tests covering normal tenant isolation and serial credential rotations.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkChannel.java
Buktal: Fixed — the store is no longer shared between credential generations. `DingTalkChannel` now takes a `Supplier<AccessTokenStore>` and calls it once per rebuild of the outbound group; the `tokenStore.clear()` on rotation is gone, and each generation's token provider is built with a store of its own. An in-flight request holding the previous `Credentials` snapshot can still fetch a token with the rotated-out credentials, but it writes into the store that snapshot owns, which nothing reads any more.

The public `fromProperties` overload still accepts a single `AccessTokenStore` — a channel built there serves the one credential it is given and does not rotate in place, which its javadoc now states. Multi-tenant wiring goes through the manager, which passes `InMemoryAccessTokenStore::new` and so never shares a store across generations or across tenants.

Pinned by `DingTalkChannelCredentialRotationTest.aRotatedAppCredentialNeverReadsThePreviousGenerationsTokenStore`: it rotates the app key and the app secret while a request on the previous generation is still in flight, has that request write the token it minted from the rotated-out secret into the store it holds, and asserts the new generation minted into a store of its own. Reverting the factory to one shared store per channel makes the test fail — the rotated generation reuses the previous generation's store (`expected: <2> but was: <1>`).

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkChannel.java
Buktal: Fixed by threading the captured snapshot through the whole request path. `DingTalkChannel` gained package-private `onInboundPayload(JsonNode, Credentials)` and `dispatch(InboundMessage, Credentials)`; the controller passes the snapshot it verified and decrypted with into the intake, which now maps with that generation's mapper and dispatches with it, so the reply leaves through the same generation's client. The single-argument overloads still read the live snapshot, so the stream-mode path (`DingTalkStreamClient`) is unaffected.

This is the interleaving you asked to see tested: `DingTalkChannelCredentialRotationTest.aCallbackKeepsTheCredentialGenerationItsRequestEnteredWith` rotates the app key and secret after the request entered, then feeds the intake with the captured snapshot. The callback is still attributed to the pre-rotation app key and its reply authenticates with that generation's cached token, so no token request is issued; the post-rotation app key and its empty store never appear. Reverting the intake to read the live snapshot fails the test at the account assertion (`expected: <app-rot> but was: <app-rot2>`).

**Coverage boundary.** That test pins the channel-level contract (`onInboundPayload(payload, snapshot)`), and it calls the two-argument overload directly — no test would fail if the controller reverted to `onInboundPayload(payload)`. The controller consults no injectable collaborator between `channel.credentials()` and the intake — the mapper and the bot-loop guard are channel internals, and deduplication is delegated to the durable intake — while `DingTalkChannel` and `DingTalkInboundMapper` are `final` with construction that is not open to tests, so there is no seam through which a test could force a rotation between the capture and the intake. Adding one would be production surface for a single test, so we are disclosing the gap rather than closing it.

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkCallbackController.java
Buktal: Fixed — the multi-tenant constructor is now annotated `@Autowired(required = false)`. An unsatisfiable non-required candidate is not a candidate, so Spring falls back to the no-argument constructor when no `DingTalkTenantChannelManager` bean exists, and prefers the manager constructor when one does.

Both directions are covered by `DingTalkCallbackControllerSpringWiringTest`, which drives a real `DefaultListableBeanFactory` with an `AutowiredAnnotationBeanPostProcessor`: with no manager bean the static registry serves the request, with one the injected manager does. Removing the annotation makes the manager direction fail (`expected: <DingTalkChannel@…> but was: <null>`) while the static direction stays green, which is the exact discrimination point.

The duplicate-mapping failure you describe is real, but it is a separate mistake, and the javadoc and the docs now say so explicitly: exposing a second `DingTalkCallbackController` bean registers the same request mappings twice and fails startup. A multi-tenant application exposes the manager bean and nothing else.

Buktal:
Four items beyond the three inline threads: one fix that had no thread of its own, two behaviours worth stating explicitly, and what the new tests do and do not pin.

**Resolution order.** `channelFor` now runs the resolve and the refresh as one critical section on the channel's own credential lock, so the snapshot a tenant keeps is the newest resolver answer, not the one whose thread happened to arrive last. The resolve before `computeIfAbsent` is still there but is no longer authoritative: it only seeds the materialization, because the channel that carries the lock does not exist until a resolve has succeeded. That is documented in the javadoc, together with the consequence for implementors — a resolver must tolerate being called more than once for a single callback. The lock lives on the channel, so it exists only for a tenant that resolved, and unresolved tenant keys taken from the callback URL cannot grow a lock map. Pinned by `DingTalkTenantChannelManagerTest.resolutionIsSerializedPerTenantSoTheNewestAnswerWins` (the resolver records concurrent entries; `maxConcurrent == 1` is non-vacuous because the terminal assertion requires the second resolve to have run and been applied) and by `concurrentLookupsDuringRotationsKeepServingTheSameChannel` (4 threads × 100 lookups with rotations every 10). One ordering consequence is worth recording: a resolver that answers `mode=http` on the first call and `mode=stream` on the second — within a single `channelFor` — now materializes the tenant's channel and then fails fast on the second answer, leaving that channel cached. The fail-fast is unchanged; what is new is that a rejected resolution can leave a materialized http channel behind. It is unreachable for a resolver that answers consistently, which the new contract language asks for, so it is recorded here rather than guarded against.

**One credential snapshot per request.** The controller captures `channel.credentials()` once and passes it into the intake, which maps and dispatches on it, so a rotation landing mid-request can no longer attribute the callback to another app key or send its reply through another client.

**`toString()` masking.** `DingTalkChannelProperties.toString()` renders `appSecret` and `aesKey` as `***`, keeping `appKey` and the routing fields visible. A resolver result reaches application and framework logging — inside an `Optional`, an exception message, a controller-advice logger — where the generated record rendering would print the secrets in full.

**Coverage boundary.** The per-request snapshot is pinned at the channel level (`DingTalkChannelCredentialRotationTest.aCallbackKeepsTheCredentialGenerationItsRequestEnteredWith`) but not at the controller level: that test calls the two-argument intake directly, and no test would fail if the controller reverted to `onInboundPayload(payload)`. The controller consults no injectable collaborator between `channel.credentials()` and the intake — the mapper and the bot-loop guard are channel internals, and deduplication is delegated to the durable intake — while `DingTalkChannel` and `DingTalkInboundMapper` are `final` with construction that is not open to tests, so there is no seam through which a test could force a rotation between the capture and the intake. Adding one would be production surface for a single test, so we are disclosing the gap rather than closing it. Two further substitutions of the same kind: store isolation is pinned deterministically (the in-flight token write is simulated by a direct `put` rather than raced), with concurrency covered separately by the manager's stress and serialization tests; and constructor selection is exercised through `DefaultListableBeanFactory` rather than a full `ApplicationContext`, since this module declares `spring-webflux` but not `spring-context` — the same constructor-resolution code path Spring itself uses.

**On duplication.** `DingTalkChannelCredentialRotationTest` is the third isomorphic copy of the same test, after WeCom's and Feishu's, and the manager/resolver/seam family they exercise exists once per provider module. The channel modules share no test-jar, so the scaffolding cannot be hoisted without one. The parallel implementation is deliberate and its shape is identical across the three, so it is reported here rather than consolidated unilaterally — a shared `channel-common` extraction is a maintainer decision, not something to fold into a provider PR.

Test counts: 26 new tests in this PR, 63 in the module, all green.

oss-maintainer (approved):
### Summary
Re-reviewed the force-pushed head (`067c4e66`). All three findings from my previous COMMENT are addressed — upgrading to approve.

- **Generation isolation of the token store** — `tokenStoreFactory` is now called once per credential generation and `tokenStore.clear()` is gone from the rotation path, so a late-finishing token request from an old outbound client can no longer repopulate the slot the new provider reads.
- **Callback spanning multiple snapshots** — the callback path now threads a single `Credentials` snapshot through `onInboundPayload` / `dispatch` / `sendReply` as an explicit parameter instead of re-reading the volatile field, so a mid-request rotation can no longer pair an old-generation authenticated payload with the new `appKey`/outbound client.
- **Spring wiring** — the tenant-aware constructor is `@Autowired(required = false)` (`DingTalkCallbackController.java:91`) and `DingTalkCallbackControllerSpringWiringTest` asserts the single-bean wiring for both static and per-tenant modes.

Rotation semantics, HTTP-mode enforcement and secret handling all look consistent with #3188/#3189. LGTM.

---

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkChannel.java
```diff
@@ -160,24 +189,42 @@ public static DingTalkChannel fromProperties(
...
-        DingTalkOutboundClient outbound =
-                new DingTalkOutboundClient(props.apiBase(), tokenProvider, props.robotCode());
-        DingTalkInboundMapper mapper = new DingTalkInboundMapper(channelId, props.appKey());
+        Objects.requireNonNull(tokenStore, "tokenStore");
+        return fromProperties(
+                channelId,
+                routing,
+                DingTalkChannelProperties.from(channelId, rawProperties),
+                idempotency,
+                properties -> tokenStore);
```
oss-maintainer, line 198: **[Warning] The fixed-store overload satisfies the per-generation contract only by never rotating — and nothing keeps that promise if a caller ever does.**

```java
}
```

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkChannel.java
```diff
@@ -302,39 +382,158 @@ void onInboundPayload(JsonNode payload) {
...
+        DingTalkOutboundClient outbound =
+                appUnchanged
+                        ? previous.outboundClient()
+                        : new DingTalkOutboundClient(
+                                properties.apiBase(),
+                                new DingTalkAccessTokenProvider(
+                                        properties.apiBase(),
+                                        properties.appKey(),
+                                        properties.appSecret(),
+                                        tokenStoreFactory.apply(properties)),
```
oss-maintainer, line 496: **[Warning] `tokenStoreFactory.apply(...)` now runs under `credentialLock`, so an application-supplied shared-storage store can block the callback path.**

Either is sufficient:

agentscope-extensions/agentscope-extensions-channel/agentscope-extensions-channel-dingtalk/src/main/java/io/agentscope/extensions/channel/dingtalk/DingTalkTenantChannelManager.java
```diff
@@ -0,0 +1,266 @@
...
+            DingTalkCredentialResolver resolver,
+            ChannelConfig routing,
+            Gateway gateway,
+            InboundEventDeduplicator idempotency) {
+        this(
+                resolver,
+                routing,
+                gateway,
+                idempotency,
+                (tenantKey, properties) -> new InMemoryAccessTokenStore());
```
oss-maintainer, line 140: **[Info] The default factory ignores both of its arguments — correct here, but the signature invites the implementation that is not.**

Discarding `tenantKey` and `properties` to return a fresh `DingTalkInMemoryAccessTokenStore`-style instance is exactly right for the default. The thing worth noting is that a reader implementing the shared-storage case is handed a `tenantKey` and a properties object, and the natural-looking memoisation — `stores.computeIfAbsent(tenantKey, k -> sharedSlot(k))` — satisfies the signature, is thread-safe, and quietly reintroduces the cross-generation slot sharing that this refactor exists to prevent. The prose javadoc does warn against it; the type does not.

oss-maintainer (commented):
### Summary
Re-reviewed the new head `f0cf8750` (one commit since my approval on `067c4e66`). Widening the seam from `Supplier<AccessTokenStore>` to `Function<DingTalkChannelProperties, AccessTokenStore>`, and threading the tenant key through the manager, is the right call: without the generation's properties an application-side shared store cannot tell one credential from the next, so the previous shape could not express the contract at all. The overload is source-compatible for `fromProperties` callers (`() -> tokenStore` → `properties -> tokenStore`) and the two adapted tests plus the new `tokenStoreFactorySeesTenantAndGenerationAndRunsOncePerGeneration` pin the intended once-per-generation behaviour, including the AES-key-only rotation arm that must *not* rebuild the store. `Check License`, `Check Module Sync` and both builds are green and the CLA is signed.

- **[Warning]** `DingTalkChannel.java:198` — the fixed-store overload now satisfies "one store per credential generation" only because that construction never rotates; `refreshCredentials` is package-private on the same class, so the invariant is a convention rather than something enforced.
- **[Warning]** `DingTalkChannel.java:496` — `tokenStoreFactory.apply(...)` runs inside `synchronized (credentialLock)` on the callback thread (and inside the `computeIfAbsent` bin lock on first materialisation), which is a new place for application-supplied store construction to block.
- **[Info]** `DingTalkTenantChannelManager.java:140` — the default factory ignores both new arguments, which is correct; the shared-storage signature it advertises is where a memoised `tenantKey`-only implementation would silently reintroduce the cross-generation sharing this refactor exists to prevent.

---

PR #3194, opened 2026-09-18
chore: bump revision to 2.0.4-SNAPSHOT

### LINKED ISSUE #3193: [enhancement]: bump revision to 2.0.4-SNAPSHOT after the 2.0.3 release
## Background

AgentScope Java `2.0.3` has been released to Maven Central (all modules, including `agentscope-bom`, now resolve to `2.0.3`. However, the `main` branch still carries:

- root `pom.xml`: `<revision>2.0.3-SNAPSHOT</revision>`
- `agentscope-dependencies-bom/pom.xml`: `<revision>2.0.3-SNAPSHOT</revision>`
- `agentscope-distribution/agentscope-bom/pom.xml`: `<revision>2.0.3-SNAPSHOT</revision>`

## Problem

Keeping `2.0.3-SNAPSHOT` on the development branch after `2.0.3` has shipped creates ambiguity:

1. **Version collision.** `2.0.3` (release) and `2.0.3-SNAPSHOT` (dev) coexist but point to different code states. Locally-installed SNAPSHOT artifacts are easy to confuse with the released ones when debugging.
2. **No development coordinate.** Work towards the next release has no reserved version; the bump would have to happen under time pressure during the next release train.
3. **Broken convention.** Standard Maven practice is to move `<revision>` to the next patch version immediately after a release, so `main` always represents "the next release under development".

## Proposal

Bump `<revision>` from `2.0.3-SNAPSHOT` to `2.0.4-SNAPSHOT` in all three places that define it:

- `pom.xml`
- `agentscope-dependencies-bom/pom.xml` (third-party dependency BOM)
- `agentscope-distribution/agentscope-bom/pom.xml` (the released `io.agentscope:agentscope-bom`, which manages AgentScope's own modules)

This is a three-file, three-line change thanks to the existing `${revision}` + `flatten-maven-plugin` setup — no per-module versions to touch.

## Optional follow-ups

- Document (or wire up) a one-command bump via `versions-maven-plugin` so future bumps are trivial:

  ```bash
  mvn org.codehaus.mojo:versions-maven-plugin:2.18.0:set-property \
      -Dproperty=revision -DnewVersion=2.0.5-SNAPSHOT -DgenerateBackupPoms=false
  ```

- (Further out) Automate this bump in the release workflow, so `main` never lags behind a published release.

## Impact

- Purely development-version housekeeping; no published artifacts are affected.
- Anyone building against `2.0.3-SNAPSHOT` should switch to the released `2.0.3`.

amyaxy:
## Summary

Bump `<revision>` from `2.0.3-SNAPSHOT` to `2.0.4-SNAPSHOT`, so that `main` immediately represents "the next release under development" now that `2.0.3` has shipped to Maven Central.

Fixes #3193 

## Changes

- `pom.xml`: `<revision>2.0.3-SNAPSHOT</revision>` → `<revision>2.0.4-SNAPSHOT</revision>`
- `agentscope-dependencies-bom/pom.xml`: same bump
- `agentscope-distribution/agentscope-bom/pom.xml`: same bump (the released `io.agentscope:agentscope-bom`)

Three files, three lines — thanks to the existing `${revision}` + `flatten-maven-plugin` setup, no per-module versions need to change.

## Motivation

1. **Version collision.** `2.0.3` (release) and `2.0.3-SNAPSHOT` (dev) coexist but point to different code states; locally-installed SNAPSHOT artifacts are easy to confuse with released ones when debugging.
2. **No development coordinate.** Work towards the next release has no reserved version.
3. **Convention.** Standard Maven practice is to advance `<revision>` right after a release, so `main` never lags behind a published version.

## Impact

- Housekeeping only — no published artifacts are affected.
- Anyone building against `2.0.3-SNAPSHOT` should switch to the released `2.0.3`.

## How to verify

- The diff touches only the three `<revision>` lines.
- Any build passes, and the effective revision resolves to `2.0.4-SNAPSHOT`:


pom.xml
```diff
@@ -27,7 +27,7 @@
     <name>AgentScope Java - Parent</name>
 
     <properties>
-        <revision>2.0.3-SNAPSHOT</revision>
+        <revision>2.0.4-SNAPSHOT</revision>
```
oss-maintainer, line 30: **[Critical] `agentscope-distribution/agentscope-bom/pom.xml` is left on `2.0.3-SNAPSHOT`.**

Three poms declare their own `revision`; this commit updates two of them:

| file | `revision` at this head |
|---|---|
| `pom.xml` | `2.0.4-SNAPSHOT` |
| `agentscope-dependencies-bom/pom.xml` | `2.0.4-SNAPSHOT` |
| `agentscope-distribution/agentscope-bom/pom.xml:62` | **`2.0.3-SNAPSHOT`** |

`agentscope-bom` is parentless (`<version>${revision}</version>` at `agentscope-bom/pom.xml:25`) and every managed entry in it is `${project.version}`, so as it stands the BOM keeps publishing `io.agentscope:agentscope-bom:2.0.3-SNAPSHOT` pinning `agentscope-core`/`agentscope-harness`/the extensions at `2.0.3-SNAPSHOT` while this bump stops producing those artifacts. A consumer that follows the documented BOM import gets the previous snapshot line (or a resolution failure once it is dropped), and `Check Module Sync` will not catch it — it validates module wiring, not the three independent `revision` properties.

The previous iteration bump behaves the way this one should: `e0674d40` ("prepare for next development iteration.") touches `pom.xml`, `agentscope-dependencies-bom/pom.xml` **and** `agentscope-distribution/agentscope-bom/pom.xml` in one commit.

To keep the next bump from regressing the same way, this is a one-line CI guard:

```bash
test "$(grep -rho '<revision>[^<]*</revision>' --include=pom.xml . | sort -u | wc -l)" -eq 1
```

*(Unrelated to the diff, so nothing to do here: `agentscope-core/pom.xml` and `agentscope-core/src/test/java/io/agentscope/core/VersionTest.java` also carry literal `2.0.3-SNAPSHOT` strings — the pom comment and the `Version.resolveVersionFrom*` unit fixtures. They are self-contained assertions, so they do not break the build; they just age on every bump.)*

---

oss-maintainer (commented):
### Summary
Bumps `revision` from `2.0.3-SNAPSHOT` to `2.0.4-SNAPSHOT` in the root pom and the dependencies BOM. The change is right in shape and the version resolution path is guarded by the existing `VersionTest` filtering checks, but it covers two of the three poms that carry the property: `agentscope-distribution/agentscope-bom/pom.xml:62` still says `2.0.3-SNAPSHOT`, so the published BOM would keep pinning a snapshot line this bump retires. One line to add, then this is good to go.

### Findings
- **[Critical]** `pom.xml:30` — the bump is incomplete; `agentscope-bom` still declares `2.0.3-SNAPSHOT`, and the last iteration bump (`e0674d40`) changed all three files together. Details plus a one-line CI guard in the inline comment.

### CI / CLA
- `license/cla`: signed. `Check License` and `Check Module Sync` pass; `build (ubuntu-latest)` / `build (windows-latest)` were still queued when this review was posted — the incomplete bump would not have failed them either way, which is why the guard above is worth adding.
- No build or test was run locally for this review; all version values come from the head blobs.

---

pom.xml
amyaxy: 已经解决

oss-maintainer, line 30: **[Warning] `agentscope-distribution/agentscope-bom/pom.xml` is left behind by the bump.**

Three poms carry their own `revision`; this commit updates two:

| file | `revision` at `551489eb` |
|---|---|
| `pom.xml:30` | `2.0.4-SNAPSHOT` |
| `agentscope-dependencies-bom/pom.xml:63` | `2.0.4-SNAPSHOT` |
| `agentscope-distribution/agentscope-bom/pom.xml:62` | **`2.0.3-SNAPSHOT`** |

`agentscope-bom` is parentless and takes its own version from that property (`agentscope-bom/pom.xml:25`), and every managed entry inside it is `${project.version}` — so it would keep publishing `agentscope-bom:2.0.3-SNAPSHOT` pinning `io.agentscope:agentscope-core:2.0.3-SNAPSHOT`, `agentscope-harness:2.0.3-SNAPSHOT` and the rest of the 2.0.3 line, which is the set of artifacts this bump stops producing. A consumer that follows the documented BOM import silently resolves the previous snapshot line (or fails once it is gone), which is exactly the failure mode a BOM is there to prevent.

The previous development-iteration bump is the precedent: `e0674d40` ("prepare for next development iteration.") changed `pom.xml`, `agentscope-dependencies-bom/pom.xml` **and** `agentscope-distribution/agentscope-bom/pom.xml` in one commit. Please add the third file.

`Check Module Sync` passes either way, because it validates module wiring rather than the three independent `revision` properties. If you want the invariant enforced instead of remembered, this is a one-liner in the same job:

```bash
test "$(grep -rho '<revision>[^<]*</revision>' --include=pom.xml . | sort -u | wc -l)" -eq 1
```

agentscope-dependencies-bom/pom.xml
```diff
@@ -60,7 +60,7 @@
     </developers>
 
     <properties>
-        <revision>2.0.3-SNAPSHOT</revision>
+        <revision>2.0.4-SNAPSHOT</revision>
```
oss-maintainer, line 63: **[Info] Nothing to change here — recording what I checked so the next bump does not have to re-derive it.**

I grepped every `pom.xml` in the tree for a literal `revision` property and for the old `2.0.3-SNAPSHOT` string. Besides the `agentscope-bom` miss in the other comment, the only remaining references are non-load-bearing: a `agentscope-core/pom.xml` comment describing what the unfiltered `version.properties` would look like (line ~271) and the self-contained fixtures in `VersionTest` (`testResolveVersionFrom*` feed their own strings to the resolver). Both are stale-looking rather than wrong, and neither is affected by this bump. `Version.VERSION` itself is resolved from the Maven-filtered `META-INF/agentscope/version.properties`, so it follows `${revision}` automatically and `VersionTest` keeps asserting the two agree.

oss-maintainer (commented):
### Summary
Routine next-development-iteration bump, `2.0.3-SNAPSHOT` → `2.0.4-SNAPSHOT`. The mechanism is right — a single `revision` property per aggregator rather than per-module versions, with `${project.version}`/`${revision}` everywhere else, which is why this is only two lines. One file is missing, and it is the one users resolve against.

### Findings
- **[Warning]** `pom.xml:30` — `agentscope-distribution/agentscope-bom/pom.xml:62` still declares `2.0.3-SNAPSHOT`, so the published BOM keeps pinning the artifact line this bump stops producing. The previous bump (`e0674d40`) changed all three poms together.
- **[Info]** `agentscope-dependencies-bom/pom.xml:63` — remaining `2.0.3-SNAPSHOT` occurrences in the tree are a stale comment in `agentscope-core/pom.xml` and self-contained `VersionTest` fixtures; neither needs action, listed so the next bump does not re-audit it.

### Checks
`license/cla` is signed and `Check License` / `Check Module Sync` pass; `build (ubuntu-latest)` and `build (windows-latest)` were still queued when this review was posted, so this is not a merge-readiness verdict — and note that neither green check covers the missed pom, which is the point of the suggestion above. Adding the third line and re-running is enough for me to call this clean.

---

oss-maintainer:
**Correction — the finding above is resolved on the current head.**

The PR was amended while this review was being written (`551489eb` → `1226ef68`) and the new head bumps all three `revision` declarations, including the one the review said was missing:

| file | `revision` at `1226ef68` |
|---|---|
| `pom.xml` | `2.0.4-SNAPSHOT` |
| `agentscope-dependencies-bom/pom.xml` | `2.0.4-SNAPSHOT` |
| `agentscope-distribution/agentscope-bom/pom.xml` | `2.0.4-SNAPSHOT` |

`grep -rho '<revision>[^<]*</revision>' --include=pom.xml . | sort -u` now returns exactly one value, so the bump matches the shape of the previous development-iteration commit (`e0674d40`) and the BOM pins the line this release actually produces. Nothing further from me on the code.

Two notes for the record, both mine rather than yours:
- Two duplicate COMMENT reviews were posted against this PR in the same sweep (`…5244130635` and `…5244260879`), both written against `551489eb`. This comment supersedes both.
- The one-line CI guard suggested above (`sort -u | wc -l` must be 1 over the `revision` properties) is still unimplemented. It is not needed for this PR anymore, but it is the thing that would have caught the missing file before it became an amend — worth a separate small PR if no one objects.

**Verdict: good to go from my side; `license/cla` signed, `Check License` and `Check Module Sync` green on `1226ef68`, the two `build` jobs were still running when this comment was posted. Merging stays with the maintainers.**

---
*Automated follow-up by github-manager-bot*

PR #3200, opened 2026-09-18
fix(formatter): make Base64 tool-result temp paths deterministic

### LINKED ISSUE #3199: [Bug]: Base64 tool-result formatting creates nondeterministic temporary paths
## Describe the bug

Formatting the same unchanged `Msg` history more than once can produce different serialized tool-result content when a `ToolResultBlock` contains Base64-backed media.

The confirmed affected paths are:

- tool-result stringification through `AbstractBaseFormatter`, reproduced through `DashScopeMultiAgentFormatter`;
- `GeminiMessageConverter`.

Before the fix, both paths materialized Base64 media with `Files.createTempFile(...)`. Each formatting pass therefore created a new randomly named temporary file, and the absolute path became part of the serialized tool-result text.

The message history and decoded media bytes were unchanged, but the formatted provider request was different.

## To Reproduce

1. Create a `ToolResultBlock` containing an `ImageBlock` backed by a `Base64Source`.
2. Add it to a tool-role `Msg`.
3. Format the same `Msg` twice through `DashScopeMultiAgentFormatter` or `GeminiMessageConverter`.
4. Compare the serialized tool-result output.

No external model or API credentials are required.

## Expected behavior

Within the same running process, identical decoded media content and the same media type should produce the same materialized path and therefore the same formatted tool-result output.

Different decoded media content should continue to produce different materialized paths.

## Actual behavior

Every formatting pass creates a new randomly named temporary file, even when the decoded media content has not changed.

## Why this affects prompt and token caching

Prompt or prefix caching generally depends on a byte-identical input prefix. Appending a new user message or tool result at the end of an unchanged history can preserve the existing cached prefix.

Before this fix, the historical tool-result text itself contained a newly generated temporary path on every formatting pass. For example:

```text
The returned image can be found at: /tmp/agentscope_123456.png
```

could become:

```text
The returned image can be found at: /tmp/agentscope_987654.png
```

even though the decoded media bytes and the logical message history were unchanged.

This changes the serialized request at a historical position. For providers that use continuous prompt or prefix caching, the prefix before the changed path may remain reusable, but the suffix starting at the changed path can no longer match the previous cached prefix. If the changed tool result appears early in a long history, a larger portion of the request may lose cache reuse.

As a result:

- reported cached-input-token counts may decrease;
- more input tokens may be processed as uncached input;
- request latency and input-token cost may increase, depending on provider caching semantics;
- cache reuse becomes less stable across repeated formatting of the same message history.

The exact cache accounting and pricing behavior is provider-specific. This issue does not claim that every provider invalidates its entire cache or that every provider exposes the same cache-token fields.

## Impact before the fix

- Unchanged message history produces different provider requests.
- A historical prefix changes even when the logical message content is unchanged.
- Provider-side prompt or prefix cache reuse may be reduced from the changed path onward.
- Cached input-token counts may decrease while more input tokens are processed as uncached input.
- Depending on provider behavior, request latency and input-token cost may increase.
- Repeated formatting creates redundant temporary files.
- Replays and debugging are less reproducible because the serialized tool-result text changes between formatting passes.

## Scope

This issue is confirmed for the specific tool-result stringification paths described above. It does not imply that every provider or every multimodal tool-result path is affected.

## Proposed fix and expected impact after the fix

The accompanying fix:

- derives the target identity from SHA-256 of the decoded media bytes;
- keeps materialized files inside a JVM-unique temporary directory;
- derives a sanitized extension from the MIME type;
- publishes complete staging files before exposing the deterministic target;
- validates existing targets and rejects symbolic links, non-regular files, and mismatched content;
- reuses the shared materialization logic from `AbstractBaseFormatter` and `GeminiMessageConverter`;
- preserves existing formatter method signatures and serialized text structure.

After the fix, formatting unchanged Base64-backed tool-result media repeatedly within the same JVM reuses the same materialized path, so the serialized tool-result text remains stable. This preserves the opportunity for providers with prompt or prefix caching to reuse the same historical prefix.

Different decoded content produces a different content-addressed target.

The fix does not guarantee cache hits, because cache policy, cache boundaries, minimum prefix lengths, expiration, and billing semantics remain provider-specific.

The deterministic guarantee is intentionally process-local. The per-process directory is not automatically cleaned up, and invalid existing targets fail closed rather than being overwritten.

## Official references

- Java `Files.createTempFile`: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Files.html#createTempFile(java.lang.String,java.lang.String,java.nio.file.attribute.FileAttribute...)
- Gemini context caching and similar request prefixes: https://ai.google.dev/gemini-api/docs/generate-content/caching?hl=en

## Environment

- AgentScope-Java: `main`
- Project version: `2.0.3-SNAPSHOT`
- Java: 17
- OS: macOS

liugy789:
## AgentScope-Java Version

`2.0.3-SNAPSHOT`, based on `main@594343c6ef29f26eea3cbbb2d6f92b1e5b9696de`

## Description

Fixes #3199

Formatting the same unchanged message history could previously produce different tool-result text when Base64-backed media was converted to a temporary-file reference.

The previous implementations in `AbstractBaseFormatter` and `GeminiMessageConverter` used `Files.createTempFile(...)`, so every formatting pass embedded a new random path into the provider request.

## Impact before this change

- The serialized prefix of an otherwise unchanged request changed between formatting passes.
- For providers using continuous prompt or prefix caching, the changed temporary path could break cache reuse from that point onward in the serialized history.
- Cached input-token counts could decrease while more input tokens were processed as uncached input.
- Depending on provider behavior, request latency and input-token cost could increase.
- Repeated formatting created redundant temporary files.
- Request replay and debugging were less reproducible.

## Changes

This change:

- adds shared content-addressed Base64 materialization to `MediaUtils`;
- derives file identity from SHA-256 of the decoded media bytes;
- stores deterministic targets inside a JVM-unique temporary directory;
- sanitizes MIME-derived file extensions;
- publishes complete staging files before exposing the deterministic target;
- uses atomic publication when supported;
- validates the target after the non-atomic fallback;
- protects publication with bounded striped locks;
- rejects symbolic links, non-regular targets, and mismatched existing content;
- wraps malformed Base64 payloads as checked materialization failures so existing formatter degradation behavior is preserved;
- preserves the original publication error when staging cleanup also fails;
- preserves the existing formatter method signatures and serialized text structure;
- reuses the shared implementation from `AbstractBaseFormatter` and `GeminiMessageConverter`.

## Impact after this change

Within the same JVM, identical decoded media content and the same media type reuse the same materialized path. The serialized tool-result text therefore remains byte-stable when the unchanged message history is formatted repeatedly.

This preserves the opportunity for providers with prompt or prefix caching to reuse the same historical prefix. It avoids actively invalidating cache reuse merely because the formatter created a new random temporary path.

The change does not guarantee cache hits. Actual cache behavior remains provider-specific and depends on cache policy, cache boundaries, minimum prefix lengths, expiration, token accounting, and billing rules.

Different decoded media content produces different content-addressed targets. The formatter remains stateless, and existing formatter method signatures and serialized text structure remain unchanged.

The deterministic guarantee is process-local. Cross-process path determinism, temporary-file TTL, global cleanup scheduling, and broader multimodal protocol changes are intentionally outside this PR.

Invalid existing targets fail closed rather than being overwritten.

## Compatibility

- Existing formatter method signatures are unchanged.
- Existing tool-result serialized text structure is unchanged.
- URL-backed media handling is unchanged.
- Provider-specific multimodal protocols are unchanged.
- No external dependencies were added.
- Identical media may be shared by multiple formatted results within the same JVM; callers must not modify or delete the returned materialized file.

## Verification

- `mvn spotless:check`: passed.
- `mvn -DskipTests ... compile` for the affected modules and reactor dependencies: passed.
- `git show --check`: passed.
- The final code tree was previously verified with 3010 tests, 0 failures, and 0 errors:
  - `MediaUtilsTest`: 41/41;
  - `DashScopeMultiAgentFormatterTest`: 37/37;
  - `GeminiMessageConverterTest`: 34/34.

A subsequent local rerun was blocked by an environment-level Mockito/Byte Buddy inline-agent attachment failure in the test JVM. The affected failure is in the local test runtime's agent-attachment mechanism; CI verification remains required before claiming the checks are green.

The tests cover:

- deterministic reuse for identical decoded content;
- equivalent padded and unpadded Base64 representations;
- different paths for different content;
- MIME-derived suffix sanitization;
- malformed Base64 degradation;
- corrupt existing-target rejection;
- symbolic-link and non-regular-target rejection;
- concurrent publication;
- explicit non-atomic publication behavior;
- JVM-unique temporary-directory placement;
- DashScope and Gemini formatter-level determinism;
- exact materialized file contents;
- preservation of successful publication when staging cleanup fails;
- preservation of the original write error when staging cleanup also fails.

## Non-goals

- temporary-file TTL;
- global cleanup scheduling;
- cross-process path determinism;
- provider-specific prompt-cache control protocols;
- broader multimodal protocol changes;
- unrelated provider formatting paths.

## Checklist

- [x] Code uses the repository's Conventional Commit format.
- [x] Regression tests cover both confirmed provider paths.
- [x] Spotless check passed.
- [x] Affected production modules compile.
- [x] Existing formatter method signatures and serialized text structure are preserved.
- [x] Prompt-cache and prefix-cache impact is documented.
- [ ] Final CI checks are passing.
- [ ] Maintainer review is complete.


agentscope-core/src/main/java/io/agentscope/core/formatter/MediaUtils.java
```diff
@@ -101,6 +116,199 @@ public static String fileToBase64(String path) throws IOException {
...
+        String extension = safeMediaExtension(mediaType);
+        Path target =
+                tempDirectory
+                        .resolve("agentscope_" + digest + "." + extension)
+                        .toAbsolutePath()
+                        .normalize();
+
+        synchronized (materializationLock(target)) {
+            if (Files.exists(target, LinkOption.NOFOLLOW_LINKS)) {
+                validateMaterializedFile(target, decodedData);
```
oss-maintainer, line 172: Non-blocking: on the reuse path the full payload is re-read and compared byte-by-byte on every format call. For large media that is re-formatted repeatedly in an agent loop (image/audio tool results), this turns the "reuse" fast path into a repeated O(size) disk read. Consider caching verified targets in-process (e.g. a small set of already-validated absolute paths, or verifying only when `Files.size(target) != decodedData.length` differs), keeping the full validation for the publication path where it is genuinely needed.

agentscope-core/src/main/java/io/agentscope/core/formatter/MediaUtils.java
```diff
@@ -101,6 +116,199 @@ public static String fileToBase64(String path) throws IOException {
...
+        }
+    }
+
+    private static Path materializedMediaDirectory() throws IOException {
+        Path directory = materializedMediaDirectory;
+        if (directory == null) {
+            synchronized (MediaUtils.class) {
+                directory = materializedMediaDirectory;
+                if (directory == null) {
+                    directory = Files.createTempDirectory("agentscope_media_");
```
oss-maintainer, line 211: Non-blocking: the per-process `agentscope_media_*` directory is never evicted or cleaned (`Runtime.deleteOnExit` is explicitly out of scope here and would not help for long-running servers). This is no worse than the previous per-call `Files.createTempFile` behaviour, but for long-lived harness/distribution processes streaming many distinct tool results it is unbounded growth in the OS temp dir. Worth a follow-up issue for a size/count-bounded eviction policy, or documenting that operators should prune `java.io.tmpdir/agentscope_media_*`.

oss-maintainer (approved):
### Summary
Deterministic, content-addressed Base64 media temp files — the previously random `Files.createTempFile` path made tool-result formatting non-reproducible and spawned a new file per call. The rewrite is correct and carefully hardened: SHA-256 content identity, striped locking, staging file + `ATOMIC_MOVE` publication with fallback, `NOFOLLOW_LINKS` checks that reject symlink/irregular targets, MIME subtype sanitisation, and existing-file validation instead of silent overwrite. Error handling preserves the primary failure and suppresses cleanup errors. Shared by both `AbstractBaseFormatter` and `GeminiMessageConverter`, so the fix lands consistently across model providers. Test coverage is excellent (content identity incl. padding, concurrency ×8, atomic and non-atomic moves, symlink rejection, cleanup-failure and SecurityException paths, plus an end-to-end DashScope determinism check). CLA signed, CI green on all platforms. Two non-blocking performance/operability notes are pinned inline; neither affects correctness. **Approved.**

---

PR #3204, opened 2026-09-18
fix(sandbox): keep Kubernetes wrap newlines around command

### LINKED ISSUE #3198: [Bug]: execute tool fails on any command whose last line is a heredoc terminator
**Describe the bug**

Any command sent through the `execute` tool (`ShellExecuteTool`) fails if its **last line is a heredoc terminator**. The command is valid shell and works verbatim when run by hand. The error that comes back is a shell syntax error that mentions neither the heredoc nor its delimiter, so it reads as "heredocs are unavailable in the sandbox".

Something is appended to the command's final line before it reaches the shell. When that line is the closing delimiter, the delimiter stops matching, the heredoc consumes the rest of the input, and the shell reports an unterminated construct at the last line of the *content*.

This is the same family as the transport/escaping bugs already fixed or reported here — #2925 (Windows docker exec), #3084 (`bash -lc` escaping in `edit()`), #2618 (base64 payload in argv) — but it needs no large payload and no Windows: it reproduces with a three-line heredoc on Linux.

**To Reproduce**

1. Call the `execute` tool with a command whose last line is the terminator:

```
cd /workspace/runs/1/work && python3 - <<'EOF'
import json
d = {'a': 1}
for k in d:
    if k in ('SRC-a','SRC-b','SRC-c'):
        print(k)
print("done")
EOF
```

2. The call fails. The same command with **one trailing newline**, or with any command after the terminator (`wc -c file`), succeeds.

Where the appended text lands can be reproduced exactly outside the framework. Wrapping the command in `( … )` with the closing paren concatenated onto the end of the string reproduces the error string, wording and line number exactly, while two other plausible appendages do not:

```python
import subprocess
lines = ["cd /tmp && python3 - <<'EOF'", "import json",
         "with open('/dev/null') as f:", "    pass", "d = {'a': 1}", "for k in d:",
         "    if k in ('SRC-a','SRC-b','SRC-c'):", "        print(k)",
         "with open('/dev/null') as f:", "    pass", 'print("done")', "EOF"]
cmd = "\n".join(lines)

for label, wrapped in [
    ("no wrapper",                 cmd),
    ("( … ) closing paren on its own line", "( " + cmd + "\n)"),
    ("( … ) closing paren appended to the last line", "( " + cmd + " )"),
    ("`< /dev/null` appended",      cmd + " < /dev/null"),
    ("`; echo EXIT:$?` appended",   cmd + " ; echo EXIT:$?"),
]:
    r = subprocess.run(["/bin/dash", "-c", wrapped], capture_output=True, text=True)
    print(label, "->", r.returncode, (r.stderr or r.stdout).strip().splitlines()[:1])
```

```
no wrapper                                      -> 0 ['done']
( … ) closing paren on its own line             -> 0 ['done']
( … ) closing paren appended to the last line   -> 2 ['/bin/dash: 12: Syntax error: end of file unexpected (expecting ")")']
`< /dev/null` appended                          -> 1 ['  File "<stdin>", line 11']
`; echo EXIT:$?` appended                       -> 1 ['  File "<stdin>", line 11']
```

Only the third matches what the tool returns, so the appended text is on the command's last line rather than on a new one.

**Expected behavior**

`execute` advertises "Execute a shell command". A command ending in a heredoc terminator is a shell command and should run. Either the command should not be modified, or whatever is appended should go on its own line.

**Error messages**

What the tool returned for the command above:

```
Exit code: 2

/bin/sh: 12: Syntax error: end of file unexpected (expecting ")")
```

Line 12 is the `EOF` line. Nothing in the message points at the heredoc, the delimiter, or the appended text.

**Environment**

- AgentScope-Java Version: 2.0.2 (`agentscope-harness`, `agentscope-core`)
- Java Version: 17
- OS: Linux (sandbox container); `/bin/sh` is dash, judging by the `sh: <line>: Syntax error` format
- Sandbox: Kubernetes (`agentscope-extensions-sandbox-kubernetes`, agent-sandbox)

**Additional context**

Where the modification happens, as far as I could trace it: `ShellExecuteTool.execute` passes `command` through unchanged unless `working_directory` is set; `SandboxBackedFilesystem.execute` forwards it to `Sandbox.exec`; `CommandExecutor` POSTs `{"command": …}` to the sandbox's `/execute` endpoint. I found no `( … )` wrapping in `agentscope-harness` or in the Kubernetes client, so it looks server-side — but the tool is the surface where it is observable, and a one-line defence here would cover every sandbox backend:

```java
// ShellExecuteTool.execute, before handing the command over
String effectiveCommand = command.endsWith("\n") ? command : command + "\n";
```

That is the same shape of defence as #3084 (base64 + stdin rather than relying on the shell's quoting). A server-side fix (append on a new line, or do not wrap) would be the real one; the trailing newline works for the behaviour observed above and would stop being needed once that lands.

**Why this is worth more than the syntax error suggests.** The failure teaches the wrong lesson. The error names nothing related to heredocs and the identical command works by hand, so an agent that hits it concludes heredocs are unavailable and switches to `write_file` a script, then `execute` the script — two tool calls for every one, for the rest of the run. In one 25-minute run of ours the agent did that three times after a single failure of this kind. It also removes `cat >> file <<'EOF'` as a way to build a file larger than one tool call can carry, which pushes the model toward writing many small files plus a merge script instead.

tengjiaozhai:
## AgentScope-Java Version

2.0.3-SNAPSHOT (based on upstream/main at 6c8ccc5c)

## Description

`KubernetesSandbox.doExec` wrapped commands as `cd <root> && (command)` with the closing `)` concatenated onto the last line of the command. When that last line was a heredoc terminator (e.g. `EOF`), the shell saw `EOF)` instead of `EOF`, so the delimiter never matched and dash reported `Syntax error: end of file unexpected (expecting ")")`.

This change puts the subshell delimiters on their own lines:

```text
cd <root> && (
<command>
)
```

Only the Kubernetes backend wraps this way; Docker / local / AgentRun paths are unchanged.

Closes #3198

## Verification

- `mvn spotless:apply` / `mvn spotless:check` for the changed sandbox-kubernetes module: passed
- `mvn -pl agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes -am test -Dtest=KubernetesSandboxTest`: 11 tests, 0 failures
- Updated the existing wrap-shape assertion and added `doExecKeepsClosingParenOffHeredocTerminatorLine`

## Checklist

- [x] Code has been formatted with mvn spotless:apply for the changed module.
- [x] Targeted tests are passing (KubernetesSandboxTest).
- [x] Related documentation has been updated (not applicable; no public API change).


agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/main/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandbox.java
```diff
@@ -88,7 +88,8 @@ public void shutdown() throws Exception {
     @Override
     protected ExecResult doExec(RuntimeContext runtimeContext, String command, int timeoutSeconds)
             throws Exception {
-        String wrapped = "cd " + shellQuote(k8sState.getWorkspaceRoot()) + " && (" + command + ")";
+        String wrapped =
+                "cd " + shellQuote(k8sState.getWorkspaceRoot()) + " && (\n" + command + "\n)";
```
oss-maintainer, line 96: **[Warning]** Non-blocking, but worth deciding on while this line is open. The new shape still produces a shell syntax error for a null/blank `command` — `cd '/workspace' && (\n\n)` fails with `syntax error near unexpected token ')'` in both dash and bash. That is unchanged from the old wrap, so it is not a regression; it just means an empty command surfaces as an opaque `sh` parse error instead of a clear failure.

Since this PR is the one that owns the wrap shape, a cheap guard would make the contract explicit:

```java
if (command == null || command.isBlank()) {
    throw new SandboxException.ExecException(2, "", "empty command rejected before exec");
}
```

(Or simply skip the subshell for blank input.) Happy to see this handled in a follow-up — it does not affect the fix here.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/test/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxTest.java
```diff
@@ -75,7 +75,25 @@ void doExecWrapsCommandWithWorkspaceCd() throws Exception {
...
+        ExecResult result = sandbox.doExec(null, command, 30);
+
+        assertEquals(0, result.exitCode());
+        ArgumentCaptor<String> cmd = ArgumentCaptor.forClass(String.class);
+        verify(commands).run(cmd.capture(), eq(Duration.ofSeconds(30)));
+        String wrapped = cmd.getValue();
+        assertEquals("cd '/workspace' && (\n" + command + "\n)", wrapped);
+        // Closing ')' must not glue onto the heredoc terminator.
+        assertTrue(wrapped.endsWith("\nEOF\n)"));
+        assertTrue(!wrapped.contains("EOF)"));
```
oss-maintainer: **[Info]** Two small test nits:

1. `assertTrue(!wrapped.contains("EOF)"))` reads better as `assertFalse(...)` — the failure message then states the positive condition that broke.
2. The newline wrap actually repairs more than the heredoc case: `echo hi # note` (the old form let the `#` comment swallow the closing `)`) and a command ending in a trailing `\` continuation both failed before and pass now. Those two shapes are the reason I would not switch to a `;`-appending variant instead. Consider folding them into this test (a `@ParameterizedTest` over last-line payloads would keep it compact) so the broader guarantee is pinned and a later refactor of the wrap cannot silently re-break them.

oss-maintainer (approved):
### Summary

Minimal, correct fix for the heredoc/`#`-comment breakage in the Kubernetes sandbox exec path: the closing `)` is moved onto its own line, so a command whose last line is a heredoc terminator (or ends in a trailing comment / line continuation) is no longer glued to `)`. Verdict: **approve** — the two items below are non-blocking.

I replayed both wrap shapes against `dash` and `bash` with the same grammar the runtime uses (`sh -c '<cd> && (<command>)'`). The new shape fixes exactly the class reported in #3198 and also repairs two shapes the old wrap broke — a command ending in `# comment` and one ending in a trailing `\` — with no behaviour change for single-line commands, trailing newlines, multi-line `if/fi` blocks, background `&`, or commands containing quoted parens. So this is a strict improvement, not a swap of one broken case for another.

Scope is right too: the wrap only exists in `KubernetesSandbox.doExec` — Docker passes `sh -c <command>` with `-w <workspaceRoot>`, AgentRun hands the command to `mcp.exec`, Daytona and E2b take a cwd — so the "only the Kubernetes backend changes" claim in the description holds. Repo-wide grep finds no remaining consumer of the old single-line shape outside the two places this PR updates. CI is green (build ubuntu/windows, Check Module Sync, Check License, codecov patch) and `license/cla` is signed.

### Findings
- **[Warning]** `KubernetesSandbox.java:92` — a null/blank `command` still reaches the shell and fails as an opaque syntax error under both the old and the new shape; not a regression, just an opportunity to make the contract explicit while this line is open.
- **[Info]** `KubernetesSandboxTest.java:96` — `assertTrue(!…)` → `assertFalse(…)`, and consider pinning the two additional shapes the newline wrap repairs (trailing `#` comment, trailing `\`) so a future refactor of the wrap cannot silently re-break them.

### Suggestions

If you want the blank-command guard, the narrow version keeps the exception type the harness already maps:

```java
String wrapped =
        "cd " + shellQuote(k8sState.getWorkspaceRoot()) + " && (\n" + command + "\n)";
```
→
```java
if (command == null || command.isBlank()) {
    throw new SandboxException.ExecException(2, "", "empty command rejected by KubernetesSandbox");
}
String wrapped =
        "cd " + shellQuote(k8sState.getWorkspaceRoot()) + " && (\n" + command + "\n)";
```

A `@ParameterizedTest` over the last line of the payload (heredoc terminator, `# comment`, trailing `\`, plain command) would cover the whole class in one place.

---

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/main/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandbox.java
tengjiaozhai: Applied: null/blank `command` now throws `SandboxException.ExecException(2, "", "empty command rejected by KubernetesSandbox")` before the wrap.

agentscope-extensions/agentscope-extensions-sandbox/agentscope-extensions-sandbox-kubernetes/src/test/java/io/agentscope/extensions/sandbox/kubernetes/KubernetesSandboxTest.java
tengjiaozhai: Applied: switched to `assertFalse`, and expanded into a `@ParameterizedTest` over last-line payloads (heredoc terminator / `# comment` / trailing `\` / plain) asserting the closing `)` stays on its own line. Also added `doExecRejectsNullOrBlankCommand`.

tengjiaozhai:
Applied review feedback from @oss-maintainer on `fee113e`:

1. **Blank-command guard** in `KubernetesSandbox.doExec` — null/blank `command` throws `SandboxException.ExecException` with `"empty command rejected by KubernetesSandbox"` before wrapping.
2. **Tests** — `assertFalse(wrapped.contains(...))`; `@ParameterizedTest` over last-line shapes (heredoc terminator / `# comment` / trailing `\` / plain); plus `doExecRejectsNullOrBlankCommand`.

`KubernetesSandboxTest`: 15 tests, 0 failures. `spotless:check` clean.

oss-maintainer (approved):
### Summary
Re-reviewed the new commit since my approval (0d96e960 → fee113e8): the incremental change rejects null/blank commands in `KubernetesSandbox.doExec` with a clear `ExecException` (stderr message) and converts the newline-wrapping assertions into a parameterized test covering heredoc, trailing-comment, trailing-backslash and plain commands. The wrap contract (`closing paren on its own line`) is now asserted generically instead of the heredoc-only case. All 6 checks pass on this head. LGTM — the added guard makes failed-exec surfacing unambiguous and the test matrix covers the previously fragile edge cases.

---

Buktal (approved):
LGTM

tengjiaozhai:
Thanks for the review! I didn’t realize you had merge access 😄

PR #3212, opened 2026-09-20
fix(harness): allow configurable force-sync timeout ceiling

### LINKED ISSUE #3166: [Feature]: Allow application-controlled force-sync subagent timeout beyond 600 seconds
In AgentScope Java 2.0.2, `AgentSpawnTool` uses a hard-coded `MAX_TIMEOUT_SECONDS = 600`. Both the model-provided `timeout_seconds` and the application-provided `CTX_FORCE_SYNC_TIMEOUT_SECONDS` are clamped to that value in `resolveEffectiveTimeoutMs()`. For a long-running remote subagent, the parent agent sometimes needs to wait for the final result in the *same run*. With `CTX_FORCE_SYNC=true`, reaching 600 seconds cancels the remote task instead of promoting it to the background. Setting the application override to 1800 still yields a 600-second deadline.

Minimal example:
```java
    .sessionId("s-1")
    .put(AgentSpawnTool.CTX_FORCE_SYNC, true)
    .put(AgentSpawnTool.CTX_FORCE_SYNC_TIMEOUT_SECONDS, 1800)
    .build();
```
Expected: the trusted application override gives the subagent up to 1800 seconds to finish. Actual: it is silently clamped to 600 seconds. This is the parent-to-subagent wait deadline, not an individual model-request timeout.

Please provide a supported way to configure a higher maximum for the application-controlled force-sync timeout, separately from the model-controlled `timeout_seconds` cap. Keeping the latter at 600 seconds would prevent the model from requesting arbitrarily long waits, while allowing an embedding application to set an explicit bounded policy (e.g. 1800 or 3600 seconds). Ideally the effective deadline should be observable, and the local/remote spawn and send paths should have tests for the configured cap.

- `timeout_seconds=0` / timeout promotion: useful for background work, but the parent run may finish before the result is available. It does not fit applications that must return the report before that run finishes.
- Patching and publishing a private `agentscope-harness` build: possible, but creates maintenance and upgrade overhead.

**Additional context**
The relevant logic is in `AgentSpawnTool.resolveTimeoutMs()` and `resolveEffectiveTimeoutMs()`; both use `Math.min(..., MAX_TIMEOUT_SECONDS)`. The force-sync setting is an application-provided `RuntimeContext` value rather than an LLM tool argument.

philo-x:
## AgentScope-Java Version

2.0.4-SNAPSHOT

## Description

Closes #3166.

Application-provided force-sync timeouts were always clamped to 600 seconds. Add `AgentSpawnTool.CTX_FORCE_SYNC_MAX_TIMEOUT_SECONDS` so a trusted application can explicitly raise or tighten its wait ceiling. An override of 1800 seconds with a ceiling of 3600 seconds now waits up to 1800 seconds.

The default application ceiling remains 600 seconds. Model-provided `timeout_seconds` remains capped at 600 even with a higher application ceiling. The application owns this ceiling policy; no additional framework duration cap is imposed beyond the Integer.MAX_VALUE representation limit. Both settings use seconds, not milliseconds. Both application settings are ignored when force sync is disabled. Invalid/non-positive ceilings fall back to 600 seconds.

Normalize numeric context values without integer overflow, including large `BigInteger`, `BigDecimal`, and integer strings. Log the effective force-sync budget and ceiling at DEBUG and clamping at INFO. Update English and Chinese documentation with examples.

### Validation

- 54 focused timeout-policy/force-sync test cases passed.
- Harness tests: 1093 discovered, 1079 passed, 14 skipped. The initial sandboxed run had 2 WebTools socket-permission errors; all 6 WebTools tests passed on a targeted rerun with localhost binding permitted. Combined final results have 0 failures/errors.
- Core SubAgentTool cancellation and timeout/retry regression tests: 2 passed.
- Spotless checks and `git diff --check` passed.
- Existing skips: 9 opt-in Docker integration tests and 5 tests disabled on macOS. The full repository-wide CI matrix was not run locally.

New path tests cover local and remote fresh spawn, persistent spawn reuse, and send. They verify application overrides outlast the model deadline, a configured lower ceiling cancels execution, and force-sync timeouts never promote to background work. Remote dependencies are mocked; short real deadlines exercise execution, while numeric tests verify 1800/3600-second and overflow boundaries without waiting half an hour.

Regression command (Java 21 on macOS; explicitly loading Mockito avoids JVM self-attachment failures):

```sh
mvn -o -pl agentscope-harness -am '-Dtest=io.agentscope.harness.**.*Test,SubAgentTool*Test' -Dsurefire.failIfNoSpecifiedTests=false -DargLine=-javaagent:$HOME/.m2/repository/org/mockito/mockito-core/5.23.0/mockito-core-5.23.0.jar test
```

## Checklist

- [x] Affected-module and related regression tests pass (scope and skips detailed above)
- [x] Javadoc comments describe the new configuration and defaults


agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java
```diff
@@ -1552,19 +1625,32 @@ static Integer forceSyncTimeoutSeconds(RuntimeContext ctx) {
...
+            } else if (value instanceof String text && !text.isBlank()) {
+                normalized = new BigDecimal(new BigInteger(text.trim()));
+            } else {
                 return null;
             }
+            if (normalized.signum() <= 0) {
+                return 0;
+            }
+            if (normalized.compareTo(BigDecimal.valueOf(Integer.MAX_VALUE)) > 0) {
+                return Integer.MAX_VALUE;
```
oss-maintainer, line 1660: **[Warning]** Saturating to `Integer.MAX_VALUE` removes the guard rail from the thing that is supposed to *be* the guard rail. Before this change the app override could never exceed `MAX_TIMEOUT_SECONDS` (600s); now the only bound on `min(override, ceiling)` is whatever the two context values say, and both accept up to 2147483647 seconds (~68 years). The realistic failure mode is not a 10-digit typo but a unit mix-up, which does not even saturate: `CTX_FORCE_SYNC_MAX_TIMEOUT_SECONDS=86400000` + `CTX_FORCE_SYNC_TIMEOUT_SECONDS=86400000` (someone writing ms into a seconds key) is a valid 2.7-year synchronous wait, and `race.orTimeout(timeoutMs, MILLISECONDS)` at `:970` will take it literally.

Could we keep an absolute bound independent of the two settings, e.g.

```java
private static final int HARD_MAX_FORCE_SYNC_SECONDS = 86_400; // 24h, documented in the constant javadoc
...
if (normalized.compareTo(BigDecimal.valueOf(HARD_MAX_FORCE_SYNC_SECONDS)) > 0) {
    log.warn("agent_spawn: force-sync value {} out of range, using {}s", value, DEFAULT_FORCE_SYNC_MAX_TIMEOUT_SECONDS);
    return DEFAULT_FORCE_SYNC_MAX_TIMEOUT_SECONDS;   // or null -> caller applies the default
}
```

Falling back to 600s on an out-of-range ceiling is also more consistent with the rest of `forceSyncMaxTimeoutSeconds`, where unset/invalid/non-positive all land on the default.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java
```diff
@@ -1521,27 +1546,75 @@ static boolean isForceSync(RuntimeContext ctx) {
...
+                        requestedSeconds,
+                        effectiveSeconds,
+                        ctx != null ? ctx.getSessionId() : null);
             }
+            log.debug(
+                    "Subagent force-sync timeout: effective={}s, ceiling={}s, source=application,"
+                            + " session={}",
+                    effectiveSeconds,
+                    appMaxSeconds,
+                    ctx.getSessionId());
```
oss-maintainer, line 1594: **[Info]** The two log statements in the same block disagree about whether `ctx` can be null: the clamp log guards it (`ctx != null ? ctx.getSessionId() : null`, `:1575`, and again at `:1596`) while this DEBUG line dereferences it directly. Nothing can reach this branch with a null `ctx` — `isForceSync(null)` is `false` and we returned at `:1562` — so the unguarded form is safe and the ternary is dead code. Worth picking one style; the mixed signal is the kind of thing that makes a later reader add a null check of their own.

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java
```diff
@@ -1552,19 +1625,32 @@ static Integer forceSyncTimeoutSeconds(RuntimeContext ctx) {
...
+     * saturate to Integer.MAX_VALUE; fractional Numbers truncate toward zero; blank/invalid or
+     * non-finite values → null.
+     */
+    private static Integer asPositiveOrZeroInt(Object value) {
+        try {
+            BigDecimal normalized;
+            if (value instanceof Number number) {
+                normalized = new BigDecimal(number.toString());
+            } else if (value instanceof String text && !text.isBlank()) {
+                normalized = new BigDecimal(new BigInteger(text.trim()));
```
oss-maintainer, line 1652: **[Info]** Asymmetric grammar between the two accepted forms: a `Number` of `1800.0` truncates and works, but the *string* `"1800.0"` fails in `new BigInteger(...)` and becomes `null`, i.e. "not configured" — so a value read from a properties file / JSON config as a decimal string silently drops back to 600s instead of 1800s. `" 1800 "` is already handled, so widening the string path to `new BigDecimal(text.trim())` looks like a one-word change that keeps the truncation and saturation logic above it intact. If integer-only strings are intentional, the javadoc at `:212` ("Accepts an `Integer`/`Number` or its string form") should say so explicitly, since the docs page already phrases it as "an integer string".

agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/AgentSpawnToolTimeoutPolicyTest.java
```diff
@@ -0,0 +1,302 @@
...
+
+    @ParameterizedTest
+    @CsvSource({"0,1", "-1,1", ",1", "600,1", "1800,600"})
+    void ceilingAlsoBoundsModelAndDefaultTimeouts(Integer modelTimeout, long expectedSeconds) {
+        RuntimeContext ctx =
+                RuntimeContext.builder()
+                        .put(AgentSpawnTool.CTX_FORCE_SYNC, true)
+                        .put(
+                                AgentSpawnTool.CTX_FORCE_SYNC_MAX_TIMEOUT_SECONDS,
+                                modelTimeout != null && modelTimeout == 1800 ? 3600 : 1)
```
oss-maintainer: **[Info]** The ceiling is derived from the *model* timeout inside the body, so the `@CsvSource` rows no longer describe the case being run: `{"600,1", "1800,600"}` actually means "ceiling 1s, model 600s -> 1s" and "ceiling 3600s, model 1800s -> 600s". Making the ceiling a real column keeps the table readable when a failure prints only the arguments:

```java
@CsvSource({"0,1,1", "-1,1,1", ",1,1", "600,1,1", "1800,3600,600"})
void ceilingAlsoBoundsModelAndDefaultTimeouts(Integer modelTimeout, Integer ceiling, long expectedSeconds) {
            .put(AgentSpawnTool.CTX_FORCE_SYNC, true)
            .put(AgentSpawnTool.CTX_FORCE_SYNC_MAX_TIMEOUT_SECONDS, ceiling)
            .build();
    assertEquals(expectedSeconds * 1000, AgentSpawnTool.resolveEffectiveTimeoutMs(modelTimeout, ctx));
}
```

agentscope-harness/src/test/java/io/agentscope/harness/agent/tool/AgentSpawnToolForceSyncTest.java
```diff
@@ -126,6 +127,132 @@ void helpers() {
         assertTrue(AgentSpawnTool.resolveEffectiveTimeoutMs(5, forceHuge) == 600_000L);
     }
 
+    @Test
+    @DisplayName("force_sync configurable ceiling and overflow hardening")
+    void configurableCeilingAndOverflow() {
```
oss-maintainer: **[Info]** Nice hardening coverage, but it now lives in the file the new dedicated suite was meant to replace: `configurableCeilingAndOverflow` re-cases ceiling-default, ceiling-tighten, non-positive override and overflow saturation, all of which `AgentSpawnToolTimeoutPolicyTest` asserts (some of them parameterized). Keeping the numeric matrix in the policy suite and leaving only the end-to-end force-sync behaviour here would give one place to edit when the policy changes. Non-blocking.

oss-maintainer (commented):
### Summary
Adds `AgentSpawnTool.CTX_FORCE_SYNC_MAX_TIMEOUT_SECONDS` so an application can raise (or tighten) the ceiling that bounds its own `CTX_FORCE_SYNC_TIMEOUT_SECONDS` override — closing #3166 — and rewrites the context-value coercion so oversized numbers saturate instead of wrapping through `Number.intValue()`. The shape of the change is right: the model-facing 600s cap stays untouched, both settings are inert while force-sync is off, and the execution-path tests cover remote/local x spawn/reuse/send rather than only the resolver. My findings are non-blocking; the one I would like a decision on is that the new ceiling has no bound of its own.

### Findings
- **[Warning]** `AgentSpawnTool.java:1644` — saturating the ceiling to `Integer.MAX_VALUE` leaves the guard rail unbounded; a seconds/milliseconds mix-up such as `86400000` is a *valid* in-range value and yields a ~2.7-year `orTimeout` instead of the pre-change 600s clamp.
- **[Info]** `AgentSpawnTool.java:1582` — the DEBUG line dereferences `ctx` while the INFO lines guard it with `ctx != null ? ... : null`; one of the two is dead code.
- **[Info]** `AgentSpawnTool.java:1640` — `Number` `1800.0` truncates and works, the string `"1800.0"` parses to null and silently means "not configured".
- **[Info]** `AgentSpawnToolTimeoutPolicyTest.java:152` — the ceiling is derived from the model timeout inside the test body, so the `@CsvSource` rows no longer describe the case being run.
- **[Info]** `AgentSpawnToolForceSyncTest.java:132` — the numeric matrix duplicates the new dedicated policy suite.

### CI / CLA
`validate`, `Check License`, `Check Module Sync` and `license/cla` are green on `afd2ba6c`; the ubuntu and windows `build` jobs were still queued when this review was written, so nothing here is verified beyond what those checks cover.

### Verdict
COMMENT, not approve — please decide on the unbounded ceiling (or state why it is acceptable) and let the build matrix finish. As a first contribution this is well above average: scoped change, tests on both the numeric policy and the real cancellation paths, and EN/ZH docs updated in step.

---

agentscope-harness/src/main/java/io/agentscope/harness/agent/tool/AgentSpawnTool.java
philo-x: Thanks for calling out the seconds/milliseconds mix-up. I am keeping the ceiling application-owned, as requested in #3166: model-provided timeout_seconds remains hard-capped at 600s, while both RuntimeContext settings are supplied by the trusted embedding application. The ceiling defaults to 600s unless explicitly configured; Integer.MAX_VALUE is a representation limit, not a recommended operational budget.

A second framework-owned duration cap (such as 24h) would introduce a new policy boundary that #3166 does not specify. I have made this ownership explicit in both public Javadocs and the EN/ZH docs, including the concrete example that 24 hours is 86400 seconds, not 86400000. Applications must validate units and select a ceiling appropriate to their workload. The existing effective-budget/clamping logs remain available.

This intentionally preserves the current application policy, including its configuration-risk tradeoff. If maintainers want an additional absolute framework maximum, its value and clamp-versus-fallback behavior should be an explicit API policy decision.

Updated in 2b8595ca.

philo-x:
@oss-maintainer Please re-review commit 2b8595ca.

I retained the application-owned ceiling policy from #3166 and clarified its trust boundary and seconds units in the public Javadocs and EN/ZH docs. I also removed the redundant null guards, made the CSV ceiling explicit, documented/tested integer-only strings (including "1800.0"), and consolidated the duplicate numeric matrix into the dedicated policy suite while preserving the distinct combinations.

Validation: 54 focused timeout/force-sync tests passed; harness combined final results are 1079 passed and 14 existing skips (1093 total); 2 core cancellation/timeout regression tests passed. Two WebTools tests initially hit sandbox socket restrictions; the whole 6-test WebTools suite passed when rerun with localhost binding permitted. Spotless and git diff --check passed. New remote CI is pending.

The detailed response on the ceiling thread explains why I did not introduce an arbitrary 24h framework cap. Please review that explicit policy decision along with the cleanups.

oss-maintainer (approved):
### Summary
Re-review after `2b8595ca` (clarify force-sync timeout policy after review). The previously raised points are all closed: the ceiling/override semantics are now an explicitly documented, application-owned policy — both constant javadocs state the seconds-not-milliseconds contract, and the `86400` vs `86400000` unit trap is spelled out in the English and Chinese doc pages, the two clamp sites now read `ctx.getSessionId()` consistently, the ceiling is a real `@CsvSource` column, and the duplicated `configurableCeilingAndOverflow` block was removed from `AgentSpawnToolForceSyncTest` in favour of the parameterized `applicationOverrideIsBoundedByConfiguredCeiling` matrix in the policy suite. **Approved** — LGTM.

### What I verified on this HEAD
- Coverage was migrated, not dropped: the 9 rows of `applicationOverrideIsBoundedByConfiguredCeiling` reproduce the assertions from the deleted 127-line test — default ceiling (unset / `0` / `-100` → 600s), raised ceiling honoured (1800 with 3600), tightening (1800 with ceiling 300 → 300s), clamp (7200 with ceiling 3600 → 3600s), non-positive override → 30s, and `Long.MAX_VALUE` override saturating above a 3600s ceiling. The ceiling-bounds-the-model-direction is carried by `ceilingAlsoBoundsModelAndDefaultTimeouts` (`600,1,1` and `1800,3600,600`), which also confirms the model path still hits its hard 600s wall regardless of a higher application ceiling.
- The overflow cases now live in the numeric matrix instead: `(double)(1L << 32) + 100`, `BigInteger 2^64+100`, `BigDecimal("18446744073709551716.5")` and `"18446744073709551616"` all resolve to `Integer.MAX_VALUE` (no wraparound), and negative/huge negatives collapse to 0 → the 30s default.
- The `Integer#MAX_VALUE` saturation is now an intentional, documented boundary rather than an implicit one. I still think a maintainer-visible second framework cap (24h) would be cheaper to operate than a prose warning, but as a trusted-application-owned policy with the unit trap documented at both the javadoc and the doc-page level, this is a defensible design call and not a blocker.
- `"1800.0"` moved into `invalidValues()`, so the Number-vs-string asymmetry is now pinned by a test rather than only by the javadoc wording.
- Import hygiene checked: after deleting the duplicated test, `AgentSpawnToolForceSyncTest.java` no longer references `assertEquals`, so removing the static import is safe (no compile break).
- CLA signed (`license/cla=success`), all required checks green on `2b8595ca` (validate, build ubuntu/windows, Check License, Check Module Sync, codecov/patch), `mergeable=MERGEABLE`; `mergeStateStatus=BLOCKED` only because a review was still required.

### Notes for maintainers
- Behaviour change is confined to the force-sync path: model-supplied `timeout_seconds` keeps its hard 600s cap, and both new context keys are ignored when force-sync is off, so default deployments see no change.
- @philo-x thanks for the thorough follow-up and for moving the numeric matrix into the dedicated policy suite — the parameterized ceiling rows are much easier to extend now.

---

Issue #919, opened 2026-03-11
[Bug]: HttpTransport默认是http2


buchixiangcai111:
<u>AgentScope-Java is an open-source project. To involve a broader community, we recommend asking your questions in English.</u>

Describe the bug
Both default implementations in HttpTransport (JdkHttpTransport and OkHttpTransport) default to using HTTP/2. When a locally deployed VLLM service does not support HTTP/2, it returns a 400 Bad Request error with a misleading validation message about a missing request body.

To Reproduce

1. code

 agent =
                    ReActAgent.builder()
                            .name("PlanAgent")
                            .sysPrompt(
                                    "You are a systematic assistant that helps users complete complex"
                                            + " tasks through structured planning.\n")
                            .model(OpenAIChatModel.builder()
                                            .apiKey("sk-vllm")  // VLLM通常接受任意key
                                            .modelName("qwen3-vl-32b-instruct")
                                            .baseUrl("http://xxx:40034/v1").httpTransport(JdkHttpTransport.builder().build())
                                            .build()
                                    )
                            .memory(memory)
                            .toolkit(toolkit)
                            .skillBox(skillBox)
                            .maxIters(50)
                            .hook(planChangeHook)
                            .planNotebook(planNotebook)
                            .build();

2. How to execute
Run the code against a local VLLM deployment that only supports HTTP/1.1.

3. See error
The request fails with a 400 Bad Request error.

Expected behavior
The HTTP transport should either:
1. Automatically fall back to HTTP/1.1 when HTTP/2 is not supported, or
2. Provide configuration options to explicitly specify the HTTP version, or
3. The error message should clearly indicate the HTTP/2 compatibility issue rather than a misleading "Field required" validation error.

Error messages

{
    "error": {
        "message": "1 validation error:\n  {'type': 'missing', 'loc': ('body',), 'msg': 'Field required', 'input': None}\n\n  File \"/usr/local/lib/python3.12/dist-packages/vllm/entrypoints/utils.py\", line 477, in create_chat_completion\n    POST /v1/chat/completions [{'type': 'missing', 'loc': ('body',), 'msg': 'Field required', 'input': None}]",
        "type": "Bad Request",
        "param": null,
        "code": 400
    }
}

Environment (please complete the following information):

• AgentScope-Java Version: [e.g. 1.0.9]

• Java Version: [e.g. 17]

• OS: [e.g. macos, windows]

Additional context
[...]

PR #3214, opened 2026-09-20
fix(transport): 明文请求默认改用 HTTP/1.1 — JDK HTTP_2 的 h2c upgrade 会导致 vLLM/uvicorn 丢弃 POST 请求体

### LINKED ISSUE #628: Failed to call OpenAI API: OpenAI API request failed with status 400

**Describe the bug**
When I try to use agent call user message with version 1.0.7, I get an error. However, it works well with version 1.0.5. And I deploy a VLLM server to provide LLM Api. 
I'm not sure it's a bug. Maybe I run it in a wrong way with the newest version. 

**To Reproduce**

1. Here is my code:
`// 1. 配置 vLLM 的 OpenAI 兼容模型
        String vllmBaseUrl = "http://localhost:8000";
        String modelName = "Qwen3-235B-A22B-Instruct-2507-AWQ"; // 必须与 vLLM 启动时的 --model 一致

        OpenAIChatModel model = OpenAIChatModel.builder()
                .baseUrl(vllmBaseUrl)
                .apiKey("EMPTY") 
                .modelName(modelName)
                .build();

        // 2. 创建 ReActAgent
                .name("VLLM-Agent")
                .sysPrompt("You are a helpful AI assistant powered by vLLM.")
                .model(model)
                .build();

        // 3. 构建用户消息
        Msg userMsg = Msg.builder()
                .role(MsgRole.USER)
                .content(TextBlock.builder().text("Hello").build())
                .build();

        // 4. 调用模型并打印结果
        try {
            Msg response = agent.call(userMsg).block();
            System.out.println("【模型回复】: " + response.getTextContent());
        } catch (Exception e) {
            System.err.println("❌ 调用失败: " + e.getMessage());
            e.printStackTrace();
        }
2. Just run this method.
3. See error
**Expected behavior**
I just want it work well.

**Error messages**
Failed to call OpenAI API: OpenAI API request failed with status 400 | {"error":{"message":"1 validation error:\n  {'type': 'missing', 'loc': ('body',), 'msg': 'Field required', 'input': None}\n\n  File \"/usr/local/lib/python3.12/dist-packages/vllm/entrypoints/utils.py\", line 477, in create_chat_completion\n    POST /v1/chat/completions [{'type': 'missing', 'loc': ('body',), 'msg': 'Field required', 'input': None}]","type":"Bad Request","param":null,"code":400}}
io.agentscope.core.model.ModelException: Failed to call OpenAI API: OpenAI API request failed with status 400 | {"error":{"message":"1 validation error:\n  {'type': 'missing', 'loc': ('body',), 'msg': 'Field required', 'input': None}\n\n  File \"/usr/local/lib/python3.12/dist-packages/vllm/entrypoints/utils.py\", line 477, in create_chat_completion\n    POST /v1/chat/completions [{'type': 'missing', 'loc': ('body',), 'msg': 'Field required', 'input': None}]","type":"Bad Request","param":null,"code":400}} [model=Qwen3-235B-A22B-Instruct-2507-AWQ, provider=openai]

- AgentScope-Java Version: 1.0.7
- Java Version: 17.0.12
- OS: windows 10
- VLLM: v0.13.0

### LINKED ISSUE #1121: [Bug]:使用openai的model调用，提示报错，用dash正常
Status: 400 | Error: {"object":"error","message":"[{'type': 'missing', 'loc': ('body',), 'msg': 'Field required', 'input': None}]","type":"Bad Request","param":null,"code":400}

xzxiaoshan:
修复 #1121  问题

---

Closes #1121
Closes #628
Refs #919（同一根因；#972 仅新增了配置项，默认路径的问题依旧未修）

**问题**：通过明文 `http://` 调用 OpenAI 兼容端点（vLLM、Ollama、SGLang 等）失败，
报 `400 {"message":"[{'type':'missing','loc':('body',),'msg':'Field required'}]"}`。
默认的 `httpVersion = HTTP_2` 会让 JDK HttpClient 在明文 URL 上尝试 h2c upgrade
（`Connection: Upgrade, HTTP2-Settings` / `Upgrade: h2c`）。uvicorn/h11
（vLLM 的服务端栈）会把带 upgrade 标记的请求当作无请求体处理，于是 FastAPI
收到空 body。curl / OkHttp / DashScope 不受影响，因为它们都不会在明文上尝试 h2c。

**修复**：`httpVersion` 改为三态 —

| 配置值         | `http://`                 | `https://`                      |
| -------------- | ------------------------- | ------------------------------- |
| `null`（默认） | HTTP/1.1（请求级降级）    | HTTP/2 经 ALPN 协商（行为不变） |
| `HTTP_1_1`     | HTTP/1.1                  | HTTP/1.1                        |
| `HTTP_2`       | HTTP/2（显式 h2c opt-in） | HTTP/2                          |

版本解析在 `buildJdkRequest` 中按请求进行（显式值同样对基于注入 `HttpClient`
构建的 transport 生效）。这与业界事实标准一致（curl / OkHttp / 浏览器：h2 仅
走 TLS）。

**迁移**：仅影响依赖默认明文 h2c 的用户 — 显式设置 `httpVersion(HTTP_2)` 即可
保留原行为。`https://` 用户无感知。

**⚠️注意⚠️**
HttpTransportConfig # getHttpVersion() 公共契约从“默认非空 HTTP_2”变成“默认 null”（null 表示未设置，框架内部做了逻辑默认 Http2 处理），会造成下游历史代码兼容性问题。请下游代码如果使用了这个get方法，在升级版本后进行 NPE 判空处理。


agentscope-core/src/main/java/io/agentscope/core/model/transport/HttpTransportConfig.java
```diff
@@ -205,7 +206,7 @@ public static class Builder {
         private Duration keepAliveDuration = Duration.ofMinutes(5);
         private boolean ignoreSsl = false;
         private ProxyConfig proxyConfig = null;
-        private HttpVersion httpVersion = HttpVersion.HTTP_2;
+        private HttpVersion httpVersion = null;
```
oss-maintainer, line 210: **[Warning] compatibility:** `getHttpVersion()` becomes nullable, which changes a public contract. Inside the monorepo the only consumer is `JdkHttpTransport.buildClient` (null-guarded here), but external code doing `config.getHttpVersion().toJdkHttpVersion()` will now NPE. Consider either (a) calling this out in the changelog / `HttpTransportConfig` class javadoc as a behavior change, or (b) adding a non-null accessor such as `getEffectiveHttpVersion(URI)` and keeping `getHttpVersion()` returning a resolved value.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -415,6 +423,24 @@ private java.net.http.HttpRequest buildJdkRequest(HttpRequest request, boolean i
...
+     * @param configured the configured version, or null for auto
+     * @param uri the request URI
+     * @return the version to set on the request, or null to inherit the client version
+     */
+    static HttpClient.Version resolveRequestVersion(HttpVersion configured, URI uri) {
+        if (configured != null) {
+            return configured.toJdkHttpVersion();
+        }
+        boolean isCleartext = uri == null || !"https".equalsIgnoreCase(uri.getScheme());
+        return isCleartext ? HttpClient.Version.HTTP_1_1 : null;
```
oss-maintainer, line 444: **[Warning] behavior change:** treating every non-`https` scheme as cleartext is right for the reported bug, but it also changes behavior for users talking to h2c-only front ends (some gRPC-web / nginx `listen http2` cleartext setups, and `absoluteURI` request targets where `getScheme()` is null). Those users now silently fall back to HTTP/1.1 with no per-scheme opt-out (the only knob is the global `httpVersion(HTTP_2)`). Suggest noting this explicitly in the PR description migration section and in `HttpVersion` javadoc: "any non-https scheme is treated as cleartext".

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -384,6 +387,11 @@ private java.net.http.HttpRequest buildJdkRequest(HttpRequest request, boolean i
 
         var builder = java.net.http.HttpRequest.newBuilder().uri(uri);
 
+        HttpClient.Version requestVersion = resolveRequestVersion(config.getHttpVersion(), uri);
+        if (requestVersion != null) {
+            builder.version(requestVersion);
+        }
```
oss-maintainer, line 396: **[Info] performance:** when a shared client is built at `HTTP_2` (client-level default, line 120) but individual requests are downgraded to `HTTP/1.1`, the JDK client cannot reuse one connection across versions, so a user hitting both an `http://` and an `https://` endpoint through the same transport keeps two connections per origin. Expected and cheap, but worth a comment near `builder.version(...)` so a future reader does not try to "unify" it.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -415,6 +423,24 @@ private java.net.http.HttpRequest buildJdkRequest(HttpRequest request, boolean i
...
+     * @param uri the request URI
+     * @return the version to set on the request, or null to inherit the client version
+     */
+    static HttpClient.Version resolveRequestVersion(HttpVersion configured, URI uri) {
+        if (configured != null) {
+            return configured.toJdkHttpVersion();
+        }
+        boolean isCleartext = uri == null || !"https".equalsIgnoreCase(uri.getScheme());
+        return isCleartext ? HttpClient.Version.HTTP_1_1 : null;
+    }
```
oss-maintainer, line 445: **[Info] scope:** the same h2c/upgrade pattern is still hard-coded for WebSockets — `JdkWebSocketTransport.buildClient` sets `HttpClient.Version.HTTP_2` unconditionally (`websocket/JdkWebSocketTransport.java:118`). If a cleartext `ws://` endpoint is affected by the same upgrade handling, please file a follow-up so the cleartext policy is consistent across transports; no change needed in this PR.

agentscope-core/src/main/java/io/agentscope/core/model/transport/JdkHttpTransport.java
```diff
@@ -113,7 +113,10 @@ public JdkHttpTransport(HttpClient client, HttpTransportConfig config) {
     private static HttpClient buildClient(HttpTransportConfig config) {
         HttpClient.Builder builder =
                 HttpClient.newBuilder()
-                        .version(config.getHttpVersion().toJdkHttpVersion())
+                        .version(
+                                config.getHttpVersion() != null
+                                        ? config.getHttpVersion().toJdkHttpVersion()
+                                        : HttpClient.Version.HTTP_2)
                         .followRedirects(Redirect.NORMAL)
```
oss-maintainer, line 123: **[Info] docs:** `httpVersion` is only honored by the JDK transport — `OkHttpTransport` ignores it (OkHttp negotiates over ALPN and has no h2c). A one-line note on `HttpTransportConfig.Builder#httpVersion` would prevent "I set HTTP_2 but the OkHttp transport still does X" reports.

agentscope-core/src/test/java/io/agentscope/core/model/transport/JdkHttpTransportTest.java
```diff
@@ -1079,14 +1080,84 @@ void testHttpVersionConfig() {
...
+                        "http://localhost:8080/v1/chat/completions"));
+        assertEquals(
+                Optional.empty(),
+                executeAndCaptureVersion(
+                        HttpTransportConfig.defaults(),
+                        "https://api.example.com/v1/chat/completions"));
+        // explicit HTTP_2 must win on the request even for an injected, non-reconfigurable client
+        assertEquals(
+                Optional.of(HttpClient.Version.HTTP_2),
+                executeAndCaptureVersion(
```
oss-maintainer, line 1140: **[Info] tests:** the new tests capture the version on the built `java.net.http.HttpRequest`, which is the right white-box check for the per-request plumbing, but they cannot see what the JDK client actually puts on the wire (it may still negotiate a different version on an existing connection). Optional follow-up: a small integration test against an in-repo cleartext HTTP server that asserts the version of the accepted request and a non-empty POST body — that pins the regression in #1121 at the protocol level.

oss-maintainer (approved):
### Summary
Fixes a real and painful failure mode: the JDK client's `HTTP_2` default made cleartext `http://` calls send an `h2c` upgrade, which uvicorn/h11 stacks (vLLM, SGLang, Ollama) answer with an empty body. Resolving the version per request in `buildJdkRequest` (so it also applies to injected, non-reconfigurable `HttpClient`s) is the correct place, `null` = "inherit the client version" keeps https behavior byte-for-byte unchanged, and the explicit `HTTP_2` opt-in preserves the old behavior for anyone who really needs it. CI (linux/windows/mac builds, codecov) is green and CLA is signed. No blocking issue found — inline notes below are non-blocking.

**Verified:** `config.getHttpVersion()` has exactly one consumer inside the repo (`JdkHttpTransport.buildClient`, null-guarded here); `OkHttpTransport` never reads it, so the new nullable default cannot NPE the monorepo. Local `mvn` build/tests were not re-run here (multi-module build exceeds the heartbeat budget) — relying on the green CI runs.

### Findings
- **[Warning]** `HttpTransportConfig.java:210` — `getHttpVersion()` is now nullable; external `config.getHttpVersion().toJdkHttpVersion()` call sites will NPE. Worth a changelog line, a javadoc note on the getter, or a non-null `getEffectiveHttpVersion(URI)` helper.
- **[Warning]** `JdkHttpTransport.java:441` — every non-`https` scheme is treated as cleartext, so h2c-only front ends (and absolute-URI requests, where `getScheme()` is null) silently move to HTTP/1.1 with no per-scheme opt-out. Please state that in the migration notes.
- **[Info]** `JdkHttpTransport.java:393` / `:120` — mixed per-request versions prevent connection reuse across versions on a shared client; and `httpVersion` is JDK-transport-only (OkHttp ignores it). A comment/javadoc line each would save future readers and support tickets.
- **[Info]** `JdkWebSocketTransport` still hard-codes `Version.HTTP_2` — consider a follow-up so the cleartext policy is consistent across transports.
- **[Info]** tests — the captures pin the request-level plumbing; an end-to-end cleartext server test asserting the negotiated version and non-empty body would lock the #1121 regression down at the protocol level.

### Suggestions
If you want to make the opt-out discoverable without reading javadoc, a tiny helper documents the policy in code:

```java
/** null = auto: HTTP/1.1 on cleartext, client-level version (HTTP/2 via ALPN) on https. */
public HttpVersion getHttpVersion() { ... }

/** Convenience for callers that must not deal with null. */
public HttpClient.Version resolveVersion(URI uri) {
    return httpVersion != null ? httpVersion.toJdkHttpVersion()
         : ("https".equalsIgnoreCase(uri.getScheme()) ? HttpClient.Version.HTTP_2
                                                       : HttpClient.Version.HTTP_1_1);
}
```

Thanks for the thorough root-cause write-up and the test coverage — that made the review easy.

---

agentscope-core/src/main/java/io/agentscope/core/model/transport/HttpTransportConfig.java
xzxiaoshan: javadoc 已注明返回 null（review 引的是改前代码）。不加 helper 是因为：解析需要 URI 的 scheme，且对注入 HttpClient 的场景会给出错误答案（实际继承注入 client 的版本）。库内调用已全部判空，遵循 getProxyConfig() 先例。

agentscope-core/src/main/java/io/agentscope/core/model/transport/HttpTransportConfig.java
```diff
@@ -205,7 +206,7 @@ public static class Builder {
...
```
jujn, line 209: HttpTransportConfig # getHttpVersion() 公共契约从“默认非空 HTTP_2”变成“默认 null”，会造成下游兼容性问题。默认配置会直接 NPE

jujn (changes requested):

agentscope-core/src/main/java/io/agentscope/core/model/transport/HttpTransportConfig.java
xzxiaoshan: 这个我知道，似乎没有更合适的处理方式了。是不是可以在PR内容或者相关发布说明中强调一下。

oss-maintainer, line 209: Re-posting on the new HEAD to add evidence for @jujn's open compatibility concern rather than repeat it: switching the builder default from `HTTP_2` to `null` makes the *absence* of a configured version observable only through a nullable getter. Inside this repo the only production reader is already guarded (`JdkHttpTransport#buildClient` checks `config.getHttpVersion() != null`), so nothing breaks in-tree — but any downstream code that did `config.getHttpVersion().toJdkHttpVersion()` (the documented, previously non-null contract of a public config class) now gets an NPE, and the change is silent for them because the type didn't change.

Two ways to keep the tri-state without overloading `null` on a public getter:

1. Add an explicit `HttpVersion.AUTO` constant and make it the builder default. `resolveEffectiveHttpVersion` already switches on the enum, so `case AUTO -> scheme == "https" ? HTTP_2 : HTTP_1_1` slots straight in, `getHttpVersion()` stays non-null, and the enum switch stays self-documenting.
2. If you'd rather not widen the public enum, change the getter to `Optional<HttpVersion>` (or add `isHttpVersionConfigured()`); both force a compile error at affected call sites instead of a runtime NPE.

Option 1 is my preference — it also removes the need for callers to know that `null` means "auto by scheme", which is currently only discoverable from the Javadoc.

oss-maintainer (commented):
### Summary
Re-reviewed the delta since my previous approval (`7007829` → `e652584`): one commit that documents the tri-state `httpVersion` semantics (scheme-based auto resolution, `HTTP_2` on cleartext = explicit h2c opt-in) and adds `testCleartextRequestAvoidsH2cUpgradeOnTheWire`, which is the right test to have — it asserts on the wire (`Upgrade` / `Connection: HTTP2-Settings` absent, body intact) instead of on the negotiated `HttpRequest.Version`, so it would actually catch issue #1121's symptom rather than a proxy for it, and the contrast half pins that explicit `HTTP_2` still attempts the upgrade. The code changes themselves are comment-only, so nothing regressed since the approval.

Not re-approving: @jujn's `CHANGES_REQUESTED` on this HEAD is still open, and it is a legitimate public-contract question about making `HttpTransportConfig#getHttpVersion()` nullable. Left one inline comment with two concrete options to resolve it; the second (adding `HttpVersion.AUTO`) also keeps the switch in `resolveEffectiveHttpVersion` exhaustive by construction.

### Checklist
- Delta is doc + test only: no behavior change since the approved commit.
- New wire-level test covers both the default (no h2c upgrade) and the explicit `HTTP_2` opt-in.
- Open: nullable `getHttpVersion()` on a public config class — see inline comment / @jujn's review.
- CI on this HEAD: license/cla, build (ubuntu/windows), spotless all green.

jujn (approved):

PR #3215, opened 2026-09-20
feat(models): expose detailed chat usage token metrics

### LINKED ISSUE #1264: [Feature]: Add reasoningTokens and cachedTokens to ChatUsage

ChatUsage currently exposes only inputTokens and outputTokens. For reasoning models, this is not enough for cost/observability analysis, because reasoning tokens and cached tokens are also important.

Add optional fields to ChatUsage, such as:
- reasoningTokens (or thinkingTokens)
- cachedTokens

jujn:
Closes #1264.

## Summary

- Extend `ChatUsage` with normalized token breakdowns:
  - `cachedTokens`
  - `cacheCreationTokens`
  - `reasoningTokens`
  - `toolUsePromptTokens`
- Propagate these fields through model parsers, streaming accumulation, structured-output aggregation, AG-UI token usage events, and OpenTelemetry tracing.
- Keep `inputTokens` and `outputTokens` as provider totals; all new breakdown fields are subsets of those totals and default to `0` when not reported.

## Provider references

| Provider | Reference | Mapped fields |
|---|---|---|
| OpenAI | [OpenAI API reference](https://platform.openai.com/docs/api-reference/chat/object) | `prompt_tokens_details.cached_tokens`, `prompt_tokens_details.cache_write_tokens`, `completion_tokens_details.reasoning_tokens` |
| DeepSeek | [DeepSeek API docs](https://api-docs.deepseek.com/api/create-chat-completion) | `prompt_cache_hit_tokens` |
| Anthropic | Anthropic Java SDK 2.52.0 | `cache_read_input_tokens`, `cache_creation_input_tokens`, `output_tokens_details.thinking_tokens` |
| Gemini | Google GenAI Java SDK 1.45.0 | `cachedContentTokenCount`, `thoughtsTokenCount`, `toolUsePromptTokenCount` |
| DashScope | [Model Studio docs](https://www.alibabacloud.com/help/zh/model-studio/qwen-api-via-dashscope) | `prompt_tokens_details.cached_tokens`, `prompt_tokens_details.cache_creation_input_tokens`, `output_tokens_details.reasoning_tokens` |
| Ollama | [Ollama API](https://github.com/ollama/ollama/blob/main/docs/api.md) | `prompt_eval_cached_count` |

## Observability

- Core and Studio tracing emit the OpenTelemetry GenAI 1.41 usage attributes for cache read, cache creation, and reasoning tokens.
- `toolUsePromptTokens` uses the AgentScope-specific attribute `agentscope.usage.tool_use_prompt_tokens` because the GenAI semantic convention does not define this field yet.

## Notes

- Anthropic Java SDK is upgraded from `2.14.0` to `2.52.0` to support `outputTokensDetails`.
- Tests cover provider usage parsing, JSON round trips, streaming aggregation, and OTel attribute extraction.


agentscope-core/src/main/java/io/agentscope/core/tracing/OtelTracingMiddleware.java
```diff
@@ -338,8 +354,16 @@ private void setToolCallIds(Span span, Set<String> callIds) {
...
-            span.setAttribute("gen_ai.usage.input_tokens", (long) usage.getInputTokens());
-            span.setAttribute("gen_ai.usage.output_tokens", (long) usage.getOutputTokens());
+            span.setAttribute(GEN_AI_USAGE_INPUT_TOKENS, (long) usage.getInputTokens());
+            span.setAttribute(GEN_AI_USAGE_OUTPUT_TOKENS, (long) usage.getOutputTokens());
+            span.setAttribute(
+                    GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS,
+                    (long) usage.getCacheCreationTokens());
+            span.setAttribute(GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS, (long) usage.getCachedTokens());
+            span.setAttribute(
+                    GEN_AI_USAGE_REASONING_OUTPUT_TOKENS, (long) usage.getReasoningTokens());
```
oss-maintainer, line 364: The new breakdown attributes are written unconditionally, so every model-call span now carries `gen_ai.usage.cache_read.input_tokens=0`, `gen_ai.usage.cache_creation.input_tokens=0`, `gen_ai.usage.reasoning.output_tokens=0` and `agentscope.usage.tool_use_prompt_tokens=0` even for providers that never report those fields. Since the Javadoc here defines `0` as "provider did not report", downstream dashboards/alerts cannot distinguish "no cache hit" from "metric unsupported", and `avg(reasoning_tokens)` over a mixed-provider fleet will be silently pulled toward 0. Consider only setting the attribute when the value is `> 0` (OTel attributes are optional, absence already means "not reported"), or adding a separate `reported` flag to `ChatUsage` if the distinction matters for tracing.

agentscope-core/src/main/java/io/agentscope/core/tracing/OtelTracingMiddleware.java
```diff
@@ -73,6 +78,17 @@ public class OtelTracingMiddleware implements MiddlewareBase {
...
 
+    private static final String GEN_AI_USAGE_INPUT_TOKENS = "gen_ai.usage.input_tokens";
+    private static final String GEN_AI_USAGE_OUTPUT_TOKENS = "gen_ai.usage.output_tokens";
+    private static final String GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS =
+            "gen_ai.usage.cache_creation.input_tokens";
+    private static final String GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS =
+            "gen_ai.usage.cache_read.input_tokens";
+    private static final String GEN_AI_USAGE_REASONING_OUTPUT_TOKENS =
+            "gen_ai.usage.reasoning.output_tokens";
+    private static final String AGENTSCOPE_USAGE_TOOL_USE_PROMPT_TOKENS =
```
oss-maintainer, line 89: Nice catch fixing the latent NPE here — the Studio `AttributesExtractors.getLLMResponseAttributes()` path used to dereference `response.getUsage()` unguarded, and the streaming aggregator can hand it a `ChatResponse` with `usage == null`, so the whole attribute extraction blew up before. The new `usage != null` guard makes the same treatment consistent in both tracing paths.

agentscope-dependencies-bom/pom.xml
```diff
@@ -79,7 +79,7 @@
         <dashscope-sdk-java.version>2.22.9</dashscope-sdk-java.version>
         <bailian.version>2.8.2</bailian.version>
         <openai-java.version>4.52.0</openai-java.version>
-        <anthropic.version>2.14.0</anthropic.version>
+        <anthropic.version>2.52.0</anthropic.version>
```
oss-maintainer, line 82: This is a 38-minor-version jump of `com.anthropic:anthropic-java` (2.14.0 -> 2.52.0) riding along inside a feature PR, and it lands in the BOM, so every downstream consumer of `agentscope-dependencies-bom` picks it up whether or not they use the Anthropic extension. Two asks: (1) split the bump into its own PR (or at least its own commit with a note in the release notes) so it can be reverted/bisected independently of the usage metrics; (2) confirm the streaming deltas still parse — `message_start` / `message_delta` usage handling is the riskiest area across that range, and the added tests here mock `outputTokensDetails()` but there is no test pinning behaviour for responses where the field is genuinely absent. Note this also supersedes the open dependabot PRs for the same dependency — worth linking/closing them so the bump history stays clear.

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-anthropic/src/main/java/io/agentscope/extensions/model/anthropic/formatter/AnthropicResponseParser.java
```diff
@@ -228,17 +237,26 @@ private static ChatResponse parseStreamEvent(
...
                     deltaUsage.inputTokens().isPresent()
-                            ? deltaUsage.inputTokens().get()
-                                    + cacheReadTokens
-                                    + deltaUsage.cacheCreationInputTokens().orElse(0L)
+                            ? deltaUsage.inputTokens().get() + cacheReadTokens + cacheCreationTokens
                             : usageState.inputTokens;
+            long reasoningTokens =
+                    deltaUsage
+                            .outputTokensDetails()
+                            .map(details -> details.thinkingTokens())
```
oss-maintainer, line 251: `cacheCreationInputTokens` is read here with `.orElse(usageState.cacheCreationTokens)` — i.e. the value captured at `message_start` is preserved if the `message_delta` payload omits it. `thinkingTokens` has no equivalent fallback, so if a deployment/gateway reports `output_tokens_details` on `message_start` but omits it on `message_delta` (the deltas do not always carry the details object), reasoning tokens collapse to 0 in the final usage. Suggest adding `int reasoningTokens` to `StreamUsageState`, populating it in the `message_start` branch next to `cacheCreationTokens`, and using `deltaUsage.outputTokensDetails().map(...).orElse((long) usageState.reasoningTokens)` so both fields behave the same way.

agentscope-core/src/main/java/io/agentscope/core/ReActAgent.java
```diff
@@ -1465,6 +1468,9 @@ private ChatUsage collectAggregatedUsage(AgentState agentState) {
                     totalInput += usage.getInputTokens();
                     totalOutput += usage.getOutputTokens();
                     totalCached += usage.getCachedTokens();
+                    totalCacheCreation += usage.getCacheCreationTokens();
```
oss-maintainer, line 1471: Aggregation test gap: the per-provider parser changes are well covered, but nothing here asserts that `collectAggregatedUsage()` sums the three new breakdowns across multiple model calls in one agent run. That is the place where a subset-of-total mistake would be most visible (a multi-step ReAct loop accumulates one `Msg` with usage per iteration), and the arithmetic is easy to pin with a small unit test: two messages with (100/20, cached 40, cacheCreation 8, reasoning 12, toolUsePrompt 5) and (50/10, cached 20, ...) -> totals 24/150/... Please add one, so a future refactor of this loop cannot silently drop a field the way `cacheCreation`/`reasoning`/`toolUsePrompt` would have been dropped before this PR.

oss-maintainer (commented):
### Summary
Extends `ChatUsage` with normalized token breakdowns (`cacheCreationTokens`, `reasoningTokens`, `toolUsePromptTokens`), wires them through the core/AG-UI/OTel paths and all five model providers, and fixes the `Msg` metadata round-trip that previously dropped `cachedTokens`. The data model is well thought out — the "subset of input/output tokens, 0 when unreported" contract is documented consistently on both the getters and the builder setters, JSON compat is preserved via the delegating 4-arg constructor, and the added `usage != null` guard in Studio's `AttributesExtractors` fixes a real latent NPE. Not blocking, but three items are worth resolving before merge.

### Findings
- **[Warning]** `OtelTracingMiddleware.java:364` — breakdown attributes are emitted unconditionally, so `0` now means both "not reported" and "genuinely zero"; spans lose that distinction for downstream dashboards.
- **[Warning]** `pom.xml:82` — `anthropic-java` 2.14.0 → 2.52.0 in the BOM inside a feature PR; split it out and confirm streaming usage parsing, and fold in the open dependabot PRs for the same dependency.
- **[Warning]** `AnthropicResponseParser.java:251` — `cacheCreationInputTokens` falls back to the value captured at `message_start`, `thinkingTokens` does not, so reasoning tokens can collapse to 0 on the `message_delta` path.
- **[Warning]** `GeminiResponseParser.java:103` — agrees with the Copilot finding: dropping the `totalTokenCount` fallback under-reports output tokens for blocked/partial responses.
- **[Info]** `ReActAgent.java:1471` — no test pins that `collectAggregatedUsage()` sums the three new breakdowns across a multi-step run.

### Suggestions
1. Guard the new OTel attributes with `> 0` (or introduce an explicit "reported" flag on `ChatUsage` if the zero/not-reported distinction is needed downstream).
2. Add `reasoningTokens` to `StreamUsageState` so the Anthropic streaming path treats it exactly like `cacheCreationTokens`.
3. Add a two-message aggregation test for `collectAggregatedUsage()` covering all six fields.

### Compatibility
`ChatUsage` gains fields only and keeps both old constructors, so binary and JSON compatibility hold; note however that `@JsonCreator` moves to the 7-arg constructor, so records persisted by an older build deserialize with the new breakdowns defaulted to `0` — worth one line in the release notes for anyone charting historical usage.

---

agentscope-extensions/agentscope-extensions-model/agentscope-extensions-model-gemini/src/main/java/io/agentscope/extensions/model/gemini/formatter/GeminiResponseParser.java
```diff
@@ -99,27 +99,16 @@ public ChatResponse parseResponse(GenerateContentResponse response, Instant star
...
-                                                                total ->
-                                                                        Math.max(
-                                                                                0,
-                                                                                total
-                                                                                        - inputTokens))
-                                                        // Without candidate or total counts,
-                                                        // thoughts are the only reported output.
-                                                        .orElse(thinkingTokens));
+                int toolUsePromptTokens = metadata.toolUsePromptTokenCount().orElse(0);
+                int outputTokens = metadata.candidatesTokenCount().orElse(0) + thinkingTokens;
```
oss-maintainer, line 103: +1 on the Copilot finding about removing the `totalTokenCount` fallback here. To be concrete about the blast radius: `outputTokens` for Gemini now reports only `candidatesTokenCount + thoughtsTokenCount`, so a response where `candidatesTokenCount` is absent (blocked/partial responses, some safety-truncated streams) reports ~0 output tokens while `getTotalTokens()` (= input + output) also collapses — cost accounting for exactly the requests that are most worth billing. If the previous derivation was removed on purpose because it double-counted thinking tokens, a `Math.max(candidateTokens + thinkingTokens, totalTokenCount - inputTokens)` style guard would keep the safety net without the over-count; otherwise please restore the `orElseGet` chain.

Aias00 (approved):

